PR #83888, opened 2020-03-31
Add BLAS/LAPACK switching mechanism

matthewbauer:
###### Motivation for this change

This is based on previous work for switching between BLAS and LAPACK
implementation in Debian[[1]] and Gentoo[[2]]. The goal is to have one way
to depend on the BLAS/LAPACK libraries that all packages must use. The
attrs “blas” and “lapack” are used to represent a wrapped BLAS/LAPACK
provider. Derivations that don’t care how BLAS and LAPACK are
implemented can just use blas and lapack directly. If you do care what
you get (perhaps for some CPP), you should verify that blas and lapack
match what you expect with an assertion.

The “blas” package collides with the old “blas” reference
implementation. This has been renamed to “blas-reference”. In
addition, “lapack-reference” is also included, corresponding to
“liblapack” from Netlib.org.

Currently, there are 3 providers of the BLAS and LAPACK interfaces:

- lapack-reference: the BLAS/LAPACK implementation maintained by netlib.org
- OpenBLAS: an optimized version of BLAS and LAPACK
- MKL: Intel’s unfree but highly optimized BLAS/LAPACK implementation

By default, the above implementations all use the “LP64” BLAS and
LAPACK ABI. This corresponds to “openblasCompat” and is the safest way
to use BLAS/LAPACK. You may received some benefits from “ILP64” or
8-byte integer BLAS at the expense of breaking compatibility with some
packages.

This can be switched at build time with an override like:

    import <nixpkgs> {
        config.allowUnfree = true;
        overlays = [(self: super: {
          lapack = super.lapack.override {
            lapackProvider = super.lapack-reference;
          };
          blas = super.blas.override {
            blasProvider = super.lapack-reference;
          };
        })];
      }

or, switched at runtime via LD_LIBRARY_PATH like:

    $ LD_LIBRARY_PATH=$(nix-build -E '(with import <nixpkgs> {}).lapack.override { lapackProvider = pkgs.mkl; is64bit = true; })')/lib:$(nix-build -E '(with import <nixpkgs> {}).blas.override { blasProvider = pkgs.mkl; is64bit = true; })')/lib ./your-blas-linked-binary

By default, we use OpenBLAS LP64 also known in Nixpkgs as
openblasCompat.

[1]: https://wiki.debian.org/DebianScience/LinearAlgebraLibraries
[2]: https://wiki.gentoo.org/wiki/Blas-lapack-switch

###### Things done

[...]

PR #301011, opened 2024-04-02
auditBlasHook: remove

philiptaron:
## Description of changes

This was added in https://github.com/NixOS/nixpkgs/commit/d83e9c95738df2e9f912c239201f50e53e40810b (part of #83888), but included substitution variables `@blas@` and `@lapack@` that were never substituted. It's not used anywhere in `nixpkgs`.

## Things done

- [x] `nixpkgs-review rev HEAD` (nothing rebuilt)


ghost (approved):

SuperSandro2000:
Or should we fix it instead? Is this hook even useful?

philiptaron:
> Or should we fix it instead? Is this hook even useful?

# 🤷🏻‍♂️ 
Not having any users in four-ish years is a pretty good indication that we can live without it, and it'll always be in the source history if someone figures out how to make it useful (ideally, with docs and worked examples in the nixpkgs tree.)

I'm not the one to do that work, though; I'm just the janitor in this PR.

a-n-n-a-l-e-e:
removing it seems fine. i think `pkgs/top-level/release-alternatives.nix` can go too, that looks like it was pre blas/lapack framework.

PR #301039, opened 2024-04-02
lib: introduce types.hexStr

Janik-Haag:
## Description of changes

This change introduces lib.types.hex, I thought that it's a common enough numer that it's worth the change.
I decided against adding it to other lib functions like lib.types.number, so the interface keeps stable.

not quite sure if there is anything else that needs to be done before this can get merged, or if there are any tests for lib.types I should be aware of.

I also tested if everything is working with the small repl example below
```nix
nix-repl> pkgs = import ./. { }

nix-repl> lib = pkgs.lib                                                                                                                                                                                            
                                                                                                                                                                                                                    
nix-repl> (lib.evalModules { modules = [ { options = { test = lib.mkOption { default = "a"; type = lib.types.hex; }; }; config = { }; } ]; }).config                                                                
{                                                                                                                                                                                                                   
  test = "a";                                                                                                                                                                                                       
}                                                                                                                                                                                                                   
                                                                                                                                                                                                                    
nix-repl> (lib.evalModules { modules = [ { options = { test = lib.mkOption { default = ""; type = lib.types.hex; }; }; config = { }; } ]; }).config                                                                 
{                                                                                                                                                                                                                   
  test = «error: A definition for option `test' is not of type `hexadecimal number'. Definition values:                                                                                                             
- In `<unknown-file>': ""»;                                                                                                                                                                                         
}
```

<!--
For package updates please link to a changelog or describe changes, this helps your fellow maintainers discover breaking updates.
For new packages please briefly describe the package or provide a link to its homepage.
-->

## Things done

<!-- Please check what applies. Note that these are not hard requirements but merely serve as information for reviewers. -->

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [X] Tested, as applicable:

<!--
To help with the large amounts of pull requests, we would appreciate your
reviews of other pull requests, especially simple package updates. Just leave a
comment describing what you have tested in the relevant package/service.
Reviewing helps to reduce the average time-to-merge for everyone.
Thanks a lot if you do!

List of open PRs: https://github.com/NixOS/nixpkgs/pulls
Reviewing guidelines: https://github.com/NixOS/nixpkgs/blob/master/pkgs/README.md#reviewing-contributions
-->

---


infinisil:
Please also add docs to https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/development/option-types.section.md and tests to https://github.com/NixOS/nixpkgs/blob/master/lib/tests/modules.sh

Janik-Haag:
@infinisil I added tests and updated the docs, as requested.

drupol (approved):
LGTM

lib/types.nix
```diff
@@ -373,6 +373,14 @@ rec {
       merge = mergeEqualOption;
     };
 
+    hex = mkOptionType {
+      name = "hex";
+      description = "hexadecimal number";
```
roberth, line 378: Is it really a number if it's in a string?
This might feel like splitting hairs, but it will save some trial and error when the `description` is all available documentation.

```suggestion
      description = "hexadecimal digits string";
```

This also implicitly excludes the prefixes that indicate the start of a hexadecimal number, such as `0x` or `#`.

lib/types.nix
```diff
@@ -373,6 +373,14 @@ rec {
...
+      descriptionClass = "noun";
+      check = x: str.check x && builtins.match "[0-9a-fA-F]+" x != null;
```
roberth, line 380: Should we allow `0x` and/or `#`?

lib/types.nix
```diff
@@ -373,6 +373,14 @@ rec {
...
+      merge = mergeEqualOption;
```
roberth, line 381: I think it'd be good to canonicalize the notation, ideally before "merging".
We generally use lower case in the Nix ecosystem.
`0x` and `#` could be stripped (if allowed in the first place).
Leading zeroes should probably not be stripped, because the hex may be data instead of a number.

nixos/doc/manual/development/option-types.section.md
```diff
@@ -160,6 +160,10 @@ merging is handled.
     may result in [precision loss](https://github.com/NixOS/nix/issues/5733).
     :::
 
+`types.hex`
+
+:   A hexadecimal number.
```
roberth, line 165: It seems that there's a lot more to say, and maybe move it into the section with string-like types?

roberth (commented):
This looks quite useful, but I don't think it's a number.

lib/types.nix
infinisil: And maybe it should be called `types.hexStr`

lib/types.nix
infinisil: I don't think so, at least for now. Can be added later if necessary

roberth: `either` can convert failures into usable behavior, and functions like `coercedTo` can make a left match vs a right match in `either` observable, so allowing more _is_ a breaking change in non-trivial compositions of types.

lib/types.nix
Janik-Haag: > We generally use lower case in the Nix ecosystem.

I'll remove the matching for uppercase letters

> 0x and # could be stripped (if allowed in the first place).

I wouldn't allow them in the first place, not sure what other people think.

> Leading zeroes should probably not be stripped, because the hex may be data instead of a number.

that's not happening as far as I'm aware.

lib/types.nix
infinisil: I'd imagine there's laws for types we could come up with to guard against this, e.g. for either:

> If `all aType.check aValues`, then
> ```nix
> (either bType aType).merge aValues
> == aType.merge aValues
> == (either aType bType).merge aValues
> ```

lib/types.nix
roberth: > that's not happening as far as I'm aware.

:+1: and that's the right behavior for hex data (as opposed to a number, which wasn't clear before)

lib/types.nix
roberth: Counterexample for the "laws":

```nix
aType = coercedTo hexStr (v: { base16 = v; }) attrs;
bType = coercedTo base64Str (v: { base64 = v; }) attrs;
# counterexample for "laws"
aValues = ["0000"];
```

Specifically,the definition is valid for both types, swapping the either runs a different coercion ie the inequalityj
```nix
(either bType aType).merge aValues
  != (either aType bType).merge aValues
```

Counterexample for allowing prefix later, reusing the types:

```nix
values = ["0x00"];

```


This causes `(either aType bType).merge` to return `{ base16 = ... }` instead the current `{ base64 = ... }` attrset.

infinisil: Oh yeah, I guess I haven't explained it fully, but that's exactly the intended point of this, because it would allow us to reason about which types should be allowed as `either` arguments and which shouldn't. And indeed your example sounds like something that shouldn't be allowed, so it seems to work. In practice we might introduce a typeclass/interface/trait hierarchy for the types and requiring `either` arguments to be of a certain typeclass/interface/trait. And given this requirement on either's argument, the law should hold.

roberth: Laws about types are useless if they don't apply to all of them. I don't see use getting rid of `either` or `coercedTo` anytime soon.

Classifying types might be useful, but it doesn't change the behavior, and it doesn't make the discussed change a non-breaking one, except for a weaker definition of non-breaking using some of that new terminology.

It'd be good to have some theory to refer to in discussions like this, but we have much lower hanging documentation fruit to prioritize.

infinisil: Yeah sure, just felt like doing a bit of an exploration, let's not spend more cycles on this :)

(and I don't think this is blocking for this PR in any way)

roberth: I agree. Hex data that's not necessarily a number (this type) is rarely prefixed anyway, so it looks like a change won't be necessary. No change implies no breaking change.

Resolving.

nixos/doc/manual/development/option-types.section.md
Janik-Haag: Added more info, let me know what you think :D

nixos/doc/manual/development/option-types.section.md
```diff
@@ -202,6 +202,12 @@ merging is handled.
 
 :   A string. Multiple definitions are concatenated with a colon `":"`.
 
+`types.hexStr`
+
+:   A hexadecimal value represented as string. The type only allows lowercase
+    values. The type also does not allow for hex prefixes like `0x` or `#`.
+
```
roberth, line 209:
```suggestion
    Leading zeroes are not stripped.
```

lib/types.nix
```diff
@@ -373,6 +373,14 @@ rec {
       merge = mergeEqualOption;
     };
 
+    hexStr = mkOptionType {
+      name = "hexStr";
+      description = "hexadecimal value represented as string";
```
roberth, line 378: I think the hexadecimal is still just a representation of something, presumably bytes.
```suggestion
      description = "bytes represented as a string of hexadecimals";
```

That reminds me, shouldn't we check that it's an even number of hex digits?
Half bytes would be rare. The only case I can think of is short 12-bit hex color codes of three hex digits, but that should just be a distinct type (and probably a type that also allows 24-bit color codes, and a `#`).

```suggestion
      description = "string of hexadecimal bytes";
```

Maybe that's a bit vague. For a more syntax oriented description:

```suggestion
      description = "string of hexadecimal pairs";
```

Or if I'm wrong about half bytes:

```suggestion
      description = "string of hexadecimal digits";
```
```suggestion
      description = "string of hexadecimals";
```

nixos/doc/manual/development/option-types.section.md
```diff
@@ -202,6 +202,12 @@ merging is handled.
 
 :   A string. Multiple definitions are concatenated with a colon `":"`.
 
+`types.hexStr`
+
+:   A hexadecimal value represented as string. The type only allows lowercase
```
roberth, line 207:
```suggestion
:   A sequence of bytes represented as a string of hexadecimals. The type only allows lowercase
```

roberth (commented):
I think we can be a bit more clear about what's represented.
Half bytes seem like a mistake we should catch.
Sorry to drag this out, but it's important for types to have a clear meaning and to be precise about it.
Types are of such a nature that all the details will be depended on rather soon, and not just here in the `nixpkgs` repo. I'm not trying to be obsessive for no good reason :)

nixos/doc/manual/development/option-types.section.md
Janik-Haag: >  Half bytes seem like a mistake we should catch.

Uh, I think we might be talking past each other. I this is a string of hex (as in characters representing hex numbers) and not a byte sequence of hex data. So you don't really have half bytes :eyes:

roberth: Yes, I did make somewhat of a leap, and I'm sorry if my presentation was messy. I'm assuming half bytes are not ok, and I think the type should check that.

Do you have a use case where the string of hexadecimal digits does not represent bytes?

infinisil: I don't think it would be beneficial to talk about half-bytes here. Just saying hexadecimal sounds good

roberth: I agree half bytes are unnecessarily awkward, and I suspect they should not be allowed. If we don't allow them, we can call this a sequence of bytes in hexadecimal; no silly language, and no silly values.
If we do have a use case for odd numbers of hex digits *with* leading zeroes, then something like "hexadecimal string" sounds good to me.

roberth: Hmm, maybe I wasn't communicating well, because I was trying to comment about the implementation, but my suggestions are on lines about documentation.
I like to use file comments because they become threads, but it seems that I need to take into account better where I put those comments.

infinisil: I'm also only getting it now: You're hinting at wanting a check to prevent against an odd number of hex digits (which would represent half-bytes), which is most likely just a bug. I've personally never encountered any programs asking for an odd number of hex digits (because indeed, hex digits are usually used to represent bytes).

I think it's fair to restrict this type to an even number of hex digits to prevent such mistakes. We can still allow an odd number of digits later if this turns out that there are use cases for an odd number of digits. Because of that though I wouldn't want this to be embedded into the documentation. So I think something like this should be good:

```suggestion
:   A string of hexadecimal digits. The type only allows lowercase
```

lib/types.nix
infinisil: In line with https://github.com/NixOS/nixpkgs/pull/301039#discussion_r1575343406:

```suggestion
      description = "string of hexadecimal digits";
```

lib/types.nix
```diff
@@ -373,6 +373,14 @@ rec {
...
+      descriptionClass = "noun";
+      check = x: str.check x && builtins.match "[0-9a-f]+" x != null;
```
infinisil, line 380: In line with https://github.com/NixOS/nixpkgs/pull/301039#discussion_r1575343406, something like this:

```suggestion
      check = x:
        str.check x
        && builtins.match "[0-9a-f]+" x != null
        # Usually hex digits are used to represent bytes.
        # A single hex digit only represents half a byte,
        # so if we have an odd number, it's likely a mistake.
        && mod (stringLength x) 2 == 0;
```

Writing this, I'm actually not sure if it's worth doing this, the error message when you get the type wrong is really not very helpful in such a case.

roberth: Right, this only outputs a bool.
What if we check it in the merge function instead, like the other, good errors?
Technically that breaks `either hexStr x` for odd-length definitions, but that is probably ok in practice.

Perhaps the module system could invoke the merge function instead of producing a generic error message? Or falling back to the generic message if the merge evaluates (`if ! t.check v then seq merged (throw genericMsg) else merged`).
(Just an idea, not a request)

infinisil: I think the proper solution is to extend `.check` to allow custom error messages.

To move forward with this PR though, I don't really care about having this check or not. I'd like to leave this decision up to @Janik-Haag.

Janik-Haag:
I quit nixpkgs and thus won't be finishing this.

roberth:
All the best to you Janik, and thank you for all your contributions.

PR #301063, opened 2024-04-02
mdadm: update homepage

ncfavier:
## Description of changes

~~Fixes the failure in https://hydra.nixos.org/build/254811279 introduced by 994232c03412:~~

```
machine: must succeed: mdadm --create --force /dev/md0 --metadata 1.2 --level=raid1 --raid-devices=2 /dev/vda5 /dev/vda6
machine # mdadm: Cannot block udev, error creating blocking file.
machine # mdadm: No such file or directory: /run/mdadm/creating-md0
```

~~Somehow mdadm 4.3 no longer creates `/run/mdadm` if not present. Bug?~~

EDIT: superseded by https://github.com/NixOS/nixpkgs/pull/301081

## Things done

- Built on platform(s)
  - [X] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
- For non-Linux: Is sandboxing enabled in `nix.conf`? (See [Nix manual](https://nixos.org/manual/nix/stable/command-ref/conf-file.html))
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:
  - [NixOS test(s)](https://nixos.org/manual/nixos/unstable/index.html#sec-nixos-tests) (look inside [nixos/tests](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests))
  - and/or [package tests](https://nixos.org/manual/nixpkgs/unstable/#sec-package-tests)
  - or, for functions and "core" functionality, tests in [lib/tests](https://github.com/NixOS/nixpkgs/blob/master/lib/tests) or [pkgs/test](https://github.com/NixOS/nixpkgs/blob/master/pkgs/test)
  - made sure NixOS tests are [linked](https://nixos.org/manual/nixpkgs/unstable/#ssec-nixos-tests-linking) to the relevant packages
- [ ] Tested compilation of all packages that depend on this change using `nix-shell -p nixpkgs-review --run "nixpkgs-review rev HEAD"`. Note: all changes have to be committed, also see [nixpkgs-review usage](https://github.com/Mic92/nixpkgs-review#usage)
- [ ] Tested basic functionality of all binary files (usually in `./result/bin/`)
[...]

PR #301081, opened 2024-04-02
mdadm: Fix hardcoded directory

amarshall:
In [upstream change][1], handling of creating `/run/mdadm/creating-%s` file was changed to fail if it was unable to create the file. This revealed that the path itself incorrectly hardcodes `/run/mdadm` instead of using `MAP_DIR` or similar. Since nixpkgs sets `RUN_DIR=/run/.mdadm` at compile time, and `MAP_DIR=$(RUN_DIR)` in upstream, the `/run/mdadm` is never created. This remedies by fixing the hard-coded directory.

[1]: https://git.kernel.org/pub/scm/utils/mdadm/mdadm.git/commit/?id=9f376da6439b07dc93ae084ab576e133b9d8d839

See also https://github.com/NixOS/nixpkgs/pull/301063 and discussion there.

Fixes the failure in https://hydra.nixos.org/build/254811279 introduced by https://github.com/NixOS/nixpkgs/commit/994232c0341269b4eda21707b781ad6a0e28af96

## Things done

`nix-build -A nixosTests.installer.swraid`

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [x] Tested, as applicable:

---

### DIFF
--- a/pkgs/os-specific/linux/mdadm/default.nix
+++ b/pkgs/os-specific/linux/mdadm/default.nix
@@ -9,7 +9,10 @@ stdenv.mkDerivation rec {
     sha256 = "sha256-QWcnrh8QgOpuMJDOo23QdoJvw2kVHjarc2VXupIZb58=";
   };
 
-  patches = [ ./no-self-references.patch ];
+  patches = [
+    ./no-self-references.patch
+    ./fix-hardcoded-mapdir.patch
+  ];
 
   makeFlags = [
     "NIXOS=1" "INSTALL=install" "BINDIR=$(out)/sbin"
--- a/pkgs/os-specific/linux/mdadm/default.nix
+++ b/pkgs/os-specific/linux/mdadm/default.nix
@@ -46,7 +46,7 @@ stdenv.mkDerivation rec {
 
   meta = with lib; {
     description = "Programs for managing RAID arrays under Linux";
-    homepage = "http://neil.brown.name/blog/mdadm";
+    homepage = "https://git.kernel.org/pub/scm/utils/mdadm/mdadm.git";
     license = licenses.gpl2;
     mainProgram = "mdadm";
     maintainers = with maintainers; [ ekleog ];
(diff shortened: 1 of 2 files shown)


pkgs/os-specific/linux/mdadm/default.nix
```diff
@@ -9,7 +9,10 @@ stdenv.mkDerivation rec {
     sha256 = "sha256-QWcnrh8QgOpuMJDOo23QdoJvw2kVHjarc2VXupIZb58=";
   };
 
-  patches = [ ./no-self-references.patch ];
+  patches = [
+    ./no-self-references.patch
```
drupol, line 13: How about adding comments?
So we know if the patch needs to be removed at the next point release. A few words of explanation would be nice.

r-vdp:
Nice! I think you should also submit this patch upstream, if you didn't plan to do so yet.

r-vdp (approved):

pkgs/os-specific/linux/mdadm/fix-hardcoded-mapdir.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/udev.c b/udev.c
+index bc4722b0..aa2a1a24 100644
+--- a/udev.c
++++ b/udev.c
+@@ -167,7 +167,7 @@ enum udev_status udev_block(char *devnm)
+ 	int fd;
+ 	char *path = xcalloc(1, BUFSIZ);
+ 
+-	snprintf(path, BUFSIZ, "/run/mdadm/creating-%s", devnm);
++	snprintf(path, BUFSIZ, "%s/creating-%s", MAP_DIR, devnm);
```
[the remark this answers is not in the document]
vcunat: The C is good.

vcunat:
@ofBorg test installer.swraid

pkgs/os-specific/linux/mdadm/default.nix
vcunat: We can't know _yet_ whether upstream will accept this change or a similar one.

drupol: Fine, but still a comment should be explaining the reason of the patch, don't you think ?

vcunat:
I don't know `mdadm` in particular, but looking through the code, docs, etc. this does seem good &ndash; so let's unblock the `nixos-unstable` channel.  Thanks 🙂

PR #301142, opened 2024-04-03
add systemd-resolved when `boot.initrd.systemd.network` is enabled

mlyxshi:
## Description of changes

Upstream systemd-resolved  part  in https://github.com/ElvishJerricco/stage1-tpm-tailscale.
@ElvishJerricco

#301333

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [x] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/nixos/modules/system/boot/networkd.nix
+++ b/nixos/modules/system/boot/networkd.nix
@@ -2976,6 +2976,7 @@ let
         "systemd-networkd.service"
         "systemd-networkd.socket"
         "systemd-network-generator.service"
+        "systemd-resolved.service"
         "network-online.target"
         "network-pre.target"
         "network.target"
@@ -2985,7 +2986,9 @@ let
         "remote-fs.target"
       ];
       systemd.users.systemd-network = {};
+      systemd.users.systemd-resolve = {};
       systemd.groups.systemd-network = {};
+      systemd.groups.systemd-resolve = {};
 
       systemd.contents."/etc/systemd/networkd.conf" = renderConfig cfg.config;
 
@@ -2995,6 +2998,10 @@ let
       systemd.sockets.systemd-networkd = {
         wantedBy = [ "initrd.target" ];
       };
+      systemd.services.systemd-resolved = {
+        wantedBy = ["initrd.target"];
+        serviceConfig.ExecStartPre = "-+/bin/ln -s /run/systemd/resolve/resolv.conf /etc/resolv.conf";
+      };
 
       systemd.services.systemd-network-generator.wantedBy = [ "sysinit.target" ];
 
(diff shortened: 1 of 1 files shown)


r-vdp:
I think you missed the line that includes the resolved config file in the initrd as well, so that we get the same config as what we have in stage 2.

mlyxshi:
> I think you missed the line that includes the resolved config file in the initrd as well, so that we get the same config as what we have in stage 2.

I actually don't know. I just think the stage 2 related stuff shouldn't be included in stage1. 

https://github.com/NixOS/nixpkgs/blob/219951b495fc2eac67b1456824cc1ec1fd2ee659/nixos/modules/system/boot/systemd/initrd.nix#L387

mlyxshi:
> I think you missed the line that includes the resolved config file in the initrd as well, so that we get the same config as what we have in stage 2.
> 
> https://github.com/ElvishJerricco/stage1-tpm-tailscale/blob/main/initrd-tailscale.nix#L37

For personal project, defining stage1 resolved config file as same as stage 2 is convenient.
However, as you can see in https://github.com/NixOS/nixpkgs/blob/master/nixos/modules/system/boot/networkd.nix#L2990, in initrd `/etc/systemd/networkd.conf` is defined by `boot.initrd.systemd.network.config` not stage 2 config.

Actually, the default `systemd-resolved` configuration is enough for most cases, people really care about resolved configuration should config it via `boot.initrd.systemd.contents"/etc/systemd/resolved.conf".text = ""`
I think we need to implement something like `boot.initrd.systemd.resolved.config` in another pr.

nixos/modules/system/boot/networkd.nix
```diff
@@ -2995,13 +2998,18 @@ let
       systemd.sockets.systemd-networkd = {
         wantedBy = [ "initrd.target" ];
       };
+      systemd.services.systemd-resolved = {
+        wantedBy = ["initrd.target"];
+        serviceConfig.ExecStartPre = "-+/bin/ln -s /run/systemd/resolve/resolv.conf /etc/resolv.conf";
```
ElvishJerricco, line 3003: I think this should be a tmpfiles rule, really.

nixos/modules/system/boot/networkd.nix
```diff
@@ -2995,13 +2998,18 @@ let
       systemd.sockets.systemd-networkd = {
         wantedBy = [ "initrd.target" ];
       };
+      systemd.services.systemd-resolved = {
+        wantedBy = ["initrd.target"];
```
ElvishJerricco, line 3002: The upstream unit file has `WantedBy=sysinit.target`, so we should probably do the same. Same goes for the stage 2 unit, really.

ElvishJerricco (changes requested):
I think this should all go in `nixos/modules/system/boot/resolved.nix` rather than `networkd.nix`

nixos/modules/system/boot/networkd.nix
mlyxshi: I have no idea how to do it properly. It would be better for you to upstream this

PR #297654, opened 2024-03-21
doc: improve fetchers overview, deduplicate readme content, follow doc conventions

DanielSidhion:
## Description of changes

This PR starts my work to update and improve the chapter on fetchers in the Nixpkgs manual. It adds more explanations on the overview of the chapter, moves content from the contributor README into the manual, and deduplicates information.

**NOTE:** this PR started as a change of the fetchers overview AND updating the documentation for `fetchurl`. After a lot of reviewing, we decided to split this PR into two. The PR concerns the description above. The second PR is #300429 and concerns only with updating the documentation for `fetchurl`.

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
- For non-Linux: Is sandboxing enabled in `nix.conf`? (See [Nix manual](https://nixos.org/manual/nix/stable/command-ref/conf-file.html))
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:
  - [NixOS test(s)](https://nixos.org/manual/nixos/unstable/index.html#sec-nixos-tests) (look inside [nixos/tests](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests))
  - and/or [package tests](https://nixos.org/manual/nixpkgs/unstable/#sec-package-tests)
  - or, for functions and "core" functionality, tests in [lib/tests](https://github.com/NixOS/nixpkgs/blob/master/lib/tests) or [pkgs/test](https://github.com/NixOS/nixpkgs/blob/master/pkgs/test)
  - made sure NixOS tests are [linked](https://nixos.org/manual/nixpkgs/unstable/#ssec-nixos-tests-linking) to the relevant packages
- [ ] Tested compilation of all packages that depend on this change using `nix-shell -p nixpkgs-review --run "nixpkgs-review rev HEAD"`. Note: all changes have to be committed, also see [nixpkgs-review usage](https://github.com/Mic92/nixpkgs-review#usage)
- [ ] Tested basic functionality of all binary files (usually in `./result/bin/`)
[...]

PR #301152, opened 2024-04-03
nixos-render-docs: make examples collapsible by default

DanielSidhion:
## Description of changes

During work on #297654 it got a bit annoying IMO to always see examples fully shown by default on the manuals. This is especially annoying when we want to add an example somewhere in the middle of some explanation, when there's no better place to move the example to (and then link to it from the text).

This PR wraps each example under a [`<details>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details) tag which makes every example collapsible, and each one is collapsed by default. It also adds a bit of CSS to make it more usable.

An example of what it looks like on my browser:
20240402T214614
And after expanding an example:
20240402T214621

The CSS is up for discussion. I tried with and without the borders after expanding, but eventually settled on borders because some examples are kinda large, and the borders help determine when an example ends. I could definitely use help improving the visual though.

Things I tested:
- Examples with and without an `id` attached to them works.
- Links to an example still work.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [x] Tested, as applicable:

---

### DIFF
--- a/doc/style.css
+++ b/doc/style.css
@@ -347,6 +347,22 @@ div.appendix div.example {
     margin-top: 1.5em;
 }
 
+div.book div.example details,
+div.appendix div.example details {
+    padding: 5px;
+}
+
+div.book div.example details[open],
+div.appendix div.example details[open] {
+    border: 1px solid #aaa;
+    border-radius: 4px;
+}
+
+div.book div.example details>summary,
+div.appendix div.example details>summary {
+    cursor: pointer;
+}
+
 div.book br.example-break,
 div.appendix br.example-break {
     display: none;
--- a/pkgs/tools/nix/nixos-render-docs/src/nixos_render_docs/html.py
+++ b/pkgs/tools/nix/nixos-render-docs/src/nixos_render_docs/html.py
@@ -219,13 +219,13 @@ class HTMLRenderer(Renderer):
     def example_open(self, token: Token, tokens: Sequence[Token], i: int) -> str:
         if id := cast(str, token.attrs.get('id', '')):
             id = f'id="{escape(id, True)}"' if id else ''
-        return f'<div class="example"><span {id} ></span>'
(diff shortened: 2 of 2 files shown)


doc/style.css
```diff
@@ -347,6 +347,22 @@ div.appendix div.example {
     margin-top: 1.5em;
 }
 
+div.book div.example details,
+div.appendix div.example details {
```
hsjobeki, line 351: Your CSS assumes a very specific HTML structure. If the HTML changes, such as adding another parent div this will break.
Id suggest '.example details' 
which targets all details with a parent of class .example
this would allow other parents as well. making it more resistent to potential future changes.

hsjobeki (commented):
Looks good to me. Seems we have highly specific css.

doc/style.css
DanielSidhion: I prefer to keep the specific CSS here - it's only used with the output of nrd, and we control the HTML structure there. It's expected that changes to nrd will be at least rendered so we can glance over the output, and if the CSS needs to be adapted (such as in this case), then we'll just change the CSS as well.

To explain why I'm taking this stance: this CSS file has a few rules that conflict on certain elements, and we rely on the more specific rule overriding the value of the less specific one.
As an exercise, I tried removing all the `div.book` and `div.appendix` specifiers from all the other rules as well, and some things got messed up.
I prefer keeping the specific rules because that's how the rest of the CSS is laid out, so it avoids subtle things from breaking if someone decides to change some other less-specific rule.

infinisil (approved):
(reviewed in the docs team)

Looking good, this is awesome, let's merge!

PR #300322, opened 2024-03-30
{nixos-container,lsb-release}: use `substitute` rather than `substituteAll`, fix cross-compilation

philiptaron:
## Description of changes

This change fixes cross-compilation of the `lsb-release` and `nixos-container` scripts, all of which used the `stdenv` environment variable `shell` instead of `runtimeShell` string as the shebang... leading to a cross-compilation mess.

There's no other change in produced output. Rationale for using `substitute` instead of `substituteAll` is here:

- https://github.com/NixOS/nixpkgs/issues/237216

Previously:
- https://github.com/NixOS/nixpkgs/pull/300308

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
- [x] Tested, as applicable:
  - [NixOS test(s)](https://nixos.org/manual/nixos/unstable/index.html#sec-nixos-tests) (look inside [nixos/tests](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests))
  - and/or [package tests](https://nixos.org/manual/nixpkgs/unstable/#sec-package-tests)
  - or, for functions and "core" functionality, tests in [lib/tests](https://github.com/NixOS/nixpkgs/blob/master/lib/tests) or [pkgs/test](https://github.com/NixOS/nixpkgs/blob/master/pkgs/test)
  - made sure NixOS tests are [linked](https://nixos.org/manual/nixpkgs/unstable/#ssec-nixos-tests-linking) to the relevant packages
- [ ] Tested compilation of all packages that depend on this change using `nix-shell -p nixpkgs-review --run "nixpkgs-review rev HEAD"`. Note: all changes have to be committed, also see [nixpkgs-review usage](https://github.com/Mic92/nixpkgs-review#usage)
- [x] Tested basic functionality of all binary files (usually in `./result/bin/`)
- [24.05 Release Notes](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2405.section.md) (or backporting [23.05](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2305.section.md) and [23.11](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2311.section.md) Release notes)
  - [ ] (Package updates) Added a release notes entry if the change is major or breaking
  - [ ] (Module updates) Added a release notes entry if the change is significant
[...]

Issue #237216, opened 2023-06-11
`substituteAll` is a terrible function and we need to stop using it / fix it respectively


roberth:
## Issue description

`substituteAll` is a bash function and a Nix function in `pkgs`.

The bash function is bad because
 - it takes variables from the environment, which is too easily affected, and often the definition of those values is through `mkDerivation` arguments, in which case it is not clear that any relation exists between the definition and the usage.
   - example: https://github.com/NixOS/nixpkgs/pull/237040#discussion_r1225384166
 -  the variables tend to be hard to grep. For one, the @ signs are omitted anywhere but in the file
    - this is a `--subst-var` complaint as well

The Nix function is even worse. Nix has semi-structured data, so the expectations are a bit higher. Variables passed to the `pkgs.substituteAll` functions all get substituted into the file, right? ... WRONG! Those variables are poured carelessly into `mkDerivation`, which assigns special meaning to them, and even ignores and replaces some, like `system`.

### Steps to reproduce

- https://github.com/NixOS/nixpkgs/pull/228133#issuecomment-1586226551

## Technical details

`substituteAll` would be useful if it wasn't such a hack. Maybe something could be achieved with structuredAttrs, so that the variables aren't taken from the environment, but from a separate, safe, json file.

There's also this. Not sure if I want to know more about the horrors inside.
 - https://github.com/NixOS/nixpkgs/issues/28086

sternenseemann:
> Maybe something could be achieved with structuredAttrs, so that the variables aren't taken from the environment, but from a separate, safe, json file.

Much simpler than structured attrs is probably just prefixing the env vars or dumping the argument to `substituteAll` into a file via `builtins.toJSON` and `passAsFile`.

AndersonTorres:
This should be still a Bash program?

roberth:
> This should be still a Bash program?

I think it's pure bash because that makes it easy to use as part of bootstrapping.
As for the `pkgs.substituteAll` derivation function, I don't think it has the same requirement, but we might as well reuse what we have.

> > Maybe something could be achieved with structuredAttrs

> prefixing the env vars or dumping the argument to `substituteAll` into a file via `builtins.toJSON` and `passAsFile`.

Or we could use command line arguments. I don't think the limitations of the process env+args memory space apply here because it's pure bash without any `execv`. Testing needed I guess.

illode:
[...]

PR #301350, opened 2024-04-03
replaceVars: init

philiptaron:
Here's an attempt at making [an easy-to-use variant](https://github.com/NixOS/nixpkgs/pull/300322#issuecomment-2033356590) of `pkgs.substitute` that suffers from none of [the pitfalls of `substituteAll`](https://github.com/NixOS/nixpkgs/issues/237216). It's named `replaceVars` since this isn't the unrelated concept of [Nix substitution](https://nixos.org/manual/nix/stable/glossary#gloss-substituter).

## Motivating transformation

From `pkgs/applications/audio/curseradio/default.nix`:

```
  patches = [
    (substituteAll {
      src = ./mpv.patch;
      inherit mpv;
    })
  ];
```

Into something like this, after `nixfmt` has its way:

```
  patches = [
    (replaceVars ./mpv.patch { inherit mpv; })
  ];
```

According to a quick grep, there's around 340 of these throughout the codebase, making up the majority of the roughly 430 uses of `substituteAll`.

Simple cases:
- `rg -A1 --no-heading --no-filename --no-line-number -tnix -F '(substituteAll {' | rg -F 'src = ./' | wc -l`

All cases:
- `rg --no-heading --no-filename --no-line-number -tnix -F 'substituteAll {' | wc -l`

### Goals

1. **Textually shorter than `substituteAll`** for the majority case, where the thing being replaced lives in the repository rather that acquired through a fetcher or part of a more complicated substitution pipeline.
2. **Uses `--replace-fail` exclusively.** I think the ergonomics of normal-looking names in the `.nix` files makes me want to continue the `name` to `@name@` substitution throughout the codebase.
3. **Has a `checkPhase`**: when there are remaining `@...@` tokens in the output, will fail. This would be wrong for the general-purpose `substitute` function but net positive for the patch substitution use case. This is the main reason for wanting `--replace-fail` exclusively.
4. **Has a name that reflects its main use case.** The name I initially chose, `substituteAttrs`, didn't hit this mark. Perhaps `replaceVars` might.
5. **Clean separation of derivation attrs and replacement attrs.** The attrsets for the derivation itself and the attrset for the substitutions cannot be confused. This is the sin that `substituteAll` commits.

## Description of changes

- [x] Introduce `replaceVars` function.
- [x] Test it

## Things done

- [x] Added tests in `pkgs/test/replace-vars/default.nix` (run with `nix-build -A test.replaceVars`)

### DIFF
--- a/pkgs/build-support/replace-vars/default.nix
+++ b/pkgs/build-support/replace-vars/default.nix
@@ -1,0 +1,77 @@
+{ lib, stdenvNoCC }:
+
+/**
+  `replaceVars` is a wrapper around the [bash function `substitute`](https://nixos.org/manual/nixpkgs/stable/#fun-substitute)
+  in the stdenv. It allows for terse replacement of names in the specified path, while checking
+  for common mistakes such as naming a replacement that does nothing or forgetting a variable which
+  needs to be replaced.
+
+  As with the [`--subst-var-by`](https://nixos.org/manual/nixpkgs/stable/#fun-substitute-subst-var-by)
+  flag, names are encoded as `@name@` in the provided file at the provided path.
+
+  Any unmatched variable names in the file at the provided path will cause a build failure.
+
+  Any remaining text that matches `@[A-Za-z_][0-9A-Za-z_'-]@` in the output after replacement
+  has occurred will cause a build failure.
+
+  # Inputs
+
+  `path` ([Store Path](https://nixos.org/manual/nix/latest/store/store-path.html#store-path) String)
+  : The file in which to replace variables.
+
+  `attrs` (AttrsOf String)
+  : Each entry in this set corresponds to a `--subst-var-by` entry in [`substitute`](https://nixos.org/manual/nixpkgs/stable/#fun-substitute).
+
+  # Example
+
+  ```nix
+  { replaceVars }:
+
+  replaceVars ./greeting.txt { world = "hello"; }
+  ```
+
+  See `../../test/replace-vars/default.nix` for tests of this function.
+*/
+path: attrs:
+
+let
+  # We use `--replace-fail` instead of `--subst-var-by` so that if the thing isn't there, we fail.
+  subst-var-by = name: value: [
+    "--replace-fail"
+    (lib.escapeShellArg "@${name}@")
+    (lib.escapeShellArg value)
+  ];
+
+  replacements = lib.concatLists (lib.mapAttrsToList subst-var-by attrs);
+in
+
+stdenvNoCC.mkDerivation {
+  name = baseNameOf (toString path);
+  src = path;
+  doCheck = true;
+  dontUnpack = true;
+  preferLocalBuild = true;
+  allowSubstitutes = false;
+
+  buildPhase = ''
+    runHook preBuild
+    substitute "$src" "$out" ${lib.concatStringsSep " " replacements}
+    runHook postBuild
+  '';
+
+  # Look for Nix identifiers surrounded by `@` that aren't substituted.
+  checkPhase =
+    let
+      regex = lib.escapeShellArg "@[a-zA-Z_][0-9A-Za-z_'-]*@";
+    in
+    ''
+      runHook preCheck
+      if grep -qe ${regex} "$out"; then
+        echo The following look like unsubstituted Nix identifiers that remain in "$out":
+        grep -oe ${regex} "$out"
+        echo Use the more precise '`substitute`' function if this check is in error.
+        exit 1
+      fi
+      runHook postCheck
+    '';
+}
--- a/pkgs/top-level/all-packages.nix
+++ b/pkgs/top-level/all-packages.nix
@@ -1335,6 +1335,8 @@ with pkgs;
 
   replaceDependency = callPackage ../build-support/replace-dependency.nix { };
 
+  replaceVars = callPackage ../build-support/replace-vars { };
+
   nukeReferences = callPackage ../build-support/nuke-references {
     inherit (darwin) signingUtils;
   };
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -43,9 +43,11 @@
 /pkgs/top-level/stage.nix                        @Ericson2314
 /pkgs/top-level/splice.nix                       @Ericson2314
 /pkgs/top-level/release-cross.nix                @Ericson2314
+/pkgs/stdenv                                     @philiptaron
 /pkgs/stdenv/generic                             @Ericson2314
 /pkgs/stdenv/generic/check-meta.nix              @Ericson2314
 /pkgs/stdenv/cross                               @Ericson2314
+/pkgs/build-support                              @philiptaron
 /pkgs/build-support/cc-wrapper                   @Ericson2314
 /pkgs/build-support/bintools-wrapper             @Ericson2314
 /pkgs/build-support/setup-hooks                  @Ericson2314
--- a/pkgs/test/default.nix
+++ b/pkgs/test/default.nix
@@ -183,5 +183,7 @@ with pkgs;
 
   systemd = callPackage ./systemd { };
 
+  replaceVars = recurseIntoAttrs (callPackage ./replace-vars { });
+
   substitute = recurseIntoAttrs (callPackage ./substitute { });
 }
--- a/pkgs/test/replace-vars/default.nix
+++ b/pkgs/test/replace-vars/default.nix
@@ -1,0 +1,74 @@
+{
+  replaceVars,
+  emptyDirectory,
+  emptyFile,
+  runCommand,
+  testers,
+}:
+let
+  inherit (testers) testEqualContents testBuildFailure;
+in
+{
+  # Success case for `replaceVars`.
+  replaceVars = testEqualContents {
+    assertion = "replaceVars";
+    actual = replaceVars ./source.txt {
+      free = "free";
+      "equal in" = "are the same in";
+      brotherhood = "shared humanity";
+    };
+
+    expected = builtins.toFile "expected" ''
+      All human beings are born free and are the same in dignity and rights.
+      They are endowed with reason and conscience and should act towards
+      one another in a spirit of shared humanity.
+
+        -- eroosevelt@humanrights.un.org
+    '';
+  };
+
+  # There might eventually be a usecase for this, but it's not supported at the moment.
+  replaceVars-fails-on-directory =
+    runCommand "replaceVars-fails" { failed = testBuildFailure (replaceVars emptyDirectory { }); }
+      ''
+        grep -e "ERROR: file.*empty-directory.*does not exist" $failed/testBuildFailure.log
+        touch $out
+      '';
+
+  replaceVars-fails-in-build-phase =
+    runCommand "replaceVars-fails"
+      { failed = testBuildFailure (replaceVars emptyFile { not-found = "boo~"; }); }
+      ''
+        grep -e "ERROR: pattern @not-found@ doesn't match anything in file.*empty-file" $failed/testBuildFailure.log
+        touch $out
+      '';
+
+  replaceVars-fails-in-check-phase =
+    runCommand "replaceVars-fails"
+      {
+        failed =
+          let
+            src = builtins.toFile "source.txt" ''
+              Header.
+              before @whatIsThis@ middle @It'sOdd2Me@ after.
+              @cannot detect due to space@
+              Trailer.
+            '';
+          in
+          testBuildFailure (replaceVars src { });
+      }
+      ''
+        grep -e "unsubstituted Nix identifiers.*source.txt" $failed/testBuildFailure.log
+        grep -F "@whatIsThis@" $failed/testBuildFailure.log
+        grep -F "@It'sOdd2Me@" $failed/testBuildFailure.log
+        grep -F 'more precise `substitute` function' $failed/testBuildFailure.log
+
+        # Shouldn't see irrelevant details.
+        ! grep -q -E -e "Header|before|middle|after|Trailer" $failed/testBuildFailure.log
+
+        # Shouldn't see the "cannot detect" version.
+        ! grep -q -F "cannot detect due to space" $failed/testBuildFailure.log
+
+        touch $out
+      '';
+}
--- a/pkgs/test/replace-vars/source.txt
+++ b/pkgs/test/replace-vars/source.txt
@@ -1,0 +1,5 @@
+All human beings are born @free@ and @equal in@ dignity and rights.
+They are endowed with reason and conscience and should act towards
+one another in a spirit of @brotherhood@.
+
+  -- eroosevelt@humanrights.un.org


pkgs/build-support/substitute/substitute-attrs.nix
```diff
@@ -0,0 +1,40 @@
...
+
+  ```nix
+  { substituteAttrs }:
+
+  substituteAttrs ./greeting.txt { world = "hello"; }
+  ```
+
+  See ../../test/substitute for tests
+*/
+srcOrAttrs: attrs:
```
a-n-n-a-l-e-e, line 24: why two arguments rather than just an attribute set?

```nix
substituteAttrs { src = ./the/source; subAttrs = { a = xxx; b = zzz; }; };
```

ghost (commented):
cool. thanks!

roberth (commented):
Thank you for picking this up!

pkgs/build-support/substitute/substitute-attrs.nix
infinisil: I'd be in favor of using an attribute set instead of curried arguments, which also allows clearing up the src/attrs thing with separately named arguments.

pkgs/build-support/substitute/substitute-attrs.nix
```diff
@@ -0,0 +1,40 @@
...
+/*
+  This is a wrapper around `substitute` in the stdenv, designed to be as easy to use as
+  `substituteAll` but have fewer pitfalls.
+
+  Attribute arguments:
+
+  - `srcOrAttrs`: either the path to the file to substitute, or an attrset that fits into `substitute`
+
+  - `attrs`: each entry in this set corresponds to a `--subst-var-by` entry in `substitute`.
+    See https://nixos.org/manual/nixpkgs/stable/#fun-substitute
```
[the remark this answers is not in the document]
infinisil: After reading [Philip's reply](https://github.com/NixOS/nixpkgs/pull/301350#issuecomment-2043165149), I'm convinced that `--subst-var-by` is better :)

pkgs/build-support/substitute/substitute-attrs.nix
```diff
@@ -0,0 +1,40 @@
+{ lib, substitute }:
+
+/*
+  This is a wrapper around `substitute` in the stdenv, designed to be as easy to use as
+  `substituteAll` but have fewer pitfalls.
+
+  Attribute arguments:
+
+  - `srcOrAttrs`: either the path to the file to substitute, or an attrset that fits into `substitute`
```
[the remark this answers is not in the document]
infinisil: I think this can be just
```suggestion
  - `path`: A [string-like value](https://nixos.org/manual/nixpkgs/unstable/#function-library-lib.strings.isStringLike) of a [store path](https://nixos.org/manual/nix/stable/store/store-path.html#store-path).
```

Then both `./foo.patch` and derivations can be passed. It's a fairly standard notion to do this.

Oh and I'd also either implement it already, or at least leave the possibility open, to support full directories as well, instead of just single files.

infinisil:
> **Has a name that reflects its main use case.** The name I initially chose, `substituteAttrs`, didn't hit this mark. Perhaps `substituteInPatch` might.

I don't think it should be named for its main use case, but rather for what it does. It doesn't do anything specific to patching, so I don't think `substituteInPatch` is a good name, I'm in favor of keeping it as `substituteAttrs`.

pkgs/build-support/substitute/substitute-attrs.nix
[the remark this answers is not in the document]
infinisil: [Philip's comment](https://github.com/NixOS/nixpkgs/pull/301350#issuecomment-2043165149) replies to this.

I think the argument order is fine as is. I've never seen the use case of mapping over such a function, it's very rare (and would still work with a `flip`). Also such a use case could also be supported with `substituteAttrs ./. { shell = runtimeShell; }` (substituting all files in the directory).

Instead `substituteAttrs ./file.sh { ... }` looks much cleaner for the majority of use cases, where it's only used once, and the attributes are much longer than the file argument.

infinisil:
> What do you think about the name `substituteInPatch` or `substituteInFile`, both of which invite a file as the next parameter?

Actually `substituteInFile` seems decent, but then it couldn't support the use case of substituting files in an entire directory. Instead how about `substituteInPath`

pkgs/top-level/all-packages.nix
```diff
@@ -1375,6 +1375,8 @@ with pkgs;
 
   substituteAllFiles = callPackage ../build-support/substitute-files/substitute-all-files.nix { };
 
+  substituteAttrs = callPackage ../build-support/substitute/substitute-attrs.nix { };
```
roberth, line 1378: Do you think in this case it'd be good to make a start with [ending a homonym](https://github.com/NixOS/nixpkgs/issues/237216#issuecomment-2045821891)?

Something like:

```suggestion
  replaceByAttrs = callPackage ../build-support/substitute/substitute-attrs.nix { };
```

infinisil: I like it!

pkgs/build-support/substitute/substitute-attrs.nix
```diff
@@ -0,0 +1,40 @@
+{ lib, substitute }:
+
+/*
+  This is a wrapper around `substitute` in the stdenv, designed to be as easy to use as
+  `substituteAll` but have fewer pitfalls.
```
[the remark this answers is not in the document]
philiptaron: Done in updated version.

pkgs/build-support/substitute/substitute-attrs.nix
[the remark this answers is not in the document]
philiptaron: I took Silvan's text.

pkgs/top-level/all-packages.nix
philiptaron: I decided to go with `substituteInPath` per Silvan's suggestion, but I _do_ like `replaceByAttrs` and could easily be swayed to that name.

roberth: How about `replaceInPath`?

> This isn't [Nix substitution](https://nixos.org/manual/nix/stable/glossary#gloss-substituter)

philiptaron: Updated.

pkgs/test/replace-in-path/default.nix
```diff
@@ -0,0 +1,74 @@
...
+        grep -e "unsubstituted Nix identifiers.*source.txt" $failed/testBuildFailure.log
+        grep -F "@whatIsThis@" $failed/testBuildFailure.log
+        grep -F "@It'sOdd2Me@" $failed/testBuildFailure.log
+        grep -F 'more precise `substitute` function' $failed/testBuildFailure.log
+
+        # Shouldn't see irrelevant details.
+        grep -q -v -e "Header|before|middle|after|Trailer" $failed/testBuildFailure.log
+
+        # Shouldn't see the "cannot detect" version.
+        grep -q -v -F "cannot detect due to space" $failed/testBuildFailure.log
```
roberth, line 70: Probably another unintentional `-v`.

roberth (commented):
Still not a big fan of the automatic `@`s, also noting that the name is very generic, admittedly due to my own suggestion.
Should it be like this?
```nix
replaceInPath { variables = { inherit bash; }; } ./foo.sh
```

That way it can be extended with arguments like
```nix
replaceInPath { replacements = { "/store/dir" = builtins.storeDir; }; } bar
```

philiptaron:
Robert, I'm pretty set on the argument order (attrset last) and the semantics (`name` = value, where name in the source is `@name@`). They correspond the actual need and usecase in nixpkgs, and they're similar to other well-known functions like `callPackage`. Considering future additions other than _maybe_ directories is, in my view, out of scope. My view: YAGNI.

On the other hand, I'm very open to other names than `replaceInPath`, especially if they help with the ill-at-ease feeling you have about `name` => `@name@`.

What about `replaceVarsInPath` or `replaceVarsByAttrs`, following the name of `--subst-var` and `--subst-var-by`, both of which have the `@name@` convention? Or `replaceAllVarsInPath`?

pkgs/test/replace-in-path/default.nix
```diff
@@ -0,0 +1,74 @@
...
+          testBuildFailure (replaceInPath src { });
+      }
+      ''
+        grep -e "unsubstituted Nix identifiers.*source.txt" $failed/testBuildFailure.log
+        grep -F "@whatIsThis@" $failed/testBuildFailure.log
+        grep -F "@It'sOdd2Me@" $failed/testBuildFailure.log
+        grep -F 'more precise `substitute` function' $failed/testBuildFailure.log
+
+        # Shouldn't see irrelevant details.
+        grep -q -v -e "Header|before|middle|after|Trailer" $failed/testBuildFailure.log
```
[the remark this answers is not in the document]
philiptaron: Thanks for the primer. I will indeed change the test to reflect the desired semantics.

emilazy:
I don’t know if I like the name `replaceInPath` but I like that it is short and relatively snappy. “Use `replaceAllVarsInPath`, not `substituteAll`” feels like an unforced error in terms of “marketing”. Ideally we would somehow give it a name that is even nicer and more appealing than `substituteAll`. `replaceVars`?

roberth:
> the argument order (attrset last)

Whoops, didn't mean to relitigate this. I just find it natural to order positional arguments from least dynamic to most dynamic, and I consider the substitutions to be the more dynamic value.

> My view: YAGNI.

We know we need all the functionality, because otherwise it wouldn't already exist in the bash functions.
The real question is how to turn all the functionality into nice interfaces.

I completely understand the need to scope things down, but by saying we won't need it, you're denying a clear need.
I'm probably just taking these five words too literally, but we wouldn't be doing any of this if words didn't matter.

Or maybe I missed something; is this part of a larger plan you have in mind?

Please take my comments with a grain of salt - I think you're more aware of the big picture here than I am.

emilazy:
FWIW I think having to write `{ variables = …; }` would also be bad for user acceptance. Not that a more general interface would be bad to have, but we should have something simple for the common case of “script or patch that needs something parameterized”, and it’s okay to optimize for the common case even if we want to expose something more advanced too.

roberth:
Fair enough then.
@emilazy makes a good point about "marketability", so if you could rename it to her suggestion `replaceVars`, that sounds good to me.

Sorry for my detour, but I think we're getting to a good place.

philiptaron:
> `replaceVars`?

I _love_ `replaceVars`.

pkgs/build-support/replace-vars/default.nix
```diff
@@ -0,0 +1,77 @@
...
+
+  Any remaining text that matches `@[A-Za-z_][0-9A-Za-z_'-]@` in the output after replacement
+  has occurred will cause a build failure.
+
+  Attribute arguments:
+
+  - `path`: A [string-like value](https://nixos.org/manual/nixpkgs/stable/#function-library-lib.strings.isStringLike) of a [store path](https://nixos.org/manual/nix/stable/store/store-path.html#store-path).
+
+  - `attrs`: each entry in this set corresponds to a `--subst-var-by` entry in `substitute`.
+    See https://nixos.org/manual/nixpkgs/stable/#fun-substitute
```
infinisil, line 21:
```suggestion
  # Inputs

  `path` ([Store Path](https://nixos.org/manual/nix/latest/store/store-path.html#store-path) String)
  : The file to substitute values in.

  `attrs` (AttrsOf String)
  : Each entry in this set corresponds to a `--subst-var-by` entry in [`substitute`](https://nixos.org/manual/nixpkgs/stable/#fun-substitute).
```

Using the [documentation conventions](https://github.com/NixOS/nixpkgs/tree/master/doc#documentation-conventions) and some adjustments:
- Don't link to `isStringLike`, I don't think it's helpful here.
- Use the `latest` Nix version link, the `stable` one doesn't seem to exist
- Give a description for `path`
- More brief linking to `substitute`

pkgs/build-support/replace-vars/default.nix
```diff
@@ -0,0 +1,77 @@
...
+
+  Example:
```
infinisil, line 23:
```suggestion
  # Example
```

This isn't 100% conform with the doc conventions, but we're not rendering this file anyways, [for now](https://github.com/nix-community/nixdoc/issues/21).

pkgs/build-support/replace-vars/default.nix
```diff
@@ -0,0 +1,77 @@
...
+
+  See `../../test/replace-vars/default.nix` for tests of this function.
+*/
+path: attrs:
+
+let
+  # We use `--replace-fail` instead of `--subst-var-by` so that if the thing isn't there, we fail.
+  subst-var-by = name: value: [
+    "--replace-fail"
+    (lib.escapeShellArg "@${toString name}@")
```
infinisil, line 39:
```suggestion
    (lib.escapeShellArg "@${name}@")
```

`name` is always a string already

pkgs/build-support/replace-vars/default.nix
```diff
@@ -0,0 +1,77 @@
...
+    (lib.escapeShellArg (toString value))
```
infinisil, line 40:
```suggestion
    (lib.escapeShellArg value)
```

`toString` is a notoriously odd function for converting values to strings (it just concats lists using `" "` without any escaping!), so I think it's better to only support [string-like](https://nixos.org/manual/nixpkgs/unstable/#function-library-lib.strings.isStringLike) values.

pkgs/build-support/replace-vars/default.nix
```diff
@@ -0,0 +1,77 @@
...
+  name = baseNameOf (toString path);
+  src = path;
+  doCheck = true;
+  dontUnpack = true;
+  preferLocalBuild = true;
+  allowSubstitutes = false;
+
+  buildPhase = ''
+    substitute "$src" "$out" ${lib.concatStringsSep " " substitutions}
+  '';
```
infinisil, line 56:
```suggestion
  buildPhase = ''
    runHook preBuild
    substitute "$src" "$out" ${lib.concatStringsSep " " substitutions}
    runHook postBuild
  '';
```

Pretty standard to support overriding with these hooks and not have them be ignored.

pkgs/build-support/replace-vars/default.nix
```diff
@@ -0,0 +1,77 @@
...
+    let
+      regex = lib.escapeShellArg "@[a-zA-Z_][0-9A-Za-z_'-]*@";
+    in
+    ''
+      if grep -qe ${regex} "$out"; then
+        echo The following look like unsubstituted Nix identifiers that remain in "$out":
+        grep -oe ${regex} "$out"
+        echo Use the more precise '`substitute`' function if this check is in error.
+        exit 1
+      fi
```
infinisil, line 69:
```suggestion
      runHook preCheck
      if grep -qe ${regex} "$out"; then
        echo The following look like unsubstituted Nix identifiers that remain in "$out":
        grep -oe ${regex} "$out"
        echo Use the more precise '`substitute`' function if this check is in error.
        exit 1
      fi
      runHook postCheck
```

pkgs/build-support/replace-vars/default.nix
```diff
@@ -0,0 +1,77 @@
...
+    '';
+
+  meta = {
+    description = "terse name replacement based on attrsets";
+    platforms = lib.platforms.all;
+    maintainers = [ lib.maintainers.philiptaron ];
+  };
```
infinisil, line 76: I'm not sure if it makes sense to have meta information here, because this doesn't describe the function, but rather the output of the function.

And since you're a committer now, the best way to become a maintainer of this is probably to add yourself to `.github/CODEOWNERS` :)

infinisil (changes requested):
I like it! I found some things to address, but otherwise I think we should merge this :)

pkgs/build-support/replace-vars/default.nix
infinisil: Just noticed that you took the original text here [from me](https://github.com/NixOS/nixpkgs/pull/301350#discussion_r1557887949) :see_no_evil: :sweat_smile:

philiptaron: Yep, it's past Silvan talking to now Silvan here 😉 I'll apply the edits.

pkgs/build-support/replace-vars/default.nix
philiptaron: Done

pkgs/build-support/replace-vars/default.nix
philiptaron: Done

pkgs/build-support/replace-vars/default.nix
philiptaron: Great call, done.

pkgs/build-support/replace-vars/default.nix
philiptaron: I'll take your advice.

infinisil (approved):
LGTM!

roberth:
Without user facing docs (https://github.com/NixOS/nixpkgs/pull/301350#discussion_r1706246442) this may as well not exist :cry:

pkgs/build-support/replace-vars/default.nix
roberth: > it just concats lists using `" "` without any escaping

What's worse, is that it converts absolute paths to strings without context, creating "references" to paths that may not exist on the target host, like `/home/user/src/foo`.
Good riddance!

philiptaron:
> Without user facing docs (https://github.com/NixOS/nixpkgs/pull/301350#discussion_r1706246442) this may as well not exist :cry: 

It needs some real life transformation experience. When I get back from camping I'm planning on doing a pass through nixpkgs on current substituteAll calls to see if it works or needs tweaks.

Then the marketing can ensue.

roberth:
Awesome. Enjoy camping!

philiptaron:
Docs are live on noogle at least: https://noogle.dev/f/pkgs/replaceVars

philiptaron:
Things I've run into so far that might motivate a `replaceVarsWith` function:

1. `isExecutable = true`, so that the replacement has the executable bit set.
2. Exceptions for the check phase: if the replacements are done in two phases, that's a blocker.

I'm just skipping derivations that have these properties.

emilazy:
> 1. `isExecutable = true`, so that the replacement has the executable bit set.

What if we made the output have the same executable bit as the input?

PR #301472, opened 2024-04-04
modules: make `apply` mergeable

Atry:
## Description of changes

As discussed [here](https://discourse.nixos.org/t/drv-parts-configure-packages-like-nixos-systems/23629) and [here](https://discourse.nixos.org/t/adding-more-control-over-overriding-default-module-definitions/4276), a restriction in module system in comparison to overlay is that we cannot visit previous value, nor transform it into a new one.

This PR aims to make `apply` mergeable, therefore modules can visit previous values by defining an `apply` function.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/lib/modules.nix
+++ b/lib/modules.nix
@@ -744,6 +744,9 @@ let
           t' = opt.options.type;
           mergedType = t.typeMerge t'.functor;
           typesMergeable = mergedType != null;
+          applySet = if (bothHave "apply")
+                        then { apply = value: opt.options.apply (res.apply value); }
+                        else {};
           typeSet = if (bothHave "type") && typesMergeable
                        then { type = mergedType; }
                        else {};
@@ -752,7 +755,6 @@ let
       if bothHave "default" ||
          bothHave "example" ||
          bothHave "description" ||
-         bothHave "apply" ||
          (bothHave "type" && (! typesMergeable))
       then
         throw "The option `${showOption loc}' in `${opt._file}' is already declared in ${showFiles res.declarations}."
@@ -772,7 +774,7 @@ let
                 then [opt.pos]
                 else [{ file = opt._file; line = null; column = null; }]);
             options = submodules;
-          } // typeSet
+          } // applySet // typeSet
     ) { inherit loc; declarations = []; declarationPositions = []; options = []; } opts;
 
   /* Merge all the definitions of an option to produce the final
--- a/nixos/doc/manual/release-notes/rl-2405.section.md
+++ b/nixos/doc/manual/release-notes/rl-2405.section.md
@@ -407,6 +407,8 @@ The pre-existing [services.ankisyncd](#opt-services.ankisyncd.enable) has been m
 
 - A stdenv's default set of hardening flags can now be set via its `bintools-wrapper`'s `defaultHardeningFlags` argument. A convenient stdenv adapter, `withDefaultHardeningFlags`, can be used to override an existing stdenv's `defaultHardeningFlags`.
 
+- The `apply` function passed to `mkOption` is now mergeable. This allows for accessing the previous value of the option, similar to the `prev`/`super` argument in `overlays`, `override` and `overrideAttrs`.
+
 - `libass` now uses the native CoreText backend on Darwin, which may fix subtitle rendering issues with `mpv`, `ffmpeg`, etc.
 
 - [Lilypond](https://lilypond.org/index.html) and [Denemo](https://www.denemo.org) are now compiled with Guile 3.0.


roberth:
I don't think we should do this, because it makes more code dependent on the import order, which is too unpredictable and un-declarative, similar to what I've argued in the linked drv-parts thread.

Atry:
Most existing overlays and overrides are commutative. The order does not matter at all in most use case.

Atry:
The use case of `apply` would be to apply a patch to a package, while not breaking other patches. For example, I want to patch a package from a flake-part.

Atry:
The use case would be to patch an existing package while not breaking other patches. In that case, the users would not depend on module application order.

Atry:
> because it makes more code dependent on the import order

Could you elabrate in what use case the code would depend on the import order?

roberth:
> The use case would be to patch an existing package while not breaking other patches. In that case, the users would not depend on module application order.

> what use case

Some examples in `haskell.lib` are
- `doCheck` and `dontCheck`, in which case the last one wins
- `buildFromSdist`, creates a copy of the package, modifies it, and uses it as the `src`. The state of the copy depends on the order; only preceding operations are included.

infinisil:
I agree with Robert here. `apply` in general should be viewed as an anti-pattern, and ideall be deprecated completely.

Atry:
> Some examples in `haskell.lib` are
> 
> * `doCheck` and `dontCheck`, in which case the last one wins
> * `buildFromSdist`, creates a copy of the package, modifies it, and uses it as the `src`. The state of the copy depends on the order; only preceding operations are included.

OK. I see your point.
In an ideal world, all packages are drv-parts, all configurations are module system, we would not need mergeable `apply`. We can remove it as an anti-pattern.

Unforunately given that `nixpkgs` still includes more packages than `dream2nix`. We still live in a world mixed of modules and overrides, we still need a way to apply a patch in the module system.

@roberth you said "That doesn’t mean that we shouldn’t have it. In fact by having them, we have a good place to document why they are the way they are, and that they’re a last resort."

roberth:
> @roberth you said "That doesn’t mean that we shouldn’t have it. In fact by having them, we have a good place to document why they are the way they are, and that they’re a last resort."

That was in reference to types `endo` and `overlay`, with checks that require either all definitions to have distinct `mkOrder` numbers. (Noting that a definition like `mkMerge ([ a b ] ++ optionals c [ d ])` is a single definition and doesn't need `mkOrder`.)

In `apply`, we don't have the luxury of `mkOrder`, and while that could be added, I doubt that we should. Adding an explicit option to perform the postprocessing has the benefit that it's documented; more discoverable. It also doesn't hurt performance.

Atry:
> with checks that require either all definitions to have distinct `mkOrder` numbers.

Suppose I want to create a NixOS module that patches `java` command line to provide `java --my-fancy-dev-flag`, so I can do this:
```
{lib, ...}: {
  options.programs.java.package = lib.mkOption {
    apply = old: old.override { ... };
  };
}
```

I want `apply` to be mergeable so that the module would not conflict with other module that also patches `java`. How can I determine the `mkOrder` number if `options.programs.java` is an `endo` or `overlay` type?

roberth:
It looks like you're recreating the traditional FHS problem of having only a single instance of a program, which has to satisfy all possible requirements simultaneously.

If you're sure that's what you want, you could improve the `programs.java` module to support all those requirements. Centralizing that makes sense because the requirements may interact. It also makes your improvements available to others, and invites others to share theirs, so you can make use of them too.

Otherwise, I would recommend nix shells for projects, and independent java packages for each service that needs one.
If some services have the same version and overrides, they may still share a store path, thanks to hashing.

If you don't want explicit support for the things you're doing, and you would rather expose the `apply` attribute through an option with a type like `endo` or `overlay`, as you say, you will have to figure out the numbers for your configuration somehow. I don't think that's exposed in `options` just yet, but maybe that could be added, perhaps in the form of raw definitions. I wonder how costly that is.

Atry:
> It looks like you're recreating the traditional FHS problem of having only a single instance of a program, which has to satisfy all possible requirements simultaneously.

The module is not necessarily a NixOS module, when it's a flake part or devenv module, the option could be a per project package, for example, [languages.python.package](https://devenv.sh/reference/options/#languagespythonpackage). In that case, it's not a single instance of a program.




> If you're sure that's what you want, you could improve the `programs.java` module to support all those requirements. Centralizing that makes sense because the requirements may interact. It also makes your improvements available to others, and invites others to share theirs, so you can make use of them too.

The build options are already available in `nixpkgs` as `override` arguments. Ideally `programs.java` should only include the runtime options of the program, not the build options.



> you will have to figure out the numbers for your configuration somehow.

What does "have to" mean?

1. This PR does not enforce the order number
2. The proposal to enforce the order number does not help the typical use case, where the `apply` function is assumed commutative.

The `doCheck` vs `dontCheck` is a good example where overrides are assumed to be commutative, but not actually 100% commutative, which is unfortunate. However, why would an enforced order number mitigate the unfortune? If we enforce the order number, what is the best practice to determine the order number?

physics-enthusiast:
How about having a 2-level merging system? First merge all the non-`apply` options. Then we treat apply itself like an option (with `mkDefault`s, `mkOverride`s, and `mkForce`s), and only apply one. Then those who want composition can set `apply.apply`. We could even add a `mkPrecedence` to `lib` to allow the creation of an arbitrarily large attrset tower of `apply`s for those who need the increased composability.

i.e.
```
mkPrecedence = n: applyFunc:
    assert n >= 1
    assert (typeOf n == "int")
    if n==1 then
        { apply = applyFunc }
    else
        { apply = mkPrecedence (n-1) applyFunc }
```

roberth:
@Atry 

> > you will have to figure out the numbers for your configuration somehow.
> 
> What does "have to" mean?
> 
>     1. This PR does not enforce the order number
> 
>     2. @roberth's proposal to enforce the order number does not help the typical use case, where the `apply` function is assumed commutative.

I was trying to illustrate why a composition of functions is not properly declarative, by showing workarounds. I don't think we need to go into much more detail here to see that the workarounds aren't great, but we could if you think we need to.

Crucially, it's important for the module system to provide sound abstractions that behave predictably, as much as possible. Every time we introduce a feature that's _not perfect, but helps a particular use case_, it will be used in cases where it's not needed, leading to annoyances and bugs. This in turn damages the reputation of the module system, making it seem unreliable.
Documentation can maybe mitigate this slightly, but not prevent this. It's self-defeating: _the module system is great, so how bad could this be? We barely use this new option anyway..._

> However, how would an enforced order number mitigate the unfortune? If we enforce the order number, what is the best practice to determine the order number?

It gives a user-visible clue as to what might be going on. I think it's a bit early to ask for best _practices_; I think something might emerge.
What I do know is that it will be unpleasant. The best way to avoid it is to just have a single definition, in which case you won't have to specify anything like `mkAfter`/`mkOrder`/etc.
That's not modular, but that's because the original problem statement (function composition) just does not have a modular solution.

roberth:
@physics-enthusiast 

> How about having a 2-level merging system? First merge all the non-`apply` options. Then we treat apply itself like an option (with `mkDefault`s, `mkOverride`s, and `mkForce`s), and only apply one.

Laziness makes it so that the way things are evaluated is very flexible by default. We probably don't need to create "phases" like that at the module level, unless you really want to generate more actual options, in which case the overhead might double, or worst case something like exponential in the depth of submodule nestings. (In case both phases need to evaluate both phases of a submodule)

Instead of building something extra into the module system, we can already do this in the cases where we need it:

```nix
{ config, ... }:
{
  options = {
    foo = mkOption {
      type = <...>;
      apply = config.fooModifiers;
    };
    fooModifiers = mkOption {
      type = types.composedFunction;  # Example, this type doesn't exist yet
    };
  };
}
```

This way you can pick how you want the functions to be merged, and we don't need to complicate the module system or make it slower by even one conditional.
It also does not require that your users learn a new module system feature. Instead they'll find these options in the module documentation where it's relevant.

physics-enthusiast:
My bad, I had misunderstood `apply` as something the option *setter* supplied, rather than the option *provider*. I still think the ability to pass a magic value that essentially says "give me whatever value this would have been, and let me modify it" as a module *user* would be cool, but that's something for perhaps a different issue another day (if I ever get around to raising it), so please do disregard my earlier comment.

PR #265409, opened 2023-11-04
passt: init at 2023_11_10

8aed:
## Description of changes

This package provides [a user-mode networking tool](https://passt.top/passt/about/) that can be used with virtual machines (`passt`) and namespaces (`pasta`), like `SLIRP` and `slirp4netns`, which are already packaged in NixOS.

It is especially useful when one wants to provide networking access to an unprivileged container.  It uses a different implementation than SLIRP, supports zero-copy  for local connections, is more optimized and generally provides higher throughput than other MTU-based solutions. When available, It is the default networking mode for rootlless podman.

## Things done

- Built on platform(s)
  - [X] x86_64-linux
  - [ ] aarch64-linux
- For non-Linux: Is sandboxing enabled in `nix.conf`? (See [Nix manual](https://nixos.org/manual/nix/stable/command-ref/conf-file.html))
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:
  - [NixOS test(s)](https://nixos.org/manual/nixos/unstable/index.html#sec-nixos-tests) (look inside [nixos/tests](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests))
  - and/or [package tests](https://nixos.org/manual/nixpkgs/unstable/#sec-package-tests)
  - or, for functions and "core" functionality, tests in [lib/tests](https://github.com/NixOS/nixpkgs/blob/master/lib/tests) or [pkgs/test](https://github.com/NixOS/nixpkgs/blob/master/pkgs/test)
  - made sure NixOS tests are [linked](https://nixos.org/manual/nixpkgs/unstable/#ssec-nixos-tests-linking) to the relevant packages
- [X] Tested basic functionality of all binary files (usually in `./result/bin/`)
- [23.11 Release Notes](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2311.section.md) (or backporting [23.05 Release notes](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2305.section.md))
  - [ ] (Package updates) Added a release notes entry if the change is major or breaking
  - [ ] (Module updates) Added a release notes entry if the change is significant
  - [ ] (Module addition) Added a release notes entry if adding a new NixOS module
- [X] Fits [CONTRIBUTING.md](https://github.com/NixOS/nixpkgs/blob/master/CONTRIBUTING.md).

PR #301553, opened 2024-04-04
podman: 4.9.3 -> 5.0.1

the-sun-will-rise-tomorrow:
## Description of changes

- https://github.com/containers/podman/releases/tag/v5.0.0
- https://github.com/containers/podman/releases/tag/v5.0.1

Depends on https://github.com/NixOS/nixpkgs/pull/265409

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [x] Tested, as applicable:

---

### DIFF
--- a/pkgs/applications/virtualization/podman/default.nix
+++ b/pkgs/applications/virtualization/podman/default.nix
@@ -17,9 +17,12 @@
 , makeWrapper
 , runtimeShell
 , symlinkJoin
+, substituteAll
 , extraPackages ? [ ]
 , runc
 , crun
+, gvisor
+, youki
 , conmon
 , slirp4netns
 , fuse-overlayfs
@@ -30,6 +33,7 @@
 , gvproxy
 , aardvark-dns
 , netavark
+, passt
 , testers
 , podman
 }:
@@ -37,9 +41,6 @@ let
   # do not add qemu to this wrapper, store paths get written to the podman vm config and break when GCed
 
   binPath = lib.makeBinPath (lib.optionals stdenv.isLinux [
-    runc
-    crun
-    conmon
     fuse-overlayfs
     util-linux
     iptables
@@ -57,21 +58,27 @@ let
       catatonit # added here for the pause image and also set in `containersConf` for `init_path`
       netavark
       slirp4netns
+      passt
     ];
   };
 in
 buildGoModule rec {
   pname = "podman";
-  version = "4.9.3";
+  version = "5.0.1";
 
   src = fetchFromGitHub {
     owner = "containers";
     repo = "podman";
     rev = "v${version}";
-    hash = "sha256-PdAXcXtc/Jl3ttWWB6TciiOwWescJ51Glhf2ZhOw550=";
+    hash = "sha256-XgLrPLswLmaB9FYXKEMLP+7KT/OY50z3JKz8DvMLrEE=";
   };
 
   patches = [
+    (substituteAll {
+      src = ./hardcode-paths.patch;
+      inherit crun runc gvisor youki conmon;
+    })
+
     # we intentionally don't build and install the helper so we shouldn't display messages to users about it
     ./rm-podman-mac-helper-msg.patch
   ];
--- a/pkgs/applications/virtualization/podman/hardcode-paths.patch
+++ b/pkgs/applications/virtualization/podman/hardcode-paths.patch
@@ -1,0 +1,109 @@
+diff --git a/vendor/github.com/containers/common/pkg/config/default.go b/vendor/github.com/containers/common/pkg/config/default.go
+index 19c4bb6bf..2743de4b2 100644
+--- a/vendor/github.com/containers/common/pkg/config/default.go
++++ b/vendor/github.com/containers/common/pkg/config/default.go
+@@ -364,75 +364,34 @@ func defaultEngineConfig() (*EngineConfig, error) {
+ 	c.Retry = 3
+ 	c.OCIRuntimes = map[string][]string{
+ 		"crun": {
+-			"/usr/bin/crun",
+-			"/usr/sbin/crun",
+-			"/usr/local/bin/crun",
+-			"/usr/local/sbin/crun",
+-			"/sbin/crun",
+-			"/bin/crun",
+-			"/run/current-system/sw/bin/crun",
++			"@crun@/bin/crun",
+ 		},
+ 		"crun-vm": {
+-			"/usr/bin/crun-vm",
+-			"/usr/local/bin/crun-vm",
+-			"/usr/local/sbin/crun-vm",
+-			"/sbin/crun-vm",
+-			"/bin/crun-vm",
+-			"/run/current-system/sw/bin/crun-vm",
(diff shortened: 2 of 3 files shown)


pkgs/applications/virtualization/podman/default.nix
```diff
@@ -57,21 +58,31 @@ let
...
   };
 
   patches = [
+    (substituteAll {
+      src = ./hardcode-paths.patch;
+      inherit crun;
+      inherit runc;
+      inherit gvisor;
+      inherit youki;
+      inherit conmon;
```
SuperSandro2000, line 83:
```suggestion
      inherit crun runc gvisor youki conmon;
```

pkgs/by-name/pa/passt/package.nix
```diff
@@ -0,0 +1,41 @@
...
+stdenv.mkDerivation {
+  pname = "passt";
+  version = "0.2023_11_10.5ec3634";
+  src = fetchgit {
+    url = "git://passt.top/passt";
+    rev = "5ec3634b07215337c2e69d88f9b1d74711897d7d";
+    hash = "sha256-76CD9PYD/NcBkmRYFSZaYl381QJjuWo0VsNdh31d6/M=";
+  };
+  nativeBuildInputs = [ ];
+  buildInputs = [];
```
SuperSandro2000, line 12:
```suggestion
```

pkgs/applications/virtualization/podman/default.nix
WxNzEMof: OK, done.

pkgs/by-name/pa/passt/package.nix
WxNzEMof: This was in https://github.com/NixOS/nixpkgs/pull/265409 which is now in master, sorry.

WxNzEMof:
Updated!

SuperSandro2000 (approved):
rootless continues to work on WSL

pkgs/applications/virtualization/podman/hardcode-paths.patch
```diff
@@ -0,0 +1,109 @@
...
+-			"/sbin/runsc",
+-			"/run/current-system/sw/bin/runsc",
++			"@gvisor@/bin/runsc",
+ 		},
+ 		"youki": {
+-			"/usr/local/bin/youki",
+-			"/usr/bin/youki",
+-			"/bin/youki",
+-			"/run/current-system/sw/bin/youki",
++			"@youki@/bin/youki",
```
SuperSandro2000, line 78: Do we really need to patch this, if the nixos path is already one line above?

pkgs/applications/virtualization/podman/default.nix
```diff
@@ -57,21 +58,27 @@ let
...
     repo = "podman";
     rev = "v${version}";
-    hash = "sha256-PdAXcXtc/Jl3ttWWB6TciiOwWescJ51Glhf2ZhOw550=";
+    hash = "sha256-XgLrPLswLmaB9FYXKEMLP+7KT/OY50z3JKz8DvMLrEE=";
   };
 
   patches = [
+    (substituteAll {
+      src = ./hardcode-paths.patch;
+      inherit crun runc gvisor youki conmon;
```
SuperSandro2000, line 79: I think we should try to limit this to what is useful. I thin kcrun is in almost every case better than runc, so we can drop that.

see https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/building_running_and_managing_containers/selecting-a-container-runtime_building-running-and-managing-containers#con_the-crun-container-runtime_selecting-a-container-runtime

pkgs/applications/virtualization/podman/hardcode-paths.patch
VPanteleev-S7: Do you mean `/run/current-system`?

I think it would be good to have this package usable outside of NixOS, but even on NixOS, it would be useful to be able to use a specific version of a dependency.

WxNzEMof: Do you mean `/run/current-system`?

I think it would be good to have this package usable outside of NixOS, but even on NixOS, it would be useful to be able to use a specific version of a dependency.

pkgs/applications/virtualization/podman/default.nix
WxNzEMof: Why would we want to limit this - to reduce the number of dependencies?

But, if `runc` is obsolete, why would upstream still continue supporting it?

Since upstream continues supporting it, wouldn't that imply that it is valuable to have it as an option?

Also, if we don't want to support `runc`, shouldn't we delete it from nixpkgs?

SuperSandro2000: > Why would we want to limit this - to reduce the number of dependencies?

Yeah, we increased the dependencies with this update a lot.

> But, if `runc` is obsolete, why would upstream still continue supporting it?

Maybe there are certain usecases when it is useful which are not relevant to nixpkgs?

SuperSandro2000 (approved):

pkgs/applications/virtualization/podman/default.nix
WxNzEMof: Would it make sense to make these dependencies (`runc`, `gvisor`, `youki`) optional?

SuperSandro2000: Yeah, I think so. I think we could right now overwrite them with null, but that is a bit dirty.

natsukium:
This PR broke darwin support.
```
error: Package ‘conmon-2.1.10’ in /nix/store/wmcbdvplkli110vmdn1ci96n8zmzh7yn-source/pkgs/applications/virtualization/conmon/default.nix:45 is not available on the requested hostPlatform:
         hostPlatform.config = "aarch64-apple-darwin"
```

mlyxshi:
`gvisor` doesn't support aarch64-linux now
```
error: Package ‘gvisor-20240401.0’ in /nix/store/v4pcs3nzx54m5bmxd39win0rgl2d2hbx-source/pkgs/by-name/gv/gvisor/package.nix:46 is not available on the requested hostPlatform:
         hostPlatform.config = "aarch64-unknown-linux-gnu"
         package.meta.platforms = [
           "x86_64-linux"
         ]

```

WxNzEMof:
@natsukium Sorry about that, if you're willing to try a patch, I think this just needs the Linux-only dependencies to be made optional (e.g. with `// optionalAttrs`) in `substituteAll`.

pkgs/applications/virtualization/podman/default.nix
WxNzEMof: https://github.com/NixOS/nixpkgs/pull/310182 does this. I left `runc` in by default on Linux, because nixpkgs actually has a test that podman works with that runtime. But, it can now be disabled.

PR #300570, opened 2024-03-31
stem: init at 1.8.2

Tochiaha:
## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
- For non-Linux: Is sandboxing enabled in `nix.conf`? (See [Nix manual](https://nixos.org/manual/nix/stable/command-ref/conf-file.html))
  - [ ] `sandbox = relaxed`
  - [x] `sandbox = true`
- [x] Tested, as applicable:
  - [NixOS test(s)](https://nixos.org/manual/nixos/unstable/index.html#sec-nixos-tests) (look inside [nixos/tests](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests))
  - and/or [package tests](https://nixos.org/manual/nixpkgs/unstable/#sec-package-tests)
  - or, for functions and "core" functionality, tests in [lib/tests](https://github.com/NixOS/nixpkgs/blob/master/lib/tests) or [pkgs/test](https://github.com/NixOS/nixpkgs/blob/master/pkgs/test)
  - made sure NixOS tests are [linked](https://nixos.org/manual/nixpkgs/unstable/#ssec-nixos-tests-linking) to the relevant packages
- [x] Tested compilation of all packages that depend on this change using `nix-shell -p nixpkgs-review --run "nixpkgs-review rev HEAD"`. Note: all changes have to be committed, also see [nixpkgs-review usage](https://github.com/Mic92/nixpkgs-review#usage)
- [x] Tested basic functionality of all binary files (usually in `./result/bin/`)
- [24.05 Release Notes](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2405.section.md) (or backporting [23.05](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2305.section.md) and [23.11](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2311.section.md) Release notes)
  - [ ] (Package updates) Added a release notes entry if the change is major or breaking
  - [ ] (Module updates) Added a release notes entry if the change is significant
  - [ ] (Module addition) Added a release notes entry if adding a new NixOS module
- [x] Fits [CONTRIBUTING.md](https://github.com/NixOS/nixpkgs/blob/master/CONTRIBUTING.md).

---

Add a :+1: [reaction] to [pull requests you find important].

[...]

PR #301559, opened 2024-04-04
stem: init at 1.8.2

Tochiaha:
## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [x] `sandbox = true`
- [x] Tested, as applicable:

---

### DIFF
--- a/pkgs/development/tools/stem/default.nix
+++ b/pkgs/development/tools/stem/default.nix
@@ -20,7 +20,7 @@ python3Packages.buildPythonPackage rec {
   meta = with lib; {
     description = "Stem is a Python controller library for Tor";
     homepage = "https://stem.torproject.org/";
-    license = with licenses.gpl3Only;
+    license = with licenses; [ gpl3Only ];
     maintainers = with maintainers; [ tochiaha ];
     mainProgram = "tor-prompt";
   };
--- a/pkgs/development/tools/stem/default.nix
+++ b/pkgs/development/tools/stem/default.nix
@@ -1,0 +1,27 @@
+{ lib
+, fetchFromGitHub
+, python3Packages
+}:
+
+python3Packages.buildPythonPackage rec {
+  pname = "stem";
+  version = "1.8.2";
+  pyproject = true;
+
+  src = fetchFromGitHub {
+    owner = "torproject";
+    repo = "stem";
+    rev = version;
+    hash = "sha256-9BXeE/sVa13jr8G060aWjc49zgDVBhjaR6nt4lSxc0g=";
+  };
+
+  nativeBuildInputs = with python3Packages; [ setuptools ];
+
+  meta = with lib; {
+    description = "Stem is a Python controller library for Tor";
+    homepage = "https://stem.torproject.org/";
+    license = with licenses.gpl3Only;
+    maintainers = with maintainers; [ tochiaha ];
+    mainProgram = "tor-prompt";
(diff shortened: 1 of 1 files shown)


Tochiaha:
#300570

imadnyc:
Hey, why did you close the previous PR and start a new one? I think I'm missing something.

Tochiaha:
> Hey, why did you close the previous PR and start a new one? I think I'm missing something.

i tried the nixpkgs-review pr  on the last one with no output, i was thinking closing and opening a new PR could solve the problem, but apparently it did not.

kirillrdy:
the reason nixpkgs-review doesn't show anything, because nothing is importing `stem`

this was reason why in previous PR, i asked, how are you testing it ?

kirillrdy:
some additional context. orignally stem was being added to by-name hierarchy, which automatically imports file

then there was discussion about stem being used as python library, which is why move to pks/development was made.

if this is indeed python library it will need to be placed in `pkgs/development/python-modules/` and imported in pkgs/top-level/python-packages

it all depends on how stem was meant to be used. @Tochiaha are you using stem for anything ? if so, how are you using it

Tochiaha:
It can be used as a standalone package and python library for flexibility we can do both but i use it as a standalone.

kirillrdy:
> @kirillrdy It can be used as a standalone package and python library for flexibility we can do both but i use it as a standalone.

1. if you want it to be ony used as an executable application, you can just add it to by-name

2. if you want to use it as python library then you need to add it to pkgs/development/python-modules/ and imported in pkgs/top-level/python-packages

if we want both, then we need to do option 2 and in by-name use [toApplication](https://nixos.org/manual/nixpkgs/stable/#topythonapplication-function)

Tochiaha:
i just checked it already has the python library option. maybe option 1 will do.

imadnyc:
Oh wow, I didn't even notice that lmao. I think then that there's no reason for this to be added as an executable, since this seems to be a library, not a standalone application no?

kirillrdy:
@imadnyc thanks for that find, we need some more automated way of checking for duplicates

imadnyc:
@kirillrdy I think discussing this may be too offtopic for here, but I can't see a way to do that. There are packages that have the same name but might be different packages and so shouldn't be auto-rejected. And checking say the source URL might be weird if it's a diff version or if it's a mirror. I think the current system is generally fine as it is, though if there's a way to do that that I'm not seeing I think making an issue would be appreciated.

kirillrdy:
> @kirillrdy I think discussing this may be too offtopic for here, but I can't see a way to do that. There are packages that have the same name but might be different packages and so shouldn't be auto-rejected. And checking say the source URL might be weird if it's a diff version or if it's a mirror. I think the current system is generally fine as it is, though if there's a way to do that that I'm not seeing I think making an issue would be appreciated.

@imadnyc yeah, we don't need fully automated, semi is fine.

as for this PR, @Tochiaha tor-prompt binary can be used from `python3Packages.stem` we can also expose it via `python3Packages.stem.toPythonApplication` via by-name

imadnyc:
Given that it's already in nixpkgs, I think it's safe to close. I still appreciate the effort @Tochiaha!

PR #181145, opened 2022-07-11
realmd: init

SohamG:
# Description of changes

Fix #180043 

Status Update 24 Nov 2023: [See this comment](https://github.com/NixOS/nixpkgs/pull/181145#issuecomment-1826176185)

## Things done

- Add realmd package.
- Add necessary nixos module.

# Pending changes:
- [ ] Enable polkit and sssd as a part of our enable option.
- [ ] Fix config file paths (perhaps use /run/current-system)
- [ ] More testing

----

- Built on platform(s)
  - [X] x86_64-linux

PR #301631, opened 2024-04-04
xmlto: use docbook-xml-ns package XSD files (offline mode)

anthonyroussel:
## Description of changes

* Substitute Docbook XSD URLs by docbook-xml-ns package XSD files
  * Required to build documentation of other nixpkgs package relying on gdbus-codegen or xmlto (like [realmd](https://github.com/NixOS/nixpkgs/pull/181145))
* Format with nixfmt-rfc-style
* Add passthru.tests.version

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/tools/typesetting/xmlto/default.nix
+++ b/pkgs/tools/typesetting/xmlto/default.nix
@@ -1,13 +1,26 @@
-{ fetchurl, fetchpatch, lib, stdenv, libxml2, libxslt
-, docbook_xml_dtd_45, docbook_xsl, flex, w3m
-, bash, getopt, makeWrapper }:
+{
+  bash,
+  docbook_xml_dtd_45,
+  docbook_xsl,
+  fetchpatch,
+  fetchurl,
+  flex,
+  getopt,
+  lib,
+  libxml2,
+  libxslt,
+  makeWrapper,
+  stdenv,
+  w3m,
+}:
 
 stdenv.mkDerivation rec {
   pname = "xmlto";
   version = "0.0.28";
+
   src = fetchurl {
     url = "https://releases.pagure.org/${pname}/${pname}-${version}.tar.bz2";
-    sha256 = "0xhj8b2pwp4vhl9y16v3dpxpsakkflfamr191mprzsspg4xdyc0i";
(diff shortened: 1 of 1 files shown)


marsam (approved):

alyssais:
Since c1ae842f7c13b3cfcbccad6da0667e0674ec5cba ("xmlto: use docbook-xml-ns package XSD files (offline mode)"), the dbus-daemon man page is garbled:

Before:

```
OPTIONS
       The following options are supported:

       --config-file=FILE
           Use the given configuration file.

       --fork
           Force the message bus to fork and become a daemon, even if the
           configuration file does not specify that it should. In most contexts
           the configuration file already gets this right, though. This option
           is not supported on Windows.

       --nofork
           Force the message bus not to fork and become a daemon, even if the
           configuration file specifies that it should. On Windows, the
           dbus-daemon never forks, so this option is allowed but does nothing.
```

After:

```
OPTIONS
       The following options are supported:

         .PP --config-file=FILE

           Use the given configuration file.

             .RE
             .PP --fork

               Force the message bus to fork and become a daemon, even if the
               configuration file does not specify that it should. In most
               contexts the configuration file already gets this right, though.
               This option is not supported on Windows.

                 .RE
                 .PP --nofork

                       Force the message bus not to fork and become a daemon,
                   even if the configuration file specifies that it should. On
                   Windows, the dbus-daemon never forks, so this option is
                   allowed but does nothing.
```

Any ideas?

jtojnar:
Fixed in https://github.com/NixOS/nixpkgs/commit/6d24a9cf8384b08625e721c5a288b15a09abf7c8.

I believe we previously relied on `findXMLCatalogs` setup hook so the file would not be downloaded but it required adding `docbook-xsl-nons` to every consumer of `xmlto`, which is probably pointless if every use of `xmlto` requires it as the name would suggest.

In other toolchains where DocBook is only one of many supported formats and not always used, having it looked up through XML catalog at runtime might be desirable to avoid the hefty 22 MiB build time closure increase in cases that do not need `docbook-xsl-nons`.

Issue #301592, opened 2024-04-04
`-Wno-incompatible-function-pointer-types` is used 523 times in fwupd dev environment


tmuehlbacher:
### Describe the bug

The dev environment for fwupd contains an inexplicably high number of the same compiler flags, causing an extreme amount of compiler warnings if using `meson setup build && ninja -C build` inside of it.

### Steps To Reproduce

```shell
$ nix print-dev-env nixpkgs#fwupd | grep -o -- '-Wno-incompatible-function-pointer-types' | wc -l
523
```

To see flood of compiler messages:

```shell
$ nix-shell '<nixpkgs>' -A fwupd
$ eval "${unpackPhase:-unpackPhase}"
$ cd $sourceRoot
$ eval "${patchPhase:-patchPhase}"
$ meson setup --prefix $out $mesonFlags build
$ ninja -C build
```

### Expected behavior

I think that `$NIX_CFLAGS_COMPILE` should not contain duplicate items. I have honestly no clue how this happens from looking at the fwupd package in nixpkgs.

### Notify maintainers

@r-vdp 

### Metadata
Please run `nix-shell -p nix-info --run "nix-info -m"` and paste the result.

```console
[user@system:~]$ nix-shell -p nix-info --run "nix-info -m"
 - system: `"x86_64-linux"`
 - host os: `Linux 6.8.2, NixOS, 24.05 (Uakari), 24.05.20240403.7781caa`
 - multi-user?: `yes`
 - sandbox: `yes`
 - version: `nix-env (Nix) 2.18.2`
 - nixpkgs: `/etc/nix/inputs/nixpkgs`
```

---

Add a :+1: [reaction] to [issues you find important].

[reaction]: https://github.blog/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/
[issues you find important]: https://github.com/NixOS/nixpkgs/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions-%2B1-desc

PR #33015, opened 2017-12-23
nixos/sway: Extend the descriptions and examples

primeos:

PR #301634, opened 2024-04-04
vala: avoid duplicating compiler args

r-vdp:
## Description of changes

See https://github.com/NixOS/nixpkgs/issues/301592

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/development/compilers/vala/setup-hook.sh
+++ b/pkgs/development/compilers/vala/setup-hook.sh
@@ -11,7 +11,7 @@ disable_incompabile_pointer_conversion_warning() {
     # Work around incompatible function pointer conversion errors with clang 16
     # by setting ``-Wno-incompatible-function-pointer-types` in an env hook.
     # See https://gitlab.gnome.org/GNOME/vala/-/issues/1413.
-    NIX_CFLAGS_COMPILE+=" -Wno-incompatible-function-pointer-types"
+    addFlagTo NIX_CFLAGS_COMPILE "-Wno-incompatible-function-pointer-types"
 }
 
 addEnvHooks "$hostOffset" disable_incompabile_pointer_conversion_warning
--- a/pkgs/stdenv/generic/setup.sh
+++ b/pkgs/stdenv/generic/setup.sh
@@ -334,6 +334,35 @@ appendToVar() {
     fi
 }
 
+# Add flag to a given variable if the variable does not yet contain the flag.
+# The variable to add to can be either an array or a string.
+# Usage: addFlagTo NIX_CFLAGS_COMPILE "-Wblabla"
+#        addFlagTo NIX_CFLAGS_COMPILE "-isystem foobar"
+#        addFlagTo NIX_CFLAGS_COMPILE "--foo=\"bar baz\""
+addFlagTo() {
+    local varname="$1"
+    # Use a different name as in appendToVar to avoid a name clash
+    local -n _nameref="$1"
+
+        # check if variable already exist and if it does then do extra checks
+    if declare -p "$1" 2> /dev/null | grep -q '^'; then
+        type="$(declare -p "$1")"
+        if [[ "$type" =~ "declare -A" ]]; then
+            echo "addFlagTo(): ERROR: trying to use addFlagTo on an associative array." >&2
+            return 1
+        fi
+    fi
+
+    shift
+
+    # If the variable already contains the flag, we don't add it again
+    if [[ "${_nameref[*]}" =~ ^(.* )*"${*}"( .*)*$ ]]; then
+        echo "${_nameref}"
+    else
+        appendToVar "${varname}" "${@}"
+    fi
+}
+
 # Accumulate into `flagsArray` the flags from the named variables.
 #
 # If __structuredAttrs, the variables are all treated as arrays
--- a/pkgs/test/stdenv/default.nix
+++ b/pkgs/test/stdenv/default.nix
@@ -96,6 +96,68 @@ let
         '';
       } // extraAttrs);
 
+  testAddFlagsTo = { name, stdenv', extraAttrs ? { } }:
+    stdenv'.mkDerivation
+      ({
+        inherit name;
+        env = {
+          string = "qux";
+        };
+
+        passAsFile = [ "buildCommand" ] ++ lib.optionals (extraAttrs ? extraTest) [ "extraTest" ];
+        buildCommand = ''
+          declare -p string
+          addFlagTo string hello
+          expected="qux hello"
+          [[ "$string" = "$expected" ]] || (echo "'\$string' was not '$expected', got '$string'" && false)
+          addFlagTo string foo
+          addFlagTo string hello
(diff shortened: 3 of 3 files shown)


pkgs/development/compilers/vala/setup-hook.sh
```diff
@@ -11,7 +11,9 @@ disable_incompabile_pointer_conversion_warning() {
     # Work around incompatible function pointer conversion errors with clang 16
     # by setting ``-Wno-incompatible-function-pointer-types` in an env hook.
     # See https://gitlab.gnome.org/GNOME/vala/-/issues/1413.
-    NIX_CFLAGS_COMPILE+=" -Wno-incompatible-function-pointer-types"
+    if echo "${NIX_CFLAGS_COMPILE}" | grep -q -F -w -v -- "-Wno-incompatible-function-pointer-types"; then
+        NIX_CFLAGS_COMPILE+=" -Wno-incompatible-function-pointer-types"
+    fi
```
alyssais, line 16: Perhaps we could avoid the grep subprocess?

```suggestion
addFlag() {
  for flag in $NIX_CFLAGS_COMPILE; do
    if [[ $flag = $1 ]; then
      return
    fi
  done
  NIX_CFLAGS_COMPILE+=" $1"
}

addFlag -Wno-incompatible-function-pointer-types
```

r-vdp: Hmm, this doesn't work well if the flag itself contains spaces though, like ` -isystem <dir>`.
We can get the flag with spaces correctly as a function argument, but then we cannot simply split the existing assembled string on spaces since that will never match. 

Like, how would we detect that `-foo -bar baz -qux` contains `-bar baz`?

We lose this info when we concatenate everything together in a flat string.
A generalisation of the grep invocation is a lot simpler to get right in this case. 

Is it worth trying to do this in pure bash then?

rhendric: We could also save the `grep` with
```suggestion
    if [[ "${NIX_CFLAGS_COMPILE}" != *-Wno-incompatible-function-pointer-types* ]]; then
        NIX_CFLAGS_COMPILE+=" -Wno-incompatible-function-pointer-types"
    fi
```
or similar.

r-vdp: I generalised the approach that @rhendric suggested and added it as a function to setup.sh, together with a test case. It seems to work well, thanks!

r-vdp:
@alyssais @reckenrode would you mind having another look at this?

pkgs/stdenv/generic/setup.sh
```diff
@@ -272,6 +272,34 @@ appendToVar() {
...
+        type="$(declare -p "$1")"
+        if [[ "$type" =~ "declare -A" ]]; then
+            echo "addFlagTo(): ERROR: trying to use addFlagTo on an associative array." >&2
+            return 1
+        fi
+    fi
+
+    shift
+
+    if [[ "${_nameref[*]}" != *"${*}"* ]]; then
```
reckenrode, line 296: This doesn’t appear to work if the flag is a substring of a flag. For example, adding `bar` to a variable containing `--with-bar` would fail.

r-vdp: Right, good catch! I added a test case for this and updated the function to handle this correctly.

r-vdp:
Is there anything I can do to help move this PR forward? It would be nice to get this fix in.

Mindavi:
Did you verify the test cases fail if they break? I find it hard to see whether it'll return a failure code if one of the tests in the middle fail.

Mindavi:
Ok, this LGTM, except that I'd really like this to be split into a `stdenv` and a `vala` commit. Makes it a lot easier to find it when someone may start bisecting.

r-vdp:
@Mindavi right, that makes sense, I splitted the commit.

Mindavi (approved):

pkgs/stdenv/generic/setup.sh
```diff
@@ -334,6 +334,35 @@ appendToVar() {
...
+        if [[ "$type" =~ "declare -A" ]]; then
+            echo "addFlagTo(): ERROR: trying to use addFlagTo on an associative array." >&2
+            return 1
+        fi
+    fi
+
+    shift
+
+    # If the variable already contains the flag, we don't add it again
+    if [[ "${_nameref[*]}" =~ ^(.* )*"${*}"( .*)*$ ]]; then
```
rhendric, line 359:
```suggestion
    if [[ "${_nameref[*]}" =~ (^| )"$2"( |$) ]]; then
```

(Assuming that only one argument is supported and consequently the `shift` above is removed.)

pkgs/development/compilers/vala/setup-hook.sh
```diff
@@ -11,7 +11,7 @@ disable_incompabile_pointer_conversion_warning() {
     # Work around incompatible function pointer conversion errors with clang 16
     # by setting ``-Wno-incompatible-function-pointer-types` in an env hook.
     # See https://gitlab.gnome.org/GNOME/vala/-/issues/1413.
-    NIX_CFLAGS_COMPILE+=" -Wno-incompatible-function-pointer-types"
+    addFlagTo NIX_CFLAGS_COMPILE "-Wno-incompatible-function-pointer-types"
```
rhendric, line 14:
```suggestion
    addFlagTo NIX_CFLAGS_COMPILE -Wno-incompatible-function-pointer-types
```

pkgs/test/stdenv/default.nix
```diff
@@ -255,6 +322,40 @@ in
...
+          declare -p list
+
+          local -a expected=(
+            "a"
+            "b"
+            "hello"
+            "\"--foo bar\""
+            bar
+            "world"
+          )
```
rhendric, line 351:
```suggestion
          local -a expected=(
            a
            b
            hello
            '"--foo bar"'
            bar
            world
          )
```

pkgs/test/stdenv/default.nix
```diff
@@ -255,6 +322,40 @@ in
...
+          for i in {-1..3}; do
+            [[ "''${list[$i]}" == "''${expected[$i]}" ]] || (echo "first element of '\$list' was not '''''${expected[$i]}, got '''''${list[$i]}'" && false)
```
rhendric, line 353: First element?

rhendric (changes requested):

fabianhjr:
hi, from researching a similar issue (lots of warnings being printed about that flag) I think that hook could be removed entirely.

https://github.com/NixOS/nixpkgs/pull/330015

r-vdp:
Closed in favour of #33015.

PR #301704, opened 2024-04-05
treewide: throw on unsupported system src access

ghost:
## Description of changes

throw rather than abort when accessing the src attribute when system is not supported

testing: accessed src attribute with unsupported system and verified throw (caught with tryEval).

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/applications/audio/renoise/default.nix
+++ b/pkgs/applications/audio/renoise/default.nix
@@ -39,7 +39,7 @@ in stdenv.mkDerivation rec {
     releasePath
   else
     let
-      platform = platforms.${stdenv.system};
+      platform = platforms.${stdenv.system} or (throw "unsupported system ${stdenv.hostPlatform.system}");
       urlVersion = lib.replaceStrings [ "." ] [ "_" ] version;
     in fetchurl {
       url =
--- a/pkgs/applications/gis/udig/default.nix
+++ b/pkgs/applications/gis/udig/default.nix
@@ -13,7 +13,7 @@ let
       hash = "sha256-Ihk3InHB3/tEYRqH2ozhokz2GN8Gfig5DJkO/8P1LJs=";
     };
   };
-  src = srcs.${stdenv.hostPlatform.system};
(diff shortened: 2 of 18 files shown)


lolbinarycat:
actually, why not factor this off into a lib function?  that way if we want to make it do something different in the future (like not throwing an error at all) it's less code to change.

additionally, for downloading native binaries, shouldn't we be using targetPlatfrom instead if hostPlatform?  the "cross compilation" logic seems fairly simple in these cases: just download the binaries for the target platform.

a-n-n-a-l-e-e:
> actually, why not factor this off into a lib function? that way if we want to make it do something different in the future (like not throwing an error at all) it's less code to change.

thought about this but i don't think there is too much benefit to a library function other than not having to re-write the message text on every use.  the `or throw` seems standard across nixpkgs and there is not currently a list of common exceptions  / errors so adding `unsupportedSystemError = system: throw "msg txt ${system}";`  would be the first and getting buy in and performing treewide replacement is out of the scope of this change, in my opinion. not sure what the different behavior would be but the throw happens on assignment to various types so if not generating an exception not sure what other correct use case is.   

> 
> additionally, for downloading native binaries, shouldn't we be using targetPlatfrom instead if hostPlatform? the "cross compilation" logic seems fairly simple in these cases: just download the binaries for the target platform.

`hostPlatform` is correct. that is where the binaries run.

https://nix.dev/tutorials/cross-compilation.html#platforms
> When compiling code, we can distinguish between the build platform, where the executable is built, and the host platform, where the compiled executable runs.

lolbinarycat (approved):
approving this because i think it's a good short-term solution, although something like a new trivial build helper for prebuilt binaries might be nice to have in the future.

PR #301712, opened 2024-04-05
incus: 0.7.0 -> 6.0.0, add lts

adamcstephens:
## Description of changes

This also makes the module use the lts by default.

- **incus-lts: init at 6.0.0**
- **incus: 0.7.0 -> 6.0.0**

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/nixos/modules/virtualisation/incus.nix
+++ b/nixos/modules/virtualisation/incus.nix
@@ -109,14 +109,16 @@ in
         {command}`incus` command line tool, among others.
       '';
 
-      package = lib.mkPackageOption pkgs "incus" { };
+      package = lib.mkPackageOption pkgs "incus-lts" { };
 
       lxcPackage = lib.mkPackageOption pkgs "lxc" { };
 
-      clientPackage = lib.mkPackageOption pkgs [
-        "incus"
-        "client"
-      ] { };
+      clientPackage = lib.mkOption {
+        type = lib.types.package;
+        default = cfg.package.client;
+        defaultText = lib.literalExpression "config.virtualisation.incus.package.client";
+        description = "The incus client package to use. This package is added to PATH.";
+      };
 
       preseed = lib.mkOption {
         type = lib.types.nullOr (lib.types.submodule { freeformType = preseedFormat.type; });
--- a/nixos/doc/manual/release-notes/rl-2405.section.md
+++ b/nixos/doc/manual/release-notes/rl-2405.section.md
@@ -65,6 +65,8 @@ Use `services.pipewire.extraConfig` or `services.pipewire.configPackages` for Pi
   }
   ```
 
+- The initial Incus LTS release (v6.0.x) is now available through `virtualisation.incus` as the default. Users who wish to continue using the non-LTS release will need to set `virtualisation.incus.package = pkgs.incus`. Stable release users are encouraged to stay on the LTS release as non-LTS releases will by default not be backported.
+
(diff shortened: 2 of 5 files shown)


nixos/modules/virtualisation/incus.nix
```diff
@@ -109,7 +109,7 @@ in
         {command}`incus` command line tool, among others.
       '';
 
-      package = lib.mkPackageOption pkgs "incus" { };
```
mkg20001, line 112: not sure if we should make the default depend on system.stateVersion, so below 24.05 it defaults to non lts

adamcstephens: This PR is also upgrading the non-lts to the same version as the lts, so I don't think it matters?

mkg20001: but it will diverge at some point in time. the idea here is to not suddenly throw people on the lts path.

adamcstephens: Well there was no LTS before, but if you feel strongly that existing users shouldn't be moved to the LTS train then I'll condition it.

mkg20001: we could also add a warning asking users to pick a default when migrating from 23.11?

adamcstephens: I just tried to add a warning, but I can't find a good way to add a warning if this is unset for `system.stateVersion < "24.05"` while also allowing users to dismiss this warning by explicitly setting `virtualisation.incus.package`. In theory we could add another option that they set, but that's an awkward UX.

I see there are two options then:
1. Fail evaluation if it's unset and they are <24.05. This isn't a great experience either since anybody upgrading or converting an older system will have to set this.
2. Just default everyone to `incus-lts`. Some unstable users may want to track the more frequent releases.

I'm back to the opinion of changing the default to LTS and just updating the release notes accordingly. If you have a solution for providing a warning, that won't fail eval, and can be easily cleared by setting `package`, I'd be open to reassessing. 

I want LTS to be the default moving forward though, given that we may not be backporting the non-LTS releases to stable. I am even considering putting a warn on the non-LTS packages in stable encouraging users to use LTS.

mkg20001: maybe builtins.trace together with an if could be used to produce a warning, but if it's this much trouble just add it to the release notes as breaking change.

adamcstephens: Do you consider this a breaking change? The impact is slowing down some user's release train, but it shouldn't break anyone. Right now it's even the same release as non-LTS.

mkg20001: whether it breaks anything is a wildcard then i suppose, but it's at least a notable change

>I want LTS to be the default moving forward though, given that we may not be backporting the non-LTS releases to stable. I am even considering putting a warn on the non-LTS packages in stable encouraging users to use LTS.

maybe a reworded version of this as part of release notes so everyone knows what's going on

adamcstephens: Done. Thanks for your feedback.

mkg20001 (approved):

adamcstephens:
@ofborg eval

Issue #257172, opened 2023-09-25
nixpkgs.dockerTools: missing /tmp 


fionera:
### Describe the bug
When running a FHSUserEnv inside a nix docker container, it fails to start as /tmp is missing by default.
As a fix the docker images could be provided with a  /tmp directory.

### Steps To Reproduce
1. Create a buildNixShellImage
2. Try using a FHSUserEnv
3. `bwrap: Failed to mount tmpfs: No such file or directory`

### Expected behavior
It should not throw an error

### Notify maintainers
@infinisil

kjeremy:
I do the following in my containers:

```nix
extraCommands = ''
  mkdir -p tmp
'';
```

infinisil:
Sounds like a good idea to provide this by default, feel free to open a PR for this!

ryota-ka:
Would it possible be reasonable to apply this also to `pkgs.dockerTools.buildLayeredImage`?

infinisil:
If it suffers from the same problem, I can't see why not :)

ryota-ka:
@infinisil Since 23.11, IIUC, it seems that making the `/tmp` directory is somewhat difficult when you use `dockerTools.{build,stream}LayeredImage`:

- w/ `extraCommands`
  - fails with `mkdir: cannot create directory '/tmp': File exists`
- w/ `fakeRootCommands`, `enableFakechroot = false`
  - `mkdir` itself succeeds but a container from the build image does NOT have `/tmp`
- w/ `fakeRootCommands`, `enableFakechroot = true`
  - fails with `tar: ./dev/shm/lttng-ust-wait-8-0: Cannot open: Permission denied` under a certain condition

I've opened #301755 to fix the problem, but I'm not sure the solution is appropriate as I'm not very familiar with the way we build a Docker image with Nix.
I would appreciate it if you could review it.

ryota-ka:
@roberth Thank you for reviewing #301755!

After some experiments, I've found that the following simple snippet just does the trick.

```nix
let
  tmp = pkgs.runCommand "tmp" { } ''
    mkdir $out
    mkdir -m 1777 $out/tmp
  '';
in
dockerTools.buildLayeredImage {
  contents = [
    tmp
  ];
}
```

Do you think it reasonable to add a trivial derivation under `pkgs.dockerTools`?
(a bikeshed problem: should it be named as `tmp`, `tmpdir`, or `tmpDir`?)

infinisil:
[...]

PR #260063, opened 2023-10-09
dockerTools: create /tmp in rootLayer

fionera:
## Description of changes

Create a /tmp directory inside the docker root layer. Closes  #257172

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
- [ ] For non-Linux: Is `sandbox = true` set in `nix.conf`? (See [Nix manual](https://nixos.org/manual/nix/stable/command-ref/conf-file.html))
- [ ] Tested, as applicable:
  - [NixOS test(s)](https://nixos.org/manual/nixos/unstable/index.html#sec-nixos-tests) (look inside [nixos/tests](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests))
  - and/or [package tests](https://nixos.org/manual/nixpkgs/unstable/#sec-package-tests)
  - or, for functions and "core" functionality, tests in [lib/tests](https://github.com/NixOS/nixpkgs/blob/master/lib/tests) or [pkgs/test](https://github.com/NixOS/nixpkgs/blob/master/pkgs/test)
  - made sure NixOS tests are [linked](https://nixos.org/manual/nixpkgs/unstable/#ssec-nixos-tests-linking) to the relevant packages
- [ ] Tested compilation of all packages that depend on this change using `nix-shell -p nixpkgs-review --run "nixpkgs-review rev HEAD"`. Note: all changes have to be committed, also see [nixpkgs-review usage](https://github.com/Mic92/nixpkgs-review#usage)
- [ ] Tested basic functionality of all binary files (usually in `./result/bin/`)
- [23.11 Release Notes](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2311.section.md) (or backporting [23.05 Release notes](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2305.section.md))
  - [ ] (Package updates) Added a release notes entry if the change is major or breaking
  - [ ] (Module updates) Added a release notes entry if the change is significant
  - [ ] (Module addition) Added a release notes entry if adding a new NixOS module
- [x] Fits [CONTRIBUTING.md](https://github.com/NixOS/nixpkgs/blob/master/CONTRIBUTING.md).

PR #301755, opened 2024-04-05
dockerTools.streamLayeredImage: make `/tmp`

ryota-ka:
## Description of changes

Make `/tmp` directory while building a Docker image with `pkgs.dockerTools.streamLayeredImage`.
(see https://github.com/NixOS/nixpkgs/issues/257172#issuecomment-2039171878 for the motivation)

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [x] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/build-support/docker/default.nix
+++ b/pkgs/build-support/docker/default.nix
@@ -949,6 +949,7 @@ rec {
           ];
           postBuild = ''
             mv $out old_out
+            mkdir -m 1777 old_out/tmp
             (cd old_out; eval "$extraCommands" )
 
             mkdir $out


roberth (changes requested):
This changes all images to have a `/tmp`, even ones that don't need them.
Each unnecessary "capability" reduces the security of an image, and for a number of images, this will be the first writable directory.

Furthermore if we were to go down this path, `streamLayeredImage` would have to grow an unbounded number of features, eventually encompassing a large part of core NixOS. This would entail a loss of efficiency and flexibility, because the lower level primitive operation would be gone.

I'd prefer a solution where we either reuse modules of NixOS (for a really high-level and standardized interface), or something like `shadowSetup` where it's the caller's responsibility to pass it in.

ryota-ka:
@roberth Unlike the files like `/etc/{group,passwd}` that is created with `shadowSetup`, the existence of `/tmp` is considered essential in POSIX/Linux systems.
Here are some facts:

- The `/tmp` directory *shall exist* on POSIX-conforming systems, according to IEEE Std 1003.1-2017[^1]
- Linux Foundation stipulates that *the `/tmp` directory must be made available for programs that require temporary files.* in File System Hierarchy 3.0[^2]

So in my opinion it isn't one of such feature requests but a bug fix.

Here, two questions seem to be clarified:

- a) Should `pkgs.dockerTools` continue to ship POSIX-incompatible Docker images lacking the `/tmp` directory?
- b) If it is undesirable that all the image have the `/tmp` directory in the perspective of security, should #260063 be also reverted?

[^1]: https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap10.html#tag_10_01
[^2]: https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch03s18.html

roberth:
Containers are not required to comply with anything except their runtime's requirements and the requirements of the application.

> * a) Should `pkgs.dockerTools` continue to ship POSIX-incompatible Docker images lacking the `/tmp` directory?

Yes, but _you_ are responsible for the shipping part. `streamLayeredImage` is just a tool, which is at the bottom of the "stack", so it should be flexible rather than opinionated.


> * b) If it is undesirable that all the image have the `/tmp` directory in the perspective of security, should

Thank you for bringing this up. I would prefer for it to be removed. I'll cross-post.

ryota-ka:
Okay, then I'll closing this. Let's keep the discussion in #257172.
Anyway, thank you for your review and being a good shepherd!

PR #301771, opened 2024-04-05
nixos/paperless: refactor to use systemd LoadCredential, Switch to systemd.tmpfiles.settings

### LINKED ISSUE #301746: nixos/paperless: tmpfiles.rules errors when when using NFS mounts for `mediaDir` or `consumptionDir`
### Describe the bug

I want to use `services.paperless` but with `mediaDir` and `consumptionDir` residing on an NFS mounted path.

My NFS server is secured to only allow access from `services.paperless.user`. I also have `root_squash` enabled on my NFS server, as is the recommended best practice.

But the paperless module uses tmpfiles to ensure those directories are created, but root does not have permission to create files in the nfs share.

```
    # example
    systemd.services.paperless-scheduler.after = ["mnt-mali-tank2-services-paperless.mount"];
    systemd.services.paperless-scheduler.bindsTo = ["mnt-mali-tank2-services-paperless.mount"];
```

Result:

```
...
setting up tmpfiles
Failed to create directory or subvolume "/mnt/mali/tank2/services/paperless/media": Permission denied
Failed to create directory or subvolume "/mnt/mali/tank2/services/paperless/consume": Permission denied
```

To be clear the nfs mounted path is `/mnt/mali/tank2/services/paperless/` so `/media` and `/consume` are inside the mount.

Notably the paperless service still works, if I ensure those directories exist myself. But the tmpfiles rules setup service still errors out.

One workaround I *thought* might work would be to set `mediaDir` and `consumptionDir` to `null` and then set those env vars myself. However the NixOS module requires those two options to be set.

Possible fixes:

1. The sub-directories could be created inside a systemd service that is running as the user.
    * this wouldn't work without some shenangiams because if the dir is something like `/var/lib/paperless/media` the user won't have permissions to make `/var/lib/paperless/`
3. Make `mediaDir` and `consumptionDir` optional so the admin can manage setting them
     * this isn't ideal because then the user also has to muck with `ReadWritePaths=`
4. Add a flag to bypass the systemd.tmpfiles.rules and let the admin manage it themselves.
     * this my favored option

I'm happy to open a PR if the maintainers like one of those options.

### Steps To Reproduce
1. Setup `services.paperless` with `mediaDir` and `consumptionDir` on nfs monts
5. ensure those directories don't exist in the nfs mount

### Expected behavior

`rebuild switch` should work without error.

### Notify maintainers

@Flakebi 
@leona-ya 
@erikarvstedt

### Metadata

```console
 - system: `"x86_64-linux"`
 - host os: `Linux 6.6.14, NixOS, 23.11 (Tapir), 23.11.20240403.1487bde`
 - multi-user?: `yes`
 - sandbox: `yes`
 - nixpkgs: `/etc/nixpkgs/channels/nixpkgs`
```

---

Ramblurr:
## Description of changes

This PR contains two commits

1. > nixos/paperless: refactor to use systemd LoadCredential
    
    This replaces the paperless-copy-password service with the use of
    systemd's LoadCredential mechanism.
    
    It is not a breaking change since it is gated behind `cfg.passwordFile`.

3. > nixos/paperless: Add opt `services.paperless.createDirs`
    
    This option resolves #301746 by allowing the admin to bypass the
    creation of the paperless directories by systemd-tmpfiles.
    
    This is necessary when, for example, those directories lie inside an NFS
    mount that the root user does not have rw access to.

    It is also not a breaking change because the default behavior is the old behavior.

Both changes are covered in the test suite.

@Flakebi
@leona-ya
@erikarvstedt

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [x] Tested, as applicable:

---

### DIFF
--- a/nixos/modules/services/misc/paperless.nix
+++ b/nixos/modules/services/misc/paperless.nix
@@ -238,6 +238,7 @@ in
         User = cfg.user;
         ExecStart = "${pkg}/bin/celery --app paperless beat --loglevel INFO";
         Restart = "on-failure";
+        LoadCredential = lib.optionalString (cfg.passwordFile != null) "PAPERLESS_ADMIN_PASSWORD:${cfg.passwordFile}";
       };
       environment = env;
 
@@ -270,7 +271,7 @@ in
       ''
       + optionalString (cfg.passwordFile != null) ''
         export PAPERLESS_ADMIN_USER="''${PAPERLESS_ADMIN_USER:-admin}"
-        export PAPERLESS_ADMIN_PASSWORD=$(cat "${cfg.dataDir}/superuser-password")
+        export PAPERLESS_ADMIN_PASSWORD=$(cat $CREDENTIALS_DIRECTORY/PAPERLESS_ADMIN_PASSWORD)
         superuserState="$PAPERLESS_ADMIN_USER:$PAPERLESS_ADMIN_PASSWORD"
         superuserStateFile="${cfg.dataDir}/superuser-state"
 
@@ -296,19 +297,6 @@ in
         PrivateNetwork = false;
       };
       environment = env;
-    };
-
-    # Reading the user-provided password file requires root access
-    systemd.services.paperless-copy-password = mkIf (cfg.passwordFile != null) {
-      requiredBy = [ "paperless-scheduler.service" ];
-      before = [ "paperless-scheduler.service" ];
-      serviceConfig = {
-        ExecStart = ''
-          ${pkgs.coreutils}/bin/install --mode 600 --owner '${cfg.user}' --compare \
-            '${cfg.passwordFile}' '${cfg.dataDir}/superuser-password'
-        '';
-        Type = "oneshot";
-      };
     };
 
     systemd.services.paperless-consumer = {
--- a/nixos/modules/services/misc/paperless.nix
+++ b/nixos/modules/services/misc/paperless.nix
@@ -220,15 +220,16 @@ in
   config = mkIf cfg.enable {
     services.redis.servers.paperless.enable = mkIf enableRedis true;
 
-    systemd.tmpfiles.rules = [
-      "d '${cfg.dataDir}' - ${cfg.user} ${config.users.users.${cfg.user}.group} - -"
-      "d '${cfg.mediaDir}' - ${cfg.user} ${config.users.users.${cfg.user}.group} - -"
-      (if cfg.consumptionDirIsPublic then
-        "d '${cfg.consumptionDir}' 777 - - - -"
-      else
-        "d '${cfg.consumptionDir}' - ${cfg.user} ${config.users.users.${cfg.user}.group} - -"
-      )
-    ];
+    systemd.tmpfiles.settings."10-paperless" = let
+      defaultRule = {
+        inherit (cfg) user;
+        inherit (config.users.users.${cfg.user}) group;
+      };
+    in {
+      "${cfg.dataDir}".d = defaultRule;
+      "${cfg.mediaDir}".d = defaultRule;
+      "${cfg.consumptionDir}".d = if cfg.consumptionDirIsPublic then { mode = "777"; } else defaultRule;
+    };
 
     systemd.services.paperless-scheduler = {
       description = "Paperless Celery Beat";


nixos/tests/paperless.nix
```diff
@@ -85,5 +91,13 @@ import ./make-test-python.nix ({ lib, ... }: {
     simple.send_monitor_command("quit")
     simple.wait_for_shutdown()
     test_paperless(postgres)
+    postgres.send_monitor_command("quit")
+    postgres.wait_for_shutdown()
+
+    with subtest("When directory creation is disabled, the service should fail to start"):
+        noCreateDirs.sleep(10)
```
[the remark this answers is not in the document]
ambroisie: Writing your own `wait_for_failed_unit` (based on `wait_for_unit`) should work well enough?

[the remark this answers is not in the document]
Ramblurr: Should I add it to the shared wait_for helpers?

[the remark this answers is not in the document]
ambroisie: I'll defer to owners of the test runner for this :shrug:.

nixos/modules/services/misc/paperless.nix
```diff
@@ -134,6 +134,17 @@ in
       description = lib.mdDoc "Whether all users can write to the consumption dir.";
     };
 
+    createDirs = mkOption {
```
SuperSandro2000, line 137: I would recommend instead to use mkForce for the specific paths in tmpfilesd as this option is very specific to your exact setup and not generally useful IMO.

erikarvstedt: I guess what @SuperSandro2000 means is to convert `tmpfiles.rules` to `tmpfiles.settings`, like so:

```nix
systemd.tmpfiles.settings."10-paperless" = let
  defaultRule = {
    inherit (cfg) user;
    inherit (config.users.users.${cfg.user}) group;
  };
in {
  "${cfg.dataDir}".d = defaultRule;
  "${cfg.mediaDir}".d = defaultRule;
  "${cfg.consumptionDir}".d = if cfg.consumptionDirIsPublic then { mode = "777"; } else defaultRule;
};
```

Then you can use the module system's builtin features to disable dir creation:
```nix
systemd.tmpfiles.settings."10-paperless" = mkForce {};
```

Yes, this approach is better than an extra option!

Ramblurr: great, thanks for the feedback. I'll refactor the PR

nixos/tests/paperless.nix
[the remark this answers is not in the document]
Ramblurr: Added.

nixos/modules/services/misc/paperless.nix
Ramblurr: Refactored, thanks for the feedback.

erikarvstedt (changes requested):
Test `When directory creation is disabled, the service should fail to start` can be removed.
This property trivially holds true for all services that use dirs created via `systemd.tmpfiles` and doesn't need to be tested.

Ramblurr:
> Test `When directory creation is disabled, the service should fail to start` can be removed. This property trivially holds true for all services that use dirs created via `systemd.tmpfiles` and doesn't need to be tested.

Test removed!

nixos/modules/services/misc/paperless.nix
```diff
@@ -220,15 +220,16 @@ in
...
-      "d '${cfg.mediaDir}' - ${cfg.user} ${config.users.users.${cfg.user}.group} - -"
-      (if cfg.consumptionDirIsPublic then
-        "d '${cfg.consumptionDir}' 777 - - - -"
-      else
-        "d '${cfg.consumptionDir}' - ${cfg.user} ${config.users.users.${cfg.user}.group} - -"
-      )
-    ];
+    systemd.tmpfiles.settings."10-paperless" = let
+      defaultRule = {
+        inherit (cfg) user;
```
SuperSandro2000, line 225:
```suggestion
```
seems to be unused

erikarvstedt:
Can you rebase this to master to fix `nixosTests.paperless`?
Currently, dependency `python3.11-moto-5.0.3` fails to build.

Ramblurr:
Done.

erikarvstedt:
The `moto` build failure only appeared when building `nixosTests.paperless` with https://github.com/NixOS/nix/pull/6530. It was unrelated to the rebase.

erikarvstedt (approved):
Nice work!

- [x] `nixosTests.paperless` succeeds on `x86_64-linux`

SuperSandro2000 (approved):

leona-ya (approved):
This is nice, thank you!

PR #301772, opened 2024-04-05
nixos/test-driver: Add Junit XML report creation

hertrste:
## Description of changes

This MR adds the possibility to generate Junit XML reports when running NixOS integration tests. Reports can be generated by adding `--junit-xml=<outfile>` to the NixOS test driver invocation. These reports can be evaluated by various tools e.g. gitlab and could help identify failing tests easier and get the corresponding log output of the failing test faster.

## Things done

The logging facilities of the NixOS test driver are refactored. There is an AbstractLogger interface and the existing logger is split into terminal and XML logging. A CompositeLogger allows the composition of multiple loggers. This allows restoring the old logging behavior of logging to the terminal and XML simultaneously. It was taken care that the existing logging is not changed at all, in order to not break anything for users relying on the format of the existing logging output.

Further, a JunitXML logger is added which can be activated via arguments to the test driver invocation. The new logger allows the creation of JunitXML reports listing success or failure of the subtests of the integration test. This allows for quick evaluation which subtest has failed.

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/nixos/lib/test-driver/default.nix
+++ b/nixos/lib/test-driver/default.nix
@@ -24,6 +24,7 @@ python3Packages.buildPythonApplication {
     coreutils
     netpbm
     python3Packages.colorama
+    python3Packages.junit-xml
     python3Packages.ptpython
     qemu_pkg
     socat
--- a/nixos/lib/test-driver/pyproject.toml
+++ b/nixos/lib/test-driver/pyproject.toml
@@ -31,6 +31,10 @@ ignore_missing_imports = true
 module = "ptpython.*"
 ignore_missing_imports = true
 
+[[tool.mypy.overrides]]
+module = "junit_xml.*"
+ignore_missing_imports = true
+
 [tool.black]
 line-length = 88
 target-version = ['py39']
--- a/nixos/lib/test-driver/test_driver/__init__.py
+++ b/nixos/lib/test-driver/test_driver/__init__.py
@@ -6,7 +6,12 @@ from pathlib import Path
 import ptpython.repl
 
 from test_driver.driver import Driver
-from test_driver.logger import JunitXMLLogger, XMLLogger, rootlog
+from test_driver.logger import (
+    CompositeLogger,
+    JunitXMLLogger,
+    TerminalLogger,
+    XMLLogger,
+)
 
 
 class EnvDefault(argparse.Action):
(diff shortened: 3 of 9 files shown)


tfc:
When we emailed about this topic, i suggested an implementation that follows the composite pattern. I apologize if we had a misunderstanding as this could be the result of me expressing myself clumsily, but the code in this PR does not really reflect the discussion from my perspective.

The concrete steps to follow the pattern, that is best to extend in the future, would be:

1. Create an abstract logger class
2. Implement the original logger in terms of this interface
3. Create an XML logger in terms of this interface
4. Create a composite logger in terms of this interface. This logger would simply keep a list of loggers and forward everything it gets to those
5. The test driver composes random combinations of loggers inside the composite logger, led by command line parameters

This has the following advantages:

- The logger class does not even have to know what XML is
- Because of this, the test driver would not need to include xml libs by default and additional supported loggers can be defined *after* the original logger implementation, as well as in different files/projects
  - Including XML alone does not hurt. Right now. But with these code changes we open the door towards more different logger implementations, and I'd like to make this easy to extend for other users who have different logging requirements.
- We can have a future where companies who use the test driver, implement their private logger implementations at home and simply inject them without the upstream implementation knowing about it.

Please also note that there are good and configurable/composable production logging frameworks in Python that already implement this logic. We might be duplicating this work with this effort, so we should have a little bit of analysis what existing library should be reused for this.

hertrste:
I have refactored the logging facilities to simply allow new loggers and composing them. I have also separated the concerns of logging to the terminal and the old XML creation.

I would argue against using external logging libraries because of our special requirements regarding the subtest and nested calls. Further, I want to avoid changing the existing log output as people may rely on it.

Please have another look.

tfc:
This is a great improvement. I'd love to see one last change: 

`AbstractLogger` that is subclassed by `TerminalLogger`, `XMLLogger`, and `JUnitXMLLogger`, does not have to have the same interface as `CompositeLogger`. If `AbstractLogger` implementations keep implementing the *bare minimum* of functionality (e.g. only `nested` but not `subtest`), we can enrich the CompositeLogger with all the extra non-atomic functionality and end up with less duplicate logic.

Can you please do that?

tfc:
Whoa i did not know that there was no setup.py before... this PR makes no sense. need to investigate

hertrste:
> AbstractLogger that is subclassed by TerminalLogger, XMLLogger, and JUnitXMLLogger, does not have to have the same interface as CompositeLogger. If AbstractLogger implementations keep implementing the bare minimum of functionality (e.g. only nested but not subtest), we can enrich the CompositeLogger with all the extra non-atomic functionality to build on top of the abstract core functionality and end up with less duplicate logic.

I simplified the `AbstractLogger` interface a bit by removing the `print_serial_logs(self, enable: bool)` function which is now only available in the `CompositeLogger`.

Unfortunately, I think removing the `subtest` call from the AbstractLogger is not possible, because the new `JunitXMLLogger` needs to differentiate between `subtest` and `nested` calls.
It would be possible to extend the `nested`  function to receive the type of nesting e.g. `nested(message, NestedType.SUBTEST)`, `nested(message, NestedType.GENERIC)` but I am not sure if this increases readability or improves the API instead of having the 2 function calls separately.

tfc:
@ofborg test login

tfc (approved):

nikstur:
@wegank since this shouldn't be a breaking change It should be fine to merge before the branch off. Do you disagree? Otherwise I'm inclined to merge.

wegank (approved):

messemar (approved):

PR #301890, opened 2024-04-05
ldproxy: init at 0.31.4

vpochapuis:
## Description of changes

Adding the [`ldproxy`](https://github.com/esp-rs/embuild/tree/master/ldproxy) rust linker tool. (At the current lastest version as of the moment of writing this PR, hence `0.31.4`).
This tool is presented as a crate within a bigger workspace parent project called [`embuild`](https://github.com/esp-rs/embuild/tree/master).
However we are only building the inner crate containing `ldproxy` which is usually installed in rust using `cargo install ldproxy`.
This tool is used in the building of embedded platforms using `esp-rs` and the `std` crate, like for the `esp32c3`.

(This package introduction was previously attempted in #190267 by @matthiasbeyer and is mostly based on it).

## Call for help

This is my first contribution to the `nixpkgs`.
I would like first to thank all the maintainers of this project which lets us have the chance to aim for reproducible builds in our daily work and personal projects. 
I have been introduced very recently to `nix`, `nixpks` and `NixOs` in my quest to automate builds of my Rust packages as well as easily reproduce my desktop and server environments.
The learning curve is steep as it was for me for Rust, the learning materials are numerous however I have found it hard to find one place to easily learn from scratch all the various concepts while seeing tangible examples of common and relatable use cases. I wish in my learning journey I could have a better view of the project and maybe one day help new comers to be introduced to this amazing ecosystem. 
As this is my first attempt of contribution to this project after a [recommendation](https://github.com/NixOS/nixpkgs/pull/190267#issuecomment-2035260188) to do so (thank you @doronbehar), I tried to read the different resources available (official website then got redirected to the README.md and CONTRIBUTING.md), and I have to say it was the most complicated process I have encountered yet in my contribution adventure (I am quite new to open-source contributions but I would like to make this a more common occurrence as I would like to give back to the communities that have already given so much to everyone).
Hence I might have made some mistakes and I hope I could be helped in improving this PR and make it up to the standards of this community.

Thank you very much for your time and attention.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:
  - [ ] Help needed: after running the command got `error: build log of '/nix/store/5dixzydh8bfsr68zbdimqmns4mz8sw00-ldproxy-0.31.4.drv^*' is not available`

---

### DIFF
--- a/pkgs/by-name/ld/ldproxy/package.nix
+++ b/pkgs/by-name/ld/ldproxy/package.nix
@@ -1,0 +1,29 @@
+{ lib, rustPlatform, fetchFromGitHub }:
+
+rustPlatform.buildRustPackage rec {
+  pname = "ldproxy";
+  version = "0.31.4";
+
+  # The source of this tools actually comes from the set of tools located in the `esp-rs:embuild` repository
+  # cf: https://github.com/esp-rs/embuild
+  src = fetchFromGitHub {
+    owner = "esp-rs";
+    repo = "embuild";
+    rev = "v${version}";
+    hash = "sha256-YH2CPb3uBlPncd+KkP25xhCVvDB7HDxJuSqWOJ1LT3k=";
+  };
+
+  cargoHash = "sha256-HobX/aBn10KPpUKRSLt2IvfFVW7YpOu5tX4zNSCz6tw=";
+
+  # However we are only interested in building the specific crate located at `/ldproxy`
+  # cf: https://github.com/esp-rs/embuild/tree/v0.31.4/ldproxy
+  buildAndTestSubdir = "ldproxy";
+
+  meta = with lib; {
+    description = "Linker Proxy: a simple tool to forward linker arguments to the actual linker executable";
+    homepage = "https://github.com/esp-rs/embuild";
+    changelog = "https://github.com/esp-rs/embuild/blob/v${version}/CHANGELOG.md";
+    license = with licenses; [ mit /* or */ asl20 ];
+    maintainers = with maintainers; [ vpochapuis ];
+  };
+}
--- a/maintainers/maintainer-list.nix
+++ b/maintainers/maintainer-list.nix
@@ -21006,6 +21006,12 @@
     githubId = 357664;
     name = "Viktor Petersson";
   };
+  vpochapuis = {
+    email = "vincent.professional@chapuis.ovh";
+    github = "vpochapuis";
+    githubId = 75721408;
+    name = "Vincent Chapuis";
+  };
   vq = {
     email = "vq@erq.se";
     github = "vq";


pkgs/by-name/ld/ldproxy/package.nix
```diff
@@ -0,0 +1,38 @@
+{ lib, rustPlatform, fetchFromGitHub, pkg-config, udev, stdenv }:
+
+# For more information on using `buildRustPackage` please take a look at:
+# https://nixos.org/manual/nixpkgs/stable/#compiling-rust-applications-with-cargo
```
doronbehar, line 4: This is too obvious.

pkgs/by-name/ld/ldproxy/package.nix
```diff
@@ -0,0 +1,38 @@
...
+
+  buildInputs = [
+    udev
+  ];
+
+  cargoSha256 = "sha256-dpdd4Yr92DrKeKrXx/OBkUQVo+bCNr+9VwXI7tvLAaU=";
+
+  meta = with lib; {
+    description = "Linker Proxy: a simple tool to forward linker arguments to the actual linker executable";
+    homepage = "https://github.com/esp-rs/embuild/tree/v0.31.4/ldproxy";
```
doronbehar, line 34:
```suggestion
    homepage = "https://github.com/esp-rs/embuild/tree/v${version}/ldproxy";
```

doronbehar (changes requested):
Looks great besides these two small commits. Welcome aboard!

doronbehar:
Note that CI reports the build to fail to fetch the `src`. It seems the `version` is wrong. https://github.com/esp-rs/embuild/releases/tag/ldproxy-v0.3.2

vpochapuis:
> Note that CI reports the build to fail to fetch the `src`. It seems the `version` is wrong. https://github.com/esp-rs/embuild/releases/tag/ldproxy-v0.3.2

Hello~
Thank you for taking a look at this PR!
Indeed the version is wrong, by wanting to stay as close as possible to the original PR i missed the fact that `embuild` now uses `tags` instead of previously using `releases`. (https://github.com/esp-rs/embuild/releases/tag/v0.31.4).
I will correct this.

pkgs/by-name/ld/ldproxy/package.nix
vpochapuis: Potentially fixed in 2668391e4ad8d4737bc4327fab3031ee2f27e6bf

pkgs/by-name/ld/ldproxy/package.nix
vpochapuis: Potentially fixed in 2668391e4ad8d4737bc4327fab3031ee2f27e6bf

pkgs/by-name/ld/ldproxy/package.nix
```diff
@@ -0,0 +1,36 @@
...
+
+  buildInputs = [
+    udev
+  ];
+
+  cargoSha256 = "sha256-dpdd4Yr92DrKeKrXx/OBkUQVo+bCNr+9VwXI7tvLAaU=";
+
+  meta = with lib; {
+    description = "Linker Proxy: a simple tool to forward linker arguments to the actual linker executable";
+    homepage = "https://github.com/esp-rs/embuild/tree/v${version}/ldproxy";
```
doronbehar, line 32: Usually this is simply set to:

```suggestion
    homepage = "https://github.com/esp-rs/embuild";
```

Usually people use a version specific URL for the changelog:

```suggestion
    homepage = "https://github.com/esp-rs/embuild";
    changelog = "https://github.com/esp-rs/embuild/blob/v${version}/CHANGELOG.md";
```

pkgs/by-name/ld/ldproxy/package.nix
```diff
@@ -0,0 +1,36 @@
...
+
+  nativeBuildInputs = [
+    pkg-config
+  ];
+
+  buildInputs = [
+    udev
+  ];
+
+  cargoSha256 = "sha256-dpdd4Yr92DrKeKrXx/OBkUQVo+bCNr+9VwXI7tvLAaU=";
```
doronbehar, line 28: Personally, I think it is best to put this attribute near the `src`, so that when a version bump comes, hashes are updated in the same diff, making it a bit smaller and easier to review. Also:

```suggestion
  cargoHash = "sha256-dpdd4Yr92DrKeKrXx/OBkUQVo+bCNr+9VwXI7tvLAaU=";
```

pkgs/by-name/ld/ldproxy/package.nix
```diff
@@ -0,0 +1,36 @@
...
+  pname = "ldproxy";
+  version = "0.31.4";
+
+  # The source of this tools actually comes from the set of tools located in the `esp-rs:embuild` repository
+  # cf: https://github.com/esp-rs/embuild
+  src = fetchFromGitHub {
+    owner = "esp-rs";
+    repo = "embuild";
+    rev = "v${version}";
+    sha256 = "sha256-CPMcFzfP/l1g04sBLWj2pY76F94mNsr1RGom1sfY23I=";
```
doronbehar, line 13: Nowadays `hash` is preferred:

```suggestion
    hash = "sha256-CPMcFzfP/l1g04sBLWj2pY76F94mNsr1RGom1sfY23I=";
```

doronbehar (commented):
More small comments I noticed all of a sudden.

doronbehar:
> Now i am wondering why importing the package on my other project and shell worked. Is it due to the hash being the same as the one i used during my previous tests?

Yes exactly. The `src` attributes are called "fixed-output derivations" where internet access is allowed when building them, but the hash you specify must match the contents you get after it finishes building. If you build such a derivation with a certain hash, and then change for instance the `version` attribute of the `src` without changing the `hash`, your `/nix/store` contains a copy of the hash of the original version, and Nix won't by itself be able to guess that perhaps the hash is wrong. Then you experience that everything is fine when in fact someone else trying to build your package will experience a hash mismatch, like you can observe now in CI:

https://github.com/NixOS/nixpkgs/pull/301890/checks?check_run_id=23528147572

There are number of ways to verify your hashes are correct. One of them is:

```
nix-build -A ldproxy.src --check
```

Or, simply set an empty hash, run `nix-build -A ldproxy` and with the editor still open verify you get the same hash.

pkgs/by-name/ld/ldproxy/package.nix
```diff
@@ -0,0 +1,36 @@
+{ lib, rustPlatform, fetchFromGitHub, pkg-config, udev, stdenv }:
```
MinerSebas, line 1: `stdenv` is unused
```suggestion
{ lib, rustPlatform, fetchFromGitHub, pkg-config, udev }:
```

doronbehar:
Before pushing and waiting for CI once more, I'd also verify with a method similar to the above that the `cargoHash` is correct - make it empty and calculate it again.

pkgs/by-name/ld/ldproxy/package.nix
```diff
@@ -0,0 +1,36 @@
...
+  # cf: https://github.com/esp-rs/embuild/tree/v0.31.4/ldproxy
+  buildAndTestSubdir = "ldproxy";
+
+  nativeBuildInputs = [
+    pkg-config
+  ];
+
+  buildInputs = [
+    udev
+  ];
```
MinerSebas, line 26: Why is this needed here? `ldproxy` should not need to depend on udev.

Commenting both out, `ldproxy` successfully build without errors, though I still need to test whether it still runs crrectly.

```suggestion
```

vpochapuis:
> > Now i am wondering why importing the package on my other project and shell worked. Is it due to the hash being the same as the one i used during my previous tests?
> 
> Yes exactly. The `src` attributes are called "fixed-output derivations" where internet access is allowed when building them, but the hash you specify must match the contents you get after it finishes building. If you build such a derivation with a certain hash, and then change for instance the `version` attribute of the `src` without changing the `hash`, your `/nix/store` contains a copy of the hash of the original version, and Nix won't by itself be able to guess that perhaps the hash is wrong. Then you experience that everything is fine when in fact someone else trying to build your package will experience a hash mismatch, like you can observe now in CI:
> 
> https://github.com/NixOS/nixpkgs/pull/301890/checks?check_run_id=23528147572
> 
> There are number of ways to verify your hashes are correct. One of them is:
> 
> ```
> nix-build -A ldproxy.src --check
> ```
> 
> Or, simply set an empty hash, run `nix-build -A ldproxy` and with the editor still open verify you get the same hash.

Thank you very much for this thorough explanation its very clear!
I will try my best to better verify this kind of issues in the future.

pkgs/by-name/ld/ldproxy/package.nix
vpochapuis: Thank you I will try to remove those and test.
The original reason for its presence is that I simply reproduced the original PR, I should have tried more to understand it to avoid this issue.

vpochapuis: Potentially fixed in 0cd4466c2521dbdc3f6d6da63416985877e4c403

pkgs/by-name/ld/ldproxy/package.nix
vpochapuis: Potentially fixed in 0cd4466c2521dbdc3f6d6da63416985877e4c403

pkgs/by-name/ld/ldproxy/package.nix
vpochapuis: Potentially fixed in 0cd4466c2521dbdc3f6d6da63416985877e4c403

pkgs/by-name/ld/ldproxy/package.nix
vpochapuis: Potentially fixed in 0cd4466c2521dbdc3f6d6da63416985877e4c403

pkgs/by-name/ld/ldproxy/package.nix
vpochapuis: Potentially fixed in 0cd4466c2521dbdc3f6d6da63416985877e4c403

doronbehar (approved):
Diff looks good, and CI is green.

doronbehar:
Welcome aboard.

vpochapuis:
> Welcome aboard.

Thank you !!
I have learnt quite a lot with this PR and you guided me well.

Have a great day!

Issue #69414, opened 2019-09-25
Implement NixOS container networking with networkd


arianvp:
We are in the process to moving to `networkd` based networking for `20.03`.

`nixos-container` module neither uses scripted networking nor systemd-based networking,
but its own set of scripts. We should move `nixos-container` to use `networkd` for setting up the network. 

Currently if you enable `networking.useNetworkd = true;` ,  `nixos-container` will stop working because the default rules that `networkd` ships sets up DHCP on the `ve-*` and `vz-*` interfaces as it assumes these are created by `systemd-nspawn` and that's the documented behaviour for the tool   (See https://github.com/NixOS/systemd/tree/nixos-v243/network)

the matches look like this:
```
# 80-container-ve.network
[Match]
Name=ve-*
Driver=veth
[Network]
# Default to using a /24 prefix, giving up to 253 addresses per virtual network.
Address=0.0.0.0/24
LinkLocalAddressing=yes
DHCPServer=yes
IPMasquerade=yes
LLDP=yes
EmitLLDP=customer-bridge

# 80-container-vz.network
[Match]
Name=vz-*
Driver=bridge
[Network]
# Default to using a /28 prefix, giving up to 13 addresses per container.
Address=0.0.0.0/28
LinkLocalAddressing=yes
DHCPServer=yes
IPMasquerade=yes
LLDP=yes
EmitLLDP=customer-bridge

```

We should come up with our own `systemd.network` files that implement all the features that `nixos-container` supports. 

However, I would also like to have vanilla `systemd-nspawn` work as documented in the systemd manpages.  

This means that the `.network` rules that `nixos-container` should generate should be:

1)  more specific:  e.g. `Match=ve-mycontainer`  vs `Match=ve-*`
2) have a lower number than `80-` such that they're executed before the ones shipped with `systemd`

Implementing this is a blocker for switching to `useNetworkd=true` by default.

Maintainer information:
```yaml
# a list of nixpkgs attributes affected by the problem
attribute:
# a list of nixos modules affected by the problem
module:
```

cc @fpletz @flokli @andir @Mic92

arianvp:
I'm not sure why this was labelled  `kind: bug`. could someone remove that for me? :)

ck3d:
If host and container uses networkd the network is configured correctly by networkd.
The the test nixos/tests/systemd-machinectl.nix tests that configuration.
I think there is no need for a special set of configuration files.

arianvp:
[...]

PR #301915, opened 2024-04-05
nixos/fileSystems: add enable option

jmbaur:
## Description of changes

Add the ability to disable filesystems set in a nixos config (enabled by default). This makes the submodules under `fileSystems` behave more like submodules in other nixos modules.

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [x] Tested, as applicable:

---

### DIFF
--- a/nixos/modules/tasks/filesystems.nix
+++ b/nixos/modules/tasks/filesystems.nix
@@ -30,6 +30,10 @@ let
   coreFileSystemOpts = { name, config, ... }: {
 
     options = {
+      enable = mkEnableOption "the filesystem mount" // {
+        default = true;
+      };
+
       mountPoint = mkOption {
         example = "/mnt/usb";
         type = nonEmptyWithoutTrailingSlash;
@@ -222,6 +226,7 @@ in
         }
       '';
       type = types.attrsOf (types.submodule [coreFileSystemOpts fileSystemOpts]);
+      apply = lib.filterAttrs (_: fs: fs.enable);
       description = ''
         The file systems to be mounted.  It must include an entry for
         the root directory (`mountPoint = "/"`).  Each
@@ -268,6 +273,7 @@ in
     boot.specialFileSystems = mkOption {
       default = {};
       type = types.attrsOf (types.submodule coreFileSystemOpts);
+      apply = lib.filterAttrs (_: fs: fs.enable);
       internal = true;
       description = ''
         Special filesystems that are mounted very early during boot.


jmbaur:
@ofborg test nixosTests.bcachefs

jmbaur:
@ofborg test nixosTests.stratis.simple

jmbaur:
@ofborg test nixosTests.overlayfs

ElvishJerricco:
It would probably be better to filter the attrs out in the option's `apply` function, so that this implicitly works on all modules consuming the `fileSystems` config, rather than having to modify each one with `enable` checks.

That said, I'm hesitant about this because I don't particularly see the use case or motivation. Most modules with and `enable` option require the user to explicitly set it to true for the thing to work. There are exceptions to this, but that's not particularly common and it's usually for good reason.

jmbaur:
> It would probably be better to filter the attrs out in the option's `apply` function, so that this implicitly works on all modules consuming the `fileSystems` config, rather than having to modify each one with `enable` checks. (Though, also, all modules should probably be using `config.system.build.fileSystems` rather than `config.fileSystems`?)

Thanks, that definitely makes the changes cleaner

> That said, I'm hesitant about this because I don't particularly see the use case or motivation. Most modules with and `enable` option require the user to explicitly set it to true for the thing to work. There are exceptions to this, but that's not particularly common and it's usually for good reason.

There are a few existing examples in nixpkgs I can think of that use this pattern, the main one I can think of being `systemd.services` (https://search.nixos.org/options?channel=unstable&from=0&size=50&sort=relevance&type=packages&query=systemd.services.%3Cname%3E.enable).

ElvishJerricco:
> There are a few existing examples in nixpkgs I can think of that use this pattern, the main one I can think of being systemd.services

Yes, as I said, there are exceptions. But I also said it's usually for a good reason. `systemd.<type>.enable`, for instance, is primarily there because it can be used to disable individual units from a package added to `systemd.packages`.

PaulGrandperrin:
@ElvishJerricco 

> That said, I'm hesitant about this because I don't particularly see the use case or motivation. Most modules with and enable option require the user to explicitly set it to true for the thing to work. There are exceptions to this, but that's not particularly common and it's usually for good reason.

Here's my use case:

I want to create some nspawn containers with NixOS on the guest (but without sharing /nix so that the guest is fully autonomous).

The issue is that the guest NixOS will try and fail to mount `/dev`, `/dev/pts`, `/dev/shm` and `/run`, which is ok because they are already mounted by the host when preparing the container.

So I want to disable them in `boot.specialFilesystems`, but that's currently impossible without overriding everything, including useful entries like `/run/key` and `/run/wrappers`.

I need something like `environment.etc.<name>.enable` to disable them without removing everything.

PaulGrandperrin:
Without this PR, I think the only solutions I have are:

- override all of `boot.specialFilesystems` and hardcode in my code the definitions of `/run/keys`, `/run/wrappers` etc

- use `disabledModules` on `filesystems.nix` and reimport a patched version using something like `(pkgs.applyPatches {...}).path + "...../filesystems.nix"` with this PR as a patch.

PaulGrandperrin:
Before this PR:
```
[root@test2:/etc/nixos]# nixos-rebuild test
building the system configuration...
activating the configuration...
setting up /etc...
mount: /dev: permission denied.
       dmesg(1) may have more information after failed mount system call.
mount: /dev/pts: permission denied.
       dmesg(1) may have more information after failed mount system call.
mount: /dev/shm: permission denied.
       dmesg(1) may have more information after failed mount system call.
mount: /run: permission denied.
       dmesg(1) may have more information after failed mount system call.
Activation script snippet 'specialfs' failed (32)
reloading user units for root...
restarting sysinit-reactivation.target
reloading the following units: dbus-broker.service
warning: error(s) occurred while switching to the new configuration
```

After this PR and disabling the failing FS:

```
[root@test2:/etc/nixos]# nixos-rebuild test
building the system configuration...
activating the configuration...
setting up /etc...
reloading user units for root...
restarting sysinit-reactivation.target
```

Lyndeno:
Any blockers for this?

PaulGrandperrin:
I don't think there's any technical blockers.

There's also no workarounds that I'm aware of.

Here's how I patch my nixpkgs with this PR to be able to have NixOS in a nspawn container: https://github.com/PaulGrandperrin/nix-systems/blob/main/nixosModules%2Fshared%2Fnspawns.nix#L182

PaulGrandperrin:
I just saw that we are not the only ones trying to get this issue solved!
@Ma27 also made a similar PR: https://github.com/NixOS/nixpkgs/pull/345899
Check out this issue for the context: https://github.com/NixOS/nixpkgs/issues/69414#issuecomment-2564755036

aanderse:
thanks @ElvishJerricco, much appreciated 🙇‍♂️

PR #301921, opened 2024-04-05
vscode-extensions.ms-vsliveshare.vsliveshare: 1.0.5900 -> 1.0.5918

### LINKED ISSUE #163202: ms liveshare macOS
**Support macOS/M1**
Currently live share is linux only. How would I start the process to add liveshare extension to support macOS and M1.

**Metadata**

The below does not work currently:
```
  vscodeExtensions = (with pkgs.vscode-extensions; [

  ]) ++ pkgs.vscode-utils.extensionsFromVscodeMarketplace [
    {
      name = "vsliveshare";
      publisher = "ms-vsliveshare";
      version = "1.0.5418";
      sha256 = "11pnfvssn2vv975liwx5g7qm0l3l6q0lcl91hxgd0bwpkdzmnr1h";
    }
```

* homepage URL:
* source URL: https://github.com/NixOS/nixpkgs/blob/634141959076a8ab69ca2cca0f266852256d79ee/pkgs/applications/editors/vscode/extensions/ms-vsliveshare-vsliveshare/default.nix
* license: mit, bsd, gpl2+ , ...
* platforms: unix, linux, darwin, m1

deviant:
## Description of changes

In addition, this cleans up the package, and enables it on platforms other than x86_64-linux (which had been supported for a while).

Fixes #163202.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [x] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/applications/editors/vscode/extensions/ms-vsliveshare.vsliveshare/default.nix
+++ b/pkgs/applications/editors/vscode/extensions/ms-vsliveshare.vsliveshare/default.nix
@@ -1,54 +1,23 @@
-# Based on previous attempts:
-#  -  <https://github.com/msteen/nixos-vsliveshare/blob/master/pkgs/vsliveshare/default.nix>
-#  -  <https://github.com/NixOS/nixpkgs/issues/41189>
-{ lib, gccStdenv, vscode-utils
-, autoPatchelfHook, bash, makeWrapper
-, curl, gcc, libsecret, libunwind, libX11, lttng-ust, util-linux
-, desktop-file-utils, xsel
-}:
+{ lib, vscode-utils, xsel }:
 
-let
-  # https://docs.microsoft.com/en-us/visualstudio/liveshare/reference/linux#install-prerequisites-manually
-  libs = [
-    # Credential Storage
-    libsecret
-
-    # NodeJS
-    libX11
-
-    # https://github.com/flathub/com.visualstudio.code.oss/issues/11#issuecomment-392709170
-    libunwind
-    lttng-ust
-    curl
-
-    # General
-    gcc.cc.lib
-    util-linux # libuuid
-  ];
-
-in ((vscode-utils.override { stdenv = gccStdenv; }).buildVscodeMarketplaceExtension {
+vscode-utils.buildVscodeMarketplaceExtension {
   mktplcRef = {
     name = "vsliveshare";
     publisher = "ms-vsliveshare";
-    version = "1.0.5900";
-    hash = "sha256-syVW/aS2ppJjg4OZaenzGM3lczt+sLy7prwsYFTDl9s=";
+    version = "1.0.5918";
+    hash = "sha256-Tk0mKydUF8M7l7NC9wEA7t2rzJWy/mq4/HvIHI2/ldQ=";
   };
-}).overrideAttrs({ buildInputs ? [], ... }: {
-  buildInputs = buildInputs ++ libs;
 
-  # Using a patch file won't work, because the file changes too often, causing the patch to fail on most updates.
-  # Rather than patching the calls to functions, we modify the functions to return what we want,
-  # which is less likely to break in the future.
   postPatch = ''
     substituteInPlace extension.js \
-      --replace "'xsel'" "'${xsel}/bin/xsel'"
+      --replace-fail '"xsel"' '"${xsel}/bin/xsel"'
   '';
 
   meta = {
-    description = "Live Share lets you achieve greater confidence at speed by streamlining collaborative editing, debugging, and more in real-time during development";
+    description = "Real-time collaborative development for VS Code";
     homepage = "https://aka.ms/vsls-docs";
+    changelog = "https://marketplace.visualstudio.com/items/MS-vsliveshare.vsliveshare/changelog";
     license = lib.licenses.unfree;
-    maintainers = [ lib.maintainers.jraygauthier lib.maintainers.V ];
-    platforms = [ "x86_64-linux" ];
+    maintainers = builtins.attrValues { inherit (lib.maintainers) jraygauthier V; };
   };
-})
+}


superherointj:
The change to how vscode-extensions works (hash) should be done in a discrete manner. (It's a valid change that I think should be done indeed.)
Still it should be a commit per package. And should not include things that are unrelated to package.

pkgs/applications/editors/vscode/extensions/default.nix
```diff
@@ -24,6 +24,7 @@
 , typst-lsp
 , typst-preview
 , autoPatchelfHook
+, xsel
```
superherointj, line 27: If this extension needs extra parameters it is ideal to keep it in a separate directory. Because it can grow, and in case the extension is removed no one remembers to remove these parameters from the global index.

pkgs/applications/editors/vscode/extensions/default.nix
```diff
@@ -3210,7 +3211,26 @@ let
...
+          hash = "sha256-Tk0mKydUF8M7l7NC9wEA7t2rzJWy/mq4/HvIHI2/ldQ=";
+        };
+
+        postPatch = ''
+          substituteInPlace extension.js \
+            --replace-fail '"xsel"' '"${xsel}/bin/xsel"'
+        '';
+
+        meta = {
+          description = "Real-time collaborative development for VS Code";
```
superherointj, line 3228: `changelog` is missing if you care to add. This is used by the bot when upgrading the package.

deviant: Added that, thanks for the suggestion!

deviant:
> The change to how vscode-extensions works (hash) should be done in a discrete manner. (It's a valid change that I think should be done indeed.) Still it should be a commit per package.

I've split this out into a separate commit now.

pkgs/applications/editors/vscode/extensions/default.nix
```diff
@@ -24,6 +24,7 @@
 , typst-lsp
 , typst-preview
 , autoPatchelfHook
+, xsel  # for ms-vsliveshare.vsliveshare
```
superherointj, line 27:
```suggestion
```

pkgs/applications/editors/vscode/extensions/default.nix
deviant: My understanding was that separate separate files were used for complex additions (lots of additional dependencies, complex patching logic, etc). Which this package used to have (it bundled a native service, which required quite a bit of support from us to make work), but no longer needs. At this point it's on par with several other extensions (`azdavis.millet`, `b4dm4n.vscode-nixpkgs-fmt`, `mgt19937.typst-preview`, etc.).

I've added a note to the import to more easily see why it's being used; is this a reasonable compromise?

superherointj:
> I've split this out into a separate commit now.

It should be a discrete change as in it's own PR because it is a meta change of how *all* vscode extensions work.

Would be good if you kept this PR about the upgrade of package: `vscode-extensions.ms-vsliveshare.vsliveshare`

To avoid conflating issues. (As in a meta change. A package upgrade.)

It's two different discussions. I'm pretty sure that the hash change is welcomed. Just how this is being done that is not ideal.

Then we can even discuss changing all sha256 to hash? Imagine that...

pkgs/applications/editors/vscode/extensions/default.nix
superherointj: This issue is not resolved. No need to mark resolved.

deviant:
> It should be a discrete change as in it's own PR because it is a meta change of how _all_ vscode extensions work.

I'll move this to a fresh PR in a moment. Although I disagree about it *affecting* other VS Code extensions, since it should be back-compatible, and is only widening the set of accepted attributes.

> I'm pretty sure that the hash change is welcomed.

Thanks, and I think this is likely as well. :)

deviant:
The `mktplcRef` change is now in #301936.

pkgs/applications/editors/vscode/extensions/default.nix
deviant: I did so such that the conversation would not get split into two locations. I think you might have missed [this](https://github.com/NixOS/nixpkgs/pull/301921#discussion_r1554294110). ^^

superherointj:
Please rebase.

deviant:
Rebase is done. :)

I think there's just the remaining question of whether this package is large/complex enough to deserve its own file, or if it's fine to leave among the rest of the simple packages. I'm not strongly for either of these, but if it's split out then I think we ought to make an effort to start moving other extensions to their own directories as well.

superherointj:
> I think there's just the remaining question of whether this package is large/complex enough to deserve its own file, or if it's fine to leave among the rest of the simple packages. I'm not strongly for either of these, but if it's split out then I think we ought to make an effort to start moving other extensions to their own directories as well.

I agree there should exist a consistent rule on this.

So far, the reasoning I have been following is:

> Whenever an extension has patches (in files), extra parameters, has components, requires more files they are moved to a discrete directory.

As this extension requires extra parameters, I'm of the opinion this should be moved to a discrete repository.

In case you are interested in moving all extensions that match above rules to a discrete directory. Keep a single commit per extension. But group many extensions in a single PR. And I can review it.

And maybe a comment explaining this rule could be added to the file.

superherointj:
But let's first handle this PR. Before considering moving other extensions.

pkgs/applications/editors/vscode/extensions/default.nix
```diff
@@ -3210,7 +3211,27 @@ let
...
+
+        postPatch = ''
+          substituteInPlace extension.js \
+            --replace-fail '"xsel"' '"${xsel}/bin/xsel"'
+        '';
+
+        meta = {
+          description = "Real-time collaborative development for VS Code";
+          homepage = "https://aka.ms/vsls-docs";
+          changelog = "https://marketplace.visualstudio.com/items/MS-vsliveshare.vsliveshare/changelog";
```
superherointj, line 3230: Usually (or hopefully) alphabetical order is followed in attrs and lists. There are exceptions like: `meta.broken` which is always last to avoid git conflicts.
Considering alphabetical order, `changelog` would be before `description`.

deviant:
> I agree there should exist a consistent rule on this.
> 
> So far, the reasoning I have been following is:
> 
> > Whenever an extension has patches (in files), extra parameters, has components, requires more files they are moved to a discrete directory.
> 
> As this extension requires extra parameters, I'm of the opinion this should be moved to a discrete repository.

Alright, that's fair enough. I've moved it back to its previous position.

> In case you are interested in moving all extensions that match above rules to a discrete directory. Keep a single commit per extension. But group many extensions in a single PR. And I can review it.
> 
> And maybe a comment explaining this rule could be added to the file.

I'd much prefer if this comment existed already, although I'm not going to embark on this refactoring presently.

pkgs/applications/editors/vscode/extensions/default.nix
deviant: This isn't true; attributes appear in a logical order in derivations. For instance, typically `pname` and then `version` can be found, then `src`, patches, and so on; and `meta` can be found at the end. Likewise, the meta has its own ordering: `changelog` appears after `homepage` and before `license`; this can be seen [in the Nixpkgs manual](https://nixos.org/manual/nixpkgs/stable/#var-meta-changelog).

pkgs/applications/editors/vscode/extensions/ms-vsliveshare.vsliveshare/default.nix
```diff
@@ -1,54 +1,23 @@
...
     substituteInPlace extension.js \
-      --replace "'xsel'" "'${xsel}/bin/xsel'"
+      --replace-fail '"xsel"' '"${xsel}/bin/xsel"'
   '';
 
   meta = {
-    description = "Live Share lets you achieve greater confidence at speed by streamlining collaborative editing, debugging, and more in real-time during development";
+    description = "Real-time collaborative development for VS Code";
     homepage = "https://aka.ms/vsls-docs";
+    changelog = "https://marketplace.visualstudio.com/items/MS-vsliveshare.vsliveshare/changelog";
```
superherointj, line 19: Would you mind following alphabetical order for `meta.changelog`?
Move `changelog` to before `description`?

Btw, `meta.broken` is an exception to alphabetical order to avoid git conflicts.

deviant: Duplicating my response to your previous review here:

> This isn't true; attributes appear in a logical order in derivations. For instance, typically `pname` and then `version` can be found, then `src`, patches, and so on; and `meta` can be found at the end. Likewise, the meta has its own ordering: `changelog` appears after `homepage` and before `license`; this can be seen [in the Nixpkgs manual](https://nixos.org/manual/nixpkgs/stable/#var-meta-changelog).

If you check that section of the manual, you can also see that `meta.broken` is listed at the end, as you have mentioned. If you want an example of this, see the [changelog attribute of the Git package](https://github.com/NixOS/nixpkgs/blob/88c619a6a356e641753668987df003f81ab6451a/pkgs/applications/version-management/git/default.nix#L383). I see that other VS Code extensions have the changelog attribute in alphabetical order, but IMO these are the ones getting it wrong, not mine.

superherointj (approved):

pkgs/applications/editors/vscode/extensions/default.nix
superherointj: I did not see this message when I wrote that message. Would be good if you did not (prematurely) mark conversation as resolved and let the reviewer do that. Because it hides the message unless the reviewer clicks on it. In your mind, I understand your argument is reasonable to you, but the other person has to catch up. Consensus is reached not when you provide a sound argument, but when the other part has reached the same understanding. And then, marking resolved makes the most sense for the reviewer.

---

Yes, attributes are in a logical order in derivations following phases order. I did not refer to that. And I did say there are exceptions. Which was not the case. And you did correct me right. VSCode just happens to not be following that. As you noticed. What I was trying to say is that it is good to use alphabetical order whenever possible in attrs/lists.

pkgs/applications/editors/vscode/extensions/ms-vsliveshare.vsliveshare/default.nix
superherointj: > Duplicating my response to your previous review here:

Thanks for pointing out that.

> I see that other VS Code extensions have the changelog attribute in alphabetical order, but IMO these are the ones getting it wrong, not mine.

Yes, you may be right.

pkgs/applications/editors/vscode/extensions/default.nix
deviant: Hi, sorry for unintentionally closing these without you getting to read them. I was incorrectly assuming that you were getting notifications for them and had seen them. For instance, I get an e-mail for each individual comment you make, which immediately shows up on my phone. I was also under this impression because there was a review I had responded to (which had not yet been marked as resolved) which you just didn't react to at all. So it was not clear that this was important.

I like to mark things as resolved once I've handled them, since not doing so has the possibility of producing situations where a reviewer will come in and comment on something I've done, I'll handle it, but they'll disappear and not respond further, making other reviewers less likely to touch the PR or merge it. This is quite annoying, I'm sure you'll appreciate.

Issue #301902, opened 2024-04-05
Warp terminal wayland 


xiro-codes:
### Describe the bug
Warp Terminal requires LD_LIBRARY_PATH to contain wayland libs to allow wayland support

### Steps To Reproduce
1. set env variable WARP_ENABLE_WAYLAND 1 
2. launch

how I got it to work
https://app.warp.dev/block/piOjTikLERuFxOZjQ16DMW

### Notify maintainers
@emilytrau  @Enzime @imadnyc 

### Metadata
Please run `nix-shell -p nix-info --run "nix-info -m"` and paste the result.

```console
[user@system:~]$ nix-shell -p nix-info --run "nix-info -m"
output here
```

---

Add a :+1: [reaction] to [issues you find important].

[reaction]: https://github.blog/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/
[issues you find important]: https://github.com/NixOS/nixpkgs/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions-%2B1-desc

imadnyc:
It does seem like it works, but I'd like to know more on how you found this. Where did you find out that that flag had to be enabled?

xiro-codes:
https://docs.warp.dev/help/known-issues 
Environmental Variables

Prefix warp-terminal with the variables (multiple can be used), and once you confirm they work, export them in your .profile/.zprofile to load on startup:

Default to Wayland: WARP_ENABLE_WAYLAND=1

Set Default GPU: e.g. MESA_D3D12_DEFAULT_ADAPTER_NAME=NVIDIA

Set Graphics APIs: e.g. WGPU_BACKEND=gl

imadnyc:
Sick! I'll make the PR now.

PR #301944, opened 2024-04-05
warp-terminal: enabled wayland support

imadnyc:
## Description of changes
Enabled wayland support thanks to some help from @xiro-codes at https://github.com/NixOS/nixpkgs/issues/301902#issuecomment-2040742810. 

Tested using xeyes, and it is running in wayland mode in wayland and in x11 mode in xorg. 

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/by-name/wa/warp-terminal/package.nix
+++ b/pkgs/by-name/wa/warp-terminal/package.nix
@@ -10,9 +10,12 @@
 , libglvnd
 , libxkbcommon
 , vulkan-loader
+, wayland
 , xdg-utils
 , xorg
 , zlib
+, makeWrapper
+, waylandSupport ? false
 }:
 
 let
@@ -35,7 +38,7 @@ linux = stdenv.mkDerivation (finalAttrs:  {
       --replace-fail /opt/ $out/opt/
   '';
 
-  nativeBuildInputs = [ autoPatchelfHook zstd ];
+  nativeBuildInputs = [ autoPatchelfHook zstd makeWrapper ];
 
   buildInputs = [
     curl
@@ -54,7 +57,7 @@ linux = stdenv.mkDerivation (finalAttrs:  {
     xorg.libxcb
     xorg.libXcursor
     xorg.libXi
-  ];
+  ] ++ lib.optionals waylandSupport [wayland];
 
   installPhase = ''
     runHook preInstall
@@ -62,6 +65,9 @@ linux = stdenv.mkDerivation (finalAttrs:  {
     mkdir $out
     cp -r opt usr/* $out
 
+  '' ++ lib.optionalString waylandSupport ''
+    wrapProgram $out/bin/warp-terminal --set WARP_ENABLE_WAYLAND 1
+  '' ++ ''
     runHook postInstall
   '';
 });


pkgs/by-name/wa/warp-terminal/package.nix
```diff
@@ -35,12 +37,13 @@ linux = stdenv.mkDerivation (finalAttrs:  {
...
   '';
 
-  nativeBuildInputs = [ autoPatchelfHook zstd ];
+  nativeBuildInputs = [ autoPatchelfHook zstd makeWrapper];
 
   buildInputs = [
     curl
     fontconfig
     stdenv.cc.cc.lib # libstdc++.so libgcc_s.so
+    wayland
```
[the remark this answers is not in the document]
imadnyc: I think that the `makeWrapper` is needed for the wayland flag, but yeah, I should move that to runtime

pkgs/by-name/wa/warp-terminal/package.nix
```diff
@@ -64,6 +67,10 @@ linux = stdenv.mkDerivation (finalAttrs:  {
 
     runHook postInstall
   '';
+
+  postInstall = ''
+    wrapProgram $out/bin/warp-terminal --set WARP_ENABLE_WAYLAND 1
```
a-n-n-a-l-e-e, line 72: what happens if this flag is set and wayland is not available? i'd suggest not wrapping so the behavior of the program is identical to what is in the docs: https://docs.warp.dev/help/known-issues#linux-1 or adding some booleans or something. idk.

imadnyc: From what I can tell, it fallsback to x11, this just sets wayland priority higher. If I run it on Xorg it's being picked up by xeyes as a proper x11 app. The thing that's linked to the known-issues you sent seems to be for WSL on windows, but I also saw this? https://github.com/warpdotdev/Warp/issues/4628. I'm confused on whether or not wayland is stable. If it isn't then I agree that it should just be a flag and then we can set it to default later.

xiro-codes: https://github.com/warpdotdev/Warp/issues/4240#issuecomment-1961993281 although i have no issue warp considers it unstable so its  not a bad idea to make it opt in rather than the default. It wayland support is very usable as of the last few updates

emilytrau, line 72: This can be part of `installPhase` instead of a seperate phase.

pkgs/by-name/wa/warp-terminal/package.nix
```diff
@@ -35,7 +37,7 @@ linux = stdenv.mkDerivation (finalAttrs:  {
       --replace-fail /opt/ $out/opt/
   '';
 
-  nativeBuildInputs = [ autoPatchelfHook zstd ];
+  nativeBuildInputs = [ autoPatchelfHook zstd makeWrapper];
```
emilytrau, line 40: Nit
```suggestion
  nativeBuildInputs = [ autoPatchelfHook zstd makeWrapper ];
```

pkgs/by-name/wa/warp-terminal/package.nix
imadnyc: Yeah, I saw that, but I was unsure about the wording. Is it unstable in WSL or just wayland in general?

imadnyc:
I made it optional, moved it to `installPhase`, and fixed up some formatting

HirschBerge:
Is there any reason this is still not accepted? I would like to use warp-terminal again...

wegank:
Looks like this breaks eval on master?

Atemu:
Please remember to check CI before merging!

imadnyc:
Sorry for the mistake, I used `++` instead of `+`. I fixed it now.

vorporeal:
Warp developer here - we don't recommend using `WARP_ENABLE_WAYLAND`, as we're still working on addressing stability/correctness issues (e.g.: windows not rendering) when using Wayland (vs. XWayland).

A bug caused in the most recent Warp release caused native Wayland support to be enabled for all users - https://github.com/warpdotdev/Warp/issues/4810 makes it clear that there are still issues for us to address.

emilytrau:
My apologies everyone. I'll strive to be more careful. 😓

Atemu:
@vorporeal thanks for chiming in! Note that the description of the PR is misleading as what was added is the *option* to enable wayland suport but it is disabled by default. Users of the Nix package will not have wayland support enabled unless they explicitly opt to do that via an override.

imadnyc:
Hey, can someone test the latest master branch and launch it on wayland? Am I the only one that can't launch it in wayland mode, even when the flag is off? Should we revert this change?

HirschBerge:
I see it hit unstable. I'm happy to test. How do i set the flag, I'm not super familiar?
Would the following, that I'm editing from something I had previously finagled to be working, be correct?
```
  nixpkgs.overlays = [
    (self: super: {
     warp-terminal = super.warp-terminal.overrideAttrs (oldAttrs: {
         waylandSupport = true;
         });
     })
  ];
```
I appreciate any corrections in case this is wrong.

annaleeleaves:
@HirschBerge you want to use `override` not `overrideAttrs` to override the function arguments, as `waylandSupport` is an argument and not an attribute.
```nix
warp-terminal = super.warp-terminal.override {  waylandSupport = true; };
```
and then you should see that `wayland` is in the `rpath` thus making `LD_LIBRARY_PATH` unnecessary.
```
$ patchelf --print-rpath /nix/store/xxxxxx-warp-terminal-xxx/opt/warpdotdev/warp-terminal/warp|xargs -d : -n 1|grep wayland
/nix/store/vsndcp99h2bzbckf75d7f0fgzwvg4ab8-wayland-1.22.0/lib
```

HirschBerge:
> @HirschBerge you want to use `override` not `overrideAttrs` to override the function arguments, as `waylandSupport` is an argument and not an attribute.
> 
> ```nix
> warp-terminal = super.warp-terminal.override {  waylandSupport = true; };
> ```
> 
> and then you should see that `wayland` is in the `rpath` thus making `LD_LIBRARY_PATH` unnecessary.
> 
> ```
> $ patchelf --print-rpath /nix/store/xxxxxx-warp-terminal-xxx/opt/warpdotdev/warp-terminal/warp|xargs -d : -n 1|grep wayland
> /nix/store/vsndcp99h2bzbckf75d7f0fgzwvg4ab8-wayland-1.22.0/lib
> ```

Unfortunately, that does not work. running the patchelf command also results in no wayland on path.

annaleeleaves:
> Unfortunately, that does not work. running the patchelf command also results in no wayland on path.

seems like you have something misconfigured. verified

```
$ nix-build --expr 'with import ./.{}; warp-terminal.override { waylandSupport = true; }' && patchelf --print-rpath result/opt/warpdotdev/warp-terminal/warp|xargs -d : -n 1|grep wayland
/nix/store/n7rclb2l3a1g3jqna0kzs70l8w3hi0bz-warp-terminal-0.2024.04.23.08.01.stable_03
/nix/store/19idnjxghxxzwmh2hrzs5j3vxw7czyza-wayland-1.22.0/lib
```

as well as:
```
$ echo 'self: super: { warp-terminal = super.warp-terminal.override {  waylandSupport = true; }; }' > ~/.config/nixpkgs/overlays/warp-terminal.nix
$ nix-build -A warp-terminal && patchelf --print-rpath result/opt/warpdotdev/warp-terminal/warp|xargs -d : -n 1|grep wayland
/nix/store/n7rclb2l3a1g3jqna0kzs70l8w3hi0bz-warp-terminal-0.2024.04.23.08.01.stable_03
/nix/store/19idnjxghxxzwmh2hrzs5j3vxw7czyza-wayland-1.22.0/lib
```

on master

MZauchner:
I am still getting the following issue when trying to sign in for the first time:
```[134483:134484:20240504,115519.292411:ERROR http_transport_libcurl.cc:118] dlopen:libcurl.so: cannot open shared object file: No such file or directory
[134483:134484:20240504,115519.292419:ERROR http_transport_libcurl.cc:118] dlopen:libcurl-gnutls.so.4: cannot open shared object file: No such file or directory
[134483:134484:20240504,115519.292420:ERROR http_transport_libcurl.cc:118] dlopen:libcurl-nss.so.4: cannot open shared object file: No such file or directory
[134483:134484:20240504,115519.292421:ERROR http_transport_libcurl.cc:118] dlopen:libcurl.so.4: cannot open shared object file: No such file or directory```

PR #184362, opened 2022-07-31
firefly-iii: init at 5.7.18, nixos/firefly-iii: init

eliandoran:
###### Description of changes

Official website: https://www.firefly-iii.org/

According to the official introduction:

> "Firefly III" is a (self-hosted) manager for your personal finances. It can help you keep track of your expenses and income, so you can spend less and save more.
Personal financial management is pretty difficult, and everybody has their own approach to it. Some people make budgets, other people limit their cashflow by throwing away their credit cards, others try to increase their current cashflow. There are tons of ways to save and earn money. Firefly III works on the principle that if you know where your money is going, you can stop it from going there.
By keeping track of your expenses and your income you can budget accordingly and save money. Stop living from paycheck to paycheck but give yourself the financial wiggle room you need.

This PR introduces both the PHP package for `firefly-iii` (via `composer2nix`), as well as the accompanying module. The module has configuration support for database, mail, nginx reverse proxy. There's also a very simple sanity check.

The work done here was heavily inspired by the work of @ymarkus and @talyz , from the `bookstack` package, which uses the same framework (Laravel). Thanks!

###### Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
- [ ] For non-Linux: Is `sandbox = true` set in `nix.conf`? (See [Nix manual](https://nixos.org/manual/nix/stable/command-ref/conf-file.html))
- [x] Tested, as applicable:
  - [NixOS test(s)](https://nixos.org/manual/nixos/unstable/index.html#sec-nixos-tests) (look inside [nixos/tests](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests))
  - and/or [package tests](https://nixos.org/manual/nixpkgs/unstable/#sec-package-tests)
  - or, for functions and "core" functionality, tests in [lib/tests](https://github.com/NixOS/nixpkgs/blob/master/lib/tests) or [pkgs/test](https://github.com/NixOS/nixpkgs/blob/master/pkgs/test)
  - made sure NixOS tests are [linked](https://nixos.org/manual/nixpkgs/unstable/#ssec-nixos-tests-linking) to the relevant packages
- [ ] Tested compilation of all packages that depend on this change using `nix-shell -p nixpkgs-review --run "nixpkgs-review rev HEAD"`. Note: all changes have to be committed, also see [nixpkgs-review usage](https://github.com/Mic92/nixpkgs-review#usage)
[...]

PR #302136, opened 2024-04-06
firefly-iii: init at 6.1.13, nixos/firefly-iii: init

savyajha:
## Description of changes

We already have a PR for firefly-iii which has not been merged in #184362. This is an attempt to push in the latest version of a package without using `composer2nix` and using the latest instructions from the [docs](https://docs.firefly-iii.org/how-to/firefly-iii/installation/self-managed/).

From the [github repo](https://github.com/firefly-iii/firefly-iii):

> "Firefly III" is a (self-hosted) manager for your personal finances. It can help you keep track of your expenses and income, so you can spend less and save more. Firefly III supports the use of budgets, categories and tags. Using a bunch of external tools, you can import data. It also has many neat financial reports available.
> 
> Firefly III should give you insight into and control over your finances. Money should be useful, not scary. You should be able to see where it is going, to feel your expenses and to... wow, I'm going overboard with this aren't I?
> 
> But you get the idea: this is your money. These are your expenses. Stop them from controlling you. I built this tool because I started to dislike money. Having money, not having money, paying bills with money, you get the idea. But no more. I want to feel "safe", whatever my balance is. And I hope this tool can help you. I know it helps me.

I'd love to get some feedback for this because I've only tested this on my machines with sqlite and postgres. While the module should setup sqlite on its own, the postgres and mysql databases have to be created by the user. There is no web-server setup included in this module either.

## Things done

- Built on platform(s)
  - [X] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [x] Tested, as applicable:

---

### DIFF
--- a/nixos/modules/module-list.nix
+++ b/nixos/modules/module-list.nix
@@ -1320,6 +1320,7 @@
   ./services/web-apps/dolibarr.nix
   ./services/web-apps/engelsystem.nix
   ./services/web-apps/ethercalc.nix
+  ./services/web-apps/firefly-iii.nix
   ./services/web-apps/fluidd.nix
   ./services/web-apps/freshrss.nix
   ./services/web-apps/galene.nix
--- a/nixos/modules/services/web-apps/firefly-iii.nix
+++ b/nixos/modules/services/web-apps/firefly-iii.nix
@@ -1,0 +1,367 @@
+{ pkgs, config, lib, ... }:
+
+let
+  inherit (lib) optionalString mkDefault mkIf mkOption mkEnableOption literalExpression;
+  inherit (lib.types) nullOr attrsOf oneOf str int bool path package enum submodule;
+  inherit (lib.strings) concatMapStringsSep removePrefix toShellVars removeSuffix hasSuffix;
+  inherit (lib.attrsets) attrValues genAttrs filterAttrs mapAttrs' nameValuePair;
+  inherit (builtins) isInt isString toString typeOf;
+
+  cfg = config.services.firefly-iii;
+
+  user = cfg.user;
+  group = cfg.group;
+
+  defaultUser = "firefly-iii";
+  defaultGroup = "firefly-iii";
+
+  artisan = "${cfg.package}/artisan";
+
+  env-file-values = mapAttrs' (n: v: nameValuePair (removeSuffix "_FILE" n) v)
+    (filterAttrs (n: v: hasSuffix "_FILE" n) cfg.settings);
+  env-nonfile-values = filterAttrs (n: v: ! hasSuffix "_FILE" n) cfg.settings;
+
+  envfile = pkgs.writeText "firefly-iii-env" ''
+    ${toShellVars env-file-values}
+    ${toShellVars env-nonfile-values}
+  '';
+
+  fileenv-func = ''
+    cp --no-preserve=mode ${envfile} /tmp/firefly-iii-env
+    ${concatMapStringsSep "\n"
+      (n: "${pkgs.replace-secret}/bin/replace-secret ${n} ${n} /tmp/firefly-iii-env")
+      (attrValues env-file-values)}
+    set -a
+    . /tmp/firefly-iii-env
+    set +a
+  '';
+
+  firefly-iii-maintenance = pkgs.writeShellScript "firefly-iii-maintenance.sh" ''
+    ${fileenv-func}
+
+    ${optionalString (cfg.settings.DB_CONNECTION == "sqlite")
+      "touch ${cfg.dataDir}/storage/database/database.sqlite"}
+    ${artisan} migrate --seed --no-interaction --force
+    ${artisan} firefly-iii:decrypt-all
+    ${artisan} firefly-iii:upgrade-database
+    ${artisan} firefly-iii:correct-database
+    ${artisan} firefly-iii:report-integrity
+    ${artisan} firefly-iii:laravel-passport-keys
+    ${artisan} cache:clear
+
+    mv /tmp/firefly-iii-env /run/phpfpm/firefly-iii-env
+  '';
+
+  commonServiceConfig = {
+    Type = "oneshot";
+    User = user;
+    Group = group;
+    StateDirectory = "${removePrefix "/var/lib/" cfg.dataDir}";
+    WorkingDirectory = cfg.package;
+    PrivateTmp = true;
+    PrivateDevices = true;
+    CapabilityBoundingSet = "";
+    AmbientCapabilities = "";
+    ProtectSystem = "strict";
+    ProtectKernelTunables = true;
+    ProtectKernelModules = true;
+    ProtectControlGroups = true;
+    ProtectClock = true;
+    ProtectHostname = true;
+    ProtectHome = "tmpfs";
+    ProtectKernelLogs = true;
+    ProtectProc = "invisible";
+    ProcSubset = "pid";
+    PrivateNetwork = false;
+    RestrictAddressFamilies = "AF_INET AF_INET6 AF_UNIX";
+    SystemCallArchitectures = "native";
+    SystemCallFilter = [
+      "@system-service @resources"
+      "~@obsolete @privileged"
+    ];
+    RestrictSUIDSGID = true;
+    RemoveIPC = true;
+    NoNewPrivileges = true;
+    RestrictRealtime = true;
+    RestrictNamespaces = true;
+    LockPersonality = true;
+    PrivateUsers = true;
+  };
+
+in {
+
+  options.services.firefly-iii = {
+
+    enable = mkEnableOption "Firefly III: A free and open source personal finance manager";
+
+    user = mkOption {
+      type = str;
+      default = defaultUser;
+      description = "User account under which firefly-iii runs.";
+    };
+
+    group = mkOption {
+      type = str;
+      default = if cfg.enableNginx then "nginx" else defaultGroup;
+      defaultText = "If `services.firefly-iii.enableNginx` is true then `nginx` else ${defaultGroup}";
+      description = ''
+        Group under which firefly-iii runs. It is best to set this to the group
+        of whatever webserver is being used as the frontend.
+      '';
+    };
+
+    dataDir = mkOption {
+      type = path;
+      default = "/var/lib/firefly-iii";
+      description = ''
+        The place where firefly-iii stores its state.
+      '';
+    };
+
+    package = mkOption {
+      type = package;
+      default = pkgs.firefly-iii;
+      defaultText = literalExpression "pkgs.firefly-iii";
+      description = ''
+        The firefly-iii package served by php-fpm and the webserver of choice.
+        This option can be used to point the webserver to the correct root. It
+        may also be used to set the package to a different version, say a
+        development version.
+      '';
+      apply = firefly-iii : firefly-iii.override (prev: {
+        dataDir = cfg.dataDir;
+      });
+    };
+
+    enableNginx = mkOption {
+      type = bool;
+      default = false;
+      description = ''
+        Whether to enable nginx or not. If enabled, an nginx virtual host will
+        be created for access to firefly-iii. If not enabled, then you may use
+        `''${config.services.firefly-iii.package}` as your document root in
+        whichever webserver you wish to setup.
+      '';
+    };
+
+    virtualHost = mkOption {
+      type = str;
+      description = ''
+        The hostname at which you wish firefly-iii to be served. If you have
+        enabled nginx using `services.firefly-iii.enableNginx` then this will
+        be used.
+      '';
+    };
+
+    poolConfig = mkOption {
+      type = attrsOf (oneOf [ str int bool ]);
+      default = {
+        "pm" = "dynamic";
+        "pm.max_children" = 32;
+        "pm.start_servers" = 2;
+        "pm.min_spare_servers" = 2;
+        "pm.max_spare_servers" = 4;
+        "pm.max_requests" = 500;
+      };
+      description = ''
+        Options for the Firefly III PHP pool. See the documentation on <literal>php-fpm.conf</literal>
+        for details on configuration directives.
+      '';
+    };
+
+    settings = mkOption {
+      description = ''
+        Options for firefly-iii configuration. Refer to
+        <https://github.com/firefly-iii/firefly-iii/blob/main/.env.example> for
+        details on supported values. All <option>_FILE values supported by
+        upstream are supported here.
+
+        APP_URL will be set by `services.firefly-iii.virtualHost`, do not
+        redefine it here.
+      '';
+      example = literalExpression ''
+        {
+          APP_ENV = "production";
+          APP_KEY_FILE = "/var/secrets/firefly-iii-app-key.txt";
+          SITE_OWNER = "mail@example.com";
+          DB_CONNECTION = "mysql";
+          DB_HOST = "db";
+          DB_PORT = 3306;
+          DB_DATABASE = "firefly";
+          DB_USERNAME = "firefly";
+          DB_PASSWORD_FILE = "/var/secrets/firefly-iii-mysql-password.txt;
+        }
+      '';
+      default = {};
+      type = submodule {
+        freeformType = attrsOf (oneOf [str int bool]);
+        options = {
+          DB_CONNECTION = mkOption {
+            type = enum [ "sqlite" "pgsql" "mysql" ];
+            default = "sqlite";
+            example = "pgsql";
+            description = ''
+              The type of database you wish to use. Can be one of "sqlite",
+              "mysql" or "pgsql".
+            '';
+          };
+          APP_ENV = mkOption {
+            type = enum [ "local" "production" "testing" ];
+            default = "local";
+            example = "production";
+            description = ''
+              The app environment. It is recommended to keep this at "local".
+              Possible values are "local", "production" and "testing"
+            '';
+          };
+          DB_PORT = mkOption {
+            type = nullOr int;
+            default = if cfg.settings.DB_CONNECTION == "sqlite" then null
+                      else if cfg.settings.DB_CONNECTION == "mysql" then 3306
+                      else 5432;
+            defaultText = ''
+              `null` if DB_CONNECTION is "sqlite", `3306` if "mysql", `5432` if "pgsql"
+            '';
+            description = ''
+              The port your database is listening at. sqlite does not require
+              this value to be filled.
+            '';
+          };
+          APP_KEY_FILE = mkOption {
+            type = path;
+            description = ''
+              The path to your appkey. The file should contain a 32 character
+              random app key. This may be set using `echo "base64:$(head -c 32
+              /dev/urandom | base64)" > /path/to/key-file`.
+            '';
+          };
+        };
+      };
+    };
+  };
+
+  config = mkIf cfg.enable {
+
+    services.firefly-iii = {
+      settings = {
+        APP_URL = cfg.virtualHost;
+      };
+    };
+
+    services.phpfpm.pools.firefly-iii = {
+      inherit user group;
+      phpPackage = cfg.package.phpPackage;
+      phpOptions = ''
+        log_errors = on
+      '';
+      settings = {
+        "listen.mode" = "0660";
+        "listen.owner" = user;
+        "listen.group" = group;
+        "clear_env" = "no";
+      } // cfg.poolConfig;
+    };
+
+    systemd.services.phpfpm-firefly-iii.serviceConfig = {
+      EnvironmentFile = "/run/phpfpm/firefly-iii-env";
+      ExecStartPost = "${pkgs.coreutils}/bin/rm /run/phpfpm/firefly-iii-env";
+    };
+
+    systemd.services.firefly-iii-setup = {
+      requiredBy = [ "phpfpm-firefly-iii.service" ];
+      before = [ "phpfpm-firefly-iii.service" ];
+      serviceConfig = {
+        ExecStart = firefly-iii-maintenance;
+        RuntimeDirectory = "phpfpm";
+        RuntimeDirectoryPreserve = true;
+      } // commonServiceConfig;
+      unitConfig.JoinsNamespaceOf = "phpfpm-firefly-iii.service";
+    };
+
+    systemd.services.firefly-iii-cron = {
+      description = "Daily Firefly III cron job";
+      script = ''
+        ${fileenv-func}
+        ${artisan} firefly-iii:cron
+      '';
+      serviceConfig = commonServiceConfig;
+    };
+
+    systemd.timers.firefly-iii-cron = {
+      description = "Trigger Firefly Cron";
+      timerConfig = {
+        OnCalendar = "Daily";
+        RandomizedDelaySec = "1800s";
+        Persistent = true;
+      };
+      wantedBy = [ "timers.target" ];
+    };
+
+    services.nginx = mkIf cfg.enableNginx {
+      enable = true;
+      recommendedTlsSettings = mkDefault true;
+      recommendedOptimisation = mkDefault true;
+      recommendedGzipSettings = mkDefault true;
+      virtualHosts.${cfg.virtualHost} = {
+        root = "${cfg.package}/public";
+        locations = {
+          "/" = {
+            tryFiles = "$uri $uri/ /index.php?$query_string";
+            index = "index.php";
+            extraConfig = ''
+              sendfile off;
+            '';
+          };
+          "~ \.php$" = {
+            extraConfig = ''
+              include ${config.services.nginx.package}/conf/fastcgi_params ;
+              fastcgi_param SCRIPT_FILENAME $request_filename;
+              fastcgi_param modHeadersAvailable true; #Avoid sending the security headers twice
+              fastcgi_pass unix:${config.services.phpfpm.pools.firefly-iii.socket};
+            '';
+          };
+        };
+      };
+    };
+
+    systemd.tmpfiles.settings."10-firefly-iii" = genAttrs [
+      "${cfg.dataDir}/storage"
+      "${cfg.dataDir}/storage/app"
+      "${cfg.dataDir}/storage/database"
+      "${cfg.dataDir}/storage/export"
+      "${cfg.dataDir}/storage/framework"
+      "${cfg.dataDir}/storage/framework/cache"
+      "${cfg.dataDir}/storage/framework/sessions"
+      "${cfg.dataDir}/storage/framework/views"
+      "${cfg.dataDir}/storage/logs"
+      "${cfg.dataDir}/storage/upload"
+      "${cfg.dataDir}/cache"
+    ] (n: {
+      d = {
+        group = group;
+        mode = "0700";
+        user = user;
+      };
+    }) // {
+      "${cfg.dataDir}".d = {
+        group = group;
+        mode = "0710";
+        user = user;
+      };
+    };
+
+    users = {
+      users = mkIf (user == defaultUser) {
+        ${defaultUser} = {
+          description = "Firefly-iii service user";
+          inherit group;
+          isSystemUser = true;
+          home = cfg.dataDir;
+        };
+      };
+      groups = mkIf (group == defaultGroup) {
+        ${defaultGroup} = {};
+      };
+    };
+  };
+}
--- a/nixos/tests/firefly-iii.nix
+++ b/nixos/tests/firefly-iii.nix
@@ -1,0 +1,26 @@
+import ./make-test-python.nix ({ lib, pkgs, ... }: {
+  name = "firefly-iii";
+  meta.maintainers = [ lib.maintainers.savyajha ];
+
+  nodes.machine = { config, ... }: {
+    environment.etc = {
+      "firefly-iii-appkey".text = "TestTestTestTestTestTestTestTest";
+    };
+    services.firefly-iii = {
+      enable = true;
+      virtualHost = "http://localhost";
+      enableNginx = true;
+      settings = {
+        APP_KEY_FILE = "/etc/firefly-iii-appkey";
+        LOG_CHANNEL = "stdout";
+        SITE_OWNER = "mail@example.com";
+      };
+    };
+  };
+
+  testScript = ''
+    machine.wait_for_unit("phpfpm-firefly-iii.service")
+    machine.wait_for_unit("nginx.service")
+    machine.succeed("curl -fvvv -Ls http://localhost/ | grep 'Firefly III'")
+  '';
+})
--- a/pkgs/by-name/fi/firefly-iii/package.nix
+++ b/pkgs/by-name/fi/firefly-iii/package.nix
@@ -1,0 +1,60 @@
+{ lib
+, fetchFromGitHub
+, buildNpmPackage
+, php83
+, nixosTests
+, dataDir ? "/var/lib/firefly-iii"
+}:
+
+let
+  pname = "firefly-iii";
+  version = "6.1.13";
+  phpPackage = php83;
+
+  src = fetchFromGitHub {
+    owner = "firefly-iii";
+    repo = "firefly-iii";
+    rev = "v${version}";
+    hash = "sha256-85zI8uCyyoCflzxDkvba6FWa9B3kh179DJfQ2Um6MGM=";
+  };
+
+  assets = buildNpmPackage {
+    pname = "${pname}-assets";
+    inherit version src;
+    npmDepsHash = "sha256-wuPUE6XuzzgKjpxZVgwh2wGut15M61WSBFG+YIZwOFM=";
+    dontNpmBuild = true;
+    installPhase = ''
+      runHook preInstall
+      npm run build
+      cp -r ./public $out/
+      runHook postInstall
+    '';
+  };
+in
+
+phpPackage.buildComposerProject (finalAttrs: {
+  inherit pname src version;
+
+  vendorHash = "sha256-CVGKyyLp5hjjpEulDNEYfljU4OgPBaFcYQQAUf6GeGs=";
+
+  passthru = {
+    inherit phpPackage;
+    tests = nixosTests.firefly-iii;
+  };
+
+  postInstall = ''
+    mv $out/share/php/${pname}/* $out/
+    rm -R $out/share $out/storage $out/bootstrap/cache $out/public
+    cp -a ${assets} $out/public
+    ln -s ${dataDir}/storage $out/storage
+    ln -s ${dataDir}/cache $out/bootstrap/cache
+  '';
+
+  meta = {
+    changelog = "https://github.com/firefly-iii/firefly-iii/releases/tag/v${version}";
+    description = "Firefly III: a personal finances manager";
+    homepage = "https://github.com/firefly-iii/firefly-iii";
+    license = lib.licenses.agpl3Only;
+    maintainers = [ lib.maintainers.savyajha ];
+  };
+})
--- a/nixos/doc/manual/release-notes/rl-2405.section.md
+++ b/nixos/doc/manual/release-notes/rl-2405.section.md
@@ -151,6 +151,8 @@ The pre-existing [services.ankisyncd](#opt-services.ankisyncd.enable) has been m
 
 - [davis](https://github.com/tchapi/davis), a simple CardDav and CalDav server inspired by Baïkal. Available as [services.davis]($opt-services-davis.enable).
 
+- [Firefly-iii](https://www.firefly-iii.org), a free and open source personal finance manager. Available as [services.firefly-iii](#opt-services.firefly-iii.enable)
+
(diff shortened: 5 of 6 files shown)


nixos/modules/module-list.nix
```diff
@@ -1320,6 +1320,7 @@
   ./services/web-apps/dolibarr.nix
   ./services/web-apps/engelsystem.nix
   ./services/web-apps/ethercalc.nix
+  ./services/web-apps/firefly-iii.nix
```
Lurkki14, line 1323: Module addition needs a release entry, as per the PR checklist

nixos/modules/services/web-apps/firefly-iii.nix
```diff
@@ -0,0 +1,355 @@
...
+    RestrictNamespaces = true;
+    LockPersonality = true;
+    PrivateUsers = true;
+  };
+
+in {
+
+  options.services.firefly-iii = {
+
+    enable = mkEnableOption "Firefly III";
```
Lurkki14, line 159: This should mention what Firefly III is

Lurkki14 (changes requested):

nixos/modules/module-list.nix
savyajha: Added

nixos/modules/services/web-apps/firefly-iii.nix
savyajha: Done!

pkgs/by-name/fi/firefly-iii/package.nix
```diff
@@ -0,0 +1,69 @@
...
+
+  installPhase = ''
+    runHook preInstall
+    mkdir -p $out
+    cp -r * $out/
+    runHook postInstall
+  '';
+
+  passthru = {
+    phpPackage = phpPackage;
```
Lurkki14, line 50:
```suggestion
    inherit phpPackage;
```

nixos/modules/services/web-apps/firefly-iii.nix
```diff
@@ -0,0 +1,355 @@
...
+  commonServiceConfig = {
+    Type = "oneshot";
+    User = user;
+    Group = group;
+    StateDirectory = "firefly-iii";
+    WorkingDirectory = "${cfg.package}";
+    PrivateTmp = true;
+    PrivateDevices = true;
+    CapabilityBoundingSet = "";
+    AmbientCapabilities = "";
```
Lurkki14, line 128: Usually setting something to an empty string isn't needed; what's the use here?

Lurkki14 (changes requested):

nixos/modules/services/web-apps/firefly-iii.nix
savyajha: It explicitly disallows any capcbilities. If these aren't specified then no capabilities are restricted.

pkgs/by-name/fi/firefly-iii/package.nix
savyajha: Made this change.

pkgs/by-name/fi/firefly-iii/package.nix
```diff
@@ -0,0 +1,67 @@
+{ stdenvNoCC
+, lib
+, fetchzip
+, php83
```
savyajha, line 4: I'm wondering whether to change this to `phpPackage ? php83` and make this overridable. Thoughts?

nixos/modules/services/web-apps/firefly-iii.nix
```diff
@@ -0,0 +1,356 @@
...
+        "listen.mode" = "0660";
+        "listen.owner" = user;
+        "listen.group" = group;
+        "clear_env" = "no";
+      } // cfg.poolConfig;
+    };
+
+    systemd.services.phpfpm-firefly-iii.serviceConfig = {
+      EnvironmentFile = "/run/phpfpm/firefly-iii-env.conf";
+      ExecStartPost = "${pkgs.runtimeShell} -c 'rm /run/phpfpm/firefly-iii-env.conf'";
```
SuperSandro2000, line 291:
```suggestion
      ExecStartPost = "${pkgs.coreutils}/bin/rm /run/phpfpm/firefly-iii-env.conf";
```
we can create scripts easily with postStart

nixos/modules/services/web-apps/firefly-iii.nix
```diff
@@ -0,0 +1,356 @@
...
+
+  #
+  # Config
+  #
+  config = mkIf cfg.enable {
+
+    # PHP
+    services.phpfpm.pools.firefly-iii = {
+      inherit user;
+      inherit group;
```
SuperSandro2000, line 276:
```suggestion
      inherit user group;
```

nixos/modules/services/web-apps/firefly-iii.nix
```diff
@@ -0,0 +1,356 @@
...
+      };
+    };
+  };
+
+  #
+  # Config
+  #
+  config = mkIf cfg.enable {
+
+    # PHP
```
SuperSandro2000, line 273:
```suggestion
  config = mkIf cfg.enable {
```

nixos/modules/services/web-apps/firefly-iii.nix
```diff
@@ -0,0 +1,356 @@
...
+              at "local".
+            '';
+          };
+          APP_URL = mkOption {
+            description = mdDoc ''
+              The root URL that you want to host Firefly III on. All URLs
+              in Firefly III will be generated using this value.
+            '';
+            example = "https://example.com";
+            default = "http://localhost";
```
SuperSandro2000, line 251:
```suggestion
```
since this is likely wrong, we should not default to it

nixos/modules/services/web-apps/firefly-iii.nix
```diff
@@ -0,0 +1,356 @@
...
+      '';
+      type = submodule {
+        freeformType = format.type;
+        options = {
+          APP_ENV = mkOption {
+            default = "local";
+            type = enum [ "production" "local" "testing" ];
+            description = mdDoc ''
+              Top level option governing Firefly's behaviour. This is best left
+              at "local".
```
SuperSandro2000, line 242:
```suggestion
              Top level option governing Firefly's behaviour. This is best left at "local".
```

nixos/modules/services/web-apps/firefly-iii.nix
```diff
@@ -0,0 +1,356 @@
...
+            '';
+          };
+          APP_URL = mkOption {
+            description = mdDoc ''
+              The root URL that you want to host Firefly III on. All URLs
+              in Firefly III will be generated using this value.
```
SuperSandro2000, line 248:
```suggestion
              The root URL that you want to host Firefly III on.
              All URLs in Firefly III will be generated using this value.
```
Lets make this easy to read for everyone

nixos/modules/services/web-apps/firefly-iii.nix
```diff
@@ -0,0 +1,356 @@
...
+        Options for firefly-iii configuration. Refer to
+        <https://github.com/firefly-iii/firefly-iii/blob/main/.env.example> for
+        details on supported values. All <option>_FILE values supported by
+        upstream are supported here.
+      '';
+      example = literalExpression ''
+        {
+          APP_ENV = "production";
+          APP_URL = "https://example.com";
+          APP_KEY = "SomeKeyOfExactly32Chars";
```
SuperSandro2000, line 224: Can we store the APP_KEY securely?

nixos/modules/services/web-apps/firefly-iii.nix
```diff
@@ -0,0 +1,356 @@
...
+    group = mkOption {
+      type = str;
+      default = defaultGroup;
+      description = mdDoc ''
+        Group under which firefly-iii runs. It is best to set this to the group
+        of whatever webserver is being used as the frontend.
+      '';
+    };
+
+    package = mkOption {
```
SuperSandro2000, line 183: Please use mkPackageOption

nixos/modules/services/web-apps/firefly-iii.nix
```diff
@@ -0,0 +1,356 @@
...
+
+    exit 0
+  '';
+
+  commonServiceConfig = {
+    Type = "oneshot";
+    User = user;
+    Group = group;
+    StateDirectory = "firefly-iii";
+    WorkingDirectory = "${cfg.package}";
```
SuperSandro2000, line 125:
```suggestion
    WorkingDirectory = cfg.package;
```

nixos/modules/services/web-apps/firefly-iii.nix
```diff
@@ -0,0 +1,356 @@
...
+    #!${pkgs.runtimeShell}
+
+    source ${cfg.package}/.env
+
+    ${fileenv-func}
+
+    ${firefly-php} ${artisan} firefly-iii:create-database
+    ${firefly-php} ${artisan} migrate --seed --no-interaction --force
+    ${firefly-php} ${artisan} firefly-iii:decrypt-all
+    ${firefly-php} ${artisan} cache:clear
```
SuperSandro2000, line 108: Isn't it enough to clear cache once?

nixos/modules/services/web-apps/firefly-iii.nix
```diff
@@ -0,0 +1,356 @@
...
+    ${firefly-php} ${artisan} firefly-iii:upgrade-database
+    ${firefly-php} ${artisan} firefly-iii:correct-database
+    ${firefly-php} ${artisan} firefly-iii:report-integrity
+    ${firefly-php} ${artisan} firefly-iii:laravel-passport-keys
+    ${firefly-php} ${artisan} cache:clear
+
+    mv /tmp/firefly-iii-env.conf /run/phpfpm/firefly-iii-env.conf
+
+    exit 0
```
SuperSandro2000, line 117: Since we don't set -e this is really dangerous as it masks failures but even when using set -e, we don't really need this

nixos/modules/services/web-apps/firefly-iii.nix
```diff
@@ -0,0 +1,356 @@
...
+    )
+
+    for e in "''${envs[@]}"; do
+      file_env "$e"
+    done
+  '';
+
+  firefly-iii-maintenance = pkgs.writeScript "firefly-iii-maintenance.sh" ''
+    #!${pkgs.runtimeShell}
+
```
SuperSandro2000, line 100:
```suggestion
  firefly-iii-maintenance = pkgs.writeShellScript "firefly-iii-maintenance.sh" ''
```

nixos/modules/services/web-apps/firefly-iii.nix
```diff
@@ -0,0 +1,356 @@
...
+  firefly-pkg = pkgs.firefly-iii.override {
+    dataDir = cfg.dataDir;
+    settings = cfg.settings;
+  };
+
+  firefly-php = "${cfg.package.phpPackage}/bin/php";
+  artisan = "${cfg.package}/artisan";
+
+  fileenv-func = ''
+    file_env() {
```
SuperSandro2000, line 25: Any chance we can simplify this?

nixos/modules/services/web-apps/firefly-iii.nix
```diff
@@ -0,0 +1,356 @@
...
+  inherit (lib.types) attrsOf oneOf str int bool path package enum submodule;
+
+  cfg = config.services.firefly-iii;
+  format = pkgs.formats.ini { };
+
+  user = cfg.user;
+  group = cfg.group;
+
+  defaultUser = "firefly-iii";
+  defaultGroup = defaultUser;
```
SuperSandro2000, line 14:
```suggestion
```
Please write this out explicit

pkgs/by-name/fi/firefly-iii/package.nix
```diff
@@ -0,0 +1,69 @@
...
+  };
+  firefly-iii-env = (fireflyEnvVars settings);
+in
+
+stdenvNoCC.mkDerivation {
+  pname = "firefly-iii";
+  inherit version;
+
+  src = fetchzip {
+    url = "https://github.com/firefly-iii/firefly-iii/releases/download/v${version}/FireflyIII-v${version}.zip";
```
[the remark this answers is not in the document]
savyajha: It used to be done using `composer`, but the author does not recommend it anymore. They recommend downloading from source now. (Source: https://docs.firefly-iii.org/how-to/firefly-iii/installation/self-managed/)

nixos/modules/services/web-apps/firefly-iii.nix
savyajha: We can, we can use `APP_KEY_FILE` instead. I'll change the example to that.

nixos/modules/services/web-apps/firefly-iii.nix
savyajha: Taking instructions from two different sources, my understanding is that it needs to be done twice:

- https://docs.firefly-iii.org/how-to/firefly-iii/upgrade/self-managed/
- https://dev.azure.com/Firefly-III/_git/MainImage?path=%2Fentrypoint-fpm.sh

nixos/modules/services/web-apps/firefly-iii.nix
savyajha: This is taken from [upstream](https://dev.azure.com/Firefly-III/_git/MainImage?path=%2Fentrypoint-fpm.sh). We probably can simplify it, but it seems to be a widely used function in the docker world.

pkgs/by-name/fi/firefly-iii/package.nix
```diff
@@ -0,0 +1,69 @@
...
+    tests = nixosTests.firefly-iii;
+  };
+
+  postInstall = ''
+    rm -R $out/storage $out/bootstrap/cache
+    ln -s ${dataDir}/storage $out/storage
+    ln -s ${dataDir}/cache $out/bootstrap/cache
+    cat <<EOF > $out/.env
+      ${firefly-iii-env}
+    EOF
```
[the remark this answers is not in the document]
savyajha: Whie the env file is easily moved into the module, I'm less sure about how to go about moving the storage and cache symlink directories to a wrapper package. `symlinkJoin` would lead to a million different symlinks given the sheer number of packages in the `vendor` directory. Should I just copy the unwrapped package wholesale?

pkgs/by-name/fi/firefly-iii/package.nix
```diff
@@ -0,0 +1,69 @@
...
+    mkKeyValue = lib.flip lib.generators.mkKeyValueDefault "=" {
+      mkValueString = v: with builtins;
+        if isInt             v then toString v
+        else if isString     v then "\"${v}\""
+        else if true  ==     v then "True"
+        else if false ==     v then "False"
+        else throw "unsupported type ${typeOf v}: ${(lib.generators.toPretty {}) v}";
+    };
+  };
+  firefly-iii-env = (fireflyEnvVars settings);
```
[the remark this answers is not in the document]
savyajha: Moved this into the module

pkgs/by-name/fi/firefly-iii/package.nix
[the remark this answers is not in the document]
savyajha: @SuperSandro2000 I tried doing this using a variety of methods, but the service doesn't work properly if the source files are symlinked in a wrapper. The only way I've found that does work is the method used by bookstack, which is to add a symlink to the data directory in the original unwrapped package.

nixos/tests/firefly-iii.nix
```diff
@@ -0,0 +1,37 @@
...
+        locations = {
+          "/" = {
+            index = "index.php";
+            tryFiles = "$uri $uri/ /index.php?$query_string";
+          };
+          "~ \.php$".extraConfig = ''
+            fastcgi_pass unix:${config.services.phpfpm.pools."firefly-iii".socket};
+            fastcgi_index index.php;
+            include ${config.services.nginx.package}/conf/fastcgi_params;
+            include ${pkgs.nginx}/conf/fastcgi.conf;
```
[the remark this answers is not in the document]
savyajha: If I don't use both here the test fails. I've switched it to use `config.services.nginx.package` everywhere.

nixos/modules/services/web-apps/firefly-iii.nix
savyajha: While I've changed it to use `mkPackageOption`, I've not yet found there to be a good method of passing something like `cfg.dataDir` to the default package there. I'd appreciate some guidance on how to do it given I'd like the default package to be `pkgs.firefly-iii.override { dataDir = cfg.dataDir; }`.

pkgs/by-name/fi/firefly-iii/package.nix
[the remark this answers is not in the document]
SuperSandro2000: It would be a lot more trustworthy if we could build the project with buildComposerProject

pkgs/by-name/fi/firefly-iii/package.nix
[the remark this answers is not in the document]
SuperSandro2000: If we can't easily do it, then we keep it like this.

nixos/tests/firefly-iii.nix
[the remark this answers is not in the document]
SuperSandro2000:
```diff
 ▶ diff ./result/conf/fastcgi_params ./result/conf/fastcgi.conf | cat
--- ./result/conf/fastcgi_params    1970-01-01 01:00:01.000000000 +0100
+++ ./result/conf/fastcgi.conf  1970-01-01 01:00:01.000000000 +0100
@@ -1,4 +1,5 @@

+fastcgi_param  SCRIPT_FILENAME    $document_root$fastcgi_script_name;
 fastcgi_param  QUERY_STRING       $query_string;
 fastcgi_param  REQUEST_METHOD     $request_method;
 fastcgi_param  CONTENT_TYPE       $content_type;
 ```

File: nixos/modules/services/web-apps/firefly-iii.nix (hunk 13 above)
Comment (SuperSandro2000) [reply nixos/modules/services/web-apps/firefly-iii.nix]: You cann add apply to the option to do that. eg https://github.com/NixOS/nixpkgs/blob/f59f29659e1db1130f87936ef9f2a11a71b3a99c/nixos/modules/programs/steam.nix#L45

File: nixos/modules/services/web-apps/firefly-iii.nix (hunk 18 above)
Comment (SuperSandro2000) [reply nixos/modules/services/web-apps/firefly-iii.nix:19]: Ah, if it is just copied from wordpress, then that's fine for me.

File: pkgs/by-name/fi/firefly-iii/package.nix (hunk 20 above)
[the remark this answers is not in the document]
Comment (PatrickDaG) [reply pkgs/by-name/fi/firefly-iii/package.nix:19]: I tried building it from source and while I still think it's possible it seems like quite a lot of work and boilerplate. As of the latest version, the project is no longer only composer but needs npm as well to build the frontend. So we would need to somehow split the project into front- and backend, build those using buildNPM and buildComposer and the merge them again somehow. As I said I think it should be possible but not trivial.

**Comment by PatrickDaG** [comment]
I would suggest adding a nginx enable option for ease of use. This is the ngixn setup that works for me:
```nix
services.nginx = mkIf cfg.enableNginx {
      enable = true;
      recommendedTlsSettings = mkDefault true;
      recommendedOptimisation = mkDefault true;
      recommendedGzipSettings = mkDefault true;
      virtualHosts.${cfg.virtualHost} = {
        root = "${package}/public";
        locations = {
          "/" = {
            tryFiles = "$uri $uri/ /index.php?$query_string";
            index = "index.php";
            extraConfig = ''
              sendfile off;
            '';
          };
          "~ \.php$" = {
            extraConfig = ''
              include ${config.services.nginx.package}/conf/fastcgi_params ;
              fastcgi_param SCRIPT_FILENAME $request_filename;
              fastcgi_param modHeadersAvailable true; #Avoid sending the security headers twice
              fastcgi_pass unix:${config.services.phpfpm.pools.firefly-iii.socket};
            '';
          };
        };
      };
```

savyajha:
@PatrickDaG I didn't do this because I believe that these configurations can vary quite a bit between different deployments and because I'm not sure whether nixos wants to privilege nginx over, say, caddy or apache. I know that a lot of projects setup nginx but I've not found it to be universal across web-apps.

And, of course, that also brings up the question of whether we should also setup mysql or postgres if the user sets that in the config.

However, if you believe strongly about this I'll add an nginx option as well.

PatrickDaG:
I think a baseline working setup for nginx is nice to have, even if people deploy differently as it can show which endpoints exist and which extra config you may need. Of course, it would be preferable to have that for all web servers, the same as it would be the best to have automatic setup for any database people may choose.

But for the sake of maintainability I would say it's enough if you setup the software stack you use and if someone wants a complete setup for e.g., Postgres and Apache, they can always post their own pull request.

If you have a very special nginx setup and thus no need for an option as part of the module I can also offer to help maintain that part.  I would use this option, making it easier to maintain, and then you don't have to maintain things you may not need.

savyajha:
@PatrickDaG Let me add the nginx option and push this again in a bit. I might also have a way of building the package from source (though it leaves a few translations out).

PatrickDaG:
Well. Funny coincidence, I was also just trying to get a build from source to work: [This](https://git.lel.lol/patrick/nix-config/src/commit/bfc7edcdde260cebb0f9dbccfa88d31863da46bd/pkgs/firefly-iii.nix) seems to work but as you said I think some translations are missing as well as some javascript 
```
Only in result_old/public/v1/js: app.js
Only in result_old/public/v1/js: app.js.LICENSE.txt
Only in result_old/public/v1/js: app_vue.js
Only in result_old/public/v1/js: app_vue.js.LICENSE.txt
Only in result_old/public/v1/js: create_transaction.js
Only in result_old/public/v1/js: create_transaction.js.LICENSE.txt
Only in result_old/public/v1/js: edit_transaction.js
Only in result_old/public/v1/js: edit_transaction.js.LICENSE.txt
Only in result_old/public/v1/js: profile.js
Only in result_old/public/v1/js: profile.js.LICENSE.txt
Only in result_old/public/v1/js/webhooks: create.js
Only in result_old/public/v1/js/webhooks: create.js.LICENSE.txt
Only in result_old/public/v1/js/webhooks: edit.js
Only in result_old/public/v1/js/webhooks: edit.js.LICENSE.txt
Only in result_old/public/v1/js/webhooks: index.js
Only in result_old/public/v1/js/webhooks: index.js.LICENSE.txt
Only in result_old/public/v1/js/webhooks: show.js
Only in result_old/public/v1/js/webhooks: show.js.LICENSE.txt
```

So far with my short tests it seems to work, but I'm fully confident.

savyajha:
So turns out that's only because those files are built when the develop branch gets built. It shouldn't be a problem until the next version (6.1) comes along.

nixos/modules/services/web-apps/firefly-iii.nix
savyajha: This doesn't use `mkPackageOption`. I'll change it to use `mkOption` as per your example.

nixos/modules/services/web-apps/firefly-iii.nix
savyajha: Changed this to a single time: it doesn't seem to make much difference.

savyajha:
@PatrickDaG I've made all the requisite changes to this, I think. Please do go over it and see if I've missed anything. If you'd like to be added as a maintainer please tell me, I shall add you and then mark this as ready for review.

pkgs/by-name/fi/firefly-iii/package.nix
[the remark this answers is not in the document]
savyajha: I've built it from source. There's a bit of file duplication, but I don't think it's too bad yet.

nixos/modules/services/web-apps/firefly-iii.nix
```diff
@@ -0,0 +1,387 @@
...
+        else if isString     v then "\"${v}\""
+        else if true  ==     v then "True"
+        else if false ==     v then "False"
+        else throw "unsupported type ${typeOf v}: ${(lib.generators.toPretty {}) v}";
+    };
+  };
+
+  firefly-iii-env = settingsFormat.generate "firefly-iii-env" cfg.settings;
+
+  touch-sqlite = if cfg.settings.DB_CONNECTION == "sqlite" then "touch ${cfg.dataDir}/storage/database/database.sqlite" else "";
```
PatrickDaG, line 104:
```suggestion
  touch-sqlite =  lib.optionalString (cfg.settings.DB_CONNECTION == "sqlite") "touch ${cfg.dataDir}/storage/database/database.sqlite" ;
```

nixos/modules/services/web-apps/firefly-iii.nix
```diff
@@ -0,0 +1,387 @@
...
+      file_env "$e"
+    done
+  '';
+
+  settingsFormat = pkgs.formats.keyValue {
+    mkKeyValue = lib.flip lib.generators.mkKeyValueDefault "=" {
+      mkValueString = v: with builtins;
+        if isInt             v then toString v
+        else if isString     v then "\"${v}\""
+        else if true  ==     v then "True"
```
PatrickDaG, line 96:
```suggestion
        else if isBool       v lib.boolToString v
```

PatrickDaG (approved):
Looks great. Thank you for all the work. I suggested two library functions that I think make the code cleaner and might be of interest, but they don't change anything of importance, so you can also just ignore them.

As I am planning to use this module for the foreseeable future, with nginx, it would be my pleasure to help maintain. But since my entry in the maintainers list is still in the progress of getting merged, I don't think you should add me and have to wait for that.

I'll just add myself if I ever get to post another pull request for this module.

nixos/modules/services/web-apps/firefly-iii.nix
savyajha: Done!

nixos/modules/services/web-apps/firefly-iii.nix
savyajha: I've removed this completely. Turns out if you use enough nix you don't need either the env file or the huge function from Wordpress.

savyajha:
@SuperSandro2000 I think I'm satisfied with this at my end. Would appreciate a once over and maybe some more testing.

pkgs/by-name/fi/firefly-iii/package.nix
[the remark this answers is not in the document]
ambroisie: :+1: on building from source if possible.

nixos/modules/services/web-apps/firefly-iii.nix
```diff
@@ -0,0 +1,356 @@
...
+  defaultGroup = "firefly-iii";
+
+  artisan = "${cfg.package}/artisan";
+
+  env-file-values = filterAttrs (n: v: hasSuffix "_FILE" n) cfg.settings;
+  env-nonfile-values = filterAttrs (n: v: ! hasSuffix "_FILE" n) cfg.settings;
+
+  fileenv-func = ''
+    ${concatLines (mapAttrsToList
+      (n: v: "export ${n}=\"${toString v}\"")
```
ambroisie, line 25: I'd use `lib.strings.toShellVars` + `lib.concatMapStringsSep "\n" (v: "export ${v}") attrNames`

nixos/modules/services/web-apps/firefly-iii.nix
```diff
@@ -0,0 +1,356 @@
...
+      "d ${cfg.dataDir}/storage/database           0700 ${user} ${group} - -"
+      "d ${cfg.dataDir}/storage/export             0700 ${user} ${group} - -"
+      "d ${cfg.dataDir}/storage/framework          0700 ${user} ${group} - -"
+      "d ${cfg.dataDir}/storage/framework/cache    0700 ${user} ${group} - -"
+      "d ${cfg.dataDir}/storage/framework/sessions 0700 ${user} ${group} - -"
+      "d ${cfg.dataDir}/storage/framework/views    0700 ${user} ${group} - -"
+      "d ${cfg.dataDir}/storage/logs               0700 ${user} ${group} - -"
+      "d ${cfg.dataDir}/storage/upload             0700 ${user} ${group} - -"
+      "d ${cfg.dataDir}/cache                      0700 ${user} ${group} - -"
+    ];
```
ambroisie, line 340: Use `systemd.tmpfiles.settings` instead, which makes it more composable.

savyajha: Made the change.

nixos/modules/services/web-apps/firefly-iii.nix
savyajha: Changed this too. The entire env file is a mess anyway, so changed the approach to use `lib.strings.toShellVar` instead and then just used plain bash loops to do everything else.

pkgs/by-name/fi/firefly-iii/package.nix
```diff
@@ -0,0 +1,60 @@
...
+  };
+in
+
+phpPackage.buildComposerProject (finalAttrs: {
+  inherit pname src version;
+
+  vendorHash = "sha256-CVGKyyLp5hjjpEulDNEYfljU4OgPBaFcYQQAUf6GeGs=";
+
+  passthru = {
+    inherit phpPackage;
```
drupol, line 41: Why is this needed?

pkgs/by-name/fi/firefly-iii/package.nix
```diff
@@ -0,0 +1,60 @@
...
+    rev = "v${version}";
+    hash = "sha256-85zI8uCyyoCflzxDkvba6FWa9B3kh179DJfQ2Um6MGM=";
+  };
+
+  assets = buildNpmPackage {
+    pname = "${pname}-assets";
+    inherit version src;
+    npmDepsHash = "sha256-wuPUE6XuzzgKjpxZVgwh2wGut15M61WSBFG+YIZwOFM=";
+    dontNpmBuild = true;
+    installPhase = ''
```
drupol, line 26: Please add missing phase hooks.

I know it's not a very important things, but I mention this to make sure we are consistent when using phases. Adding those phase hooks by default also help people when they do a search in the repository, to avoid misleading future potential contributors.

drupol (commented):
Nice !

Opinionated: I wonder why `nginx` is used by default, `caddy` is so much greater.

pkgs/by-name/fi/firefly-iii/package.nix
savyajha: So that the exact same package can be used by php-fpm in the module. That way, in case firefly-iii requires a different set of extensions in a future iteration we can rest easy knowing this variable references that same php package.

drupol: Excellent, thanks, TIL !

drupol: I think this is something we can improve in the PHP Composer Builder, right ?

pkgs/by-name/fi/firefly-iii/package.nix
savyajha: I don't exactly get what you mean. Does `buildNpmPackage` not automatically run checkPhase, fixupPhase etc.?

pkgs/by-name/fi/firefly-iii/package.nix
savyajha: By automatically exporting this variable? Yup, I'd agree.

pkgs/by-name/fi/firefly-iii/package.nix
drupol: If you are overriding `configurePhase`, `buildPhase`, `checkPhase`, `installPhase` or any other phase, you should not forget about explicitly running pre and post-phase hooks like [their](https://github.com/NixOS/nixpkgs/blob/d71a03ad695407dd482ead32d3eddff50092a1c3/pkgs/stdenv/generic/setup.sh#L953) [original](https://github.com/NixOS/nixpkgs/blob/d71a03ad695407dd482ead32d3eddff50092a1c3/pkgs/stdenv/generic/setup.sh#L1008) [definitions](https://github.com/NixOS/nixpkgs/blob/d71a03ad695407dd482ead32d3eddff50092a1c3/pkgs/stdenv/generic/setup.sh#L1037) [do](https://github.com/NixOS/nixpkgs/blob/d71a03ad695407dd482ead32d3eddff50092a1c3/pkgs/stdenv/generic/setup.sh#L1078).

It is generally expected that an appropriate pre-phase hook (e.g. `preBuild`) hook will run at the beginning of a phase (e.g. `buildPhase`) and post-phase hook (e.g. `postBuild`) will run at the end. Hooks are normally ran as a part of a phase so if you override a phase as a whole, you will need to add `runHook hookName` calls (e.g. `runHook preBuild`) manually.

Having phases run pre/post-phase hooks is important because many setup hooks insert their own code into them – omitting a hook might therefore prevent some setup hooks required for proper functionality of a package from running. Additionally, hooks are often inserted by developers into the package expression and by users when overriding a package using `overrideAttrs`. Not running them can thus cause confusion why their code is not executed.

## Examples

### Before

```nix
  installPhase = ''
    your commands
  '';
```

### After

```nix
  installPhase = ''
    runHook preInstall

    your commands

    runHook postInstall
  '';
```

Source: https://github.com/jtojnar/nixpkgs-hammering/blob/main/explanations/missing-phase-hooks.md

pkgs/by-name/fi/firefly-iii/package.nix
drupol: https://github.com/NixOS/nixpkgs/pull/305256

pkgs/by-name/fi/firefly-iii/package.nix
savyajha: @drupol It's already been done. :) The `installPhase` in the package is:

```
runHook preInstall
npm run build
cp -r ./public $out/
runHook postInstall
'';
```

drupol: Oh sorry. Ignore this.

pkgs/by-name/fi/firefly-iii/package.nix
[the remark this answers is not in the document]
savyajha: It's built from source now.

savyajha:
@drupol Is there anything else I need to do here before this is ready for a merge?

stackcoder:
Hey @savyajha,
thank you, for your great work packing firelfy-iii for us! I already tried to set it up using your module and might provide some little feedback on this:

In the current implementation nix fails (even when bundled nginx is not used) without setting the parameter `firefly-iii.virtualHost`. Especially when using a reverse proxy, a default value of `localhost` just works fine in my setups.
```
… while evaluating the option `firefly.services.firefly-iii.virtualHost':
error: The option `firefly.services.firefly-iii.virtualHost' is used but not defined.
```

Also `app.js` got delivered without correct mime-type, so the web app is broken when the `X-Content-Type-Options: nosniff` header is set. I'm still working out what's causing it, maybe someone else will stumble across it.

savyajha:
@PatrickDaG Can you let me know what these options are? While I’ve been dogfooding this module, I might not have covered everything. I was also of the opinion that the defaults (set my the application itself) are sensible.

I will get the changes into a PR this weekend.

nixos/modules/services/web-apps/firefly-iii.nix
```diff
@@ -0,0 +1,367 @@
...
+        };
+      };
+    };
+  };
+
+  config = mkIf cfg.enable {
+
+    services.firefly-iii = {
+      settings = {
+        APP_URL = cfg.virtualHost;
```
PatrickDaG, line 247: This should be an mkDefault. So far, that's tho only one I've found.

stackcoder:
> I think the mime-type mismatch is actually a 404 not found. In the old version of this package the `app.js` was not built correctly leading to nginx returning a 404 which has mime-type html resulting in the error you see. I've just tested the newest version, and it seems to be fixed.

I can confirm updating from `6.1.13` -> `6.1.15` fixes the `app.js` issue.

nixos/modules/services/web-apps/firefly-iii.nix
savyajha: Why would you want the virtual host and the APP URL to be different?

PatrickDaG: I have another reverse proxy in front of the firefly host, that does a bunch of preprocessing in the request, e.g. ssl termination. So I need the APP_URL to resolve to the first proxy using https while the virtual host itself ist done using plain http.

PatrickDaG: In that case, the `virtualHost` is no longer actually a virtual host compatible option. I would leave it as is, as I feel like with weird string replacements we remove options from end users rather than enabling them to configure it appropriately.

savyajha: The only thing this string replacement does is removes the "http(s)://" string from the front and the trailing backslash from the virtual host. It's nothing more than a sanity check. If the virtual host definition doesn't have "http(s)://" in the front then the string remains unmodified.

PR #268796, opened 2023-11-20
secure-delete: init at 3.1

xdettlaff:
## Description of changes

Gutmann method based tools for securely wiping data from files, free disk space, swap and memory: srm, sfill, sswap and sdmem.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
- For non-Linux: Is sandboxing enabled in `nix.conf`? (See [Nix manual](https://nixos.org/manual/nix/stable/command-ref/conf-file.html))
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:
  - [NixOS test(s)](https://nixos.org/manual/nixos/unstable/index.html#sec-nixos-tests) (look inside [nixos/tests](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests))
  - and/or [package tests](https://nixos.org/manual/nixpkgs/unstable/#sec-package-tests)
  - or, for functions and "core" functionality, tests in [lib/tests](https://github.com/NixOS/nixpkgs/blob/master/lib/tests) or [pkgs/test](https://github.com/NixOS/nixpkgs/blob/master/pkgs/test)
  - made sure NixOS tests are [linked](https://nixos.org/manual/nixpkgs/unstable/#ssec-nixos-tests-linking) to the relevant packages
- [ ] Tested compilation of all packages that depend on this change using `nix-shell -p nixpkgs-review --run "nixpkgs-review rev HEAD"`. Note: all changes have to be committed, also see [nixpkgs-review usage](https://github.com/Mic92/nixpkgs-review#usage)
- [ ] Tested basic functionality of all binary files (usually in `./result/bin/`)
- [23.11 Release Notes](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2311.section.md) (or backporting [23.05 Release notes](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2305.section.md))
  - [ ] (Package updates) Added a release notes entry if the change is major or breaking
  - [ ] (Module updates) Added a release notes entry if the change is significant
  - [ ] (Module addition) Added a release notes entry if adding a new NixOS module
- [ ] Fits [CONTRIBUTING.md](https://github.com/NixOS/nixpkgs/blob/master/CONTRIBUTING.md).

PR #302269, opened 2024-04-07
thc-secure-delete: init at 3.1.1

Tochiaha:
## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [x] `sandbox = true`
- [x] Tested, as applicable:

---

### DIFF
--- a/pkgs/by-name/th/thc-secure-delete/package.nix
+++ b/pkgs/by-name/th/thc-secure-delete/package.nix
@@ -1,0 +1,43 @@
+{ lib
+, stdenv
+, fetchFromGitHub
+, installShellFiles
+, pkg-config
+}:
+
+stdenv.mkDerivation (finalAttrs: {
+  pname = "thc-secure-delete";
+  version = "3.1.1";
+
+  src = fetchFromGitHub {
+    owner = "gordrs";
+    repo = "thc-secure-delete";
+    rev = "v${finalAttrs.version}";
+    hash = "sha256-hvWuxFkzhOSCplPtyjRtn36bIk6KdPBcpr3lAmiAyfE=";
+  };
+
+  nativeBuildInputs = [ pkg-config installShellFiles ];
+
+  makeFlags = [
+    "CC=${stdenv.cc.targetPrefix}cc"
+  ];
+
+  installPhase = ''
+    runHook preInstall
+
+    install -Dt $out/bin/ -m755 srm sdmem sswap sfill
+    installManPage *.1
+
+    runHook postInstall
+  '';
+
+  meta = with lib; {
+    description = "THC's Secure Delete tools";
+    homepage = "https://github.com/gordrs/thc-secure-delete";
+    changelog = "https://github.com/gordrs/thc-secure-delete/blob/v${finalAttrs.version}/CHANGES";
+    license = licenses.gpl2Only;
+    maintainers = with maintainers; [ tochiaha ];
+    mainProgram = "srm";
+    platforms = platforms.all;
+  };
+})


Aleksanaa:
The latest change was already 5 years ago. Do you really want it (for a good reason)? After reading the readme I think `shred` (installed on NixOS by default) should cover its use cases.

Tochiaha:
> The latest change was already 5 years ago. Do you really want it (for a good reason)? After reading the readme I think `shred` (installed on NixOS by default) should cover its use cases.

does shred really cover all its use cases, i know srm is already packaged but what about sdmem sfill and sswap they apply different techniques though compared to shred.

pkgs/by-name/th/thc-secure-delete/package.nix
```diff
@@ -0,0 +1,36 @@
+{ lib
+, stdenv
+, fetchFromGitHub
+, pkg-config
+}:
+
+stdenv.mkDerivation rec {
```
drupol, line 7: I think we should use the `finalAttrs` pattern here.

```nix
mkDerivation rec {
   foo = 1;
   bar = foo + 1;
}
```

pkgs/by-name/th/thc-secure-delete/package.nix
```diff
@@ -0,0 +1,36 @@
...
+  src = fetchFromGitHub {
+    owner = "gordrs";
+    repo = "thc-secure-delete";
+    rev = "v${version}";
+    hash = "sha256-hvWuxFkzhOSCplPtyjRtn36bIk6KdPBcpr3lAmiAyfE=";
+  };
+
+  build-systems = [ pkg-config ];
+
+  installPhase = ''
```
drupol, line 20: If you are overriding `configurePhase`, `buildPhase`, `checkPhase`, `installPhase` or any other phase, you should not forget about explicitly running pre and post-phase hooks like [their](https://github.com/NixOS/nixpkgs/blob/d71a03ad695407dd482ead32d3eddff50092a1c3/pkgs/stdenv/generic/setup.sh#L953) [original](https://github.com/NixOS/nixpkgs/blob/d71a03ad695407dd482ead32d3eddff50092a1c3/pkgs/stdenv/generic/setup.sh#L1008) [definitions](https://github.com/NixOS/nixpkgs/blob/d71a03ad695407dd482ead32d3eddff50092a1c3/pkgs/stdenv/generic/setup.sh#L1037) [do](https://github.com/NixOS/nixpkgs/blob/d71a03ad695407dd482ead32d3eddff50092a1c3/pkgs/stdenv/generic/setup.sh#L1078).

It is generally expected that an appropriate pre-phase hook (e.g. `preBuild`) hook will run at the beginning of a phase (e.g. `buildPhase`) and post-phase hook (e.g. `postBuild`) will run at the end. Hooks are normally ran as a part of a phase so if you override a phase as a whole, you will need to add `runHook hookName` calls (e.g. `runHook preBuild`) manually.

Having phases run pre/post-phase hooks is important because many setup hooks insert their own code into them – omitting a hook might therefore prevent some setup hooks required for proper functionality of a package from running. Additionally, hooks are often inserted by developers into the package expression and by users when overriding a package using `overrideAttrs`. Not running them can thus cause confusion why their code is not executed.

## Examples

### Before

```nix
  installPhase = ''
    your commands
  '';
```

### After

```nix
  installPhase = ''
    runHook preInstall

    your commands

    runHook postInstall
  '';
```

Source: https://github.com/jtojnar/nixpkgs-hammering/blob/main/explanations/missing-phase-hooks.md

drupol (changes requested):

drupol (changes requested):
Build is broken on Darwin...

pkgs/by-name/th/thc-secure-delete/package.nix
```diff
@@ -0,0 +1,38 @@
...
+    repo = "thc-secure-delete";
+    rev = finalAttrs.version;
+    hash = "sha256-hvWuxFkzhOSCplPtyjRtn36bIk6KdPBcpr3lAmiAyfE=";
+  };
+
+  build-systems = [ pkg-config ];
+
+  installPhase = ''
+    runHook preInstall
+    mkdir -p $out/bin
```
drupol, line 22: Remove this, it's not needed.

pkgs/by-name/th/thc-secure-delete/package.nix
```diff
@@ -4,35 +4,39 @@
...
   pname = "thc-secure-delete";
-  version = "3.1.1";
+  inherit version;
 
   src = fetchFromGitHub {
     owner = "gordrs";
     repo = "thc-secure-delete";
     rev = finalAttrs.version;
-    hash = "sha256-hvWuxFkzhOSCplPtyjRtn36bIk6KdPBcpr3lAmiAyfE=";
+    sha256 = "hvWuxFkzhOSCplPtyjRtn36bIk6KdPBcpr3lAmiAyfE=";
```
drupol, line 19: Prefer `hash = ...` to `sha256 = ...` Both currently work, but the former will make it easier if the algorithm ever needs to change.

pkgs/by-name/th/thc-secure-delete/package.nix
```diff
@@ -4,35 +4,39 @@
 , pkg-config
 }:
 
+let
+  version = "3.1.1";
```
drupol, line 8: There's no need to have a `let...in` clause, use `finalAttrs.version` when you need the `version`.

pkgs/by-name/th/thc-secure-delete/package.nix
```diff
@@ -4,35 +4,39 @@
...
+    install -Dm755 sswap $out/bin
+    install -Dm755 sdmem $out/bin
     runHook postInstall
   '';
 
   meta = with lib; {
     description = "THC's Secure Delete tools";
     homepage = "https://github.com/gordrs/thc-secure-delete";
-    changelog = "https://github.com/gordrs/thc-secure-delete/blob/finalAttrs.rev/CHANGES";
+    changelog = "https://github.com/gordrs/thc-secure-delete/blob/finalAttrs.version/CHANGES";
```
drupol, line 37:
```suggestion
    changelog = "https://github.com/gordrs/thc-secure-delete/blob/${finalAttrs.version}/CHANGES";
```

drupol (changes requested):

pkgs/by-name/th/thc-secure-delete/package.nix
Tochiaha: so exactly how am i going to output the binaries if i remove mkdir -p $out/bin?

pkgs/by-name/th/thc-secure-delete/package.nix
Tochiaha: I introduced the let...in because it keeps on resolving to mirror not found error.

drupol (changes requested):

MikaelFangel:
Supersedes: #268796

pkgs/by-name/th/thc-secure-delete/package.nix
drupol: ... ?

drupol (approved):
Thanks!

drupol (dismissed):
There's a build issue on Darwin, can you check and eventually fix it?

Tochiaha:
> There's a build issue on Darwin, can you check and eventually fix it?

Had to check alternative, is the recent file changes acceptable?

pkgs/by-name/th/thc-secure-delete/package.nix
```diff
@@ -0,0 +1,47 @@
...
+  src = fetchFromGitHub {
+    owner = "gordrs";
+    repo = "thc-secure-delete";
+    rev = "v${finalAttrs.version}";
+    hash = "sha256-hvWuxFkzhOSCplPtyjRtn36bIk6KdPBcpr3lAmiAyfE=";
+  };
+
+  build-systems = [ pkg-config gcc ];
+
+  buildPhase = ''
```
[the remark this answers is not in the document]
reckenrode: If upstream supports building with clang, it should probably be substituting `$(CC)` for GCC instead of using the path to GCC. Also, substituting a dependency directly into the build phase probably breaks cross-compilation and definitely prevents `overrideSDK` from working should it be needed in the future.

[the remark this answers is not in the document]
reckenrode: This should also probably be overriding `postPatch`, so it can use the `buildPhase` for `make`.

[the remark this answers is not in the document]
reckenrode: And `--replace-fail` instead of `--replace` in case the `Makefile` changes and the substitution stops happening.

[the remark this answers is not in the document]
2xsaiko: And if it doesn't support building with clang, do the same as @reckenrode said, but use gccStdenv instead of stdenv. You should not use stdenvNoCC here in any case.

pkgs/by-name/th/thc-secure-delete/package.nix
```diff
@@ -0,0 +1,47 @@
...
+  version = "3.1.1";
+
+  src = fetchFromGitHub {
+    owner = "gordrs";
+    repo = "thc-secure-delete";
+    rev = "v${finalAttrs.version}";
+    hash = "sha256-hvWuxFkzhOSCplPtyjRtn36bIk6KdPBcpr3lAmiAyfE=";
+  };
+
+  build-systems = [ pkg-config gcc ];
```
2xsaiko, line 19: What is this supposed to do? This isn't used anywhere by nixpkgs or (as far as I can see unless github search just didn't show me anything) the upstream project. If you need pkg-config, use this:

```suggestion
  nativeBuildInputs = [ pkg-config ];
```

dblsaiko (changes requested):

pkgs/by-name/th/thc-secure-delete/package.nix
```diff
@@ -0,0 +1,47 @@
...
+    make
+  '';
+
+  installPhase = ''
+    runHook preInstall
+    install -Dm755 srm $out/bin
+    install -Dm755 sfill $out/bin
+    install -Dm755 sswap $out/bin
+    install -Dm755 sdmem $out/bin
+    install -Dm755 sfill.1 $out/share/man/man1/sfill.1.gz
```
viraptor, line 32: Those files are not actually gzipped, are they?

pkgs/by-name/th/thc-secure-delete/package.nix
```diff
@@ -0,0 +1,41 @@
...
+  version = "3.1.1";
+
+  src = fetchFromGitHub {
+    owner = "gordrs";
+    repo = "thc-secure-delete";
+    rev = "v${finalAttrs.version}";
+    hash = "sha256-hvWuxFkzhOSCplPtyjRtn36bIk6KdPBcpr3lAmiAyfE=";
+  };
+
+  build-systems = [ pkg-config ];
```
wegank, line 18:
```suggestion
  nativeBuildInputs = [ pkg-config ];
```

pkgs/by-name/th/thc-secure-delete/package.nix
```diff
@@ -0,0 +1,46 @@
+{ lib
+, stdenvNoCC # x86_64-darwin and aarch64-darwin cant find C compiler
```
wegank, line 2:
```suggestion
, stdenv
```

pkgs/by-name/th/thc-secure-delete/package.nix
```diff
@@ -0,0 +1,46 @@
...
+  src = fetchFromGitHub {
+    owner = "gordrs";
+    repo = "thc-secure-delete";
+    rev = "v${finalAttrs.version}";
+    hash = "sha256-hvWuxFkzhOSCplPtyjRtn36bIk6KdPBcpr3lAmiAyfE=";
+  };
+
+  build-systems = [ pkg-config gcc ];
+
+  buildPhase = ''
```
wegank, line 20: This can be dropped once `makeFlags` is added.

pkgs/by-name/th/thc-secure-delete/package.nix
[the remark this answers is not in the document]
wegank: It builds with Clang, just needs `CC=${stdenv.cc.targetPrefix}cc` in `makeFlags`.

pkgs/by-name/th/thc-secure-delete/package.nix
```diff
@@ -0,0 +1,46 @@
...
+    install -Dm755 sswap $out/bin
+    install -Dm755 sdmem $out/bin
+    install -Dm755 sfill.1 $out/share/man/man1/sfill.1.gz
+    install -Dm755 srm.1 $out/share/man/man1/srm.1.gz
+    install -Dm755 smem.1 $out/share/man/man1/smem.1.gz
+    install -Dm755 sswap.1 $out/share/man/man1/sswap.1.gz
+    runHook postInstall
+  '';
+
+  meta = with lib; {
```
wegank, line 38: `meta.platforms` is missing, probably should be set to `lib.platforms.unix` or `lib.platforms.all`.

pkgs/by-name/th/thc-secure-delete/package.nix
wegank: Not resolved, cc means Clang on macOS.

pkgs/by-name/th/thc-secure-delete/package.nix
```diff
@@ -0,0 +1,45 @@
+{ lib
+, gccStdenv # x86_64-darwin and aarch64-darwin cant find C compiler
```
wegank, line 2: Just use the normal `stdenv` and add `makeFlags` as mentioned. Using GCC on Darwin is highly unexpected.

pkgs/by-name/th/thc-secure-delete/package.nix
```diff
@@ -0,0 +1,46 @@
+{ lib
+, stdenvNoCC # x86_64-darwin and aarch64-darwin cant find C compiler
+, fetchFromGitHub
+, pkg-config
+}:
+
+stdenvNoCC.mkDerivation (finalAttrs: {
```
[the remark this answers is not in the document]
Tochiaha: using gccStdenv to prevent build failure when looking for gcc during compilation.

pkgs/by-name/th/thc-secure-delete/package.nix
Tochiaha: Ok

pkgs/by-name/th/thc-secure-delete/package.nix
wegank: Actually the whole `buildPhase`.

pkgs/by-name/th/thc-secure-delete/package.nix
```diff
@@ -0,0 +1,46 @@
...
+  '';
+
+  meta = with lib; {
+    description = "THC's Secure Delete tools";
+    homepage = "https://github.com/gordrs/thc-secure-delete";
+    changelog = "https://github.com/gordrs/thc-secure-delete/blob/${finalAttrs.version}/CHANGES";
+    license = licenses.gpl2Only;
+    maintainers = with maintainers; [ tochiaha ];
+    mainProgram = "srm";
+    platforms = platforms.all
```
wegank, line 44:
```suggestion
    platforms = platforms.all;
```

pkgs/by-name/th/thc-secure-delete/package.nix
```diff
@@ -0,0 +1,46 @@
...
+    install -Dm755 srm.1 $out/share/man/man1/srm.1.gz
+    install -Dm755 smem.1 $out/share/man/man1/smem.1.gz
+    install -Dm755 sswap.1 $out/share/man/man1/sswap.1.gz
+    runHook postInstall
+  '';
+
+  meta = with lib; {
+    description = "THC's Secure Delete tools";
+    homepage = "https://github.com/gordrs/thc-secure-delete";
+    changelog = "https://github.com/gordrs/thc-secure-delete/blob/${finalAttrs.version}/CHANGES";
```
wegank, line 40:
```suggestion
    changelog = "https://github.com/gordrs/thc-secure-delete/blob/v${finalAttrs.version}/CHANGES";
```

pkgs/by-name/th/thc-secure-delete/package.nix
Tochiaha: yes they are

pkgs/by-name/th/thc-secure-delete/package.nix
```diff
@@ -0,0 +1,46 @@
...
+  installPhase = ''
+    runHook preInstall
+    install -Dm755 srm $out/bin
+    install -Dm755 sfill $out/bin
+    install -Dm755 sswap $out/bin
+    install -Dm755 sdmem $out/bin
+    install -Dm755 sfill.1 $out/share/man/man1/sfill.1.gz
+    install -Dm755 srm.1 $out/share/man/man1/srm.1.gz
+    install -Dm755 smem.1 $out/share/man/man1/smem.1.gz
+    install -Dm755 sswap.1 $out/share/man/man1/sswap.1.gz
```
2xsaiko, line 33: Right now this copies the binaries so that $out/bin is the executable directly, also these commands can be shorter and man pages should probably not have the x bit

```
$ results/thc-secure-delete/bin --help
results/thc-secure-delete/bin: illegal option -- -
smem v3.1 (c) 1997-2003 by van Hauser / THC <vh@thc.org>
```

```suggestion
    install -d $out/bin $out/share/man/man1
    install -m755 srm sfill sswap sdmem $out/bin
    install -m644 sfill.1 srm.1 smem.1 sswap.1 $out/share/man/man1
```

dblsaiko (changes requested):

pkgs/by-name/th/thc-secure-delete/package.nix
Tochiaha: can you check the man pages again this time.

2xsaiko: It looks fine now even though I'm not exactly sure how :^) You're not adding the .gz ending explicitly anymore, I assume it gets added by the standard mkDerivation builder or something

dblsaiko (approved):

wegank (approved):

pkgs/by-name/th/thc-secure-delete/package.nix
```diff
@@ -0,0 +1,41 @@
...
+
+  makeFlags = [
+    "CC=${stdenv.cc.targetPrefix}cc"
+  ];
+
+  installPhase = ''
+    runHook preInstall
+    install -d $out/bin $out/share/man/man1
+    install -m755 srm sdmem sswap sfill $out/bin/
+    install -m644 *.1 $out/share/man/man1/
```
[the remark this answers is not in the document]
2xsaiko: then you probably don't need to explicitly create $out/share/man/man1 either, I assume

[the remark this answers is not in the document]
Aleksanaa: Yes

pkgs/by-name/th/thc-secure-delete/package.nix
```diff
@@ -0,0 +1,42 @@
...
+  nativeBuildInputs = [ pkg-config installShellFiles ];
+
+  makeFlags = [
+    "CC=${stdenv.cc.targetPrefix}cc"
+  ];
+
+  installPhase = ''
+    runHook preInstall
+    install -d $out/bin
+    install -m755 srm sdmem sswap sfill $out/bin/
```
eclairevoyant, line 28:
```suggestion
    install -Dm755 srm sdmem sswap sfill -t $out/bin/
```

Tochiaha: I previously tried this method but package build fails so i use the install -d and install -m saperately,

eclairevoyant: note the `-D` and `-t`.
it builds fine on my end.

Tochiaha: Yeah, the build works with -t.

eclairevoyant (approved):

PR #234201, opened 2023-05-26
coding-conventions: include the preceding upstream version

7c6f434c:
###### Description of changes

The only thing that seems clear after all the versioning discussions is that we do want the previous upstream-assigned version in the beginning when packaging something untagged.

###### Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
- [x] For non-Linux: Is `sandbox = true` set in `nix.conf`? (See [Nix manual](https://nixos.org/manual/nix/stable/command-ref/conf-file.html))
- N/A Tested basic functionality of all binary files (usually in `./result/bin/`)
- [ ] Fits [CONTRIBUTING.md](https://github.com/NixOS/nixpkgs/blob/master/CONTRIBUTING.md).

PR #302397, opened 2024-04-07
tiramisu: 2.0.20211107 -> 2.0-unstable-2023-03-29

### LINKED ISSUE #217845: Update request: tiramisu 2.0.20211107 → unstable-2022-11-19
The project seems slow with releasing the latest changes, which (among other things) includes [proper JSON formatted output](https://github.com/Sweets/tiramisu/pull/42)

- Package name: tiramisu
- Latest released version: 2.0.20211107, I suggest using the latest commit instead since there is no newer release yet

- Current version on the unstable channel: 2.0.20211107
- Current version on the stable/release channel: 2.0.20211107

- [X] Checked the [nixpkgs pull requests](https://github.com/NixOS/nixpkgs/pulls)

**Notify maintainers**

@wishfort36 @fortuneteller2k 

-----

JohnRTitor:
## Description of changes

tiramisu: 2.0.20211107 -> unstable-2023-03-29
Use a commit since the last release, to get the latest changes as the latest changes have not been released yet
Changelog: https://github.com/Sweets/tiramisu/compare/2.0.20211107...master
Closes https://github.com/NixOS/nixpkgs/issues/217845

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [x] Tested, as applicable:

---

Pinging maintainers: @wishfort36 @moni

### DIFF
--- a/pkgs/applications/misc/tiramisu/default.nix
+++ b/pkgs/applications/misc/tiramisu/default.nix
@@ -2,13 +2,15 @@
 
 stdenv.mkDerivation rec {
   pname = "tiramisu";
+  # FIXME: once a newer release in upstream is available
   version = "2.0-unstable-2023-03-29";
 
   src = fetchFromGitHub {
     owner = "Sweets";
-    repo = pname;
+    repo = "tiramisu";
+    # FIXME: use the current HEAD commit as upstream has no releases since 2021
     rev = "5dddd83abd695bfa15640047a97a08ff0a8d9f9b";
-    sha256 = "sha256-owYk/YFwJbqO6/dbGKPE8SnmmH4KvH+o6uWptqQtpfI=";
+    hash = "sha256-owYk/YFwJbqO6/dbGKPE8SnmmH4KvH+o6uWptqQtpfI=";
   };
 
   buildInputs = [ glib ];
--- a/pkgs/applications/misc/tiramisu/default.nix
+++ b/pkgs/applications/misc/tiramisu/default.nix
@@ -2,13 +2,13 @@
 
 stdenv.mkDerivation rec {
   pname = "tiramisu";
-  version = "2.0.20211107";
+  version = "2.0-unstable-2023-03-29";
 
   src = fetchFromGitHub {
     owner = "Sweets";
     repo = pname;
-    rev = version;
-    sha256 = "1n1x1ybbwbanibw7b90k7v4cadagl41li17hz2l8s2sapacvq3mw";
+    rev = "5dddd83abd695bfa15640047a97a08ff0a8d9f9b";
+    sha256 = "sha256-owYk/YFwJbqO6/dbGKPE8SnmmH4KvH+o6uWptqQtpfI=";
   };
 
   buildInputs = [ glib ];


pkgs/applications/misc/tiramisu/default.nix
```diff
@@ -2,13 +2,13 @@
 
 stdenv.mkDerivation rec {
   pname = "tiramisu";
-  version = "2.0.20211107";
+  version = "unstable-2023-03-29";
```
kirillrdy, line 5:
```suggestion
  version = "2.0.20211107-unstable-2023-03-29";
```

https://github.com/NixOS/nixpkgs/blob/master/pkgs/README.md#package-naming

pkgs/applications/misc/tiramisu/default.nix
```diff
@@ -2,13 +2,13 @@
...
 
   src = fetchFromGitHub {
     owner = "Sweets";
     repo = pname;
-    rev = version;
-    sha256 = "1n1x1ybbwbanibw7b90k7v4cadagl41li17hz2l8s2sapacvq3mw";
+    rev = "5dddd83abd695bfa15640047a97a08ff0a8d9f9b";
+    sha256 = "sha256-owYk/YFwJbqO6/dbGKPE8SnmmH4KvH+o6uWptqQtpfI=";
```
kirillrdy, line 11:
```suggestion
    hash = "sha256-owYk/YFwJbqO6/dbGKPE8SnmmH4KvH+o6uWptqQtpfI=";
```

https://nixos.org/manual/nixpkgs/unstable/#fetchfromgithub

> hash is currently preferred

wegank:
> Use a commit since the last release

Can this be provided as a patch via `fetchpatch`?

JohnRTitor:
> > Use a commit since the last release
> 
> Can this be provided as a patch via `fetchpatch`?

That would require [multiple patches](https://github.com/Sweets/tiramisu/compare/2.0.20211107...master), unless we squash them all into one.

wegank:
Alright, let's go unstable then.

pkgs/applications/misc/tiramisu/default.nix
JohnRTitor: Totally missed that, thanks for noticing!

kirillrdy:
there are only 5 PRs merged, if we find specific PRs that users need then its feasible to go via fetchpatch,

also maybe ping upstream for a realease ?

pkgs/applications/misc/tiramisu/default.nix
JohnRTitor: "2.0.20211107-unstable-2023-03-29" seems kind of confusing with two dates and all.
Will "2.0-unstable-2023-03-29" suffice?

kirillrdy: yeah, it does look a bit "sub-optimal"

but here `2.0.20211107` refers to a tag.

maybe others can chime in ?

JohnRTitor:
Upstream dev does not seem to be active on GitHub since last year, can still @Sweets them though.
Squashing all recent commits into a patch since the last release seems kind of weird to me. These are upstream changes after all.

pkgs/applications/misc/tiramisu/default.nix
wegank: As the author of #234201, @7c6f434c, may I ask you for help?

7c6f434c: Technically, upstream has 2.0 and _two_ 2.0.X releases based on datestamps. So if we want both to have an unstable word-mark and proper version sorting in case of a plausible 2.0.2024MMDD upstream tag, we need to do this not so nicely looking version with two dates. This is not a question of wording in the documentation, this is a logical implication.

But then the documentation says «should» not «must» for a good reason. So something in the direction of 2.0.20240101-unstable with a comment in the expression would not be completely unreasonable, if people involved with the package agree.

JohnRTitor: I see. So will changing the version to 2.0.20240408-unstable (today's date) suffice?

kirillrdy: @JohnRTitor I think we are arriving at `2.0.20240408-unstable`, except I would probably use date of commit not today's date.

@7c6f434c re wording in documentation. it also says `The date must be in "YYYY-MM-DD" format."`

so maybe an other option is `2.0-unstable-YYYY-MM-DD`. whichever option we choose we will need to leave a comment

JohnRTitor: Something along the lines of `2.0.20240408-unstable` would be more consistent with the upstream versioning scheme.
2.0-unstable-YYYY-MM-DD is also a viable option that is consistent with Nix versioning scheme.

7c6f434c: I would say that 2.0.20240408-unstable is breaking a «should» and representing it as «_faking_ the upstream tag in _their_ date format» not «creating a downstream tag based on upstream tag and date». 

2.0-unstable-X has the ordering drawbacks, although it might be irrelevant for the specific package.

JohnRTitor: Changed to `2.0.20240101-unstable`.

JohnRTitor: Can this be resolved now?

wegank: No, certainly not `20240101`.

Both `2.0-unstable-2023-03-29` and `2.0.20211107-unstable-2023-03-29` should be OK, choose the one you prefer.

JohnRTitor: Apologies for misunderstanding. I decided to go with `2.0-unstable-2023-03-29`.

pkgs/applications/misc/tiramisu/default.nix
```diff
@@ -2,13 +2,13 @@
 
 stdenv.mkDerivation rec {
   pname = "tiramisu";
-  version = "2.0.20211107";
+  version = "2.0-unstable-2023-03-29";
```
kirillrdy, line 5: can you leave a comment why version is not `2.0.20211107-unstable-2023-03-29`

JohnRTitor: Sure. I amended the commit. Because it would be confusing to have two dates in the version.

JohnRTitor: @kirillrdy Please check if it's fine.

kirillrdy (approved):

kirillrdy:
@JohnRTitor let's give some time for others to have a look

wegank (approved):

Sweets:
> Upstream dev does not seem to be active on GitHub since last year, can still @Sweets them though. Squashing all recent commits into a patch since the last release seems kind of weird to me. These are upstream changes after all. Then again, if we did that, we'd be essentially backporting those changes to last release.

Most recent commits have all been collapsed into a formal release.

Issue #141239, opened 2021-10-11
googleearth-pro fails to start due to proj linking error


squalus:
### Describe the bug
`googleearth-pro` has a runtime linking error that prevents it from starting.

`libgisingest.so: undefined symbol: pj_set_searchpath`

NixOS/nixpkgs `master@3f405ae`
### Steps To Reproduce
`env NIXPKGS_ALLOW_INSECURE=1 nix-build -A googleearth-pro`

```$ /nix/store/a45c4w83y4a6lp794jf5qswfz8078nng-googleearth-pro-7.3.3.7786/bin/googleearth-pro
/nix/store/a45c4w83y4a6lp794jf5qswfz8078nng-googleearth-pro-7.3.3.7786/bin/.googleearth-pro-wrapped_: symbol lookup error: /nix/store/a45c4w83y4a6lp794jf5qswfz8078nng-googleearth-pro-7.3.3.7786/opt/google/earth/pro/libgisingest.so: undefined symbol: pj_set_searchpath
```

### Additional context
Linking `gdal_2` to `proj_7` instead of `proj` seems to fix the problem and allows the program to start. After it starts, it then just shows a blank black screen where the 3d view should be. Updating to the latest googleearth-pro release didn't fix anything. 

### Notify maintainers
@friedelino @SCOTT-HAMILTON 

### Metadata
Please run `nix-shell -p nix-info --run "nix-info -m"` and paste the result.

```console
 - system: `"x86_64-linux"`
 - host os: `Linux 5.4.151, NixOS, 21.05.3740.ce7a1190a0f (Okapi)`
 - multi-user?: `yes`
 - sandbox: `yes`
 - version: `nix-env (Nix) 2.3.15`
 - channels(root): `"nixos-21.05.3740.ce7a1190a0f, nixpkgs-unstable-21.11pre320922.ee084c02040"`
 - nixpkgs: `/nix/var/nix/profiles/per-user/root/channels/nixos`

```

Maintainer information:
```yaml
# a list of nixpkgs attributes affected by the problem
attribute: googleearth-pro
# a list of nixos modules affected by the problem
module:
```

squalus:
The current package goes to some effort to use nixpkgs-provided dependencies instead of the ones bundled in the .deb. Using the bundled libraries instead makes it run correctly for me: https://github.com/squalus/nixpkgs/commit/5ff814823da9ec48ef6c5390877ba33f43618695

[...]

PR #302468, opened 2024-04-08
googleearth-pro: more specific vulnerability description

gnubufferoverflows:
The current description of the vulnerabilities in Google Earth Pro is rather vague, so I have added a more detailed description to help someone evaluate if it is safe to use or not. This is according to:

https://github.com/NixOS/nixpkgs/issues/141239#issuecomment-942911083

### DIFF
--- a/pkgs/applications/misc/googleearth-pro/default.nix
+++ b/pkgs/applications/misc/googleearth-pro/default.nix
@@ -120,6 +120,6 @@ mkDerivation rec {
     license = licenses.unfree;
     maintainers = with maintainers; [ friedelino shamilton ];
     platforms = platforms.linux;
-    knownVulnerabilities = [ "Includes vulnerable bundled libraries." ];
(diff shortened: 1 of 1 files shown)


pkgs/applications/misc/googleearth-pro/default.nix
```diff
@@ -120,6 +120,6 @@ mkDerivation rec {
     license = licenses.unfree;
     maintainers = with maintainers; [ friedelino shamilton ];
     platforms = platforms.linux;
-    knownVulnerabilities = [ "Includes vulnerable bundled libraries." ];
+    knownVulnerabilities = [ "Includes old and vulnerable versions of openssl, ffmpeg, gdal, and proj, which are included with Google Earth Pro." ];
```
Aleksanaa, line 123:
```suggestion
    knownVulnerabilities = [ "Includes old and vulnerable bundled libraries: openssl, ffmpeg, gdal, and proj." ];
```

Aleksanaa (approved):

PR #302645, opened 2024-04-08
nixos/samba: only create /etc/samba/smb.conf when samba is enabled

anthonyroussel:
## Description of changes

Introduced with commit 7435db4f89823

`/etc/samba/smb.conf` file is installed in the system environment even when `services.samba.enable` is disabled

This file should only be installed when Samba is enabled.
This PR is a requirement to start the work on Samba AD DC NixOS module.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
- For non-Linux: Is sandboxing enabled in `nix.conf`? (See [Nix manual](https://nixos.org/manual/nix/stable/command-ref/conf-file.html))
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [x] Tested, as applicable:
  - [NixOS test(s)](https://nixos.org/manual/nixos/unstable/index.html#sec-nixos-tests) (look inside [nixos/tests](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests))
  - and/or [package tests](https://github.com/NixOS/nixpkgs/blob/master/pkgs/README.md#package-tests)
  - or, for functions and "core" functionality, tests in [lib/tests](https://github.com/NixOS/nixpkgs/blob/master/lib/tests) or [pkgs/test](https://github.com/NixOS/nixpkgs/blob/master/pkgs/test)
  - made sure NixOS tests are [linked](https://nixos.org/manual/nixpkgs/unstable/#ssec-nixos-tests-linking) to the relevant packages
- [ ] Tested compilation of all packages that depend on this change using `nix-shell -p nixpkgs-review --run "nixpkgs-review rev HEAD"`. Note: all changes have to be committed, also see [nixpkgs-review usage](https://github.com/Mic92/nixpkgs-review#usage)
- [x] Tested basic functionality of all binary files (usually in `./result/bin/`)
- [24.05 Release Notes](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2405.section.md) (or backporting [23.05](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2305.section.md) and [23.11](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2311.section.md) Release notes)
  - [ ] (Package updates) Added a release notes entry if the change is major or breaking
[...]

PR #279917, opened 2024-01-09
nixos/samba: add system-samba.slice

h7x4:
## Description of changes

Part of #279915

Ran `nix build .#nixosTests.samba`

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
- For non-Linux: Is sandboxing enabled in `nix.conf`? (See [Nix manual](https://nixos.org/manual/nix/stable/command-ref/conf-file.html))
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [x] Tested, as applicable:
  - [NixOS test(s)](https://nixos.org/manual/nixos/unstable/index.html#sec-nixos-tests) (look inside [nixos/tests](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests))
  - and/or [package tests](https://nixos.org/manual/nixpkgs/unstable/#sec-package-tests)
  - or, for functions and "core" functionality, tests in [lib/tests](https://github.com/NixOS/nixpkgs/blob/master/lib/tests) or [pkgs/test](https://github.com/NixOS/nixpkgs/blob/master/pkgs/test)
  - made sure NixOS tests are [linked](https://nixos.org/manual/nixpkgs/unstable/#ssec-nixos-tests-linking) to the relevant packages
- [ ] Tested compilation of all packages that depend on this change using `nix-shell -p nixpkgs-review --run "nixpkgs-review rev HEAD"`. Note: all changes have to be committed, also see [nixpkgs-review usage](https://github.com/Mic92/nixpkgs-review#usage)
- [ ] Tested basic functionality of all binary files (usually in `./result/bin/`)
- [24.05 Release Notes](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2405.section.md) (or backporting [23.05](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2305.section.md) and [23.11](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2311.section.md) Release notes)
  - [ ] (Package updates) Added a release notes entry if the change is major or breaking
  - [ ] (Module updates) Added a release notes entry if the change is significant
  - [ ] (Module addition) Added a release notes entry if adding a new NixOS module
[...]

PR #302681, opened 2024-04-08
nixos/samba: migrate to structural settings (RFC42)

anthonyroussel:
## Description of changes

See [RFC42](https://github.com/NixOS/rfcs/pull/42)

* Migrated services.samba.extraConfig to services.samba.settings
* Renamed services.samba.enableNmbd to nmbd.enable
* Renamed services.samba.enableWinbind to winbindd.enable
* Added services.samba.smbd.enable
  * This option will disable the smbd daemon
  * Required to continue the work on samba-ad-dc SystemD service
* Split SystemD configuration and sync it with upstream

Also see https://github.com/NixOS/nixpkgs/pull/302645

TODO / Work in progress:

- [x] Improve new module documentation
- [x] Fix SystemD services dependencies

Also planned later:

- [ ] Lint module with nixfmt-rfc-style (in another PR)
- [ ] Remove `with lib;` (in another PR)
- [ ] Migrate enable options to mkEnableOption, and maybe disable them by default (in another PR)
- [ ] Add samba-ad-dc support and ensureDomain provisioner (in another PR)

This PR is a requirement to start the work on Samba AD DC NixOS module.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---


anthonyroussel:
@ofborg test samba

nixos/modules/services/network-filesystems/samba.md
```diff
@@ -0,0 +1,35 @@
...
+
+```nix
+{
+  services.samba = {
+    enable = true;
+    openFirewall = true;
+    settings = {
+      public = {
+        path = "/public";
+        "read only" = true;
```
anthonyroussel, line 27:
```suggestion
        "read only" = "yes";
```

nixos/tests/samba.nix
```diff
@@ -1,46 +1,49 @@
...
               "guest ok" = "yes";
               comment = "Public samba share.";
             };
+          };
         };
-    };
+      };
+  };
 
   # client# [    4.542997] mount[777]: sh: systemd-ask-password: command not found
```
anthonyroussel, line 38: Maybe we can remove this line now

nixos/modules/services/network-filesystems/samba.nix
```diff
@@ -241,6 +176,100 @@ in
...
+          ];
+
+          partOf = [ "samba.target" ];
+          wantedBy = [ "samba.target" ];
+
+          environment.LD_LIBRARY_PATH = config.system.nssModules.path;
+
+          serviceConfig = {
+            ExecReload = "${pkgs.coreutils}/bin/kill -HUP $MAINPID";
+            ExecStart = "${cfg.package}/sbin/winbindd --foreground --no-process-group ${escapeShellArgs cfg.winbindd.extraArgs}";
```
anthonyroussel, line 263:
```suggestion
            ExecStart = "${cfg.package}/sbin/winbindd --foreground --no-process-group ${lib.escapeShellArgs cfg.winbindd.extraArgs}";
```

anthonyroussel:
Fixed merge conflict

anthonyroussel:
Rebased against master branch and fixed merge conflict

nixos/doc/manual/release-notes/rl-2411.section.md
```diff
@@ -42,6 +42,8 @@
   for `stateVersion` ≥ 24.11. (It was previously using SQLite for structured
   data and the filesystem for blobs).
 
+- `samba`' `services.samba.extraConfig` setting has been removed and converted to the freeform type option named [`services.samba.settings`](#opt-services.samba.settings) according to RFC42.
```
SuperSandro2000, line 45:
```suggestion
- The `services.samba.extraConfig` setting has been removed and converted to the freeform type option named [`services.samba.settings`](#opt-services.samba.settings) according to RFC42.
```

nixos/modules/services/network-filesystems/samba.nix
```diff
@@ -3,35 +3,12 @@
...
-      [global]
-      security = ${cfg.securityType}
-      passwd program = /run/wrappers/bin/passwd %u
-      invalid users = ${smbToString cfg.invalidUsers}
-
-      ${cfg.extraConfig}
-
-      ${smbToString (map shareConfig (attrNames cfg.shares))}
-    '');
+  samba = cfg.package;
```
SuperSandro2000: While we are at it, please drop this very generic variable and replace all occurences with cfg.package

nixos/modules/services/network-filesystems/samba.nix
```diff
@@ -67,6 +44,19 @@ in
...
+      Use services.samba.settings instead.
+
+      This is part of the general move to use structured settings instead of raw
+      text for config as introduced by RFC0042:
+      https://github.com/NixOS/rfcs/blob/master/rfcs/0042-config-option.md
+    '')
+    (lib.mkRemovedOptionModule [ "services" "samba" "extraConfig" ] "Use services.samba.settings instead.")
+    (lib.mkRemovedOptionModule [ "services" "samba" "invalidUsers" ] ''Use services.samba.settings.global."invalid users" instead.'')
+    (lib.mkRemovedOptionModule [ "services" "samba" "securityType" ] ''Use services.samba.settings.global."security type" instead.'')
+    (lib.mkRemovedOptionModule [ "services" "samba" "shares" ] "Use services.samba.settings instead.")
```
SuperSandro2000, line 33: Is the format of those similar? if so we could use mkRenamedOption

nixos/modules/services/network-filesystems/samba.nix
```diff
@@ -169,25 +124,31 @@ in
...
+            security = "user";
+            "passwd program" = "/run/wrappers/bin/passwd %u";
+            "invalid users" = "root";
+          };
+          public = {
+            path = "/srv/public";
+            "read only" = "yes";
+            browseable = "yes";
+            "guest ok" = "yes";
+            comment = "Public samba share.";
```
SuperSandro2000, line 125: Lets quote everything to be more consistent

nixos/doc/manual/release-notes/rl-2411.section.md
```diff
@@ -44,6 +44,10 @@
 
 - `samba`' `services.samba.extraConfig` setting has been removed and converted to the freeform type option named [`services.samba.settings`](#opt-services.samba.settings) according to RFC42.
 
+- `samba`' The following options of the Samba module were renamed:
+  - `enableWinbindd` -> [`winbindd.enable`](#opt-services.samba.winbindd.enable),
+  - `enableNmbd` -> [`nmbd.enable`](#opt-services.samba.nmbd.enable).
+
```
SuperSandro2000, line 50:
```suggestion
```
That's not note worthy for the changelog. People will be informed by the module system already

nixos/modules/services/network-filesystems/samba.nix
```diff
@@ -74,38 +74,22 @@ in
...
       };
 
-      openFirewall = mkOption {
-        type = types.bool;
-        default = false;
-        description = ''
-          Whether to automatically open the necessary ports in the firewall.
-        '';
-      };
+      openFirewall = lib.mkEnableOption "automatically open the necessary ports in the firewall.";
```
SuperSandro2000, line 52: That would now be completed to ``Wether to enable automatically open the necessary ports in the firewall..``.

nixos/modules/services/network-filesystems/samba.nix
```diff
@@ -124,19 +108,12 @@ in
...
-        default = false;
-        type = types.bool;
-        description = ''
-          Whether to enable the WINS NSS (Name Service Switch) plug-in.
-          Enabling it allows applications to resolve WINS/NetBIOS names (a.k.a.
-          Windows machine names) by transparently querying the winbindd daemon.
-        '';
-      };
+        Enabling it allows applications to resolve WINS/NetBIOS names (a.k.a.
+        Windows machine names) by transparently querying the winbindd daemon.
```
SuperSandro2000, line 108:
```suggestion
        Windows machine names) by transparently querying the winbindd daemon
```

nixos/modules/services/network-filesystems/samba.nix
```diff
@@ -191,6 +176,100 @@ in
...
+          wantedBy = [ "samba.target" ];
+          wants = [ "network-online.target" ];
+
+          environment.LD_LIBRARY_PATH = config.system.nssModules.path;
+
+          serviceConfig = {
+            ExecReload = "${pkgs.coreutils}/bin/kill -HUP $MAINPID";
+            ExecStart = "${cfg.package}/sbin/nmbd --foreground --no-process-group ${lib.escapeShellArgs cfg.nmbd.extraArgs}";
+            LimitCORE = "infinity";
+            PIDFile = "/run/nmbd.pid";
```
SuperSandro2000, line 200:
```suggestion
            PIDFile = "/run/samba/nmbd.pid";
```
We should also try to add some hardening

nixos/modules/services/network-filesystems/samba.nix
```diff
@@ -191,6 +176,100 @@ in
...
+          wants = [ "network-online.target" ];
+
+          environment.LD_LIBRARY_PATH = config.system.nssModules.path;
+
+          serviceConfig = {
+            ExecReload = "${pkgs.coreutils}/bin/kill -HUP $MAINPID";
+            ExecStart = "${cfg.package}/sbin/smbd --foreground --no-process-group ${lib.escapeShellArgs cfg.smbd.extraArgs}";
+            LimitCORE = "infinity";
+            LimitNOFILE = 16384;
+            PIDFile = "/run/smbd.pid";
```
SuperSandro2000, line 235: same here

nixos/modules/services/network-filesystems/samba.nix
```diff
@@ -191,6 +176,100 @@ in
...
+          partOf = [ "samba.target" ];
+          wantedBy = [ "samba.target" ];
+
+          environment.LD_LIBRARY_PATH = config.system.nssModules.path;
+
+          serviceConfig = {
+            ExecReload = "${pkgs.coreutils}/bin/kill -HUP $MAINPID";
+            ExecStart = "${cfg.package}/sbin/winbindd --foreground --no-process-group ${lib.escapeShellArgs cfg.winbindd.extraArgs}";
+            LimitCORE = "infinity";
+            PIDFile = "/run/winbindd.pid";
```
SuperSandro2000, line 265: and here

nixos/modules/services/network-filesystems/samba.nix
```diff
@@ -191,6 +176,100 @@ in
         networking.firewall.allowedTCPPorts = mkIf cfg.openFirewall [ 139 445 ];
         networking.firewall.allowedUDPPorts = mkIf cfg.openFirewall [ 137 138 ];
       })
-    ];
 
+      (lib.mkIf cfg.nmbd.enable {
```
SuperSandro2000, line 180: I think now got a lot of duplication. Maybe we could work with a common attrset instead?

SuperSandro2000:
Looks already pretty solid

nixos/modules/services/network-filesystems/samba.nix
anthonyroussel: From my point of view, I'm not sure it's worth it.

There was a common attr set before, however the configuration values of the 3 services is quite different between each.

That's why I split the configuration of each service in this PR.

anthonyroussel:
@SuperSandro2000 Hello! Thank you very much for your review on this PR :)
I've handled all your suggestions.

anthonyroussel:
Merge conflict coming from https://github.com/NixOS/nixpkgs/pull/279917

nixos/doc/manual/release-notes/rl-2411.section.md
```diff
@@ -187,6 +187,8 @@
 
 - The `replay-sorcery` package and module was removed as it unmaintained upstream. Consider using `gpu-screen-recorder` or `obs-studio` instead.
 
+- The `services.samba.extraConfig` setting has been removed and converted to the freeform type option named [`services.samba.settings`](#opt-services.samba.settings) according to RFC42.
```
bachp, line 190: Please also mention the removal of `configText`.

And for completeness it might also be worth to mentioned the renamed options here.

nixos/modules/services/network-filesystems/samba.nix
```diff
@@ -3,232 +3,162 @@
...
+      Use services.samba.settings instead.
+
+      This is part of the general move to use structured settings instead of raw
+      text for config as introduced by RFC0042:
+      https://github.com/NixOS/rfcs/blob/master/rfcs/0042-config-option.md
+    '')
+    (lib.mkRemovedOptionModule [ "services" "samba" "extraConfig" ] "Use services.samba.settings instead.")
+    (lib.mkRenamedOptionModule [ "services" "samba" "invalidUsers" ] [ "services" "samba" "settings" "global" "invalid users" ])
+    (lib.mkRenamedOptionModule [ "services" "samba" "securityType" ] [ "services" "samba" "settings" "global" "security type" ])
+    (lib.mkRemovedOptionModule [ "services" "samba" "shares" ] "Use services.samba.settings instead.")
```
bachp, line 33:
```suggestion
    (lib.mkRenamedOptionModule [ "services" "samba" "shares" ] [ "services" "samba" "settings" ])
```

I think we can just rename shares to settings. As far as I see the format didn't change. This makes one less breaking change. This way only users that used `extraConfig` or `configText` are affected.

bachp (changes requested):
I tested this in a migrations scenario and it workes.

I added some suggestions to make it even less invasive.

nixos/doc/manual/release-notes/rl-2411.section.md
anthonyroussel: I modified rl-2411.section.md to mention that configText has been removed in favor of services.samba.settings freeform type option.

And also added the renamed options here.

nixos/modules/services/network-filesystems/samba.nix
anthonyroussel: Thank you so much for the review @bachp! 

You're right! I changed that line to mkRenamedOptionModule.

anthonyroussel:
Hello @trofi, I noticed that you've recently modified the Samba package.

So I thought you might be interested in reviewing this pull request to migrate the NixOS Samba module to RFC42 :)

bachp (approved):

nixos/modules/services/network-filesystems/samba.nix
```diff
@@ -252,6 +182,103 @@ in
...
+            Type = "notify";
+          };
+
+          unitConfig.RequiresMountsFor = "/var/lib/samba";
+
+          restartTriggers = [ configFile ];
+        };
+      })
+
+      (lib.mkIf cfg.smbd.enable {
```
skeuchel, line 217: It seems the options here are now always enabled. `cfg.smbd.enable` defaults to `true`. Shouldn't this include `cfg.enable` (which defaults to false)? E.g. `mkIf (cfg.enable && cfg.smbd.enable)`? @anthonyroussel

bachp: @skeuchel I was under the impression that the whole block will not be processed if `cfg.enable` on line 186 is false. Did you see smbd being enabled now without enabling the samba module?

anthonyroussel: Thank you for the review, I hadn't thought about it.
I will fix this in another PR.

skeuchel: > @skeuchel I was under the impression that the whole block will not be processed if `cfg.enable` on line 186 is false. Did you see smbd being enabled now without enabling the samba module?

Yes, a nightly cross build started failing because it was building `samba` all of a sudden, which apparently fails to cross-compile.

anthonyroussel: https://github.com/NixOS/nixpkgs/pull/340272

nixos/modules/services/network-filesystems/samba.nix
```diff
@@ -3,35 +3,12 @@
...
+  settingsFormat = pkgs.formats.ini { };
+  configFile = settingsFormat.generate "smb.conf" cfg.settings;
 
-  configFile = pkgs.writeText "smb.conf"
-    (if cfg.configText != null then cfg.configText else
-    ''
-      [global]
-      security = ${cfg.securityType}
-      passwd program = /run/wrappers/bin/passwd %u
-      invalid users = ${smbToString cfg.invalidUsers}
```
yshui, line 29: if the user follows the instruction to use `services.samba.settings`, these 2 options are lost without an instruction on how to migrate them.

bachp: @yshui I think `invalid users` and `securtiy` are renamed. The only option thet is indeed missing is the `passwd program`.

yshui: @bachp right, but `invalidUsers` had a default value, so if the user didn't config anything they would still get a `invalid users = root` by default. after this migration that's gone, and the user won't get any warnings about that.

anthonyroussel: @yshui Thank you for your feedback! :)
Yes you're right, I shouldn't break the default values like I did here.

What do you think if I put them back in the submodule options?

```nix
       settings = lib.mkOption {
-        type = lib.types.submodule { freeformType = settingsFormat.type; };
-        default = {};
+        type = lib.types.submodule {
+          freeformType = settingsFormat.type;
+          options = {
+            "global"."security" = lib.mkOption {
+              type = lib.types.enum [ "auto" "user" "domain" "ads" ];
+              default = "user";
+              description = "Samba security type.";
+            };
+            "global"."invalid users" = lib.mkOption {
+              type = lib.types.str;
+              default = "root";
+              description = "List of users who are denied to login via Samba.";
+            };
+            "global"."passwd program" = lib.mkOption {
+              type = lib.types.str;
+              default = "/run/wrappers/bin/passwd %u";
+              description = "FIXME";
+            };
+          };
+        };
+        default = {
+          "global" = {
+            "security" = "user";
+            "passwd program" = "/run/wrappers/bin/passwd %u";
+            "invalid users" = "root";
+          };
+        };
         example = {
           "global" = {
             "security" = "user";
```

anthonyroussel: https://github.com/NixOS/nixpkgs/pull/340500

misuzu:
Looks like the `global` section must be always on top and after this change it can end up after shares declarations which breaks everything...

anthonyroussel:
@misuzu Hello, thank you for pointing this out.

The [smb.conf documentation](https://www.samba.org/samba/docs/current/man-html/smb.conf.5.html) does not mention anything about the section ordering, especially regarding the special global section.

I was able to reproduce the issue you mentioned with the following configuration:

```nix
  services.samba = {
    enable = true;
    openFirewall = true;
    settings = {
      global = {
        "guest ok" = true;
      };
      "public" = {
        "path" = "/public";
        "browseable" = "yes";
        "comment" = "Public samba share.";
      };
      "Shared" = {
        "path" = "/shared";
        "browseable" = "yes";
        "comment" = "Shared samba share.";
      };
    };
  };
};
```

In this configuration, the public share will allow guest access, while the Shared share won't.

I see a few possible ways to address this issue:

* Split the settings into `globalSettings` and `shareSettings`, ensuring that `globalSettings` appears first in the configuration file and `shareSettings` after.
* Use a customized version of `pkgs.formats.iniWithGlobalSection` to automatically place the `global section` (and its `[global]` header) at the top of the configuration file.
* Revert the PR and restore the previous behavior.

Does anyone have any other ideas?

bachp:
@anthonyroussel 

There is no need to split the section from a user perspective. We can also do this via somthing along the lines of:

```
  globalSettings = { global = cfg.settings.global; };
  shareSettings = builtins.removeAttrs cfg.settings [ "global" ];
```

However I'm not sure how to tell `formats.ini` to put this two attrsets in the right order.

bachp:
@misuzu @anthonyroussel 

I think I found a solution in https://github.com/NixOS/nixpkgs/pull/341257

I just ran the samba module test but did not try yet with the above reproducer if it really fixes the issue.

nixos/modules/services/network-filesystems/samba.nix
```diff
@@ -3,232 +3,162 @@
...
-        PIDFile = "/run/${appName}.pid";
-        Type = "notify";
-        NotifyAccess = "all"; #may not do anything...
-        Slice = "system-samba.slice";
-      };
-      unitConfig.RequiresMountsFor = "/var/lib/samba";
-
-      restartTriggers = [ configFile ];
-    };
+  settingsFormat = pkgs.formats.ini { };
```
endocrimes, line 8: Samba settings aren't strict INI so this breaks things - specifically configuring extensions like `fruit:appl = yes` for enabling the Darwin compatibility layer.

anthonyroussel: Hello! Thank you for the feedback.

Could you provide an example where this configuration isn’t working as expected, please?

To configure `fruit:appl = yes` for `vfs_fruit`, you can use this syntax:

```nix
services.samba.settings.global."fruit:appl" = "yes";
```

brookst:
Hi. I've update the NixOS Wiki article: https://nixos.wiki/index.php?title=Samba&diff=13441&oldid=13357
I think that updates the snippets to be compatible with these changes. Perhaps someone with more familiarity ought to look over that article to see if any info has gone out-of-date.

Cheers!

bachp:
Thank you @brookst.

There is also https://wiki.nixos.org/wiki/Samba which is linked from the official nixos.org page. Maybe it makes sense to update there too.

PR #302688, opened 2024-04-08
openssh: move Kerberos support into a dedicated package

LeSuisse:
## Description of changes

The `openssh` and `openssh_hpn` packages are now built without the Kerberos support by default in an effort to reduce the attack surface.

The Kerberos support is likely used only by a fraction of the total users (I'm guessing mainly users integrating SSH in an Active Directory env) so dropping it should not impact too many users. It should also be noted that the Kerberos/GSSAPI auth is disabled by default in the configuration. `opensshWithKerberos` and `openssh_hpnWithKerberos` are added in order to provide an easy migration path for users needing this support.

The `openssh_gssapi` package is kept untouched.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [x] Tested, as applicable:

---

### DIFF
--- a/nixos/modules/services/networking/ssh/sshd.nix
+++ b/nixos/modules/services/networking/ssh/sshd.nix
@@ -708,6 +708,10 @@ in
 
     assertions = [{ assertion = if cfg.settings.X11Forwarding then cfgc.setXAuthLocation else true;
                     message = "cannot enable X11 forwarding without setting xauth location";}
+                  { assertion = (builtins.match "(.*\n)?(\t )*[Kk][Ee][Rr][Bb][Ee][Rr][Oo][Ss][Aa][Uu][Tt][Hh][Ee][Nn][Tt][Ii][Cc][Aa][Tt][Ii][Oo][Nn][ |\t|=|\"]+yes.*" "${configFile}\n${cfg.extraConfig}") != null -> cfgc.package.withKerberos;
+                    message = "cannot enable Kerberos authentication without using a package with Kerberos support";}
+                  { assertion = (builtins.match "(.*\n)?(\t )*[Gg][Ss][Ss][Aa][Pp][Ii][Aa][Uu][Tt][Hh][Ee][Nn][Tt][Ii][Cc][Aa][Tt][Ii][Oo][Nn][ |\t|=|\"]+yes.*" "${configFile}\n${cfg.extraConfig}") != null -> cfgc.package.withKerberos;
+                    message = "cannot enable GSSAPI authentication without using a package with Kerberos support";}
                   (let
                     duplicates =
                       # Filter out the groups with more than 1 element
--- a/pkgs/tools/networking/openssh/common.nix
+++ b/pkgs/tools/networking/openssh/common.nix
@@ -22,7 +22,7 @@
 , pam
 , libredirect
 , etcDir ? null
-, withKerberos ? true
+, withKerberos ? false
 , withLdns ? true
 , libkrb5
 , libfido2
@@ -177,9 +177,12 @@ stdenv.mkDerivation {
     "sysconfdir=\${out}/etc/ssh"
   ];
 
-  passthru.tests = {
-    borgbackup-integration = nixosTests.borgbackup;
-    openssh = nixosTests.openssh;
+  passthru = {
+    inherit withKerberos;
+    tests = {
+      borgbackup-integration = nixosTests.borgbackup;
+      openssh = nixosTests.openssh;
+    };
   };
 
   meta = with lib; {
--- a/pkgs/top-level/all-packages.nix
+++ b/pkgs/top-level/all-packages.nix
@@ -11358,12 +11358,21 @@ with pkgs;
     etcDir = "/etc/ssh";
   };
 
+  opensshWithKerberos = openssh.override {
+    withKerberos = true;
+  };
+
   openssh_hpn = opensshPackages.openssh_hpn.override {
     etcDir = "/etc/ssh";
   };
 
+  openssh_hpnWithKerberos = openssh_hpn.override {
+    withKerberos = true;
+  };
+
   openssh_gssapi = opensshPackages.openssh_gssapi.override {
     etcDir = "/etc/ssh";
+    withKerberos = true;
   };
 
   ssh-copy-id = callPackage ../tools/networking/openssh/copyid.nix { };
--- a/nixos/doc/manual/release-notes/rl-2411.section.md
+++ b/nixos/doc/manual/release-notes/rl-2411.section.md
@@ -14,6 +14,9 @@
 
 - `nginx` package no longer includes `gd` and `geoip` dependencies. For enabling it, override `nginx` package with the optionals `withImageFilter` and `withGeoIP`.
 
+- `openssh` and `openssh_hpn` are now compiled without Kerberos 5 / GSSAPI support in an effort to reduce the attack surface of the components for the majority of users. Users needing this support can
+  use the new `opensshWithKerberos` and `openssh_hpnWithKerberos` flavors (e.g. `programs.ssh.package = pkgs.openssh_gssapi`).
+
 - `nvimpager` was updated to version 0.13.0, which changes the order of user and
   nvimpager settings: user commands in `-c` and `--cmd` now override the
   respective default settings because they are executed later.


tomfitzhenry (changes requested):
Glad to see this! I have made a similar change for my own machines, and glad to see a reduced out-the-box attack surface for all!

I left a comment re this being a breaking change (that could break remote access), and a few smaller suggestions/questions.

Also, any thoughts on whether to disable `withPAM` by default? That's something that users would have to opt into anyway, right?

nixos/modules/services/networking/ssh/sshd.nix
```diff
@@ -385,6 +385,20 @@ in
                 Specifies whether keyboard-interactive authentication is allowed.
               '';
             };
+            KerberosAuthentication = mkOption {
```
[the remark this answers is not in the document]
LeSuisse: I mainly added them to be able to use them in the assertions. Changed that in the latest version of the patch to directly check the final config file (and I do not want to break existing config that might have set something lik e `KerberosAuthentication = "yes"`).

pkgs/top-level/all-packages.nix
```diff
@@ -11519,12 +11519,21 @@ with pkgs;
     etcDir = "/etc/ssh";
   };
 
+  opensshWithKerberos = openssh.override {
```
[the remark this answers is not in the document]
LeSuisse: Maybe but that's basically the case for all existing variants so the situation is not really worse with this contribution. Having a meaningful integration test could be interesting but it should probably be written and maintained by someone having an actual Kerberos/GSSAPI setup.

pkgs/tools/networking/openssh/common.nix
```diff
@@ -22,7 +22,7 @@
 , pam
 , libredirect
 , etcDir ? null
-, withKerberos ? true
+, withKerberos ? false
```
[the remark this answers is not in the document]
LeSuisse: Adding a warning that will force all users to do switch the package is more disruptive and users needing the Kerberos support they will continue to have the warning.

I adjusted the assertions to check the generated config so now we should also catch situations where the Kerberos/GSSAPI auth have been enabled in `extraConfig`. It means that building the configuration will throw an error if you also did not provide a package with the Kerberos support enabled. That should prevent users to be be locked out.

I'm not going to touch PAM at this stage, it likely far more usages than Kerberos and is also enabled by default in the current config:
https://github.com/NixOS/nixpkgs/blob/1042fd8b148a9105f3c0aca3a6177fd1d9360ba5/nixos/modules/services/networking/ssh/sshd.nix#L641

LeSuisse:
Thanks for the review!

tomfitzhenry (approved):

tomfitzhenry (changes requested):
Sorry, I just noticed if `services.openssh.extraConfig = "KerberosAuthentication = yes";` is added to the machine config in nixosTests.openssh, then the test still passes. I had expected that would fail to build, due to the assertion.

Any ideas what's going on there?

LeSuisse:
Thanks, fixed.

I forgot about `=` as a separator between the keyword and argument.

nixos/modules/services/networking/ssh/sshd.nix
```diff
@@ -690,6 +690,10 @@ in
 
     assertions = [{ assertion = if cfg.settings.X11Forwarding then cfgc.setXAuthLocation else true;
                     message = "cannot enable X11 forwarding without setting xauth location";}
+                  { assertion = (builtins.match "(.*\n)?(\t )*[Kk][Ee][Rr][Bb][Ee][Rr][Oo][Ss][Aa][Uu][Tt][Hh][Ee][Nn][Tt][Ii][Cc][Aa][Tt][Ii][Oo][Nn][ |\t|=|\"]+yes.*" "${configFile}\n${cfg.extraConfig}") != null -> cfgc.package.withKerberos;
```
SuperSandro2000, line 693: Don't we have structured settings here? Those regexes are kinda ugly.

pkgs/tools/networking/openssh/common.nix
```diff
@@ -168,8 +168,11 @@ stdenv.mkDerivation {
     "sysconfdir=\${out}/etc/ssh"
   ];
 
-  passthru.tests = {
-    borgbackup-integration = nixosTests.borgbackup;
+  passthru = {
```
SuperSandro2000, line 171: We should definitely add a test with kerberos to passthru

pkgs/top-level/all-packages.nix
```diff
@@ -11519,12 +11519,21 @@ with pkgs;
...
+    withKerberos = true;
+  };
+
   openssh_hpn = opensshPackages.openssh_hpn.override {
     etcDir = "/etc/ssh";
   };
 
+  openssh_hpnWithKerberos= openssh_hpn.override {
```
SuperSandro2000, line 11530: Missing space before =

nixos/modules/services/networking/ssh/sshd.nix
LeSuisse: No we do not because we also have to cover `extraConfig`.

pkgs/tools/networking/openssh/common.nix
LeSuisse: A NixOS test? We are not in a worse place than what we currently have on `master` and I am probably not the best person to write this test and to maintain it

https://github.com/NixOS/nixpkgs/pull/302688#discussion_r1563165964

LeSuisse:
No, my initial version was not correct. It appears that separating keyword from argument with a `=` is acceptable and works as you would expect it to be. It is mentioned in the sshd_config(5) man page of OpenBSD but not in the one of openssh-portable.

LeSuisse:
Putting this to draft for now, I will rebase and re-open after branch-off.

LeSuisse:
Rebased, we should be good.

tomfitzhenry (approved):

hmenke:
I find this extremely inconsiderate. Now I essentially have the choice:

1. Whenever I log into a system that has Kerberos login, make sure that I'm using an `ssh` binary from the `openssh_gssapi` package and explicitly write `ssh -o GSSAPIAuthentication=yes`.
2. Make an overlay that replaces `openssh` with `openssh_gssapi` system-wide and rebuild like a 1000 packages everytime I update.
3. Live with the constant spam of
    ```
    /home/henri/.ssh/config line 9: Unsupported option "gssapiauthentication"
    /home/henri/.ssh/config line 10: Unsupported option "gssapidelegatecredentials"
    ```
   that is coming now from `ssh`, `git`, `mosh`, `rsync`, and all the other tools that use SSH internally.

Also the statement "in an effort to reduce the attack surface" definitely needs citation. Is there *any* other Linux distro out there that compiles OpenSSH without Kerberos support?

Please revert this and backport to stable.

hmenke:
I did a quick survey as to which Linux distros enabled OpenSSH with Kerberos support by default. Those are **Debian** (and by extension its many derivatives like Ubuntu), **Fedora** (i.e. also on RHEL, CentOS, AlmaLinux, RockyLinux, and other derivatives), **Arch Linux** (which includes the currently extremely popular CachyOS and EndeavourOS), **OpenSuSE**. Case in point though, it is disabled by default Gentoo, because the global USE flag `kerberos` is not set by default.

While the statement "should not impact too many users" is probably true, it completely [breaks my workflow](https://xkcd.com/1172/) and I don't see myself mass-rebuilding packages all the time.

hmenke:
No, `programs.ssh.package = pkgs.opensshWithKerberos;` does not work for me, because this does not propagate to the hundreds of packages that have `openssh` as a dependency. The only way to do this is a mass-rebuild.

> Maintainer of OpenSSH: https://cybervillains.com/@djm/112192735215160932

But this is about supply chain attacks and not about the security of Kerberos in itself. Also, if that was actually a concern, then OpenSSH would vendor all their dependencies, which they don't do, so this is all just meaningless security theater driven by the momentary scare of the xz-utils incident.

hmenke:
I'm counting 112 packages that (probably) depend on `openssh` (on NixOS 24.11)
```
$ find pkgs/ -iname '*.nix' -type f -not -iwholename 'pkgs/top-level/*' | xargs grep -l openssh | wc -l
112
```
The most prominent example is probably Git and as you know there are like a thousand packages that depend on Git.

Making those packages depend on `programs.ssh.package` doesn't really solve the problem, no? This will still cause mass-rebuilds when `programs.ssh.package` deviates from the default. Not to mention that  `programs.ssh.package` is not available to derivations in `pkgs/`.

Wouldn't it have been a better compromise to introduce an `opensshWithoutKerberos` package and set `services.openssh.package` to that by default? Then the attack surface on the server is reduced, while the client is still able to login to other systems that do require these other authentication methods. And if I need Kerberos on the server I only have to change the package in a single place without forcing a mass-rebuild.

SuperSandro2000:
> that is coming now from ssh, git, mosh, rsync, and all the other tools that use SSH internally.

you probably want to overwrite the the openssh inputs of those packages instead of doing a global overwrite. Also many tools also use libssh.

hmenke:
> you probably want to overwrite the the openssh inputs of those packages instead of doing a global overwrite. Also many tools also use libssh.

Sure, then I will have `git' = git.override { openssh = opensshWithKerberos; }`, but next thing I'll have to do is `pass' = pass.override { git = git'; }` and so on and so forth. It'll just keep spiraling out of control this way with extra overrides everywhere.

As far as I know libssh has no overlap with openssh. When I searched nixpkgs, I explicitly searched for "openssh". Packages depending on libssh were not affected by this change.

SuperSandro2000:
> Sure, then I will have git' = git.override { openssh = opensshWithKerberos; }, but next thing I'll have to do is pass' = pass.override { git = git'; } and so on and so forth. It'll just keep spiraling out of control this way with extra overrides everywhere.

The problem is probably here that pass is forcing it's version of git onto PATH. It should suffix PATH instead of prefixing.

hmenke:
> The problem is probably here that pass is forcing it's version of git onto PATH. It should suffix PATH instead of prefixing.

Maybe, but this was just an example. There are surely many others. Is it really necessary to fix perhaps hundreds of derivations instead of just rolling back this (in my opinion completely unnecessary) change?

SuperSandro2000:
For git only the GUI needs it https://github.com/NixOS/nixpkgs/blob/master/pkgs/applications/version-management/git/default.nix#L84

IMO, just take that from PATH, too. Also reduces rebuild counts.

SuperSandro2000:
You can just set GIT_SSH to overwrite it without triggering a rebuild.

hmenke:
And what do I do with all the other programs that wrap OpenSSH, that don't have an environment variable to override it?
```
$ git grep -l '${openssh}/bin' -- pkgs/
pkgs/applications/networking/remote/x2goclient/default.nix
pkgs/applications/version-management/git/default.nix
pkgs/by-name/as/assh/package.nix
pkgs/by-name/bo/borgbackup/package.nix
pkgs/by-name/di/diskrsync/package.nix
pkgs/by-name/hs/hss/package.nix
pkgs/by-name/mo/mosh/package.nix
pkgs/by-name/ps/pssh/package.nix
pkgs/by-name/rs/rset/package.nix
pkgs/by-name/ss/ssh-ident/package.nix
pkgs/by-name/ss/sshfs-fuse/common.nix
pkgs/by-name/ti/tigervnc/package.nix
pkgs/by-name/un/unix-privesc-check/package.nix
pkgs/desktops/gnome/extensions/gsconnect/default.nix
pkgs/development/perl-modules/ham/default.nix
pkgs/development/python-modules/clustershell/default.nix
pkgs/development/python-modules/fipy/default.nix
pkgs/development/python-modules/sfepy/default.nix
```

tomfitzhenry:
> And what do I do with all the other programs that wrap OpenSSH, that don't have an environment variable to override it?

hmenke@, sorry this impacted your usecases. After 1y, which usecases do you have that are impacted by this?

I'm open to `pkgs.openssh` having Kerberos, but `services.openssh.package = pkgs.opensshWithoutKerberos`. That way the client would have Kerberos support, but the server wouldn't. This addresses the server security concerns, which keeping Kerberos support for applications that use the client. I've not thought through compatibility issues of this, but just an idea.

hmenke:
I use Kerberos to log into machines at work. Those have home directories on NFSv4 with Kerberos, so I can't have passwordless login with SSH keys. Instead for passwordless login I obtain a Kerberos ticket on my local machine and use `GSSAPIAuthentication` to present that ticket to the server.

In my NixOS configuration I already set
```nix
{
  programs.mosh.package = pkgs.mosh.override {
    openssh = pkgs.opensshWithKerberos;
  };
  programs.ssh.package = pkgs.opensshWithKerberos;
}
```
which means that I am constantly building `mosh` from source again.

For `pass` I would need a full-blown nixpkgs overlay to replace the `openssh` dependency with a non-broken one. Right now, I always get to see

```console
$ pass git push
/home/henri/.ssh/config line 10: Unsupported option "gssapiauthentication"
/home/henri/.ssh/config line 11: Unsupported option "gssapidelegatecredentials"
Everything up-to-date
```

and that even though I use `IgnoreUnknown`, but I guess unknown ≠ unsupported.

```console
$ cat .ssh/config
Include ~/.ssh/config.d/*.conf

Host *
IgnoreUnknown GSSAPIAuthentication,GSSAPIDelegateCredentials
Compression yes
ControlMaster auto
ControlPath ~/.ssh/control-%L%h%p%r.sock
ControlPersist 600
ForwardX11 no
GSSAPIAuthentication yes
GSSAPIDelegateCredentials no
ServerAliveCountMax 3
ServerAliveInterval 300
```

PR #302802, opened 2024-04-09
linuxManualConfig: Use a pure-Nix readConfig

dramforever:
## Description of changes

Use a pure-Nix implementation to parse the configfile so that IFD can be
avoided in many cases.

The original decade-old readConfig function uses a bash script and requires an
IFD even when readFile would suffice. This is really unnecessary for many uses
of linuxManualConfig.

I'm not sure whether this is a good way to do it. Please advise on code organization and whether the interface should work like this.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/doc/packages/linux.section.md
+++ b/doc/packages/linux.section.md
@@ -56,7 +56,6 @@ An example of this is building a kernel for use in a VM or micro VM. You can use
   configfile = ./path_to_config_file;
   linux = pkgs.linuxManualConfig {
     inherit version src configfile;
-    allowImportFromDerivation = true;
   };
 }
 ```
@@ -74,10 +73,11 @@ If necessary, the version string can be slightly modified to explicitly mark it 
   configfile = ./path_to_config_file;
   linux = pkgs.linuxManualConfig {
     inherit version modDirVersion src configfile;
-    allowImportFromDerivation = true;
   };
 }
 ```
+
+If `configfile` is a derivation instead of a path, pass `allowImportFromDerivation = true;` to linuxManualConfig to allow reading the config file to detect options from it.
 
 :::
 
--- a/pkgs/os-specific/linux/kernel/manual-config.nix
+++ b/pkgs/os-specific/linux/kernel/manual-config.nix
@@ -8,15 +8,7 @@ let
   lib_ = lib;
   stdenv_ = stdenv;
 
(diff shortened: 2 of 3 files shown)


dramforever:
Ideally this is 0 rebuild (disregarding docs)... If it isn't I messed something up

dramforever:
I also checked a random config I had and this yielded exactly the same derivation as the original code.

tpwrules:
I've shipped an implementation [here](https://github.com/tpwrules/nixos-apple-silicon/blob/d47afc3f0f8b3078c818da8609c41340af61a2ec/apple-silicon-support/packages/linux-asahi/default.nix) for about a year and not had any problems, but never quite got around to submitting it upstream.

I think there is some additional subtlety around combining with kernel patch options and string options.

dramforever:
I do handle strings. These are everywhere and it wouldn't work without.

I think `configfile` is supposed to come after patches have been applied and in fact completely override patch options? You just give it a `.config` file instead of extraConfig in the format expected by nixpkgs. At least the old one didn't bother.

You're right about the last point, we should probably just ignore `allowImportFromDerivation` flag altogether, but idk if that would be good, or if we should just delete it altogether.

tpwrules:
I was just commenting that one of the uncertainties I had about my changes is whether or not it handled all varieties of string quoting (if multiple are even supported) in all circumstances, etc. It hasn't been tested or examined very thoroughly.

It does look like in the normal case extraConfig in patches is handled by the generic configuration stuff. I had to add it separately because my stuff doesn't use that (though it should in the future).

It's unclear if the config passed by the generic builder is even right?

I think this PR is okay, it's just not as much as I'd like. It's a step in the right direction though. I need to double check that it works properly with my stuff in that repo.

dramforever:
ah, yes, i almost forgot. i just delegated string parsing to `builtins.fromJSON` which may or may not be a bad idea.... i think the string extraction regex should be correct, but the parsing may need some more care for people who have say quotes in cflags or something.

tpwrules:
I'm trying to look if there's a defined syntax for that, I don't know what the possibilities are. My code just strips quotes.

PR #302591, opened 2024-04-08
nixos/steam: support using system fonts in Steam interface

eclairevoyant:
## Description of changes

Without this fix, all Chinese and Japanese text in the Steam interface display as boxes ("tofu").

As we use the existing fonts on the system, this avoids closure impact.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
- For non-Linux: Is sandboxing enabled in `nix.conf`? (See [Nix manual](https://nixos.org/manual/nix/stable/command-ref/conf-file.html))
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:
  - [NixOS test(s)](https://nixos.org/manual/nixos/unstable/index.html#sec-nixos-tests) (look inside [nixos/tests](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests))
  - and/or [package tests](https://github.com/NixOS/nixpkgs/blob/master/pkgs/README.md#package-tests)
  - or, for functions and "core" functionality, tests in [lib/tests](https://github.com/NixOS/nixpkgs/blob/master/lib/tests) or [pkgs/test](https://github.com/NixOS/nixpkgs/blob/master/pkgs/test)
  - made sure NixOS tests are [linked](https://nixos.org/manual/nixpkgs/unstable/#ssec-nixos-tests-linking) to the relevant packages
- [ ] Tested compilation of all packages that depend on this change using `nix-shell -p nixpkgs-review --run "nixpkgs-review rev HEAD"`. Note: all changes have to be committed, also see [nixpkgs-review usage](https://github.com/Mic92/nixpkgs-review#usage)
- [ ] Tested basic functionality of all binary files (usually in `./result/bin/`)
- [24.05 Release Notes](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2405.section.md) (or backporting [23.05](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2305.section.md) and [23.11](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2311.section.md) Release notes)
  - [ ] (Package updates) Added a release notes entry if the change is major or breaking
  - [ ] (Module updates) Added a release notes entry if the change is significant
[...]

PR #302862, opened 2024-04-09
nixos/steam: include fonts from system

jonringer:
## Description of changes
My attempt at https://github.com/NixOS/nixpkgs/pull/302591

Was able to successfully evaluate my desktop with these changes.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/nixos/modules/programs/steam.nix
+++ b/nixos/modules/programs/steam.nix
@@ -55,6 +55,10 @@ in {
             then [ package ] ++ extraPackages
             else [ package32 ] ++ extraPackages32;
         in prevLibs ++ additionalLibs;
+        extraPkgs = pkgs: let
+          prevPkgs = if prev ? extraPkgs then prev.extraPkgs pkgs else [ ];
+        in prevPkgs ++ config.fonts.packages;
+
       } // optionalAttrs (cfg.gamescopeSession.enable && gamescopeCfg.capSysNice)
       {
         buildFHSEnv = pkgs.buildFHSEnv.override {


eclairevoyant (commented):
It evals, but as I alluded to on the revert PR it turns out to not be a good solution. It just chooses a font at random without really respecting fontconfig.

jonringer:
> It evals, but as I alluded to on the revert PR it turns out to not be a good solution. It just chooses a font at random without really respecting fontconfig.

Interesting... I only have english fonts, so I lack the ability to e2e test this.

oxalica:
Could we add a smoke test for the steam module? Since it's unfree, maybe we could override `programs.steam.package` with a fake one for tests.

oxalica:
Just tested this PR. It evals and works for me. I have noto-fonts-cjk-* in system fonts and steam can correctly render Chinese characters now.

> It evals, but as I alluded to on the revert PR it turns out to not be a good solution. It just chooses a font at random without really respecting fontconfig.

I think something is better than nothing though.

eclairevoyant:
> I think something is better than nothing though.

I strongly disagree, I would simply document the `extraPkgs` arg instead of using something like this. Some fonts are unredable if used by Steam, we cannot have Steam randomly just pick a font

oxalica:
> Some fonts are unreadable if used by Steam

Could you elaborate what exact issue do you run into and what fonts are you using?

eclairevoyant:
Let me put it this way. If I have [Wingdings](https://en.wikipedia.org/wiki/Wingdings) installed on my system, does that mean I should have to navigate my computer if it chooses to display text in Wingdings? There are many fonts that do not display well at various resolutions. Simply _installing_ a font does not mean I want applications to suddenly use it as the _default_.

eclairevoyant:
Here's some samples below.

1. `config.font.packages`
2. `pkgs.wqy_zenhei`
3. `pkgs.source-han-sans`

In the Chinese text the problem is visible enough:

swappy-20240409_215134
swappy-20240409_215619
swappy-20240409_220237

In the Japanese text, the problem is worse because it picked a cursive font for kana:

swappy-20240409_215050
swappy-20240409_215735
swappy-20240409_220218

This might not demonstrate my issue clearly depending on your monitor setup, but I think the differences in stroke width are obvious on _my_ monitor at least.

I mean, maybe my eyes are already screwed anyway, but I find sample 1 (what this PR proposes) far more unreadable, at the distance I sit at from the monitor. Sample 2, I could live with it, but @Aleksanaa reported in https://github.com/NixOS/nixpkgs/pull/302591#issuecomment-2043123395 that some other users have issues even with wenquanyi. So to me, the solution would be to document how to override the Steam package and let the user choose whichever font is suitable for their screen.

oxalica:
> In the Chinese text the problem is visible enough:

Thanks for the explanation. But my point is that, comparing to the current situation on master, even a random font is better, because there is a non-zero chance that it happens to be readable. But it's zero chance to be readable if everything is tofu.

> Or maybe there is some way to apply the fontconfig to Steam as well, for proper control of the fonts chosen.

Of course, it would be better if we make fontconfig work. But I'm not sure how steam handles font rendering and why it's not working. It seems that `/etc/fonts` does exist in the FHS wrapper environment.

eclairevoyant:
> my point is that, comparing to the current situation on master, even a random font is better, because there is a non-zero chance that it happens to be readable. But it's zero chance to be readable if everything is tofu.

My point is that, if a user encounters tofu, they can override with the `extraPkgs` arg, choose some font they are happy with, and move on. If we forcibly put all the fonts on the system in, how will they fix that situation?

Atemu:
This could perhaps simply be an option then: `programs.steam.includeSystemFonts`. Its description could also point you at how to add your own fonts instead.

eclairevoyant:
> This could perhaps simply be an option then: `programs.steam.includeSystemFonts`. Its description could also point you at how to add your own fonts instead.

Seems like a good solution; let's go with that.

Issue #274191, opened 2023-12-14
Package request: ethercat


17steen:
**Project description**

This is a library for interfacing with hardware that uses the ethercat protocol.

It contains a kernel driver, a userland library, and some command line tools.

**Metadata**

* homepage URL: https://etherlab.org/en_GB/ethercat
* source URL: https://gitlab.com/etherlab.org/ethercat
* license: GNU LGPLv2.1
* platforms: linux

---

Add a :+1: [reaction] to [issues you find important].

[reaction]: https://github.blog/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/
[issues you find important]: https://github.com/NixOS/nixpkgs/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions-%2B1-desc

hugosenari:
This is related to #196556 ?

PR #302959, opened 2024-04-09
ethercat: Kernel module, documentation, NixOS module and VM test

stv0g:
## Description of changes

This PR packages the kernel modules, documentation and SystemD unit file of the IgH Ethercat master.

See also: https://github.com/NixOS/nixpkgs/issues/274191

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [x] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [x] Tested, as applicable:

---

### DIFF
--- a/pkgs/by-name/et/ethercat/common.nix
+++ b/pkgs/by-name/et/ethercat/common.nix
@@ -1,0 +1,60 @@
+{
+  lib,
+  fetchFromGitLab,
+  stdenv,
+  gitUpdater,
+  autoreconfHook,
+  pkg-config,
+}:
+stdenv.mkDerivation (finalAttrs: {
+  pname = "ethercat";
+  version = "1.6.2";
+
+  src = fetchFromGitLab {
+    owner = "etherlab.org";
+    repo = "ethercat";
+    rev = "refs/tags/${finalAttrs.version}";
+    hash = "sha256-NgRyvNvHy04jr7ieOscBYULRdWJ7YuHbuYbRrSfXYRU=";
+    fetchSubmodules = true;
+    leaveDotGit = true;
+  };
+
+  configureFlags = [
+    # Features
+    "--enable-eoe=yes"
+    "--enable-cycles=yes"
+    "--enable-rtmutex=yes"
+    "--enable-hrtimer=yes"
+    "--enable-regalias=yes"
+    "--enable-refclkop=yes"
+    "--enable-tty=no" # Is broken in Kernel 6.6
+    "--enable-wildcards"
+
+    # Debugging
+    "--enable-debug-if=yes"
+    "--enable-debug-ring=no" # Is broken
+
+    # Devices
+    "--enable-generic"
+    "--enable-ccat"
+  ];
+
+  nativeBuildInputs = [
+    autoreconfHook
+    pkg-config
+  ];
+
+  enableParallelBuilding = true;
+  separateDebugInfo = true;
+
+  passthru.updateScript = gitUpdater { };
+
+  meta = {
+    description = "IgH EtherCAT Master for Linux";
+    homepage = "https://etherlab.org/ethercat";
+    changelog = "https://gitlab.com/etherlab.org/ethercat/-/blob/1.6-devel/NEWS";
+    license = lib.licenses.gpl2Plus;
+    maintainers = [ lib.maintainers.stv0g ];
+    platforms = lib.platforms.linux;
+  };
+})
--- a/pkgs/by-name/et/ethercat/common.nix
+++ b/pkgs/by-name/et/ethercat/common.nix
@@ -65,7 +65,7 @@ stdenv.mkDerivation (finalAttrs: {
   meta = {
     description = "IgH EtherCAT Master for Linux";
     homepage = "https://etherlab.org/ethercat";
-    changelog = "https://gitlab.com/etherlab.org/ethercat/-/blob/1.6-devel/NEWS";
+    changelog = "https://gitlab.com/etherlab.org/ethercat/-/blob/${finalAttrs.version}/NEWS";
     license = lib.licenses.gpl2Plus;
     maintainers = [ lib.maintainers.stv0g ];
     platforms = lib.platforms.linux;
--- a/pkgs/by-name/et/ethercat/common.nix
+++ b/pkgs/by-name/et/ethercat/common.nix
@@ -1,10 +1,13 @@
 {
-  lib,
-  fetchFromGitLab,
-  stdenv,
-  gitUpdater,
   autoreconfHook,
+  fetchFromGitLab,
+  gitUpdater,
+  lib,
+  nixosTests,
   pkg-config,
+  stdenv,
+
+  # Options
   withDocs ? false,
 }:
 stdenv.mkDerivation (finalAttrs: {
@@ -60,7 +63,12 @@ stdenv.mkDerivation (finalAttrs: {
   enableParallelBuilding = true;
   separateDebugInfo = true;
 
-  passthru.updateScript = gitUpdater { };
+  passthru = {
+    updateScript = gitUpdater { };
+    tests = {
+      inherit (nixosTests) ethercat;
+    };
+  };
 
   meta = {
     description = "IgH EtherCAT Master for Linux";
--- a/pkgs/by-name/et/ethercat/common.nix
+++ b/pkgs/by-name/et/ethercat/common.nix
@@ -5,19 +5,32 @@
   gitUpdater,
   autoreconfHook,
   pkg-config,
+  withDocs ? false,
 }:
 stdenv.mkDerivation (finalAttrs: {
   pname = "ethercat";
   version = "1.6.2";
 
-  src = fetchFromGitLab {
-    owner = "etherlab.org";
-    repo = "ethercat";
-    rev = "refs/tags/${finalAttrs.version}";
-    hash = "sha256-NgRyvNvHy04jr7ieOscBYULRdWJ7YuHbuYbRrSfXYRU=";
-    fetchSubmodules = true;
-    leaveDotGit = true;
-  };
+  src = fetchFromGitLab (
+    {
+      owner = "etherlab.org";
+      repo = "ethercat";
+      rev = "refs/tags/${finalAttrs.version}";
+
+    }
+    // (
+      if withDocs then
+        {
+          hash = "sha256-b04T33eYmlCkLE+bVdYkUGtNaQDUZ+xlDdXM2w3Me3k=";
+          fetchSubmodules = true;
+          leaveDotGit = true;
+        }
+      else
+        {
+          hash = "sha256-NgRyvNvHy04jr7ieOscBYULRdWJ7YuHbuYbRrSfXYRU=";
+        }
+    )
+  );
 
   configureFlags = [
     # Features
--- a/pkgs/by-name/et/ethercat/kernel.nix
+++ b/pkgs/by-name/et/ethercat/kernel.nix
@@ -1,0 +1,199 @@
+{
+  autoreconfHook,
+  pkg-config,
+  stdenv,
+  pkgs,
+  lib,
+  kernel,
+}:
+let
+  deviceKernelCompatabilities = {
+    "6.6" = [ "igc" ];
+    "6.4" = [
+      "e100"
+      "igc"
+    ];
+    "6.1" = [
+      "8139too"
+      "bcmgenet"
+      "e100"
+      # "e1000" # Is broken
+      "e1000e"
+      "igb"
+      # "igc" # Is broken
+      "r8169"
+    ];
+    "5.15" = [
+      "8139too"
+      "e100"
+      "e1000"
+      "e1000e"
+      "igb"
+      "r8169"
+    ];
+    "5.14" = [
+      "8139too"
+      "bcmgenet"
+      "e100"
+      "e1000"
+      "e1000e"
+      "igb"
+      "igc"
+      "r8169"
+    ];
+    "5.10" = [
+      "8139too"
+      "bcmgenet"
+      "e100"
+      "e1000"
+      "e1000e"
+      "igb"
+      "r8169"
+    ];
+    "5.4" = [
+      "e100"
+      "e1000e"
+    ];
+    "4.19" = [ "igb" ];
+    "4.4" = [
+      "8139too"
+      "e100"
+      "e1000"
+      "e1000e"
+      "igb"
+      "r8169"
+    ];
+    "3.18" = [ "igb" ];
+    "3.16" = [
+      "8139too"
+      "e100"
+      "e1000"
+      "e1000e"
+      "r8169"
+    ];
+    "3.14" = [
+      "8139too"
+      "e100"
+      "e1000"
+      "e1000e"
+      "r8169"
+    ];
+    "3.12" = [
+      "8139too"
+      "e100"
+      "e1000"
+      "e1000e"
+      "r8169"
+    ];
+    "3.10" = [
+      "8139too"
+      "e100"
+      "e1000"
+      "e1000e"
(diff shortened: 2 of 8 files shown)


pkgs/by-name/et/ethercat/kernel.nix
```diff
@@ -0,0 +1,175 @@
...
+  pkg-config,
+  stdenv,
+  lib,
+  kernel,
+  ethercat,
+}:
+let
+  common = callPackage ./common.nix { };
+
+  deviceKernelCompatabilities = {
```
17steen, line 13: Not sure what device kernel compatiblities means, but i don't see "generic" here, is it implied that it's compatible with all kernels?

stv0g: Exactly, the `generic` and `ccat` drivers are supported irrespective of the Kernel version.
They are activated by default here:

https://github.com/NixOS/nixpkgs/pull/302959/files#diff-b399f63d866c72ee8718c3af74adc6f25eff65ef6ffa93db3f77fd01974ee85eR40

stv0g:
Some feedback is welcome about my usage of `common.nix` in this PR.

It is currently just a plain set of shared attributes of the package and kernel-module derivations.
But I think maybe even cleaner if common itself is a derivation, and we simply override attributes specific for the package and the kernel module?

Using the exact same derivation does not work as standard packages are build independently from a kernel...

pkgs/by-name/et/ethercat/kernel.nix
17steen: great!, thanks for the clarification

17steen (approved):
Not sure I'm versed enough in Nix to give a proper review,

Now from looking at it it looks good, i'm wondering why choosing to package a development build instead of the stable branch (stable-1.5), but that's your choice of course

17steen:
> > i'm wondering why choosing to package a development build instead of the stable branch (stable-1.5), but that's your choice of course
> 
> Thats a valid question.. I've opted to package the current development version due to these reasons:
> 
>     * The last stable release [1.5.2](https://gitlab.com/etherlab.org/ethercat/-/tags/1.5.2) is really outdated. I was released 11 years ago.
> 
>     * There were several packaging related changes in 1.6-devel regarding the building of documentation. I wanted to package it for the new changes rather the old to avoid redoing it once 1.6 becomes stable.
> 
>     * I've seen quite some activity in the repo recently which gives me to believe that 1.6 could become stablized soon at which point I will bump the package in Nixpkgs.

The tag is old but the stable branch is still maintained, getting frequent bugfixes.
But indeed if you think 1.6 stabilisation is near, it's not worth the hassle to support the older version,

stv0g:
Hey @simonkampe,

do you maybe have some minutes to review this PR? I would really like to see the Kernel module beeing merged in Nixpkgs :)

simonkampe (approved):
Looks fine by me, if we want to test it with some actual ethercat devices I can do that but that will take a little while.

stv0g:
I've rebased the PR to latest master. We are now using the upstream release v1.6.0 instead of the previous devel version.

stv0g:
I've updated/rebased the PR to the latest master.

Additional reviews or a merge to master would be very welcome :)

pkgs/by-name/et/ethercat/kernel.nix
```diff
@@ -0,0 +1,199 @@
...
+  supportedDrivers = deviceKernelCompatabilities.${kernelVersionMajorMinor};
+in
+stdenv.mkDerivation (
+  finalAttrs:
+  (
+    with pkgs.ethercat;
+    {
+      inherit version;
+      inherit src;
+      inherit meta;
```
SuperSandro2000:
```suggestion
      inherit version src meta;
```

pkgs/by-name/et/ethercat/kernel.nix
```diff
@@ -0,0 +1,199 @@
...
+  kernelVersionMajorMinor = builtins.concatStringsSep "." (
+    lib.lists.take 2 (builtins.splitVersion kernel.version)
+  );
+
+  supportedDrivers = deviceKernelCompatabilities.${kernelVersionMajorMinor};
+in
+stdenv.mkDerivation (
+  finalAttrs:
+  (
+    with pkgs.ethercat;
```
SuperSandro2000: We do not want to use with's over such a large scope

pkgs/by-name/et/ethercat/kernel.nix
```diff
@@ -0,0 +1,199 @@
+{
+  autoreconfHook,
+  pkg-config,
+  stdenv,
+  pkgs,
+  lib,
+  kernel,
+}:
+let
+  deviceKernelCompatabilities = {
```
SuperSandro2000, line 14: We must comment the source of this

pkgs/by-name/et/ethercat/package.nix
```diff
@@ -15,7 +22,9 @@ stdenv.mkDerivation (finalAttrs: {
     owner = "etherlab.org";
     repo = "ethercat";
     rev = "refs/tags/${finalAttrs.version}";
-    hash = "sha256-fvuyADZzH1lHo1TkGd9s3n2n8y9vsGgs/ZOakIw48WI=";
+    hash = "sha256-w1aoznLjCHh+dN2fyfpKpzq68B4D/rert2XFpTxk/F4=";
+    fetchSubmodules = true;
+    leaveDotGit = true;
```
SuperSandro2000: Can't we mock the git information? this is now likely sometimes unreproducible

pkgs/by-name/et/ethercat/package.nix
```diff
@@ -57,6 +67,39 @@ stdenv.mkDerivation (finalAttrs: {
...
+    make -C documentation
+
+    cp documentation/*.pdf $doc
+  '';
+
+  postFixup = ''
+    mv $out/{share,lib,etc} $bin
+    find $out
+    echo
+    find $bin
```
SuperSandro2000:
```suggestion
```

pkgs/by-name/et/ethercat/package.nix
```diff
@@ -57,6 +67,39 @@ stdenv.mkDerivation (finalAttrs: {
...
+
+    echo "Build Doxygen LaTeX docs"
+    make -C doxygen-output/latex
+    cp -r doxygen-output/latex/refman.pdf $doc/ethercat_ref.pdf
+
+    echo "Build LateX manual"
+    mkdir -p documentation/external
+    make -C documentation
+    make -C documentation index
+    make -C documentation
```
SuperSandro2000: Not sure if the build time is worth it. Also pdf's could be a bit bigger.

pkgs/by-name/et/ethercat/kernel.nix
stv0g: @SuperSandro2000 I think you somehow reviewer an outdated commit of the PR.

Its already fixed in 14a1b28727fac1532a2b21269c0f5f0d317e4405

pkgs/by-name/et/ethercat/kernel.nix
stv0g: Resolved in 14a1b28727fac1532a2b21269c0f5f0d317e4405

pkgs/by-name/et/ethercat/kernel.nix
stv0g: Resolved with

https://github.com/NixOS/nixpkgs/blob/14a1b28727fac1532a2b21269c0f5f0d317e4405/pkgs/by-name/et/ethercat/kernel.nix#L13-L14

pkgs/by-name/et/ethercat/package.nix
stv0g: The Git information is used in the documentation here:

https://github.com/NixOS/nixpkgs/blob/14a1b28727fac1532a2b21269c0f5f0d317e4405/pkgs/by-name/et/ethercat/package.nix#L61

I am not sure how I could mock the Git commit date or hash..

pkgs/by-name/et/ethercat/package.nix
stv0g: Fixed in 14a1b28727fac1532a2b21269c0f5f0d317e4405

pkgs/by-name/et/ethercat/package.nix
stv0g: I made the building of the LaTex documentation optional in 7324d105a131255c2baa97ff5949ab009776eec4

This also allows us to drop the `leaveDotGit` and `fetchSubmodule` options from `fetchFromGitHub`.

pjungkamp (commented):
You should probably build the documentation in the `buildPhase`, e.g. in `postBuild`, and only move the files to `$doc` in the `postInstall` hook. I haven't built this derivation yet, is the documentation reproducable? It seems to include the build time in the pdf.

The PDF files should be copied to `$doc/share/doc/`, so that they are actually merged in `/run/current-system/sw/share/doc` if you have `documentation.doc.enable = true` in your NixOS config.

pkgs/by-name/et/ethercat/package.nix
```diff
@@ -1,38 +1,90 @@
...
-, lib
-, pkg-config
-, stdenv
-, fetchFromGitLab
-, gitUpdater
+{
+  callPackage,
+  lib,
+  autoreconfHook,
+  systemd,
```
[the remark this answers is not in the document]
stv0g: Good point. I addressed this in 020e74babc88fbf7f156265f1714d1254fbe5ff2

pkgs/by-name/et/ethercat/package.nix
```diff
@@ -1,38 +1,90 @@
...
+      babel-french
+      ;
   };
+in
+common.overrideAttrs (
+  finalAttrs: previousAttrs: {
+    buildInputs = [ systemd ];
 
-  separateDebugInfo = true;
+    nativeBuildInputs = previousAttrs.nativeBuildInputs ++ lib.optional withDocs [
```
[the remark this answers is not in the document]
stv0g: Resolved with 2850d8272119b6314ffbfcf52c0ebad76db7d401

stv0g:
Can we somehow include the building of the documentation still in the Hydra tests? E.g. building the derivation with the `withDocs` argument?
I've have tested it now only locally.

stv0g:
> You should probably build the documentation in the `buildPhase`, e.g. in `postBuild`, and only move the files to `$doc` in the `postInstall` hook. I haven't built this derivation yet, is the documentation reproducable? It seems to include the build time in the pdf.
> 
> The PDF files should be copied to `$doc/share/doc/`, so that they are actually merged in `/run/current-system/sw/share/doc` if you have `documentation.doc.enable = true` in your NixOS config.

Resolved with 2850d8272119b6314ffbfcf52c0ebad76db7d401

Necropaw:
Thank you for your understanding, as I am a NixOS noob with no previous NixOS experience. 

I got to the point, that I can run `ethercatctl start`. But this results in the error: 
```
/run/current-system/sw/bin/ethercatctl: line 108: /sbin/modprobe: No such file or directory
```

Looking at the `ethercatctl` script it is no wonder, as some execs are expected at the "default" locations:
```
LSMOD=/sbin/lsmod
MODPROBE=/sbin/modprobe
RMMOD=/sbin/rmmod
MODINFO=/sbin/modinfo
IP=/bin/ip
``` 

Any idea how this can be solved?

stv0g:
@Necropaw Can you test it with my latest fix in https://github.com/NixOS/nixpkgs/pull/302959/commits/89eb8c6af54338c4f8acb0372485ef7085b9d811?

Necropaw:
@stv0g this also rewrites the shebang and results (on my system) in the first line in `ethercatctl`:
```
#!/nix/store/izpf49b74i15pcr9708s3xdwyqs4jxwl-bash-5.2p32/nix/store/vg4vfjx3iigcp9d762ciqkjibz3hgfq1-iproute2-6.10.0/bin/bash
```

stv0g:
Sorry, my fault. Fixed in 5c3fa7f454ff7860ab26b6208a0f9f8ccc118293

stv0g:
@SuperSandro2000 Do you maybe have some bandwidth to revisit this PR? I have addressed all your comments.

pkgs/by-name/et/ethercat/common.nix
```diff
@@ -0,0 +1,61 @@
...
+  src = fetchFromGitLab {
+    owner = "etherlab.org";
+    repo = "ethercat";
+    rev = "refs/tags/${finalAttrs.version}";
+    hash = "sha256-Ki8WUKUW4O9SNFTVIXrWpL6suCu7j2nLQtpWVsUoiK0=";
+    fetchSubmodules = withDocs;
+    leaveDotGit = withDocs;
+  };
+
+  meta = {
```
SuperSandro2000, line 23: meta should be last

pkgs/by-name/et/ethercat/common.nix
```diff
@@ -0,0 +1,61 @@
...
+    description = "IgH EtherCAT Master for Linux";
+    homepage = "https://etherlab.org/ethercat";
+    changelog = "https://gitlab.com/etherlab.org/ethercat/-/blob/1.6-devel/NEWS";
```
SuperSandro2000, line 26: We should substitute the version number in here

pkgs/by-name/et/ethercat/common.nix
```diff
@@ -0,0 +1,61 @@
...
+  pname = "ethercat";
+  version = "1.6.1";
+
+  src = fetchFromGitLab {
+    owner = "etherlab.org";
+    repo = "ethercat";
+    rev = "refs/tags/${finalAttrs.version}";
+    hash = "sha256-Ki8WUKUW4O9SNFTVIXrWpL6suCu7j2nLQtpWVsUoiK0=";
+    fetchSubmodules = withDocs;
+    leaveDotGit = withDocs;
```
SuperSandro2000, line 20: This will not work as the hash changes

stv0g: I've reworked this in da57d7998d6f3b3b8e009f53050f062ad77bf291

pkgs/by-name/et/ethercat/common.nix
stv0g: Fixed in b2a4c30a695f671775395b2786e66b754e2ceb68

pkgs/by-name/et/ethercat/common.nix
stv0g: Fixed in e792a4370cc0140573d19c4013ce7b375f464d10

stv0g:
I also added a new NixOS module to configure the EtherCAT kernel module in b51ca578ae500cbcabd7af2df97ce2deb9ef00d8, as well as a NixOS VM test alongside it in 7e95ba9df377bfe5f5463a325e201c53af067609 which verifies that the kernel module can be loaded and the userspace tooling detects a dummy EtherCAT master interface.

DanielIndictor:
> Looks fine by me, if we want to test it with some actual ethercat devices I can do that but that will take a little while.

I was able to communicate with real Ethercat devices (specifically, using the `libethercat` userspace driver as well as the `ethercat` terminal program) using this PR. I used it to blink the outputs of a Beckhoff EL2001 and read PDO/SDO information of a Beckhoff EL2001, Beckhoff EK1100, and ClearPath-EC motor. If you just needed a third party confirmation of this PR's functionality, I hope my testimony can suffice.

stv0g:
Hi all,

there are now two confirmed tests of this module with real hardware (mine and @DanielIndictor).
And we also already had a couple of review rounds.

Do we still see a chance to get this merged? For those who already reviewed the changes: could you please approve the PR?

PR #302970, opened 2024-04-09
cargo-deps: remove

philiptaron:
## Description of changes

The repository has been deleted, and no alternate can be found. https://crates.io/crates/cargo-deps still exists.

## Things done

- [x] `nixpkgs-review rev HEAD` (leaf package removed)


pkgs/top-level/aliases.nix
```diff
@@ -151,6 +151,7 @@ mapAliases ({
   cadence = throw "cadence has been removed from nixpkgs, as it was archived upstream"; # Added 2023-10-28
   cask = emacs.pkgs.cask; # Added 2022-11-12
   canonicalize-jars-hook = stripJavaArchivesHook; # Added 2024-03-17
+  cargo-deps = throw "cargo-deps has been removed as the repository is deleted"; # Added 2024-04-09
```
donovanglover, line 154: Should we mention the [alternatives](https://crates.io/crates/cargo-deps#alternatives) in the readme?

philiptaron: Sorry, which readme?

matthiasbeyer:
It rather seems that the github account was taken over by some 3rd party

image

(or was that there before?)

I think we should definitively remove it, because if there was indeed a takeover and the third party (re-)publishes the repository with malicious code, we might have a _real_ issue if we (auto) update the package with that malicious code.

Lets remove it.

matthiasbeyer (approved):

PR #301416, opened 2024-04-04
cuda-modules: apply lessons learned from cross-compilation attempts

ConnorBaker:
> [!Important]
>
> To-do:
> - [x] Review, merge, and rebase on #301176
> - [x] Revert/fix change made in #301180 which breaks discovery of multiple-outputs
>   - See https://github.com/NixOS/nixpkgs/pull/301180#issuecomment-2043734826
> - [x] Ensure the switch to `cmakeFlagsArray` in the setup hooks does not cause problems by (for example) overriding user-specified values in `cmakeFlags` because of the order of string interpolation in the CMake setup hook.
> - [x] Address feedback.

> [!Note]
>
> - Setup hook rewrites were split into #302694
> - Documentation on cross-compilation attempts was split into #303122

## Description of changes

Supersedes #279952 and #275560.

This PR instead adds fixes and updates accumulated over the course of work on #279952 and #275560. Work related specifically to rewriting the setup hooks is covered by #302694.

- Access `*Platform` through `stdenv` rather than directly from `pkgs`
- Update uses of `substituteInPlace` to `--replace-fail` to fix deprecation
- Add CMake utilities `cmakeCudaArchitectures` and `cmakeCudaArchitecturesString`
- Always provide explicit argument to `return` so it does not use the return value of the previously executed command (which would sometimes cause failures when evaluating bash functions)
- Fixed an error which could occur in `setupCudaHook` wherein attempting to read an empty marker file would fail
- Updated `nccl` and `saxpy`
- Overhaul of CUDA overrides for more consistent structure and use of attributes such as `brokenConditions`

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
- For non-Linux: Is sandboxing enabled in `nix.conf`? (See [Nix manual](https://nixos.org/manual/nix/stable/command-ref/conf-file.html))
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:
  - [NixOS test(s)](https://nixos.org/manual/nixos/unstable/index.html#sec-nixos-tests) (look inside [nixos/tests](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests))
  - and/or [package tests](https://nixos.org/manual/nixpkgs/unstable/#sec-package-tests)
[...]

PR #303122, opened 2024-04-10
cudaPackages: add documentation on cross-compilation

ConnorBaker:
> [!Important]
>
> To-do:
> - [ ] Revise language in `pkgs/development/cuda-modules/ON-CROSS-COMPILATION.md`, adding examples of prior work and possible routes for further investigation.

## Description of changes

Split off from #301416 with the goal of discussing and revising docs on cross-compilation separately from otherwise unrelated fixes and improvements to `cudaPackages`.

CC @NixOS/cuda-maintainers @samuela @SomeoneSerge @kmittman

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/development/cuda-modules/ON-CROSS-COMPILATION.md
+++ b/pkgs/development/cuda-modules/ON-CROSS-COMPILATION.md
@@ -1,0 +1,168 @@
+# On cross-compilation
+
+> [!NOTE]
+>
+> I am also hesitant to discourage future contributors by declaring this
+> impossible. Perhaps we could document precisely the project state, attempted
+> solutions, and technical considerations and leave open the question how to
+> move forward?
+>
+> - @samuela
+
+## Summary
+
+The CUDA redistributables[^1], individually packaged pieces of software
+historically only made available through the CUDA Toolkit, do not appear
+equipped for cross-compilation. NVIDIA's documentation makes it seem possible
+only with special versions of the Debian CUDA Toolkit installer[^2].
+
+Given the NixOS CUDA Team's focus on moving to the CUDA redistributables and
+eventually dropping support for the CUDA Toolkit runfile installer, adding
+support for cross-compilation requires further research and analysis. That said,
+contributions improving the state of support for cross-compilation are more than
+welcome!
+
+## Background
+
+### Cross-compilation
+
+Cross-compilation holds varying levels of importance for the diverse user base
+of Nixpkgs. It allows for the building of software for a different architecture
+than the one on which the build is taking place. This is particularly useful for
+embedded systems, where one might want to use a powerful `x86_64` machine, but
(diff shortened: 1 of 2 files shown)


ConnorBaker:
Specific comments and points to address, from #301416.

A superficial analysis of the differences between redistributables and the CUDA Toolkit run file/Debian installer:

> Superficially, I can say there's something going on with `host-*` and `target-*` subdirectories present in the CUDA Toolkit (of all kinds, I believe, cross or not).
>
> In the non-cross CUDA Toolkit, on `x86_64-linux` (that's the NixOS system name) there will be a `host-linux-x64` and `target-linux-x64` subdirectory in the root of the CUDA Toolkit.
>
> In the cross compilation CUDA Toolkit, the values might well be something like `host-linux-x64` with `target-linux-sbsa` (ARM servers) or `target-linux-aarch64` (Jetson devices).
>
> I'd have to spin up an Ubuntu system to install the cross-compilation version of the CUDA Toolkit and trace through the flow of CMake to see how it's driving cross-compilation.
>
> I fear that beyond the presence of these host and target directories, the binaries in the cross-compilation version of the CUDA Toolkit are different from those offered in the redistributables -- though I haven't done any sort of analysis on the individual binaries themselves, I worry they're compiled differently from NVIDIA's sources in such a way that we cannot mimic the changes necessary for cross-compilation in an ad-hoc manner.
>
> At any rate, it would require more time than I'm able to allocate.
>
> @ConnorBaker in https://github.com/NixOS/nixpkgs/pull/301416#discussion_r1551774908

Cross-packages should be a subset of the native packages:

> Hmmm, I'm not entirely sure why these issues are cropping up, seems to be related to the `targets/` directory symlinks.
>
> The binaries are the same, they are built once per platform/version and then packaged into each format; some formats (Conda and Python wheels) involve `patchelf`.
>
> I would expect the "binary archive" tarballs, Debian packages, and RPM packages to be bit-identical.
>
> The cross-packages are more-or-less a subset of the native packages; some libraries provide stub files, while others use the native runtime.
>
> @kmittman in https://github.com/NixOS/nixpkgs/pull/301416#discussion_r1556084671

On future approaches to a minimum working example:

> We really should (have) start(ed) with an MWE: piecing together an nvcc derivation sufficient to run `nvcc` directly on a single file (saxpy but without cmake), manually copying stuff from archives for different systems, without hooks or much automation
>
> @SomeoneSerge in https://github.com/NixOS/nixpkgs/pull/301416#discussion_r1558575644

PR #297901, opened 2024-03-22
Switch the `pkgs/by-name` check to a separate repository

infinisil:
## Description of changes

This completes the move of the `pkgs/by-name` tooling to the new https://github.com/NixOS/nixpkgs-check-by-name repository, see https://github.com/NixOS/nixpkgs/issues/286559 for motivation.

> [!Note]
> I'm leaving in both `pkgs/test/nixpkgs-check-by-name` to make it easier to review the removal in a separate PR, and `pkgs/test/nixpkgs-check-by-name/scripts/pinned-tool.json` because it's needed for CI of this PR to succeed.

This work is sponsored by [Antithesis](https://antithesis.com/) :sparkles:

## Things done
- [x] Test it in CI: https://github.com/tweag/nixpkgs/pull/87
- [x] Test it locally: `./maintainers/scripts/check-by-name.sh master`
- [x] Update all comments (especially in the workflow)

---

Add a :+1: [reaction] to [pull requests you find important].

[reaction]: https://github.blog/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/
[pull requests you find important]: https://github.com/NixOS/nixpkgs/pulls?q=is%3Aopen+sort%3Areactions-%2B1-desc

PR #303141, opened 2024-04-10
CODEOWNERS: Fix check-by-name path

infinisil:
Was moved in https://github.com/NixOS/nixpkgs/pull/297901

I'm not getting notifications otherwise :)


philiptaron (commented):
# 🦉 
Watch that `check-by-name` like a hawk. Still wish it were named `nixpkgs-vet`. 😉

PR #303253, opened 2024-04-11
nixos/restic: Add optional global read capability

busyboredom:
## Description of changes

Add the option to use a security wrapper with elevated read permissions for restic. When enabled, restic can backup files owned by other users without the need to run as root.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [x] Tested, as applicable:

---

### DIFF
--- a/nixos/modules/services/backup/restic.nix
+++ b/nixos/modules/services/backup/restic.nix
@@ -300,7 +300,7 @@ in
       mapAttrs'
         (name: backup: nameValuePair "restic-backups-${name}" {
            source = "${backup.package}/bin/restic";
-           owner = "${backup.user}";
+           owner = backup.user;
            group = "restic";
            permissions = "u=rwx,g=,o=";
            capabilities = "cap_dac_read_search=+ep";
--- a/nixos/modules/services/backup/restic.nix
+++ b/nixos/modules/services/backup/restic.nix
@@ -256,6 +256,14 @@ in
             having to manually specify most options.
           '';
         };
+
+        elevatedReadPermission = lib.mkOption {
+          type = lib.types.bool;
+          default = false;
+          description = lib.mdDoc ''
+            Whether to use a security wrapper to allow backing up data owned by other users.
+          '';
+        };
       };
     }));
     default = { };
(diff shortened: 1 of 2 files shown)


busyboredom:
Seems like a good security improvement for folks who need to backup data owned by multiple users. For example, on a server with separate users per service. The alternative is to have a separate repo per user and give up deduplication between users, but I was unhappy with that option.

Happy to add tests if this is a change we want to go forward with (just gotta figure out how testing works first, I'm new here).

NickCao:
> backup files owned by other users without the need to run as root

This still feels like a security nightmare. Since everyone on the machine (even nobody), would be able to read the sensitive files of other users.

busyboredom:
> everyone on the machine (even nobody), would be able to read the sensitive files of other users.

Only the user(s) with access to the wrapped restic can use it. So if you do:
```
users.users.alice = {
  isNormalUser = true;
};

users.users.bob = {
  isNormalUser = true;
};

services.restic.backup.elevated-backup = {
   ...    
    user = "alice";
    repository = userRepository;
    elevatedReadPermission = true;
};
```

Then Alice will have access to user data through restic, but bob will not. The idea is that Alice would be a single-purpose system user just for backups. Without my change, alice has to be root.

nixos/modules/services/backup/restic.nix
```diff
@@ -287,12 +295,24 @@ in
       assertion = (v.repository == null) != (v.repositoryFile == null);
       message = "services.restic.backups.${n}: exactly one of repository or repositoryFile should be set";
     }) config.services.restic.backups;
+    users.groups.restic = {};
+    security.wrappers =
+      mapAttrs'
+        (name: backup: nameValuePair "restic-backups-${name}" {
+           source = "${backup.package}/bin/restic";
+           owner = "${backup.user}";
```
MinerSebas, line 303:
```suggestion
           owner = backup.user;
```

busyboredom: I've applied your suggestion, thanks!

PR #271161, opened 2023-11-30
treewide: support NIX_SSL_CERT_FILE as an impureEnvVar

timbertson:
## Description of changes

**tl;dr** this PR makes `NIX_SSL_CERT_FILE` the preferred way to control CA certificates throughout nix, by allowing it to be set as in impureEnvVar, including it in proxyImpureEnvVars, and ensuring fetchers support it.

----

Nix (the executable) as well as much of its packaged software respects `NIX_SSL_CERT_FILE` as the preferred way to specify a custom set of root certificates. Within derivations, this is typically done by the setup hook of `cacert`.

However, there's no way to inject this variable from _outside_ (with impure env vars used in fetchers). This is required for most use cases of an an https proxy, as the proxy's own certificate will not likely be trusted by nix's builtin cacert package.

Functionality for injecting a custom trust store has been added to `fetchgit`, but because the setuphook will always override `NIX_SSL_CERT_FILE`, a different envvar had to be used (`NIX_GIT_SSL_CAINFO`). I have [an open PR](https://github.com/NixOS/nixpkgs/pull/266643) to add this same customisation to the go module fetcher.

However, it'd be preferable if we didn't invent a new envvar, and instead made `NIX_SSL_CERT_FILE` the single way to control this setting for both fetchers and nix itself.

For this to work, we need two small changes:
 - `cacert` should only set this variable if it's not already set, so that any impure version is not overwritten
   - this is in a setuphook so it's a mass rebuild
 - add `"NIX_SSL_CERT_FILE"` to `proxyImpureEnvVars` (fetcher.nix). It's not strictly a proxy-only variable, but:
   - this needs to be overridden for all HTTPS proxy use cases, and most things are https these days. Currently fetcher support for custom certs is patchy, because they're all doing their own thing (see below notes)
   - using the certificates the system has explicitly set for nix (if any) is inline with user expectations even outside proxy setups, and this only affects fixed-output derivations anyway

Various fetchers current (and updated) behaviour is oulined below:

 - fetchgit: previously added explicit support for `NIX_GIT_SSL_CAINFO` as the impure overrideable version. This is still supported, but the standard `NIX_SSL_CERT_FILE` now works and is preferred.
 - fetchurl: passes `--insecure` to curl for a fixed-output derivation, so I guess it doesn't need modification
 - fetchgomodule: previously no support for custom certs for git dependencies, now sets `GIT_SSL_CAINFO` to the value of `NIX_SSL_CERT_FILE` using whatever's provided by the user / setup hook

[...]

PR #303307, opened 2024-04-11
treewide: support NIX_SSL_CERT_FILE as an impureEnvVar

timbertson:
(this is a reopened version of https://github.com/NixOS/nixpkgs/pull/271161 for technical reasons)

## Description of changes

**tl;dr** this PR makes `NIX_SSL_CERT_FILE` the preferred way to control CA certificates throughout nix, by allowing it to be set as in impureEnvVar, including it in proxyImpureEnvVars, and ensuring fetchers support it.

----

Nix (the executable) as well as much of its packaged software respects `NIX_SSL_CERT_FILE` as the preferred way to specify a custom set of root certificates. Within derivations, this is typically done by the setup hook of `cacert`.

However, there's no way to inject this variable from _outside_ (with impure env vars used in fetchers). This is required for most use cases of an an https proxy, as the proxy's own certificate will not likely be trusted by nix's builtin cacert package.

Functionality for injecting a custom trust store has been added to `fetchgit`, but because the setuphook will always override `NIX_SSL_CERT_FILE`, a different envvar had to be used (`NIX_GIT_SSL_CAINFO`). I have [an open PR](https://github.com/NixOS/nixpkgs/pull/266643) to add this same customisation to the go module fetcher.

However, it'd be preferable if we didn't invent a new envvar, and instead made `NIX_SSL_CERT_FILE` the single way to control this setting for both fetchers and nix itself.

For this to work, we need two small changes:
 - `cacert` should only set this variable if it's not already set, so that any impure version is not overwritten
   - this is in a setuphook so it's a mass rebuild
 - add `"NIX_SSL_CERT_FILE"` to `proxyImpureEnvVars` (fetcher.nix). It's not strictly a proxy-only variable, but:
   - this needs to be overridden for all HTTPS proxy use cases, and most things are https these days. Currently fetcher support for custom certs is patchy, because they're all doing their own thing (see below notes)
   - using the certificates the system has explicitly set for nix (if any) is inline with user expectations even outside proxy setups, and this only affects fixed-output derivations anyway

Various fetchers current (and updated) behaviour is oulined below:

 - fetchgit: previously added explicit support for `NIX_GIT_SSL_CAINFO` as the impure overrideable version. This is still supported, but the standard `NIX_SSL_CERT_FILE` now works and is preferred.
 - fetchurl: passes `--insecure` to curl for a fixed-output derivation, so I guess it doesn't need modification
 - fetchgomodule: previously no support for custom certs for git dependencies, now sets `GIT_SSL_CAINFO` to the value of `NIX_SSL_CERT_FILE` using whatever's provided by the user / setup hook

Looking through other fetchers, it seems most of them either have their own schemes or delegate to fetchurl. So this isn't a wide-reaching code change, but I think it's important to support `NIX_SSL_CERT_FILE` as an impureEnvVar, so that we don't force fetchers to invent their own envvars for this purpose.

## Things done

Since this is a mass-rebuild, I tested locally by only applying these changes to the version of `cacert` used by fetchers. With these modifications I tested fetching over https (using custom certs) with:
 - fetchurl
 - fetchgit (over https)
 - go modules fetch (git https)  

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

### DIFF
--- a/pkgs/build-support/go/module.nix
+++ b/pkgs/build-support/go/module.nix
@@ -136,6 +136,7 @@ in
         exit 10
       fi
 
+      export GIT_SSL_CAINFO=$NIX_SSL_CERT_FILE
       ${if finalAttrs.proxyVendor then ''
         mkdir -p "''${GOPATH}/pkg/mod/cache/download"
         go mod download
--- a/pkgs/data/misc/cacert/setup-hook.sh
+++ b/pkgs/data/misc/cacert/setup-hook.sh
@@ -1,7 +1,7 @@
-export NIX_SSL_CERT_FILE=@out@/etc/ssl/certs/ca-bundle.crt
+export NIX_SSL_CERT_FILE="${NIX_SSL_CERT_FILE:-@out@/etc/ssl/certs/ca-bundle.crt}"
 
 # compatibility
 #  - openssl
-export SSL_CERT_FILE=@out@/etc/ssl/certs/ca-bundle.crt
+export SSL_CERT_FILE=$NIX_SSL_CERT_FILE
 #  - Haskell x509-system
-export SYSTEM_CERTIFICATE_PATH=@out@/etc/ssl/certs/ca-bundle.crt
+export SYSTEM_CERTIFICATE_PATH=$NIX_SSL_CERT_FILE
--- a/doc/build-helpers/fetchers.chapter.md
+++ b/doc/build-helpers/fetchers.chapter.md
@@ -157,6 +157,12 @@ Here are security considerations for this scenario:
 
   In more concrete terms, if you use any other hash, the [`--insecure` flag](https://curl.se/docs/manpage.html#-k) will be passed to the underlying call to `curl` when downloading content.
 
+## Proxy usage {#sec-pkgs-fetchers-proxy}
+
+Nixpkgs fetchers can make use of a http(s) proxy. Each fetcher will automatically inherit proxy-related environment variables (`http_proxy`, `https_proxy`, etc) via [impureEnvVars](https://nixos.org/manual/nix/stable/language/advanced-attributes#adv-attr-impureEnvVars).
+
+The environment variable `NIX_SSL_CERT_FILE` is also inherited in fetchers, and can be used to provide a custom certificate bundle to fetchers. This is usually required for a https proxy to work without certificate validation errors.
+
 []{#fetchurl}
 ## `fetchurl` {#sec-pkgs-fetchers-fetchurl}
 
--- a/lib/fetchers.nix
+++ b/lib/fetchers.nix
@@ -9,6 +9,9 @@
     # by definition pure.
     "http_proxy" "https_proxy" "ftp_proxy" "all_proxy" "no_proxy"
     "HTTP_PROXY" "HTTPS_PROXY" "FTP_PROXY" "ALL_PROXY" "NO_PROXY"
+
+    # https proxies typically need to inject custom root CAs too
+    "NIX_SSL_CERT_FILE"
   ];
 
 }
--- a/pkgs/build-support/fetchgit/default.nix
+++ b/pkgs/build-support/fetchgit/default.nix
@@ -67,7 +67,7 @@ stdenvNoCC.mkDerivation {
   builder = ./builder.sh;
   fetcher = ./nix-prefetch-git;
 
-  nativeBuildInputs = [ git ]
+  nativeBuildInputs = [ git cacert ]
     ++ lib.optionals fetchLFS [ git-lfs ];
 
   outputHashAlgo = if hash != "" then null else "sha256";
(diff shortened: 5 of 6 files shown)


timbertson:
@roberth would you mind taking another look at this now that I've rebased?
(I had to close the old PR which you commented on previously, https://github.com/NixOS/nixpkgs/pull/271161#issuecomment-2049183843)

pkgs/build-support/go/module.nix
```diff
@@ -114,6 +114,7 @@ let
         exit 10
       fi
 
+      export GIT_SSL_CAINFO="''${NIX_SSL_CERT_FILE}"
```
3noch, line 117:
```suggestion
      export GIT_SSL_CAINFO=$NIX_SSL_CERT_FILE
```

pkgs/data/misc/cacert/setup-hook.sh
```diff
@@ -1,7 +1,7 @@
-export NIX_SSL_CERT_FILE=@out@/etc/ssl/certs/ca-bundle.crt
+export NIX_SSL_CERT_FILE="${NIX_SSL_CERT_FILE:-@out@/etc/ssl/certs/ca-bundle.crt}"
 
 # compatibility
 #  - openssl
-export SSL_CERT_FILE=@out@/etc/ssl/certs/ca-bundle.crt
+export SSL_CERT_FILE="$NIX_SSL_CERT_FILE"
 #  - Haskell x509-system
-export SYSTEM_CERTIFICATE_PATH=@out@/etc/ssl/certs/ca-bundle.crt
+export SYSTEM_CERTIFICATE_PATH="$NIX_SSL_CERT_FILE"
```
3noch, line 7:
```suggestion
export SYSTEM_CERTIFICATE_PATH=$NIX_SSL_CERT_FILE
```

pkgs/data/misc/cacert/setup-hook.sh
```diff
@@ -1,7 +1,7 @@
-export NIX_SSL_CERT_FILE=@out@/etc/ssl/certs/ca-bundle.crt
+export NIX_SSL_CERT_FILE="${NIX_SSL_CERT_FILE:-@out@/etc/ssl/certs/ca-bundle.crt}"
 
 # compatibility
 #  - openssl
-export SSL_CERT_FILE=@out@/etc/ssl/certs/ca-bundle.crt
+export SSL_CERT_FILE="$NIX_SSL_CERT_FILE"
```
3noch, line 5:
```suggestion
export SSL_CERT_FILE=$NIX_SSL_CERT_FILE
```

3noch (changes requested):

timbertson:
Suggestions applied and rebased, thanks for the review @3noch

Quince-Pie (approved):

timbertson:
Rebased again. I'd love to get this PR merged if anyone knows how to get any attention to it. It's been reviewed by a few people since I raised it nearly a year ago (as #271161), but I don't know who has authority to actually merge it. It is a bulk-rebuild, but it's also pretty small and should be a no-op in the typical case of "not using an HTTPS proxy".

lukegb (approved):
This change seems reasonable - there's some potential for slightly wonky things happening inside a nix-shell, but I have no serious objection and I can't think of any particular circumstances in which that would make much of a difference.

pkgs/data/misc/cacert/setup-hook.sh
```diff
@@ -1,7 +1,7 @@
-export NIX_SSL_CERT_FILE=@out@/etc/ssl/certs/ca-bundle.crt
+export NIX_SSL_CERT_FILE="${NIX_SSL_CERT_FILE:-@out@/etc/ssl/certs/ca-bundle.crt}"
```
arcnmx, line 1: Would it make sense to check that the path in the env var is readable first before prioritizing its value over the built one?
I believe this change requires it to now be included in `extra-sandbox-paths`, which is unlikely to be the case in existing deployments using `NIX_SSL_CERT_FILE` prior to this change - and thus will suddenly point to something that doesn't exist as far as the builder is concerned.

timbertson: > Would it make sense to check that the path in the env var is readable first before prioritizing its value over the built one?

Maybe? It wouldn't hurt, but I'm also surprised that you'd need to.

> I believe this change requires it to now be included in extra-sandbox-paths, which is unlikely to be the case in existing deployments using NIX_SSL_CERT_FILE prior to this change

For build-daemon use cases, I've needed to work hard to set NIX_SSL_CERT_FILE. Either add it to the builder's env intentionally, or use the [`inpure-env` config](https://nix.dev/manual/nix/2.22/command-ref/conf-file#conf-impure-env). So I assumed it wouldn't be accidentally set.

Maybe this is more common in a nixos deployment? It might be good to understand your use case and what happens after these changes, to confirm whether the right fix is "ignore NIX_SSL_CERT_FILE when it's not readable", or "don't do that thing you're doing". Maybe raise an issue and link it here for better visibility?

PR #303365, opened 2024-04-11
vscode-extensions: add documentation and move extensions with parameters to a directory

superherointj:
vscode-extensions:

1. Document the criteria I have been using to review vscode extensions.

2. Avoid to keep adding new extensions with new parameters at the global index, because these parameters don't get removed whenever the extension gets updated/removed. It is extra work to do that kind of checking and it also gets messy as the number of packages grow. Doing this facilitates upkeep and avoids trivial feedback on reviews.

CC @deviant @jonringer

### DIFF
--- a/pkgs/applications/editors/vscode/extensions/README.md
+++ b/pkgs/applications/editors/vscode/extensions/README.md
@@ -1,0 +1,37 @@
+# Visual Studio Code Extensions
+
+## Conventions for adding new extensions
+
+* Extensions are named in the **lowercase** version of the extension's unique identifier. Which is found on the marketplace extension page, and is the name under which the extension is installed by VSCode under `~/.vscode`.
+  Extension location should be: ${lib.strings.toLower mktplcRef.publisher}.${lib.string.toLower mktplcRef.name}
+
+* Move extension to a discrete directory whenever the extension needs extra parameters/packages (at top of the file) or other files (such as patches, update script, components). Global index file parameters/packages should be utilities shared by many extensions. Extension specific parameters/packages should not be in the global index page.
+
+* Currently `nixfmt-rfc-style` formatter is being used to format the VSCode extensions.
+
+* Respect `alphabetical order` whenever adding extensions. On disorder, please, kindly open a PR re-establishing the order.
+
+* Avoid [unnecessary](https://nix.dev/guides/best-practices.html#with-scopes) use of `with`, particularly `nested with`.
+
+* Use `hash` instead of `sha256`.
+
+* On `meta` field:
+  - add a `changelog`.
+  - `description` should mention it is a Visual Studio Code extension.
+  - `downloadPage` is the VSCode marketplace URL.
+  - `homepage` is the source-code URL.
+  - verify `license` in upstream.
+
+* On commit messages:
+  - Naming convention for:
+    - Adding a new extension:
+
+      > vscode-extensions.publisher.extension-name: init 1.2.3
+      >
+      > Release: https://github.com/owner/project/releases/tag/1.2.3
+    - Updating an extension:
+
+      > vscode-extensions.publisher.extension-name: 1.2.3 -> 2.3.4
+      >
+      > Release: https://github.com/owner/project/releases/tag/2.3.4
+  - Multiple extensions can be added in a single PR, but each extension requires it's own commit.
--- a/pkgs/applications/editors/vscode/extensions/default.nix
+++ b/pkgs/applications/editors/vscode/extensions/default.nix
@@ -9,7 +9,6 @@
 , jdk
 , llvmPackages
 , llvmPackages_14
-, nixpkgs-fmt
 , protobuf
 , jq
 , shellcheck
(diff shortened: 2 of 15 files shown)


pkgs/applications/editors/vscode/extensions/default.nix
deviant, line 1: I would add a comment at the top of this file referring to the documentation and summarizing the criteria. I think it's likely that people won't become aware of the convention, otherwise. Repeating the most important information here is nice because it reduces the effort required to find the criteria, making it less likely that people will inadvertently miss reading about them if they skip reading the documentation.

pkgs/applications/editors/vscode/extensions/README.md
```diff
@@ -0,0 +1,14 @@
+# Visual Studio Code Extensions
+
+## Conventions for adding new extensions
+
+* Extensions are named in the **lowercase** version of the extension's unique identifier. Which is found on the marketplace extension page, and is the name under which the extension is installed by VSCode under `~/.vscode`.
+  Extension location should be: ${lib.strings.toLower mktplcRef.publisher}.${lib.string.toLower mktplcRef.name}
+
+* Move extension to a discrete directory whenever the extension needs `extra parameters` and other files (such as `patches` or `components`). The parameters that exist in the global index file are utilities shared by many extensions. Parameters that are specific to an extension or that are not globally used shouldn't be added to the global index page.
+
+* Currently `nixfmt-rfc-style` formatter is being used to format the VSCode extensions.
```
deviant, line 10: I don't know much about `nixfmt-rfc-style`, is this a new standard formatter being used everywhere, or is this just something you're using? If there isn't consensus on this in general. I'd probably drop it.

pkgs/applications/editors/vscode/extensions/README.md
```diff
@@ -0,0 +1,14 @@
+# Visual Studio Code Extensions
+
+## Conventions for adding new extensions
+
+* Extensions are named in the **lowercase** version of the extension's unique identifier. Which is found on the marketplace extension page, and is the name under which the extension is installed by VSCode under `~/.vscode`.
+  Extension location should be: ${lib.strings.toLower mktplcRef.publisher}.${lib.string.toLower mktplcRef.name}
+
+* Move extension to a discrete directory whenever the extension needs `extra parameters` and other files (such as `patches` or `components`). The parameters that exist in the global index file are utilities shared by many extensions. Parameters that are specific to an extension or that are not globally used shouldn't be added to the global index page.
```
deviant, line 8: I would clarify what is meant by "extra parameters"; perhaps instead say "if it would require additional parameters to be added at the top of the file". I don't think it's *unambiguous* per se, since there's only really one place in the file where there are parameters, but the current wording leads me (and I imagine will lead other readers also) to think more about the extension expression itself, which will probably be situated partway through the file.

deviant (changes requested):
Thanks for this! It looks good so far.

pkgs/applications/editors/vscode/extensions/README.md
superherointj: https://github.com/NixOS/rfcs/pull/166

pkgs/applications/editors/vscode/extensions/default.nix
superherointj: I considered adding the comment to the main file. But I couldn't make it much shorter (or nice). And then, I realized that I was duplicating information and ideally it should be a single source of truth. So instead, I have replaced the instructions on how to add extension with a reference for the documentation on adding the extension. So that, whenever the person considers adding an extension, the hint for instructions is there. It avoids mixing code and documentation in a badly manner. Let me know if you think it's good enough. Or if you want changes.

I wonder if I should have added the full URL for README.

pkgs/applications/editors/vscode/extensions/README.md
superherointj: I have reworded it. Feel free to provide further feedback.

ratsclub (approved):
The extensions I use worked just fine with this change and I like the structure.

superherointj:
Fixed a conflict. This sort of change conflicts often. Would be nice to get this merged sooner. But then, it needs a review. As I made changes, it requires another round of reviews. Thank you.

drupol (approved):
LGTM, I don't like the idea that `default.nix` _MUST_ be edited when adding a new extension in its own directory.
Wouldn't it be possible to make something similar to `pkgs/by-name` instead?

superherointj:
> LGTM, I don't like the idea that `default.nix` _MUST_ be edited when adding a new extension in its own directory. Wouldn't it be possible to make something similar to `pkgs/by-name` instead?

Makes sense.

I don't have this solution atm. What I have done here is to move individual extensions that have extension specific arguments/packages to a discrete directory. Which could be a step helpful in that direction. I suggest we merge this as it is for the moment. And then, consider a new PR for automatically inferring packages from directory structure.

drupol:
Fine. Can you call `@ofborg build <name> <name>` for each modified extension just to check if it still builds fine?

superherointj:
@ofborg build vscode-extensions.ms-python.vscode-pylance vscode-extensions.betterthantomorrow.calva vscode-extensions.jackmacwindows.craftos-pc vscode-extensions.azdavis.millet vscode-extensions.eugleo.magic-racket vscode-extensions.asciidoctor.asciidoctor-vscode vscode-extensions.myriad-dreamin.tinymist vscode-extensions.kamadorueda.alejandra vscode-extensions.foxundermoon.shell-format vscode-extensions.mgt19937.typst-preview vscode-extensions.nvarner.typst-lsp vscode-extensions.b4dm4n.vscode-nixpkgs-fmt vscode-extensions.timonwong.shellcheck

PR #303388, opened 2024-04-11
nixos/paperless: override enabled tesseract languages with the in paperless configured ones

SuperSandro2000:
## Description of changes

Reduces required closure size dramatically when only a few languages are required.

```
all: ε → ∅, -1039163.4 KiB
deu.traineddata: ∅ → ε, +15075.8 KiB
eng.traineddata: ∅ → ε, +22916.8 KiB
equ.traineddata: ∅ → ε, +2199.3 KiB
osd.traineddata: ∅ → ε, +10315.4 KiB
tesseract: -31.5 KiB
```

## Things done

Tested with and without PAPERLESS_OCR_LANGUAGE being configured.

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [x] Tested, as applicable:

---

### DIFF
--- a/nixos/modules/services/misc/paperless.nix
+++ b/nixos/modules/services/misc/paperless.nix
@@ -3,7 +3,6 @@
 with lib;
 let
   cfg = config.services.paperless;
-  pkg = cfg.package;
 
   defaultUser = "paperless";
   defaultFont = "${pkgs.liberation_ttf}/share/fonts/truetype/LiberationSerif-Regular.ttf";
@@ -25,7 +24,7 @@ let
   } // optionalAttrs (cfg.settings.PAPERLESS_ENABLE_NLTK or true) {
     PAPERLESS_NLTK_DIR = pkgs.symlinkJoin {
       name = "paperless_ngx_nltk_data";
-      paths = pkg.nltkData;
+      paths = cfg.package.nltkData;
     };
   } // optionalAttrs (cfg.openMPThreadingWorkaround) {
     OMP_NUM_THREADS = "1";
@@ -38,7 +37,7 @@ let
   manage = pkgs.writeShellScript "manage" ''
     set -o allexport # Export the following env vars
     ${lib.toShellVars env}
-    exec ${pkg}/bin/paperless-ngx "$@"
+    exec ${cfg.package}/bin/paperless-ngx "$@"
   '';
 
   # Secure the services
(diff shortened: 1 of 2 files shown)


leona-ya (approved):
This seems quite nice. Still, I would like someone else to also have a look at it.

pkgs/applications/office/paperless-ngx/default.nix
```diff
@@ -37,17 +39,21 @@ let
...
+        uvicorn = prev.uvicorn.overridePythonAttrs (_: {
+          version = "0.25.0";
+          src = fetchFromGitHub {
+            owner = "encode";
+            repo = "uvicorn";
+            rev = "0.25.0";
+            hash = "sha256-ng98DTw49zyFjrPnEwfnPfONyjKKZYuLl0qduxSppYk=";
+          };
+        });
+      })
```
ambroisie, line 55: Should we perhaps switch the order, _just in case_ someone wants to override `uvicorn`?

ambroisie (approved):
LGTM.

erikarvstedt (changes requested):
Why not simply override `ocrmypdf` in the paperless pkg, [like so](https://github.com/erikarvstedt/nixpkgs/commit/3d804ec78ce86a91e7d96410cc65d58c88100104)?

Maybe keep the `pkg = cfg.package`  variable. It shortens some complex expressions, also the diff from this PR gets much smaller. Medium preference.

leona-ya:
> Why not simply override `ocrmypdf` in the paperless pkg, [like so](https://github.com/erikarvstedt/nixpkgs/commit/3d804ec78ce86a91e7d96410cc65d58c88100104)?

This should not produce the same result, as in Sandro's version `ocrmypdf.override { tesseract = tesseract5; }` is using the tesseract defined above in the rec clause, while your version just uses the normal input

erikarvstedt:
No, my version is equivalent.
The `tesseract5` input in the `paperless-ngx` pkg is exactly the customized tesseract pkg from the rec attrs.

leona-ya:
Oh my bad, one more override than I thought about, sorry :/

SuperSandro2000:
Moved the overwrite like in the linked patch. Thanks!

leona-ya (approved):

nixos/modules/services/misc/paperless.nix
```diff
@@ -200,7 +199,17 @@ in
...
+      apply = pkg: pkg.override {
+        tesseract5 = pkg.tesseract5.override {
+          # always enable detection modules
+          enableLanguages = if cfg.settings ? PAPERLESS_OCR_LANGUAGE then
+            [ "equ" "osd" ]
+              ++ lib.splitString "+" cfg.settings.PAPERLESS_OCR_LANGUAGE
+          else null;
+        };
+      };
+    };
```
mweinelt, line 212: This breaks overridePythonAttrs used on the package.

SuperSandro2000: What should then be used instead? When I last looked at this, it wasn't easily possible to do an override in the package for the python package set and here.

mweinelt: It could be an override on the default package, that users need to carry over.

SuperSandro2000: Since the override is a bit more complicated, I am not sure that this is the best option.

PR #303393, opened 2024-04-11
sparrow : 1.8.4 -> 1.9.0

emmanuelrosa:
## Description of changes

This PR contains three commits:

1. Version bump of sparrow from 1.8.4 to 1.9.0.
2. Migration from stdenv to stdenvNoCC
3. GPG verification of upstream tarball package.
4. Removal of the Nix package updater.

### Version bump

The 1.8.5 version bump is trivial. The only thing which is not obvious is that upstream updated to HWI 3.0.0. This Nix package doesn't use the bundled HWI from upstream, and instead uses HWI from Nixpkgs. However, this PR applies the commits on a version of Nixpkgs which does include HWI 3.0.0, therefore it means the application's expectations.

Version 1.9.0 removes the Samurai Whirlpool functionality.

Release notes:

- https://github.com/sparrowwallet/sparrow/releases/tag/1.8.5
- https://github.com/sparrowwallet/sparrow/releases/tag/1.9.0

### stdenvNoCC

This package contains a number of derivations, yet only one of them needs a C compiler. And that single derivation needs a library which comes with the compiler, not the compiler itself. Therefore, the 2nd commit replaces `stdenv` with `stdenvNoCC`, except for that one instance mentioned.

### GPG verification

It's quite common for developers of Bitcoin-related applications to provide GPG signatures of their releases. Up to this point, the updater script for this package has had GPG verification.  But a downside to verification via an update script is that there's no proof that the script was used to update the package.

The 3rd commit in this PR adds GPG verification within the use of `fetchurl`, leading to build-time verification. In short, if the Nix package builds, then it's also GPG verified. 

The verification provided by the update script differs from the build-time verification in the fact that the updater gets the signer's public key from a URL, while the build-time verification uses a copy of the signer's key included in this repo (added by this PR).

If the signer's public key ever changes, then it could lead to a situation where the updater successfully produces a verified update, but the build fails because the build-time verification will attempt to verify with the wrong public key. Such a situation will require my manual intervention to resolve.

### Removal of package updater

I removed the package updater because GPG verification requires additional hashes to be updated, and the updater won't be able to do that.

## Things done

- Built on platform(s)
  - [X] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---


RaghavSood:
Result of `nixpkgs-review pr 303393` run on aarch64-darwin [1](https://github.com/Mic92/nixpkgs-review)

RaghavSood:
Well, that wasn't helpful - let me try on a linux box shortly

cbergqvist (approved):
ACK de1719e41a719e4043fefa4eb370f5677d632e79!

Tested by checking out @emmanuelrosa's branch and applying to own system, running Sparrow and verifying that the about dialog reads 1.9.0.

Used `gpg --import pkgs/applications/blockchains/sparrow/publickey.asc` to check key file and confirmed that it is from the upstream/application developer, Craig Raw.

(Was actually attempting to run the previous update.sh today to see if I could update locally, but ran into issues. Great to see there was already a PR considering the Samourai situation).

RaghavSood (approved):

fossadmin:
Is 1.9.0 available? I'm running NixOS unstable with flakes and it still shows as 1.8.4. Still kind of new to NixOS so not sure if I need to run some kind of update or something. I did sudo nixos-rebuild switch --flake . but nothing changed after the rebuild. 
Thanks!

emmanuelrosa:
It's available on the master branch, but it probably hasn't made it to the unstable branch yet.

RaghavSood:
It should be on unstable now: https://status.nixos.org/

You may need up update your flake.lock and make sure your nixpkgs input points to unstable

fossadmin:
Cool, thanks!

PR #303442, opened 2024-04-11
nixos/gns3-server: fix ubridge support

### LINKED ISSUE #292258: nixos/gns3-server: ubridge doesn't work
### Describe the bug
When I try to start docker appliences I get the following error:
```
uBridge requires root access or the capability to interact with network adapters
```

I viewed module source, and noticied that `ubridge_path` is set to `/nix/store/...` package, rather than `/run/wrappers/bin/ubridge` which gives those capabilities.

I have created PR (#292095) to fix this. But then run into another problem, it finds the wrapper, but fails when it tries to verify version with an error.
```
Could not determine uBridge version for /run/wrappers/bin/ubridge
```

After few patches to `gns3-server` code, I got a more informative error:
```
Could not determine uBridge version, error: failed to inherit capabilities: Operation not permitted
```

I have searched for this error, and it originates from [nixos/modules/security/wrappers/wrapper.c](https://github.com/NixOS/nixpkgs/blob/9f21aa90cb8c078969543956d88c19696b646743/nixos/modules/security/wrappers/wrapper.c#L110)

Unfortunatly at this point of diagnostic my knowledge is lacking with respect to capabilites, and C.

Also `gns3-server` package works fine when started as a user with `ubridge` group (with the wrapper in `ubridge_path`). It looks like it has something to do with systemd, systemd service configuration or hardening. I tried commenting out the later, but it didn't change much.

### Steps To Reproduce
1. Enable `services.gns3-server` and `services.gns3-server.ubridge`
2. Create any Docker applience
3. Try to start Docker applience

### Expected behavior
Docker applience start without errors

### Notify maintainers

@anthonyroussel

### Metadata

```console
 - system: `"x86_64-linux"`
 - host os: `Linux 6.1.79, NixOS, 24.05 (Uakari), 24.05.20240226.13aff9b`
 - multi-user?: `yes`
 - sandbox: `yes`
 - channels(dsuetin): `""`
 - nixpkgs: `/nix/store/1gpgihsi21bni03hy6i5ccw9lfrhnbqi-source`
```

---

anthonyroussel:
## Description of changes

Fixes #292258
See [this comment](https://github.com/NixOS/nixpkgs/issues/292258#issuecomment-2050392583) for explanations.

The usage of SUID wrappers is incompatible with SystemD DynamicUser & hardenings.
And GNS3 needs to call ubridge via its SUID Wrapper to work.

:warning: These changes are not backwards compatible, but are needed to fix the module. You will need to run `chown -R gns3:gns3 /var/lib/gns3 /etc/gns3 /var/log/gns3` to fix the permissions (see release notes).

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [x] Tested, as applicable:

---

### DIFF
--- a/nixos/modules/services/networking/gns3-server.nix
+++ b/nixos/modules/services/networking/gns3-server.nix
@@ -150,7 +150,7 @@ in {
         };
       }
       (lib.mkIf (cfg.ubridge.enable) {
-        Server.ubridge_path = lib.mkDefault (lib.getExe cfg.ubridge.package);
+        Server.ubridge_path = lib.mkDefault "/run/wrappers/bin/ubridge";
       })
       (lib.mkIf (cfg.auth.enable) {
         Server = {
--- a/nixos/modules/services/networking/gns3-server.nix
+++ b/nixos/modules/services/networking/gns3-server.nix
@@ -233,14 +233,27 @@ in {
         User = "gns3";
         WorkingDirectory = "%S/gns3";
 
+        # Required for ubridge integration to work
+        #
+        # GNS3 needs to run SUID binaries (ubridge)
+        # but NoNewPrivileges breaks execution of SUID binaries
+        DynamicUser = false;
+        NoNewPrivileges = false;
+        RestrictSUIDSGID = false;
+        PrivateUsers = false;
+
         # Hardening
-        DeviceAllow = lib.optional flags.enableLibvirtd "/dev/kvm";
+        DeviceAllow = [
+          # ubridge needs access to tun/tap devices
+          "/dev/net/tap rw"
+          "/dev/net/tun rw"
+        ] ++ lib.optionals flags.enableLibvirtd [
+          "/dev/kvm"
+        ];
         DevicePolicy = "closed";
         LockPersonality = true;
         MemoryDenyWriteExecute = true;
-        NoNewPrivileges = true;
         PrivateTmp = true;
-        PrivateUsers = true;
         # Don't restrict ProcSubset because python3Packages.psutil requires read access to /proc/stat
         # ProcSubset = "pid";
         ProtectClock = true;
@@ -261,8 +274,7 @@ in {
         ];
         RestrictNamespaces = true;
         RestrictRealtime = true;
-        RestrictSUIDSGID = true;
-        UMask = "0077";
+        UMask = "0022";
       };
     };
   };
--- a/nixos/modules/services/networking/gns3-server.nix
+++ b/nixos/modules/services/networking/gns3-server.nix
@@ -129,7 +129,14 @@ in {
       }
     ];
 
+    users.groups.gns3 = { };
+
     users.groups.ubridge = lib.mkIf cfg.ubridge.enable { };
+
+    users.users.gns3 = {
+      group = "gns3";
+      isSystemUser = true;
+    };
 
     security.wrappers.ubridge = lib.mkIf cfg.ubridge.enable {
       capabilities = "cap_net_raw,cap_net_admin=eip";
@@ -206,7 +213,6 @@ in {
       serviceConfig = {
         ConfigurationDirectory = "gns3";
         ConfigurationDirectoryMode = "0750";
-        DynamicUser = true;
         Environment = "HOME=%S/gns3";
         ExecReload = "${pkgs.coreutils}/bin/kill -HUP $MAINPID";
         ExecStart = "${lib.getExe cfg.package} ${commandArgs}";
--- a/nixos/doc/manual/release-notes/rl-2411.section.md
+++ b/nixos/doc/manual/release-notes/rl-2411.section.md
@@ -298,6 +298,15 @@
   a static `user` and `group`. The `writablePaths` option has been removed and
   the models directory is now always exempt from sandboxing.
 
+- The `gns3-server` service now runs under the `gns3` system user
+  instead of a dynamically created one via `DynamicUser`.
+  The use of SUID wrappers is incompatible with SystemD's `DynamicUser` setting,
+  and GNS3 requires calling ubridge through its SUID wrapper to function properly.
+  This change requires to manually move the following directories:
+    * from `/var/lib/private/gns3` to `/var/lib/gns3`
+    * from `/var/log/private/gns3` to `/var/log/gns3`
+  and to change the ownership of these directories and their contents to `gns3` (including `/etc/gns3`).
+
 - Legacy package `stalwart-mail_0_6` was dropped, please note the
   [manual upgrade process](https://github.com/stalwartlabs/mail-server/blob/main/UPGRADING.md)
   before changing the package to `pkgs.stalwart-mail` in


anthonyroussel:
@ofborg test gns3-server

dani0854:
I am still getting
```
Could not start auxiliary console process: [Errno 2] No such file or directory: 'script'
```

Seems to be this line of code:
https://github.com/GNS3/gns3-server/blob/5bab4131e186fbaa08943987fa9f7deb6a409642/gns3server/compute/docker/docker_vm.py#L530C1-L530C25

It looks like GNS3 depends on a `script` util, not included in NixOS by default.
[unixtools.script](https://search.nixos.org/packages?channel=23.11&from=0&size=50&sort=relevance&type=packages&query=script) provides that util (as well as some others), but simply adding it to system environment doesn't seem to work, not sure why.

I guess it might require a patch to the package in order to work. Or it has to be available in the environment.
There are also some other commands it depends on (not a complete list):
* docker
* telnet
* vncviewer
* remote-viewer (spice)

dani0854 (approved):
Apart from above, those changes do fix the ubridge problem in the issue.

dani0854:
Unfortunately there seemed to many layers of problems hidden under each other, and it might be quite tricky to get GNS3 fully working with docker on NixOS.

Perhaps a more general issue to get GNS3 to work with docker is needed.

anthonyroussel:
> Unfortunately there seemed to many layers of problems hidden under each other, and it might be quite tricky to get GNS3 fully working with docker on NixOS.
> 

@dani0854 Thanks for the review.

I made a PR to fix the "Auxiliary console not available" error: https://github.com/NixOS/nixpkgs/pull/303472

nixos/modules/services/networking/gns3-server.nix
```diff
@@ -226,37 +232,6 @@ in {
           ++ lib.optional cfg.ubridge.enable "ubridge";
         User = "gns3";
         WorkingDirectory = "%S/gns3";
-
-        # Hardening
```
anthonyroussel, line 230: Add a comment stating that these SystemD hardenings are incompatible with GNS3 (DynamicUsers, and NoNewPrivileges)

SuperSandro2000:
Please implement the above suggestion

lugeha:
I just tried this in a container too,

It also requires PrivateUsers = lib.mkForce false;

anthonyroussel:
```
DeviceAllow = [
  "/dev/null rw"
  "/dev/net/tun rw"
  "/dev/net/tap rw"
] ++ lib.optionals flags.enableLibvirtd [
  "/dev/kvm"
];
```

is also required for ubridge to work (nio tap devices).

anthonyroussel:
Unfortunately, these changes are not backwards compatible.
And there is no official way within SystemD to revert the DynamicUser setting.

Here are the commands I use to migrate the GNS3 state files out of /var/lib/private:

```shell
rm /var/lib/gns3
mv /var/lib/private/gns3 /var/lib/gns3
chown -R gns3:gns3 /var/lib/gns3

rm /var/log/gns3
mv /var/log/private/gns3 /var/log/gns3
chown -R gns3:gns3 /var/log/gns3

chown -R gns3:gns3 /etc/gns3
```

haras-unicorn:
I was trying to run some docker services with this thing and came to the conclusion that UMask needs to be "0022" or something along those lines.
The reason is that if you try to run a docker service that runs via a user other than root that they don't have the permissions to start. The GNS3 init script tries to run another one of the scripts as that user via su. The other script is put there via a bind to a subdir in the GNS3 working directory. If UMask is 0077 as it is currently that other script doesn't have permissions so that the user declared in the docker image can execute it and the docker image entrypoint can never start.
I hope this all makes sense. I added this to the config I posted above if anyone wants to get around this:
```nix
      UMask = pkgs.lib.mkForce "0022";
```

anthonyroussel:
@haras-unicorn Thanks for the review. I changed UMask to its default value (0022).

TafkaMax:
Will this be taken into account for the 24.11 release?

nixos/modules/services/networking/gns3-server.nix
```diff
@@ -227,14 +233,25 @@ in {
...
+        # but NoNewPrivileges breaks execution of SUID binaries
+        DynamicUser = lib.mkForce false;
+        NoNewPrivileges = lib.mkForce false;
+        RestrictSUIDSGID = lib.mkForce false;
+        PrivateUsers = lib.mkForce false;
+
         # Hardening
-        DeviceAllow = lib.optional flags.enableLibvirtd "/dev/kvm";
+        DeviceAllow = [
+          "/dev/net/tap rw" # ubridge needs access to tap devices
```
dani0854, line 247:
```suggestion
          # ubridge needs access to tap and tun devices
          "/dev/net/tap rw"
          "/dev/net/tun rw"
```
Looks like you need tun as well, otherwise I still get the following error:
```
Error while sending command 'bridge add_nio_tap bridge0 tap-gns3-e0': unable to create NIO TAP for bridge 'bridge0': uBridge version 0.9.18 running with libpcap version 1.10.4 (with TPACKET_V3)
Hypervisor TCP control server started (IP 0.0.0.0 port 37651).
create_nio_tap: unable to open TAP device tap-gns3-e0 (Operation not permitted)
```

Interesting enough, on my device it still works with only tun, and without tap.
Either the error message is wrong, or something else is going on behind scene, my networking knowledge is not deep enough.

anthonyroussel: @dani0854 Thank you for the review!

Oops sorry, I forgot to force push the commit adding `/dev/net/tun rw` to `DeviceAllow`.

`/dev/net/tun rw` is required for ubridge: https://github.com/GNS3/ubridge/blob/2f7798c48e718b44e210d16cf0592dbaa085e9be/src/nio_tap.c#L67

anthonyroussel:
* Rebased the PR against master branch
* Also added some release notes stating that this is a breaking change required to fix the GNS3 module

nixos/modules/services/networking/gns3-server.nix
```diff
@@ -227,14 +233,27 @@ in {
...
         WorkingDirectory = "%S/gns3";
 
+        # Required for ubridge integration to work
+        #
+        # GNS3 needs to run SUID binaries (ubridge)
+        # but NoNewPrivileges breaks execution of SUID binaries
+        DynamicUser = lib.mkForce false;
+        NoNewPrivileges = lib.mkForce false;
+        RestrictSUIDSGID = lib.mkForce false;
+        PrivateUsers = lib.mkForce false;
```
Aleksanaa, line 243: Why are these using `lib.mkForce`? If users have to change them to adapt their needs, do they have to use `lib.mkVmOverride`? Just `false` would be okay.

anthonyroussel: Thank you for your review! :)

We set this to false because GNS3 is compatible with these hardening flags.

I changed these flags to:

```suggestion
        DynamicUser = false;
        NoNewPrivileges = false;
        RestrictSUIDSGID = false;
        PrivateUsers = false;
```

dani0854 (approved):
Ubridge seems to work correctly now as far I have tested.

Although there are many other problems I have discovered with docker based appliances, but I think it's better to merge this PR, and start a separate issue/pr for those problems.

PR #303682, opened 2024-04-12
imgui: ship vcpkg' (cmake) distribution info

SomeoneSerge:
## Description of changes

"Dear ImGui" is one instance of a project where upstream chooses to only distribute sources and headers without any metadata related to building or packaging (cmake "modules", pkg-config, ...). In practice, building and managing the dependencies of imgui, or including imgui in projects built by common build tools (cmake, meson, ...) are not trivial. One way to consume it, besides vendoring, is via the cmake rules maintained in a centralized repository like vcpkg (conan, ...). Presumably, it's even a quite common method.

A wider adoption of the vcpkg-like approach by upstream projects would be ideal for Nixpkgs purposes: it establishes a common interface and a dependency injection mechanism that help creating consistent package sets, fix points, modeling dependencies as independent derivations, etc. This common interface is in fact pure cmake and vcpkg-agnostic. Possibly worth a mention that vcpkg is licensed under MIT.

I suggest we try out the practice of distributing vcpkg' packaging info in situations where upstream doesn't provide any.

In this PR I adapt `imgui` and related packages to reuse `vcpkg`' `CMakeLists.txt` instead of custom phases.

Affected packages:
- glfw3
- imgui
  - implot
  - imnodes

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---


pkgs/by-name/vc/vcpkg/package.nix
```diff
@@ -16,6 +17,14 @@ stdenvNoCC.mkDerivation (finalAttrs: {
     hash = "sha256-HMK3sebq/9TuxHQ75+5UIMvN09cPWmq7TFBBwRY4X7o=";
   };
 
+  patches = [
```
gracicot, line 20: I don't think applying patch to vcpkg ports are a good idea.

First, I'm not even sure this patch will hit mainstream and I want to keep this package as vanilla as possible. 

Second, those patches can be applied by user via [overlay ports](https://learn.microsoft.com/en-us/vcpkg/concepts/overlay-ports). If those patch are needed by one of your project, no need to patch the official package, just add it to your git until maintain accepts the patch.

Unless you have a good argument why we should patch vcpkg ports (and therefore breaking the built-in baseline) I would abstain to diverge from upstream.

SomeoneSerge: Uh-huh, the intent here actually wasn't to ship these patches to vcpkg' users, but to imgui users. The direct implementation would've been to `applyPatches` in `imgui/default.nix`, but the more scalable solution is to create a fixpoint. I used the `vcpkg` package as one.

I'll revert the vcpkg change

SomeoneSerge: I updated the description and I no longer touch `pkgs.vcpkg`. There's no need to synchronize any patches yet

gracicot: If those changes benefit users of imgui even for those who use vcpkg as their package managers, I would recommend also sending the patch to upstream vcpkg.

SomeoneSerge: The upstream PR is at https://github.com/microsoft/vcpkg/pull/38159, although probably not yet in shape that upstream would accept.

pkgs/development/libraries/imgui/default.nix
```diff
@@ -1,8 +1,77 @@
...
+  # The intent is to mirror vcpkg's flags[^1],
+  # but we limit our attention to Linux and glfw3 for now
+  # [^1]: https://github.com/microsoft/vcpkg/blob/095ee06e7f60dceef7d713e3f8b1c2eb10d650d7/ports/imgui/CMakeLists.txt#L33-L108
+  withGLLoader ? !libGL.meta.unsupported,
+  withFeatures ?
+    lib.optionals (!glfw.meta.unsupported) [ "glfw" ]
+    ++ lib.optionals (!SDL2.meta.unsupported) [ "sdl2" ]
+    ++ lib.optionals (!stdenv.isDarwin) [ "opengl3" ]
+    ++ lib.optionals stdenv.isDarwin [ "metal" ]
+    ++ lib.optionals stdenv.hostPlatform.isAndroid [ "android" ],
```
SomeoneSerge, line 24: I'm quite unhappy about making this a list instead of individual `bool` attributes, but OTOH there's the issue of making a consistent mapping from the nix expression's arguments to vcpkg/cmake options and I'd rather not over-engineer it

nixos/doc/manual/release-notes/rl-2405.section.md
```diff
@@ -572,6 +572,11 @@ The pre-existing [services.ankisyncd](#opt-services.ankisyncd.enable) has been m
 
 - The `mpich` package expression now requires `withPm` to be a list, e.g. `"hydra:gforker"` becomes `[ "hydra" "gforker" ]`.
 
+- To facilitate dependency injection, the `imgui` package now builds a static
+  archive using vcpkg' CMake rules. It now installs "impl" headers selectively
+  instead of by a wildcard. Use `imgui.src` to if you just want to access the
+  unpacked sources.
```
SuperSandro2000, line 578:
```suggestion
- To facilitate dependency injection, the `imgui` package now builds a static archive using vcpkg' CMake rules.
  It now installs "impl" headers selectively instead of by a wildcard. Use `imgui.src` to if you just want to access the unpacked sources.
```
easier to read and we don't have an 80 line length limit in nixpkgs anyway

pkgs/by-name/im/implot/demos/default.nix
```diff
@@ -0,0 +1,72 @@
...
+  version = "unstable-2023-08-20";
+
+  src = fetchFromGitHub {
+    owner = "epezent";
+    repo = "implot_demos";
+    rev = "f33219d3ade192a2333d4a32e749842010952a29";
+    hash = "sha256-Xq0kVk8qncj/BzQyIX/l1OLvPSQJU8ckTxIIfLZdO/g=";
+  };
+
+
```
SuperSandro2000, line 30:
```suggestion

```

pkgs/by-name/im/imnodes/package.nix
```diff
@@ -0,0 +1,56 @@
...
+    "out"
+    "dev"
+  ];
+
+  src = fetchFromGitHub {
+    owner = "Nelarius";
+    repo = "imnodes";
+    rev = "8563e1655bd9bb1f249e6552cc6274d506ee788b";
+    hash = "sha256-E7NNCxYq9dyVvutWbpl2a+D2Ap2ErvdYHBDqpX0kb0c=";
+  };
```
SuperSandro2000, line 27:
```suggestion
  };

```

pkgs/by-name/im/implot/demos/default.nix
```diff
@@ -0,0 +1,72 @@
...
+      hash = "sha256-jYdM8NuwbZk7INKo2wqMAbjLMTPdrAdM4Kr3xmtquIY=";
+    })
+  ];
+
+  cmakeFlags = [ (lib.cmakeBool "FETCHCONTENT_FULLY_DISCONNECTED" true) ];
+
+  nativeBuildInputs = [
+    cmake
+    pkg-config
+  ];
```
SuperSandro2000, line 50:
```suggestion
  ];

```

pkgs/by-name/im/implot/demos/default.nix
```diff
@@ -0,0 +1,72 @@
...
+  src = fetchFromGitHub {
+    owner = "epezent";
+    repo = "implot_demos";
+    rev = "f33219d3ade192a2333d4a32e749842010952a29";
+    hash = "sha256-Xq0kVk8qncj/BzQyIX/l1OLvPSQJU8ckTxIIfLZdO/g=";
+  };
+
+
+  patches = [
+    # cmake
```
SuperSandro2000, line 32:
```suggestion
    # demos&app: fix build, missing symbols, etc 
```

pkgs/by-name/im/implot/demos/default.nix
```diff
@@ -0,0 +1,72 @@
...
+    (fetchpatch {
+      url = "https://github.com/epezent/implot_demos/commit/85a59612c102f8da97d6ead04f528f4c88f4ef9a.patch";
+      hash = "sha256-HRhfC3TUwz9Mv+1ODabaDxTWUaj4Nx1iH7C6vjUPo2s=";
+    })
+
+    # undefined symbols
```
SuperSandro2000, line 38:
```suggestion
    # CMakeLists.txt: try find_package first 
```

pkgs/by-name/im/imnodes/package.nix
```diff
@@ -0,0 +1,56 @@
...
+  patches = [
+    # CMake install rules
```
SuperSandro2000, line 29:
```suggestion
    # Add CMake install rules
```

pkgs/by-name/im/implot/demos/default.nix
```diff
@@ -0,0 +1,72 @@
...
+  ];
+  buildInputs =
+    [
+      imgui
+      imnodes
+      implot
+      (lib.getDev fmt)
+      iir1
+      (lib.getDev openssl)
+      (lib.getDev curl)
```
SuperSandro2000, line 59: Isn't this chosen automatically?

pkgs/by-name/im/implot/package.nix
```diff
@@ -0,0 +1,41 @@
...
+  src = fetchFromGitHub {
+    owner = "epezent";
+    repo = "implot";
+    rev = "v${version}";
+    hash = "sha256-/wkVsgz3wiUVZBCgRl2iDD6GWb+AoHN+u0aeqHHgem0=";
+  };
+
+  cmakeRules = "${vcpkg.src}/ports/implot";
+  postPatch = ''
+    cp "$cmakeRules"/CMakeLists.txt ./
```
SuperSandro2000, line 24:
```suggestion
  postPatch = ''
    cp "${vcpkg.src}/ports/implot"/CMakeLists.txt ./
```
or is this used as an env?

pkgs/development/libraries/imgui/default.nix
```diff
@@ -11,21 +80,32 @@ stdenv.mkDerivation rec {
     sha256 = "sha256-iTs8bcCAR8lvN1tcKmwyB3CcFmqR1QpqdGqdER920rM=";
   };
 
-  dontBuild = true;
+  cmakeRules = "${vcpkgSource}/ports/imgui";
+  postPatch = ''
+    cp "$cmakeRules"/{CMakeLists.txt,*.cmake.in} ./
```
SuperSandro2000, line 85:
```suggestion
  postPatch = ''
    cp "${vcpkgSource}/ports/imgui"/{CMakeLists.txt,*.cmake.in} ./
```

pkgs/development/libraries/imgui/default.nix
```diff
@@ -1,8 +1,77 @@
...
+  libGL,
+  darwin,
+  vcpkg,
+  SDL2,
+
+  # The intent is to mirror vcpkg's flags[^1],
+  # but we limit our attention to Linux and glfw3 for now
+  # [^1]: https://github.com/microsoft/vcpkg/blob/095ee06e7f60dceef7d713e3f8b1c2eb10d650d7/ports/imgui/CMakeLists.txt#L33-L108
+  withGLLoader ? !libGL.meta.unsupported,
+  withFeatures ?
```
SuperSandro2000, line 19: Not allowing to overwrite a single decision but only everything is no that great

pkgs/development/libraries/imgui/demo/default.nix
```diff
@@ -0,0 +1,31 @@
...
+stdenv.mkDerivation {
+  pname = "${imgui.pname}-demo";
+  inherit (imgui) version;
+
+  nativeBuildInputs = [ cmake ];
+  buildInputs = [
+    (imgui.override { withFeatures = lib.unique (imgui.withFeatures ++ [ "glfw" ]); })
+  ];
+
+  src = "${imgui.src}/examples/example_glfw_opengl3";
```
SuperSandro2000, line 17:
```suggestion
```

pkgs/development/libraries/imgui/demo/default.nix
```diff
@@ -0,0 +1,31 @@
...
+  lib,
+  stdenv,
+  cmake,
+  imgui,
+}:
+
+stdenv.mkDerivation {
+  pname = "${imgui.pname}-demo";
+  inherit (imgui) version;
+
```
SuperSandro2000, line 11:
```suggestion

  src = "${imgui.src}/examples/example_glfw_opengl3";

```

pkgs/by-name/im/implot/package.nix
SomeoneSerge: This is so as to keep phases agnostic of the hashes. I see no good reason to render the hashes into the scripts

pkgs/by-name/im/implot/demos/default.nix
SomeoneSerge: I wasn't sure. Openssl's first output is `bin`, but I suppose there might be some magic in mkDerivation to always peek `outputDev` for `buildInputs`. I'll test

pkgs/development/libraries/imgui/default.nix
SomeoneSerge: The passthru tests include an example which uses e.g. `unique (prev.withFeatures ++ [ ... ])` but that is clearly much less flexible than individual attributes.

One obvious alternative is to make the lambda take `IMGUI_BUILD_GLFW_BINDING, ...` as arguments, directly mirroring vcpkg's options. The issue with that is I'd have to be selective: I wouldn't want to declare `IMGUI_BUILD_ALLEGRO5_BINDING` or `IMGUI_BUILD_DX12_BINDING` because I don't actually plan to be implementing support for that (someone else might, I won't).

I could really use some advise or suggestions on this one

pkgs/by-name/im/implot/demos/default.nix
SuperSandro2000: Yes, there is

pkgs/by-name/im/implot/package.nix
SuperSandro2000: I can't follow you.

Usually turning nix variables into bash variables is an anti pattern we try to avoid.

SomeoneSerge: Introducing the auxiliary variable like this gives the opportunity to `implot.overrideAttrs (_: { cmakeRules = ./my-custom-cmake; })` which otherwise would require a careful surgery on the `postPatch`. I've been following a similar pattern outside the tree and thought it wouldn't do harm to do the same here.

>  an anti pattern we try to avoid.

Yesssss setup-hooks are an anti-pattern, let's make nix evaluation faster and abolish setup hooks 😆

pkgs/development/libraries/imgui/default.nix
```diff
@@ -1,8 +1,66 @@
...
+  IMGUI_BUILD_DX12_BINDING ? false,
+  IMGUI_BUILD_GLFW_BINDING ? !glfw.meta.unsupported && !libGL.meta.unsupported,
+  IMGUI_BUILD_GLUT_BINDING ? false,
+  IMGUI_BUILD_METAL_BINDING ? stdenv.isDarwin,
+  IMGUI_BUILD_OPENGL2_BINDING ? false,
+  IMGUI_BUILD_OPENGL3_BINDING ?
+    IMGUI_BUILD_SDL2_BINDING || IMGUI_BUILD_GLFW_BINDING || IMGUI_BUILD_GLUT_BINDING,
+  IMGUI_BUILD_OSX_BINDING ? false,
+  IMGUI_BUILD_SDL2_BINDING ?
+    !IMGUI_BUILD_GLFW_BINDING && !SDL2.meta.unsupported && !libGL.meta.unsupported,
```
SomeoneSerge, line 33: I decided to replace lists and maps with the boilerplate. It's stupid and easy to get lost in when editing, but it makes more sense as an interface and mimics vcpkg options exactly. Most of these options currently set `broken = true`, but the default instantiations like SDL/glfw3 on linux work as expected

pkgs/by-name/im/implot/demos/default.nix
```diff
@@ -0,0 +1,78 @@
...
+    (fetchpatch {
+      url = "https://github.com/epezent/implot_demos/commit/4add0433a46ed5e2099e1af1a77e8055e49230d0.patch";
+      hash = "sha256-jYdM8NuwbZk7INKo2wqMAbjLMTPdrAdM4Kr3xmtquIY=";
+    })
+
+    # CMake: link libGL from the "app" target
+    (fetchpatch {
+      url = "https://github.com/epezent/implot_demos/commit/6742e4202858eb85bd0d67ca5fa15a7a07e6b618.patch";
+      hash = "sha256-h4EJ9u1iHLYkGHgxSynskkuCGmY6mmvKdZSRwHJKerY=";
+    })
```
SomeoneSerge, line 47: vcpkg' CMakeLists.txt for imgui doesn't `target_link_libraries(... OpenGL::GL)`. I wasn't sure if it should, so for now I made the demo link libGL explicitly

pkgs/by-name/im/implot/package.nix
SuperSandro2000: You should be able to use finalAttrs to allow overrides like that, too.

pkgs/development/libraries/imgui/default.nix
SomeoneSerge: Ditched lists and passthru in favour of boilerplate attributes

pkgs/by-name/im/implot/package.nix
SomeoneSerge: That's an option, but it's not really any better than eliminating the template rendering and treating `postPatch` as a normal setup hook

nixos/doc/manual/release-notes/rl-2405.section.md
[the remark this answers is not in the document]
SomeoneSerge: I'd like to merge tomorrow unless there are objections.


@SuperSandro2000 we haven't reached the agreement on the `finalAttrs` matter but we can keep the question open and if we end up choosing against bash variables I'll just open a follow-up PR

SomeoneSerge:
Well if there should be any objections they can be resolved in future PRs.

RE: Ofborg

- Aarch64-linux builds fine locally (can't post a nixpkgs-review because my aarch64 builder doesn't have enough storage for the whole thing 🙈)
- I can't test Darwin, it's something to follow up on

wegank:
> I can't test Darwin, it's something to follow up on

And I guess this PR does break imgui on darwin....

SomeoneSerge:
> And I guess this PR does break imgui on darwin....

Well, "breaks" is maybe an overly strong way to put it, given that previously this derivation didn't actually "build" imgui. As I mentioned in the update, clients that prefer the "add_subdirectory" approach can still use `imgui.src`.

I'll try to clean this up rn

wegank:
Opened #315917, it's just missing some frameworks.

PR #303703, opened 2024-04-12
minipicolisp: init at 2024-04-12

nat-418:
MiniPicoLisp is an embedded, "pure" variant of PicoLisp.


ocfox:
No value putting it in nixpkgs.

PR #303724, opened 2024-04-12
rPackages.immunotation: cache external URLs

Kupac:
## Description of changes

Immunotation uses 3 external URL-s, and stores them variables in the package's namespace. Package installation is not possible without internet connection, and skipping load test doesn't help, because this happens _before_ the load test.

I use fetchurl to get those external deps, and replace them in the code with a patch + substituteInPlace.

This should work in the long term for the  two MHC allele lists, but I'm not sure about the geographical region list. We'll have to see how frequently the website changes. I only found the copyright2024 notice problematic, because it could change every year, which means we'd have to update the hash. I think that would still be acceptable. If it's more frequent, we'll have to mark this as broken, I think. 

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/development/r-modules/default.nix
+++ b/pkgs/development/r-modules/default.nix
@@ -1318,6 +1318,36 @@ let
       buildInputs = [ cacert ] ++ attrs.buildInputs;
     });
 
+
+    immunotation = let
+      MHC41alleleList = fetchurl {
+        url = "https://services.healthtech.dtu.dk/services/NetMHCpan-4.1/allele.list";
+        hash = "sha256-CRZ+0uHzcq5zK5eONucAChXIXO8tnq5sSEAS80Z7jhg=";
+      };
+
+      MHCII40alleleList = fetchurl {
+        url = "https://services.healthtech.dtu.dk/services/NetMHCIIpan-4.0/alleles_name.list";
+        hash = "sha256-K4Ic2NUs3P4IkvOODwZ0c4Yh8caex5Ih0uO5jXRHp40=";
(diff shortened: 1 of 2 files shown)


b-rodrigues (approved):
builds and loads on aarch64-linux

lkupcsik:
@ofborg build rPackages.immunotation for @jbedo

Kupac:
Ok, this won't work, the hash of hla6006a already changed

b-rodrigues:
Stupid question: could we use an archived page for hla6006a? does the actual data of interest change or is it just the website ?

Kupac:
That could be an option, actually. Normally it is retrieved once, at install time, so it's not that different here. Other possibility could be to retrieve it in .onLoad() but that would require some  more excessive patching.

Kupac:
We could use an archive.org snapshot as a source. There are small changes from time to time, but these probably don't affect the list of valid geographical regions, only some count values, which are not needed by the script at this stage. However, it should be updated once in a while to a new snapshot. I'd prefer to move this whole connection stuff to .onLoad, like in many other packages. I don't see why not. It's just not very straightforward to do in nix.

https://web.archive.org/web/diff/20231207065757/20240205120933/www.allelefrequencies.net/hla6006a.asp

PR #303764, opened 2024-04-13
Add nixos-needsreboot to NixOS

thefossguy:
## Description of changes

This PR adds my stupid package that compares the current and latest NixOS generations and creates a file to let you know if you "_need_" to reboot or not.

- Built on platform(s)
  - [ ] x86_64-linux
  - [x] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/nixos/modules/services/misc/nixos-needsreboot.nix
+++ b/nixos/modules/services/misc/nixos-needsreboot.nix
@@ -1,0 +1,29 @@
+{ config, lib, pkgs, ... }:
+let
+  cfg = config.services.nixos-needsreboot;
+in
+{
+  options = {
+    services.nixos-needsreboot = {
+      enable = lib.mkEnableOption "nixos-needsreboot, which determines if you need to reboot your NixOS machine";
+    };
+  };
+
+  config = lib.mkIf cfg.enable {
+    assertions = [{
+      assertion = config.system.autoUpgrade.enable == true;
+      message  = "Please enable 'system.autoUpgrade'.";
+    }];
+
+    systemd.services.nixos-needsreboot = {
+      description = "Determine if you need to reboot your NixOS machine";
+      requires = [ "nixos-upgrade.service" ];
+      wantedBy = [ "multi-user.target" ];
+      serviceConfig = {
+        ExecStart = lib.getExe pkgs.nixos-needsreboot;
+      };
+    };
+  };
+
+  meta.maintainers = with lib.maintainers; [ thefossguy ];
+}
--- a/pkgs/by-name/ni/nixos-needsreboot/package.nix
+++ b/pkgs/by-name/ni/nixos-needsreboot/package.nix
@@ -1,0 +1,26 @@
+{ fetchFromGitHub
+, lib
+, rustPlatform
+}:
+
+rustPlatform.buildRustPackage rec {
+  pname = "nixos-needsreboot";
+  version = "0.1.10";
+
+  src = fetchFromGitHub {
+    owner = "thefossguy";
+    repo = "nixos-needsreboot";
+    rev = version;
+    hash = "sha256-0MpqqkkpTMiq5xgzSOqBZMLUfVhtfEfRXWZzt2tqRKI=";
+  };
+
+  cargoHash = "sha256-LzO1kkrpWTjLnqs0HH5AIFLOZxtg0kUDIqXCVKSqsAc=";
+
+  meta = with lib; {
+    description = "Determine if you need to reboot your NixOS machine";
+    homepage = "https://github.com/thefossguy/nixos-needsreboot";
+    license = licenses.gpl2Only;
+    maintainers = with maintainers; [ thefossguy ];
+    mainProgram = "nixos-needsreboot";
+  };
+}
--- a/nixos/doc/manual/release-notes/rl-2405.section.md
+++ b/nixos/doc/manual/release-notes/rl-2405.section.md
@@ -80,6 +80,8 @@ Use `services.pipewire.extraConfig` or `services.pipewire.configPackages` for Pi
 
 <!-- To avoid merge conflicts, consider adding your item at an arbitrary place in the list instead. -->
 
+- [nixos-needsreboot](https://github.com/thefossguy/nixos-needsreboot), determines if you need to reboot your NixOS machine.
+
 - [ownCloud Infinite Scale Stack](https://owncloud.com/infinite-scale-4-0/), a modern and scalable rewrite of ownCloud.
 
 - [Handheld Daemon](https://github.com/hhd-dev/hhd), support for gaming handhelds like the Legion Go, ROG Ally, and GPD Win. Available as [services.handheld-daemon](#opt-services.handheld-daemon.enable).
--- a/nixos/modules/module-list.nix
+++ b/nixos/modules/module-list.nix
@@ -753,6 +753,7 @@
   ./services/misc/nix-gc.nix
   ./services/misc/nix-optimise.nix
   ./services/misc/nix-ssh-serve.nix
+  ./services/misc/nixos-needsreboot.nix
   ./services/misc/novacomd.nix
   ./services/misc/ntfy-sh.nix
   ./services/misc/nzbget.nix


thefossguy:
Thank you for the review @Aleksanaa! I made the changes you suggested and force-pushed :)

nixos/modules/services/misc/nixos-needsreboot.nix
```diff
@@ -0,0 +1,24 @@
...
+    };
+  };
+
+  config = lib.mkIf cfg.enable {
+    systemd.services.nixos-needsreboot = {
+      description = "Determine if you need to reboot your NixOS machine";
+      requires = [ "nixos-upgrade.service" ];
+      wantedBy = [ "multi-user.target" ];
+      serviceConfig = {
+        ExecStart = "${pkgs.nixos-needsreboot}/bin/nixos-needsreboot";
```
SuperSandro2000, line 18: We can use lib.getExe here

SuperSandro2000 (commented):
Using rust for such a simple comparison is overkill. We can easily check this in 5 lines of bash if the kernel booted is different from the latest generation. Also systemd could also be re-execed.

thefossguy:
> Using rust for such a simple comparison is overkill. We can easily check this in 5 lines of bash if the kernel booted is different from the latest generation.

Why I wrote it in Rust: "When all you have is a hammer, everything looks like a nail" :)

I did think of using Bash but when releasing it to others, thought of using Rust because people might want to check for other pieces of software that are critical [to them] (ssh). And Rust, though over-engineered for this simple problem, might end up being a good-enough bet over Bash.

I'm open to further discussion if you think otherwise.

> Also systemd could also be re-execed.

This is less for "you absolutely must reboot" and more like "hey, this updated, might wanna reboot when you get free".

nixos/modules/services/misc/nixos-needsreboot.nix
thefossguy: I assumed that was for instances when the user had an option to select the package. Should I still use `lib.getExe` given that there's only one package?

thefossguy:
_friendly ping_ @SuperSandro2000

nixos/modules/services/misc/nixos-needsreboot.nix
dotlambda: Yes.
```suggestion
ExecStart = lib.getExe pkgs.nixos-needsreboot;
```

pkgs/by-name/ni/nixos-needsreboot/package.nix
```diff
@@ -0,0 +1,26 @@
...
+}:
+
+rustPlatform.buildRustPackage rec {
+  pname = "nixos-needsreboot";
+  version = "0.1.5";
+
+  src = fetchFromGitHub {
+    owner = "thefossguy";
+    repo = "nixos-needsreboot";
+    rev = "${version}";
```
dotlambda, line 13:
```suggestion
    rev = version;
```
or
```suggestion
    rev = "refs/tags/${version}";
```

nixos/modules/services/misc/nixos-needsreboot.nix
```diff
@@ -0,0 +1,24 @@
...
+  options = {
+    services.nixos-needsreboot = {
+      enable = lib.mkEnableOption "nixos-needsreboot, which determines if you need to reboot your NixOS machine";
+    };
+  };
+
+  config = lib.mkIf cfg.enable {
+    systemd.services.nixos-needsreboot = {
+      description = "Determine if you need to reboot your NixOS machine";
+      requires = [ "nixos-upgrade.service" ];
```
dotlambda, line 15: So we should probably add an assertion that `config.system.autoUpgrade.enable == true`.

thefossguy: Interesting, in which case, should that assertion be under `options`?

```diff
-      enable = lib.mkEnableOption "nixos-needsreboot, which determines if you need to reboot your NixOS machine";
+      enable = config.system.autoUpgrade.enable;
```

dotlambda: It's definitely still needed. Not everyone enabling auto upgrades wants to use nixos-needsreboot.

dotlambda: There is still no assertion.

thefossguy: Oh, assertion. I was sleepy and misread+mistook it for depending on `config.system.autoUpgrade.enable`. Sorry, my bad lol

thefossguy:
Bump. Is this ready for merging?

nixos/modules/services/misc/nixos-needsreboot.nix
```diff
@@ -0,0 +1,29 @@
...
+{
+  options = {
+    services.nixos-needsreboot = {
+      enable = lib.mkEnableOption "nixos-needsreboot, which determines if you need to reboot your NixOS machine";
+    };
+  };
+
+  config = lib.mkIf cfg.enable {
+    assertions = [{
+      assertion = config.system.autoUpgrade.enable == true;
```
dotlambda, line 14:
```suggestion
      assertion = config.system.autoUpgrade.enable;
```

nixos/modules/services/misc/nixos-needsreboot.nix
```diff
@@ -0,0 +1,29 @@
...
+      message  = "Please enable 'system.autoUpgrade'.";
+    }];
+
+    systemd.services.nixos-needsreboot = {
+      description = "Determine if you need to reboot your NixOS machine";
+      requires = [ "nixos-upgrade.service" ];
+      wantedBy = [ "multi-user.target" ];
+      serviceConfig = {
```
dotlambda, line 22: Do we want to do any hardening?

dotlambda (commented):
Please squash the last two commits.

nixos/modules/services/misc/nixos-needsreboot.nix
thefossguy: I'm not sure what you mean by hardening

jian-lin: https://wiki.nixos.org/wiki/Systemd/Hardening

An example: https://github.com/NixOS/nixpkgs/blob/c1a16412b44d133c7f6a37ad993bfd7323bd9180/nixos/modules/services/hardware/kanata.nix#L117-L148

thefossguy: I don't think so. It only compares a few filesystem paths and determines version from it. The only write operation it performs is to write what critical "package" was upgraded that necessitates a reboot in the `/var/run/reboot-required` file.

thefossguy: That said, what do you think about hardening? Do you feel like it's needed, if yes, what do you want? I'm open to hearing it!

dotlambda: Make everything but that file read only, block network access, etc.

thefossguy: ack, will push again with these three fixes:
1. [fix assertion ](https://github.com/NixOS/nixpkgs/pull/303764#discussion_r1584672534)
2. [squash last two commits](https://github.com/NixOS/nixpkgs/pull/303764#pullrequestreview-2031100030)
3. [enable hardening](https://github.com/NixOS/nixpkgs/pull/303764#discussion_r1587055159)

nixos/modules/services/misc/nixos-needsreboot.nix
SuperSandro2000: So this module absolutely requires system.autoUpgrade? If not we could just make this conditional and I first thought that this would also work without auto upgrades.

thefossguy: You can run the standalone binary from the shell if you want to manually trigger it, but to trigger the module, I don't see anything other than `system.autoUpgrade`.

PR #303811, opened 2024-04-13
doc: migrate filesets to doc-comment format

hsjobeki:
## Description of changes

lib.filesets comments migrated to doc-comments.

Manual steps:

- [x] File description headings should start with h1 not h2.  (consistency across doc-comments)
- [x] reformatted some of the arguments/structured args documentation manually. Since they would be lost with our current automatism.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/lib/fileset/default.nix
+++ b/lib/fileset/default.nix
@@ -218,8 +218,12 @@ in {
 
     : The file set to trace.
 
-    This argument can also be a path,
-    which gets [implicitly coerced to a file set](#sec-fileset-path-coercion).
+      This argument can also be a path,
+      which gets [implicitly coerced to a file set](#sec-fileset-path-coercion).
+
+    `val`
+
+    : The value to return.
 
     # Type
 
@@ -271,8 +275,8 @@ in {
 
     : The file set to trace and return.
 
-    This argument can also be a path,
-    which gets [implicitly coerced to a file set](#sec-fileset-path-coercion).
+      This argument can also be a path,
+      which gets [implicitly coerced to a file set](#sec-fileset-path-coercion).
 
     # Type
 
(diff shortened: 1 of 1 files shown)


lib/fileset/default.nix
```diff
@@ -194,27 +211,40 @@ in {
...
-    The file set to trace.
+
+    # Inputs
+
+    `fileset`
+
+    : The file set to trace.
 
     This argument can also be a path,
     which gets [implicitly coerced to a file set](#sec-fileset-path-coercion).
```
DanielSidhion, line 222: This content is for the same argument. The last 2 lines should have two extra spaces of indentation so it's rendered correctly.

lib/fileset/default.nix
```diff
@@ -194,27 +211,40 @@ in {
...
-    */
-    fileset:
+
```
DanielSidhion, line 223:
```suggestion

    `val`

    : The value to return.

```

lib/fileset/default.nix
```diff
@@ -234,34 +264,47 @@ in {
...
-    The file set to trace and return.
+
+    # Inputs
+
+    `fileset`
+
+    : The file set to trace and return.
 
     This argument can also be a path,
     which gets [implicitly coerced to a file set](#sec-fileset-path-coercion).
```
DanielSidhion, line 275: This content is for the same argument. The last 2 lines should have two extra spaces of indentation so it's rendered correctly.

lib/fileset/default.nix
```diff
@@ -286,83 +329,98 @@ in {
...
-      toSource {
-        root = "/nix/store/...-source";
-        fileset = ./.;
-      }
-      => <error>
+    # Inputs
+
+    Takes an attribute set with the following attributes
+
+    ## root (required, Path)
```
DanielSidhion, line 336: Ideally this shouldn't be a heading, but instead continue to be a definition list just like with other functions. Is there a reason why you wanted to turn them into headings?

lib/fileset/default.nix
```diff
@@ -632,55 +758,72 @@ in {
...
+      `hasExt ext` is true only if `hasSuffix ".${ext}" name`.
 
-        This also means that e.g. for a file with name `.gitignore`,
-        `hasExt "gitignore"` is true.
+      This also means that e.g. for a file with name `.gitignore`,
+      `hasExt "gitignore"` is true.
 
-      Other attributes may be added in the future.
-    */
+    Other attributes may be added in the future.
```
DanielSidhion, line 785: Also missing extra indentation on many of the lines here, they should all be part of the content for `predicate`.

DanielSidhion (commented):
@hsjobeki I didn't make suggestions to fix the line indentation because I'm assuming you want to get these fixed in the tooling you're using for the conversion. The only suggestion I made is for a bit of extra content in the description of one of the functions, just because that function's description mentions two arguments, but only one is caught by the tooling.

lib/fileset/default.nix
hsjobeki: I wanted to try a better way than doing a nested definition list. Because this function only takes on input which is an attribute set with some attributes.

This would be

```md

structured function argument

: root
    : definition of root
    where must the next line start ?
: fileset
    : definition of fileset
```


I always mess up the indentation with nested definition lists-

DanielSidhion: In this particular case, you won't need a nested definition list though, right? You could convert the heading style into a single definition list:

```
`root` (Path; _required_)
: definition of root

`fileset` (FileSet; _required_)
: definition of fileset
```

As for the indentation, I always follow the same [indentation rule](https://spec.commonmark.org/0.31.2/#list-items) as normal markdown lists and nested lists: make sure the content of the definition starts at the same level of indentation:

```
`attributeSet`
: Definition of attribute set.
  Note that it starts in column 2 of the text (counting from 0), so any lines that follow it and should be part of the definition must also have at least 2 spaces at the beginning.

  Line breaks are ok, as long as the content still has the 2 spaces at the beginning.

  `attribute1`
  : This is a definition for an attribute of `attributeSet`.
    Note that it starts in column 4 now, so any lines that should be part of `attribute` must also have at least 4 spaces at the beginning.

    As always, line breaks are still ok, as long as the content has 4 spaces at the beginning.

  `attribute2`
  : Since `attribute2` started with only 2 spaces of indentation, it's part of the content of `attributeSet` instead of the content for `attribute1`.
```

20240417T043740

You can't go wrong by ensuring content always starts on the same column.

hsjobeki:
@DanielSidhion I'll add the indentation fixes to the tooling. Most of it was done manually anyways. I dont have tooling to automatically migrate structured arguments.
This is still an open question, so i did it by hand.

hsjobeki:
@DanielSidhion indentation should be fine now.

DanielSidhion (approved):
Thanks for the changes! Also rendered the manual locally and checked that nothing was broken.

PR #283460, opened 2024-01-24
pkgsStatic: Pass hostPlatform.gcc attribute

rodarima:
To build the security wrappers[1] the pkgsStatic stdenv is used, so the binaries are static. However, the hostPlatform may have gcc attributes that are *required* to build binaries so they can run on the host platform. In particular, this is the case when using gcc.arch, which ends up injecting -march=... in the gcc wrapper. Those attributes are not contained in hostPlatform.parsed.

This change sets the same gcc attributes found in the hostPlatform for the pkgsStatic cross system, so it can build binaries with the same gcc flags.

---

This problem has ocurred to me while cross compiling a RISC-V NixOS system from an x86 machine to run in QEMU with the compressed instructions disabled (using `-cpu rv64,c=false`). What happens when the -march flags are not propagated to the gcc wrapper is that the binaries end up being compiled with compressed instructions which cause a SIGILL like the following:

```
$ mount
[52637.699560] mount[2334]: unhandled signal 4 code 0x1 at 0x000000000001048c in mount[10000+e000]
[52637.702177] CPU: 19 PID: 2334 Comm: mount Not tainted 6.1.72 #1-NixOS
[52637.703517] Hardware name: riscv-virtio,qemu (DT)
[52637.705091] epc : 000000000001048c ra : 000000000004a0dc sp : 00ffffffc5588bf0
[52637.706789]  gp : 000000000001f800 tp : 00ffffffabed63a0 t0 : ffffffffffffffff
[52637.708603]  t1 : 0000000000030a8c t2 : 00000000001952a8 s0 : 00000000001f29c0
[52637.710191]  s1 : 00000000001f29c0 a0 : 0000000000000000 a1 : 00000000001807c0
[52637.711965]  a2 : 0000000000180af0 a3 : 0000000000000000 a4 : 0000000000116010
[52637.713687]  a5 : 0000000000000000 a6 : 00000000001807c0 a7 : 00000000000000dd
[52637.715348]  s2 : 000000000010cefc s3 : 00000000001805c0 s4 : 00000000001807c0
[52637.717137]  s5 : 00ffffffac1a8d50 s6 : 0000000000180af0 s7 : 00000000001807c0
[52637.718808]  s8 : 0000000000117890 s9 : 0000000000000000 s10: 00000000001134c0
[52637.720686]  s11: 00000000001134cc t3 : 00ffffffabf9f348 t4 : 918a9092d091969d
[52637.722370]  t5 : 0000000000000010 t6 : 0000000000000007
[...]

PR #136549, opened 2021-09-03
pkgsStatic: make pkgsMusl a no-op

r-burns:
When pkgsStatic is already musl, we can let pkgsMusl simply be a no-op.

As the interaction of pkgsStatic and pkgsMusl machinery is currently broken
(see discussion in https://github.com/NixOS/nixpkgs/issues/114510), this fixes pkgsStatic.pkgsMusl.stdenv.

###### Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
- [ ] For non-Linux: Is `sandbox = true` set in `nix.conf`? (See [Nix manual](https://nixos.org/manual/nix/stable/#sec-conf-file))
- [ ] Tested via one or more NixOS test(s) if existing and applicable for the change (look inside [nixos/tests](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests))
- [ ] Tested compilation of all packages that depend on this change using `nix-shell -p nixpkgs-review --run "nixpkgs-review wip"`
- [ ] Tested execution of all binary files (usually in `./result/bin/`)
- [21.11 Release Notes (or backporting 21.05 Release notes)](https://github.com/NixOS/nixpkgs/blob/master/CONTRIBUTING.md#generating-2111-release-notes)
  - [ ] (Package updates) Added a release notes entry if the change is major or breaking
  - [ ] (Module updates) Added a release notes entry if the change is significant
  - [ ] (Module addition) Added a release notes entry if adding a new NixOS module
- [ ] Fits [CONTRIBUTING.md](https://github.com/NixOS/nixpkgs/blob/master/CONTRIBUTING.md).

Issue #114510, opened 2021-02-26
pkgsMusl.pkgsStatic.stdenv is broken


r-burns:
**Describe the bug**
The gcc used for pkgsMusl.pkgsStatic.stdenv fails to build on x86_64-linux, due to a faulty bintools setup, as pointed out by Sandro here: https://github.com/NixOS/nixpkgs/pull/112096#issuecomment-783143041.

I think the reason for this is that the binutils hostPlatform is different from the targetPlatform (host is dynamic musl, target is static) but the configs are both `x86_64-unknown-linux-musl`. So our package definitions expect the resulting bintools to have a `x86_64-unknown-linux-musl` targetPrefix, but they do not because the autotools configure script only sees matching host/target configs, so it produces unprefixed `ld`, `as`, etc binaries.

**To Reproduce**
`nix-build '<nixpkgs>' -A pkgsMusl.pkgsStatic.stdenv`

**Notify maintainers**
Nixpkgs cross maintainers: @matthewbauer @Ericson2314 

**Metadata**
 - system: `"x86_64-linux"`
 - host os: `Linux 5.4.87, NixOS, 21.03pre268206.536fe36e23a (Okapi)`
 - multi-user?: `yes`
 - sandbox: `yes`
 - version: `nix-env (Nix) 2.3.10`
 - channels(root): `"nixos-21.03pre268206.536fe36e23a"`
 - nixpkgs: `/nix/var/nix/profiles/per-user/root/channels/nixos`

Maintainer information:
```yaml
# a list of nixpkgs attributes affected by the problem
attribute:
  - binutils
  - gcc
# a list of nixos modules affected by the problem
module:
```

samueldr:
Isn't `pkgsStatic` already musl? I ask, but I know it is.

```
nix-repl> pkgsStatic.targetPlatform.isMusl
true
```

I'm not sure what the difference *should* be when comparing `pkgsMusl.pkgsStatic.*` vs. `pkgsStatic.*` *considering pkgsStatic is using musl*.

To me, it *feels* like `pkgsMusl.pkgsStatic` should be a no-op, but obviously (as I'm observing) it is not.

r-burns:
Musl is not necessarily static, so pkgsMusl can be dynamic. I think pkgsStatic implies musl but only because glibc is bad at static linking. So I think pkgsStatic.pkgsMusl would be redundant but pkgsMusl.pkgsStatic would not be.

samueldr:
[...]

PR #303849, opened 2024-04-13
pkgs/top-level: make package sets composable

### LINKED ISSUE #281596: Build failure: pkgsCross.mingwW64.pkgsStatic.stdenv
### Steps To Reproduce

1. build *pkgsCross.mingwW64.pkgsStatic.stdenv*

### Build log

```
error: builder for '/nix/store/v1ycfkiivjszsl8qagsw2v86r81amwxs-mingw-w64-static-x86_64-w64-windows-gnu-10.0.0-headers.drv' failed with exit code 1;
       last 10 log lines:
       > checking for x86_64-w64-windows-gnu-strip... no
       > checking for strip... no
       > checking for a race-free mkdir -p... /nix/store/khndnv11g1rmzhzymm1s5dw7l2ld45bc-coreutils-9.4/bin/mkdir -p
       > checking for gawk... gawk
       > checking whether make sets $(MAKE)... yes
       > checking whether make supports nested variables... yes
       > checking whether to enable maintainer-specific portions of Makefiles... no
       > checking build system type... x86_64-unknown-linux-gnu
       > checking host system type... Invalid configuration 'x86_64-w64-windows-gnu': Kernel 'windows' not known to work with OS 'gnu'.
       > configure: error: /nix/store/9vafkkic27k7m4934fpawl6yip3a6k4h-bash-5.2-p21/bin/bash ./build-aux/config.sub x86_64-w64-windows-gnu failed
```

### Metadata

```console
 - system: `"x86_64-linux"`
 - host os: `Linux 6.6.10-zen1, NixOS, 24.05 (Uakari), 24.05.20240115.c3e128f`
 - multi-user?: `yes`
 - sandbox: `yes`
 - channels(root): `"nixos"`
 - channels(user): `"nixgl, nixpkgs"`
 - nixpkgs: `/nix/store/11zbgb8j7wnnccbbjcq0q556h28g7p4r-source`
```

---

wolfgangwalther:
## Description of changes

The various `pkgsXYZ` top-level package sets did not pass `localSystem` / `crossSystem` options to lower levels, so far. This change propagates the original arguments, i.e. `localSystem0` / `crossSystem0` to lower levels. Those include the overrides made by an upper package set.

Example: `pkgsStatic` adds `crossSystem = { isStatic = true }`. `pkgsStatic.pkgsMusl` previously lost this setting, because it didn't propagate `crossSystem0`, yet.

The first commit removes / rewrites some outdated comments, which referred to the cross and i686 package sets. Meanwhile, we have a lot more, so I rewrote those comments to not say anything about specific sets anymore.

The second commit adds some tests to confirm composability of package sets. This helped me coming up with the fixes, I'm not sure whether it should be kept. It takes about 8-9 seconds for me to run. It also helps showing what the remaining commits fix by uncommenting the previously broken tests.

The remaining commits fix those composability issues. This:
- will fix #281596 (@yshui, @wegank)
- could possibly fix #283460 with a more principal approach (@rodarima, @alyssais, @K900)
- supersedes the fix in #136549 (@r-burns, @rnhmjoj)

~~One open question for me is how to deal with `pkgsXYZ.pkgsCross`. This has practical challenges (#114510), but is also conceptually unsound: In a way `pkgsCross` expects to reset `crossSystem` from scratch, because the whole point about this layer is to start with "well known" example configurations. IMHO, it's fine to make modifications on top of that. e.g. `pkgsCross.XYZ.pkgsLLVM` or `pkgsCross.XYZ.pkgsStatic`, but it makes little sense to have `pkgsLLVM.pkgsCross...` or `pkgsStatic.pkgsCross...`. Probably completely unusable is `pkgsCross.XYZ.pkgsCross.ABC`.~~

~~One idea would be to just restrict `pkgsCross` to be used at the top-level, for example by adding something like this in each package set, including `pkgsCross` itself: [...]~~

~~That would also disallow `pkgsMusl.pkgsCross.XYZ`, which I think is about the only remotely reasonable use-case here: "cross compiling from a native musl environment". I'm not sure whether anyone uses that, though, and whether it's worth supporting at all.~~

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/test/top-level/stage.nix
+++ b/pkgs/test/top-level/stage.nix
@@ -1,0 +1,128 @@
+# run like this:
+#  nix-build pkgs/test/top-level/stage.nix
+{
+  localSystem ? {
+    system = builtins.currentSystem;
+  },
+}:
+
+with import ../../top-level { inherit localSystem; };
+
+let
+  # To silence platform specific evaluation errors
+  discardEvaluationErrors = e: (builtins.tryEval e).success -> e;
+
+  # Basic test for idempotency of the package set, i.e:
+  # Applying the same package set twice should work and
+  # not change anything.
+  isIdempotent = set: discardEvaluationErrors (pkgs.${set}.stdenv == pkgs.${set}.${set}.stdenv);
+
+  # Some package sets should be noops in certain circumstances.
+  # This is very similar to the idempotency test, but not going
+  # via the super' overlay.
+  isNoop =
+    parent: child:
+    discardEvaluationErrors (
+      (lib.getAttrFromPath parent pkgs).stdenv == (lib.getAttrFromPath parent pkgs).${child}.stdenv
+    );
+
+  allMuslExamples = builtins.attrNames (
+    lib.filterAttrs (_: system: lib.hasSuffix "-musl" system.config) lib.systems.examples
+  );
+
+  allLLVMExamples = builtins.attrNames (
+    lib.filterAttrs (_: system: system.useLLVM or false) lib.systems.examples
+  );
+
+  # A package set should only change specific configuration, but needs
+  # to keep all other configuration from previous layers in place.
+  # Each package set has one or more key characteristics for which we
+  # test here. Those should be kept, even when applying the "set" package
+  # set.
+  isComposable =
+    set:
+    (
+      # Can't compose two different libcs...
+      builtins.elem set [ "pkgsLLVMLibc" ]
+      || discardEvaluationErrors (
+        pkgsCross.mingwW64.${set}.stdenv.hostPlatform.config == "x86_64-w64-mingw32"
+      )
+    )
+    && (
+      # Can't compose two different libcs...
+      builtins.elem set [ "pkgsLLVMLibc" ]
+      || discardEvaluationErrors (pkgsCross.mingwW64.${set}.stdenv.hostPlatform.libc == "msvcrt")
+    )
+    && discardEvaluationErrors (pkgsCross.ppc64-musl.${set}.stdenv.hostPlatform.gcc.abi == "elfv2")
+    && discardEvaluationErrors (
+      builtins.elem "trivialautovarinit" pkgs.pkgsExtraHardening.${set}.stdenv.cc.defaultHardeningFlags
+    )
+    && discardEvaluationErrors (pkgs.pkgsLLVM.${set}.stdenv.hostPlatform.useLLVM)
+    && (
+      # Can't compose two different libcs...
+      builtins.elem set [
+        "pkgsMusl"
+        "pkgsStatic"
+      ]
+      || discardEvaluationErrors (pkgs.pkgsLLVMLibc.${set}.stdenv.hostPlatform.isLLVMLibc)
+    )
+    && discardEvaluationErrors (pkgs.pkgsArocc.${set}.stdenv.hostPlatform.useArocc)
+    && discardEvaluationErrors (pkgs.pkgsZig.${set}.stdenv.hostPlatform.useZig)
+    && discardEvaluationErrors (pkgs.pkgsLinux.${set}.stdenv.buildPlatform.isLinux)
+    && (
+      # Can't compose two different libcs...
+      builtins.elem set [ "pkgsLLVMLibc" ]
+      || discardEvaluationErrors (pkgs.pkgsMusl.${set}.stdenv.hostPlatform.isMusl)
+    )
+    && discardEvaluationErrors (pkgs.pkgsStatic.${set}.stdenv.hostPlatform.isStatic)
+    && discardEvaluationErrors (pkgs.pkgsi686Linux.${set}.stdenv.hostPlatform.isx86_32)
+    && discardEvaluationErrors (pkgs.pkgsx86_64Darwin.${set}.stdenv.hostPlatform.isx86_64);
+in
+
+# Appends same defaultHardeningFlags again on each .pkgsExtraHardening - thus not idempotent.
+# assert isIdempotent "pkgsExtraHardening";
+# TODO: Remove the isDarwin condition, which currently results in infinite recursion.
+# Also see https://github.com/NixOS/nixpkgs/pull/330567#discussion_r1894653309
+assert (stdenv.hostPlatform.isDarwin || isIdempotent "pkgsLLVM");
+# TODO: This currently results in infinite recursion, even on Linux
+# assert isIdempotent "pkgsLLVMLibc";
+assert isIdempotent "pkgsArocc";
+assert isIdempotent "pkgsZig";
+assert isIdempotent "pkgsLinux";
+assert isIdempotent "pkgsMusl";
+assert isIdempotent "pkgsStatic";
+assert isIdempotent "pkgsi686Linux";
+assert isIdempotent "pkgsx86_64Darwin";
+
+# TODO: fails
+# assert isNoop [ "pkgsStatic" ] "pkgsMusl";
+# TODO: fails because of ppc64-musl
+# assert lib.all (sys: isNoop [ "pkgsCross" sys ] "pkgsMusl") allMuslExamples;
+assert lib.all (sys: isNoop [ "pkgsCross" sys ] "pkgsLLVM") allLLVMExamples;
+
+# TODO: fails
+# assert isComposable "pkgsExtraHardening";
+# assert isComposable "pkgsLLVM";
+# assert isComposable "pkgsLLVMLibc";
+# assert isComposable "pkgsArocc";
+# TODO: unexpected argument 'bintools' - uncomment once https://github.com/NixOS/nixpkgs/pull/331011 is done
+# assert isComposable "pkgsZig";
+# TODO: attribute 'abi' missing
+# assert isComposable "pkgsMusl";
+# TODO: fails
+# assert isComposable "pkgsStatic";
+# assert isComposable "pkgsi686Linux";
+
+# Special cases regarding buildPlatform vs hostPlatform
+# TODO: fails
+# assert discardEvaluationErrors (pkgsCross.gnu64.pkgsMusl.stdenv.hostPlatform.isMusl);
+# assert discardEvaluationErrors (pkgsCross.gnu64.pkgsi686Linux.stdenv.hostPlatform.isx86_32);
+# assert discardEvaluationErrors (pkgsCross.mingwW64.pkgsLinux.stdenv.hostPlatform.isLinux);
+# assert discardEvaluationErrors (pkgsCross.aarch64-darwin.pkgsx86_64Darwin.stdenv.hostPlatform.isx86_64);
+
+# pkgsCross should keep upper cross settings
+# TODO: fails
+# assert discardEvaluationErrors (with pkgsStatic.pkgsCross.gnu64.stdenv.hostPlatform; isGnu && isStatic);
+# assert discardEvaluationErrors (with pkgsLLVM.pkgsCross.musl64.stdenv.hostPlatform; isMusl && useLLVM);
+
+emptyFile
--- a/pkgs/test/top-level/stage.nix
+++ b/pkgs/test/top-level/stage.nix
@@ -94,35 +94,36 @@ assert isIdempotent "pkgsStatic";
 assert isIdempotent "pkgsi686Linux";
 assert isIdempotent "pkgsx86_64Darwin";
 
-# TODO: fails
-# assert isNoop [ "pkgsStatic" ] "pkgsMusl";
-# TODO: fails because of ppc64-musl
-# assert lib.all (sys: isNoop [ "pkgsCross" sys ] "pkgsMusl") allMuslExamples;
+assert isNoop [ "pkgsStatic" ] "pkgsMusl";
+assert lib.all (sys: isNoop [ "pkgsCross" sys ] "pkgsMusl") allMuslExamples;
 assert lib.all (sys: isNoop [ "pkgsCross" sys ] "pkgsLLVM") allLLVMExamples;
 
-# TODO: fails
-# assert isComposable "pkgsExtraHardening";
-# assert isComposable "pkgsLLVM";
+assert isComposable "pkgsExtraHardening";
+assert isComposable "pkgsLLVM";
+# TODO: Results in infinite recursion
 # assert isComposable "pkgsLLVMLibc";
-# assert isComposable "pkgsArocc";
+assert isComposable "pkgsArocc";
 # TODO: unexpected argument 'bintools' - uncomment once https://github.com/NixOS/nixpkgs/pull/331011 is done
 # assert isComposable "pkgsZig";
-# TODO: attribute 'abi' missing
-# assert isComposable "pkgsMusl";
-# TODO: fails
-# assert isComposable "pkgsStatic";
-# assert isComposable "pkgsi686Linux";
+assert isComposable "pkgsMusl";
+assert isComposable "pkgsStatic";
+assert isComposable "pkgsi686Linux";
 
 # Special cases regarding buildPlatform vs hostPlatform
 # TODO: fails
 # assert discardEvaluationErrors (pkgsCross.gnu64.pkgsMusl.stdenv.hostPlatform.isMusl);
 # assert discardEvaluationErrors (pkgsCross.gnu64.pkgsi686Linux.stdenv.hostPlatform.isx86_32);
-# assert discardEvaluationErrors (pkgsCross.mingwW64.pkgsLinux.stdenv.hostPlatform.isLinux);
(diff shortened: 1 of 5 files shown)


K900:
Eval is broken :(

K900:
That said, this is generally good stuff and we should land this (though maybe after 24.05?)

lf-:
I don't have time to review this right now but I just want to say, thank you so much for writing this, and your commit messages are great!

Ericson2314:
Thank for doing a very thorough job documenting your work, @wolfgangwalther, but I am a little worried about passing this information. it was originally very intentional that I transformed the ugly/bad `localSystem` and `crossSystem` to other better arguments things in the process of going from `pkgs/top-level/default.nix` -> `pkgs/top-level/stage.nix`.

I think I see why you are doing this, but another way to look at this is that having the other package sets mashed into `stage.nix` like this is --- while convenient --- and architectural abomination. The package sets are are supposed to be intended, and the proper way to get them is just to call nixpkgs with different arguments. Having some convience example calls is fine, but it was a mistake to ever make this part of the package set. `pkgsCross` being bad makes it unsurprising to me that supporting it forces further architectural imperfections. It's "virally" the wrong thing.

pkgs/os-specific/linux/busybox/sandbox-shell.nix
```diff
@@ -1,9 +1,15 @@
...
 # Minimal shell for use as basic /bin/sh in sandbox builds
 busybox.override {
   enableStatic = true;
   enableMinimal = true;
+
+  useMusl =
+    # GNU hosts use glibc which is difficult to statically link against.
+    # Prefer musl if the host platform supports it.
+    stdenv.hostPlatform.isGnu && lib.meta.availableOn stdenv.hostPlatform musl;
+
```
Ericson2314, line 12: This is a regression of sorts I think, we should just delete `useMusl` in `busybox`

lib/systems/default.nix
```diff
@@ -71,8 +71,6 @@ let
     final = {
       # Prefer to parse `config` as it is strictly more informative.
       parsed = parse.mkSystemFromString (if args ? config then args.config else args.system);
-      # Either of these can be losslessly-extracted from `parsed` iff parsing succeeds.
-      system = parse.doubleFromSystem final.parsed;
       config = parse.tripleFromSystem final.parsed;
```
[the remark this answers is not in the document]
Ericson2314: Adam Joseph's old PR had something to fix the MinGW cases, btw. We should continue just fixing that.

pkgs/test/top-level/default.nix
```diff
@@ -44,4 +44,85 @@ lib.recurseIntoAttrs {
...
+
+      assert isNoop [ "pkgsStatic" ] "pkgsMusl";
+      assert lib.all (sys: isNoop [ "pkgsCross" sys ] "pkgsMusl") allMuslExamples;
+      assert lib.all (sys: isNoop [ "pkgsCross" sys ] "pkgsLLVM") allLLVMExamples;
+
+      assert isComposable "pkgsExtraHardening";
+      assert isComposable "pkgsLLVM";
+      assert isComposable "pkgsArocc";
+      # TODO: unexpected argument 'bintools'
+      # assert isComposable "pkgsZig";
```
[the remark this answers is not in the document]
RossComputerGuy: I believe #331011 might fix it.

aforemny:
Reading through the comments, I feel settling on one of [@wolfgangwalther's proposed solutions](https://github.com/NixOS/nixpkgs/pull/303849#issuecomment-2132304066) would be the way forward.

I very much agree that the current interface suggests arbitrary composability, and if the implementation is not supposed to live up to that suggestion, the interface should be changed to be more clear, for instance by removing nested `pkgsStatic`, `pkgsCross`, or have them throw with an error message detailing how to achieve the intended outcome.

I am personally leaning towards making the current interface work, ie. have all of `pkgsStatic`, `pkgsCross` compose well, as it is very easy to use and would be a backwards-compatible solution.

However, I do not feel qualified making a decision here.

@Ericson2314 could you provide some guidance on progressing here? I think first we should settle on the solution that we want (change the interface or change the implementation), and then see if this PR does address that.

I am interested in this, is there a way I could help out?

pkgs/top-level/stage.nix
```diff
@@ -176,62 +176,80 @@ let
...
+    # helper function to call nixpkgsFun to make sure each package set
+    # references itself as a no-op and to prevent nested pkgsCross.
+    # Nested pkgsCross doesn't make sense fundamentally, because it replaces
+    # the full crossSystem argument, thus starting "from scratch".
+    createPackageSet = name: { overlays ? [], ... }@args:
+      nixpkgsFun (args // {
+        overlays = [
+          (self: super: {
+            "${name}" = super;
+            pkgsCross = throw "Nested pkgsCross not allowed.";
```
aforemny, line 190: Why not allow this if it is well-defined semantically? Otherwise, it would be a breaking change:

```
❯ nom-build default.nix -A pkgsCross.mingwW64.hello
Finished at 13:39:42 after 0s
/nix/store/7k2v0zk1f3aypihalihygm6bzcly8v38-hello-x86_64-w64-mingw32-2.12.1
❯ nom-build default.nix -A pkgsCross.mingwW64.pkgsCross.mingwW64.hello
Finished at 13:39:31 after 0s
/nix/store/7k2v0zk1f3aypihalihygm6bzcly8v38-hello-x86_64-w64-mingw32-2.12.1
```

aforemny: I agree. The following example should NOT produce the same hashes:

```
❯ nix-build default.nix -A pkgsCross.aarch64-multiplatform.pkgsCross.gnu64.hello
/nix/store/qdaj356l85gyahgvlwmj9g955pkpsfgv-hello-2.12.1
❯ nix-build default.nix -A hello
/nix/store/qdaj356l85gyahgvlwmj9g955pkpsfgv-hello-2.12.1
```

r-burns:
> Those attributes exist and don't work well, so leaving them in a broken state is not helpful.

Agree 100%. In an ideal world maybe we'd have a better mechanism for doing this and we wouldn't be using pkgsCross internally at all. But as it is we rely on these and they're broken, so why not improve them a bit? It seems your changes are quite well-contained, so I'm not as worried about the "virality" of this PR.

It seems that a lot of good changes to the cross machinery stall out because it's hard to gain confidence in them. Maybe that's what's happening here? I'd be happy to help out with testing this on a variety of cross-compilation scenarios if that will improve confidence in the implications of this PR.

Mic92:
What is the evaluation memory increase of this pull request? On other pull requests we don't go out of memory.

pwaller:
I can imagine that eval'ing all these cross package sets in one expression could be a problem, since each one does instantiate a lot of stuff and require a lot of memory individually. Perhaps it would be necessary to test these in separate expressions/invocations of the interpreter? Is that possible?

Mic92:
> I can imagine that eval'ing all these cross package sets in one expression could be a problem, since each one does instantiate a lot of stuff and require a lot of memory individually. Perhaps it would be necessary to test these in separate expressions/invocations of the interpreter? Is that possible?

We would have the same issue in hydra, so it would be good, if we can avoid this somehow.

pwaller:
Found a difference on ccb185ad547c8c735cbbfe1cf6b33bf3d4f23d3d:

```
$ nix eval --system x86_64-linux -f . pkgsCross.aarch64-multiplatform.pkgsStatic.hello
«derivation /nix/store/6m5pmwdbw6y8zlk3kkdf0k8ln6kkp23g-hello-static-aarch64-unknown-linux-musl-2.12.1.drv»

$ nix eval --system x86_64-linux -f . pkgsStatic.pkgsCross.aarch64-multiplatform.hello
«derivation /nix/store/l8gjyp3x83s3hffn2dikj6xkalg7hxzb-hello-static-aarch64-unknown-linux-gnu-2.12.1.drv»
```

It seems the libc switch to musl is getting lost, though the -static marker is still present on the package name.

pwaller:
Thanks for clarifying!

pkgs/test/top-level/stage.nix
```diff
@@ -0,0 +1,116 @@
...
+with import ../../top-level { inherit localSystem; };
+
+let
+  # To silence platform specific evaluation errors
+  discardEvaluationErrors =
+    e:
+    let
+      res = builtins.tryEval e;
+    in
+    res.success == res.value;
```
m-bdf, line 18:
```suggestion
  discardEvaluationErrors = e: (builtins.tryEval e).success -> e;
```

If I am not mistaken the two are equivalent for booleans, but the second version is more easily understood and will throw an error if e is not a boolean.

pkgs/test/top-level/stage.nix
```diff
@@ -0,0 +1,116 @@
...
+
+  allLLVMExamples = builtins.attrNames (
+    lib.filterAttrs (
+      _:
+      {
+        useLLVM ? false,
+        ...
+      }:
+      useLLVM
+    ) lib.systems.examples
```
m-bdf, line 46:
```suggestion
    lib.filterAttrs (_: system: system.useLLVM or false) lib.systems.examples
```

More readable than argument destructuring when auto-formatting with nixfmt.

pkgs/test/top-level/stage.nix
```diff
@@ -0,0 +1,116 @@
...
+  # This is very similar to the idempotency test, but not going
+  # via the super' overlay.
+  isNoop =
+    parent: child:
+    discardEvaluationErrors (
+      (lib.getAttrFromPath parent pkgs).stdenv == (lib.getAttrFromPath parent pkgs).${child}.stdenv
+    );
+
+  allMuslExamples = builtins.attrNames (
+    lib.filterAttrs (_: { config, ... }: lib.hasSuffix "-musl" config) lib.systems.examples
```
m-bdf, line 35:
```suggestion
    lib.filterAttrs (_: system: lib.hasSuffix "-musl" system.config) lib.systems.examples
```

m-bdf (approved):

m-bdf:
pkgsMusl seems broken on NixOS:

```nix
$ nix repl github:wolfgangwalther/nixpkgs/compose-pkgs-sets
> system = lib.nixosSystem { system = "x86_64-linux"; modules = []; }
> system.pkgs.pkgsMusl.stdenv.hostPlatform.libc
"glibc"
```

nixpkgs master prints "musl" as expected.

pwaller:
> Not familiar too much with `lib.nixosSystem`, yet. That's flake-specific, right?

It's the same thing as evaluating a nixos system which you can do with `nix-instantiate --eval` from a nixpkgs checkout; I see the same issue there, which is interesting:

```
$ nix-instantiate --eval --arg configuration {} ./nixos -A pkgs.pkgsMusl.stdenv.hostPlatform.libc
glibc
```

pkgs/test/top-level/stage.nix
```diff
@@ -0,0 +1,104 @@
...
+    && discardEvaluationErrors (pkgs.pkgsArocc.${set}.stdenv.hostPlatform.useArocc)
+    && discardEvaluationErrors (pkgs.pkgsZig.${set}.stdenv.hostPlatform.useZig)
+    && discardEvaluationErrors (pkgs.pkgsLinux.${set}.stdenv.buildPlatform.isLinux)
+    && discardEvaluationErrors (pkgs.pkgsMusl.${set}.stdenv.hostPlatform.isMusl)
+    && discardEvaluationErrors (pkgs.pkgsStatic.${set}.stdenv.hostPlatform.isStatic)
+    && discardEvaluationErrors (pkgs.pkgsi686Linux.${set}.stdenv.hostPlatform.isx86_32)
+    && discardEvaluationErrors (pkgs.pkgsx86_64Darwin.${set}.stdenv.hostPlatform.isx86_64);
+in
+
+# Appends same flags over and over again
```
roberth, line 62: What flags? I only see package set selection

pkgs/test/top-level/stage.nix
```diff
@@ -0,0 +1,104 @@
...
+with import ../../top-level { inherit localSystem; };
+
+let
+  # To silence platform specific evaluation errors
+  discardEvaluationErrors = e: (builtins.tryEval e).success -> e;
+
+  # Basic test for idempotency of the package set, i.e:
+  # Applying the same package set twice should work and
+  # not change anything.
+  isIdempotent = set: discardEvaluationErrors (pkgs.${set}.stdenv == pkgs.${set}.${set}.stdenv);
```
roberth, line 18: `stdenv` is a derivation, and `==` on derivations is limited to `outPath`.
So this would be equivalent and avoid confusion:
```suggestion
  isIdempotent = set: discardEvaluationErrors (pkgs.${set}.stdenv.outPath == pkgs.${set}.${set}.stdenv.outPath);
```

However, I think you do want to check more:

```suggestion
  isIdempotent = set: discardEvaluationErrors (pkgs.${set}.stdenv.outPath == pkgs.${set}.${set}.stdenv.outPath
    && pkgs.${set}.stdenv.hostPlatform == pkgs.${set}.${set}.stdenv.hostPlatform
    && pkgs.${set}.stdenv.buildPlatform == pkgs.${set}.${set}.stdenv.buildPlatform
  );
```

pkgs/test/top-level/stage.nix
```diff
@@ -0,0 +1,104 @@
...
+  localSystem ? {
+    system = builtins.currentSystem;
+  },
+}:
+
+with import ../../top-level { inherit localSystem; };
+
+let
+  # To silence platform specific evaluation errors
+  discardEvaluationErrors = e: (builtins.tryEval e).success -> e;
```
roberth, line 13: Extensive use of this makes the test less effective.

pkgs/top-level/stage.nix
```diff
@@ -176,155 +176,104 @@ let
...
+      crossSystem =
+        (lib.systems.systemToAttrs (lib.defaultTo prevArgs.localSystem prevArgs.crossSystem or null)) // crossAttrs;
+    });
+    # This is only cross when we are already cross, otherwise local.
+    # For the case of "native cross", i.e. pkgsCross.gnu64 on a x86_64-linux system, we need to adjust **both**
+    # localSystem **and** crossSystem, otherwise they're out of sync.
+    mkHybridPkgs = name: hybridAttrs: mkPkgs name (prevArgs: let
+        newSystem = (lib.systems.systemToAttrs (lib.defaultTo prevArgs.localSystem prevArgs.crossSystem or null)) // hybridAttrs;
+      in lib.optionalAttrs (stdenv.hostPlatform == stdenv.buildPlatform) { localSystem = newSystem; }
+      // lib.optionalAttrs (prevArgs ? crossSystem) { crossSystem = newSystem; }
```
roberth, line 199: Cross-ness is not an intrinsic, but derived property of a build+host configuration.
By differentiating on the presence of `crossSystem`, we're extending the configuration space from build+host to build+host+cross, which complicates the interface and/or user interface of Nixpkgs as a function.

Concretely, a configuration where the `localSystem` and `crossSystem` are identical values should be indistinguishable from a configuration where only `localSystem` was passed.
I believe this is currently not entirely the case, which is a bug.

This particularly line with expression `prevArgs?crossSystem` seems either affected by or prone to that bug.

pkgs/top-level/stage.nix
```diff
@@ -176,155 +176,104 @@ let
...
-  # - pkgsCross.<system> where system is a member of lib.systems.examples
-  # - pkgsMusl
-  # - pkgsi686Linux
-  otherPackageSets = self: super: {
+  # these will instantiate a new version of allPackages.
+  otherPackageSets = let
+    mkPkgs = name: fn: nixpkgsFun (prevArgs: let nixpkgsArgs = fn prevArgs; in nixpkgsArgs // {
+      overlays = [
+        (self': super': {
+          "${name}" = super';
```
m-bdf, line 184: Is this overlay still needed now that the package sets are composable?

It forces accessing an already accessed package set to be a no-op, which it now already is without the overlay for simple cases (e.g. `pkgsStatic.pkgsStatic`), and I believe is not desirable for more complex cases (e.g. when using `buildPackages`):

```nix
> pkgs.pkgsStatic.buildPackages.pkgsStatic.stdenv.hostPlatform.libc
"glibc"
```

m-bdf: I've found that without the overlay above, using a package set in an overlay stack overflows:

```nix
> (pkgs.extend (final: prev: { inherit (final.pkgsStatic) hello; })).hello
error: stack overflow; max-call-depth exceeded
```

So a mechanism to avoid infinite recursion is indeed still needed, however I think the current one is not ideal, as it leads to unwanted no-ops (see the example in my previous comment).

K900:
This broke container tests, and probably everything else that uses stub-ld: https://hydra.nixos.org/build/286424899

PR #303865, opened 2024-04-13
sourcegit: init at 2025.1

### LINKED ISSUE #303855: Package request: SourceGit
**Project description**

Modern cross-platform Git GUI built with C#.

* Supports Windows/macOS/Linux
* Opensource/Free
* Fast
* English/简体中文
* Built-in light/dark themes
* Visual commit graph
* Supports SSH access with each remote
* GIT commands with GUI
  * Clone/Fetch/Pull/Push...
  * Branches
  * Remotes
  * Tags
  * Stashes
  * Submodules
  * Archive
  * Diff
  * Save as patch/apply
  * File histories
  * Blame
  * Revision Diffs
* GitFlow support

**Metadata**

* homepage URL: https://github.com/sourcegit-scm/sourcegit
* source URL: https://github.com/sourcegit-scm/sourcegit
* license: mit
* platforms: linux, darwin

---

TomaSajt:
## Description of changes

Closes https://github.com/NixOS/nixpkgs/issues/303855
Related: https://github.com/sourcegit-scm/sourcegit/issues/71

This PR adds 1 package: `sourcegit`

The app is mostly usable on Linux, however I don't know how good it is on Darwin

There's a patch that makes the platform specific logic for finding git on MacOS the same as on Linux (a.k.a. search the PATH)

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/by-name/so/sourcegit/package.nix
+++ b/pkgs/by-name/so/sourcegit/package.nix
@@ -1,0 +1,117 @@
+{
+  lib,
+  stdenv,
+  buildDotnetModule,
+  fetchFromGitHub,
+  dotnetCorePackages,
+  copyDesktopItems,
+  makeDesktopItem,
+  libicns,
+
+  libXcursor,
+  libXext,
+  libXi,
+  libXrandr,
+
+  git,
+  xdg-utils,
+}:
+
+buildDotnetModule (finalAttrs: {
+  pname = "sourcegit";
+  version = "2025.01";
+
+  src = fetchFromGitHub {
+    owner = "sourcegit-scm";
+    repo = "sourcegit";
+    tag = "v${finalAttrs.version}";
+    hash = "sha256-6FO0WRUcWm7xnuw6Br97+cWZhvzIOesg/eANZzeyxZo=";
+  };
+
+  patches = [ ./fix-darwin-git-path.patch ];
+
+  dotnet-sdk = dotnetCorePackages.sdk_9_0;
+  dotnet-runtime = dotnetCorePackages.runtime_9_0;
+
+  nugetDeps = ./deps.json;
+
+  projectFile = [ "src/SourceGit.csproj" ];
+
+  executables = [ "SourceGit" ];
+
+  dotnetFlags = [
+    "-p:DisableUpdateDetection=true"
+    "-p:DisableAOT=true"
+  ];
+
+  nativeBuildInputs = lib.optionals stdenv.hostPlatform.isLinux [
+    copyDesktopItems
+    libicns
+  ];
+
+  # these are dlopen-ed at runtime
+  # libXi is needed for right-click support
+  # not sure about what the other ones are needed for, but I'll include them anyways
+  runtimeDeps = [
+    libXcursor
+    libXext
+    libXi
+    libXrandr
+  ];
+
+  # Note: users can use `.overrideAttrs` to append to this list
+  runtimePathDeps = [
+    git
+    xdg-utils
+  ];
+
+  # add fallback binaries to use if the user doesn't already have them in their PATH
+  preInstall = ''
+    makeWrapperArgs+=(
+      --suffix PATH : ${lib.makeBinPath finalAttrs.runtimePathDeps}
+    )
+  '';
+
+  desktopItems = [
+    (makeDesktopItem {
+      name = "SourceGit";
+      exec = "SourceGit";
+      icon = "SourceGit";
+      desktopName = "SourceGit";
+      terminal = false;
+      comment = finalAttrs.meta.description;
+    })
+  ];
+
+  postInstall =
+    lib.optionalString stdenv.hostPlatform.isLinux ''
+      # extract the .icns file into multiple .png files
+      # where the format of the .png file names is App_"$n"x"$n"x32.png
+
+      icns2png -x build/resources/app/App.icns
+
+      for f in App_*x32.png; do
+        res=''${f//App_}
+        res=''${res//x32.png}
+        install -Dm644 $f "$out/share/icons/hicolor/$res/apps/SourceGit.png"
+      done
+    ''
+    + lib.optionalString stdenv.hostPlatform.isDarwin ''
+      install -Dm644 build/resources/app/App.icns $out/Applications/SourceGit.app/Contents/Resources/App.icns
+
+      substitute build/resources/app/App.plist $out/Applications/SourceGit.app/Contents/Info.plist \
+        --replace-fail "SOURCE_GIT_VERSION" "${finalAttrs.version}"
+
+      mkdir -p $out/Applications/SourceGit.app/Contents/MacOS
+      ln -s $out/bin/SourceGit $out/Applications/SourceGit.app/Contents/MacOS/SourceGit
+    '';
+
+  meta = {
+    changelog = "https://github.com/sourcegit-scm/sourcegit/releases/tag/${finalAttrs.src.tag}";
+    description = "Free & OpenSource GUI client for GIT users";
+    homepage = "https://github.com/sourcegit-scm/sourcegit";
+    license = lib.licenses.mit;
+    mainProgram = "SourceGit";
+    maintainers = with lib.maintainers; [ tomasajt ];
+  };
+})
--- a/pkgs/by-name/so/sourcegit/deps.json
+++ b/pkgs/by-name/so/sourcegit/deps.json
@@ -1,0 +1,232 @@
+[
+  {
+    "pname": "Avalonia",
+    "version": "11.0.0",
+    "hash": "sha256-7QE0MtD1QDiG3gRx5xW33E33BXyEtASQSw+Wi3Lmy3E="
+  },
+  {
+    "pname": "Avalonia",
+    "version": "11.2.3",
+    "hash": "sha256-NUoyXJkIsgbkcKFVb10VRafM4ViHs801c/7vhu3ssUY="
+  },
+  {
+    "pname": "Avalonia.Angle.Windows.Natives",
+    "version": "2.1.22045.20230930",
+    "hash": "sha256-RxPcWUT3b/+R3Tu5E5ftpr5ppCLZrhm+OTsi0SwW3pc="
+  },
+  {
+    "pname": "Avalonia.AvaloniaEdit",
+    "version": "11.1.0",
+    "hash": "sha256-K9+hK+4aK93dyuGytYvVU25daz605+KN54hmwQYXFF8="
+  },
+  {
+    "pname": "Avalonia.BuildServices",
+    "version": "0.0.29",
+    "hash": "sha256-WPHRMNowRnYSCh88DWNBCltWsLPyOfzXGzBqLYE7tRY="
+  },
+  {
+    "pname": "Avalonia.Controls.ColorPicker",
+    "version": "11.2.3",
+    "hash": "sha256-z3ZHxVSOoOjqq+5G71jnGN1Y0i3YpAkox7cj3lNr6kg="
+  },
+  {
+    "pname": "Avalonia.Controls.DataGrid",
+    "version": "11.2.3",
+    "hash": "sha256-jIJvuYN0iym/WeOC0C7z5xj5kCZSXGoeLQ/q5qQfewM="
+  },
+  {
+    "pname": "Avalonia.Desktop",
+    "version": "11.2.3",
+    "hash": "sha256-srtZi+kDbhRtMl33l91zssBWETU5oHodKbbWyfEsb/I="
+  },
+  {
+    "pname": "Avalonia.Diagnostics",
+    "version": "11.2.3",
+    "hash": "sha256-DIGkaBff+C3BLwedw5xteR5lfzb6ecxiLt12eJVgLQc="
+  },
+  {
+    "pname": "Avalonia.Fonts.Inter",
+    "version": "11.2.3",
+    "hash": "sha256-ySsCXVpjqjCX/uYkwluSfrAoBtuq9k7fC1bFjxKC9/Q="
+  },
+  {
+    "pname": "Avalonia.FreeDesktop",
+    "version": "11.2.3",
+    "hash": "sha256-3sNemBmZE06w2ul87T5HrEeHUxXMOa9MfQhpI4AoxDY="
+  },
+  {
+    "pname": "Avalonia.Native",
+    "version": "11.2.3",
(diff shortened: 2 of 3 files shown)


julius-boettger:
I just tried to test if all those `runtimeDeps` are really necessary, and it seems to run fine without `libXi`, `libXext`, `libXrandr` and `glew`. My testing was more rough than thorough, so I can't promise that removing these `runtimeDeps` really doesn't change anything, but I thought it could be of interest.

TomaSajt:
> I just tried to test if all those `runtimeDeps` are really necessary, and it seems to run fine without `libXi`, `libXext`, `libXrandr` and `glew`. My testing was more rough than thorough, so I can't promise that removing these `runtimeDeps` really doesn't change anything, but I thought it could be of interest.

Yeah, looks like this works for me too. (I just copied this list from another derivation using AvaloniaUI)

julius-boettger:
`libXi` is definitely necessary, right-click-menus don't work without it. But I haven't figured out what `libXext` and `libXrandr` do here yet.

TomaSajt:
Tried to update to the latest version (8.26) however it looks like something broke in a commit that bumps used the AvaloniaUI version (https://github.com/sourcegit-scm/sourcegit/commit/ef0cf7657183a1ddf20387bcd262be7f52e89b8c) (Part of 8.22->8.23 changes)

TomaSajt:
Seems like https://github.com/NixOS/nixpkgs/pull/340493 fixed the Avalonia issue
Also, the latest version of the app no longer uses the version of Avalonia that had the issue, so it doesn't really matter anyway.

GAsplund:
I actually tried to create my own version of this package, but using their provided AppImage instead, and that worked really well without any real issues. Maybe that would be a better approach, at least for Linux? Granted, it would be more difficult to make it work on Darwin.

TomaSajt:
> I actually tried to create my own version of this package, but using their provided AppImage instead, and that worked really well without any real issues. Maybe that would be a better approach, at least for Linux? Granted, it would be more difficult to make it work on Darwin.

You may use the AppImage version, if you want to, however, if possible, we should package things from source.
Is there something that does not work for you with the source based package?

GAsplund:
> > I actually tried to create my own version of this package, but using their provided AppImage instead, and that worked really well without any real issues. Maybe that would be a better approach, at least for Linux? Granted, it would be more difficult to make it work on Darwin.
> 
> You may use the AppImage version, if you want to, however, if possible, we should package things from source. Is there something that does not work for you with the source based package?

My main consideration was whether wrapping the AppImage or building from source would be the most effective in terms of maintenance and usability, but ultimately it might be better to build directly from source to be compatible with more platforms. I will admit however, I have not tried this since the 8.8 init, for which I only recall the issues with right-click menus

DocBrown101:
Is the only thing missing here the formatting of the deps.nix or why does it not continue?

```
Please go to the Nixpkgs root directory, run `nix-shell`, then:
nixfmt 'pkgs/by-name/so/sourcegit/deps.nix'
```

TomaSajt:
> deps.nix

I don't think there was a decision made yet about whether or not we should keep generated file formatted, so this is not a blocker.

I am still waiting for someone to review/merge this PR.
I can send this again into the review requests Matrix channel.

pkgs/by-name/so/sourcegit/package.nix
```diff
@@ -0,0 +1,87 @@
...
+}:
+
+let
+  binPath = lib.makeBinPath [
+    git
+    xdg-utils
+  ];
+in
+
+buildDotnetModule rec {
```
SuperSandro2000, line 22:
```suggestion

buildDotnetModule rec {
```

pkgs/by-name/so/sourcegit/package.nix
```diff
@@ -0,0 +1,87 @@
...
+      icon = "SourceGit";
+      desktopName = "SourceGit";
+      terminal = false;
+      comment = meta.description;
+    })
+  ];
+
+  preInstall = ''
+    makeWrapperArgs+=(
+      --suffix PATH : ${binPath}
```
SuperSandro2000, line 66:
```suggestion
      --suffix PATH : ${lib.makeBinPath [ git xdg-utils ]}
```

TomaSajt: nixfmt splits this into multiple lines, that's why I extracted it. It doesn't look terrible as a multiline, but it's not pretty.
If you think it's better I can move the `let in` above the wrapping logic.

TomaSajt: Also, what's your opinion about this being suffix vs prefix?

DocBrown101:
Probably the nix code has to be formatted manually until the generator can generate valid nix code.
FYI: @fpletz

pkgs/by-name/so/sourcegit/package.nix
SuperSandro2000: I would recommend to not hack around the formatting by moving things into let ins and similar.

I would suffix it to always prefer local available tools or git with more functions or scripts around xdg-open for other environment.

TomaSajt: Thanks for the feedback.
I don't completely agree with not having formatting hacks: having a multi-line expression interpolated into a multi-line string is a bit difficult to parse with your head. Though in this case it's not unreadable, so I did went with it.

TomaSajt:
> Probably the nix code has to be formatted manually until the generator can generate valid nix code.

I don't think we've come to a conclusion about formatting generated nix files.
Even if the CI doesn't pass it can be merged.

pitkling:
Thanks for the PR! Just a note here: I tested this on Darwin. SourceGit builds fine and starts without problems. However, it does not show the differences introduced by current changes or when selecting two different commits (it just says `NO CHANGES OR ONLY EOL CHANGES`).

It seems that is an upstream problem of version 8.30.1, since I got the same behavior with the upstream release from their GitHub page. The current upstream release (8.41) shows the diffs correctly. So it might be worth bumping the version used in this PR (in the hope that it's just a simple version bump).

pitkling:
I tested the version bump to 8.41 in a [personal fork](https://github.com/pitkling/nixpkgs/commit/0313457510182a1d51a96905cbf4a170f4e3e743), went without hiccups. Diff view is now working fine on Darwin 🙂.

pitkling:
FYI: I bumped the version in [my fork](https://github.com/pitkling/nixpkgs/tree/sourcegit-init) of this PR to 8.44 and disabled automatic update detection (which is a new option in 8.44). Also, I added the creation of a proper Darwin application bundle.

@TomaSajt: Feel free to integrate those changes if you want to (the changes are fairly minimal). If not that's also fine, then I'll create a corresponding PR once this PR is accepted.

TomaSajt:
> FYI: I bumped the version in [my fork](https://github.com/pitkling/nixpkgs/tree/sourcegit-init) of this PR to 8.44 and disabled automatic update detection (which is a new option in 8.44). Also, I added the creation of a proper Darwin application bundle.
> 
> @TomaSajt: Feel free to integrate those changes if you want to (the changes are fairly minimal). If not that's also fine, then I'll create a corresponding PR once this PR is accepted.

Thanks for your work! :) I'll update my PR

oddlama (approved):
## `nixpkgs-review` result

Command: `nixpkgs-review pr 303865`

---
### `x86_64-linux`
<details>
  <ul>
    <li>sourcegit</li>
  </ul>
</details>

pkgs/by-name/so/sourcegit/package.nix
```diff
@@ -0,0 +1,109 @@
...
+      icns2png -x build/resources/App.icns
+      for f in App_*x32.png; do
+          res=''${f//App_}
+          res=''${res//x32.png}
+          install -Dm644 $f "$out/share/icons/hicolor/$res/apps/SourceGit.png"
+      done
+    ''
+    + lib.optionalString stdenv.hostPlatform.isDarwin ''
+      install -Dm644 build/resources/app/App.icns $out/Applications/SourceGit.app/Contents/Resources/App.icns
+      sed "s/SOURCE_GIT_VERSION/${version}/g" build/resources/app/App.plist > $out/Applications/SourceGit.app/Contents/Info.plist
```
r-vdp, line 96: This will make the version very annoying to override. Would you mind using the `finalAttrs` pattern here such that we can override `version` and `src` and have the rest of the expression use the right values automatically?

TomaSajt: AFAIK buildDotnetModule doesnt support finalAttrs

TomaSajt: Also, this is just some metadata and not the end of the world if it becomes outdated.

TomaSajt: But, now that think about it: we could use $version (bash variable substitution) instead which would essentially be the same as finalAttrs

r-vdp: It does support it ;) 

https://github.com/NixOS/nixpkgs/pull/331398

TomaSajt: Couldn't resist looking at the implementation and leaving some feedback: https://github.com/NixOS/nixpkgs/pull/331398#discussion_r1903378702

I have also tried to add a "good" fixed-point system for `buildRustPackage`, and I'm still searching for an "ideal" solution.

DimitarNestorov:
Seems to be working on Darwin. BTW, 2025.1 was just released https://github.com/sourcegit-scm/sourcegit/releases/tag/v2025.01

image

DimitarNestorov (approved):
LGTM
image

pkgs/by-name/so/sourcegit/package.nix
```diff
@@ -0,0 +1,112 @@
...
+
+      substitute build/resources/app/App.plist $out/Applications/SourceGit.app/Contents/Info.plist \
+        --replace-fail "SOURCE_GIT_VERSION" "${finalAttrs.version}"
+
+      mkdir -p $out/Applications/SourceGit.app/Contents/MacOS
+      ln -s $out/bin/SourceGit $out/Applications/SourceGit.app/Contents/MacOS/SourceGit
+    '';
+
+  meta = {
+    changelog = "https://github.com/sourcegit-scm/sourcegit/releases/tag/${finalAttrs.src.tag}";
```
MattSturgeon, line 105: This would break if someone overrides the package with a `src` that doesn't have a `tag`

Maybe that's not an issue, because `meta.changelog` is probably only read for in-tree packages, and is likely unused for out-of-tree overrides.

I think using `"v${finalAttrs.version}"` would be ok though.

pkgs/by-name/so/sourcegit/package.nix
```diff
@@ -0,0 +1,112 @@
...
+          xdg-utils
+        ]
+      }
+    )
+  '';
+
+  postInstall =
+    lib.optionalString stdenv.hostPlatform.isLinux ''
+      icns2png -x build/resources/App.icns
+      for f in App_*x32.png; do
```
MattSturgeon, line 88: Why `App_*x32.png` here? Shouldn't it be `App_*.png`, else it'll only match `App_32x32.png` right?

TomaSajt: It looks like the command was silently failing, since I needed to use the `build/resources/app/App.icns`
(why does the command silently fail...)

After that, running `ls *.png` gives this:
```
App_128x128x32.png  App_32x32x32.png    App_64x64x32.png
App_256x256x32.png  App_512x512x32.png
```

pretty weird format, so I'll write a comment explaining the format

pkgs/by-name/so/sourcegit/package.nix
TomaSajt: The same could be said for using `src.rev` as well, and that's pretty common

IMO it's fine, since we're not evaling `meta` anyways.

MattSturgeon: > The same could be said for using `src.rev` as well, and that's pretty common

I've normally seen that done _without_ using `finalAttrs`, so it wouldn't be affected by overrides.

I haven't managed to find any guidance on when we should use fixed vs overridable references to attrs.

Instead, I usually see strong opinions banded around without any real explanation or justification 😥

TomaSajt: Yeah, working fixed-point packages is a bit annoying, especially when dealing with FOD hashes not being invalidated (because of the default `name = "source"`).

I don't know how much fixed-point packages worsen the eval performance, but I like them otherwise.

The main issue is that we are in a state inbetween totally fixed-point package definitions and (mostly) static package definitions. And new packages are being written in both styles every day.

It doesn't help that most wrappers for mkDerivation ATM don't support passing in a fixed-point attrset.

DocBrown101:
What is currently blocking the merge? I would be very happy about this package!

TomaSajt:
> What is currently blocking the merge?

Nothing, really.

---

@MattSturgeon could you re-review?

In case there are no problems I'll self-merge.

pkgs/by-name/so/sourcegit/package.nix
```diff
@@ -0,0 +1,116 @@
...
+  ];
+
+  # add fallback binaries to use if the user doesn't already have them in their PATH
+  preInstall = ''
+    makeWrapperArgs+=(
+      --suffix PATH : ${
+        lib.makeBinPath [
+          git
+          xdg-utils
+        ]
```
MattSturgeon, line 80: I'd prefer this use an overridable attr, e.g. `makeBinPath finalAttrs.runtimeInputs` (or `dependencies`, or something)?

This could make customization easier for end-users.

MattSturgeon (approved):
Sorry, I held off approving as I'm not familiar with the package and don't have time to test it out.

What I understand of the diff looks fine to me though.

pkgs/by-name/so/sourcegit/package.nix
TomaSajt: Might be useful, especially since on darwin, the PATH isn't always inherited when running the program.

TomaSajt:
Added an `overrideAttrs`-able list for path-deps

MattSturgeon (approved):

PR #303927, opened 2024-04-14
buildNeovimPlugin: apply luarocks -> plugin conversion to dependencies

mrcjkb:
## Description of changes

Some luarocks packages are built with custom build backends (e.g. luarocks-build-rust-mlua).
When `buildNeovimPlugin` overrides such a lua derivation's config to use a flat hierarchy, this breaks packages with a custom build backend, which can be fixed by applying the change to all `nativeBuildInputs`.

Open questions (cc @teto @GaetanLepage):

- [ ] We probably want to only apply the overrie to luarocks packages (some `nativeBuildInputs` might be non-luarocks packages - `toml-edit` is an example).
  Which attribute makes most sense to check for? Maybe `env.LUAROCKS_CONFIG`?
- [ ] I've also applied the override to the `propagatedBuildInputs` (should also be selectively applied to luarocks packages).
  Does this make sense? If yes, it should probably be applied to all `propagatedBuildInputs` recursively...
- [ ] Does this work with native lua libraries? (To be tested...)

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/applications/editors/neovim/build-neovim-plugin.nix
+++ b/pkgs/applications/editors/neovim/build-neovim-plugin.nix
@@ -19,14 +19,31 @@ in
     let
       originalLuaDrv = lua.pkgs.${luaAttr};
 
+      extraConfig = ''
+        -- to create a flat hierarchy
+        lua_modules_path = "lua"
+      '';
+
       luaDrv = originalLuaDrv.overrideAttrs (oa: {
         version = attrs.version or oa.version;
         rockspecVersion = oa.rockspecVersion;
 
-        extraConfig = ''
-          -- to create a flat hierarchy
-          lua_modules_path = "lua"
-        '';
+        # the flat hierarchy needs to be applied to build backends,
+        # otherwise luarocks won't find them.
+        nativeBuildInputs = map (drv: drv.overrideAttrs (_: {
+          inherit extraConfig;
+        })) (oa.nativeBuildInputs or []);
+
+        # TODO: This should probably be applied recursively to luarocks packages
+        propagatedBuildInputs = map (drv: drv.overrideAttrs (oa': {
+          inherit extraConfig;
+          nativeBuildInputs = oa'.nativeBuildInputs or [] ++ [
+            lua.pkgs.luarocksMoveDataFolder
+          ];
+        })) (oa.propagatedBuildInputs or []);
+
+        inherit extraConfig;
+
       });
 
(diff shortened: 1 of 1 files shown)


mrcjkb:
> what nix-build command can trigger the issue ?

I've encountered this issue when attempting to package [orgmode](https://luarocks.org/modules/kristijanhusak/orgmode), which depends on [tree-sitter-orgmode](https://luarocks.org/modules/kristijanhusak/tree-sitter-orgmode), which has [luarocks-build-treesitter-parser](https://luarocks.org/modules/neorocks/luarocks-build-treesitter-parser) as a build dependency.

teto:
I dont think this will be necessary after https://github.com/NixOS/nixpkgs/pull/322145 but let's see. Is your orgmode attempt public ? I could give it a try

mrcjkb:
> I dont think this will be necessary after #322145 but let's see. Is your orgmode attempt public ? I could give it a try

Nope, I hadn't opened a PR yet.
I'd have to dig it up again.

mrcjkb:
Linking: 

- https://github.com/NixOS/nixpkgs/pull/356517

pkgs/applications/editors/neovim/build-neovim-plugin.nix
```diff
@@ -19,14 +19,31 @@ in
...
+        nativeBuildInputs = map (drv: drv.overrideAttrs (_: {
+          inherit extraConfig;
+        })) (oa.nativeBuildInputs or []);
+
+        # TODO: This should probably be applied recursively to luarocks packages
+        propagatedBuildInputs = map (drv: drv.overrideAttrs (oa': {
+          inherit extraConfig;
+          nativeBuildInputs = oa'.nativeBuildInputs or [] ++ [
+            lua.pkgs.luarocksMoveDataFolder
+          ];
```
khaneliman, line 42: This seems we need to selectively apply ? I was trying to change out locally and it was causing luajit to fail to build.

mrcjkb: Hmm, I guess so, since luarocks doesn't actually have official support for luajit, afaict (there's no `--lua-version jit` option, for instance).

PR #304033, opened 2024-04-14
wrapCC, wrapBintools, stdenv: make Nix code a bit more consistent

sternenseemann:
## Description of changes

Simple cleanups that aim to make the stdenv/wrapper code more consistent and less confusing. These should be easy to review and uncontroversial. ~~A follow-up change concerning darwin is noted as `FIXME`, but I think that's better reviewed separately.~~ See #304034.

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/stdenv/linux/default.nix
+++ b/pkgs/stdenv/linux/default.nix
@@ -328,12 +328,12 @@ in
     assert isBuiltByBootstrapFilesCompiler prevStage.patchelf;
     stageFun prevStage {
       name = "bootstrap-stage-xgcc";
-      overrides = final: prev: {
+      overrides = self: super: {
         inherit (prevStage) ccWrapperStdenv coreutils gnugrep gettext bison texinfo zlib gnum4 perl patchelf;
         ${localSystem.libc} = getLibc prevStage;
-        gmp      = prev.gmp.override { cxx = false; };
+        gmp = super.gmp.override { cxx = false; };
         gcc-unwrapped =
-          (prev.gcc-unwrapped.override (commonGccOverrides // {
+          (super.gcc-unwrapped.override (commonGccOverrides // {
             # The most logical name for this package would be something like
             # "gcc-stage1".  Unfortunately "stage" is already reserved for the
             # layers of stdenv, so using "stage" in the name of this package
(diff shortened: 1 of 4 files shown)


pkgs/stdenv/linux/default.nix
```diff
@@ -328,12 +328,12 @@ in
     assert isBuiltByBootstrapFilesCompiler prevStage.patchelf;
     stageFun prevStage {
       name = "bootstrap-stage-xgcc";
-      overrides = final: prev: {
```
Ericson2314, line 331: I thought some people were trying to move away from `self` `super`?

Ericson2314 (commented):
The rest looks good!

pkgs/stdenv/linux/default.nix
sternenseemann: See my commit message:

> As of late, `final: prev: …` for overlays has become more prevalent in
> newer code. This is also exhibited in some code (presumably added
> recently) in stdenv. This change is not about any merits or demerits of
> any naming convention, but rather aims to make the nomenclature in
> stdenv bootstrapping consistent to lessen confusion.
>
> I've chosen to stick to `self: super: …` convention because:
>
> 1. It is more common in the code as it stands.
>
> 2. Using `final: prev: …` makes the code more confusing, als it causes
>    `prev` to be in scope alongside `prevStage`. `prevStage` actually
>    bears no relation to `prev` even though their naming suggests it,
>    making it easy to confuse them (the former is the (final) package set
>    of the previous stage while the latter is just the `prev`/`super` of
>    the overlay “chaining” on a completely fresh package set, i.e. `prev`
   doesn't even relate to the previous stage's `overrides` argument).

I'd be open to change it at a later date, though.

Ericson2314: OK that's fair enough reasoning for me. If the bootstrap is finally made non-bad (not so many stages, uses cross stuff), we can revisit these things.

PR #293768, opened 2024-03-06
python3Packages.onnxruntime: fix build with cudaSupport

NiklasGollenstede:
## Description of changes

Another follow-up to #260745 (and #291874). (ping @autrimpo)

The related python package failed to build (for me on `nixos-unstable`). I don't actually recall whether it failed during the build itself or some of the tests.

Anyway, I guessed a bit and this change (adding some of the `buildInputs` of `onnxruntime` itself to the python package) fixed the build (and any tests it may or may not run).

The python module so far did not (need to?) know about `cudaSupport` (but failed to build if `onnxruntime` was built with that).
I decided to leave it at that, and inherit the properties (`cudaSupport` and `cudaPackages`) from the base `onnxruntime`(`.passthru`). Not sure if that is the correct choice.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
- For non-Linux: Is sandboxing enabled in `nix.conf`? (See [Nix manual](https://nixos.org/manual/nix/stable/command-ref/conf-file.html))
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:
  - [NixOS test(s)](https://nixos.org/manual/nixos/unstable/index.html#sec-nixos-tests) (look inside [nixos/tests](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests))
  - and/or [package tests](https://nixos.org/manual/nixpkgs/unstable/#sec-package-tests)
  - or, for functions and "core" functionality, tests in [lib/tests](https://github.com/NixOS/nixpkgs/blob/master/lib/tests) or [pkgs/test](https://github.com/NixOS/nixpkgs/blob/master/pkgs/test)
  - made sure NixOS tests are [linked](https://nixos.org/manual/nixpkgs/unstable/#ssec-nixos-tests-linking) to the relevant packages
- [ ] Tested compilation of all packages that depend on this change using `nix-shell -p nixpkgs-review --run "nixpkgs-review rev HEAD"`. Note: all changes have to be committed, also see [nixpkgs-review usage](https://github.com/Mic92/nixpkgs-review#usage)
- [ ] Tested basic functionality of all binary files (usually in `./result/bin/`)
[...]

PR #206742, opened 2022-12-18
nativeCheckInputs

symphorien:
###### Description of changes

Fixes https://github.com/NixOS/nixpkgs/issues/161570

tl;dr: checkInputs used to be added to nativeBuildInputs, which means they are added to PATH but one cannot link to them when strictDeps is set. As linking to check-only libs is possible (quite common for ocaml for example) we need to distinguish
* checkInputs, added to buildInputs when doCheck is set. One can link against them.
* nativeCheckInputs, added to nativeBuildInputs when doCheck is set. They are added to PATH.
(same for installCheck)

The PR contains manual commits to adapt the builders (mkDerivation, buildPythonPackage notably), a commit to use the new feature in ocamlPackages.batteries, and a treewide search-and-replace from checkInputs to nativeCheckInputs.

**as a result hashes are unchanged** except for the reverse-dependencies of ocamlPackages.batteries (8).

Supersedes https://github.com/NixOS/nixpkgs/pull/185406/files

This PR will inevitably accumulate merge conflicts. When it obtains enough positive reviews, I will regenerate the treewide commit and merge immediately. Maybe this can be done just after staging-next is merged?

###### Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
- [ ] For non-Linux: Is `sandbox = true` set in `nix.conf`? (See [Nix manual](https://nixos.org/manual/nix/stable/command-ref/conf-file.html))
- [ ] Tested, as applicable:
  - [NixOS test(s)](https://nixos.org/manual/nixos/unstable/index.html#sec-nixos-tests) (look inside [nixos/tests](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests))
  - and/or [package tests](https://nixos.org/manual/nixpkgs/unstable/#sec-package-tests)
  - or, for functions and "core" functionality, tests in [lib/tests](https://github.com/NixOS/nixpkgs/blob/master/lib/tests) or [pkgs/test](https://github.com/NixOS/nixpkgs/blob/master/pkgs/test)
  - made sure NixOS tests are [linked](https://nixos.org/manual/nixpkgs/unstable/#ssec-nixos-tests-linking) to the relevant packages
[...]

PR #304069, opened 2024-04-14
onnxruntime: CUDA fixes

autrimpo:
## Description of changes

Fixes for onnxruntime with CUDA from #293768 and helping various people build the package on "normal" hardware.

#293768 adds missing inputs for the python package to successfully auto-patchelf the onnxruntime shared library (thank you @NiklasGollenstede!).

I've moved `gtest` back to its original place - it is only needed for the tests. We disable *running* them for CUDA builds, so the dependency was not being pulled, but it is still required for *building* them. This is now handled correctly - disabling `checkPhase`  also disables building the tests, which should also slightly speed up the build.

Finally, I've set a limit for nvcc threads to 1 - upstream uses a [function in their build script](https://github.com/microsoft/onnxruntime/blob/7ec51f0a13d5d8cbd796de33276bf04210ce6176/tools/ci_build/build.py#L930) to figure out a reasonable value. Given that they expect >= 4GB RAM *per thread*, I don't think trying to tie it to `NIX_BUILD_CORES` is a good idea, as that would eat up all the RAM on most "consumer" machines (8 cores are not that uncommon, 64GB+ RAM on such machines is). In my testing with a couple of builds, I did not notice a significant slowdown with this set, I'd even say it was slightly faster for some reason. The great benefit is many more people should now be able to safely build this.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/development/libraries/onnxruntime/default.nix
+++ b/pkgs/development/libraries/onnxruntime/default.nix
@@ -224,6 +224,7 @@ effectiveStdenv.mkDerivation rec {
   '';
 
   passthru = {
+    inherit cudaSupport cudaPackages; # for the python module
     protobuf = protobuf_21;
     tests = lib.optionalAttrs pythonSupport {
       python = python3Packages.onnxruntime;
--- a/pkgs/development/libraries/onnxruntime/default.nix
+++ b/pkgs/development/libraries/onnxruntime/default.nix
@@ -133,7 +133,6 @@ effectiveStdenv.mkDerivation rec {
     nlohmann_json
     microsoft-gsl
   ] ++ lib.optionals pythonSupport (with python3Packages; [
-    gtest
     numpy
     pybind11
     packaging
@@ -150,7 +149,9 @@ effectiveStdenv.mkDerivation rec {
     cuda_cudart
   ]);
 
-  nativeCheckInputs = lib.optionals pythonSupport (with python3Packages; [
+  nativeCheckInputs = [
+    gtest
+  ] ++ lib.optionals pythonSupport (with python3Packages; [
     pytest
     sympy
     onnx
@@ -179,7 +180,7 @@ effectiveStdenv.mkDerivation rec {
     "-DFETCHCONTENT_SOURCE_DIR_SAFEINT=${safeint}"
     "-DFETCHCONTENT_TRY_FIND_PACKAGE_MODE=ALWAYS"
     "-Donnxruntime_BUILD_SHARED_LIB=ON"
-    "-Donnxruntime_BUILD_UNIT_TESTS=ON"
+    (lib.cmakeBool "onnxruntime_BUILD_UNIT_TESTS" doCheck)
     "-Donnxruntime_ENABLE_LTO=ON"
     "-Donnxruntime_USE_FULL_PROTOBUF=OFF"
     (lib.cmakeBool "onnxruntime_USE_CUDA" cudaSupport)
--- a/pkgs/development/libraries/onnxruntime/default.nix
+++ b/pkgs/development/libraries/onnxruntime/default.nix
@@ -191,6 +191,7 @@ effectiveStdenv.mkDerivation rec {
     (lib.cmakeFeature "FETCHCONTENT_SOURCE_DIR_CUTLASS" "${cutlass}")
     (lib.cmakeFeature "onnxruntime_CUDNN_HOME" "${cudaPackages.cudnn}")
     (lib.cmakeFeature "CMAKE_CUDA_ARCHITECTURES" cudaArchitecturesString)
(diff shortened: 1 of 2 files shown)


pkgs/development/libraries/onnxruntime/default.nix
```diff
@@ -151,6 +150,7 @@ effectiveStdenv.mkDerivation rec {
   ]);
 
   nativeCheckInputs = lib.optionals pythonSupport (with python3Packages; [
+    gtest
```
jonringer, line 153: this should probably be moved outside of the pythonSupport flag as it doesn't exist under python package set.

autrimpo: done

pkgs/development/libraries/onnxruntime/default.nix
SomeoneSerge, line 1: >  I don't think trying to tie it to NIX_BUILD_CORES is a good idea, as that would eat up...


Can't say this is very convincing given that `NIX_BUILD_CORES` can be controlled by the user 🤔

pkgs/development/libraries/onnxruntime/default.nix
SomeoneSerge: > I've moved gtest back to its original place - it is only needed for the tests. 

Ugh there it goes again... so, why is it `nativeCheckInputs` and not `checkInputs`, and how does this work if we _hypothetically_ want to enable tests (using `hostPlatform.emulator`, presumably) in cross-compilation?

autrimpo: Was like that before I started touching it 🤷 You're right though, it should probably go in `checkInputs`.

https://github.com/NixOS/nixpkgs/blob/3392575c4008b9a98dc69e68d02a031b379d5a81/pkgs/development/libraries/onnxruntime/default.nix#L129-L130

pkgs/development/libraries/onnxruntime/default.nix
autrimpo: But what if the user wants to use multiple cores for the rest of the build? From my testing, nvcc_threads only had effect during parts of the build that were building CUDA objects and not the rest of the C++ compilation. Tying it together severely limits the number of cores you can use unless you have tons of RAM, resulting in net slower builds than using all cores for C++ and 1 for CUDA.

autrimpo: Ideally it would be a separate setting but I'm not aware of anything.

autrimpo: Actually, thinking about it now, in onnx `NIX_BUILD_CORES` is passed as `make -j` in the build. Not sure exactly how the make targets are set up, but wouldn't that hypothetically result in `NIX_BUILD_CORES` CUDA jobs each with `NIX_BUILD_CORES` threads?

pkgs/development/libraries/onnxruntime/default.nix
SomeoneSerge: I'm all sleepy, but this is probably relevant and I plan to acquaint tomorrow: https://github.com/NixOS/nixpkgs/pull/206742/files#diff-9538c800780031db3dfa7746f5a36fbbc895c60fdc896385bedd36940927427dR51

pkgs/development/libraries/onnxruntime/default.nix
SomeoneSerge: > But what if the user wants to use multiple cores for the rest of the build? From my testing, nvcc_threads only

Oh I see. Sure let's stay with your current solution

SomeoneSerge (approved):

SomeoneSerge:
```bash
❯ nom build -I nixpkgs=flake:github:autrimpo/nixpkgs/python-onnxruntime-cuda -f '<nixpkgs>' python3Packages.onnxruntime --print-out-paths
Finished at 08:01:38 after 1s
/nix/store/sm9fnczvjzlanydicfm8x106x26w6k12-python3.11-onnxruntime-1.16.3
```

Nixpkgs config:

<details>

```nix
{
  cudaSupport = true;
  cudaCapabilities = [
    "7.0"
    "8.0"
    "8.6"
  ];
  cudaEnableForwardCompat = false;
  allowUnfreePredicate =
    p:
    builtins.all (
      license:
      builtins.elem license.shortName [
        "CUDA EULA"
        "cuDNN EULA"
      ]
    ) (p.meta.licenses or [ p.meta.license ]);
}
```

</details>

Issue #300735, opened 2024-04-01
[Tracking] Remove all uses of `lib.mdDoc`


philiptaron:
Since https://github.com/NixOS/nixpkgs/pull/237557 was merged, `lib.mdDoc` has been a no-op. Let's remove the places that use it.

```
$ rg -w '(lib\.)?mdDoc' -o --no-filename | sort | uniq -c
  12356 lib.mdDoc
   1242 mdDoc
```

hojerst:
as this seems to break the current home-manager project - wouldn't it be reasonable to wait a bit longer / issue a warning when this is used instead of removing it alltogether?

eclairevoyant:
> wouldn't it be reasonable to wait a bit longer / issue a warning

There is a warning, unfortunately it's broken.

```
nix-repl> lib.options.mdDoc "test"
trace: warning: lib.mdDoc was removed from nixpkgs. Option descriptions are now in Markdown by default, you can remove any remaining uses of it.
"test"

nix-repl> lib.mdDoc "test"
error: attribute 'mdDoc' missing
       at «string»:1:1:
            1| lib.mdDoc "test"
             | ^
```

philiptaron:
> as this seems to break the current home-manager project - wouldn't it be reasonable to wait a bit longer / issue a warning when this is used instead of removing it alltogether?

I'm mostly interested in removing the _use_ of `lib.mdDoc` in `nixpkgs` itself. Obviously there's a much longer period where it remains accessible from `lib`, likely originating from `lib/deprecated.nix`.

philiptaron:
Closed by https://github.com/NixOS/nixpkgs/pull/303841 (thanks @stuebinm!)

stuebinm:
ah sorry for the breakage everyone — I added that warning assuming it would be fine, but missed that `mdDoc` was only exposed via an `inherit` into `lib.mdDoc` (which of course got automatically removed along with all the other inherits).

eclairevoyant:
No worries, it'll be fixed soon enough. Thanks for the original PR; maybe this will get HM to remove the usages sooner too ;)

eclairevoyant:
Giving a reasonable deprecation period is how we should handle anything removed from `lib`. In this case it was removed from hm and nix-darwin fairly quickly, but apparently we still have [a lot more consumers using this function](https://github.com/search?q=mddoc+lang%3Anix&type=code). At the very least they should know why they are getting this error for now.

[...]

PR #304277, opened 2024-04-15
lib: add mdDoc back in to provide proper warning; set explicit timeline for removal

eclairevoyant:
## Description of changes

Ref. https://github.com/NixOS/nixpkgs/issues/300735#issuecomment-2056850598

Most uses of `mdDoc` would be via `lib.mdDoc`, not `lib.options.mdDoc`, so we want to keep `lib.mdDoc` available for consumers to actually receive a useful warning and have enough time to remove usages.

Also, we should provide an explicit timeline for full removal, which I've done in the warning message.

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/lib/default.nix
+++ b/lib/default.nix
@@ -146,7 +146,7 @@ let
       scrubOptionValue literalExpression literalExample
       showOption showOptionWithDefLocs showFiles
       unknownModule mkOption mkPackageOption mkPackageOptionMD
-      literalMD;
+      mdDoc literalMD;
     inherit (self.types) isType setType defaultTypeMerge defaultFunctor
       isOptionType mkOptionType;
     inherit (self.asserts)
--- a/lib/options.nix
+++ b/lib/options.nix
@@ -404,7 +404,7 @@ rec {
      Kept here to alert downstream users who may not be aware of the migration's
      completion that it should be removed from modules.
   */
-  mdDoc = lib.warn "lib.mdDoc was removed from nixpkgs. Option descriptions are now in Markdown by default, you can remove any remaining uses of it.";
+  mdDoc = lib.warn "lib.mdDoc will be removed from nixpkgs in 24.11. Option descriptions are now in Markdown by default; you can remove any remaining uses of lib.mdDoc.";
 
   /* For use in the `defaultText` and `example` option attributes. Causes the
      given MD text to be inserted verbatim in the documentation, for when


philiptaron (approved):
I think this is directionally the right way to deprecate a symbol from `lib`. There's a whole file of these in `lib/deprecated.nix`. I wonder how to make this one not just be a one-off.

eclairevoyant:
Does that mean we should move the impl from `options.nix` to `deprecated.nix` as well?

philiptaron:
> Does that mean we should move the impl from `options.nix` to `deprecated.nix` as well?

I think so? After that, pulling that impl into options.nix through `inherit (lib.misc) mdDoc` or similar. For some reason, the filename is `deprecated.nix` but the exported name is `misc`. 🤔 

Personally I'm fine to merge as-is, just putting the thought out there about the general "how do things in `lib get deprecated?" question.

eclairevoyant:
That's fair.
I checked for out-of-tree usages, and maybe moving to `deprecated.nix` is not a huge issue, since it seems like almost no one is using `lib.options.mdDoc` anyway:

https://github.com/search?q=lang%3Anix+%22lib.options.mdDoc%22&type=code
https://github.com/search?q=lang%3Anix+%22%28lib.options%29%22+%22mdDoc%22&type=code

lib/default.nix
```diff
@@ -146,7 +146,7 @@ let
       scrubOptionValue literalExpression literalExample
       showOption showOptionWithDefLocs showFiles
       unknownModule mkOption mkPackageOption mkPackageOptionMD
-      literalMD;
```
philiptaron, line 149: You didn't pull it from `self.misc`, so this code will break.

philiptaron (changes requested):

lib/default.nix
eclairevoyant: Thanks for catching. Double-checked now that both forms of mdDoc still work.

philiptaron (approved):
OK, this PR:

1. Moves the deprecation warning from `lib/options.nix` to `lib/deprecated.nix`.
2. Re-introduces the symbol `mdDoc` exported from `lib`.
3. Removes the symbol from `lib.options`.

The final step is the only one that I think might want to be changed, by adding a `inherit (lib.misc) mdDoc` in `lib/options.nix`. But I'm fine if it doesn't.

eclairevoyant:
> adding a `inherit (lib.misc) mdDoc` in `lib/options.nix`

Done, added a comment there as well.

philiptaron (approved):

roberth:
I've done many `lib` reviews, but I can't remember ever interacting with `deprecated.nix`. We've been doing deprecations in their normal locations - right in the sublibrary where it always was.
I think we should get rid of `deprecated.nix`, in a separate PR, and avoid it here.

eclairevoyant:
Okay, thanks for the note; I'll just drop the last commit then.

roberth (approved):

philiptaron (approved):
I still wonder about a coherent deprecation policy for `lib`. As originally noted, I'm fine deprecating in place.

roberth:
> I think we should get rid of `deprecated.nix`, in a separate PR, and avoid it here.

Can't quite remove it anytime soon, but this should help:

- #305752

benaryorg:
Given that the explicit timeline (`lib.mdDoc will be removed from nixpkgs in 24.11`) was introduced in this issue I wanted to ask whether there are plans to actually remove it, given that the release preparations for 24.11 are currently ongoing.?

The only consumer of `lib.mdDoc` that I currently know of (colmena) has code checking for its existence and falls back to an identity function for its implementation, so at least in my (comparatively small? I guess?) codebase removal is absolutely feasible though I don't know much about code outside that.
Either way getting that message on 24.11 might cause confusion, so removal or adjusting the warning text would be cool.

Issue #265058, opened 2023-11-02
[Tracking] Documentation: Remove `nix-env` from the Nixpkgs manual


fricklerhandwerk:
## Problem

Beginners are still nudged to use `nix-env`, although that will not do what they expect or need from their Nix setup soon after.
Lots of parts in the Nixpkgs manual are quite old, from a time where declarative patterns were not as much evolved yet. Today we have more predictable alternatives, and should at the very least implicitly recommend ephemeral environments wherever possible.

This was discussed with the @nixos/documentation-team

Related: https://github.com/NixOS/nix/issues/7769

## Proposal

Remove all occurrences of `nix-env`. And you can help!

## How to help

- Check for [open PRs related to `nix-env`](https://github.com/NixOS/nixpkgs/pulls?q=is%3Apr+is%3Aopen+nix-env)
- Open the [latest Nixpkgs manual](https://nixos.org/manual/nixpkgs/unstable/) and search for instances of `nix-env` not being worked on yet
- Take any occurrence that seems easy enough to replace with `nix-shell`
- In a clone of the Nixpkgs repo, start the manual preview that will automatically update on changes:

  ```console
  $ cd doc
  $ nix-shell 
  [nix-shell]$ devmode
  ```
- Make the change and **work through the examples** to make sure they actually work

There may be sections that need a complete rewrite to start making sense. Ideally the change should still focus on simply avoiding `nix-env`. If you're interested in a particular topic, consider owning that piece of documentation and reworking it to be up-to-date with current best practices.

## Work done

- https://github.com/NixOS/nixpkgs/pull/258595
- https://github.com/NixOS/nixpkgs/pull/257276
- https://github.com/NixOS/nixpkgs/pull/255947
- https://github.com/NixOS/nixpkgs/pull/255131
- https://github.com/NixOS/nixpkgs/pull/257861
- https://github.com/NixOS/nixpkgs/pull/265057

PR #304307, opened 2024-04-15
doc: move note on configuring Nixpkgs in NixOS to the NixOS manual

fricklerhandwerk:
that NixOS manual section talks a lot about Nixpkgs package
configuration, which really should not be there but rather in the
Nixpkgs manual itself. but this is a rabbit hole for another time.

Related: https://github.com/NixOS/nixpkgs/issues/265058

---

### DIFF
--- a/doc/using/configuration.chapter.md
+++ b/doc/using/configuration.chapter.md
@@ -1,6 +1,7 @@
 # Global configuration {#chap-packageconfig}
 
-Nix comes with certain defaults about what packages can and cannot be installed, based on a package's metadata. By default, Nix will prevent installation if any of the following criteria are true:
+Nix comes with certain defaults about which packages can and cannot be installed, based on a package's metadata.
+By default, Nix will prevent installation if any of the following criteria are true:
 
 -   The package is thought to be broken, and has had its `meta.broken` set to `true`.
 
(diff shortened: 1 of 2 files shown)


doc/using/configuration.chapter.md
```diff
@@ -10,19 +11,11 @@ Nix comes with certain defaults about what packages can and cannot be installed,
...
-
-```nix
-{
-  nixpkgs.config = {
-    allowUnfree = true;
-  };
-}
-```
 
 However, this does not allow unfree software for individual users. Their configurations are managed separately.
```
DanielSidhion, line 20: This line should also be removed, it was part of the explanation for the configuration you removed.

nixos/doc/manual/configuration/customizing-packages.section.md
```diff
@@ -1,11 +1,23 @@
 # Customising Packages {#sec-customising-packages}
 
-Some packages in Nixpkgs have options to enable or disable optional
-functionality or change other aspects of the package.
+The Nixpkgs configuration for a NixOS system is set by the {option}`nixpkgs.config` option.
+
+:::{.example}
+# Globally allow unfree packages
```
DanielSidhion, line 6: Since this is now an example and is a self-contained "block", might also be a good idea to move that sentence ("However, this does not allow unfree software...") here after a slight adjustment to make it more understandable.

fricklerhandwerk:
@DanielSidhion thanks, good catch!

PR #304313, opened 2024-04-15
fish: remove `python` from dependencies

drupol:
Hello everyone,

Today, while optimizing the closure size of a Home-Manager profile, I discovered that including Python3 significantly increased the size. Specifically, Python3 set the `fish` derivation's size to 220MB.

```
❯ nix path-info --recursive --size --closure-size --human-readable $(nix build --impure --print-out-paths --expr '(import ./. { system = "x86_64-linux";}).fish')
...
/nix/store/rswmlkz46y7svb399kfzvp3gdihml8mg-fish-3.7.1            14.6M  220.8M
...
```

Upon examining the [Fish derivation](https://github.com/NixOS/nixpkgs/blob/90055d5e616bd943795d38808c94dbf0dd35abe8/pkgs/shells/fish/default.nix#L23), I found a parameter `usePython` set to `true` by default. 

When I build Fish with `usePython` set to `false`, its size dropped to 92MB.

```
programs.fish.package = pkgs.fish.override({
  usePython = false;
});
```

```
❯ nix path-info --recursive --size --closure-size --human-readable $(nix build --impure --print-out-paths --expr '(import ./. { system = "x86_64-linux";}).fish.override({ usePython=true; })')
...
/nix/store/mgmn7cyjriy65z27hijfma2habq4779y-fish-3.7.1            14.6M   92.3M
...
```

It appears that Python is primarily used to run `fish_config`, the web-based interface for configuring Fish.

Given this, I opened this thread to discuss whether the `usePython` flag should default to `false` to optimize the closure size for users who do not require the web configuration interface. In this Nix context, Fish is usually configured through Home-Manager and/or custom config files. I guess not so many people are aware of this config interface anyway, but I might be wrong.

This PR is here as a draft for the moment until we find a way to fix this gracefully.

See the related Discourse thread: https://discourse.nixos.org/t/fish-shell-includes-python-by-default/43421/5

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/shells/fish/default.nix
+++ b/pkgs/shells/fish/default.nix
@@ -1,9 +1,7 @@
 { stdenv
 , lib
 , fetchurl
-, fetchpatch
 , coreutils
-, which
 , gnused
 , gnugrep
 , groff
@@ -18,9 +16,6 @@
 , cmake
 , fishPlugins
 , procps
-
-# used to generate autocompletions from manpages and for configuration editing in the browser
-, usePython ? true
 
 , runCommand
 , writeText
@@ -270,14 +265,6 @@ let
           -i $out/share/fish/functions/{__fish_print_packages.fish,__fish_print_addresses.fish,__fish_describe_command.fish,__fish_complete_man.fish,__fish_complete_convert_options.fish} \
              $out/share/fish/completions/{cwebp,adb,ezjail-admin,grunt,helm,heroku,lsusb,make,p4,psql,rmmod,vim-addons}.fish
 
-    '' + optionalString usePython ''
-      cat > $out/share/fish/functions/__fish_anypython.fish <<EOF
-      function __fish_anypython
-          echo ${python3.interpreter}
-          return 0
-      end
-      EOF
-
     '' + optionalString stdenv.isLinux ''
       for cur in $out/share/fish/functions/*.fish; do
         sed -e "s|/usr/bin/getent|${getent}/bin/getent|" \
@@ -335,7 +322,7 @@ let
             ) | ${lib.getExe gnugrep} -q 'a href="http://localhost.*Start the Fish Web config'
           '';
           in
-          runCommand "test-web-config" { } ''
+          runCommand "test-web-config" { buildInputs = [ python3 ]; } ''
             HOME=$(mktemp -d)
             ${fish}/bin/fish ${fishScript} && touch $out
           '';


pkgs/shells/fish/default.nix
```diff
@@ -306,40 +291,6 @@ let
     };
 
     passthru = {
-      shellPath = "/bin/fish";
```
jtojnar, line 309: Shells must have `shellPath` or they will not be recognized by NixOS!

pkgs/shells/fish/default.nix
```diff
@@ -306,40 +291,6 @@ let
...
-      tests = {
-        nixos = lib.optionalAttrs stdenv.isLinux nixosTests.fish;
```
jtojnar, line 311: Do not drop this either.

pkgs/shells/fish/default.nix
```diff
@@ -306,40 +291,6 @@ let
...
-
-        # Test the fish_config tool by checking the generated splash page.
-        # Since the webserver requires a port to run, it is not started.
-        fishConfig =
-          let fishScript = writeText "test.fish" ''
```
jtojnar, line 316: We could fix the test by adding python to `PATH`.

pkgs/shells/fish/default.nix
```diff
@@ -270,14 +263,6 @@ let
           -i $out/share/fish/functions/{__fish_print_packages.fish,__fish_print_addresses.fish,__fish_describe_command.fish,__fish_complete_man.fish,__fish_complete_convert_options.fish} \
              $out/share/fish/completions/{cwebp,adb,ezjail-admin,grunt,helm,heroku,lsusb,make,p4,psql,rmmod,vim-addons}.fish
 
-    '' + optionalString usePython ''
```
jtojnar, line 273: Python is also used by many auto-completions (including [composer](https://github.com/fish-shell/fish-shell/pull/6225)). But all those and `fish_config` find Python using the `__fish_anypython` function, which looks like it should be able to [discover Python from `PATH`](https://github.com/fish-shell/fish-shell/blob/9af6a64fd2bd132ac1866af846fdc8368e086206/share/functions/__fish_anypython.fish#L5-L6).

`share/fish/tools/create_manpage_completions.py` currently uses `#!/usr/bin/env python3` in shebang so that is not affected either.

So looks like this PR will only make it take slightly longer to find Python and break those features in environments without Python, not sure if any exist in practice.

jtojnar (changes requested):

pkgs/shells/fish/default.nix
drupol: I opened this PR to received feedback such as yours. Now I will adjust the PR to make something usable and lighter.

pkgs/shells/fish/default.nix
drupol: Fixed.

pkgs/shells/fish/default.nix
drupol: Restored.

pkgs/shells/fish/default.nix
drupol: I fixed it by adding it to the `buildInputs` in the `runCommand`. Is it OK for you?

adamcstephens:
I understand the concern, but prefer we match the expected out of the box experience which includes the web interface. It seems pretty simple to reduce size if you're striving to do so.

Can we switch to python minimal? Does this save much at all?

jtojnar:
`python3Minimal` does not contain some things that are generally taken for granted. It would need to be thoroughly tested.

And if you have `python3` installed (which is the case on most systems), the system closure will be actually larger due to multiple instances of Python.

drupol:
Fair enough. I don't agree much but OK ... I'll use `usePython = false;` then.

r-vdp:
Do you have many reverse dependencies for fish? Because on my system it builds pretty fast and doesn't really have any reverse dependencies. 

So you could also just override it in your home manager profile without needing to change this for all nixpkgs users.

PR #304316, opened 2024-04-15
all-packages: Add firstAvailableOnHost function

mattpolzin:
## Description of changes

This is a preliminary PR to propose an idea. I don't know if it's a good idea, I don't know where the function should live, and I am seeking feedback.

The `firstAvailableOnHost` function aids in producing top-level derivations where not all host platforms are supported by the same derivation within `nixpkgs`. In practice, the derivations are the same, but they must be built differently depending on the host.

It is conceptually the same as a top-level if-statement choosing between two derivations, but it supports any number of possible variants of the derivation and it merges the platforms together. This is a key advantage -- no matter what host the derivation is evaluated on, the list of platforms supported will be complete.

My desire for this function in part comes from the frustration that currently any nixpkgs derivation that depends on `electron` and uses `electron.meta.platforms` as its own `meta.platforms` will look like it supports different platforms depending on where the derivation is evaluated. Evaluate on a Darwin machine and it works, but evaluate it on a Linux machine and it looks like Darwin is not even supported. This latter situation happens to cause `ofBorg` to think such packages should not be built for Darwin.

I've marked a couple of other top-level packages as "candidates" but I noted a handful of additional locations where this would not immediately work because the derivations being chosen between all stated that they supported all platforms, but I suspect that the more robust solution would be to refactor those to use a function like this as well and let the individual derivations being chosen between state if they support a given architecture instead of that information only living in a conditional in the top-level file.

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/top-level/all-packages.nix
+++ b/pkgs/top-level/all-packages.nix
@@ -15497,6 +15497,7 @@ with pkgs;
 
   fasmg = callPackage ../development/compilers/fasmg { };
 
+  # candidate
   fbc = if stdenv.hostPlatform.isDarwin then
     callPackage ../development/compilers/fbc/mac-bin.nix { }
   else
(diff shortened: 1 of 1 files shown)


mattpolzin:
@teutat3s I am just broadening my request for feedback on this hypothetical refactor. Any thoughts (other than I assume this function should not ultimately be defined where it is in the all-packages file)?

pkgs/top-level/all-packages.nix
```diff
@@ -18084,11 +18085,23 @@ with pkgs;
...
+  # and merge all meta.platforms in the resulting derivation.
+  firstAvailableOnHost = packages:
+    let platforms = lib.unique (lib.concatMap (p: p.meta.platforms) packages);
+    in let firstAvailable =
+         lib.findFirst (lib.meta.availableOn stdenv.hostPlatform)
+                       (let p = lib.head packages;
+                            name = p.pname or p.name;
+                        in (throw "${name} not supported on ${stdenv.hostPlatform.system}"))
+                       packages;
+       in lib.recursiveUpdate firstAvailable { meta.platforms = platforms; };
```
alyssais, line 18098: We would also need to do something about badPlatforms.  Currently if both packages have the same badPlatforms that information would be incorrectly discarded.

alyssais (commented):
I'm not sure about this, because the problem here is that then the platform list is not in sync with the rest of `meta`.  Often when we do this we have distinct implementations with completely different metadata (e.g. udev or iconv).

mattpolzin:
Good points @alyssais. I would be potentially interested in continuing down the path of seeing if a similar function could be workable, but I wonder if a totally different approach would be better: finding a way to have Hydra and/or ofborg check platforms supported having set the host platform to each of a set of options might also solve the problem I described originally of incorrectly limiting platforms for a “package” (by which I mean a particular attribute in all-packages, where `electron` is conceptually one package even though it’s one of two alternatives under the hood).

mattpolzin:
I'll close this. I think it's not the right way to go about solving my problem.

PR #284551, opened 2024-01-28
Add `types.attrTag`

roberth:
## Description of changes

The purpose of `attrTag` is to describe values that are already structured with this exact syntax, in RFC 42 settings.

It can also be used as an alternative to other sum type representations that aren't merged yet (edit: in some cases). This may or may not be nice, but that's irrelevant because `attrTag` is needed for a specific purpose anyway, as mentioned.
My goal here is only to introduce _mechanism_, not _policy_. For instance, if we want to pick a favorite sum type representation, or formulate some other kinds of recommendations, that can be done separately, later, when we have practical experience with multiple sum types.

I've updated the docs to lay out the new concept and how it relates to the existing "union-like" types.

Other sum type in development (not suitable for my use case as mentioned)
- #254790

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
- For non-Linux: Is sandboxing enabled in `nix.conf`? (See [Nix manual](https://nixos.org/manual/nix/stable/command-ref/conf-file.html))
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:
  - [NixOS test(s)](https://nixos.org/manual/nixos/unstable/index.html#sec-nixos-tests) (look inside [nixos/tests](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests))
  - and/or [package tests](https://nixos.org/manual/nixpkgs/unstable/#sec-package-tests)
  - or, for functions and "core" functionality, tests in [lib/tests](https://github.com/NixOS/nixpkgs/blob/master/lib/tests) or [pkgs/test](https://github.com/NixOS/nixpkgs/blob/master/pkgs/test)
  - made sure NixOS tests are [linked](https://nixos.org/manual/nixpkgs/unstable/#ssec-nixos-tests-linking) to the relevant packages
- [ ] Tested compilation of all packages that depend on this change using `nix-shell -p nixpkgs-review --run "nixpkgs-review rev HEAD"`. Note: all changes have to be committed, also see [nixpkgs-review usage](https://github.com/Mic92/nixpkgs-review#usage)
- [ ] Tested basic functionality of all binary files (usually in `./result/bin/`)
[...]

PR #304328, opened 2024-04-15
lib.attrsets.matchAttrTag: init

infinisil:
## Description of changes

As suggested in https://github.com/NixOS/nixpkgs/pull/284551#pullrequestreview-1961816244, this is a useful utility function for working with values from `lib.types.attrTag`:

```
lib.matchAttrTag "my.option" value.my.option {
  tag1 = tagValue: "Tag1: ${tagValue}";
  tag2 = tagValue: "Tag2: ${tagValue}";
}
```

This was written in [Nix Hour 68](https://www.youtube.com/watch?v=Y6epvlDxrCQ&list=PLyzwHTVJlRc8yjlx4VR4LU5A5O44og9in&index=1)

Ping @roberth

## Things done

- [x] Add documentation
- [x] Add tests
- [ ] Update `lib.types.attrTag` docs to mention this function

---

### DIFF
--- a/lib/attrsets.nix
+++ b/lib/attrsets.nix
@@ -5,6 +5,8 @@
 
 let
   inherit (builtins) head length;
+  inherit (lib.asserts) assertMsg;
+  inherit (lib.generators) toPretty;
   inherit (lib.trivial) mergeAttrs warn;
   inherit (lib.strings) concatStringsSep concatMapStringsSep escapeNixIdentifier sanitizeDerivationName;
   inherit (lib.lists) foldr foldl' concatMap elemAt all partition groupBy take foldl;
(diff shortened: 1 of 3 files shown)


lib/attrsets.nix
```diff
@@ -1726,6 +1728,75 @@ rec {
...
+      n = number: "It's the number ${toString number}";
+      s = str: "It's the string ${str}";
+    }
+    => "It's the string Paul"
+
+    matchAttrTag "test" { unknown = null; } {
+      n = number: "It's the number ${toString number}";
+      s = str: "It's the string ${str}";
+    }
+    => error: test: No handler for tag "unknown"
```
roberth, line 1778: `"test"` doesn't seem like a great example. It'd help to have a more realistic use case. When it's a realistic example, it's easier to see what's expected from the context.

When this error occurs in the context of a module that defines an option of type `attrTag ...`, this case will only occur due to programming error, and I don't think the module author should have the foresight to clarify that in the context.

In other cases, you probably do want the context to be quite general.
Maybe we should have another function that takes an option instead of its value? Doesn't work for attrTags when used as part of type constructors like `attrsOf` though.

lib/attrsets.nix
```diff
@@ -1726,6 +1728,75 @@ rec {
...
+      n = number: "It's the number ${toString number}";
+      s = str: "It's the string ${str}";
+    }
+    => error: test: Value has not exactly one tag: [ "n" "s" ]
+    ```
+
+    :::
+  */
+  matchAttrTag =
+    context: value: handlers:
```
roberth, line 1790:
```suggestion
    context: handlers: value:
```

Often, the `handlers` are more part of the program than part of the data, so arranging them like this lets us take advantage of currying, in things like `map`, pipes, or eta-reduced function definitions.

infinisil: I think the most common case is to _not_ curry tbh, and you can still use `flip` if you need it. With the `value: handlers:` order we're also mirroring how matching looks like in many other languages:

```rust
match value {
  Foo { .. } => 0
  Bar { .. } => 1
}
```

```haskell
case value of
  Foo { .. } -> 0
  Bar { .. } -> 1
```

So I'd really prefer leaving it like that

lib/attrsets.nix
infinisil: What if the merged result of `attrTag` is the matching function itself :thinking: 

```nix
config.some.option {
  n = number: "It's the number ${toString number}";
  s = str: "It's the string ${str}";
}
```

This would be doable effectively in a backwards-compatible way using `__functor`! And this way we should be able to get a good error message for free. Not a big fan of `__functor` in general, but in this case I'm really wondering..

pluiedev:
Oh this looks *really* nice. I make extensive use of `attrTag` and this would make the processing logic much easier (at least, it would be much simpler than [whatever this is](https://github.com/pluiedev/plasma-manager/blob/7446cfaaf17d48487822085f00dcd1fae49d4746/modules/widgets/default.nix) :sweat_smile:)

lib/attrsets.nix
pluiedev: The annoying thing is that if you do want to lib.flip, you have to write it like `flip (matchAttrTag "context") [...]` which looks a bit funny to me

infinisil: I guess if we go for https://github.com/NixOS/nixpkgs/pull/304328#discussion_r1598758011 this won't be a problem anymore :D

lib/attrsets.nix
roberth: For my use case, settings, I need to not have it:
```
       … while evaluating attribute '__functor'

error: cannot convert a function to JSON
```

infinisil: I guess it also wouldn't be very backwards compatible, because any code that asserts for the attribute set to only have one attribute would fail.

lib/attrsets.nix
roberth: **Pipe friendly**:

```nix
map (matchAttrTag "context") myList
```
and
```nix
value |> matchAttrTag "context" {
  ...
}
```

**Not pipe friendly**:

```nix
map (flip (matchAttrTag "context")) myList
```
and
```nix
matchAttrTag "context" value {
  ...
}
```

roberth: Also, fwiw, avoiding `flip` avoids a runtime cost.

Issue #164630, opened 2022-03-17
Cabal/Haskell: Importing exports Use of `data-dir` leads to large closure


m4dc4p:
### Describe the bug
When an executable uses the "paths" module exported by a library, the closure of that executable is very large. It includes packages such as GHC and its documentation.

### Steps To Reproduce
1. Clone the repo https://github.com/m4dc4p/nix-cabal-data-dir-surprise
2. Run `nix-store --query -R --include-outputs $(nix-build -A app)`. 
3. Notice closure is small & GHC does not appear.
4. Run `nix-store --query -R --include-outputs $(nix-build -A big-app)`. 
5. Closure has many more entries, including GHC (/nix/store/...-ghc-8.10.7) and its documentation (`/nix/store/...-ghc-8.10.7-doc`)

### Expected behavior

I would not expect that using the exported "paths" module from a library would make the closure so much larger. This is pretty impactful when building docker images. In the above repo, the docker image for `big-app` is 425MB, while the docker image for `app` is only 24M:

```
$ ls -lah $(nix-build -A docker-big-app)
-r--r--r--  1 justin.bailey  admin   475M Dec 31  1969 /nix/store/0bvvq9sdw9bwiqbwyaidxj0wwkjmdif5-docker-image-docker-big-app.tar.gz
$ ls -lah $(nix-build -A docker-app)
-r--r--r--  1 justin.bailey  admin    26M Dec 31  1969 /nix/store/kbs7i25jnnlfcd0i9wxq68q3i829323m-docker-image-docker-app.tar.gz
```

### Additional context

`app/Main.hs` conditionally imports the `Paths_library` module from the `library` package. If the module is imported and used, the closure blows up. This is due to a dependency found in the `package.conf.d` directory installed with library. 

```
$ nix-store --query -R --include-outputs $(nix-build -A big-app) | grep ghc-8.10.7
/nix/store/4ycwcc6rmyx2bchr7zbigfiyklqdkfik-ghc-8.10.7-doc
/nix/store/qmcdw0bq6590082f3g6sqflqvlyvap5g-ghc-8.10.7
$ nix why-depends --precise $(nix-build -A big-app) /nix/store/4ycwcc6rmyx2bchr7zbigfiyklqdkfik-ghc-8.10.7-doc
/nix/store/j17x9cwaj4japr0mwk8nqvjgx7i6aysf-app-0.1.0.0
[...]

PR #304352, opened 2024-04-15
justStaticExecutables: Forbid references to GHC

9999years:
## Description of changes

This makes `justStaticExecutables` error if the produced store path contains references to GHC. This is almost always erroneous and due to the generated `Paths_*` module being imported. This helps prevent `justStaticExecutables` from producing binaries with closure sizes in the gigabytes.

See: https://github.com/NixOS/nixpkgs/issues/164630

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [x] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/development/haskell-modules/generic-builder.nix
+++ b/pkgs/development/haskell-modules/generic-builder.nix
@@ -702,14 +702,6 @@ stdenv.mkDerivation ({
     runHook postInstallIntermediates
   '';
 
-  disallowedRequisites =
-    disallowedRequisites
-    ++ (
-      if disallowGhcReference
-      then [ghc]
-      else []
-    );
-
   passthru = passthru // rec {
 
     inherit pname version disallowGhcReference;
@@ -873,10 +865,19 @@ stdenv.mkDerivation ({
 // optionalAttrs (args ? dontStrip)              { inherit dontStrip; }
 // optionalAttrs (postPhases != [])              { inherit postPhases; }
 // optionalAttrs (stdenv.buildPlatform.libc == "glibc"){ LOCALE_ARCHIVE = "${glibcLocales}/lib/locale/locale-archive"; }
+// optionalAttrs (disallowedRequisites != [] || disallowGhcReference) {
+  disallowedRequisites =
+    disallowedRequisites
+    ++ (
+      if disallowGhcReference
+      then [ghc]
+      else []
+    );
+}
 
 # Implicit pointer to integer conversions are errors by default since clang 15.
 # Works around https://gitlab.haskell.org/ghc/ghc/-/issues/23456.
-// lib.optionalAttrs (stdenv.hasCC && stdenv.cc.isClang) {
+// optionalAttrs (stdenv.hasCC && stdenv.cc.isClang) {
   NIX_CFLAGS_COMPILE = "-Wno-error=int-conversion";
 }
 )
--- a/pkgs/development/haskell-modules/generic-builder.nix
+++ b/pkgs/development/haskell-modules/generic-builder.nix
@@ -110,6 +110,28 @@ in
   # build products from that prior build as a starting point for accelerating
   # this build
 , previousIntermediates ? null
+  # References to these store paths are forbidden in the produced output.
+, disallowedRequisites ? []
+  # Whether to allow the produced output to refer to `ghc`.
+  #
+  # This is used by `haskell.lib.justStaticExecutables` to help prevent static
+  # Haskell binaries from having erroneous dependencies on GHC.
+  #
+  # Generated `Paths_*` modules include paths for the runtime library
+  # directory (and similar) of the package being built. If the `Paths_*`
+  # module is imported, this creates a dependency from the static binary
+  # being built to the _library_ being built (which is dynamically linked
+  # and depends on the GHC used to build it).
+  #
+  # To avoid this:
+  # 1. Build the impacted derivation.
+  # 2. Run `strings` on the built binary of the impacted derivation to
+  #    locate the store paths it depends on.
+  # 3. Add `remove-references-to -t ${bad-store-path-in-binary}` to the
+  #    impacted derivation's `postInstall`.
+  #
+  # See: https://github.com/NixOS/nixpkgs/issues/164630
+, disallowGhcReference ? false
 , # Cabal 3.8 which is shipped by default for GHC >= 9.3 always calls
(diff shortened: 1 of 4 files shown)


maralorn:
I like this.

A bit sad that this makes justStaticExecutables better than enableSeparateBinOutput. I forgot whether there is a per output option regarding requisites.

We will need to make sure that this doesn’t break any of the Haskell binaries that we ship in nixpkgs so this definitely needs to merge into haskell-updates.

9999years:
@maralorn Agreed, how do I do that? Just change the merge target?

sternenseemann (changes requested):
`disallowedRequisites` needs to become an argument to `haskellPackages.mkDerivation` and transparently passed on to `stdenv.mkDerivation` in there. Otherwise `disallowedRequisites` will only be set depending on override order, i.e. any `overrideCabal`-based function applied after `justStaticExecutables` would revert the change again, as it uses a different fix point from `overrideAttrs`.

9999years:
@sternenseemann I'm not sure how to make that work; I can add `disallowedRequisites` to the generic builder, but then I don't have access to the `passthru.compiler` attribute. I can add a `ghc` argument with a default value, but then I can't access that in the override unless it's changed from the default due to https://github.com/NixOS/nix/issues/334

sternenseemann:
@9999years `passthru.compiler` is set in that function, so it is available as the `ghc` argument passed in.

sternenseemann:
Ah, I get what you mean. In principle you can literally add an `disallowGhcReference` argument tbh.

sternenseemann:
Or `drv: let old = drv; in overrideCabal { disallowedRequisites = [ old.passthru.compiler ]; } old` should also work?

9999years:
OK, I took a shot at it, let me know what you think.

I also checked that the docs render correctly:

image

9999years:
@sternenseemann Still looking for a review on this!

pkgs/development/haskell-modules/generic-builder.nix
```diff
@@ -20,6 +21,11 @@ in
 , version, revision ? null
 , sha256 ? null
 , src ? fetchurl { url = "mirror://hackage/${pname}-${version}.tar.gz"; inherit sha256; }
+  # The GHC to build this package with.
+  #
+  # `haskell.lib.justStaticExecutables` will add this to `disallowedRequisites`
+  # unless `needsGHCAtRuntime` is set.
+, ghc ? originalGHC
```
sternenseemann, line 28: Why are you threading it through in this way? It doesn't really make sense that you can freely pass `ghc = …` to `haskellPackages.mkDerivation` as that already has a predetermined GHC and no other would work. You can simply rely on `ghc` as it is already in scope, can't you?

pkgs/development/haskell-modules/generic-builder.nix
```diff
@@ -20,6 +21,11 @@ in
 , version, revision ? null
 , sha256 ? null
 , src ? fetchurl { url = "mirror://hackage/${pname}-${version}.tar.gz"; inherit sha256; }
+  # The GHC to build this package with.
+  #
+  # `haskell.lib.justStaticExecutables` will add this to `disallowedRequisites`
+  # unless `needsGHCAtRuntime` is set.
```
sternenseemann, line 27: (I think the argument name is outdated here.)

sternenseemann (commented):
Sorry for the wait!

pkgs/development/haskell-modules/generic-builder.nix
9999years: Hmm. I guess you're right that this is usually redundant, but I see no reason why it wouldn't work. I'll remove it for now.

9999years:
PTAL

maralorn:
This looks reasonable and simple. Really appreciate the documentation work! Thank you.

sternenseemann (commented):
This looks good to me! Last bit of busywork would be to add an entry to the changelog noting the change so people upgrading aren't confused if their derivations should start to fail.

pkgs/development/haskell-modules/generic-builder.nix
```diff
@@ -645,9 +667,17 @@ stdenv.mkDerivation ({
...
+    ++ (
+      if disallowGhcReference
+      then [ghc]
+      else []
+    );
+
   passthru = passthru // rec {
 
-    inherit pname version;
+    inherit pname version disallowGhcReference;
```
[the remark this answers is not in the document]
9999years: So that other derivations or Nix code can introspect it. It might be useful, especially for debugging.

9999years:
Sure, where does the changelog entry go?

sternenseemann:
> Sure, where does the changelog entry go?

`nixos/doc/manual/release-notes/rl-2405.section.md`, in the breaking changes section somewhere.

9999years:
Thanks, added a note to the breaking changes section. Let me know how it looks!

sternenseemann:
I've realized that we can't backport this anymore since this is technically a breaking change. I've moved the changelog entry to the 24.11 release notes as a result. We *could* backport a version of this which adds the new arguments to `haskellPackages.mkDerivation`, but doesn't change the behavior of `justStaticExecutables`.

I've also noticed that we haven't documented `disallowedRequisites` and `disallowedGhcReference` to the `mkDerivation` documentation in the manual, but maybe that's okay since `disallowedRequisites` is undocumented for `stdenv.mkDerivation` as well…

sternenseemann (approved):

sternenseemann:
@9999years Evals newer than 1806586 of [haskell-updates](https://hydra.nixos.org/jobset/nixpkgs/haskell-updates) contain your change. There are already a few regressions I've seen on aarch64-darwin. This is not really surprising as GHC can't properly eliminate dead code on that platform. It may be prudent to ignore `disallowGhcReference` on that platform.

9999years:
> There are already a few regressions I've seen on aarch64-darwin

Which attributes are impacted?

sternenseemann:
@9999years They should accumulate [here](https://hydra.nixos.org/eval/1806586#tabs-now-fail). Note that a lot of builds are still pending so the list will get longer as time goes on.

PR #304497, opened 2024-04-16
buildMavenPackage: use nativeBuildInputs in .fetchedMavenDeps as well

doronbehar:
## Description of changes

I encountered it in my attempt to build a certain java package from source - where I needed `gettext` to be available during that `mvn` command.
## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/development/tools/build-managers/apache-maven/build-package.nix
+++ b/pkgs/development/tools/build-managers/apache-maven/build-package.nix
@@ -28,7 +28,7 @@ let
 
     nativeBuildInputs = [
       maven
-    ];
+    ] ++ args.nativeBuildInputs or [ ];
 
     buildPhase = ''
       runHook preBuild


raboof (approved):
Since this is a FOD, any added complexity/variance in the FOD increases the risk of accidentally changing the behavior (and thus the resulting hash) without noticing it (as you still had the already-built 'old' hash in your cache). For that reason I would prefer introducing a new parameter to pass additional dependencies to the deps' `nativeBuildInputs`.

This change still LGTM, feel free to merge as-is or apply this suggestion.

doronbehar:
Thanks for the review! I ran this command:

```
nix build --rebuild -Lf. \
  $(dirname $(git grep -l '\.buildMavenPackage rec {' pkgs/ ) | \grep -v keycloak | awk -F/ '{print $NF".fetchedMavenDeps"}')

**Comment by doronbehar** [comment]
Sorry for an incomplete comment giving an email notification, will update when I get the results of the above command :).

**Comment by raboof** [comment]
> ```
> nix build --rebuild -Lf. \
>   $(dirname $(git grep -l '\.buildMavenPackage rec {' pkgs/ ) | \grep -v keycloak | awk -F/ '{print $NF".fetchedMavenDeps"}')
> ```

That is pretty cool ;).

Even if this reports no changed hashes, I'm not sure that changes my opinion, though: to be safe, you'd have to run something similar whenever any of the `nativeBuildInputs` (or their dependencies) change. Since AFAICT we don't even have an obvious way to tell when that happens (not within nixpkgs, let alone when using `buildMavenPackage` outside of nixpkgs), that seems pretty unlikely to happen. I'd prefer keeping the scope/complexity of the FOD as small as possible as a general good practice - though I guess this whole FOD already goes against that ideal ;).

That said, as mentioned before, I don't feel too strongly about this, so if you decide you prefer to apply the change as it is I'm happy to merge it, just LMK.

**Comment by doronbehar** [comment]
OK so here is my report on the potentially changed hashes:

I ran this command:

```
nix build --rebuild .#keycloak.plugins.keycloak-metrics-spi.fetchedMavenDeps
dirname $(git grep -l '\.buildMavenPackage rec {' pkgs/ ) | \
  grep -v keycloak | \
  awk -F/ '{print $NF".fetchedMavenDeps"}' | while read a; do \
  nix build --rebuild .#"$a"
done
```

**With, and without my patch**, I get the same output:

image

Sorry for the screenshot, I ran this on a server with good internet connection to speed things up.

I found a few more attributes by `grep`ing `buildMavenPackage {`:

```
nix build --keep-going --rebuild .#{s3proxy,openrefine,forge-mtg,certificate-ripper.src}.fetchedMavenDeps
```

And the above reported no hash mismatch.

-----

In general, I think that when an FOD builder is implemented in Nixpkgs, we should be able to be confident enough that it is not sensitive to such changes in `nativeBuildInputs`, i.e that it is deterministic. After the tests I ran above, I think that `buildMavenPackage` is somewhat sensitive to certain changes I can't yet put my finger on, but not to the change I propose here. Hence I feel confident in merging this.

**Comment by doronbehar** [comment]
For reference, the attribute that had a hash mismatch is `lemminx.fetchMavenDeps`, and `dbeaver.fetchedMavenDeps` and `jd-cli.fetchedMavenDeps` both completely failed to build.

The complete list of attributes I checked is:

```
dbeaver
ns-usbloader
tabula-java
digital
gephi
bytecode-viewer
ktfmt
lemminx
mariadb-connector-java
quark-goldleaf
schemaspy
global-platform-pro
java-language-server
scenebuilder
runelite
h2
keycloak-metrics-spi
keycloak-restrict-client-auth
scim-for-keycloak
slipstream
jugglinglab
cryptomator
jd-cli
openrefine
certificate-ripper
s3proxy
forge-mtg
scim-keycloak-user-storage-spi
```

**Comment by raboof** [comment]
> In general, I think that when an FOD builder is implemented in Nixpkgs, we should be able to be confident enough that it is not sensitive to such changes in nativeBuildInputs, i.e that it is deterministic.

That will often be true, but is not guaranteed: the `nativeBuildInputs` could have had a change in behavior that leads to a different output, and that would go unnoticed for a FOD. I suspect you're already aware, I don't intend to be pedantic, but for completeness an example:

```
{ pkgs ? import <nixpkgs> {} }:

let
  myBuildInput = pkgs.writeScriptBin "myNativeBuildInput" ''
    echo 'something something';
  '';
in
pkgs.stdenv.mkDerivation {
  pname = "fod-demo";
  version = "0.0.1";

  nativeBuildInputs = [ myBuildInput ];

  dontUnpack = true;

  installPhase = ''
    runHook preInstall

    mkdir -p $out
    myNativeBuildInput > $out/foo

    runHook postInstall
  '';

  outputHashMode = "recursive";
  outputHash = "sha256-eH+msMfBpAP20n0wytxsitbkihmJ+hRIgknWlhfGQqo=";
}
```

If you build this, and then change the implementation of `myBuildInput`, it's easy to miss that you also need to update the output hash.

> I think that buildMavenPackage is somewhat sensitive to certain changes I can't yet put my finger on

I agree it can likely be sensitive to things like changes on Maven Central.

> Hence I feel confident in merging this.

Great, thanks for the improvement!

doronbehar:
Yes I was aware of the behavior you explained. It is basics of FOD. Thanks for writing it down here, for future readers of this thread. :clinking_glasses: .

PR #304721, opened 2024-04-17
lldap-cli: init at 0-unstable-2024-05-15

ibizaman:
## Description of changes

New package https://github.com/Zepmann/lldap-cli

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/by-name/ll/lldap-cli/package.nix
+++ b/pkgs/by-name/ll/lldap-cli/package.nix
@@ -1,0 +1,63 @@
+{
+  lib,
+  stdenvNoCC,
+  fetchFromGitHub,
+  resholve,
+
+  bash,
+  coreutils,
+  curl,
+  gnugrep,
+  gnused,
+  jq,
+  lldap,
+  util-linux,
+}:
+
+resholve.mkDerivation {
+  pname = "lldap-cli";
+  version = "0-unstable-2024-05-15";
+
+  src = fetchFromGitHub {
+    owner = "ibizaman";
+    repo = "lldap-cli";
+    rev = "65c4102ab0b314593d69c02a803a11ba1853978f";
+    hash = "sha256-Ip1FkDrIsphi0GjP805Y8Gii16x7v/fsHtDUOzxie0w=";
+  };
+
+  solutions = {
+    default = {
+      scripts = [ "bin/lldap-cli" ];
+      interpreter = lib.getExe bash;
+      inputs = [
+        coreutils
+        curl
+        gnugrep
+        gnused
+        jq
+        lldap
+        util-linux
+      ];
+      keep = {
+        "$value" = true;
+      };
+    };
+  };
+
+  installPhase = ''
+    runHook preInstall
+
+    install -Dm555 $src/lldap-cli $out/bin/lldap-cli
+
+    runHook postInstall
+  '';
+
+  meta = {
+    description = "Command line interface for LLDAP";
+    homepage = "https://github.com/ibizaman/lldap-cli";
+    license = lib.licenses.gpl3;
+    mainProgram = "lldap-cli";
+    maintainers = with lib.maintainers; [ ibizaman ];
+    platforms = lib.platforms.all;
+  };
+}


pkgs/by-name/ll/lldap-cli/package.nix
```diff
@@ -0,0 +1,56 @@
...
+
+, curl
+, jq
+, makeWrapper
+, lldap
+, util-linux
+}:
+
+stdenvNoCC.mkDerivation (finalAttrs: rec {
+  name = "lldap-cli";
```
pluiedev, line 13: Prefer a combination of `pname` and `version`, instead of `name`. If a package is unversioned, the convention is to set the version to `0-unstable-YYYY-MM-DD`, where `YYYY-MM-DD` is where the date of the commit used. [See the section on package naming in `pkgs/README.md`.](https://github.com/NixOS/nixpkgs/blob/c10fb0df312867b0ae47a580818b317a9ae61f21/pkgs/README.md#package-naming)

In this case, `version` should be `0-unstable-2024-05-15`.

pkgs/by-name/ll/lldap-cli/package.nix
```diff
@@ -0,0 +1,56 @@
...
+    mkdir -p $out/share/lldap-cli
+    mkdir -p $out/bin
+
+    cp lldap-cli $out/bin
+    chmod +x $out/bin/lldap-cli
+
+    substituteInPlace $out/bin/lldap-cli --replace "curl" "${curl}/bin/curl"
+    substituteInPlace $out/bin/lldap-cli --replace "jq" "${jq}/bin/jq"
+    substituteInPlace $out/bin/lldap-cli --replace "column" "${util-linux}/bin/column"
+    substituteInPlace $out/bin/lldap-cli --replace "lldap_set_password" "${lldap}/bin/lldap_set_password"
```
pluiedev, line 43: There's probably a more elegant way of doing this. Maybe try replacing the shebang with a nix-shell with these dependencies?

pkgs/by-name/ll/lldap-cli/package.nix
```diff
@@ -0,0 +1,56 @@
...
+
+    runHook postInstall
+  '';
+
+  meta = {
+    description = "LLDAP-CLI is a command line interface for LLDAP.";
```
pluiedev, line 49:
```suggestion
    description = "Command line interface for LLDAP";
```

pkgs/by-name/ll/lldap-cli/package.nix
```diff
@@ -0,0 +1,56 @@
...
+, fetchFromGitHub
+
+, curl
+, jq
+, makeWrapper
+, lldap
+, util-linux
+}:
+
+stdenvNoCC.mkDerivation (finalAttrs: rec {
```
pluiedev, line 12: Why is there a `rec` alongside `finalAttrs`?

pkgs/by-name/ll/lldap-cli/package.nix
```diff
@@ -0,0 +1,56 @@
...
+    substituteInPlace $out/bin/lldap-cli --replace "jq" "${jq}/bin/jq"
+    substituteInPlace $out/bin/lldap-cli --replace "column" "${util-linux}/bin/column"
+    substituteInPlace $out/bin/lldap-cli --replace "lldap_set_password" "${lldap}/bin/lldap_set_password"
+
+    runHook postInstall
+  '';
+
+  meta = {
+    description = "LLDAP-CLI is a command line interface for LLDAP.";
+    homepage = "https://github.com/Zepmann/lldap-cli";
```
pluiedev, line 50: This does not match the fork you're basing the package on, which is https://github.com/ibizaman/lldap-cli!

pluiedev (changes requested):

pkgs/by-name/ll/lldap-cli/package.nix
ibizaman: FWIW I based this upon other packages that has this. But you're right, this makes no sense.

pkgs/by-name/ll/lldap-cli/package.nix
ibizaman: :man_facepalming:

pkgs/by-name/ll/lldap-cli/package.nix
ibizaman: I took "inspiration" from the [manual](https://nixos.org/manual/nixpkgs/stable/#fun-makeWrapper). Specifically:

> # Be advised that paths often should be patched in directly
> # (via string replacements or in `configurePhase`).

That being said, I [just realized](https://nixos.org/manual/nixpkgs/stable/#fun-substitute) I can pass multiple `--replace` arguments to `substituteInPlace`. What do you think of:

```nix
substituteInPlace $out/bin/lldap-cli \
    --replace "curl" "${curl}/bin/curl" \
    --replace "jq" "${jq}/bin/jq" \
    --replace "column" "${util-linux}/bin/column" \
    --replace "lldap_set_password" "${lldap}/bin/lldap_set_password"
```

pluiedev: The thing is that it's still pretty much doing the same thing and it's not exactly extensible. That said, I just realized that you could just add these to the nativeBuildInputs and the script should be able to use them just fine

ibizaman: Happy to try something else but I don't have any ideas.

pluiedev: Actually, try using something like https://github.com/nix-community/patsh or https://github.com/abathur/resholve. I wouldn't recommend using the latter anymore, though, since it depends on Python 2.7 which has been EOL for like 3-4 years now

ibizaman: Used resholve. It actually found a bug :tada: https://github.com/ibizaman/lldap-cli/commit/65c4102ab0b314593d69c02a803a11ba1853978f

ibizaman:
Friendly ping just in case you didn't see the update @pluiedev

pluiedev (approved):

pkgs/by-name/ll/lldap-cli/package.nix
```diff
@@ -0,0 +1,62 @@
...
+, coreutils
+, curl
+, gnugrep
+, gnused
+, jq
+, lldap
+, util-linux
+}:
+
+resholve.mkDerivation rec {
```
drupol, line 16: I think we should use the `finalAttrs` pattern here.

```nix
mkDerivation rec {
   foo = 1;
   bar = foo + 1;
}
```

pkgs/by-name/ll/lldap-cli/package.nix
```diff
@@ -0,0 +1,62 @@
...
+  pname = "lldap-cli";
+  version = "0-unstable-2024-05-15";
+
+  src = fetchFromGitHub {
+    owner = "ibizaman";
+    repo = pname;
```
drupol, line 22: Please don't use this.

`repo = pname;` is nice for DRY but creates a binding that goes too far.

See further information about this here: https://github.com/nix-community/nixpkgs-lint/issues/21

drupol (changes requested):
I added 2 minor comments, and I will need you to pass your new file through `nixfmt-rfc-style`.

pkgs/by-name/ll/lldap-cli/package.nix
abathur: Like other builders besides mkDerivation itself, resholve doesn't support finalAttrs yet:
- abathur/resholve#107

drupol: Ho yes, my mistake. Ignore my comment.

ibizaman: Thank you for the thorough explanation!

ibizaman:
@drupol I answered all review comments.
Also, I passed the file through the formatter, it changed the import comma style.

drupol (changes requested):
The commit log message should be the same as the PR title.

Also, there's a merge conflict...

alyssais:
https://github.com/NixOS/nixpkgs/pull/326131

PR #304741, opened 2024-04-17
[staging] python/hooks/python-imports-check-hook: use $python output if existing

msanft:
## Description of changes

Right now, `pythonImportsCheck` only recognizes `$out`. While this is the relevant output for most Python packages, there also are packages that define both `$out` and `$py` (packages providing both a Python library as well as executables), such as [dnf4](https://github.com/NixOS/nixpkgs/blob/master/pkgs/development/python-modules/dnf4/default.nix). This makes `pythonImportsCheck` work with such packages by checking in the `$py` with precedence over the `$out` output, if it exists.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/development/interpreters/python/hooks/python-imports-check-hook.sh
+++ b/pkgs/development/interpreters/python/hooks/python-imports-check-hook.sh
@@ -6,8 +6,13 @@ pythonImportsCheckPhase () {
 
     if [ -n "$pythonImportsCheck" ]; then
         echo "Check whether the following modules can be imported: $pythonImportsCheck"
-        export PYTHONPATH="$out/@pythonSitePackages@:$PYTHONPATH"
-        ( cd $out && eval "@pythonCheckInterpreter@ -c 'import os; import importlib; list(map(lambda mod: importlib.import_module(mod), os.environ[\"pythonImportsCheck\"].split()))'" )
+        pythonImportsCheckOutput=$out
+        if [ -n "$python" ]; then
+            echo "Using python specific output \$python for imports check"
+            pythonImportsCheckOutput=$python
+        fi
+        export PYTHONPATH="$pythonImportsCheckOutput/@pythonSitePackages@:$PYTHONPATH"
+        ( cd $pythonImportsCheckOutput && eval "@pythonCheckInterpreter@ -c 'import os; import importlib; list(map(lambda mod: importlib.import_module(mod), os.environ[\"pythonImportsCheck\"].split()))'" )
     fi
 }
 


pkgs/development/interpreters/python/hooks/python-imports-check-hook.sh
```diff
@@ -6,8 +6,13 @@ pythonImportsCheckPhase () {
 
     if [ -n "$pythonImportsCheck" ]; then
         echo "Check whether the following modules can be imported: $pythonImportsCheck"
-        export PYTHONPATH="$out/@pythonSitePackages@:$PYTHONPATH"
-        ( cd $out && eval "@pythonCheckInterpreter@ -c 'import os; import importlib; list(map(lambda mod: importlib.import_module(mod), os.environ[\"pythonImportsCheck\"].split()))'" )
+        checkPath=$out
```
SuperSandro2000, line 9: We should use something none generic to not overwrite potential other things from checkPhase. How about pythonImportsCheckOutput?

pkgs/development/interpreters/python/hooks/python-imports-check-hook.sh
```diff
@@ -6,8 +6,13 @@ pythonImportsCheckPhase () {
...
+        if [ -n "$py" ]; then
```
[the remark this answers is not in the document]
msanft: Okay. I've added both now. This could theoretically lead to weird behaviour when both are defined (since $python now takes precedence over $py in my implementation), but I think it can be assumed that packages should either not define both, or make libraries importable in both. What do you think?

pkgs/development/interpreters/python/hooks/python-imports-check-hook.sh
```diff
@@ -6,8 +6,17 @@ pythonImportsCheckPhase () {
...
     if [ -n "$pythonImportsCheck" ]; then
         echo "Check whether the following modules can be imported: $pythonImportsCheck"
-        export PYTHONPATH="$out/@pythonSitePackages@:$PYTHONPATH"
-        ( cd $out && eval "@pythonCheckInterpreter@ -c 'import os; import importlib; list(map(lambda mod: importlib.import_module(mod), os.environ[\"pythonImportsCheck\"].split()))'" )
+        pythonImportsCheckOutput=$out
+        if [ -n "$py" ]; then
+            echo "Using python specific output \$py for imports check"
+            pythonImportsCheckOutput=$py
+        fi
+        if [ -n "$python" ]; then
```
katexochen, line 14: This should be an `elif` instead, so we don't echo twice in case `$py` and `$python` are defined.

katexochen (approved):
Change lgtm, wdyt @SuperSandro2000 ?

Last staging-next cycle on master for 24.05 has started, so we should be good merging this into staging without affecting the release.

mweinelt (commented):
I think for python packages the `lib/site-packages` directory tree should always be in `out`. If you want to expose additional executables, use the  `bin` output.

msanft:
> I think for python packages the `lib/site-packages` directory tree should always be in `out`. If you want to expose additional executables, use the `bin` output.

What are `$py` / `$python` for then though?

mweinelt:
If we'd add a default output to the import check hook it should probably be `python`, not `py`.

msanft:
> If we'd add a default output to the import check hook it should probably be python, not py. And then packages with the py output should probably be normalized, if they want to use the hook.

Did so.

Issue #65718, opened 2019-08-01
Do not set DESTDIR in makeFlags


jtojnar:
[`DESTDIR`](https://www.gnu.org/software/automake/manual/html_node/DESTDIR.html) is used by some package managers as a poor man’s sandbox, transparent to the package. Setting `DESTDIR=$(out) PREFIX=` in `makeFlags`
will place the files in the correct directory but the program itself will not be aware of the files’ location. For example, if the source code is hardcoding `PREFIX "foo"`, it will try to open `foo` instead of `$out/foo`.

This is usually done when `make install` tries to create a `$(DESTDIR)/var` directory but as described above, it can have a negative side effect.

With `autotools` you can set `installFlags = [ "localstatedir=${placeholder "out"}/var" ];` or `installFlags = [ "sysconfdir=${placeholder "out"}/etc" ];` that should fix most instances of the issue. When the project using hand-written `Makefile`s, you need to check the files to see how to replace them. In the worst case, especially when the paths are hard-coded in the build files, you will need to patch the build files.

matthewbauer:
Some documentation on this could also go a long way. I've written up some basic docs:

------

The global configure policy of Nixpkgs (which should be set for you when using autotools, but not necessarily other build systems) is:

- prefix: ```"$(out)"```
- destdir: ```""```
- sysconfdir: ```"/etc"```
- localstatedir: ```"/var"```
- disable static, enable shared

It's a fairly common mistake to do ```DESTDIR=$(out) PREFIX=$(out)```, which is incorrect and ends up getting you a directory structure that looks like ```/nix/store/...-.../nix/store/...-.../...```.

To make sure installs go to the right place, you sometimes have to override make install flags like:

```
installFlags = [ "sysconfdir=$(out)/etc" "localstatedir=$out/var" ];

```

jtojnar:
This is a best practices issue, we still need to find a better place for it.

AndersonTorres:
Is this sprintable?

jtojnar:
Yes. Each expression needs to be evaluated separately so the labour can be divided.

jonringer:
Can we update the nixpkgs manual to include this under the https://nixos.org/manual/nixpkgs/stable/#var-stdenv-makefile section and close this issue?

It would also be a lot more discoverable

SuperSandro2000:
Can someone help me with #1218 ?

PR #304761, opened 2024-04-17
bcachefs-tools: 1.4.1 -> 1.7.0

JohnRTitor:
## Description of changes

- Update: bcachefs-tools: 1.4.1 -> 1.7.0
- Changelog: https://github.com/koverstreet/bcachefs-tools/compare/v1.4.1...v1.7.0
- Adding myself as maintainer
- Reformat the file using `nixfmt-rfc-style`

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [x] Tested, as applicable:

---
Pinging maintainers: @davidak @Madouura 

---


pkgs/tools/filesystems/bcachefs-tools/default.nix
```diff
@@ -43,14 +43,6 @@ stdenv.mkDerivation (finalAttrs: {
     makeWrapper
   ];
 
-  cargoRoot = "rust-src";
-  cargoDeps = rustPlatform.importCargoLock {
-    lockFile = ./Cargo.lock;
-    outputHashes = {
-      "bindgen-0.64.0" = "sha256-GNG8as33HLRYJGYe0nw6qBzq86aHiGonyynEM7gaEE4=";
```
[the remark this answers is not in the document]
JohnRTitor: I tried building with `${src}/Cargo.lock`, didn't work as it would fail to find the file.

Removing `importCargoLock` will fail to build the rust code.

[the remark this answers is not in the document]
alyssais: We could use `cargoHash`.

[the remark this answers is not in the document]
JohnRTitor: > We could use `cargoHash`.

Doesn't look like it's working.
```
┃        > unpacking source archive /nix/store/qyrbgxh64055v3v93dbasp8x6s2xhg4r-source
┃        > source root is source
┃        > Executing cargoSetupPostUnpackHook
┃        > unpacking source archive
┃        > do not know how to unpack source archive
```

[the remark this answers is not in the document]
JohnRTitor: > I tried building with `${src}/Cargo.lock`, didn't work as it would fail to find the file.

Nevermind, I should have used `finalAttrs.src`. Successfully built. Pushing.

JohnRTitor:
@ofborg eval
@ofborg build bcachefs-tools

pkgs/tools/filesystems/bcachefs-tools/default.nix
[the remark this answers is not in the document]
JohnRTitor: Looks like ofborg eval is failing with latest changes. Successfully builds though.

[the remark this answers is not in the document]
alyssais: It's failing because it's import-from-derivation, which is not allowed in Nixpkgs.

> We are not using `rustPlatform.buildRustPackage` to build a "rust-only package" here.

Ah, then what we need is `rustPlatform.fetchCargoTarball`.

JohnRTitor:
@ofborg build bcachefs-tools

pkgs/tools/filesystems/bcachefs-tools/default.nix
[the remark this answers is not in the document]
JohnRTitor: Nice, Ofborg eval working with fetchCargoTarball.
As for the update script, is there a script for automatically updating `hash` in `fetchCargoTarball`?

JohnRTitor:
@Mic92 @alyssais is this ready to be approved? Latest bcachefs fixes and features also depend on `bcachefs-tools` package.

alyssais:
Please squash the latest commit into the update, so there's not an intermediate point where it's using IFD.

JohnRTitor:
@alyssais is this fine now? I decided to keep the last two commits separately, for obvious reasons.

pkgs/tools/filesystems/bcachefs-tools/default.nix
```diff
@@ -65,20 +57,27 @@ stdenv.mkDerivation (finalAttrs: {
...
-  checkFlags = [ "BCACHEFS_TEST_USE_VALGRIND=no" ];
+  cargoDeps = rustPlatform.fetchCargoTarball {
+    src = finalAttrs.src;
+    hash = "sha256-RsRz/nb8L+pL1U4l6RnvqeDFddPvcBFH4wdV7G60pxA=";
+  };
 
   makeFlags = [
-    "PREFIX=${placeholder "out"}"
+    "DESTDIR=${placeholder "out"}"
+    "PREFIX="
```
alyssais, line 67: Are you sure this is right? https://github.com/NixOS/nixpkgs/issues/65718

pkgs/tools/filesystems/bcachefs-tools/default.nix
```diff
@@ -65,20 +57,27 @@ stdenv.mkDerivation (finalAttrs: {
...
     "VERSION=${finalAttrs.version}"
     "INITRAMFS_DIR=${placeholder "out"}/etc/initramfs-tools"
-    "BCACHEFS_FUSE=${toString fuseSupport}"
-  ];
+  ] ++ lib.optional fuseSupport "BCACHEFS_FUSE=1";
```
alyssais, line 70: Why this change?

pkgs/tools/filesystems/bcachefs-tools/default.nix
```diff
@@ -65,20 +57,27 @@ stdenv.mkDerivation (finalAttrs: {
...
+
+
+  # FIXME: Try enabling this once the default linux kernel is at least 6.7
+  doCheck = false; # needs bcachefs module loaded on builder
 
   preCheck = lib.optionalString (!fuseSupport) ''
     rm tests/test_fuse.py
   '';
+  checkPhase = "./target/release/bcachefs version";
```
alyssais, line 79: What's the point of setting this if `doCheck` is `false`?

pkgs/tools/filesystems/bcachefs-tools/default.nix
JohnRTitor: IIRC, When fixing the build errors, I encountered "a libfuse3 not found error". I initially looked at bcachefs build.nix, to see what has changed leading to this. BCACHEFS_FUSE=0 had weird behaviour like trying to link unavailable library anyway.
With this change, this should fix the bug.

pkgs/tools/filesystems/bcachefs-tools/default.nix
JohnRTitor: So a solution for this would be adding autotools to inputs and then using `installFlags = [ "localstatedir=${placeholder "out"}/var" ]; ` ?

I have not encountered any issues during normal usage with the new binary though.

pkgs/tools/filesystems/bcachefs-tools/default.nix
JohnRTitor: I have added this check for future, when default LTS kernel becomes > 6.7. Then we can enable doCheck = true and that will run this script

alyssais: But that overrides the default checkPhase, so now it won't run any of the tests it would have run before.

What's the motivation for testing that `bcachefs version` works?  It feels like not a very meaningful test unless I'm missing something.

pkgs/tools/filesystems/bcachefs-tools/default.nix
alyssais: No, because this package doesn't use autotools, and localstatedir would only cover that one directory, not the whole thing.

The right thing to do is generally set PREFIX, and not set DESTDIR, like it was before.  Why did you change it?

pkgs/tools/filesystems/bcachefs-tools/default.nix
JohnRTitor: Oh, I thought I would add a step to checkPhase not override it. While `bcachefs --version` is not an extensive check, it could show if there is a fatal runtime error due to missing libraries/packages.

I'll revert this so default checkPhase is used, if enabled that is.

pkgs/tools/filesystems/bcachefs-tools/default.nix
JohnRTitor: Looks like it was a oversight on my part. I was trying to cleanup this config by looking at the build.nix from the repository. It modify any existing behaviour at first glance, so I thought it was okay to include the change.

I'll revert this change.

alyssais:
Test failure on x86_64-linux should probably be investigated.  Is it just racy perhaps?

JohnRTitor:
> Test failure on x86_64-linux should probably be investigated. Is it just racy perhaps?

When I ran the test locally it did not actually "fail", it tested everything including in a VM, but then it took a lot of time at the finishing stage, [leading to the timeout](https://github.com/NixOS/nixpkgs/pull/304761/checks?check_run_id=24206442471).

JohnRTitor:
@Mic92 @alyssais do I need to make any other changes or is this ready to be merged?

onny (approved):

onny:
Result of `nixpkgs-review pr 304761` run on x86_64-linux [1](https://github.com/Mic92/nixpkgs-review)
<details>
  <ul>
    <li>bcachefs-tools</li>
  </ul>
</details>

alyssais (approved):

onny:
Thank you!

PR #304756, opened 2024-04-17
icu: Make static linking possible

NorfairKing:
## Description of changes

Make static linking possible for `icu`.

## Things done

- Built on platform(s)
  - [X] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
- [ ] Tested, as applicable:
  - [NixOS test(s)](https://nixos.org/manual/nixos/unstable/index.html#sec-nixos-tests) (look inside [nixos/tests](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests))
  - and/or [package tests](https://github.com/NixOS/nixpkgs/blob/master/pkgs/README.md#package-tests)
  - or, for functions and "core" functionality, tests in [lib/tests](https://github.com/NixOS/nixpkgs/blob/master/lib/tests) or [pkgs/test](https://github.com/NixOS/nixpkgs/blob/master/pkgs/test)
  - made sure NixOS tests are [linked](https://nixos.org/manual/nixpkgs/unstable/#ssec-nixos-tests-linking) to the relevant packages
- [ ] Tested compilation of all packages that depend on this change using `nix-shell -p nixpkgs-review --run "nixpkgs-review rev HEAD"`. Note: all changes have to be committed, also see [nixpkgs-review usage](https://github.com/Mic92/nixpkgs-review#usage)

  ^ This took ages, and filled my screen, so I didn't complete it.

---

Add a :+1: [reaction] to [pull requests you find important].

[reaction]: https://github.blog/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/
[pull requests you find important]: https://github.com/NixOS/nixpkgs/pulls?q=is%3Aopen+sort%3Areactions-%2B1-desc

PR #304772, opened 2024-04-17
icu: Make static linking possible

NorfairKing:
## Description of changes

Make it possible to statically link against `icu`.
See #304756

## Things done

- Built on platform(s)
  - [X] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/development/libraries/icu/make-icu.nix
+++ b/pkgs/development/libraries/icu/make-icu.nix
@@ -1,6 +1,6 @@
 { stdenv, lib, buildPackages, fetchurl, fixDarwinDylibNames, testers }:
 
-{ version, hash, patches ? [], patchFlags ? [] }:
+{ version, hash, patches ? [], patchFlags ? [], withStatic ? stdenv.hostPlatform.isStatic }:
 
 let
   # Cross-compiled icu4c requires a build-root of a native compile
@@ -36,9 +36,12 @@ let
       sed -e 's/LDFLAGSICUDT=-nodefaultlibs -nostdlib/LDFLAGSICUDT=/' -i config/mh-linux
     '';
 
+    dontDisableStatic = withStatic;
+
     configureFlags = [ "--disable-debug" ]
       ++ lib.optional (stdenv.isFreeBSD || stdenv.isDarwin) "--enable-rpath"
-      ++ lib.optional (stdenv.buildPlatform != stdenv.hostPlatform) "--with-cross-build=${nativeBuildRoot}";
+      ++ lib.optional (stdenv.buildPlatform != stdenv.hostPlatform) "--with-cross-build=${nativeBuildRoot}"
+      ++ lib.optional withStatic "--enable-static";
 
     enableParallelBuilding = true;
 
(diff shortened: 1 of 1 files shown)


pkgs/development/libraries/icu/make-icu.nix
```diff
@@ -1,6 +1,6 @@
 { stdenv, lib, buildPackages, fetchurl, fixDarwinDylibNames, testers }:
 
-{ version, hash, patches ? [], patchFlags ? [] }:
+{ version, hash, patches ? [], patchFlags ? [], withStatic ? false }:
```
symphorien, line 3:
```suggestion
{ version, hash, patches ? [], patchFlags ? [], withStatic ? stdenv.hostPlatform.isStatic }:
```

symphorien (commented):
what is the advantage of this flag in comparison to pkgsStatic.icu  ? (which already builds it seems)

NorfairKing:
> Are the rebuilds just because of dontDisableStatic now being false instead of missing (by default)?

I guess so, but that surprises me as well.

NorfairKing:
> what is the advantage of this flag in comparison to pkgsStatic.icu ? (which already builds it seems)

I'd like to be able to statically link outside of `pkgsStatic` and "just" using `dontDisableStatic = true` doesn't cut it for `icu`.

NorfairKing:
What are the steps to get this merged?

7c6f434c:
If you drop that line, the change becomes useless, but does the output path return to the old one?

NorfairKing:
@7c6f434c It took me a while to figure out what you meant, but no the path is different:

```
git checkout HEAD^
nix build .#icu && ls -l result
lrwxrwxrwx 1 syd users 54 Apr 19 00:57 result -> /nix/store/qwnmd1a9zxfjqgkhv3jyyf86r9jm4gsq-icu4c-73.2
git checkout icu-static
nix build .#icu && ls -l result                  131ms
lrwxrwxrwx 1 syd users 54 Apr 19 00:57 result -> /nix/store/c1m796by7343zg40d90qzr30lgc2k60f-icu4c-73.2
```

7c6f434c:
But if it changes anything about postInstall while `withStatic` is not enabled, that means we don't actually understand what it changes, and given that pkgsStatic.icu works if really needed, I'd want to understand which parts here are actually non-no-op in which case…

NorfairKing:
@7c6f434c is there a way to say `nix whats-different`?

NorfairKing:
@7c6f434c I just confirmed: It's the presence of the `dontDisableStatic` env var that changes the store hash. That's unavoidable though.

7c6f434c:
It is avoidable, but in an ugly way. I wanted it checked that this is the only rebuild-triggering change, though.

NorfairKing:
@7c6f434c  I just realised that I added the argument to the wrong function. It should be the former, so that it's overridable, not the later.

7c6f434c:
If you add the argument to the upper function, you need to re-call the resulting function after overriding and pass it the source hash again. don't you?

NorfairKing:
@7c6f434c Hm didn't think about it enough maybe. In any case I can't override static now so I still can't use this with `pkgsMusl` yet :P

PR #304861, opened 2024-04-17
wrapCC, wrapBintools: correctly cross-compile if buildPlatform != hostPlatform

sternenseemann:
## Description of changes

The wrapper scripts have two main dependencies (aside from `cc` and `bintools` which are a given):

- The shell used for executing the scripts, known as `shell` thus far.
- `expand-response-params`, a small C program which requires a *wrapped* compiler to be built.

Previously, these programs would be essentially taken from/built with the package set used to build the package set that the wrapper itself is a part of (usually `buildPackages`). This is incorrect, as it only “works” if `hostPlatform == buildPlatform` (relative to the wrapper) or it will be impossible to execute the dependencies.

This has the consequence that if you e.g. build `pkgsCross.aarch64-multiplatform.gcc`, it will be impossible to execute the resulting (wrapped) compiler due to exec format errors.

This weird situation stems from the stdenv bootstrapping process where we have the concrete problem that building a shell and C program requires a wrapped C compiler, but we also require these to build the wrapper. The solution is to either do without the dependencies or take them from the previous bootstrapping stage and, ultimately, `bootstrapTools`. These is sound in this situation since during (initial) stdenv bootstrapping `buildPlatform == hostPlatform == targetPlatform`. It is, however, incorrect to have this leak into the main package set where cross compilation may happen.

The solution to this problem is to make the wrappers behave correctly in the main package set (taking cross into account) and worry about bootstrapping later. This is exactly what I've done:

- Instead of `shell` which defaults to `stdenvNoCC.shell` (which is usually equivalent to `buildPackages.runtimeShell`), we use `runtimeShell` which is an attribute that exposes the default shell we use on the host platform.
- Instead of ad hoc building `expand-response-params` inline in the wrapper derivation by trying to pull a `stdenv` out of `buildPackages` (so that we are effectively using `buildPackages.buildPackages.gcc/clang`), we just take it as an ordinary input (which works perfectly in the main package set) and let the caller worry about bootstrapping.

In a second step I've reflected these changes in the darwin and linux stdenvs, making sure that no derivation hash changes by inserting extra bootstrapping instructions where necessary (the single rebuild is caused by the newly added `expand-response-params` attribute). Some of these may actually be unnecessary (i.e. we may be able to build a proper `runtimeShell` earlier etc.), but this will have to be investigated separately as that would cause a stdenv rebuild.

This change is best reviewed commit by commit.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [x] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/build-support/bintools-wrapper/default.nix
+++ b/pkgs/build-support/bintools-wrapper/default.nix
@@ -8,7 +8,8 @@
 { name ? ""
 , lib
 , stdenvNoCC
-, bintools ? null, libc ? null, coreutils ? null, shell ? stdenvNoCC.shell, gnugrep ? null
+, runtimeShell
+, bintools ? null, libc ? null, coreutils ? null, gnugrep ? null
 , netbsd ? null, netbsdCross ? null
 , sharedLibraryLoader ?
   if libc == null then
@@ -419,7 +420,8 @@ stdenvNoCC.mkDerivation {
   env = {
     # for substitution in utils.bash
     expandResponseParams = "${expand-response-params}/bin/expand-response-params";
-    shell = getBin shell + shell.shellPath or "";
+    # TODO(@sternenseemann): rename env var via stdenv rebuild
+    shell = (getBin runtimeShell + runtimeShell.shellPath or "");
     gnugrep_bin = optionalString (!nativeTools) gnugrep;
     wrapperName = "BINTOOLS_WRAPPER";
     inherit dynamicLinker targetPrefix suffixSalt coreutils_bin;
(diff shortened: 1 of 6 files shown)


sternenseemann:
cc @alexandretunstall

sternenseemann:
I think the `pkgs/by-name` CI error can be ignored, since I wouldn't want to move `expand-response-params` out of `build-support` for now.

Ericson2314 (approved):
Great work!

Issue #292468, opened 2024-02-29
Documentation: style guide for use of `with`


lolbinarycat:
## Problem

i see a lot of `license = with licenses; [ mit ]` in PRs recently.

compared to the simpler `license = licenses.mit`, this is both longer and more syntactically complex, taking longer to read for humans, and possibly longer to evaluate too.

## Proposal
this should be avoided for single-license packages (which is most of them).

this would not apply to `meta.maintainers`, for the following reason:
* it is more likely to have multiple values
* it is much more likely to have extra values added in the future.
* `maintainers` is plural, which implies a list

i would be willing to write this into `pkgs/README.md` but i'm not going to open a PR until i've gotten some feedback, it wouldn't make sense for me to single-handedly decide what the best practice is.

---

Add a :+1: [reaction] to [issues you find important].

[reaction]: https://github.blog/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/
[issues you find important]: https://github.com/NixOS/nixpkgs/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions-%2B1-desc

eclairevoyant:
relevant: #208242

lolbinarycat:
[the nix best practices page](https://nix.dev/guides/best-practices#with-scopes) seems to imply that `with` should basically never be used, this seems very extreme and would require rewriting basically every file in nixpkgs.

if we instead just recommend limiting its use to `meta = with lib; { }` and `maintainers = with maintainers; [ ]`, that would allow most unproblematic usages of `with` to continue, while discouraging unusual and sometimes problematic usages.

AndersonTorres:
> i see a lot of `license = with licenses; [ mit ]` in PRs recently.

In my opinion, license should be a list, rather than "a list of licenses or a single license". The "singleton", even if being more typical, looks like an exception and not a rule.

So, as a personal preference, I am writing `meta.license` as a list unconditionally.

>  maintainers is plural, which implies a list

License should be called licenses, precisely because there is at least one multiple-licensed project.

>  seems to imply that with should basically never be used, this seems very extreme and would require rewriting basically every file in nixpkgs.

There is more: we didn't find a decent replacement for `with` yet. The best we can do is to limit its use to predictable and easy-to-not-be-fooled examples.

To me, the exceptions are single-line statements, like

1. Lists, with the restriction that the `with` applies to each and every element:
    ```nix
       with X; [ A B C D ];
       # is shorter than
[...]

Issue #277994, opened 2024-01-01
Tracking issue: superfluous use of `pname`


AndersonTorres:
There is a huge quantity of appearances of patterns like ` . . . = pname`, ` . . . = finalAttrs.pname`, `${pname}` and alike in Nixpkgs.

After some amount of ~heated~ discussions with @SuperSandro2000, I come to the conclusion that this should be discouraged.

There are some reasons I can enumerate for it:

0. Many, if not all, appearances of `pname` as an rvalue are basically constants - indeed they are constants since Nix is [a functional language](https://www.defmacro.org/2006/06/19/fp.html). They can be hardcoded with no loss (and some very small performance gains since the compiler will not need to consult the symbol tables).

   - It can be argued that other attributes change constantly, like the ubiquitous `version`.
     However, they change based on a familiar modification pattern from the upstream project, whereas things like `pname` rarely change.

1. Related to  the point above, this use of pname creates otherwise superfluous in-code dependencies 
   1. This overcomplicates the static reasoning about the code
   2. This overcomplicates the use of `override*` functionalities
   3. The issues above are especially more pronounced in the context of [finalAttrs design pattern](https://github.com/NixOS/nixpkgs/pull/119942)

2. Sometimes - indeed, most of the time - things like GitHub's `repo` and other source origins are not identical to `pname`. Some examples:

    - OpenMSX and Bochs.
       - By convention `pname` is all-lowercase.
         - An exception consecrated by the use is, curiously,  SDL.
    - In the old times, PIL - the Python Imaging Library - had `pname = pil` and `src.url = ". . ./Imaging-${version}.tar.gz"`
    - Before migration to a dedicated forge, the MIR Project was called Unity-8.
      After being passed from Canonical to the community, it was renamed Lomiri, but the forge did not change his name. Indeed the old, now-frozen repo is still called https://github.com/ubports/unity8.
      Now it lives over https://gitlab.com/ubports/development/core/lomiri

4. Nix is not a macro language like M4 or the CPP preprocessor. 
   1. Just because two things have the same value, we should not treat them as if they are somehow connected
   2. Just because two things are somehow connected, this connection should not necessarily be converted into the code
   4. Pushing to the absurd and linking to the point above, it would be insane to parameterize each occurrence of `pname`.

References:

-  https://github.com/nix-community/nixpkgs-lint/issues/21

oxalica:
> 3\. 3. This overcomplicates the use of `override*` functionalities

[...]

Issue #178468, opened 2022-06-21
Enable `config.strictDepsByDefault = true` by default


trofi:
`strictDeps` helps keeping cleaner `PATH`s for cross-compiled packages. But it's also useful to have for native builds: it helps writing nix derivations that have better chance of working when cross-compiled later.

A bunch of packages will need to sort out their `buildInputs / nativeBuildInputs` depends. Or more precise `depsBuildBuild & co.` finer grained equivalents.

https://github.com/NixOS/nixpkgs/blob/913ee4f295c6a0cc801effd3c72906639a31fdb0/pkgs/stdenv/generic/make-derivation.nix#L345-L346

erdnaxe:
libxml2 seems to break with `config.strictDepsByDefault = true`. Adding `python` in `nativeBuildInputs` fixes the issue.
I am not really confident to propose this as a patch as I don't understand why this exists: `pythonSupport ? enableShared && stdenv.buildPlatform == stdenv.hostPlatform`.

Artturin:
> libxml2 seems to break with `config.strictDepsByDefault = true`. Adding `python` in `nativeBuildInputs` fixes the issue. I am not really confident to propose this as a patch as I don't understand why this exists: `pythonSupport ? enableShared && stdenv.buildPlatform == stdenv.hostPlatform` (in pkgs/development/libraries/libxml2/default.nix).

python is probably being disabled because cross-compilation enables strictDeps and so breaks the python detection

i'll fix it

erdnaxe:
There seems to be a similar issue to libxml2 with libxslt.

I am building a aarch64-linux NixOS system from master using `config.strictDepsByDefault = true`. I will update the following list with the files in which I had to add `strictDeps = false;` to get a working system:

 * [ ] pkgs/development/libraries/libxml2/default.nix
 * [ ] pkgs/development/libraries/libxslt/default.nix
 * [ ] pkgs/os-specific/linux/kmod/default.nix

Artturin:
libxml2, libxslt: https://github.com/NixOS/nixpkgs/pull/194941

Artturin:
perlPackages.{Po4a,AlienLibxml2},kmod,liblouis https://github.com/NixOS/nixpkgs/pull/218503

AndersonTorres:
Ward because I need to learn this.

infinisil:
[...]

PR #119942, opened 2021-04-20
stdenv.mkDerivation: overlay style overridable recursive attributes

roberth:
###### Motivation for this change

 - Allows to get rid of confusing `rec { }` in `mkDerivation` calls. Overriding an attribute in this style causes derived attributes to be overridden as expected.

 - Allows `passthru` attributes to reference the outputs

 - Fixes #119407, the problem where if you use `overrideAttrs`, the tests in `passthru` still use the old package.

 - Should allow cleanup of python and rust packaging functions by using "overlay" layers on a prototype derivation, rather than the current workarounds.

See updated manual for usage explanation.

Update: this problem was also brought up in this discussion: https://github.com/NixOS/nixpkgs/pull/119731#pullrequestreview-638309038

Update: "overlay" part minus `public` was also discovered and implemented by lilyball https://github.com/NixOS/nixpkgs/pull/94198

###### Things done

- [ ] Tested using sandboxing **not applicable:** changes are evaluation-only ([nix.useSandbox](https://nixos.org/nixos/manual/options.html#opt-nix.useSandbox) on NixOS, or option `sandbox` in [`nix.conf`](https://nixos.org/nix/manual/#sec-conf-file) on non-NixOS linux)
- Built on platform(s)
   - [ ] NixOS
   - [ ] macOS
   - [ ] other Linux distributions
- [ ] Tested via one or more NixOS test(s) if existing and applicable for the change (look inside [nixos/tests](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests))
- [ ] Tested compilation of all pkgs that depend on this change using `nix-shell -p nixpkgs-review --run "nixpkgs-review wip"`
- [ ] Tested execution of all binary files (usually in `./result/bin/`)
- [ ] Determined the impact on package closure size (by running `nix path-info -S` before and after)
- [x] Ensured that relevant documentation is up to date
- [x] Fits [CONTRIBUTING.md](https://github.com/NixOS/nixpkgs/blob/master/.github/CONTRIBUTING.md).

PR #304942, opened 2024-04-18
Mednafen refactors

AndersonTorres:
## Description of changes

Closes https://github.com/NixOS/nixpkgs/pull/304028

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---


pkgs/by-name/me/mednaffe/package.nix
```diff
@@ -0,0 +1,52 @@
...
+  autoreconfHook,
+  fetchFromGitHub,
+  gtk3,
+  mednafen,
+  pkg-config,
+  stdenv,
+  wrapGAppsHook,
+}:
+
+stdenv.mkDerivation (finalAttrs: {
```
RossComputerGuy, line 12: Why `(finalAttrs: {` instead of just `rec {`?

pkgs/by-name/me/mednafen-server/package.nix
```diff
@@ -0,0 +1,30 @@
+{
+  lib,
+  stdenv,
+  fetchurl,
+}:
+
+stdenv.mkDerivation (finalAttrs: {
```
RossComputerGuy, line 7: Same here. Why `(finalAttrs: {` instead of just `rec {`?

pkgs/by-name/me/mednafen-server/package.nix
```diff
@@ -0,0 +1,30 @@
...
+    hash = "sha256-uJmxaMW+bydfAXq8XDOioMoBOLUsi5OT2Tpbbotsp3Y=";
+  };
+
+  strictDeps = true;
+
+  postInstall = ''
+    install -m 644 -Dt $out/share/mednafen-server standard.conf
+  '';
+
+  meta = {
```
RossComputerGuy, line 22: If you make `meta = with lib; {` then you can remove `lib.` from anything below this line.

pkgs/by-name/me/mednafen/package.nix
```diff
@@ -1,33 +1,39 @@
...
+  libjack2,
+  libsamplerate,
+  libsndfile,
+  pkg-config,
+  stdenv,
+  zlib,
 }:
 
-stdenv.mkDerivation rec {
+stdenv.mkDerivation (finalAttrs: {
```
RossComputerGuy, line 22: It was just fine with `rec` before. Why the change?

pkgs/by-name/me/mednafen/package.nix
```diff
@@ -1,33 +1,39 @@
...
-    url = "https://mednafen.github.io/releases/files/${pname}-${version}.tar.xz";
+    url = "https://mednafen.github.io/releases/files/mednafen-${finalAttrs.version}.tar.xz";
     hash = "sha256-2j+88Ch3+b4PAov6XRy1npU6QEm5D+fjk4ijOG2fNi4=";
   };
 
-  nativeBuildInputs = [ pkg-config ];
+  outputs = [ "out" "doc" ];
+
+  nativeBuildInputs = [
+    SDL2
```
RossComputerGuy, line 34: Why is `SDL2` in `nativeBuildInputs` if it is in `buildInputs`?

pkgs/by-name/me/mednafen/package.nix
```diff
@@ -48,16 +54,21 @@ stdenv.mkDerivation rec {
...
+
   postInstall = ''
-    mkdir -p $out/share/doc
-    mv Documentation $out/share/doc/mednafen
+    mkdir -p $doc/share/doc
+    mv Documentation $doc/share/doc/mednafen
   '';
 
-  meta = with lib; {
+  meta = {
```
RossComputerGuy, line 71: Why the removal of `with lib;`?

RossComputerGuy (commented):
Good start, though needs some improvements. Another thing to note, please stick with the [git commit naming conventions we use](https://github.com/NixOS/nixpkgs/blob/master/CONTRIBUTING.md#commit-conventions). You can use `git rebase -i` to squash and rename commits so the history looks cleaner and you can follow the convention.

pkgs/by-name/me/mednafen/package.nix
AndersonTorres: When using `strictDeps = true`, we need two SDLs: the one that runs `sdl-config` from building platform and the other that installs SDL libs to target platform.

https://discourse.nixos.org/t/always-be-cross-compiling-slides-by-matthew-bauer-pdf/30287

pkgs/by-name/me/mednaffe/package.nix
RossComputerGuy: Because the majority of other packages use rec and achieve the same result as what wrapping the attributes in a function does.

pkgs/by-name/me/mednafen-server/package.nix
AndersonTorres: `with` does not work like the distributive axiom of a field (as suggested by your remark), and its rules are unintuitive.

https://nix.dev/anti-patterns/language#with-attrset-expression

https://github.com/NixOS/nix/issues/490

https://github.com/NixOS/nixpkgs/issues/208242

pkgs/by-name/me/mednaffe/package.nix
AndersonTorres: > Because the majority of other packages use rec

Popularity does not make a good argument here, or else things like treewide banishing unquoted URLs from the codebase would never happen.

> and achieve the same result as what wrapping the attributes in a function does

Although being spread over the whole Nixpkgs codebase, `rec` is an anti-pattern:

https://nix.dev/guides/best-practices#recursive-attribute-set-rec

Then, `finalAttrs` design pattern (a.k.a. __overlay style overridable recursive attributes__) was merged more than 3 years ago, to overcome some of the problems of `rec`.

> Also, what does this change fix or improve that required the removal of rec and wrapping it in a function?

#119942

RossComputerGuy: > Popularity does not make a good argument here, or else things like treewide banishing unquoted URLs from the codebase would never happen.

I am aware, however I believe the change should be made with a purpose.

> Although being spread over the whole Nixpkgs codebase, rec is an anti-pattern:
>
> https://nix.dev/guides/best-practices#recursive-attribute-set-rec
>
> Then, `finalAttrs` design pattern (a.k.a. overlay style overridable recursive attributes) was merged more than 3 years ago, to overcome some of the problems of `rec`.

Is there an RFC or issue about treewide removal of `rec`? The way it is upstream works fine currently. That nix.dev page recommends using `let...in` as an option.

pkgs/by-name/me/mednafen-server/package.nix
RossComputerGuy: That last one is about overuse of `with`. In this case, I wouldn't consider using `with lib;` for assigning values in `meta` to be an overuse.

pkgs/by-name/me/mednafen/package.nix
AndersonTorres: How exactly?

If in the sense of "it works on my machine, so it is not needed", then it is not needed.

If in the sense of "bad with it, worse without it", it is seriously needed.

Among many things, `strictDeps` enforces a cleaner separation of rules and concerns among inputs and phases - a thing that is easily blurred in native compilation but can be very tricky in cross compilation.

https://nixos.org/manual/nixpkgs/stable/#ssec-stdenv-dependencies-reference

https://github.com/NixOS/nixpkgs/issues/178468

RossComputerGuy: Understandable, I wasn't aware that this was an issue with `nativeBuildInputs` and `buildInputs`.

pkgs/by-name/me/mednaffe/package.nix
AndersonTorres: > I am aware, however I believe the change should be made with a purpose.

"Getting rid of `rec` (and other bad design patterns)" is a purpose, I suppose.

> Is there an RFC or issue about treewide removal of rec?

Reasoning backwards it is.

First, technically there is no need of a "treewide-rec-banish RFC" to banish `rec` from Nixpkgs.

Second, refactoring the expressions I maintain is in my discretion.

Third, objectively the code is not worse than before; it is still perfectly readable and understandable.

Fourth, and related to second, subjectively the code is better since it avoids pitfalls from bad design patterns, and it is better to me to maintain and improve it.

> The way it is upstream works fine currently.

Not a sufficient reason to hinder code refactoring.

> That nix.dev page recommends using let...in as an option.

Yes, and I do use it sometimes. 

But this nix.dev tutorial recommends let-in because, among other things, its purpose is to use more general approaches, especially in external codebases that will not necessarily have direct access to Nixpkgs.

Also, things like #119942 did not exist in that time.

RossComputerGuy: Alright, I understand. 👍

pkgs/by-name/me/mednafen-server/package.nix
AndersonTorres: Until a nasty `with` scope invasion appears:


NixOS/nixpkgs#292468 (comment)

pkgs/by-name/me/mednafen/package.nix
```diff
@@ -1,33 +1,39 @@
...
 }:
 
-stdenv.mkDerivation rec {
+stdenv.mkDerivation (finalAttrs: {
   pname = "mednafen";
   version = "1.29.0";
 
   src = fetchurl {
-    url = "https://mednafen.github.io/releases/files/${pname}-${version}.tar.xz";
+    url = "https://mednafen.github.io/releases/files/mednafen-${finalAttrs.version}.tar.xz";
```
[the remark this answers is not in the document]
AndersonTorres: https://github.com/NixOS/nixpkgs/issues/277994

RossComputerGuy (approved):

Issue #295029, opened 2023-11-17
Attaching more kinds of data to packages?


abathur:
Curious what you think about taking something I've been doing with resholve and applying it more broadly within nixpkgs. I'll start with a little background on what I've been doing, and _then_ I'll get to the point :)

---
One of the trickier problems resholve has to wrestle with is how to find/replace commands lurking in the arguments to other commands (like `find ... -exec rm {} +`).

It's the kind of problem we can only take bites out of:

1. First we perform some static analysis and use it to ignore the safest subset of executables: those we have scanning rules for and the scanner finds no evidence of a well-known exec API.

2. Then we scan for executables that look like exec wrappers to apply the analysis from the underlying executable to the wrapper.

3. Since the first analysis produces a lot of false positives (the executable uses exec but not on a CLI argument, executables we have no rules for scanning, etc.), there are systems for overriding the automated analysis with human judgements. (Everything left at this point is singled out for more scrutiny--but that's beyond the scope of my query.)

resholve itself doesn't actually generate the analyses. It just accepts a directory with text files that encode the information. The proof-of-concept implements this as a separate tool called `binlore` which includes a Nix function to run the analyses on a single input and output each analysis as `$out/execers` and the other as `$out/wrappers`. (I realize this is similar to `$out/nix-support/*`, but I didn't want to dump something ~unofficial/tangential in that namespace.)

`binlore` also bundles a ~centralized collection of overrides to apply when they are a prefix match for the package name. It's also possible to add decentralized overrides where the resholve API is used.

---

I think it might be interesting to explore this idea more broadly/officially in nixpkgs. 

I imagine this taking the form of a generic mechanism for attaching details like this (and probably a community mechanism for debating what to include?) for a few reasons:

- An override mechanism that works by string matching at the left hand of package names has obvious problems. It's a compromise that enabled me to work independently without a tight coupling to package versions, but it would be much more accurate to be able to directly add overrides to expressions/packages. 

- Some information is cheaper and/or more accurate when generated at build time.
	- All of the wrapper generators, for example, could publish a wrapper->wrapped mapping as they go. This can hold whether it's a shell wrapper, or a binary wrapper, or a wrapper in some other interpreted language.

	- binlore performs static analysis directly on the executable because it's the lowest-common-denominator, but some decisions can inevitably be simpler or better with build-time access to source.

[...]

PR #304956, opened 2024-04-18
binlore: migrate override lore to package passthru

abathur:
# Background

> Note: trying to keep this from sprawling; there's a longer background/motivation in the issue below if you want more context:
> - #295029

resholve ~needs metadata about which executables are wrappers and which might exec their args.

I implemented these concepts in a separate ~package called `binlore` because I suspect there are other uses of this (and similar) metadata (and I didn't want to bog down on a quest to add metadata of unproven utility to core nixpkgs machinery).

> Aside: I named this metadata `lore`. I'm not married to the name, but I do think it needs something to avoid confusion with package.meta. I'll use `lore` from here on.

`binlore` uses YARA rules and post-processing logic to scan executables and generate lore. It also merges manual overrides (if available) with generated lore.

(The current implementation is quick but not very smart, so it needs manual overrides when generated lore is wrong.)

# Problem

The manual overrides mentioned above are bundled with binlore's source, but this has scaled poorly. Moving them into nixpkgs is the goal of this PR. 

<details><summary>The main reasons it scales poorly are:</summary>

- People working on nixpkgs can't readily self-service fixes/improvements.

- Some obvious limitations of the longest-pname-prefix-match lore override mechanism have started to bite in past several months.

- Now that xdg-utils is packaged with resholve, updates to overrides in binlore's source must go through staging.

</details>

# Description

This PR moves lore overrides from binlore's source repo to the passthru of the target packages. 

Broad strokes:

- Change how lore is merged/gathered:
	- Stop applying overrides from binlore's source. 
	- Merge lore from a package's `$out/nix-support/<loretype>` if it exists. This mechanism prepares binlore to collect lore generated by `makeWrapper` and friends. 

	    (I'm not including those changes here. It's a huge rebuild that makes all of this harder to test, so I'll save it for a follow-up PR.)

	- Merge lore from `<package>.lore` if it exists.

	    (Happy to debate this name. I shied away from using `overrideLore` since `override` is more or less a contract in nixpkgs by this point and I don't want to mislead.)

- Add a utility function (`binlore.synthesize`) for ~synthesizing lore overrides. It rolls in a small Shell DSL for efficiently expressing overrides.

    (Happy to debate this name. I again shied away from using `override` since that term is more or less a contract in nixpkgs by this point.) 

- Reimplement most overrides using `passthru.lore = binlore.synthesize ...;`. (I'm not porting _all_ overrides. I'm paring them back to the set needed for nixpkgs and other public repos found via search.)

- Expand resholve's passthru tests to better exercise the machinery.

# TODO

Opened as draft to gather feedback. Will probably reopen against staging when it feels ~ready. For now, hoping to figure out:

- [ ] roughly how much consensus/consent this will take
- [ ] how much documentation this needs and where it should be
- [ ] whether any names/terms should change (ideally before I write docs per above)

My branch for this is based on a recent nixpkgs-unstable commit and I generally test with `nix-build -A resholve.tests --arg config '{ allowBroken = false; }'`.

### DIFF
--- a/pkgs/by-name/pd/pdf2odt/package.nix
+++ b/pkgs/by-name/pd/pdf2odt/package.nix
@@ -49,6 +49,10 @@ resholve.mkDerivation rec {
       imagemagick
       zip
     ];
+    execer = [
+      # zip can exec; confirmed 2 invocations in pdf2odt don't
+      "cannot:${zip}/bin/zip"
+    ];
   };
 
   meta = with lib; {
--- a/pkgs/development/libraries/libarchive/default.nix
+++ b/pkgs/development/libraries/libarchive/default.nix
@@ -23,6 +23,9 @@
 , cmake
 , nix
 , samba
+
+# for passthru.lore
+, binlore
 }:
 
 assert xarSupport -> libxml2 != null;
@@ -124,4 +127,11 @@ stdenv.mkDerivation (finalAttrs: {
   passthru.tests = {
     inherit cmake nix samba;
   };
+
+  # bsdtar is detected as "cannot" because its exec is internal to
+  # calls it makes into libarchive itself. If binlore gains support
+  # for detecting another layer down into libraries, this can be cut.
+  passthru.lore = (binlore.synthesize finalAttrs.finalPackage ''
+    execer can bin/bsdtar
+  '');
 })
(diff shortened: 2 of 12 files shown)


abathur:
@ofborg build resholve.tests

abathur:
@ofborg build resholve.tests

abathur:
There's a merge conflict, but that's fine for now--it doesn't affect the main questions here, and I already plan to close and reopen against staging once this is actually ready for review.

PR #305004, opened 2024-04-18
nixos/nh: coerce flake option to string

PerchunPak:
## Description of changes

I don't know why no one reported this but for me Nix copies path to /nix/store, so I would like to use string for `$FLAKE`.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
    - My system evaluated with that commit, so I think it counts.
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

cc @viperML

---

### DIFF
--- a/nixos/modules/programs/nh.nix
+++ b/nixos/modules/programs/nh.nix
@@ -17,6 +17,7 @@ in
     flake = lib.mkOption {
       type = lib.types.nullOr lib.types.path;
       default = null;
+      apply = toString;
       description = ''
         The path that will be used for the `FLAKE` environment variable.
 


nixos/modules/programs/nh.nix
```diff
@@ -15,7 +15,7 @@ in
     package = lib.mkPackageOption pkgs "nh" { };
 
     flake = lib.mkOption {
-      type = lib.types.nullOr lib.types.path;
+      type = with lib.types; nullOr (either path str);
```
PerchunPak, line 18: Please correct me if this is not a right way to write such type. I am still pretty new to Nix

viperML:
What value are you using that is not supported?

PerchunPak:
> What value are you using that is not supported?

I tried:
```nix
environment.variables = {
  "FOO" = /etc/nixos;
  # hacky way to say /home/perchun/dotfiles without hardcoding /home/perchun
  # if `/etc/nixos` doesn't work, then just `/home/perchun/dotfiles` won't either
  "BAR" = /. + config.home-manager.users.perchun.home.homeDirectory + "/dotfiles";
};
```
Both were copied to `/nix/store`:
```bash
~ ❯ echo $FOO
/nix/store/92iarj2qa7kfalx247sqa333i995zycr-nixos
~ ❯ echo $BAR
/nix/store/kclrsp4bmz1c1385ljjyk6458i0cr10r-dotfiles
```

PerchunPak:
Looks like home-manager encouraged same problem
https://github.com/nix-community/home-manager/commit/e6e49ad73c7d8cbd9f6622f0a3330ce7b71c8a86

PerchunPak:
Hmm, I remember it raised an error about invalid type, but I probably did something wrong. It works now.

Can I update this PR to add `apply = toString;`?

viperML:
toString conversion is performed by `environment.sessionVariables` https://github.com/NixOS/nixpkgs/blob/53a2c32bc66f5ae41a28d7a9a49d321172af621e/nixos/modules/config/system-environment.nix#L35

PerchunPak:
You are right, but for some reason after adding `apply = toString` Nix didn't copy it to /nix/store. I have no idea why though

nixos/modules/programs/nh.nix
eclairevoyant: Actually the `path` type already permits strings:
https://github.com/NixOS/nixpkgs/blob/2cf972b6039f6cec292b018c434719ff73bddbd2/lib/types.nix#L535

viperML:
I think this PR should be closed

PerchunPak:
But currently flake option doesn't work (if you provide path), you can check it by yourself

Why this fix works doesn't make sense for me too

viperML:
```
$ cat configuration.nix
{
  config,
  ...
}:
{
  boot.isContainer = true;

  programs.nh = {
    enable = true;
    flake = /etc/nixos;
  };
}

$ nix build -I nixos-config=configuration.nix -I nixpkgs=. -f ./nixos config.system.build.toplevel

$ grep FLAKE result/etc/set-environment
export FLAKE="/etc/nixos"
```

If anything, you can't access /etc/nixos with flake's evaluation sandbox, in which case you can use a string type, not a path type (`"/etc/nixos"`)

viperML:
Also note that toString has a different implementation that string interpolation:

```
$ cat flake.nix
{
  outputs = _: {
    a = builtins.toString /etc/nixos;
    b = "${/etc/nixos}";
  };
}

$ nix eval .#a
"/etc/nixos"

$ nix eval .#b
"/nix/store/92iarj2qa7kfalx247sqa333i995zycr-nixos"
```

https://github.com/NixOS/nix/issues/7327

PerchunPak:
I don't understand what are you trying to say

```
$ cat configuration.nix
{
  config,
  ...
}:
{
  boot.isContainer = true;

  programs.nh = {
    enable = true;
    flake = /etc/nixos;
  };
}
$ git add configuration.nix
$ git checkout nh-fix-flake-type
$ nix build -I nixos-config=configuration.nix -I nixpkgs=. -f ./nixos config.system.build.toplevel
$ grep FLAKE result/etc/set-environment
export FLAKE="/etc/nixos"
$ git checkout upstream/nixos-unstable
$ nix build -I nixos-config=configuration.nix -I nixpkgs=. -f ./nixos config.system.build.toplevel
$ grep FLAKE result/etc/set-environment
export FLAKE="/nix/store/92iarj2qa7kfalx247sqa333i995zycr-nixos"
```

eclairevoyant:
What viper is saying is that `environment.variables.apply` is implemented as so:
https://github.com/NixOS/nixpkgs/blob/9d7339877a6d615fad1a0dd488e9d2dc7197fa63/nixos/modules/config/shells-environment.nix#L46

So rather than providing a path (which will undergo string interpolation), provide the string directly.

In other words, if you set `programs.nh.flake = "/etc/nixos";` the PR should not be necessary.

PerchunPak:
That is right, it will work, but there is no reason to silently fail when user provides path.

`FLAKE` variable is used as a path to a directory to the system configuration, which then will be used in `nh os switch` or similar commands. If `nh os switch` will silently use copied immutable configuration from `/nix/store`, a user that accidentally set up path as path, not string, will be confused as to why `nh os switch` doesn't see any changes in configuration:

```bash
$ grep flake modules/nh.nix
    flake = /etc/nixos;  # user didn't know it MUST be a string!
$ nh os switch
# because there is no argument after the word `switch`, nh will try to get configuration from
# $FLAKE variable. but $FLAKE points to /nix/store, which means it was just copied from /etc/nixos,
# and changes that user maid in `/etc/nixos` won't apply to previously copied copy in `/nix/store` to which
# $FLAKE now points. that means, `nh os switch` will not find any changes until a user specifies
# `/etc/nixos` explicitly as an argument
```

Even though this is how environment variables in nix usually work, users should not to have to know how `flake` option works under the hood. Providing a path type to this option should not lead to unexpected and invalid behavior, but instead, the provided value should be fixed or an error should be raised. I cannot come up with a scenario, where user would want `$FLAKE` to point to `/nix/store`.

eclairevoyant:
> users should not to have to know how `flake` option works under the hood

Okay, then put another way: this is more consistent with how paths vs strings are treated in general; paths are copied to the store, path strings aren't. `environment.etc` is another example of this.

If you want to _discourage_ or _disallow_ store paths then IMO using a warning/assertion makes more sense than silently behaving differently.

PerchunPak:
> If you want to discourage or disallow store paths then IMO using a warning/assertion makes more sense than silently behaving differently.

So will you accept the PR if I implement this solution?

viperML:
I cannot reproduce this: <https://github.com/NixOS/nixpkgs/pull/305004#issuecomment-2066155280> 

```
$ nix build github:viperML/nixpkgs-305004-repro#nixosConfigurations.default.config.system.build.toplevel
trace: warning: system.stateVersion is not set, defaulting to 24.05. Read why this matters on https://nixos.org/manual/nixos/stable/options.html#opt-system.stateVersion.

$ grep FLAKE result/etc/set-environment
export FLAKE="/etc/nixos"
```

PerchunPak:
Bisected fix to https://github.com/NixOS/nixpkgs/commit/c5371710de5d244b45e0c3e689e3afef1dddf990

Anyway, thanks for your reviews!

viperML:
As a final note, maybe it would be nice to have a `flakeref` type in `lib.types` 🙂

PerchunPak:
Should I reopen it due to https://github.com/NixOS/nixpkgs/pull/308779/commits/ed38b9699a3a6680eb323cfbe0f199295ce3e3b7 (from https://github.com/NixOS/nixpkgs/pull/308779)?

PR #305121, opened 2024-04-18
Add NixOS options: fileSystems.<name>.enable and boot.specialFileSystems.<name>.enable

PaulGrandperrin:
I need this, and I really can't find a nice workaround without maintaining a local fork of `nixpkgs` so here we are :)

**What this does**

this adds:

`fileSystems.<name>.enable`
and 
`boot.specialFileSystems.<name>.enable`

both are enabled by default, so everything works as before.

(I only need the second one, but they share the same submodule, so both are modified)

**Why I need this**

I want to create some nspawn containers with NixOS on the guest (but without sharing /nix so that the guest is fully autonomous).

The issue is that the guest NixOS will try and fail to mount `/dev`, `/dev/pts`, `/dev/shm` and `/run`, which is ok because they are already mounted by the host when preparing the container.

So I want to disable them in `boot.specialFilesystems` but that currently impossible without overriding everything, including useful entries like `/run/key` and `/run/wrappers`.

I need something like `environment.etc.<name>.enable` to disable them without removing everything.

**About this PR**

I created this PR as a draft to get feedback on the idea. The current code is mainly for illustration.

I don't really know who to mention, so here are the 2 last committers to this file: @dasJ @ElvishJerricco, please add anyone who might be interested.

### DIFF
--- a/nixos/modules/tasks/filesystems.nix
+++ b/nixos/modules/tasks/filesystems.nix
@@ -11,7 +11,10 @@ let
   isNonEmpty = s: (builtins.match "[ \t\n]*" s) == null;
   nonEmptyStr = addCheckDesc "non-empty" types.str isNonEmpty;
 
-  fileSystems' = toposort fsBefore (attrValues config.fileSystems);
+  enabledFileSystems = lib.filterAttrs (n: v: v.enable) config.fileSystems;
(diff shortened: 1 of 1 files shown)


eclairevoyant:
seems there is some overlap with #301915

PaulGrandperrin:
@eclairevoyant oh indeed! Only 2 weeks old :)

PR #188278, opened 2022-08-25
tlpui: init at 1.5.0

GeorgesAlkhouri:
###### Description of changes

Added tlpui as a nix package.

> A GTK user interface for TLP written in Python 
https://github.com/d4nj1/TLPUI

###### Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
- [ ] For non-Linux: Is `sandbox = true` set in `nix.conf`? (See [Nix manual](https://nixos.org/manual/nix/stable/command-ref/conf-file.html))
- [ ] Tested, as applicable:
  - [NixOS test(s)](https://nixos.org/manual/nixos/unstable/index.html#sec-nixos-tests) (look inside [nixos/tests](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests))
  - and/or [package tests](https://nixos.org/manual/nixpkgs/unstable/#sec-package-tests)
  - or, for functions and "core" functionality, tests in [lib/tests](https://github.com/NixOS/nixpkgs/blob/master/lib/tests) or [pkgs/test](https://github.com/NixOS/nixpkgs/blob/master/pkgs/test)
  - made sure NixOS tests are [linked](https://nixos.org/manual/nixpkgs/unstable/#ssec-nixos-tests-linking) to the relevant packages
- [ ] Tested compilation of all packages that depend on this change using `nix-shell -p nixpkgs-review --run "nixpkgs-review rev HEAD"`. Note: all changes have to be committed, also see [nixpkgs-review usage](https://github.com/Mic92/nixpkgs-review#usage)
- [ ] Tested basic functionality of all binary files (usually in `./result/bin/`)
- [22.11 Release Notes (or backporting 22.05 Release notes)](https://github.com/NixOS/nixpkgs/blob/master/CONTRIBUTING.md#generating-2211-release-notes)
  - [ ] (Package updates) Added a release notes entry if the change is major or breaking
  - [ ] (Module updates) Added a release notes entry if the change is significant
  - [ ] (Module addition) Added a release notes entry if adding a new NixOS module
  - [ ] (Release notes changes) Ran `nixos/doc/manual/md-to-db.sh` to update generated release notes
- [ ] Fits [CONTRIBUTING.md](https://github.com/NixOS/nixpkgs/blob/master/CONTRIBUTING.md).

PR #305278, opened 2024-04-19
tlpui: init at 1.6.5

LordGrimmauld:
## Description of changes

- ~~link the empty default config of tlp into the legacy config path of tlp to make it work even without explicit config in /etc/tlp.conf~~
- ~~move nixos tlp module to use `/etc/tlp.d/` as to not break the tlp config created imperatively with tlpui in /etc/tlp.conf~~
- init tlpui at 1.6.5

This PR reuses work by @GeorgesAlkhouri in #188278, but with an updated nix package to use pyproj and an updated patchset to use the tlp default config in the tlp store path if the explicit tlp config in /etc/tlp.conf does not yet exist.

I tested this PR on my local system and it works well as far as i can tell.

~~I understand this is a relatively big scale for tlp as a whole, not just a simple package addition, feedback is welcome.~~

Edit: i adjusted this pull request. It no longer needs to modify the tlp nixos module.
Instead, the tlp nixos module is now completely untouched and still applies its config in the global config location of /etc/tlp.conf. tlpui places its config files in /etc/tlp.d/30-tlpui.conf instead.
This approach has a couple interesting quirks but is an overall improvement:

-   currently, it outright crashes if /etc/tlp.d does not yet exist as a directory.
-   the changes made in tlpui do persist even after uninstalling tlpui
-   nix settings for tlp takes precedence over tlpui, which means at least the end result can still be fully determined by nix even with left-over config files.
    Ideally, I'd write a nixos module for tlpui instead, with tlpui writing to a special location and the nixos module symlinking the config result into place ensuring both tlp.d exists and is being properly cleaned up on uninstall.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/by-name/tl/tlpui/package.nix
+++ b/pkgs/by-name/tl/tlpui/package.nix
@@ -1,0 +1,70 @@
+{
+  cairo,
+  fetchFromGitHub,
+  gobject-introspection,
+  gtk3,
+  lib,
+  pciutils,
+  python3Packages,
+  substituteAll,
+  tlp,
+  usbutils,
+  wrapGAppsHook,
+}:
+python3Packages.buildPythonPackage rec {
+  pname = "tlpui";
+  version = "1.6.5";
+  pyproject = true;
+
+  src = fetchFromGitHub {
+    owner = "d4nj1";
+    repo = "TLPUI";
+    rev = "refs/tags/tlpui-${version}";
+    hash = "sha256-pgzGhf2WDRNQ2z0hPapUJA5MLTKq92UlgjC+G78T/4s=";
+  };
+
+  patches = [
+    (substituteAll {
+      src = ./path.patch;
+      inherit tlp;
+    })
+  ];
+
+  # ignore test/test_tlp_settings.py asit relies on opening a gui which is non-trivial
+  pytestFlagsArray = [ "--ignore=test/test_tlp_settings.py" ];
+  nativeCheckInputs = [
+    gobject-introspection
+    python3Packages.pytestCheckHook
+  ];
+
+  build-system = [
+    wrapGAppsHook
+    python3Packages.poetry-core
+  ];
+
+  buildInputs = [ tlp ];
+
+  dependencies = [
+    gobject-introspection
+    gtk3
+    pciutils
+    python3Packages.pycairo
+    python3Packages.pygobject3
+    python3Packages.pyyaml
+    usbutils
+  ];
+
+  meta = {
+    changelog = "https://github.com/d4nj1/TLPUI/releases/tag/tlpui-${version}";
+    description = "A GTK user interface for TLP written in Python";
+    homepage = "https://github.com/d4nj1/TLPUI";
+    license = lib.licenses.gpl2Only;
+    longDescription = ''
+      The Python scripts in this project generate a GTK-UI to change TLP configuration files easily.
+      It has the aim to protect users from setting bad configuration and to deliver a basic overview of all the valid configuration values.
+    '';
+    platforms = lib.platforms.linux;
+    mainProgram = "tlpui";
+    maintainers = with lib.maintainers; [ grimmauld ];
+  };
+}
--- a/pkgs/by-name/tl/tlpui/path.patch
+++ b/pkgs/by-name/tl/tlpui/path.patch
@@ -1,0 +1,57 @@
+diff --git a/tlpui/file.py b/tlpui/file.py
+index f0f3ecb..a9ad7f8 100644
+--- a/tlpui/file.py
++++ b/tlpui/file.py
+@@ -26,7 +26,7 @@ def get_tlp_config_defaults(tlpversion: str):
+     tlpconfig_defaults = extract_default_tlp_configs(f"{settings.workdir}/defaults/tlp-{tlpversion}.conf")
+ 
+     # update default values with intrinsic ones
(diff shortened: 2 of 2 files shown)


pkgs/by-name/tl/tlpui/package.nix
```diff
@@ -0,0 +1,69 @@
...
+  dependencies = [
+    gobject-introspection
+    gtk3
+    pciutils
+    python3Packages.pycairo
+    python3Packages.pygobject3
+    python3Packages.pyyaml
+    usbutils
+  ];
+
```
Aleksanaa, line 55:
```suggestion
  dontWrapGApps = true;
  
  preFixup = ''
    makeWrapperArgs+=("''${gappsWrapperArgs[@]}")
  '';
```

To avoid double wrapping.

pkgs/by-name/tl/tlpui/package.nix
```diff
@@ -0,0 +1,69 @@
...
+, gtk3
+, lib
+, pciutils
+, python3Packages
+, substituteAll
+, tlp
+, usbutils
+, wrapGAppsHook
+}:
+python3Packages.buildPythonPackage rec {
```
Aleksanaa, line 12:
```suggestion
python3Packages.buildPythonApplication rec {
```

pkgs/by-name/tl/tlpui/package.nix
LordGrimmauld: Interesting. I addessed this, but is there a good way to tell how often things are being wrapped to prevent double wraps?

pkgs/by-name/tl/tlpui/package.nix
LordGrimmauld: Oops, that was a goof, good catch

pkgs/by-name/tl/tlpui/package.nix
Aleksanaa: Oh, I mean, you can still tell by e.g. `ls -a ./result/bin`

LordGrimmauld:
TODO: I do need to do more testing in terms of how tlpui reads default values. At a quick glance everything seemed fine, but i can't quite tell. I need to especially crossreference with a tlpui install on a known-good distro.

GeorgesAlkhouri (approved):

LordGrimmauld:
Update:
I ran this ever since implementing without issues. It is not possible to reset to "clean" defaults however. Config fields that were at any point defined in tlpui will always be defined in tlpui. Further, uninstalling tlpui does not remove the config fields defined by it. This might in itself be an issue in regards to nix inherent concept to only have things affect the system that are also listed in the config.
Likely the cleaner option would be to have nix still define the config in `/etc/tlp.conf` and somehow figure out how to coerce tlpui to place its configs in `/etc/tlp.d`. This would still mean uninstalling tlpui would have the tlpui config file remain on the system, but the nix defined config would take priority.

If anyone familiar enough with tlpui to know how we can set it to use a config in /etc/tlp.d/* ? One option would be to patch the path lookup in https://github.com/d4nj1/TLPUI/blob/master/tlpui/settingshelper.py#L23 but that might not be the best of ideas

Aleksanaa:
> This might in itself be an issue in regards to nix inherent concept to only have things affect the system that are also listed in the config.

Yes, it's intended. Sorry I cannot test it for now.

LordGrimmauld:
All good, take your time. These are chaotic and turbulent times and this is not a pressing issue. I can run this PR just fine until we have a better method and a more stable situation overall....

LordGrimmauld:
Alright, i adjusted this pull request. It no longer needs to modify the tlp nixos module.
Instead, the tlp nixos module is now completely untouched and still applies its config in the global config location of `/etc/tlp.conf`. tlpui places its config files in `/etc/tlp.d/30-tlpui.conf` instead.
This approach has a couple interesting quirks but is an overall improvement:
- currently, it outright crashes if `/etc/tlp.d` does not yet exist as a directory.
- the changes made in tlpui do persist even after uninstalling tlpui
- nix settings for tlp takes precedence over tlpui, which means at least the end result can still be fully determined by nix even with left-over config files.
Ideally, I'd write a nixos module for tlpui instead, with tlpui writing to a special location and the nixos module symlinking the config result into place ensuring both tlp.d exists and is being properly cleaned up on uninstall.

pkgs/by-name/tl/tlpui/package.nix
```diff
@@ -0,0 +1,70 @@
...
+  fetchFromGitHub,
+  gobject-introspection,
+  gtk3,
+  lib,
+  pciutils,
+  python3Packages,
+  substituteAll,
+  tlp,
+  usbutils,
+  wrapGAppsHook,
```
Aleksanaa, line 12: It has been renamed to `wrapGAppsHook3` to avoid confusion.

Aleksanaa (commented):
Also should avoid double wrapping (See https://github.com/NixOS/nixpkgs/pull/303848)

Aleksanaa:
> currently, it outright crashes if /etc/tlp.d does not yet exist as a directory.

This is logically correct, but it is better to mention it in `longDescription`.

> the changes made in tlpui do persist even after uninstalling tlpui

This is also acceptable.

By the way, I'd recommend always adding new module after package, in two separate PRs, to improve the efficiency of reviewing.

PR #305286, opened 2024-04-19
nixos/devpi-server: init

cafkafk:
## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [x] Tested, as applicable:

---

### DIFF
--- a/nixos/modules/services/misc/devpi-server.nix
+++ b/nixos/modules/services/misc/devpi-server.nix
@@ -1,0 +1,128 @@
+{
+  pkgs,
+  lib,
+  config,
+  ...
+}:
+with lib;
+let
+  cfg = config.services.devpi-server;
+
+  secretsFileName = "devpi-secret-file";
+
+  stateDirName = "devpi";
+
+  runtimeDir = "/run/${stateDirName}";
+  serverDir = "/var/lib/${stateDirName}";
+in
+{
+  options.services.devpi-server = {
+    enable = mkEnableOption "Devpi Server";
+
+    package = mkPackageOption pkgs "devpi-server" { };
+
+    primaryUrl = mkOption {
+      type = types.str;
+      description = "Url for the primary node. Required option for replica nodes.";
+    };
+
+    replica = mkOption {
+      type = types.bool;
+      default = false;
+      description = ''
+        Run node as a replica.
+        Requires the secretFile option and the primaryUrl to be enabled.
+      '';
+    };
+
+    secretFile = mkOption {
+      type = types.nullOr types.path;
+      default = null;
+      description = ''
+        Path to a shared secret file used for synchronization,
+        Required for all nodes in a replica/primary setup.
+      '';
+    };
+
+    host = mkOption {
+      type = types.str;
+      default = "localhost";
+      description = ''
+        domain/ip address to listen on
+      '';
+    };
+
+    port = mkOption {
+      type = types.port;
+      default = 3141;
+      description = "The port on which Devpi Server will listen.";
+    };
+
+    openFirewall = mkEnableOption "opening the default ports in the firewall for Devpi Server";
+  };
+
+  config = mkIf cfg.enable {
+
+    systemd.services.devpi-server = {
+      enable = true;
+      description = "devpi PyPI-compatible server";
+      documentation = [ "https://devpi.net/docs/devpi/devpi/stable/+d/index.html" ];
+      wants = [ "network-online.target" ];
+      wantedBy = [ "multi-user.target" ];
+      after = [ "network-online.target" ];
+      # Since at least devpi-server 6.10.0, devpi requires the secrets file to
+      # have 0600 permissions.
+      preStart =
+        ''
+          cp ${cfg.secretFile} ${runtimeDir}/${secretsFileName}
+          chmod 0600 ${runtimeDir}/*${secretsFileName}
+
+          if [ -f ${serverDir}/.nodeinfo ]; then
+            # already initialized the package index, exit gracefully
+            exit 0
+          fi
+          ${cfg.package}/bin/devpi-init --serverdir ${serverDir} ''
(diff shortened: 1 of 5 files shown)


nixos/modules/services/misc/devpi-server.nix
```diff
@@ -0,0 +1,125 @@
...
+  ...
+}:
+with lib;
+let
+  cfg = config.services.devpi-server;
+  serverDir = "/var/lib/devpi";
+in
+{
+  options.services.devpi-server = {
+    enable = mkEnableOption (mdDoc "Devpi Server");
```
anthonyroussel, line 14:
```suggestion
    enable = mkEnableOption "Devpi Server";
```

mdDoc expresions can be removed

nixos/modules/services/misc/devpi-server.nix
```diff
@@ -0,0 +1,125 @@
...
+      type = types.bool;
+      default = false;
+      description = ''
+        Run node as a replica.
+        Requires the secretFile option and the primaryUrl to be enabled.
+      '';
+    };
+
+    secretFile = mkOption {
+      type = types.str;
```
anthonyroussel, line 38:
```suggestion
      type = types.nullOr types.path;
```

I think type here should be null or path.

nixos/modules/services/misc/devpi-server.nix
```diff
@@ -0,0 +1,125 @@
...
+                "--port=${builtins.toString cfg.port}"
+              ]
+              ++ (
+                if cfg.secretFile != "" then
+                  [
+                    "--secretfile=\${CREDENTIALS_DIRECTORY}/devpi-secret-file"
+                  ]
+                else
+                  [ ]
+              )
```
anthonyroussel, line 101:
```suggestion
              ++ lib.optionals (!cfg.secretFile) [
                "--secretfile=\${CREDENTIALS_DIRECTORY}/devpi-secret-file"
              ]
```

This can be simplified with lib.optionals.
What do you think?

nixos/tests/devpi-server.nix
```diff
@@ -0,0 +1,35 @@
+import ./make-test-python.nix ({ pkgs, ...}:
```
anthonyroussel, line 1: Don't forget to also bind your NixOS test to the devpi-server package:

```nix
  passthru.tests.devpi-server = nixosTests.devpi-server;
```

in pkgs/development/tools/devpi-server/default.nix

nixos/modules/services/misc/devpi-server.nix
```diff
@@ -0,0 +1,125 @@
...
+      type = types.port;
+      default = 3141;
+      description = "The port on which Devpi Server will listen.";
+    };
+
+    openFirewall = mkOption {
+      type = types.bool;
+      default = false;
+      description = "Opens the port used by the firewall.";
+    };
```
anthonyroussel, line 64:
```suggestion
    openFirewall = mkEnableOption "opening the default ports in the firewall for Devpi Server";
```

You can also use mkEnableOption here :)

nixos/modules/services/misc/devpi-server.nix
```diff
@@ -0,0 +1,125 @@
...
+  };
+
+  config = mkIf cfg.enable {
+
+    systemd.services.devpi-server = {
```
anthonyroussel, line 69:
```suggestion
    systemd.services.devpi-server = {
      description = "devpi PyPI-compatible server";
      documentation = [ "https://devpi.net/docs/devpi/devpi/stable/+d/index.html" ];
```

We can also improve the SystemD service with some description, documentation.
And also some SystemD hardenings like the Debian team has done.

See https://salsa.debian.org/python-team/packages/devpi-server/-/blob/debian/master/debian/devpi-server.service

What do you think?

nixos/modules/services/misc/devpi-server.nix
```diff
@@ -0,0 +1,125 @@
...
+{
+  options.services.devpi-server = {
+    enable = mkEnableOption (mdDoc "Devpi Server");
+
+    package = mkOption {
+      type = with types; package;
+      default = pkgs.devpi-server;
+      defaultText = lib.literalExpression "pkgs.devpi-server";
+      description = "Which devpi-server package to use.";
+    };
```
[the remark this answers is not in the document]
cafkafk: Didn't know about this, thanks, will use it instead!

nixos/modules/services/misc/devpi-server.nix
cafkafk: This makes a lot more sense this way, I prefer this!

nixos/modules/services/misc/devpi-server.nix
cafkafk: Good idea, I'll take most of these, except `ReadWritePaths` since I think that might cause issues.

nixos/modules/services/misc/devpi-server.nix
cafkafk: I've refactored it to use null or path, but I'm unsure if this might mean that secrets will have to be stored in the nix store, which isn't ideal...

cafkafk:
Thanks for the review, I learned about a lot of things I didn't know existed :D

srhb (changes requested):
Mostly looks great, however:

I can't actually make the `secretFile` option work. It appears devpi has opinions:tm: on the mode of the file -- it must be 0600, and the `LoadCredential` creates the file with group access too.

I think we will have to copy it to a runtime directory or something and ensure the mode is correct, and perhaps set the `secretFile` in the nixos test as well to surface issues here.

srhb (approved):

PR #305474, opened 2024-04-20
mmtf-cpp: add darwin support

natsukium:
## Description of changes

This PR makes mmtf-cpp support darwin.
I'd like to use it with pymol.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [x] x86_64-darwin
  - [x] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [x] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/development/libraries/science/chemistry/mmtf-cpp/default.nix
+++ b/pkgs/development/libraries/science/chemistry/mmtf-cpp/default.nix
@@ -19,7 +19,7 @@ stdenv.mkDerivation rec {
     description = "A library of exchange-correlation functionals with arbitrary-order derivatives";
     homepage = "https://github.com/rcsb/mmtf-cpp";
     license = licenses.mit;
-    platforms = platforms.linux;
+    platforms = platforms.unix;
     maintainers = [ maintainers.sheepforce ];
   };
 }
--- a/pkgs/development/libraries/science/chemistry/mmtf-cpp/default.nix
+++ b/pkgs/development/libraries/science/chemistry/mmtf-cpp/default.nix
@@ -1,13 +1,13 @@
 { stdenv, lib, fetchFromGitHub, cmake, msgpack } :
 
-stdenv.mkDerivation rec {
+stdenv.mkDerivation (finalAttrs: {
   pname = "mmtf-cpp";
   version = "1.1.0";
 
   src = fetchFromGitHub  {
     owner = "rcsb";
-    repo = pname;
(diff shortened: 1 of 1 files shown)


pkgs/development/libraries/science/chemistry/mmtf-cpp/default.nix
```diff
@@ -19,7 +19,7 @@ stdenv.mkDerivation rec {
     description = "A library of exchange-correlation functionals with arbitrary-order derivatives";
     homepage = "https://github.com/rcsb/mmtf-cpp";
     license = licenses.mit;
-    platforms = platforms.linux;
+    platforms = platforms.unix;
```
kirillrdy, line 22: should this be `with lib.platforms; linux ++ darwin;` ?

Aleksanaa: We have no reason why it can't work on other BSDs?

kirillrdy: I am pro BSD support, it's just there is no evidence that it works on BSD's.

Maybe to be more specific, it was obvious to me that we want all platforms but it may or may not work on platforms other than what CI builds. In those cases do we want to guess of the best or list what we know for sure ?

Aleksanaa: On those platforms with lower support tiers, we just normally assume it works. If someone found any problems, they can later fix that. If we don't, the maintainers of these platforms need to add tons of attrs to make them work (which will very likely be rejected by Nixpkgs).

kirillrdy: thank you, that's was pretty much my question. is it documented anywhere ?

natsukium: Since the package has no platform-specific features and is built with platform-independent tools, I expect it to work well on other UNIX systems.
If it doesn't work, we shouldn't care.

I don't know if it is documented or if there is a common consensus among the committers, but the tier of the systems can be found here.
https://github.com/NixOS/rfcs/blob/master/rfcs/0046-platform-support-tiers.md

sheepforce (approved):
Fine for me, thanks! :)

kirillrdy (approved):

PR #303173, opened 2024-04-10
crystal: 1.11 -> 1.14

will:
## Current Description of changes

Release notes: https://crystal-lang.org/2024/10/09/1.14.0-released/

## Old Description of changes

Release notes: https://crystal-lang.org/2024/04/09/1.12.0-released

From the notes: 

> Our own LLVM extension library libllvm_ext is no longer required when linking LLVM >= 18 because all APIs are now directly exposed in libllvm directly. The Makefile automatically skips building libllvm_ext.o on LLVM 18 and above.

So I've updated the llvm package for this version to llvm 18. From talking with people on the Crystal core team, there is very small chance that the 16-bit alignment for 128-bit integers in llvm18 can cause problems on x86 systems (not aarch64 since they already were using 16-bit alignment), and if so we should go down to llvm 17.

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [x] aarch64-darwin
- For non-Linux: Is sandboxing enabled in `nix.conf`? (See [Nix manual](https://nixos.org/manual/nix/stable/command-ref/conf-file.html))
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:
  - [NixOS test(s)](https://nixos.org/manual/nixos/unstable/index.html#sec-nixos-tests) (look inside [nixos/tests](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests))
  - and/or [package tests](https://github.com/NixOS/nixpkgs/blob/master/pkgs/README.md#package-tests)
  - or, for functions and "core" functionality, tests in [lib/tests](https://github.com/NixOS/nixpkgs/blob/master/lib/tests) or [pkgs/test](https://github.com/NixOS/nixpkgs/blob/master/pkgs/test)
  - made sure NixOS tests are [linked](https://nixos.org/manual/nixpkgs/unstable/#ssec-nixos-tests-linking) to the relevant packages
- [x] Tested compilation of all packages that depend on this change using `nix-shell -p nixpkgs-review --run "nixpkgs-review rev HEAD"`. Note: all changes have to be committed, also see [nixpkgs-review usage](https://github.com/Mic92/nixpkgs-review#usage)

Result of `nixpkgs-review` run on aarch64-darwin [1](https://github.com/Mic92/nixpkgs-review)
<details>
  <summary>1 package marked as broken and skipped:</summary>
  <ul>
[...]

PR #282286, opened 2024-01-20
collision: 3.6.0 -> 3.8.0

sund3RRR:
## Description of changes
Update version and install `nautilus-python` extension.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [x] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
- For non-Linux: Is sandboxing enabled in `nix.conf`? (See [Nix manual](https://nixos.org/manual/nix/stable/command-ref/conf-file.html))
  - [ ] `sandbox = relaxed`
  - [x] `sandbox = true`
- [ ] Tested, as applicable:
  - [NixOS test(s)](https://nixos.org/manual/nixos/unstable/index.html#sec-nixos-tests) (look inside [nixos/tests](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests))
  - and/or [package tests](https://nixos.org/manual/nixpkgs/unstable/#sec-package-tests)
  - or, for functions and "core" functionality, tests in [lib/tests](https://github.com/NixOS/nixpkgs/blob/master/lib/tests) or [pkgs/test](https://github.com/NixOS/nixpkgs/blob/master/pkgs/test)
  - made sure NixOS tests are [linked](https://nixos.org/manual/nixpkgs/unstable/#ssec-nixos-tests-linking) to the relevant packages
- [x] Tested compilation of all packages that depend on this change using `nix-shell -p nixpkgs-review --run "nixpkgs-review rev HEAD"`. Note: all changes have to be committed, also see [nixpkgs-review usage](https://github.com/Mic92/nixpkgs-review#usage)
- [x] Tested basic functionality of all binary files (usually in `./result/bin/`)
- [24.05 Release Notes](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2405.section.md) (or backporting [23.05](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2305.section.md) and [23.11](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2311.section.md) Release notes)
  - [ ] (Package updates) Added a release notes entry if the change is major or breaking
  - [ ] (Module updates) Added a release notes entry if the change is significant
  - [ ] (Module addition) Added a release notes entry if adding a new NixOS module
[...]

PR #243914, opened 2023-07-17
gi-crystal: init at 0.17.0

sund3RRR:
###### Description of changes
GI Crystal is a binding generator used to generate Crystal bindings for GObject based libraries using [GObject Introspection](https://gi.readthedocs.io/en/latest/).
https://github.com/hugopl/gi-crystal

###### Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
- [ ] For non-Linux: Is `sandbox = true` set in `nix.conf`? (See [Nix manual](https://nixos.org/manual/nix/stable/command-ref/conf-file.html))
- [ ] Tested, as applicable:
  - [NixOS test(s)](https://nixos.org/manual/nixos/unstable/index.html#sec-nixos-tests) (look inside [nixos/tests](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests))
  - and/or [package tests](https://nixos.org/manual/nixpkgs/unstable/#sec-package-tests)
  - or, for functions and "core" functionality, tests in [lib/tests](https://github.com/NixOS/nixpkgs/blob/master/lib/tests) or [pkgs/test](https://github.com/NixOS/nixpkgs/blob/master/pkgs/test)
  - made sure NixOS tests are [linked](https://nixos.org/manual/nixpkgs/unstable/#ssec-nixos-tests-linking) to the relevant packages
- [ ] Tested compilation of all packages that depend on this change using `nix-shell -p nixpkgs-review --run "nixpkgs-review rev HEAD"`. Note: all changes have to be committed, also see [nixpkgs-review usage](https://github.com/Mic92/nixpkgs-review#usage)
- [x] Tested basic functionality of all binary files (usually in `./result/bin/`)
- [23.11 Release Notes](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2311.section.md) (or backporting [23.05 Release notes](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2305.section.md))
  - [ ] (Package updates) Added a release notes entry if the change is major or breaking
  - [ ] (Module updates) Added a release notes entry if the change is significant
  - [ ] (Module addition) Added a release notes entry if adding a new NixOS module
- [ ] Fits [CONTRIBUTING.md](https://github.com/NixOS/nixpkgs/blob/master/CONTRIBUTING.md).

PR #305484, opened 2024-04-20
gi-crystal: remove package

sund3RRR:
## Description of changes
Remove `gi-crystal` package because it is useless now as we have `copyShardDeps` flag in `buildCrystalPackage` func, so `gi-crystal` works now without any patches and packaging.
Since both packages that depend on gi-crystal are [broken](https://github.com/NixOS/nixpkgs/pull/303173#issuecomment-2049162658), we can remove `gi-crystal` now.
But it is better to wait until [collision](https://github.com/NixOS/nixpkgs/pull/282286) and [rtfm](https://github.com/NixOS/nixpkgs/pull/305481) are merged.

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [x] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---


UlyssesZh:
I think it may be better to add a `useGiCrystal` argument to `crystal.buildCrystalPackage`, and enabling that argument will
- toggle on `copyShardDeps` by default,
- add `gobject-introspection` to `nativeBuildInputs`, and
- add the following to build phase:
```shell
mkdir bin
cd lib/gi-crystal
shards build -Dpreview_mt --release --no-debug
cd ../..
cp lib/gi-crystal/bin/gi-crystal bin
```
This may reduce some common codes. The only thing that needs to be patched is then removing `shards install` commands in the Makefile.

UlyssesZh:
> Remove `gi-crystal` package because it is useless now as we have `copyShardDeps` flag in `buildCrystalPackage` func, so `gi-crystal` works now without any patches and packaging.

I don't know whether `gi-crystal` should be considered useless. A benefit of having  `gi-crystal` is to make gi-crystal a separate derivation from the package (say `collision`) that uses gi-crystal, so we don't have to compile gi-crystal every time we build `collision`.

sund3RRR:
> I think it may be better to add a `useGiCrystal` argument to `crystal.buildCrystalPackage`, and enabling that argument will
> 
> * toggle on `copyShardDeps` by default,
> * add `gobject-introspection` to `nativeBuildInputs`, and
> * add the following to build phase:
> 
> ```shell
> mkdir bin
> cd lib/gi-crystal
> shards build -Dpreview_mt --release --no-debug
> cd ../..
> cp lib/gi-crystal/bin/gi-crystal bin
> ```
> 
> This may reduce some common codes. The only thing that needs to be patched is then removing `shards install` commands in the Makefile.

These are good thoughts and I like it. But I think there's something wrong with the crystal on the Nix.
You know, like... we need to set the `shards.nix` file, then set `copyShardDeps = true`, then set `useGiCrystal = true;`, and then add a couple of patches, because nothing works with Crystal on nix.

While the "I use Arch btw" guys just do `shards install && shards build` and push their package to the AUR, we need...
1) Download Collision, try to package...
2) Realize that something is wrong, because it just won’t work
3) Take a couple of days to understand why
4) Got it! Take a couple of days to figure out how to fix this...
5) gi-crystal package with some patches!
6) Understand that this is a bad idea because every package that depends on it must have its own gi
-crystal because the version must match the version in shard.yml.
7) Take a couple of days to figure out how to fix this...
8) Make a PR to the main buildCrystalPackage function
9) Wait a couple of months until someone checks and merges the PR
10) FINALLY the collision package is correct, but you have to add this shit:
``` bash
mkdir bin
cd lib/gi-crystal
building shards -Dpreview_mt --release --no-debug
cd ../..
cp lib/gi-crystal/bin/gi-crystal bin
```
because you still can't run `shards install` and you need to build it manually :)))))))

I've spent so much time trying to package crystal applications because we have over-engineering for... over-engineering!

While the Arch Linux guys have a human environment on their system, we create a declarative shards.nix file on top of the declarative shard.yml and shard.lock LOCK FILE. We have a lock file, why do we need another lock????

I feel like a gun without bullets. I have the dependencies, but I can't use them because gi-crystal can't run from /nix/store, and blake3.cr doesn't build because Crystal2nix doesn't run scripts after installation.

While people use the default human-like environment, we can't use `shards install` because it is less reproducible than nix.

You know, it gets really crazy when we talk about reproducibility. We store deps into /nix/store, but we need the `copyShardDeps` flag to restore the normal project structure, because without it it doesn't work. Why do we use this shit if we still need to restore the project structure and copy it? Why don't we just use `shards install` during the download phase????

Because we don't like non-reproducible builds. But do we love so many hacks and patches? Do we really need all this over-engineering? It's not normal for you to need so much time just to run a program.

Going back to the `useGiCrystal` flag, I think it's better to create another builder called `buildGiCrystalApplication` or something like that.
Like python with `buildPythonPackage` and `buildPythonApplication`.
Add a `preBuild` hook to create and run gi-crystal.

But I have to do so much work for this, so... I don't know, it would be great if we just run `shards install`...

UlyssesZh:
This situation with Crystal packages is actually the result of multiple factors, not just nix. The `lib` dir shouldn't need write access by design; shards should not manage build time dependencies by design, either.

I am actually surprised that, when you packaged `gi-crystal` in the first place (#243914), you didn't propose the changes to the upstream (or maybe you did?). Making gi-crystal a standalone build tool instead of a dependency managed by shards is so much better.

Having `buildGiCrystalApplication` is fine. I just thought adding a flag to `buildCrystalPackage` would be less work.

sund3RRR:
I think having gi-crystal in nixpkgs is a bad idea because this thing not backwards compatible. Means gi-crystal 0.23.0 may not work if the package expects gi-crystal 0.22.0.

Another reason is that it is now a patched gi-crystal with changed logic. We have a problem: we need to support our patches, which Hugo won't support.

If Hugo decides to generate bindings outside of the lib/ folder, we'll be fine with it. But we shouldn't apply these hard patches, our goal in my opinion is just to pack it.

I don't want to add another flag to `buildCrystalPackage` just because it will take a long time waiting for someone to review and merge it. I want to fix all packages related to this issue as quickly as possible because the release of NixOS 24.05 Uakari is approaching.

UlyssesZh:
> I don't want to add another flag to `buildCrystalPackage` just because it will take a long time waiting for someone to review and merge it. I want to fix all packages related to this issue as quickly as possible because the release of NixOS 24.05 Uakari is approaching.

I think it should be fine to just do it in this PR.

sund3RRR:
@ofborg eval

sund3RRR:
@GrahamcOfBorg eval

UlyssesZh:
Why close?

sund3RRR:
Cause lusasew made a [cool](https://github.com/NixOS/nixpkgs/pull/312290/commits/b4c133a87402c57471faf31c693ba08a983077d1) update script for that package. I don't want to make another flag or builder for buildCrystalPackage func. But maybe someone wants to use gi-crystal as a nix package for development. So I decided to leave this package in nixpkgs.

PR #305509, opened 2024-04-20
pcsx2: 1.7.5587 -> 1.7.5779

SuperSamus:
## Description of changes

Diff: https://github.com/PCSX2/pcsx2/compare/v1.7.5587...v1.7.5779

~Depends on #305508 (which is the base of this branch).~ Merged.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---


SuperSamus:
Waiting to know whether it's better to patch `shaderc` inside the package or in `all-packages`. (#309064)

hrdinka:
@SuperSamus thanks a lot for continuously updating this package! As always I am very happy about your contribution.

Regarding the `shaderc` patch: IMHO it should be overriden in `all-packages.nix`. There are many other packages doing it this way as well. The `pcsx2` derivation should be agnostic of its input and not try to override them, it cannot know what exact derivations are passed as input. It should basically just ask for a `shaderc` package and be happy with whatever it gets. The caller should make sure to pass the right one. If a package needs a special version of an input, it is always done via `callPackage` in `all-packages.nix` and this here is not different.

Further a `fetch*` variant should be preferred and used whenever possible like in the duckstation PR to not bloat nixpkgs unnecessarily.

I would however not be happy about moving it to the bottom in `all-packages.nix` like it was proposed in the duckstation PR. I appreciate its good intentions but in the end this will mess up the ordering in `all-packages.nix` even more since I expect only few people to notice and adhere to it. In the light of RCF140, eventually all default package calls will be removed from `all-packages.nix` anyway.

pkgs/by-name/pc/pcsx2/package.nix
```diff
@@ -1,65 +1,67 @@
...
+    rev = "b3a788e16ea12efac006cbbe1ece45b6b9b34326";
+    sha256 = "sha256-Uvpz2Gpj533Sr6wLruubZxssoXefQDey8GHIDKWhW3s=";
   };
+  inherit (qt6Packages)
+    qtbase
+    qtsvg
+    qttools
+    qtwayland
+    wrapQtAppsHook
+    ;
```
SuperSandro2000, line 45: qt6Packages.callPackages should be used instead of inheriting here, like it was done before.

SuperSandro2000 (dismissed):

pkgs/by-name/pc/pcsx2/package.nix
AndersonTorres: Why?

AndersonTorres:
```
Run if result/bin/nixpkgs-check-by-name --base "$base" .; then
- Because pkgs/by-name/pc/pcsx2 exists, the attribute `pkgs.pcsx2` must be defined like

    pcsx2 = callPackage ./../by-name/pc/pcsx2/package.nix { /* ... */ };

  However, in this PR, a different `callPackage` is used. See the definition in pkgs/top-level/all-packages.nix:2752:

    pcsx2 = qt6Packages.callPackage ../by-name/pc/pcsx2/package.nix {
      shaderc = shaderc.overrideAttrs (old: {
        patches = (old.patches or [ ]) ++ [
          (fetchpatch {
            url = "file://${pcsx2.src}/.github/workflows/scripts/common/shaderc-changes.patch";
            hash = "sha256-cD/gZwo7ARoePbEsHGHV34sx6MomfNOM7BSZDH1NlZM=";
            excludes = [
              "CHANGES"
              "CMakeLists.txt"
              "libshaderc/CMakeLists.txt"
            ];
          })
        ];
      });
    };
```

pkgs/top-level/all-packages.nix
```diff
@@ -2755,7 +2755,21 @@ with pkgs;
 
   pcem = callPackage ../applications/emulators/pcem { };
 
-  pcsx2 = qt6Packages.callPackage ../applications/emulators/pcsx2 { };
+  pcsx2 = callPackage ../by-name/pc/pcsx2/package.nix {
+    shaderc = shaderc.overrideAttrs (old: {
+      patches = (old.patches or [ ]) ++ [
+        (fetchpatch {
+          url = "file://${pcsx2.src}/.github/workflows/scripts/common/shaderc-changes.patch";
+          hash = "sha256-cD/gZwo7ARoePbEsHGHV34sx6MomfNOM7BSZDH1NlZM=";
```
hrdinka, line 2763: Hash should be `sha256-Ps/D+CdSbjVWg3ZGOEcgbpQbCNkI5Nuizm4E5qiM9Wo=`

hrdinka (changes requested):

pkgs/by-name/pc/pcsx2/package.nix
SuperSandro2000: Same reason as in the other PR. pkgs-by-name is not meant to replace existing wrappers with more copy paste code and for those it is totally fine to keep the current way.

pkgs/by-name/pc/pcsx2/package.nix
```diff
@@ -107,12 +108,9 @@ llvmPackages_17.stdenv.mkDerivation rec {
...
     let
-      libs = lib.makeLibraryPath ([
-        vulkan-loader
-      ] ++ cubeb.passthru.backendLibs);
-    in [
-      "--prefix LD_LIBRARY_PATH : ${libs}"
-    ];
+      libs = lib.makeLibraryPath ([ vulkan-loader ] ++ cubeb.passthru.backendLibs);
+    in
+    [ "--prefix LD_LIBRARY_PATH : ${libs}" ];
```
SuperSandro2000, line 113:
```suggestion
  qtWrapperArgs = [
    "--prefix LD_LIBRARY_PATH : ${lib.makeLibraryPath ([ vulkan-loader ] ++ cubeb.passthru.backendLibs)}"
  ];
```

pkgs/top-level/all-packages.nix
```diff
@@ -2755,7 +2755,21 @@ with pkgs;
...
+          hash = "sha256-Ps/D+CdSbjVWg3ZGOEcgbpQbCNkI5Nuizm4E5qiM9Wo=";
+          excludes = [
+            "CHANGES"
+            "CMakeLists.txt"
+            "libshaderc/CMakeLists.txt"
+          ];
+        })
+      ];
+    });
+  };
```
SuperSandro2000, line 2772: IMO we should move that inside the package. Otherwise it is easily missed and forgotten.

hrdinka: I agree with you but I don’t know ho to do this better? What is your solution to this?

When a custom dependency is needed, it should be passed just like this in `all-packages.nix`. This is how it is done for any derivation that needs different inputs. The derivation itself just uses whatever it transparently gets handed to. Of course here is more inline code due the patching, but there are also other packages doing it like this.

Patching an transparently passed dependency within the derivation (as initially proposed) is not a good solution (see my previous comment).

We could add a new non-existing input to the package with a default value (e.g. `shaderc-patched`). This way it is in pcsx2 package file and it could still be overridden by users. However if a new derivation is added named the same it will break. Further this breaks with the convention of how it is done with every other derivation.

So overall this seems to be the cleanest and best approach to me but I would be happy about suggestions.

hrdinka:
@SuperSamus thanks a lot for your work so far! Currently merging is blocked since there still are change requests that need to be addressed.

SuperSamus:
I'm waiting on what they decide to do on  #309064 (which has the same exact problem).

hrdinka:
Ok, all clear

pkgs/top-level/all-packages.nix
SuperSandro2000: > When a custom dependency is needed, it should be passed just like this in `all-packages.nix`. This is how it is done for any derivation that needs different inputs. The derivation itself just uses whatever it transparently gets handed to. Of course here is more inline code due the patching, but there are also other packages doing it like this.

It can also be placed in a let in.

hrdinka: Yes, but the problem is that the possibility of overriding `shaderc` is either lost or it does unexpected stuff with your own supplied `shaderc` package.

Initially (in this PR) the derivation got `shaderc` passed via an argument and patched it within the derivation. By doing so `shaderc` was theoretically overridable. It would however mess with the package supplied by the user and could fail in a patch conflict. Thus rendering this a bad solution because it does things where they should not be done leading to unexpected behavior.

In the duckstation PR `shaderc` is not passed as an argument (only indirectly via `callPackage`). So it is not overridable as usually, which is not great either.

The current solution makes the derivation behave as it should, without any unexpected behavior or caveats. Only problem is the patch that could easily be overlooked. So I suggest to fix this exact problem by simply putting a comment besides the `shaderc` argument that reminds about this fact.

AndersonTorres: My rationale for duckstation was that the Duckstation developers are treating shaderc as a vendored dependency.
The difference is that they are not providing the whole Git subproject themselves, but a patch.

Then we should basically create a new package, `shaderc-specifically-patched-for-duckstation`.
But since this package is not meant to be useful outside duckstation, it is not exposed to top-level.

SuperSamus:
Rebased, done what was done with DuckStation, and backported upstream patches to fix startup crashes

(Current master requires some stuff that nixpkgs doesn't have yet, like an updated SDL2 and Shaderc. One step at a time, this PR has been open long enough already.)

pkgs/by-name/pc/pcsx2/package.nix
```diff
@@ -133,8 +144,8 @@ llvmPackages_17.stdenv.mkDerivation rec {
     '';
     homepage = "https://pcsx2.net";
     license = with licenses; [ gpl3 lgpl3 ];
```
AndersonTorres, line 146:
```suggestion
    license = with licenses; [ gpl3Plus lgpl3Plus ];
```

pkgs/by-name/pc/pcsx2/package.nix
```diff
@@ -2,64 +2,76 @@
...
+    })
+    (fetchpatch {
+      url = "https://github.com/PCSX2/pcsx2/commit/92b707db994f821bccc35d6eef67727ea3ab496b.patch";
+      sha256 = "sha256-HWJ8KZAY/qBBotAJerZg6zi5QUHuTD51zKH1rAtZ3tc=";
+    })
   ];
 
   cmakeFlags = [
     "-DDISABLE_ADVANCE_SIMD=ON"
     "-DUSE_LINKED_FFMPEG=ON"
```
AndersonTorres, line 73:
```suggestion
    (lib.cmakeBool "DISABLE_ADVANCE_SIMD"  true)
    (lib.cmakeBool "USE_LINKED_FFMPEG"  true)
```

pkgs/by-name/pc/pcsx2/package.nix
```diff
@@ -2,64 +2,76 @@
...
-    "-DPCSX2_GIT_REV=v${version}"
+    "-DPCSX2_GIT_REV=v${finalAttrs.version}"
```
AndersonTorres, line 74:
```suggestion
    (lib.cmakeFeature "PCSX2_GIT_REV" finalAttrs.src.rev)
```

AndersonTorres (commented):
Some minor cleanup.

pkgs/by-name/pc/pcsx2/package.nix
```diff
@@ -2,64 +2,76 @@
...
     fetchSubmodules = true;
-    rev = "v${version}";
-    sha256 = "sha256-PCZ1r6x28Z5FEVMXWm4oxpTknz/XEiwo0rRGhn4B33g=";
+    rev = "v${finalAttrs.version}";
+    sha256 = "sha256-WiwnP5yoBy8bRLUPuCZ7z4nhIzrY8P29KS5ZjErM/A4=";
   };
 
   patches = [
     ./define-rev.patch
+    (fetchpatch {
```
AndersonTorres, line 61: Oh man, I did not see these!

Please comment why each patch is needed here.

AndersonTorres (approved):
Besides, LGTM

AndersonTorres (approved):

AndersonTorres:
P.S.: we should have a team for emulators.

SuperSamus:
Anything left to do before this can be merged?

pkgs/by-name/pc/pcsx2/package.nix
```diff
@@ -132,9 +144,9 @@ llvmPackages_17.stdenv.mkDerivation rec {
       PC, with many additional features and benefits.
     '';
     homepage = "https://pcsx2.net";
-    license = with licenses; [ gpl3 lgpl3 ];
+    license = with licenses; [ gpl3Plus lgpl3Plus ];
     maintainers = with maintainers; [ hrdinka govanify ];
     mainProgram = "pcsx2-qt";
     platforms = platforms.x86_64;
```
AndersonTorres, line 150: Just per curiousity,  what about setting this as Linux?

hrdinka (approved):

hrdinka:
Tested locally. Thanks to everyone and thanks @SuperSamus for maintaining pcsx2 🙂

I had to dismiss the outdated review which blocked the merge.

PR #305555, opened 2024-04-20
Add section about nixpkgs-merge-bot

Scriptkiddi:


FliegendeWurst (approved):

infinisil (approved):
We could debate whether this is exactly the right place for these docs, but it doesn't really matter, because having this is better than not having this in any case, so let's go :)

PR #305619, opened 2024-04-20
pythonCatchConflictsHook: prevent exponential worst-case

### LINKED ISSUE #305599: `pythonCatchConflictsHook` has exponential time complexity
### Describe the bug

Since #287957, `pythonCatchConflictsHook` traverses all paths in the DAG<sup>[1]</sup> formed by `propagatedBuildInputs` using depth first search. This is fine if the graph is approximately a tree, but in the worst case with a highly connected graph visiting all paths has exponential time complexity.

Most packages in nixpkgs have tree-like dependency graphs, but in my [nix-ros-overlay](https://github.com/lopsided98/nix-ros-overlay) project, packages tends to have much more complicated dependency graphs. This is because this project attempts to automatically build thousands of packages that are designed to run on traditional Linux distributions where dependencies are implicitly propagated. Therefore, the only way to make this feasible is to make heavy use of `propagatedBuildInputs`.

The result is that packages that used to take only a few seconds to build now take spend several hours running `pythonCatchConflictsHook`.

<sup>[1]</sup>Not technically acyclic, but we don't follow loops

### Steps To Reproduce
1. `nix build github:lopsided98/nix-ros-overlay#iron.urdfdom-py --override-input nixpkgs github:NixOS/nixpkgs/4cba8b53da471aea2ab2b0c1f30a81e7c451f4b6`
2. Observe that the build spends a very long time running `pythonCatchConflictsHook`

### Possible solutions

The simplest way to fix this is to keep track of visited paths, and only search each path once. This means that only one path to each package will be traversed, rather than all of the possible paths. This doesn't affect conflict detection, but will make the error output only print one out of possibly multiple dependency chains that lead to the conflicting package. If this is an acceptable solution, I have an implementation that I can submit as a PR.

If printing all the dependency chains is a hard requirement, a more complex solution could involve constructing a reversed graph in memory with edges from child->parent, then using that graph to find all possible dependency chains in the case that a conflict is actually found.

### Notify maintainers

@DavHau @phaer

---

lopsided98:
## Description of changes

The hook performs a depth first search on the graph defined by `propagatedBuildInputs`. This traverses all paths through the graph, except for any cycles. In the worst case with a highly connected graph, this search can take exponential time. In practice, this means that in cases with long dependency chains and multiple packages depending on the same package, the hook can take several hours to run.

Avoid this problem by keeping track of already visited paths and only visiting each path once. This makes the search complete in linear time.

The visible effect of this change is that, if a conflict is found, only one dependency chain that leads to the conflicting package is printed, rather than all the possible dependency chains.

Now that only one dependency chain will be found for each package, a few other simplifications are possible and are included in a separate commit.

Also, add a test to demonstrate the change in output with multiple dependency chains, and fix a minor bug in parsing `nix-support/propagated-build-inputs`.

The tests pass and I have tested building a few Python packages.

Fixes #305599

cc @DavHau @phaer

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [x] Tested, as applicable:

---

### DIFF
--- a/pkgs/development/interpreters/python/catch_conflicts/catch_conflicts.py
+++ b/pkgs/development/interpreters/python/catch_conflicts/catch_conflicts.py
@@ -58,7 +58,7 @@ def find_packages(store_path: Path, site_packages_path: str, parents: List[str])
     # recursively add dependencies
     if propagated_build_inputs.exists():
         with open(propagated_build_inputs, "r") as f:
-            build_inputs: List[str] = f.read().strip().split(" ")
+            build_inputs: List[str] = f.read().split()
             for build_input in build_inputs:
                 find_packages(Path(build_input), site_packages_path, parents + [build_input])
 
--- a/pkgs/development/interpreters/python/catch_conflicts/catch_conflicts.py
+++ b/pkgs/development/interpreters/python/catch_conflicts/catch_conflicts.py
@@ -5,7 +5,7 @@ import sys
 import os
 from typing import Dict, List, Set, Tuple
 do_abort: bool = False
-packages: Dict[str, Dict[str, List[Dict[str, List[str]]]]] = collections.defaultdict(list)
+packages: Dict[str, Dict[str, Dict[str, List[str]]]] = collections.defaultdict(dict)
 found_paths: Set[Path] = set()
 out_path: Path = Path(os.getenv("out"))
 version: Tuple[int, int] = sys.version_info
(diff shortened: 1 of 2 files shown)


phaer (approved):
LGTM! Maybe we should target the python-updates branch here?

lopsided98:
I updated the target branch

phaer (approved):

DavHau (approved):
LGTM
Thanks for the fix. It looks good to be merged.

I think it would be better if it would display the first 5 paths leading to a collision instead of just one. I guess this would still be in linear time and it would allow to better grasp the complexity of the underlying dependency conflict and make it easier to debug it.
But I don't want to block this. This could be done in a separate PR

Issue #273442, opened 2023-12-11
Build failure: pkgsCross.aarch64-darwin.stdenv


t0rr3sp3dr0:
### Steps To Reproduce

1. build *pkgsCross.aarch64-darwin.stdenv*

### Build log

```console
user@macos-sonoma ~ % nix-build '<nixpkgs>' -A pkgsCross.aarch64-darwin.stdenv -I nixpkgs=channel:nixpkgs-unstable --show-trace
error:
       … while calling the 'derivationStrict' builtin

         at /derivation-internal.nix:9:12:

            8|
            9|   strict = derivationStrict drvAttrs;
             |            ^
           10|

       … while evaluating derivation 'stdenv-darwin'
         whose name attribute is located at /nix/store/js53b3xz6z97jv87knhw1mdgdc4p845z-source/pkgs/stdenv/generic/default.nix:93:14

       … while evaluating attribute 'defaultNativeBuildInputs' of derivation 'stdenv-darwin'

         at /nix/store/js53b3xz6z97jv87knhw1mdgdc4p845z-source/pkgs/stdenv/generic/default.nix:126:14:

          125|
          126|       inherit initialPath shell
             |              ^
          127|         defaultNativeBuildInputs defaultBuildInputs;

       … while evaluating derivation 'aarch64-apple-darwin-clang-wrapper-16.0.6'
         whose name attribute is located at /nix/store/js53b3xz6z97jv87knhw1mdgdc4p845z-source/pkgs/stdenv/generic/make-derivation.nix:348:7

       … while evaluating attribute 'depsTargetTargetPropagated' of derivation 'aarch64-apple-darwin-clang-wrapper-16.0.6'

         at /nix/store/js53b3xz6z97jv87knhw1mdgdc4p845z-source/pkgs/stdenv/generic/make-derivation.nix:403:7:

          402|       propagatedBuildInputs       = elemAt (elemAt propagatedDependencies 1) 1;
          403|       depsTargetTargetPropagated  = elemAt (elemAt propagatedDependencies 2) 0;
             |       ^
          404|

       … while evaluating derivation 'libcxx-aarch64-apple-darwin-16.0.6'
         whose name attribute is located at /nix/store/js53b3xz6z97jv87knhw1mdgdc4p845z-source/pkgs/stdenv/generic/make-derivation.nix:348:7

       … while evaluating attribute 'stdenv' of derivation 'libcxx-aarch64-apple-darwin-16.0.6'

[...]

PR #305729, opened 2024-04-21
argc: support cross-compilation

ShamrockLee:
## Description of changes

* Move the Nix expression under name-based package directories.
* Format the Nix expression with `nixfmt` in accordance with NixOS/rfcs#166.
* Refactor the shell-completion-file-installation step and 
    * Determine if host executables can be run on the build platform via `stdenv.buildPlatform.canExecute stdenv.hostPlatform`.
    * If so, add `"${!outputBin}/bin"` into  the `PATH` used by `installShellCompletion`. (Such `PATH` value is named `INSTALL_SHELL_FILE_PATH`, so that we could generate and install man pages after argc 1.16).
    * If not, add `argcNative` into `nativeBuildInputs` and `disallowedRequisites`.
*   Add package test for cross-compilation.

## Things done

- Built on platform(s)
  - [X] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [X] Tested, as applicable:

---


ShamrockLee:
The Darwin cross-compilation test cases are affected by #273442 and fails to evaluate (infinite recursion). Here are the reason why I still add them:
- OfBorg ignores evaluation failure of `<pkg>.passthru.tests`, and shows green as long as the package itself builds.
- Potential fixes are being worked on. Darwin argc.tests should benefit from them.

Mindavi:
Think this looks ok, but would like another pair of eyes

pkgs/by-name/ar/argc/package.nix
```diff
@@ -0,0 +1,64 @@
...
+
+  passthru = {
+    tests = {
+      cross = (nixpkgsFun {
+        localSystem = stdenv.buildPlatform;
+        crossSystem =
+          (if stdenv.buildPlatform.isAarch64 then "x86_64" else "aarch64")
+          + "-"
+          + stdenv.buildPlatform.parsed.kernel.name;
+      }).argc;
```
SuperSandro2000, line 48: Why not use pkgsCross?

pkgs/top-level/all-packages.nix
```diff
@@ -4183,7 +4183,9 @@ with pkgs;
 
   aptdec = callPackage ../development/libraries/aptdec { };
 
-  argc = callPackage ../development/tools/argc { };
+  argc = callPackage ../by-name/ar/argc/package.nix {
+    nixpkgsFun = args: import ./. ({ inherit config overlays; } // args);
```
SuperSandro2000, line 4187: I don't think we should do that here. Evaluating nixpkgs is very expensive.

pkgs/by-name/ar/argc/package.nix
```diff
@@ -0,0 +1,64 @@
...
+let
+  canExecuteHost = stdenv.buildPlatform.canExecute stdenv.hostPlatform;
+in
+rustPlatform.buildRustPackage rec {
+  pname = "argc";
+  version = "1.14.0";
+
+  src = fetchFromGitHub {
+    owner = "sigoden";
+    repo = pname;
```
SuperSandro2000, line 20:
```suggestion
    repo = "argc";
```

pkgs/by-name/ar/argc/package.nix
```diff
@@ -0,0 +1,64 @@
...
+  };
+
+  cargoHash = "sha256-D1T9FWTvwKtAYoqFlR2OmLRLGWhPJ9D8J7lq/QKcBoM=";
+
+  nativeBuildInputs = [ installShellFiles ] ++ lib.optional (!canExecuteHost) argc;
+
+  postInstall = ''
+    INSTALL_SHELL_FILE_PATH="${lib.optionalString canExecuteHost ''''${!outputBin}/bin''${PATH:+:}''}''${PATH-}"
+
+    PATH="$INSTALL_SHELL_FILE_PATH" installShellCompletion --cmd argc \
```
SuperSandro2000, line 32: We don't need to indirect the bin output here, as the package only has one output.

Instead of doing complex shell logic here, we can just turn argc into a variable, put it in a let in and set it to $out/bin/argc if we are not doing cross and otherwise to the binary for the executing platform.

pkgs/by-name/ar/argc/package.nix
```diff
@@ -0,0 +1,64 @@
...
+      --bash <(argc --argc-completions bash) \
+      --fish <(argc --argc-completions fish) \
+      --zsh <(argc --argc-completions zsh)
+  '';
+
+  disallowedRequisites = lib.optional (!canExecuteHost) argc.__spliced.buildHost or argc;
```
SuperSandro2000, line 38: Why the or here? Either we don't need the value or the splicing is incorrect in which case things will not work either way.

pkgs/by-name/ar/argc/package.nix
```diff
@@ -0,0 +1,64 @@
...
+        crossSystem =
+          (if stdenv.buildPlatform.isAarch64 then "x86_64" else "aarch64")
+          + "-"
+          + stdenv.buildPlatform.parsed.kernel.name;
+      }).argc;
+    };
+  };
+
+  meta = with lib; {
+    description = "A command-line options, arguments and sub-commands parser for bash";
```
SuperSandro2000, line 53:
```suggestion
    description = "Command-line options, arguments and sub-commands parser for bash";
```

pkgs/by-name/ar/argc/package.nix
```diff
@@ -0,0 +1,64 @@
+{
+  lib,
+  rustPlatform,
+  stdenv, # Determin platforms.
+  fetchFromGitHub,
+  nixpkgsFun, # Re-evaluate Nixpkgs.
+  argc, # Get native argc for shell completion generation during cross-compilation.
```
SuperSandro2000, line 7:
```suggestion
  stdenv,
  fetchFromGitHub,
  nixpkgsFun,
  argc,
```
This is all clear from the package definition in the rest of the file.

SuperSandro2000 (changes requested):

pkgs/by-name/ar/argc/package.nix
ShamrockLee: An alternative way is to add `argcNative` as an argument and specify it as `buildPackages.argc` inside `all-packages.nix`.

ShamrockLee:
@SuperSandro2000 Some diffs mentioned in the reviews are caused by the formatter. Should I also refactor them in this PR?

pkgs/by-name/ar/argc/package.nix
ShamrockLee: Which `pkgsCross` attribute should be chosen to test cross-compilation from `aarch64-linux` to `x86_64`?

pkgs/by-name/ar/argc/package.nix
ShamrockLee: Address by adding an argument `argcNative`,

pkgs/by-name/ar/argc/package.nix
SuperSandro2000: ``pkgsCross.gnu64`` should be able to do that.

Also we can omit the x86_64 condition. On the correct platforms this is a noop.

pkgs/by-name/ar/argc/package.nix
SuperSandro2000: We can just use buildPackages.argc. That's pretty common to do and has less indirection.

SuperSandro2000:
> Some diffs mentioned in the reviews are caused by the formatter. Should I also refactor them in this PR?

I would prefer to do less formatting changes to make the review easier but whatever you prefer.

ShamrockLee:
> > Some diffs mentioned in the reviews are caused by the formatter. Should I also refactor them in this PR?
> 
> I would prefer to do less formatting changes to make the review easier but whatever you prefer.

Next time I'll split out the formatting changes when working on similar features.

Given that the the formatted changes has been reviewed, I'm adding a commit to address them.

pkgs/by-name/ar/argc/package.nix
ShamrockLee: Addressed. Thanks a lot!

SuperSandro2000 (approved):

ShamrockLee:
@SuperSandro2000 Many thanks to you for reviewing this PR and polishing it!

PR #305753, opened 2024-04-21
solidpython2: init at 2.1.0

jonboh:
## Description of changes

Add solidpython2 python library.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/development/python-modules/solidpython2/default.nix
+++ b/pkgs/development/python-modules/solidpython2/default.nix
@@ -1,0 +1,49 @@
+{ buildPythonPackage
+, fetchFromGitHub
+, lib
+, ply
+, setuptools
+, poetry-core
+, withOpenSCAD ? false
+, openscad
+}:
+buildPythonPackage rec {
+  pname = "solidpython2";
+  version = "2.1.0";
+  pyproject = true;
+  src = fetchFromGitHub {
+    owner = "jeff-dh";
+    repo = "SolidPython";
+    rev = "v${version}";
+    hash = "sha256-Tq3hrsC2MmueCqChk6mY/u/pCjF/pFuU2o3K+qw7ImY=";
+  };
+
+  # NOTE: this patch makes tests runnable outside the source-tree
+  # - it uses diff instead of git-diff
+  # - modifies the tests output to resemble the paths resulting from running inside the source-tree
+  # - drop the openscad image geneneration tests, these don't work on the nix sandbox due to the need for xserver
+  patches = [ ./difftool_tests.patch ];
+
+  propagatedBuildInputs = lib.optionals withOpenSCAD [openscad];
+
+  build-system = [
+    poetry-core
+  ];
+  dependencies = [
+    ply
+    setuptools
+  ];
+  pythonImportsCheck = [ "solid2" ];
+  checkPhase = ''
+    runHook preCheck
+    python $TMPDIR/source/tests/run_tests.py
+    runHook postCheck
+  '';
+
+  meta = with lib; {
+    homepage = "https://github.com/jeff-dh/SolidPython";
+    description = "A python frontend for solid modelling that compiles to OpenSCAD";
+    license = licenses.lgpl2Plus;
+    maintainers = with maintainers; [ jonboh ];
+  };
+}
--- a/pkgs/development/python-modules/solidpython2/difftool_tests.patch
+++ b/pkgs/development/python-modules/solidpython2/difftool_tests.patch
@@ -1,0 +1,35 @@
+diff --git a/tests/examples_test.py b/tests/examples_test.py
+index 77204fd..9784389 100644
+--- a/tests/examples_test.py
++++ b/tests/examples_test.py
+@@ -48,14 +48,19 @@ class ExamplesTest(unittest.TestCase):
+             subprocess.check_call(["python3", f.as_posix()])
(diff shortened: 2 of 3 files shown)


pkgs/development/python-modules/solidpython2/default.nix
```diff
@@ -0,0 +1,34 @@
...
+  version = "2.1.0";
+  pyproject = true;
+  src = fetchFromGitHub {
+    owner = "jeff-dh";
+    repo = "SolidPython";
+    rev = "v${version}";
+    hash = "sha256-Tq3hrsC2MmueCqChk6mY/u/pCjF/pFuU2o3K+qw7ImY=";
+  };
+
+  nativeBuildInputs = [
```
dani0854, line 19:
```suggestion
  build-system = [
```

pkgs/development/python-modules/solidpython2/default.nix
```diff
@@ -0,0 +1,34 @@
...
+    poetry-core
+  ];
+  propagatedBuildInputs = [
```
dani0854, line 22:
```suggestion
  dependencies = [
```

pkgs/development/python-modules/solidpython2/default.nix
```diff
@@ -0,0 +1,34 @@
...
+    ply
+    setuptools
```
dani0854, line 24: Are you sure that's a run time python dependency, normally it's a build time dependency

pkgs/development/python-modules/solidpython2/default.nix
dani0854, line 1: Project does seem to contain tests, under test directory.
It's a good idea to run them, unfortunately by the look of it they meant to be run on source git repository.

`pythonImportsCheck` can also be used. But it looks like there are some non-python dependencies missing, such as `openscad`, which won't be tested by that.

dani0854 (changes requested):

pkgs/development/python-modules/solidpython2/default.nix
jonboh: I've been able to patch the tests so that they are runnable outside the source tree :)

jonboh: I've been able to patch the tests so that we are able to run the text output comparison part of the tests.
However I wasn't able to get the `openscad` part of the tests running. I've tried running the tests with `xvfb-run`, but `openscad` fails to generate the images with `glXChooseFBConfig failed`. I was able to run the tests in my desktop, so it seems that this is an issue of configuring the xserver correctly with xvfb-run.
But once OpenScad releases a new version this shouldn't be a problem, as it seems that there has been work in making `openscad` work in a headless way https://github.com/openscad/openscad/issues/4613#issuecomment-2146592621.
So I think its ok to drop that part of the tests until they are runnable without the need of xserver.

dani0854: I think it's ok to drop that part.

Though the question left, is `openscad` a runtime dependency of `solidpython2`, or just used for tests.

pkgs/development/python-modules/solidpython2/default.nix
```diff
@@ -0,0 +1,43 @@
...
+    poetry-core
+  ];
+  dependencies = [
+    ply
+    setuptools
+  ];
+  pythonImportsCheck = [ "solid2" ];
+  checkPhase = ''
+    python $TMPDIR/source/tests/run_tests.py
+  '';
```
dani0854, line 35:
```suggestion
  checkPhase = ''
    runHook preCheck

    python $TMPDIR/source/tests/run_tests.py

    runHook postCheck
  '';
```

dani0854 (changes requested):

pkgs/development/python-modules/solidpython2/default.nix
jonboh: There is some functionality in `solidpython2` that requires `openscad`, like rendering stls.
However in general you don't need to have `openscad` to use `solidpython2` and generate OpenSCAD code. Although you will need `openscad` to generate usable artifacts from that code.
In my case I dont have `openscad` from nixpkgs installed in my system, in order to process the output of SolidPython I use one of the development snapshots available in https://openscad.org/downloads.html
So I would say that it isn't necessary to add `openscad` as a dependency of the library itself, but let me know if you consider otherwise

dani0854: I am not really an expert in this. But perhaps if it's considered as an optional dependency, it can be declared using `withOpenSCAD ? false`, and then added via `lib.optional` to `propagatedBuildInputs`.

jonboh: yes, I think that's reasonable, thanks for the pointer on how to do the optional dependency :)

dani0854: Though I might be wrong about it, but I guess other reviewers can correct me if that's not the recommended approach here

dani0854 (approved):

SuperSandro2000:
@ofborg build python312Packages.solidpython2

PR #305836, opened 2024-04-21
nixos/doc: fix typo in abstractions.section.md

gabbieblue:
I found this while perusing the manual thinking about whether or not to install NixOS on my new laptop. It threw me for a loop for a second but as best I can tell this meant to be ".org" like the rest of the example, and not ".com"

Please let me know if this isn't the right place for this PR, I don't use GitHub and I couldn't find any instructions telling me otherwise


iblech (approved):
Awesome, thank you for your contribution! Indeed, this should be fixed. I acknowledge that this typo can cause confusion.

I would merge this if I had the permissions.

Aleksanaa (approved):
Please reword to `nixos/doc: fix typo in abstractions.section.md`

PR #305864, opened 2024-04-22
netbsd: move packages into their own modules

ali-abrar:
## Description of changes

Reorganized netbsd packages to follow the pattern established in the freebsd tree.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---


pkgs/os-specific/bsd/netbsd/default.nix
```diff
@@ -7,8 +7,6 @@
 }:
```
Ericson2314, line 7: I think some of these imports above are no longer needed?

pkgs/os-specific/bsd/netbsd/default.nix
```diff
@@ -28,9 +26,12 @@ let
...
   otherSplices = generateSplicesForMkScope "netbsd";
-  f = (self: let
-    inherit (self) mkDerivation;
-  in {
+  f = (self: lib.packagesFromDirectoryRecursive {
+    callPackage = self.callPackage;
+    directory = ./pkgs;
+  } // (let inherit (self) mkDerivation; in {
+
+  inherit defaultMakeFlags fetchNetBSD netbsdSetupHook;
```
Ericson2314, line 34: Can we move these 3 from the let block above directly into the inherit? I think nothing else is using them / if things are, we can get them via `self.<name>` instead? we can remove the `let..in` then.

pkgs/os-specific/bsd/netbsd/default.nix
```diff
@@ -28,9 +26,12 @@ let
...
 
   # Why do we have splicing and yet do `nativeBuildInputs = with self; ...`?
   #
```
Ericson2314, line 37: This comment is now partially addressed, and should be combined with the "Overridden arguments avoid cross package-set..." comment in FreeBSD's `default.nix`. We put the combined comment in one place, and reference it in the other place.

pkgs/os-specific/bsd/netbsd/default.nix
Ericson2314: Also now that we basically have all-new and shorter content within this `{....}` we should indent two more spaces to fix the indentation that was broken :).

Ericson2314: Finally `netbsdSetupHook` can be a package, see `freebsdSetupHook` which is done with a `package.nix` and shell script file,

pkgs/os-specific/bsd/netbsd/default.nix
ali-abrar: removed

pkgs/os-specific/bsd/netbsd/default.nix
ali-abrar: done

pkgs/os-specific/bsd/netbsd/default.nix
ali-abrar: updated

Ericson2314 (approved):

alyssais:
This has broken `pkgsCross.x86_64-freebsd.evdev-proto`:

```
error:
         at <nix/derivation-internal.nix>:9:12:
            8|
            9|   strict = derivationStrict drvAttrs;
             |            ^
           10|

       … while evaluating derivation 'evdev-proto-x86_64-unknown-freebsd13-6.7'
         whose name attribute is located at /home/qyliss/src/nixpkgs/pkgs/stdenv/generic/make-derivation.nix:331:7

       … while evaluating attribute 'ABI_FILE' of derivation 'evdev-proto-x86_64-unknown-freebsd13-6.7'
         at /home/qyliss/src/nixpkgs/pkgs/os-specific/bsd/freebsd/evdev-proto/default.nix:31:3:
           30|
           31|   ABI_FILE = runCommandCC "abifile" {} "$CC -shared -o $out";
             |   ^
           32|   CLEAN_FETCH_ENV = true;

       error: infinite recursion encountered
       at /home/qyliss/src/nixpkgs/pkgs/os-specific/bsd/netbsd/pkgs/mkDerivation.nix:63:25:
           62|   MKRELRO = "no";
           63| } // lib.optionalAttrs (stdenv'.cc.isClang or false) {
             |                         ^
           64|   HAVE_LLVM = lib.versions.major (lib.getVersion stdenv'.cc.cc);
```

PR #305902, opened 2024-04-22
python3Packages.uxsim: init at 1.1.1

vinnymeller:
## Description of changes

Adds the [UXsim](https://github.com/toruseo/UXsim) Python library

image
Animation created using this package in a dev shell
Just kidding it's not animated here. I pasted in a GIF :(

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/development/python-modules/uxsim/add-qt-plugin-path-to-env.patch
+++ b/pkgs/development/python-modules/uxsim/add-qt-plugin-path-to-env.patch
@@ -1,0 +1,24 @@
+diff --git a/uxsim/__init__.py b/uxsim/__init__.py
+index 01e1ad1..de1f0fd 100644
+--- a/uxsim/__init__.py
++++ b/uxsim/__init__.py
+@@ -1,8 +1,14 @@
+-from .uxsim import *
+-from .utils import *
++import os
++
+ from .analyzer import *
++from .utils import *
++from .uxsim import *
++
++# Only set our own plugin path if it's not already set
++if not os.getenv("QT_PLUGIN_PATH"): 
++    os.environ["QT_PLUGIN_PATH"] = "$NIX_QT_PLUGIN_PATH"
+ 
+ __version__ = "1.1.1"
+ __author__ = "Toru Seo"
+ __copyright__ = "Copyright (c) 2023 Toru Seo"
+-__license__ = "MIT License"
+\ No newline at end of file
++__license__ = "MIT License"
+
--- a/pkgs/development/python-modules/uxsim/default.nix
+++ b/pkgs/development/python-modules/uxsim/default.nix
@@ -1,0 +1,69 @@
+{
+  lib,
+  buildPythonPackage,
+  fetchFromGitHub,
+  setuptools,
+  wheel,
+  qt5,
+  hackgen-font,
+  python3,
+  matplotlib,
+  numpy,
+  pandas,
+  pillow,
+  pyqt5,
+  scipy,
+  tqdm,
+}:
+buildPythonPackage rec {
+  pname = "uxsim";
+  version = "1.1.1";
+  pyproject = true;
+
+  src = fetchFromGitHub {
+    owner = "toruseo";
+    repo = "UXsim";
+    rev = "v${version}";
+    hash = "sha256-qPAFodvx+Z7RsRhzdTq1TRsbvrUFaqRJZxYg3FM6q8A=";
+  };
+
+  patches = [
+    ./add-qt-plugin-path-to-env.patch
+  ];
+
+  nativeBuildInputs = [
+    setuptools
+    wheel
+  ];
+
+  propagatedBuildInputs = [
+    matplotlib
+    numpy
+    pandas
+    pillow
+    pyqt5
+    scipy
+    tqdm
+  ];
+
+  pythonImportsCheck = ["uxsim"];
+
+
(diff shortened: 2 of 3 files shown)


vinnymeller:
Result of `nixpkgs-review pr 305902` run on x86_64-linux [1](https://github.com/Mic92/nixpkgs-review)
<details>
  <ul>
    <li>python311Packages.uxsim</li>
    <li>python311Packages.uxsim.dist</li>
  </ul>
</details>

pkgs/development/python-modules/uxsim/default.nix
```diff
@@ -0,0 +1,69 @@
...
+
+  pythonImportsCheck = ["uxsim"];
+
+
+  # QT_PLUGIN_PATH is required to be set for the program to produce its images
+  # our patch sets it to $NIX_QT_PLUGIN_PATH if QT_PLUGIN_PATH is not set
+  # and here we replace this string with the actual path to qt plugins
+  postInstall = ''
+    substituteInPlace $out/${python3.sitePackages}/uxsim/__init__.py \
+      --replace-warn '$NIX_QT_PLUGIN_PATH' '${qt5.qtbase.bin}/${qt5.qtbase.qtPluginPrefix}'
```
kirillrdy, line 57: why not replace-fail ?

vinnymeller: an oversight I forgot to fix, ty!

pkgs/development/python-modules/uxsim/add-qt-plugin-path-to-env.patch
```diff
@@ -0,0 +1,24 @@
...
+-from .uxsim import *
+-from .utils import *
++import os
++
+ from .analyzer import *
++from .utils import *
++from .uxsim import *
++
++# Only set our own plugin path if it's not already set
++if not os.getenv("QT_PLUGIN_PATH"):
```
kirillrdy, line 15: this can be replaced by makeWrapper --set-defailt argument

https://nixos.org/manual/nixpkgs/unstable/#fun-makeWrapper

pkgs/development/python-modules/uxsim/default.nix
```diff
@@ -0,0 +1,69 @@
...
+  numpy,
+  pandas,
+  pillow,
+  pyqt5,
+  scipy,
+  tqdm,
+}:
+buildPythonPackage rec {
+  pname = "uxsim";
+  version = "1.1.1";
```
kirillrdy, line 20: 1.3 is out, but can be done as separate PR

pkgs/development/python-modules/uxsim/default.nix
```diff
@@ -0,0 +1,69 @@
...
+  ];
+
+  pythonImportsCheck = ["uxsim"];
+
+
+  # QT_PLUGIN_PATH is required to be set for the program to produce its images
+  # our patch sets it to $NIX_QT_PLUGIN_PATH if QT_PLUGIN_PATH is not set
+  # and here we replace this string with the actual path to qt plugins
+  postInstall = ''
+    substituteInPlace $out/${python3.sitePackages}/uxsim/__init__.py \
```
kirillrdy, line 56: this won't be necessary when using wrapper

pkgs/development/python-modules/uxsim/add-qt-plugin-path-to-env.patch
kirillrdy: on a second thought... that will not work for python module

kirillrdy:
@ofbrog build python3Packages.uxsim

pkgs/development/python-modules/uxsim/add-qt-plugin-path-to-env.patch
vinnymeller: @kirillrdy i was very unsure of what the preferred method might be here since it is a module. i don't particularly like the way i've done it here but it does work. i would be happy if you or anyone else has a better suggestion :slightly_smiling_face:

kirillrdy: i don't think there is a better option.

you could suggest your patch upstream ?

vinnymeller: hmm does it really make sense to go upstream in this case? idk its kind of convoluted

the patch sets env `QT_PLUGIN_PATH` to `$NIX_QT_PLUGIN_PATH` which is just a dummy string that only exists to get replaced in in the derivation with `substituteInPlace` with the real path

the way the patch currently works, if upstreamed, would just set `QT_PLUGIN_PATH` to a dummy string which I suppose wouldn't break anything but also probably isn't something I'd want to add to my own codebase

idk :zany_face:

kirillrdy: yeah, i guess you are right

kirillrdy:
@vinnymeller waiting for CI to pass will merge, unless you want to make further changes

vinnymeller:
im ok with it going as is, but am going to keep looking to see if theres a nicer way to handle the env var for the future

PR #305999, opened 2024-04-22
nixos: support `systemd-veritysetup-generator`

### LINKED ISSUE #285254: `/` does not need to be mounted if `roothash` is specified in the kernel command line
### Describe the bug
The initrd builder of NixOS makes assertions on [`fileSystems."/"` being specified](https://github.com/NixOS/nixpkgs/blob/c33213823590ca56d8c7aadb125dd62ab3d139ac/nixos/modules/system/boot/stage-1.nix#L688-L690) when building an initrd (i.e. its fstab, from which [`systemd-fstab-generator`](https://www.freedesktop.org/software/systemd/man/latest/systemd-fstab-generator.html#) [creates the mount units](https://github.com/NixOS/nixpkgs/blob/3dd36d266986d9c7ae39dca0e7b4a264c80f7fa9/nixos/doc/manual/configuration/file-systems.chapter.md?plain=1#L17)). When using a dm-verity system, with the `roothash` option specified in the kernel command line, [`systemd-veritysetup-generator`](https://www.freedesktop.org/software/systemd/man/latest/systemd-veritysetup-generator.html) will already have taken care of creating the mount unit, causing the subsequent try to create the unit by `systemd-fstab-generator` to fail. This makes it hard to build immutable, image-based appliances where one would want a dm-verity protected rootfs created through veritysetup without overrides to the module.

### Steps  To Reproduce
1. Build an image with a verity partition and the `roothash` option specified in the kernel command line.
2. Boot the image.
3. See the error message in stage 1:
  ```console
Welcome to NixOS 24.05 (Uakari) (Initrd)!

[    0.406772] systemd-fstab-generator[69]: Failed to create unit file '/run/systemd/generator/sysroot.mount', as it already exists. Duplicate entry in '/nix/store/2y5fanvgc23zfqpq0pjcm0v8hmfv452s-initrd-fstab'?
[    0.409288] (sd-e[67]: /nix/store/vfmf8qh892jfl107hih0yfnic00byjgj-systemd-254.6/lib/systemd/system-generators/systemd-fstab-generator failed with exit status 1.
```

### Expected behavior
Assertion should not hit or be able to be suppressed.

### Additional context
I'm not entirely sure what the best fix for this would be. Options I have thought about include:
- Adding a suppressor option for the assertion, which is probably the least involved fix, but also has the worst UI.
- Adding a `boot.initrd.systemd.veritysetup` (or similar) option, that specifies the necessary command line options and additional units for using veritysetup.
- Adding a `withVeritysetup` option to the `systemd` package, similar as it is done with [`withCryptsetup`](https://github.com/NixOS/nixpkgs/blob/a5ae45e45af0a697a219eb4da2827fe3cf4e9374/pkgs/os-specific/linux/systemd/default.nix#L98). NixOS modules could then check whether the attribute is enabled and include services and disable assertions based on that.

### Notify maintainers

Adding @ElvishJerricco, as he seems to be the most active maintainer of stage1 / systemd-initrd.

### Metadata

```console
 - system: `"x86_64-linux"`
 - host os: `Linux 6.7.2, NixOS, 24.05 (Uakari), 24.05.20240127.160b762`
 - multi-user?: `yes`
 - sandbox: `yes`
```

---

msanft:
## Description of changes

This adds support for using verity-protected root filesystems on NixOS, implemented by [`systemd-veritysetup-generator`](https://www.freedesktop.org/software/systemd/man/latest/systemd-veritysetup-generator.html). This generator looks for verity partitions (e.g. as created by `systemd-repart`) and mounts them automatically. It acts as an alternative to the other generators supported at the moment. (Namely `systemd-gpt-auto-generator` and `systemd-fstab-generator`).

The tentative API looks as follows:
```nix
config.boot.initrd.systemd.root = "verity" # This option specifies how root should be mounted at all.
config.boot.initrd.systemd.verityRootHash = "..." # This option specifies the root hash of the verity partition, which systemd uses to identify the correct device to mount.
```

I'm not sure yet whether this API is good. I couldn't think of anything else you'd need to have configurability of yet, but I'm happy to get some feedback on this.

Closes #285254

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [x] Tested, as applicable:

---

### DIFF
--- a/nixos/modules/system/boot/systemd/initrd.nix
+++ b/nixos/modules/system/boot/systemd/initrd.nix
@@ -72,7 +72,12 @@ let
     "timers.target"
     "umount.target"
     "systemd-bsod.service"
-  ] ++ cfg.additionalUpstreamUnits;
+  ] ++ cfg.additionalUpstreamUnits
+  ++ lib.optional (cfg.root == "verity") [
+    "veritysetup-pre.target"
+    "veritysetup.target"
+    "remote-veritysetup.target"
+  ];
 
   upstreamWants = [
     "sysinit.target.wants"
@@ -211,7 +216,7 @@ in {
     };
 
     root = lib.mkOption {
-      type = lib.types.enum [ "fstab" "gpt-auto" ];
+      type = lib.types.enum [ "fstab" "gpt-auto" "verity" ];
       default = "fstab";
       example = "gpt-auto";
       description = ''
@@ -220,6 +225,25 @@ in {
         allow specifying the root file system itself this
         way. Instead, the `fstab` value is used in order to interpret
         the root file system specified with the `fileSystems` option.
+      '';
+    };
+
+    verityRootHash = lib.mkOption {
+      type = lib.types.str;
+      default = "";
+      example = "6ed332bcfa615381511d4d5ba44a293bb476f368f7e9e304f0dff50230d1a85b";
+      description = ''
+        The root hash of the verity-protected root file system.
+
+        If set, the root file system will be mounted via DM-verity through
+        {manpage}`systemd-veritysetup-generator(8)` and no other root file
+        system mount option such as `fileSystems."/"` should be supplied.
+
+        Note that currently, for NixOS activation to succeed, certain parts of the root
+        file system must be writable, so writable (overlay) mounts need to be put in place
+        for activation to succeed. Furthermore, for dm-verity to work, the required udev rules
+        need to be put in place, which can conveniently be done by enabling the
+        {option}`boot.initrd.services.lvm.enable` option.
       '';
     };
 
@@ -237,7 +261,7 @@ in {
 
     initrdBin = mkOption {
       type = types.listOf types.package;
-      default = [];
+      default = [ ];
       description = ''
         Packages to include in /bin for the stage 1 emergency shell.
       '';
@@ -356,7 +380,15 @@ in {
     assertions = [
       {
         assertion = cfg.root == "fstab" -> any (fs: fs.mountPoint == "/") (builtins.attrValues config.fileSystems);
-        message = "The ‘fileSystems’ option does not specify your root file system.";
+        message = "The `fileSystems` option does not specify your root file system.";
+      }
+      {
+        assertion = cfg.root == "verity" -> cfg.verityRootHash != "";
+        message = "The `verityRootHash` option must be set when `root` is set to `verity`.";
+      }
+      {
+        assertion = cfg.root == "verity" -> !(builtins.hasAttr "/" config.fileSystems);
+        message = "The `fileSystems.\"/\"` option should not be set when `root` is set to `verity`.";
       }
     ] ++ map (name: {
       assertion = lib.attrByPath name (throw "impossible") config.boot.initrd == "";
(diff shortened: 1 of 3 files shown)


msanft:
@ofborg test verity-protected-root

flokli:
This looks like a nice way into the right direction! Also thanks for writing a test for it :heart: .

malt3 (approved):

nixos/modules/system/boot/systemd/initrd.nix
```diff
@@ -211,7 +216,7 @@ in {
     };
 
     root = lib.mkOption {
-      type = lib.types.enum [ "fstab" "gpt-auto" ];
+      type = lib.types.enum [ "fstab" "gpt-auto" "verity" ];
```
nikstur, line 219: I don't believe this is the right abstraction. You still need an `fstab` or `gpt-auto` for this to work because `systemd-veritysetup-generator` only opens `dm` devices. It doesn't mount anything and indeed you could configure your mounts with either the fstab or gpt-auto when you use dm-verity.

nixos/tests/verity-protected-root.nix
nikstur, line 1: This can be done much simpler, by avoiding to build full images. See: https://github.com/nikstur/nix-store-veritysetup-generator/blob/main/nix/tests/nix-store-veritysetup-generator.nix

nixos/tests/verity-protected-root.nix
```diff
@@ -0,0 +1,226 @@
...
+    # Provides 'veritysetup' to check if the verity-protected device
+    # has been mapped correctly.
+    environment.systemPackages = with pkgs; [ cryptsetup ];
+
+    boot.initrd = {
+      kernelModules = [ "overlay" ];
+      supportedFilesystems = [ "erofs" ];
+
+      # This adds some udev rules that are required for dm-verity to work.
+      services.lvm.enable = true;
```
nikstur, line 53: I find it odd that you need to enable this yourself. This should be enabled when you enable dm-verity support.

nikstur (changes requested):
Very cool that you're interested in dm-verity protected NixOS. That's something I've been working on for while as well!

My biggest issue with this PR is that it's all very entangled with the root kernel command line parameter while it doesn't have to be and probably shouldn't be.

Instead of adding another variant to the root enum, I believe we should add a separate module that you can enable. See https://github.com/nikstur/nix-store-veritysetup-generator/blob/main/nix/modules/verity.nix

If you really want a dm-verity protected root, you can then:

- enable this module
- set the `roothash=` kernel commandline 
- and set `fileSystems."/"` to use `/dev/mapper/root` and whichever fs options (e.g. `ro`) you need.

That's all that should be needed and still allows to use dm-verity differently, e.g. to protect  `/usr` or `/nix/store` with [nix-store-veritysetup-generator](https://github.com/nikstur/nix-store-veritysetup-generator/tree/main)

IMO the test shows (by it's sheer complexity), how ill-suited a verity protected root fs is for NixOS. We'd get the same (or better) security properties if we just protect the Nix store and use some form of impermanence for the root. The simplest form of impermanence would be a tmpfs.  What I've already tested and which works remarkable well is to mount the Nix store under `/usr/nix/store` and then bind mount it to `/nix/store`. Then you can just use the systemd veritysetup-generator.

nixos/tests/verity-protected-root.nix
msanft: Yeah, I agree. But looking around, I think it's uncommon to have modules enable things in other modules. Is there some kind of agreement around this?

nixos/modules/system/boot/systemd/initrd.nix
msanft: It does mount stuff (see `roothash` in https://www.freedesktop.org/software/systemd/man/latest/systemd-veritysetup-generator.html), and it's incompatible with both `root=gpt-auto` and `root=fstab` unfortunately. I'd very much liked the latter to not be the case, but it is. 

Re "automatically mounting", also see [this Matrix discussion](https://matrix.to/#/!DBFhtjpqmJNENpLDOv:nixos.org/$WfP271DEhtCF47-AzBuos4kitpTA8IeO9r1HMHkOHls?via=nixos.org&via=matrix.org&via=tchncs.de).

nixos/modules/system/boot/systemd/initrd.nix
```diff
@@ -356,7 +380,15 @@ in {
...
-        message = "The ‘fileSystems’ option does not specify your root file system.";
+        message = "The `fileSystems` option does not specify your root file system.";
+      }
+      {
+        assertion = cfg.root == "verity" -> cfg.verityRootHash != "";
+        message = "The `verityRootHash` option must be set when `root` is set to `verity`.";
+      }
+      {
+        assertion = cfg.root == "verity" -> !(builtins.hasAttr "/" config.fileSystems);
+        message = "The `fileSystems.\"/\"` option should not be set when `root` is set to `verity`.";
```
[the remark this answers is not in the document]
msanft: I think that's not the case when using `roothash`. 

However, I'm more and more unsure about whether `roothash` is the thing to do, as it also requires a DDI. If we used `systemd.verity_root_data=, systemd.verity_root_hash=`, we could probably offer a universal interface that doesn't mount anything, see [this Matrix discussion](https://matrix.to/#/!DBFhtjpqmJNENpLDOv:nixos.org/$WfP271DEhtCF47-AzBuos4kitpTA8IeO9r1HMHkOHls?via=nixos.org&via=matrix.org&via=tchncs.de)

PR #306026, opened 2024-04-22
lib.trivial: rename NIX_ABORT_ON_WARN to NIXPKGS_ABORT_ON_WARN

stuebinm:
## Description of changes

Renames the `NIX_ABORT_ON_WARN` environment variable to `NIXPKGS_ABORT_ON_WARN`to make it consistent with other environment variables read by nixpkgs to configure its behaviour (e.g. `NIXPKGS_ALLOW_INSECURE` etc.). The previous naming appeared to suggest that the variable is read by Nix itself, which it is not. This is especially a source of confusion when used together with Nix's pure evaluation mode, in which case setting the variable has no effect.

The old `NIX_ABORT_ON_WARN` is still honoured, but produces a deprecation warning. This warning's check is added as a wrapper to the entire lib.trivial attrset, as it should be checked (and, if necessary, printed) regardless of whether or not a use of `lib.warn` is encountered, but whenever the `lib.warn` function is available (i.e. whenever `lib` itself is evaluated), so that users of `NIX_ABORT_ON_WARN` will see it even if their code does not currently produce any warnings. Likewise, this means that the warning does not actually abort evaluation, as otherwise downstream users of `NIX_ABORT_ON_WARN` would be immediately presented with an unrecoverable error.

This deprecation warning and honouring of the older behaviour can then be removed in a future release.

## Downsides

For reasons which remain mysterious to me, the deprecation warning is printed eight times if `nixpkgs` is imported as a flake:

~~~
$ NIX_ABORT_ON_WARN=1 nix build .#hello --impure
trace: warning: NIX_ABORT_ON_WARN has been renamed to NIXPKGS_ABORT_ON_WARN. NIX_ABORT_ON_WARN is still honored, but deprecated and will be removed from nixpkgs after the 24.11 release.
trace: warning: NIX_ABORT_ON_WARN has been renamed to NIXPKGS_ABORT_ON_WARN. NIX_ABORT_ON_WARN is still honored, but deprecated and will be removed from nixpkgs after the 24.11 release.
trace: warning: NIX_ABORT_ON_WARN has been renamed to NIXPKGS_ABORT_ON_WARN. NIX_ABORT_ON_WARN is still honored, but deprecated and will be removed from nixpkgs after the 24.11 release.
trace: warning: NIX_ABORT_ON_WARN has been renamed to NIXPKGS_ABORT_ON_WARN. NIX_ABORT_ON_WARN is still honored, but deprecated and will be removed from nixpkgs after the 24.11 release.
trace: warning: NIX_ABORT_ON_WARN has been renamed to NIXPKGS_ABORT_ON_WARN. NIX_ABORT_ON_WARN is still honored, but deprecated and will be removed from nixpkgs after the 24.11 release.
trace: warning: NIX_ABORT_ON_WARN has been renamed to NIXPKGS_ABORT_ON_WARN. NIX_ABORT_ON_WARN is still honored, but deprecated and will be removed from nixpkgs after the 24.11 release.
trace: warning: NIX_ABORT_ON_WARN has been renamed to NIXPKGS_ABORT_ON_WARN. NIX_ABORT_ON_WARN is still honored, but deprecated and will be removed from nixpkgs after the 24.11 release.
trace: warning: NIX_ABORT_ON_WARN has been renamed to NIXPKGS_ABORT_ON_WARN. NIX_ABORT_ON_WARN is still honored, but deprecated and will be removed from nixpkgs after the 24.11 release.
~~~

but only once if imported in the traditional way:

~~~
$ NIX_ABORT_ON_WARN=1 nix build -f . hello
trace: warning: NIX_ABORT_ON_WARN has been renamed to NIXPKGS_ABORT_ON_WARN. NIX_ABORT_ON_WARN is still honored, but deprecated and will be removed from nixpkgs after the 24.11 release.
~~~

This is only a minor annoyance, but does perhaps (?) point to an unrelated issue with how the nixpkgs flake handles importing `lib`.

## Alternatives

There are a few possible placements for the deprecation warning & its check in the code:
 - inside `lib.warn` itself: seems the most logical place to put it, but it would then only be shown if a warning is actually produced. As (presumably) most users of `NIX_ABORT_ON_WARN` will strive to keep their code free of warnings, this could lead to them never seeing the deprecation message.
 - outside of `lib/`, e.g. in `pkgs/top-level/impure.nix`: i am very hesitant to touch that file, since even slight mistakes would possibly have large consequences. However, it is the only place i could find where, if placed there, the warning is only ever printed once even when using nixpkgs as a flake.

## Question

Should this change be mentioned in the release notes? I was unable to find any clear guidelines on what (if any) changes to `lib` should get an entry there.

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/lib/trivial.nix
+++ b/lib/trivial.nix
@@ -12,7 +12,20 @@ let
     version
     versionSuffix
     warn;
-in {
+
+  # We have to use builtins here in place of lib.warn, as otherwise its
+  # definition would depend on itself. Further, we do not want this to
+  # actually abort even if NIX_ABORT_ON_WARN is set, as otherwise legacy
+  # uses of this variable would immediately result in an unrecoverable
+  # error where there should be a deprecation warning.
+  printDeprecationWarning =
+    if builtins.getEnv "NIX_ABORT_ON_WARN" != "" then
+      builtins.trace
+        "[1;31mwarning: NIX_ABORT_ON_WARN has been renamed to NIXPKGS_ABORT_ON_WARN. NIX_ABORT_ON_WARN is still honored, but deprecated and will be removed from nixpkgs after the 24.11 release.[0m"
+    else
+      a: a;
+
+in printDeprecationWarning {
 
   ## Simple (higher order) functions
 
@@ -698,8 +711,10 @@ in {
     ```
   */
   warn =
-    if lib.elem (builtins.getEnv "NIX_ABORT_ON_WARN") ["1" "true" "yes"]
-    then msg: builtins.trace "[1;31mwarning: ${msg}[0m" (abort "NIX_ABORT_ON_WARN=true; warnings are treated as unrecoverable errors.")
+    if lib.elem (builtins.getEnv "NIXPKGS_ABORT_ON_WARN") [ "1" "true" "yes" ]
+    # this latter clause is deprecated. When removing, also remove printDeprecationWarning above.
+    || lib.elem (builtins.getEnv "NIX_ABORT_ON_WARN") [ "1" "true" "yes" ]
+    then msg: builtins.trace "[1;31mwarning: ${msg}[0m" (abort "NIXPKGS_ABORT_ON_WARN=true; warnings are treated as unrecoverable errors.")
     else msg: builtins.trace "[1;31mwarning: ${msg}[0m";
 
   /**
--- a/lib/trivial.nix
+++ b/lib/trivial.nix
@@ -19,11 +19,15 @@ let
   # uses of this variable would immediately result in an unrecoverable
   # error where there should be a deprecation warning.
   printDeprecationWarning =
-    if builtins.getEnv "NIX_ABORT_ON_WARN" != "" then
+    if 2405 <= oldestSupportedRelease && builtins.getEnv "NIX_ABORT_ON_WARN" != "" then
       builtins.trace
         "[1;31mwarning: NIX_ABORT_ON_WARN has been renamed to NIXPKGS_ABORT_ON_WARN. NIX_ABORT_ON_WARN is still honored, but deprecated and will be removed from nixpkgs after the 24.11 release.[0m"
     else
       a: a;
+
+  # Temporarily moved here to prevent lib.trivial from referencing itself.
+  # See the `inherit oldestSupportedRelease' further down.
+  oldestSupportedRelease = 2311;
 
 in printDeprecationWarning {
 
@@ -390,9 +394,8 @@ in printDeprecationWarning {
     This release number allows deprecation warnings to be implemented such that
     they take effect as soon as the oldest release reaches end of life.
   */
-  oldestSupportedRelease =
     # Update on master only. Do not backport.
-    2311;
+  inherit oldestSupportedRelease;
 
   /**
     Whether a feature is supported in all supported releases (at the time of
--- a/nixos/modules/misc/documentation.nix
+++ b/nixos/modules/misc/documentation.nix
@@ -101,7 +101,7 @@ let
           libPath = filter (pkgs.path + "/lib");
           pkgsLibPath = filter (pkgs.path + "/pkgs/pkgs-lib");
           nixosPath = filter (pkgs.path + "/nixos");
-          NIX_ABORT_ON_WARN = warningsAreErrors;
+          NIXPKGS_ABORT_ON_WARN = warningsAreErrors;
           modules =
             "[ "
             + concatMapStringsSep " " (p: ''"${removePrefix "${modulesPath}/" (toString p)}"'') docModules.lazy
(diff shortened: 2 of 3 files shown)


lib/trivial.nix
```diff
@@ -12,7 +12,20 @@ let
...
     warn;
-in {
+
+  # We have to use builtins here in place of lib.warn, as otherwise its
+  # definition would depend on itself. Further, we do not want this to
+  # actually abort even if NIX_ABORT_ON_WARN is set, as otherwise legacy
+  # uses of this variable would immediately result in an unrecoverable
+  # error where there should be a deprecation warning.
+  printDeprecationWarning =
+    if builtins.getEnv "NIX_ABORT_ON_WARN" != "" then
```
infinisil, line 22:
```suggestion
    if lib.isInOldestRelease 2405 && builtins.getEnv "NIX_ABORT_ON_WARN" != "" then
```

This needs some code rearranging to get working without infinite recursion, but this makes sure that there's a transition period where stable third party projects can switch to `NIXPKGS_ABORT_ON_WARN` without warnings for end-users.

lib/trivial.nix
```diff
@@ -698,8 +711,9 @@ in {
     ```
   */
   warn =
-    if lib.elem (builtins.getEnv "NIX_ABORT_ON_WARN") ["1" "true" "yes"]
-    then msg: builtins.trace "[1;31mwarning: ${msg}[0m" (abort "NIX_ABORT_ON_WARN=true; warnings are treated as unrecoverable errors.")
+    if lib.elem (builtins.getEnv "NIXPKGS_ABORT_ON_WARN") [ "1" "true" "yes" ]
+    || lib.elem (builtins.getEnv "NIX_ABORT_ON_WARN") [ "1" "true" "yes" ]
```
infinisil, line 715: Let's have a comment here to mention that this one is deprecated, referring to `printDeprecationWarning`

infinisil (changes requested):
Just minor thoughts, but overall this is looking good to me! This is a really tricky change, but your explanation makes a lot of sense!

The multiple messages for Flakes can be fixed by putting the `builtins.trace` before even the `{ lib }:` part of the file (`builtins.trace "..." ({ lib }: ...)`). But this does feel a bit too ugly..

So indeed I think it would be better to figure out why the Flake handling appears to import `lib` that many times. But this can be done in a separate PR :)

Ping @roberth as the one who introduced this environment variable

roberth:
While your reasoning makes sense, my original goal was to build this functionality into Nix.
I haven't followed through on that, because at the time, there wasn't a Nix team yet, and it proved really hard to get anything merged.
So my preferred way forward is to still build this into Nix.

1. Add `builtins.warn`, which logs at the correct Nix log level instead of just ANSI codes at "info" level.
2. Add a Nix config option `abort-on-warn` that defaults to the `NIX_ABORT_ON_WARN` value.
3. Change `lib.warn` to `builtins.warn or ...`, where the current implementation will serve as a good polyfill.

This PR would not prevent that from happening, but it would be messier, as users first have to learn to use `NIXPKGS_ABORT_ON_WARN`, and then switch back to `NIX_ABORT_ON_WARN` or `--abort-on-warn`. If we leave Nixpkgs as is, there's no need to learn anything.

Also there's an argument to be made that `lib` is fairly universal, and users might not perceive themselves as interacting with Nixpkgs when they need this behavior. For instance, they might be trying to solve a problem in a module system application like home-manager, which is not Nixpkgs.

lib/trivial.nix
stuebinm: I have done so, though aesthetically I don't particularly like the result. If you can think of any nicer way to do this, please let me know / feel free to edit this branch.

lib/trivial.nix
stuebinm: I've added one

lorenzleutgeb:
When I found out about this environment variable recently (https://github.com/ngi-nix/ngipkgs/issues/155 and consequently https://discourse.nixos.org/t/disallow-warnings-on-nix-flake-check/38964) I had this change in mind, too, but did not follow through. I support @stuebinm's reasoning. No matter what happens upstream in Nix, this environment variable should have the prefix `NIXPKGS_` not `NIX_` (as explained). If at some point a new built-in or similar appears, then there'll just have to be another change. I am one of the few (?) "users" of this environment variable and don't mind re-learning.

roberth:
I've just opened https://github.com/NixOS/nix/pull/10592. Sorry for dropping the ball earlier.
I think the team can merge this after the release, if not sooner.

PR #306072, opened 2024-04-22
workflows: add concurrency group to check-by-name workflow

willbush:
## Description of changes

Noticed yesterday that multiple runs can happen at the same time.
On event that triggers the workflow for the matching concurrency group, any previous runs of the workflow in the same group will be cancelled.

## Things done

---


infinisil (approved):
Briefly tested it in https://github.com/tweag/nixpkgs/pull/92, seems to work!

(this is necessary because this PR itself uses the workflow file from the base branch, so this file itself is not even evaluated)

PR #291339, opened 2024-02-25
GNOME: 45 → 46

amaxine:
This draft PR starts update work for the upcoming GNOME 46 release.

- Schedule: https://wiki.gnome.org/FortySix
- Project: TBD
- Previous major release: #247766
- Tool for reviewing changes: https://github.com/jtojnar/nonemast
- Release announcement: https://release.gnome.org/46/
- Hydra jobset: https://hydra.nixos.org/jobset/nixpkgs/gnome

### How to help

- test running your system from this branch (warning: you might not be able to roll back to 45 without dropping all configuration)
- test running new versions of software from this branch
- check hydra builds
  1. look at the latest evaluation in the [jobset](https://hydra.nixos.org/jobset/nixpkgs/gnome)
  2. check the _newly/still failing jobs_ tab, look for red crosses ❌ 
  3. open the individual failures' logs
  4. if you see any failures that look like dependency issue or anything that could be caused by this PR, report it in a comment, or open a PR against gnome-46 branch, if you are able to.
- report any and all problems!

### Needs work

- [ ] Test and fix darwin stuff
  - [x] [ofborg-eval-nixpkgs-unstable-jobset](https://gist.github.com/GrahamcOfBorg/4371e895f4088a18c02e16ed9ff8c319) fails (gst-plugins-base)
- [ ] g-c-c g-r-d doesn't support password generation: ` WARNING: Failed to open file: Error opening file /usr/share/dict/linux.words: No such file or directory`
  - consider setting `environment.wordlist.enable = true` as part of g-r-d? not enough by itself, would still need to patch
- [x] keyring unlocking doesn't use native prompts all of a sudden?
  - cannot repro
- [x] figure out if there's been any implementation changes that require something on our side for mounting from g-o-a - mounting dav/google accounts?
  - all good
- [x] snapshot crashes on startup/can't open stream
  - https://gitlab.gnome.org/GNOME/snapshot/-/issues/134 maybe this
- [ ] Try the new `gi-compile-repository`, `gi-decompile-typelib` and `gi-inspect-typelib`
  - Decide which output they should belong to (they are in `-bin` right now)
  - Figure out how the gobject-introspection-wrapped wrapper might be used here (?? tbh idk)
- [x] loupe/glycin crashes when loading images
  - seems to be sandboxing/bubblewrap related - we need to fix the sandbox patch, the bind mounts make no sense in the context of nix
[...]

PR #306080, opened 2024-04-22
staging-next 2024-04-22

vcunat:
According to the [24.05 schedule](https://github.com/NixOS/nixpkgs/issues/303285), this is the last staging-next iteration that might contain breaking changes to release critical packages.

##### Helpful links

- workflow docs: https://github.com/NixOS/nixpkgs/blob/master/CONTRIBUTING.md#staging
- important jobs:
  https://hydra.nixos.org/job/nixpkgs/staging-next/unstable#tabs-constituents
  https://hydra.nixos.org/job/nixos/staging-next-small/tested
- jobsets:
  https://hydra.nixos.org/jobset/nixpkgs/staging-next
  https://hydra.nixos.org/jobset/nixos/staging-next-small
- nix-review reports: https://malob.github.io/nix-review-tools-reports/
- previous staging-next: https://github.com/NixOS/nixpkgs/pull/303527
- - - 
##### Mass breakages
_nothing notable so far_


vcunat:
- [ ] `python3Packages.urwid` tests regressed, probably after glib or related upgrades
  https://hydra.nixos.org/build/257645050/nixlog/2/tail
  (all four platforms, both python versions, review says ~300 builds block because of that)

bobby285271:
> https://github.com/NixOS/nixpkgs/pull/306080#issuecomment-2080449253

Adding `gobject-introspection` to its `nativeCheckInputs` seems to fix the build, I wonder if this can be related to https://github.com/NixOS/nixpkgs/pull/291339#issuecomment-1987088766.

vcunat:
Oh, that's a move forward.  Though it does make me wonder... if this is just about fixing the tests (check inputs) or some of the glib-related python bindings are actually missing a runtime/propagated dependency.

jtojnar:
Yeah, sounds like the issue of GLib bindings not being automatically pulled into `pygobject3` by `libgirepository` any more (see also https://github.com/NixOS/nixpkgs/pull/291339#issuecomment-2049784712). We should actually need just `glib` and `gobject-introspection.setupHook` but adding `gobject-introspection` propagates both.

I believe it should be possible to use `pygobject3` without `GLib` bindings in theory so propagating `glib` is not exactly correct. But since all practical libraries will probably need `GLib` or `GObject` anyway, it might still be a a good idea to propagate `glib` from `pygobject3`.

We probably want to propagate `gobject-introspection.setupHook` from `pygobject3` all the time because without the environment variables set by the setup hook, no bindings will be available. Not sure how setup hooks interact with propagation though.

I do not think we do not want to propagate `gobject-introspection` since it now just contains typelibs for bunch of third-party libraries (e.g. cairo, fontconfig, xrandr…) which are not needed most of the time, the old libgirepository (which is usually only linked against) and build tools (needed for producing bindings, not for consuming them).

vcunat:
IIRC setup hooks should combine well, i.e. get run even for inputs that were (transitively) propagated to the current build.

vcunat:
Uh, the dependency on variables pointing to GIR... I suspect that will be cumbersome for python3 stuff, e.g. now this thing won't work by itself anymore: `from gi.repository import GLib` (you'll need a wrapper).

Propagating just `gobject-introspection.setupHook` wouldn't fix `python3Packages.pydbus` build, so let me propagate it whole for now.  (feel free to improve)

SuperSandro2000:
> Uh, the dependency on variables pointing to GIR... I suspect that will be cumbersome for python3 stuff, e.g. now this thing won't work by itself anymore: `from gi.repository import GLib` (you'll need a wrapper).

To my knowledge you always needed the wrapper for applications for gobject stuff to work.

jtojnar:
> Propagating just `gobject-introspection.setupHook` wouldn't fix `python3Packages.pydbus` build, so let me propagate it whole for now. (feel free to improve)

As mentioned [above](https://github.com/NixOS/nixpkgs/pull/306080#issuecomment-2081520383), you would need both the setup hook and `glib`. `gobject-introspection` itself is not needed.

> To my knowledge you always needed the wrapper for applications for gobject stuff to work.

If you just needed GLib (or other bindings shipped with gobject-introspection), wrapper was not needed since `${gobject-introspection}/lib/girepository-1.0` is implicitly on typelib path. Now that GLib bindings were moved to `glib` package, `${glib}/lib/girepository-1.0` must be added to `GI_TYPELIB_PATH` by the wrapper (or `gobject-introspection` setup hook for build time).

vcunat:
I now tried `gobject-introspection.setupHook glib` but it didn't suffice for `python3Packages.pydbus`.

jtojnar:
Just tried it and it works for me:

```patch
diff --git a/pkgs/development/python-modules/pydbus/default.nix b/pkgs/development/python-modules/pydbus/default.nix
index fc8965fd3538..8e47b4489e17 100644
--- a/pkgs/development/python-modules/pydbus/default.nix
+++ b/pkgs/development/python-modules/pydbus/default.nix
@@ -3,6 +3,8 @@
 , fetchFromGitHub
 , setuptools
 , pygobject3
+, gobject-introspection
+, glib
 }:
 
 buildPythonPackage rec {
@@ -24,12 +26,17 @@ buildPythonPackage rec {
 
   nativeBuildInputs = [
     setuptools
+    gobject-introspection.setupHook
   ];
 
   propagatedBuildInputs = [
     pygobject3
   ];
 
+  buildInputs = [
+    glib
+  ];
+
   pythonImportsCheck = [
     "pydbus"
     "pydbus.generic"
```

vcunat:
I was only modifying pygobject3.  Maybe I didn't choose the right combination of _native_ and _propagated_.

jtojnar:
If I am reading the [reference](https://nixos.org/manual/nixpkgs/stable/#ssec-stdenv-dependencies-reference) right, I believe, if `pygobject3` goes to `buildInputs` (`(0,1)`), `gobject-introspection.setupHook` should go to `depsBuildBuildPropagated` (`(-1,-1)`) to result in `(-1,0)` (`nativeBuildInputs`), and `glib` should go to `depsHostHostPropagated` (`(0,0)`), if we want it in `(0,1)` (`buildInputs`).

jtojnar:
I tried https://github.com/NixOS/nixpkgs/pull/308488 and it fixes `pydbus` but it only works with `glib` in `depsBuildBuildPropagated`, not in `depsHostHostPropagated` I would expect.

vcunat:
Well, it's not great, bringing a few hundred build regressions (closest view probably [here](https://hydra.nixos.org/eval/1806057?compare=1806045#tabs-now-fail) so far), mostly because of python packages &ndash; but:
- I see nothing critical
- just waiting wouldn't help much, most likely
- ZHF will be soon, so hopefully the issues will attract more attention and get reduced.

PR #306091, opened 2024-04-22
mockoon: 6.1.0 -> 8.1.1

PrimaMateria:
## Description of changes

- https://mockoon.com/releases/6.2.0/
- https://mockoon.com/releases/7.0.0/
- https://mockoon.com/releases/8.0.0/

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

Comments: 

- There is no mockoon test inside /nixos/tests
- There is no passthru test defined in the current package.
- `nix-shell -p nixpkgs-review --run "nixpkgs-review rev HEAD"` was always dying with message `<Signals.SIGKILL: 9>`.
- tested that I pointed nixpkgs in my locale flake to modified local nixpkgs directory and installed and run the new version of this package successfully this way
 

---


imadnyc:
Is there a reason that this isn't building from the `.deb` instead? `.appimage` replicates dependencies, so generally it's better to build from the `.deb`. Also, if it's just going to be a simple version bumping, consider adding `nix-update` so that it auto updates next time? Like in here: https://github.com/NixOS/nixpkgs/pull/300581.

PrimaMateria:
There are `.deb` assets available only for `arm64` and `amd64`, but none for `x86_64`. I'll attempt to add the `nix-update`.

There's a common issue when utilizing GitHub release assets as they're removed for older versions. I attempted to build from the source, but unfortunately, I was unsuccessful. It's an electron app within a monorepo managed by lerna.

PrimaMateria:
@imadnyc I have added nix-update-script and run following:

```
$ nix-shell maintainers/scripts/update.nix --argstr package mockoon
this derivation will be built:
  /nix/store/2ifkiymixwwpwm34s28c7hnlvlxmllwj-packages.json.drv
building '/nix/store/2ifkiymixwwpwm34s28c7hnlvlxmllwj-packages.json.drv'...

Going to be running update for following packages:
 - mockoon-8.0.0

Press Enter key to continue...

Running update for:
 - mockoon-8.0.0: UPDATING ...
 - mockoon-8.0.0: DONE.

Packages updated!
```

But nothing happened. At this moment, the latest is 8.1.1. The [repository](https://github.com/mockoon/mockoon) uses tags. Can it be that it doesn't work because `appimageTools.wrapType2` is used?

imadnyc:
https://github.com/NixOS/nixpkgs/blob/38c01297e7ec11f7b9e3f2cae7d6fcec6cc767ec/doc/stdenv/stdenv.chapter.md?plain=1#L516-L523

So here, we can see that nix-update-script is just a wrapper for [nix-update](https://github.com/Mic92/nix-update). There, we can see that it detects a host of different sources,  including github. So why doesn't it work?

```
  src = fetchurl {
    url = "https://github.com/mockoon/mockoon/releases/download/v${version}/mockoon-${version}.x86_64.AppImage";
    hash = "sha256-mhUjV8yFXS76kJDj28VeIv4/PlnKos/Ugo9k3RHnRaM=";
  };
```
Here, you're passing in the raw URL, and nix has no idea that it's getting a github link -- this could be from google.com as far as it's concerned. So `update-script` doesn't affect it. The way that we make nix that it's getting a github source is by using `fetchFromGithub`.


```
  src = fetchFromGitHub {
    owner = "mockoon";
    repo = "mockoon";
    rev = "8.0.0";
    hash = "sha256-whatever";
  };
```

Here's the issue though, it can't get the prebuilt binary this way, only the source. So how do we get around this?

We could either 
  - make our own script, and pass that to `passthru.updateScript`
  - build from source
 
The easier approach is to create a custom script, something that takes the JSON of the github tags and replaces the string, like here https://github.com/NixOS/nixpkgs/blob/38c01297e7ec11f7b9e3f2cae7d6fcec6cc767ec/pkgs/applications/finance/denaro/update.sh#L8. 

The better approach is to build from source, something that I'd have to look into later. It doesn't seem that mockoon gives build instructions, so this might be a challenge. That said, as for the `.deb`, `amd64` is `x86_64`, so you could just use that one.

imadnyc:
I wouldn't bother if you're gonna build from source later anyway -- there's no reason to if it's not going to last a few updates anyway. I think this would look good if you just removed the `updateScript`, and then we could work on it in the next PR.

PrimaMateria:
I just finished the update.sh when I read your last comment. I can remove it then, because even it didn't work as your example from denaro. The problem is that `appimageTools.wrapType2`, for to me unknow reason, removes the version attribute from the derivation, but adds it to the `pname`. I modified the update.sh to extract the old version from the `pname` and then instead of `update-source-version` I used `sed`. At the end it worked, but I am unsure if this is a valid way.

PrimaMateria:
Wait, please don't merge. I have just realized I didn't update the hash.

PrimaMateria:
Fixed. And also I realized that the update.sh I wrote is then wrong because it doesn't calculate new hash, it just uses `sed` to overwrite the version in the nix file. Anyway, update.sh is now removed.

PrimaMateria:
@dit7ya Anything else left before it can be merged?

pkgs/tools/networking/mockoon/default.nix
```diff
@@ -1,15 +1,16 @@
...
 
 let
   pname = "mockoon";
-  version = "6.1.0";
+  version = "8.1.1";
 
   src = fetchurl {
-    url = "https://github.com/mockoon/mockoon/releases/download/v${version}/mockoon-${version}.AppImage";
-    hash = "sha256-harZU3TTIzfJoY/jAQI0dm7YSOr24Y9xk9L5ZaBLdD8=";
+    url = "https://github.com/mockoon/mockoon/releases/download/v${version}/mockoon-${version}.x86_64.AppImage";
```
pbsds, line 12: please update `meta.platforms` to reflect the reduced platform support

PrimaMateria: I've updated the platforms to an intersection of Linux and x86_64

pkgs/tools/networking/mockoon/default.nix
```diff
@@ -1,15 +1,16 @@
 { lib
 , appimageTools
 , fetchurl
+, nix-update-script
```
pbsds, line 4:
```suggestion
```

Please either remove it or use it

pbsds (commented):
Could you squash the commits?

dit7ya (changes requested):

wegank:
Done in #381182.

PR #306159, opened 2024-04-23
swhkd: add package and module

lolbinarycat:
## Description of changes
yet another attempt at adding swhkd to nixpkgs, since all the previous attempts seem to have stagnated.

additionally, there's a few things this PR has over the others i've seen:
1. properly install manpages and polkit policy
2. split-output derivation 
3. nixos module

currently the nixos module doesn't do much besides writing to swhkdrc (although that is quite important, as swhkd will fail to start if the global config file doesn't exist).  users will have to manually launch swhkd via pkexec, or put something in their shell login script.  if someone who understands the nixos security model can tell me the right way to autostart it, that would be appreciated.

the package is built from the main branch, as the Makefile for the latest stable release is not good enough for nix.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/nixos/modules/services/wayland/swhkd.nix
+++ b/nixos/modules/services/wayland/swhkd.nix
@@ -5,23 +5,17 @@ let
   cfg = config.services.swhkd;
 in {
   options.services.swhkd = {
-    enable = mkEnableOption (lib.mdDoc "simple wayland hotkey daemon");
+    enable = mkEnableOption "simple wayland hotkey daemon";
     package = mkPackageOption pkgs "swhkd" {};
-    installManpages = mkOption {
-      type = lib.types.bool;
-      default = true;
-      description = "install swhkd manual pages";
-    };
     swhkdrc = mkOption {
-      type = lib.types.str;
+      type = lib.types.lines;
       default = "";
-      description = "contents of the system-wide swhkdrc file";
+      description = "contents of the system-wide swhkdrc file. See {manpage}`swhkd(5) for more details on the config format.";
     };
   };
 
   config = mkIf cfg.enable {
-    environment.systemPackages =
-      [ cfg.package.bin cfg.package.out ] ++ optional cfg.installManpages cfg.package.man;
+    environment.systemPackages = [ cfg.package ];
     environment.etc."swhkd/swhkdrc".text = cfg.swhkdrc;
     security.polkit.enable = true;
   };
--- a/nixos/modules/services/wayland/swhkd.nix
+++ b/nixos/modules/services/wayland/swhkd.nix
@@ -15,7 +15,7 @@ in {
   };
 
   config = mkIf cfg.enable {
-    environment.systemPackages = [ cfg.package ];
+    environment.systemPackages = [ cfg.package ] ++ [ cfg.package.out ];
     environment.etc."swhkd/swhkdrc".text = cfg.swhkdrc;
     security.polkit.enable = true;
   };
--- a/nixos/modules/services/wayland/swhkd.nix
+++ b/nixos/modules/services/wayland/swhkd.nix
@@ -1,0 +1,28 @@
+{ config, pkgs, lib, ... }:
+
+let
+  inherit (lib) mkIf mkOption mkEnableOption mkPackageOption optional;
+  cfg = config.services.swhkd;
+in {
+  options.services.swhkd = {
+    enable = mkEnableOption (lib.mdDoc "simple wayland hotkey daemon");
+    package = mkPackageOption pkgs "swhkd" {};
+    installManpages = mkOption {
+      type = lib.types.bool;
+      default = true;
+      description = "install swhkd manual pages";
+    };
+    swhkdrc = mkOption {
+      type = lib.types.str;
+      default = "";
+      description = "contents of the system-wide swhkdrc file";
+    };
+  };
+
+  config = mkIf cfg.enable {
+    environment.systemPackages =
+      [ cfg.package.bin cfg.package.out ] ++ optional cfg.installManpages cfg.package.man;
+    environment.etc."swhkd/swhkdrc".text = cfg.swhkdrc;
+    security.polkit.enable = true;
+  };
+}
--- a/pkgs/by-name/sw/swhkd/package.nix
+++ b/pkgs/by-name/sw/swhkd/package.nix
@@ -1,0 +1,68 @@
+{ lib
+, rustPlatform
+, fetchurl
+, fetchFromGitHub
+, pkg-config
+, scdoc
+, libgcc
+, systemd
+}:
+rustPlatform.buildRustPackage rec {
+    pname = "swhkd";
+    version = "1.2.1-unstable-2024-04-06";
+
+    # split-output derivation, since there's a fair amount of associated data for
+    # pkexec and such.
+    outputs = [ "bin" "man" "out" ];
+
+    src = fetchFromGitHub {
+        owner = "waycrate";
+        repo = "swhkd";
+        # build from master, since the 1.2.1 makefile is unsutible for packaging
+        rev = "f8519a54900d72492a6c036b32e472c108d44dbf";
+        hash = "sha256-zyGyZOG8gAtsRkzSRH1M777fPv1wudbVsBrSTJ5CBnY=";
+      };
+
+    nativeBuildInputs = [
+      scdoc
+      pkg-config
+    ];
+
+    # the makefile tries to set the ownership of a file to root.
+    # this will fail, but files are owned by root anyways.
+    postPatch = "sed -ie 's/-o root//' Makefile";
+
+    buildPhase = ''
+      runHook preBuild
+      make build
+      runHook postBuild
+    '';
+
+    installPhase = ''
+      runHook preInstall
+
+      mkdir -p $out $bin $man/share
+      make DESTDIR=$out MAN1_DIR=/share/man/man1 MAN5_DIR=/share/man/man5 TARGET_DIR=/bin install
+      mv $out/bin $bin/bin
+      mv $out/share/man $man/share/man
+      mv $out/usr/share/polkit-1 $out/share/polkit-1
+      rm -r $out/etc $out/usr
+
+      runHook postInstall
+    '';
+
+    cargoHash = "sha256-d/61hdyooYuqfOSTUcxVUJVhG98uexgPk7h6N1ptIgQ=";
+
+    buildInputs = [
+      systemd
+      libgcc
+    ];
+
+    meta = with lib; {
+        description = "A drop-in replacement for sxhkd that works with wayland";
+        homepage = "https://github.com/waycrate/swhkd";
+        changelog = "https://github.com/waycrate/swhkd/blob/${src.rev}/CHANGELOG.md";
+        license = licenses.bsd2;
+        maintainers = with maintainers; [ binarycat ];
+      };
+  }
--- a/pkgs/by-name/sw/swhkd/package.nix
+++ b/pkgs/by-name/sw/swhkd/package.nix
@@ -16,12 +16,12 @@ rustPlatform.buildRustPackage rec {
     outputs = [ "bin" "man" "out" ];
 
     src = fetchFromGitHub {
-        owner = "waycrate";
-        repo = "swhkd";
-        # build from master, since the 1.2.1 makefile is unsutible for packaging
-        rev = "f8519a54900d72492a6c036b32e472c108d44dbf";
-        hash = "sha256-zyGyZOG8gAtsRkzSRH1M777fPv1wudbVsBrSTJ5CBnY=";
-      };
+      owner = "waycrate";
+      repo = "swhkd";
+      # build from master, since the 1.2.1 makefile is unsutible for packaging
+      rev = "f8519a54900d72492a6c036b32e472c108d44dbf";
+      hash = "sha256-zyGyZOG8gAtsRkzSRH1M777fPv1wudbVsBrSTJ5CBnY=";
+    };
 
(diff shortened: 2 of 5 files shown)


pkgs/by-name/sw/swhkd/package.nix
```diff
@@ -0,0 +1,68 @@
...
+    # pkexec and such.
+    outputs = [ "bin" "man" "out" ];
+
+    src = fetchFromGitHub {
+        owner = "waycrate";
+        repo = "swhkd";
+        # build from master, since the 1.2.1 makefile is unsutible for packaging
+        rev = "f8519a54900d72492a6c036b32e472c108d44dbf";
+        hash = "sha256-zyGyZOG8gAtsRkzSRH1M777fPv1wudbVsBrSTJ5CBnY=";
+      };
```
SuperSandro2000, line 24:
```suggestion
      owner = "waycrate";
      repo = "swhkd";
      # build from master, since the 1.2.1 makefile is unsutible for packaging
      rev = "f8519a54900d72492a6c036b32e472c108d44dbf";
      hash = "sha256-zyGyZOG8gAtsRkzSRH1M777fPv1wudbVsBrSTJ5CBnY=";
    };
```

pkgs/by-name/sw/swhkd/package.nix
```diff
@@ -0,0 +1,68 @@
...
+      libgcc
+    ];
+
+    meta = with lib; {
+        description = "A drop-in replacement for sxhkd that works with wayland";
+        homepage = "https://github.com/waycrate/swhkd";
+        changelog = "https://github.com/waycrate/swhkd/blob/${src.rev}/CHANGELOG.md";
+        license = licenses.bsd2;
+        maintainers = with maintainers; [ binarycat ];
+      };
```
SuperSandro2000, line 67:
```suggestion
      description = "A drop-in replacement for sxhkd that works with wayland";
      homepage = "https://github.com/waycrate/swhkd";
      changelog = "https://github.com/waycrate/swhkd/blob/${src.rev}/CHANGELOG.md";
      license = licenses.bsd2;
      maintainers = with maintainers; [ binarycat ];
    };
```

pkgs/by-name/sw/swhkd/package.nix
```diff
@@ -0,0 +1,68 @@
...
+      rm -r $out/etc $out/usr
+
+      runHook postInstall
+    '';
+
+    cargoHash = "sha256-d/61hdyooYuqfOSTUcxVUJVhG98uexgPk7h6N1ptIgQ=";
+
+    buildInputs = [
+      systemd
+      libgcc
```
SuperSandro2000, line 58: It looks odd that we need libgcc here.

pkgs/by-name/sw/swhkd/package.nix
```diff
@@ -0,0 +1,68 @@
...
+    '';
+
+    installPhase = ''
+      runHook preInstall
+
+      mkdir -p $out $bin $man/share
+      make DESTDIR=$out MAN1_DIR=/share/man/man1 MAN5_DIR=/share/man/man5 TARGET_DIR=/bin install
+      mv $out/bin $bin/bin
+      mv $out/share/man $man/share/man
+      mv $out/usr/share/polkit-1 $out/share/polkit-1
```
SuperSandro2000, line 48: To move things between outputs, please use moveToOutput. Also I think manpages should be moved automatically in fixupPhase

pkgs/by-name/sw/swhkd/package.nix
```diff
@@ -0,0 +1,68 @@
...
+      pkg-config
+    ];
+
+    # the makefile tries to set the ownership of a file to root.
+    # this will fail, but files are owned by root anyways.
+    postPatch = "sed -ie 's/-o root//' Makefile";
+
+    buildPhase = ''
+      runHook preBuild
+      make build
```
SuperSandro2000, line 37: This could easily fail with how we use buildRustPackage. We should probably call ``./scripts/build-polkit-policy.sh`` by hand in postBuild

pkgs/by-name/sw/swhkd/package.nix
```diff
@@ -0,0 +1,68 @@
...
+      };
+
+    nativeBuildInputs = [
+      scdoc
+      pkg-config
+    ];
+
+    # the makefile tries to set the ownership of a file to root.
+    # this will fail, but files are owned by root anyways.
+    postPatch = "sed -ie 's/-o root//' Makefile";
```
SuperSandro2000, line 33:
```suggestion
    postPatch = ''
      sed -ie 's/-o root//' Makefile
    '';
```

pkgs/by-name/sw/swhkd/package.nix
```diff
@@ -0,0 +1,68 @@
...
+, libgcc
+, systemd
+}:
+rustPlatform.buildRustPackage rec {
+    pname = "swhkd";
+    version = "1.2.1-unstable-2024-04-06";
+
+    # split-output derivation, since there's a fair amount of associated data for
+    # pkexec and such.
+    outputs = [ "bin" "man" "out" ];
```
SuperSandro2000, line 16: but don't we need always all that data? Then it would be easier to keep everything in out.

nixos/modules/services/wayland/swhkd.nix
```diff
@@ -0,0 +1,28 @@
+{ config, pkgs, lib, ... }:
+
+let
+  inherit (lib) mkIf mkOption mkEnableOption mkPackageOption optional;
+  cfg = config.services.swhkd;
+in {
+  options.services.swhkd = {
+    enable = mkEnableOption (lib.mdDoc "simple wayland hotkey daemon");
```
SuperSandro2000, line 8:
```suggestion
    enable = mkEnableOption "simple wayland hotkey daemon";
```

nixos/modules/services/wayland/swhkd.nix
```diff
@@ -0,0 +1,28 @@
...
+    package = mkPackageOption pkgs "swhkd" {};
+    installManpages = mkOption {
+      type = lib.types.bool;
+      default = true;
+      description = "install swhkd manual pages";
+    };
```
SuperSandro2000, line 14:
```suggestion

```
 man pages are being disabled by setting documentation.man.enable to false

nixos/modules/services/wayland/swhkd.nix
```diff
@@ -0,0 +1,28 @@
...
+    swhkdrc = mkOption {
+      type = lib.types.str;
+      default = "";
+      description = "contents of the system-wide swhkdrc file";
+    };
+  };
+
+  config = mkIf cfg.enable {
+    environment.systemPackages =
+      [ cfg.package.bin cfg.package.out ] ++ optional cfg.installManpages cfg.package.man;
```
SuperSandro2000, line 24:
```suggestion
    environment.systemPackages = [ cfg.package ];
```

nixos/modules/services/wayland/swhkd.nix
```diff
@@ -0,0 +1,28 @@
...
+  options.services.swhkd = {
+    enable = mkEnableOption (lib.mdDoc "simple wayland hotkey daemon");
+    package = mkPackageOption pkgs "swhkd" {};
+    installManpages = mkOption {
+      type = lib.types.bool;
+      default = true;
+      description = "install swhkd manual pages";
+    };
+    swhkdrc = mkOption {
+      type = lib.types.str;
```
SuperSandro2000, line 16:
```suggestion
      type = lib.types.lines;
```
we probably want to merge that, don't we?

nixos/modules/services/wayland/swhkd.nix
```diff
@@ -0,0 +1,28 @@
...
+      default = "";
+      description = "contents of the system-wide swhkdrc file";
```
SuperSandro2000, line 18:
```suggestion
      description = "contents of the system-wide swhkdrc file. See {manpage}`swhkd(5) for more details on the config format.";
```

SuperSandro2000 (changes requested):

pkgs/by-name/sw/swhkd/package.nix
lolbinarycat: considering the previous PR got two approvals while not generating any of that data, it would seem not.

pkgs/by-name/sw/swhkd/package.nix
lolbinarycat: moveToOutput is seemingly undocumented, and i think relying on implicit behavior would make the derivation harder to understand.

pkgs/by-name/sw/swhkd/package.nix
lolbinarycat: rust does use libgcc to unwind the stack.  at worst this is mildly redundant.

nixos/modules/services/wayland/swhkd.nix
lolbinarycat: this is incorrect, `out` is not in `meta.outputsToInstall`.

lolbinarycat:
@Shinyzenith one thing i don't quite understand is the role of polkit in swhkd.  it seems like some sort of setuid wrapper, as is obviously required due to wayland's lack of a global keybind system.  however, it seems to be doing something extra beyond just changing users, as simply running `swhkd` as root will lead to an error mentioning pkexec.

setuid security generally doesn't work as well on nixos, due to the nature of the nix store.  i think the most ideomatic way to handle this in nixos would be simply using a systemd system service.

nixos pkexec seems to always prompt for a password unless run as root, so for now i'm just testing the whole module by running `pkexec swhkd` as root.

Shinyzenith:
> @Shinyzenith one thing i don't quite understand is the role of polkit in swhkd. it seems like some sort of setuid wrapper, as is obviously required due to wayland's lack of a global keybind system. however, it seems to be doing something extra beyond just changing users, as simply running `swhkd` as root will lead to an error mentioning pkexec.

We are using polkit to retain some user related information such as invoking UID to avoid some CVE's which previously plagued us.

Just FYI as part of our GSoC program, we do have a student who has been looking into ways to better our privilege escalation system while making the process much easier for users. I totally get your concerns 🤝.

> setuid security generally doesn't work as well on nixos, due to the nature of the nix store. i think the most ideomatic way to handle this in nixos would be simply using a systemd system service.

I wasn't aware of this.. I usually try to keep my code init system independent however I'd be more than happy to add some conditional build flags to enable this for you.

> nixos pkexec seems to always prompt for a password unless run as root, so for now i'm just testing the whole module by running `pkexec swhkd` as root.

Yeah this does not sound good. I would hold off on the PR in this case -- the entire role of polkicykit file is so users can spawn the daemon in their window managers rc files without the need for a pkexec prompt while retaining invoking uid info -- Forcing this interaction in root space defeats the point.

lolbinarycat:
> Just FYI as part of our GSoC program, we do have a student who has been looking into ways to better our privilege escalation system while making the process much easier for users. I totally get your concerns 🤝.

what about opening all the device files as root, opening the socket, then using `setuid()` to drop all filesystem permissions?  since permissions are only checked when opening files, not when writing to them, that seems like it would work fairly well, as long as none of the libraries you depend on want to create tempfiles or something similar.

> I wasn't aware of this.. I usually try to keep my code init system independent however I'd be more than happy to add some conditional build flags to enable this for you.

it shouldn't be necessary to make the actual rust code do anything else, just writing a simple service file (i suppose i could actually just do this in the nixos module)

Shinyzenith:
> what about opening all the device files as root, opening the socket, then using `setuid()` to drop all filesystem permissions? since permissions are only checked when opening files, not when writing to them, that seems like it would work fairly well, as long as none of the libraries you depend on want to create tempfiles or something similar.

This could work ( needs testing obviously ) but the doubt remains of which user to de-escalate to...we need to know the invoking UID to avoid some nasty CVE's.

lolbinarycat:
@Shinyzenith why not just leave that up to the invoking user via a command line flag? you could even make that flag default to PKEXEC_UID, so existing setups will keep working.

pkgs/by-name/sw/swhkd/package.nix
SuperSandro2000: That's the standard way to move files from different outputs and emulating that through other means is discouraged, especially if some functionality is missing https://github.com/NixOS/nixpkgs/blob/master/pkgs/build-support/setup-hooks/multiple-outputs.sh#L105-L149

There is maybe some documentation missing, but that doesn't allow disregarding functions from stdenv.

pkgs/by-name/sw/swhkd/package.nix
SuperSandro2000: True but we require quite some extra code because of the split outputs and removing them would make the module quite a bit simpler and we don't loose much.

Zhaith-Izaliel:
Is this PR still being worked on ? I tried reproducing the derivation on my config and it doesn't build swhkd anymore due to somehow missing `libudev` during build (when it's supposed to be in `systemd` however).

pkgs/by-name/sw/swhkd/package.nix
ivandimitrov8080: I think what he meant here is that libgcc goes into the runtime deps instead of the build deps for this derivation @lolbinarycat

heyzec:
swhkd now no longer requires `pkexec` after this https://github.com/waycrate/swhkd/pull/270, so this PR needs to be updated

PR #306273, opened 2024-04-23
uhd: fix for python module failing to install

systemofapwne:
## Description of changes

This PR fixes the python module failing to install by patching the upstream build process.

I kept the enablePythonApi flag off however, to respect established uhd package default options.

Detail of this fix:
- It avoids upstreams build process to use pythons setuptools directly by omitting the virtual-env check.
- It rewrites upstreams derived installation path for the python module (/nix/store/<hash>-uhd/../<hash>-python-3.XY-env/lib/python-3.XY/site-packages -> /nix/store/<hash>-uhd/lib/python-3.XY/site-packages) to avoid permission problems.

## Things done

Only tested on x86_64-linux
- Built on platform(s)
  - [X] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/applications/radio/uhd/python-api.patch
+++ b/pkgs/applications/radio/uhd/python-api.patch
@@ -1,0 +1,20 @@
+diff --git a/host/python/CMakeLists.txt b/host/python/CMakeLists.txt
+index fdf612747..a67631bc0 100644
+--- a/host/python/CMakeLists.txt
++++ b/host/python/CMakeLists.txt
+@@ -174,6 +174,7 @@ add_custom_command(OUTPUT ${TIMESTAMP_FILE}
+   DEPENDS ${PYUHD_FILES})
+ 
+ add_custom_target(pyuhd_library ALL DEPENDS ${TIMESTAMP_FILE} pyuhd)
++set(HAVE_PYTHON_VIRTUALENV FALSE)
+ if(HAVE_PYTHON_VIRTUALENV)
+   message(STATUS "Python virtual environment detected -- Ignoring UHD_PYTHON_DIR.")
+     # In virtualenvs, let setuptools do its thing
+@@ -192,6 +193,7 @@ else()
+             OUTPUT_VARIABLE UHD_PYTHON_DIR OUTPUT_STRIP_TRAILING_WHITESPACE
+         )
+     endif(NOT DEFINED UHD_PYTHON_DIR)
++    string(REGEX REPLACE "^\\.\\./[^/]+/" "" UHD_PYTHON_DIR ${UHD_PYTHON_DIR})
+     file(TO_CMAKE_PATH ${UHD_PYTHON_DIR} UHD_PYTHON_DIR)
+ 
+     message(STATUS "Utilizing the python install directory: ${CMAKE_INSTALL_PREFIX}/${UHD_PYTHON_DIR}")
--- a/pkgs/applications/radio/uhd/default.nix
+++ b/pkgs/applications/radio/uhd/default.nix
@@ -8,12 +8,6 @@
 , boost
 , ncurses
 , enableCApi ? true
-# Although we handle the Python API's dependencies in pythonEnvArg, this
-# feature is currently disabled as upstream attempts to run `python setup.py
-# install` by itself, and it fails because the Python's environment's prefix is
-# not a writable directly. Adding support for this feature would require using
-# python's pypa/build nad pypa/install hooks directly, and currently it is hard
-# to do that because it all happens after a long buildPhase of the C API.
 , enablePythonApi ? false
 , python3
 , buildPackages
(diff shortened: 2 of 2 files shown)


hexchen:
@ofborg eval

bjornfor (approved):
LGTM. Didn't test.

pkgs/applications/radio/uhd/python-api.patch
```diff
@@ -0,0 +1,20 @@
+diff --git a/host/python/CMakeLists.txt b/host/python/CMakeLists.txt
+index fdf612747..a67631bc0 100644
+--- a/host/python/CMakeLists.txt
++++ b/host/python/CMakeLists.txt
+@@ -174,6 +174,7 @@ add_custom_command(OUTPUT ${TIMESTAMP_FILE}
+   DEPENDS ${PYUHD_FILES})
+ 
+ add_custom_target(pyuhd_library ALL DEPENDS ${TIMESTAMP_FILE} pyuhd)
++set(HAVE_PYTHON_VIRTUALENV FALSE)
```
doronbehar, line 9: Perhaps instead of this, and the `string(REGEX REPLACE`, we can modify the command upstream uses here:

https://github.com/EttusResearch/uhd/blob/50fa3baa2e11ea3b30d5a7e397558e9ae76d8b00/host/python/CMakeLists.txt#L179-L180

? What do you think? Perhaps upstream will also accept such a change..

doronbehar (changes requested):
Hello @systemofapwne ! Thanks for debugging this issue, and attempting to solve it. Also, welcome to Nixpkgs.

> It rewrites upstreams derived installation path for the python module (`/nix/store/-uhd/../-python-3.XY-env/lib/python-3.XY/site-packages` -> `/nix/store/-uhd/lib/python-3.XY/site-packages`) to avoid permission problems.

I read upstream's CMake files to get some context for your patch, and I think that now, since we have a better understanding of how they work, we can suggest to them a fix that will be acceptable by them as well.

My main comment is below. In general, I dislike using such patches, as they are easy to become inapplicable when versions are bumped. I also read upstream's source code for their `master` branch, vs the `v4.6.0.0` tag, and I'm afraid your patch will become a maintenance burden on the next uhd release, or the releases afterwards.

Additionally, I think that if this `enablePythonApi` feature will be properly fixed, we can set it to `true` by default, so that we will be forced to test that it works on every release of `uhd`.

systemofapwne:
> Hello @systemofapwne ! Thanks for debugging this issue, and attempting to solve it. Also, welcome to Nixpkgs.
> 
> > It rewrites upstreams derived installation path for the python module (`/nix/store/-uhd/../-python-3.XY-env/lib/python-3.XY/site-packages` -> `/nix/store/-uhd/lib/python-3.XY/site-packages`) to avoid permission problems.
> 
> I read upstream's CMake files to get some context for your patch, and I think that now, since we have a better understanding of how they work, we can suggest to them a fix that will be acceptable by them as well.
> 
> My main comment is below. In general, I dislike using such patches, as they are easy to become inapplicable when versions are bumped. I also read upstream's source code for their `master` branch, vs the `v4.6.0.0` tag, and I'm afraid your patch will become a maintenance burden on the next uhd release, or the releases afterwards.
> 
> Additionally, I think that if this `enablePythonApi` feature will be properly fixed, we can set it to `true` by default, so that we will be forced to test that it works on every release of `uhd`.

I have absolutely the same view on this as you, since I noticed the same change between master and the uhd-4.6 branch. A change of upstream is most welcome. E.g. by allowing an override for the VENV detection by a build parameter.

To the regexp: I think this mostly became necessary due to this packages structure of handling the python env: https://github.com/NixOS/nixpkgs/blob/ac83fcb6eea91e193d73bc1ea57cd3c85f34fa8c/pkgs/applications/radio/uhd/default.nix#L44 . For UHD 3.5, this regexp was not necessary (https://github.com/NixOS/nixpkgs/blob/nixos-21.11/pkgs/applications/radio/uhd/3.5.nix).

Little sidenote: UHD 3.5 (nixpkgs 21.11) was properly installing the uhd python bindungs due to the VENV check failing upstream and the python-env of this package being differently defined. That together lead to my patch.

pkgs/applications/radio/uhd/python-api.patch
systemofapwne: Do you have a suggestion on how to utilize upstreams direct call to setuptools for this package here? I am rather fresh to nix and still learning about the details.

doronbehar:
OK so I investigated this a bit further, and I found a solution already supported by upstream's cmake option `UHD_PYTHON_DIR`. I opened #307435 as an alternative to this PR. I also checked that `gnuradio{,3_{8,9}}{,Minimal}`, which all depend on `uhdMinimal` (that doesn't install the Python API), still build with this change. Your approval that it produces a desired layout is welcome.

PR #306283, opened 2024-04-23
haskellPackages.mkDerivation: Replace testTarget with testTargets

TeofilC:
## Description of changes

testTarget takes a space separated list of test suites to run. We introduce a new testTargets argument that instead takes a list, and add a backwards compatibility shim.

cc @sternenseemann  you mentioned you were interested in a change like this.

I've tested building a couple of packages and it seems to evaluate to the same thing. 

## Things done

- Built on platform(s)
  - [X] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/development/haskell-modules/configuration-nix.nix
+++ b/pkgs/development/haskell-modules/configuration-nix.nix
@@ -75,7 +75,7 @@ self: super: builtins.intersectAttrs super {
       chmod +x "$out/bin/haskell-language-server"
       '';
     testToolDepends = [ self.cabal-install pkgs.git ];
-    testTarget = "func-test"; # wrapper test accesses internet
+    testTargets = [ "func-test" ]; # wrapper test accesses internet
     preCheck = ''
       export PATH=$PATH:$PWD/dist/build/haskell-language-server:$PWD/dist/build/haskell-language-server-wrapper
       export HOME=$TMPDIR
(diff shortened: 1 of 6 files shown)


maralorn:
I like this change.

Is there a way to actually throw a warning if someone explicitly sets the old value?

TeofilC:
> Is there a way to actually throw a warning if someone explicitly sets the old value?

This sounds good. I'll make it emit a warning (using `lib.warn`) if a user tries to use the old setting.

TeofilC:
Sorry for this taking so long. I've now added a warning @maralorn. 
It looks like `cabal2nix` has some hard-coded overrides that always generate `testTarget` attrs for some packages, so we'd have to change that to avoid the deprecation message with something like https://github.com/NixOS/cabal2nix/pull/646

pkgs/development/haskell-modules/configuration-nix.nix
```diff
@@ -701,7 +701,7 @@ self: super: builtins.intersectAttrs super {
 
   # Not running the "example" test because it requires a binary from lsps test
   # suite which is not part of the output of lsp.
-  lsp-test = overrideCabal (old: { testTarget = "tests func-test"; }) super.lsp-test;
+  lsp-test = overrideCabal (old: { testTargets = [ "tests func-test" ]; }) super.lsp-test;
```
sternenseemann, line 704:
```suggestion
  lsp-test = overrideCabal (old: { testTargets = [ "tests" "func-test" ]; }) super.lsp-test;
```

pkgs/development/haskell-modules/generic-builder.nix
```diff
@@ -408,6 +410,10 @@ let
       exec "$@"
     '';
 
+  testTargetsString =
+    lib.warnIf (testTarget != "") "haskell-modules/generic-builder.testTarget is deprecated. Use testTargets instead"
```
sternenseemann, line 414:
```suggestion
    lib.warnIf (testTarget != "") "haskellPackages.mkDerivation: testTarget is deprecated. Use testTargets instead"
```

pkgs/development/haskell-modules/generic-builder.nix
```diff
@@ -408,6 +410,10 @@ let
...
+    (lib.concatStringsSep " " testTargets);
```
sternenseemann, line 415: In a follow up change, we could change this to `lib.escapeShellArgs` (which of course causes a rebuild).

sternenseemann (commented):
This looks great, thanks!

- Needs a changelog entry for NixOS 25.05
- Please add a `haskellPackages.mkDerivation: ` prefix to the commit message.

sternenseemann (approved):

sternenseemann:
https://github.com/NixOS/nixpkgs/pull/381187 implements a compatible interface for 24.11 so expressions can work with both unstable/25.05 and 24.11.

PR #306292, opened 2024-04-23
mediathekview: build from source

jensbin:
## Description of changes

- Rewrote to build from source
- Added .desktop item
- Fix dependencies (`libnotify`, `libXxf86vm`, `libglvnd`)
- Adding missing java parameters

## Things done

Built and execution successfully and is usable.

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin

---


pkgs/applications/video/mediathekview/default.nix
```diff
@@ -1,49 +1,93 @@
...
+    hash = "sha256-wVISldd5kixrCVatqI+ZUamh23Kul1QaTk+TDk55NMI=";
   };
 
+  mvnParameters = toString [
+    "-Dorg.slf4j.simpleLogger.log.org.apache.maven.cli.transfer.Slf4jMavenTransferListener=warn"
+    "-Dmaven.test.skip=true"
+    {
+      x86_64-linux = "-Plinux_64bit";
+      aarch64-linux = "-Plinux_arm_64bit";
+    }.${stdenv.hostPlatform.system}
```
Moredread, line 28: I guess this would fail on other platforms.  We should change meta.platforms to those two as well. Do you have an idea why the hash isn't the same? Are there sharedibs that are included into the jar?

pkgs/applications/video/mediathekview/default.nix
```diff
@@ -1,49 +1,93 @@
...
-    makeWrapper ${jre}/bin/java $out/bin/MediathekView \
-      --add-flags "-jar $out/lib/MediathekView.jar" \
-      --suffix LD_LIBRARY_PATH : "${libraryPath}"
+    install -m644 /build/source/target/MediathekView.jar $out/lib
+    cp ${src}/res/MediathekView.svg $out/share/icons/hicolor/512x512/apps/de.mediathekview.MediathekView.svg
 
-    makeWrapper ${jre}/bin/java $out/bin/MediathekView_ipv4 \
-      --add-flags "-Djava.net.preferIPv4Stack=true -jar $out/lib/MediathekView.jar" \
-      --suffix LD_LIBRARY_PATH : "${libraryPath}"
+    makeShellWrapper ${jdk}/bin/java $out/bin/mediathek \
```
Moredread, line 55: Upstream this is MediathekView now. I think we should change it as well.

pkgs/applications/video/mediathekview/default.nix
```diff
@@ -1,49 +1,93 @@
...
+      --add-flags "--enable-preview" \
+      --add-flags "-Dfile.encoding=UTF-8" \
+      --add-flags "-XX:+UseShenandoahGC" \
+      --add-flags "-XX:ShenandoahGCHeuristics=compact" \
+      --add-flags "-XX:+UseStringDeduplication" \
+      --add-flags "-XX:MaxRAMPercentage=50.0" \
+      --add-flags "--add-opens java.desktop/sun.awt.X11=ALL-UNNAMED" \
+      --add-flags "-DexternalUpdateCheck" \
```
Moredread, line 63: Does this enable or disable update checks in the app itself?

Moredread (changes requested):

pkgs/applications/video/mediathekview/default.nix
jensbin: This is by default set in appimage → https://github.com/mediathekview/MediathekView/blob/master/scripte/appimage.sh.

"Check whether or not we are using Install4j's external update mechanism." In the UI mentions with this parameter set, that an external updater is used.

pkgs/applications/video/mediathekview/default.nix
jensbin: Do you mean to include the mediathek_ipv4? This is not part of the appimage instructions, but mentioned in the troubleshoot guide. My intention was to be close to appimage creation and if somebody runs into IPv6 issues to use overwrite. But I am open to suggestions.

pkgs/applications/video/mediathekview/default.nix
jensbin: tbh, I haven't had a deeper look into this and I am not sure how to handle this properly for different architectures. Any suggestion welcome.

Is it acceptable to change:
```
-    platforms = lib.platforms.linux;
+    platforms = [ "x86_64-linux" "aarch64-linux" ];

```

Or completely remove `aarch64-linux`?

avdv (approved):

PR #306320, opened 2024-04-23
stdenv: make checkDependencyList tolerate null deps

yaitskov:
Hi,

Looks like a bug or deadcode i.e.
The null check should got first must be dropped at all because `isDerivation` is not null safe.


FliegendeWurst (commented):
LGTM. Please make sure your commit message follows CONTRIBUTING.md

philiptaron (approved):

RossComputerGuy (approved):

Issue #271307, opened 2023-11-30
`lib.fileset` should have a way to filter directories


infinisil:
## Problem

Sometimes one needs to filter out directories with a specific name throughout the source directory.

For example, you might have a structure like
```
- foo
  - target
- bar
  - baz
    - target
  - qux
    - target
```

And you want to remove all of the `target` directories.

Currently, the `lib.fileset` library does not have any function to make this work. To do that currently you need to work around it with either:
- ```nix
  lib.fileset.fromSource (lib.sources.cleanSourceWith {
    src = ./.
    filter = pathString: type:
      # untested..
      type != "directory" || baseNameOf pathString != "target";
  }
  ```
- Some recursive `readDir` and `lib.fileset.unions` I don't want to write out

## Requirements

1. It must not recurse into filtered-out directories for performance reasons
2. It must be intuitive to use

## Potential solutions

### Directory name predicate

```nix
directoryFilter (dir: dir.name == "target") ./.
```

This creates a file set containing all files within directories named "target". E.g. `./foo/target/some/file.txt` would be included because one of `foo`, `target` and `some` satisfies the predicate. Does it satisfy the requirements?

1. :heavy_check_mark: It doesn't need to recurse into the filtered out directories, but the selection needs to be inverted to actually filter out the ones with `target`:
   ```nix
   difference ./. (directoryFilter (dir:
     dir.name == "target"
   ) ./.)
   ```
2. :x: This is _not_ intuitive, because one might expect an inverted predicate to return the complement set:
   ```nix
   directoryFilter (dir:
     dir.name != "target"
   ) ./.
   ```

   But this doesn't work: This would include `./foo/target/some/file.txt`, since `foo` satisfies the predicate.

### Give the `fileFilter` predicate access to the file's directories

Like already described in https://github.com/NixOS/nixpkgs/issues/269504. This would allow filtering certain directories:

```nix
fileFilter (file:
  # Only include files whose directory components don't have `target`
  ! elem "target" file.dirComponents
) ./.
```

This way `./foo/target/some/file.nix` would not be included, because `target` is in `./foo/target/some`. Does it satisfy the requirements?

1. :x: No, this does have to recurse for every file, because the predicate needs files. So this would be very slow.
2. :heavy_check_mark: Yes this is intuitive. One can also invert the predicate and get the expected result.

### Does a solution satisfying both requirements exist?

I don't know yet

JRMurr:
I would love to have a way to do this!

[...]

Issue #269504, opened 2023-11-23
`lib.fileset.fileFilter` should allow filtering by file subpath/components


infinisil:
## Issue description
Currently the predicate of [`lib.fileset.fileFilter`](function-library-lib.fileset.fileFilter) only gets `name`, `type` (and [soon `hasExt`](https://github.com/NixOS/nixpkgs/pull/266362)).

There's no way to filter by the subpath/components of the file, something like this:
```nix
lib.fileset.fileFilter (file:
  lib.lists.hasSuffix [ "foo" "bar" ] file.components
) ./.
```

(`hasSuffix` currently also doesn't exist)

Some ideas:
- `file.components :: [ String ]`: File components, always at least one element for the file's name
- `file.subpath :: String`: The file's [subpath](https://nixos.org/manual/nixpkgs/unstable/#function-library-lib.path.subpath.isValid)
- `file.dirComponents :: [ String ]`: Directory components, can be empty for files in the root level
- `file.dirSubpath :: String`: Directory subpath, can be `./.` for files in the root level

This feature request is the motivation behind https://github.com/NixOS/nixpkgs/pull/267384.

roberth:
This would be the most powerful function, and users catch on to that. If the function can do anything, so can they. However, their sets will require more computation at evaluation time, and there's a good chance they'll be harder to read.

Meanwhile, I haven't seen a concrete use case that needs such a function. Do you have a real world motivating use case for this?
("parity with filterSource" would be one or a few degrees short of _real world_.)

infinisil:
Indeed, I can't think of a good motivation for this right now. In fact I can only think of a reason why we shouldn't do this:

People might try this:
```nix
lib.fileset.fileFilter (file:
  head file.components == ".git"
) ./.
```

But that's really bad for performance.

infinisil:
I do think we need something else for `cleanSource` parity. I think it should be something like `lib.fileset.directoryFilter`, which allows recursion to stop early.

roberth:
An alternative solution that is equally powerful is for the users to do some `readDir` and returning file sets from that.
Might not feel "integrated", but seems quite efficient, because it's constructive rather than O(nm) steps where `m` is the complexity of the predicate reaching its return value.

roberth:
Not sure about pings tbh :sweat_smile:. I feel like it's a nicely rounded library now, so I'm just say a lot of "do we really need this" and how about just this + that.
[...]

PR #306371, opened 2024-04-23
Make parent dir available in fileset.fileFilter

andrewhamon:
## Description of changes

Currently, fileFilter only allows filtering based based on a files base name + file type.

This is a bit limiting if you want to include files based on the name of the directory they reside in.

I bumped in to this when using fileFilter to make a minimal set of files to make Cargo happy in a Rust workspace - specifically, I needed to pull in any `.rs` file that lived under a `bin/`, `examples/`, or `benches/` folder before Cargo would stop complaining. (Note that I am trying to avoid pulling in all rust files - I'm working in a large Rust workspace but trying to package a single small member of the workspace).

This change allows me to write filters like such as this:

```
file.hasExt "rs" && (pkgs.lib.hasSuffix  "bin" (toString file.path))
```

## Open questions
- should `dir` be a Path (as implemented it is) or should it be converted to stringified path relative to the filter root?

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### Benchmark results
There is a preexisting benchmark script. I ran `./benchmark.sh master` and got the following results:

```
Benchmarking expression toSource { root = ./.; fileset = ./.; }
Mean CPU time 0.179607 (σ = 0.00527474) for 10 runs is 100.03230315958318% (σ = 3.669259422572901%) of the old value 0.179549 (σ = 0.00394593)
Statistic .gc.totalBytes (19776544) is 100.0001% (+32) of the old value 19776512
1 stats differ between the current tree and master

Benchmarking expression toSource { root = ./.; fileset = unions (mapAttrsToList (name: value: ./. + "/${name}") (builtins.readDir ./.)); }
Mean CPU time 0.105958 (σ = 0.00257414) for 10 runs is 101.95621842675006% (σ = 2.7418005135301344%) of the old value 0.103925 (σ = 0.00119843)
0 stats differ between the current tree and master
```

This doesn't seem to have a large impact on performance if not using `dir`

### DIFF
--- a/lib/fileset/default.nix
+++ b/lib/fileset/default.nix
@@ -668,6 +668,8 @@ in {
 
       - `name` (String): The name of the file
 
+      - `dir` (Path): The path of the directory containing the file.
+
       - `type` (String, one of `"regular"`, `"symlink"` or `"unknown"`): The type of the file.
         This matches result of calling [`builtins.readFileType`](https://nixos.org/manual/nix/stable/language/builtins.html#builtins-readFileType) on the file's path.
 
--- a/lib/fileset/internal.nix
+++ b/lib/fileset/internal.nix
@@ -825,11 +825,11 @@ rec {
   _fileFilter = predicate: root:
     let
       # Check the predicate for a single file
-      # Type: String -> String -> filesetTree
-      fromFile = name: type:
+      # Type: Path -> String -> String -> filesetTree
+      fromFile = dir: name: type:
         if
           predicate {
-            inherit name type;
+            inherit dir name type;
             hasExt = ext: hasSuffix ".${ext}" name;
 
             # To ensure forwards compatibility with more arguments being added in the future,
@@ -848,7 +848,7 @@ rec {
           if type == "directory" then
             fromDir (path + "/${name}")
           else
-            fromFile name type
+            fromFile path name type
         ) (readDir path);
 
       rootType = pathType root;


infinisil (changes requested):
Thanks for the PR!

Unfortunately I already explored this option in https://github.com/NixOS/nixpkgs/issues/269504, where we figured that this shouldn't be done like that because it leads to poorer performance than expected. In your example, it would have to recurse throughout all `bin` directories, even though they won't be included in the result.

A more general tracking issue for this use case is https://github.com/NixOS/nixpkgs/issues/271307, for which I think this is the best way forward: https://github.com/NixOS/nixpkgs/issues/271307#issuecomment-1839940023

andrewhamon:
Re: performance, it seems like passing the path unaltered (as I do in this PR) is close to zero cost for any use cases that don't do anything with `dir`.

> In your example, it would have to recurse throughout all bin directories, even though they won't be included in the result.

I'm a little confused... doesn't fileFilter already recurse through all directories? Since it passes every file to the predicate?

andrewhamon:
After reading https://github.com/NixOS/nixpkgs/issues/269504 a little more closely you gave this example:

```
lib.fileset.fileFilter (file:
  head file.components == ".git"
) ./.
```

I see how that is very wasteful, but that is a very different situation that my example - I am already in a situation where it makes sense to use fileFilter (thus paying the performance penalty).

This is the full example of how I'm using this:

```nix
  fileSetForCrate = fileset.toSource {
    root = ../..;
    fileset = fileset.unions [
      (fileset.fileFilter (file:
      (file.name) == "Cargo.toml"
        || file.name == "Cargo.lock"
        || file.name == "main.rs"
        || file.name == "lib.rs"
        || file.name == "build.rs"
        || (file.hasExt "rs" && (pkgs.lib.hasSuffix  "bin" (toString file.path)))
        || (file.hasExt "rs" && (pkgs.lib.hasSuffix  "examples" (toString file.path)))
        || (file.hasExt "rs" && (pkgs.lib.hasSuffix  "benches" (toString file.path)))
      ) ../..)
    ];
  };
  ```

**Comment by andrewhamon** [comment]
I think I agree that your preferred solution (https://github.com/NixOS/nixpkgs/issues/271307#issuecomment-1839940023) is a better fit. To be clear, for my needs it will still have the same performance, but that API seems harder to accidentally misuse.

**Comment by infinisil** [comment]
Performance is a bit more tricky because the fileset combinators are lazy. So e.g. while `fileFilter` by itself always has to recurse throughout the entire directory, if you combine it with e.g. `intersection`, it won't have to. Here's a fun example that creates a file set which contains literally _every_ Nix file on your system, but then limits it to just a single directory:

```
nix-repl> :l <nixpkgs/lib>
Added 428 variables.

nix-repl> :a fileset
Added 13 variables.

nix-repl> nixFiles = fileFilter (file: file.hasExt "nix") /.

nix-repl> trace (intersection nixFiles ~/src/nixpkgs/main/lib/fileset)
trace: /home/tweagysil/src/nixpkgs/main/lib/fileset
trace: - default.nix (regular)
trace: - internal.nix (regular)
trace: - mock-splitRoot.nix (regular)
«lambda @ /nix/store/fmb2gkj0jyd4nzi7xzkp5k1c9pz45v5s-source/lib/fileset/default.nix:225:8»
```

The same will happen with the proposed `withoutDirectories` function.

infinisil:
> Re: performance, it seems like passing the path unaltered (as I do in this PR) is close to zero cost for any use cases that don't do anything with `dir`. The benchmarks seem to confirm this.

This would be problematic because it would allow filesets to depend on the absolute location of the local directory. It's a very intentional design choice to make sure this cannot happen to avoid problems relating to that. This would be good to add to the [design goals](https://github.com/NixOS/nixpkgs/blob/master/lib/fileset/README.md#goals) of the library.

So we cannot pass absolute paths because of that, but relative paths would be fine (even if they're a bit slower). Efficiency by itself is notably not a design goal of the library :)

PR #306407, opened 2024-04-23
ffado: Remove Qt dependency

jtojnar:
## Description of changes

This slims up `pipewire` build closure.

The full package is available as `ffado-mixer`.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/os-specific/linux/ffado/default.nix
+++ b/pkgs/os-specific/linux/ffado/default.nix
@@ -7,7 +7,6 @@
   dbus_cplusplus,
   desktop-file-utils,
   fetchurl,
-  fetchpatch,
   glibmm,
   libavc1394,
   libconfig,
@@ -31,7 +30,7 @@ let
 in
 mkDerivation rec {
   pname = "ffado";
-  version = "2.4.8";
+  version = "2.4.9";
 
   outputs = [
     "out"
@@ -41,7 +40,7 @@ mkDerivation rec {
 
   src = fetchurl {
     url = "http://www.ffado.org/files/libffado-${version}.tgz";
-    hash = "sha256-0iFXYyGctOoHCdc232Ud80/wV81tiS7ItiS0uLKyq2Y=";
+    hash = "sha256-xELFL60Ryv1VE7tOhGyFHxAchIT4karFRe0ZDo/U0Q8=";
   };
 
   prePatch = ''
@@ -49,20 +48,7 @@ mkDerivation rec {
       --replace /lib/modules/ "/run/booted-system/kernel-modules/lib/modules/"
   '';
 
-  patches = [
-    # fix installing metainfo file
-    ./fix-build.patch
-
-    (fetchpatch {
-      name = "musl.patch";
-      url = "http://subversion.ffado.org/changeset?format=diff&new=2846&old=2845";
-      stripLen = 2;
-      hash = "sha256-iWeYnb5J69Uvo1lftc7MWg7WrLa+CGZyOwJPOe8/PKg=";
-    })
-  ];
-
   nativeBuildInputs = [
-    desktop-file-utils
     (scons.override { python3Packages = python311.pkgs; })
     pkg-config
     which
@@ -76,7 +62,6 @@ mkDerivation rec {
     "DEBUG=False"
     "ENABLE_ALL=True"
     "BUILD_TESTS=True"
-    "WILL_DEAL_WITH_XDG_MYSELF=True"
     "BUILD_MIXER=True"
     "UDEVDIR=${placeholder "out"}/lib/udev/rules.d"
     "PYPKGDIR=${placeholder "out"}/${python.sitePackages}"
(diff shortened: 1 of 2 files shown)


SuperSandro2000:
The new nixfmt is making reviewing this commit by commit actually hard...

jtojnar:
Hiding whitespace changes will help.

SuperSandro2000 (commented):
The approach looks reasonable to me and the mixer is probably not required to be there for pipewire but I can't properly review the changes on mobile with the formatting changes all in between...

> Hiding whitespace changes (or using difftastic) will help.

That's not possible in the mobile app 🙃

pkgs/os-specific/linux/ffado/default.nix
```diff
@@ -63,23 +71,26 @@ mkDerivation rec {
...
+    ];
 
   prefixKey = "PREFIX=";
   sconsFlags = [
     "DEBUG=False"
     "ENABLE_ALL=True"
     "BUILD_TESTS=True"
     "WILL_DEAL_WITH_XDG_MYSELF=True"
-    "BUILD_MIXER=True"
+    "BUILD_MIXER=${if withMixer then "True" else "False"}"
```
jtojnar, line 93: Ideally, we would only build the mixer in the `mixer` package but it is not easy to prevent libffado from building.

And the mixer will try to launch `ffado-dbus-server` so we would need to be able to specify path to it from the libffado prefix.

uninsane:
i've confirmed on my end that pipewire works fine without the mixer using a [similar](https://github.com/uninsane/nixpkgs/compare/master...uninsane:nixpkgs:wip-ffado-cross?expand=1), but not identical patch. the changes here look reasonable, and once this lands i'll rebase what i've got there to make ffado cross compile. then with Mindavi's appstream cross PR i think we'll have the whole suite of basic `wrapGAppsHook4` packages cross compiling out-of-the-box :)

nitpick: since it includes all of the base `ffado`, should it be named `ffado-with-mixer` instead of `ffado-mixer`?

jtojnar:
Will probably wait for 2.4.9, which will include all the patches used.

avnik:
Any chances we merge it?  Curiously, we have `gtk3/4` depends on qt* via pipewire and ffado. Would be nice to cut this

Mic92:
Doesn't look good for ffado's upstream development. Their subversion is down: http://subversion.ffado.org/

Mic92:
I cherry-picked the first two commits into its own pull request: https://github.com/NixOS/nixpkgs/pull/318999

Mic92:
@K900 you added ffado support to pipewire in the first place to nixpkgs presumably because of https://gitlab.freedesktop.org/pipewire/pipewire/-/commit/3c44634dd80975f13def3c09a1f05aaf9ad08110
But do we actually need this feature? Have we any confirmed firewire audio user?

K900:
There are definitely users, but Pipewire only requires the library, so moving the mixer to another package is good IMO.

Mic92:
> There are definitely users, but Pipewire only requires the library, so moving the mixer to another package is good IMO.

Ok. Can you than point me to one that would also test changes on real hardware? Otherwise I cannot be sure that I am not breaking anything.

K900:
Not off the top of my head, sorry :( I think if Pipewire builds, it should be fine.

jtojnar:
> Doesn't look good for ffado's upstream development. Their subversion is down: [http://subversion.ffado.org](http://subversion.ffado.org/)

I [reported it to upstream](https://sourceforge.net/p/ffado/mailman/ffado-devel/thread/20240425221620.213112-1-jtojnar%40gmail.com/#msg58764904) a while ago and I vaguely remember upstream acknowledging that but cannot find the reply.

New patches are still being accepted over the mailing list just fine.

> Do we still want to wait?

I do not think there is any hurry, the metainfo is not really used by anything so waiting for 2.4.9 should be fine.

The only critical thing is https://ffado.org/posts/ffado-2.4.8-tarball_fix/ 🙀 I still need to write an e-mail telling upstream off.

Though, I just noticed the hash update was reverted so it is probably still broken: https://github.com/NixOS/nixpkgs/pull/317396

pkgs/os-specific/linux/ffado/default.nix
```diff
@@ -41,43 +45,33 @@ mkDerivation rec {
...
+  nativeBuildInputs =
+    [
+      desktop-file-utils
+      scons
+      pkg-config
+      which
+    ]
+    ++ lib.optionals withMixer [
+      python
+      python3.pkgs.pyqt5
```
K900, line 65: Is it actually necessary here and not just in withPackages?

K900 (approved):
LGTM other than one small nit

pkgs/os-specific/linux/ffado/default.nix
jtojnar: Looks like I added it in 8b1816ec5ab6b2ef5eb9c218322ff4742d5d0e99. I believe this is for `pyuic5` program but not sure.

SuperSandro2000 (approved):
Didn't test but lgtm

pkgs/os-specific/linux/ffado/default.nix
```diff
@@ -41,43 +45,32 @@ mkDerivation rec {
...
-    (scons.override { python3Packages = python311.pkgs; })
-    pkg-config
-    which
-    python
-    python.pkgs.pyqt5
-    wrapQtAppsHook
-  ];
+  nativeBuildInputs =
+    [
+      (scons.override { python3Packages = python311.pkgs; })
```
K900, line 58: Do we still need the scons pin?

jtojnar: Yes, the `SConstruct` still pulls in `import distutils.sysconfig`, which was removed in Python 3.12.

K900: Cool. Can you add a comment for that so whoever updates this next knows to remove it?

PR #306430, opened 2024-04-24
Revert "workflows: add concurrency group to check-by-name workflow"

willbush:
This reverts commit 75600cde6bb25b0e04d2a10d0fb9aaf975e1da81.

## Description of changes

@infinisil I don't think it's working as expected.

- https://github.com/NixOS/nixpkgs/pull/306427
- https://github.com/NixOS/nixpkgs/pull/306418

---

### DIFF
--- a/.github/workflows/check-by-name.yml
+++ b/.github/workflows/check-by-name.yml
@@ -16,13 +16,6 @@ on:
     # so it shouldn't be a problem
     types: [opened, synchronize, reopened, edited]
 
-# Create a check-by-name concurrency group based on the branch name. if a new
-# commit is pushed to the main branch while a previous run is still in progress,
-# the previous run will be cancelled and the new one will start.
-concurrency:
-  group: check-by-name-${{ github.ref }}
-  cancel-in-progress: true
-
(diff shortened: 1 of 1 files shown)


willbush:
Hmm maybe it does run after all queued checks complete. Still looking for an example of it being canceled with no other pending checks.

infinisil:
Huh weird, yeah let's revert for now, thanks for double-checking this!

infinisil:
Ahh: 
> Canceling since a higher priority waiting request for 'check-by-name-refs/heads/master' exists

I guess we should use `${{ github.event.pull_request.number }}` instead.

willbush:
> [Ahh](https://github.com/NixOS/nixpkgs/actions/runs/8808862830):
> 
> > Canceling since a higher priority waiting request for 'check-by-name-refs/heads/master' exists
> 
> I guess we should use `${{ github.event.pull_request.number }}` instead.

Ah that makes sense!

infinisil:
Btw I restarted all the cancelled builds manually (were only a couple), I think it should all be resolved now

PR #306457, opened 2024-04-24
nixos/forgejo: Small cleanups and service capability changes

### LINKED ISSUE #306205: Forgejo module changes openssh settings even when using built in ssh server
### Describe the bug

The option to enable the built in ssh server is `services.forgejo.settings.server.START_SSH_SERVER`, however the module checks for `services.forgejo.settings.START_SSH_SERVER`

Here's the line that does the check:
https://github.com/NixOS/nixpkgs/blob/6143fc5eeb9c4f00163267708e26191d1e918932/nixos/modules/services/misc/forgejo.nix#L631

### Steps To Reproduce
1. Enable forgejo and set `services.forgejo.settings.server.START_SSH_SERVER` to true
2. Set `services.openssh.settings.AcceptEnv` to anything

### Expected behavior
There's no conflicting definition, and my system config builds just fine

### Screenshots
image

### Notify maintainers

@emilylange @bendlas 

### Metadata

```console
 - system: `"x86_64-linux"`
 - host os: `Linux 6.1.86, NixOS, 23.11 (Tapir), 23.11.20240422.ea97158`
 - multi-user?: `yes`
 - sandbox: `yes`
 - channels(root): `""`
 - nixpkgs: `/nix/store/g861759ghxxwvyfdbv17xf3iahgm8rcb-source`

```

---

pyrox0:
Also resolves #306205 and adds myself as a maintainer to the module.

Note: This won't eval cleanly until #306349, which fixes my maintainer attribute name, is merged. However, I've checked and it does work properly with that merged into the tree.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [x] Tested, as applicable:

---

### DIFF
--- a/nixos/modules/services/misc/forgejo.nix
+++ b/nixos/modules/services/misc/forgejo.nix
@@ -660,11 +660,9 @@ in
       environment = {
         USER = cfg.user;
         HOME = cfg.stateDir;
-        # `GITEA_` prefix until https://codeberg.org/forgejo/forgejo/issues/497
-        # is resolved.
-        GITEA_WORK_DIR = cfg.stateDir;
-        GITEA_CUSTOM = cfg.customDir;
-      } // lib.listToAttrs (map (e: lib.nameValuePair e.env "%d/${e.env}") secrets);
+        FORGEJO_WORK_DIR = cfg.stateDir;
+        FORGEJO_CUSTOM = cfg.customDir;
+      };
     };
 
     services.openssh.settings.AcceptEnv = mkIf (!cfg.settings.START_SSH_SERVER or false) "GIT_PROTOCOL";
@@ -690,10 +688,8 @@ in
       environment = {
         USER = cfg.user;
         HOME = cfg.stateDir;
-        # `GITEA_` prefix until https://codeberg.org/forgejo/forgejo/issues/497
-        # is resolved.
-        GITEA_WORK_DIR = cfg.stateDir;
-        GITEA_CUSTOM = cfg.customDir;
+        FORGEJO_WORK_DIR = cfg.stateDir;
+        FORGEJO_CUSTOM = cfg.customDir;
       };
 
       serviceConfig = {
--- a/nixos/modules/services/misc/forgejo.nix
+++ b/nixos/modules/services/misc/forgejo.nix
@@ -725,5 +725,5 @@ in
   };
 
   meta.doc = ./forgejo.md;
-  meta.maintainers = with lib.maintainers; [ bendlas emilylange ];
+  meta.maintainers = with lib.maintainers; [ bendlas emilylange pyrox0 ];
 }
--- a/nixos/modules/services/misc/forgejo.nix
+++ b/nixos/modules/services/misc/forgejo.nix
@@ -325,6 +325,18 @@ in
                 description = "Disable external SSH feature.";
               };
 
+              START_SSH_SERVER = (mkEnableOption "START_SSH_SERVER" //{
+                description = ''
+                  When enabled, use Forgejo's built-in SSH server instead of the
+                  hosts' existing SSH server ({option}`services.openssh.enable`).
+                  ::: {.note}
+                  This likely requires manually changing the value of either
+                  {option}`services.openssh.ports` or
+                  {option}`services.forgejo.settings.server.SSH_LISTEN_PORT` so they don't collide.
+                  :::
+                '';
+              });
+
               SSH_PORT = mkOption {
                 type = types.port;
                 default = 22;
@@ -655,6 +667,10 @@ in
         SystemCallFilter = [ "~@cpu-emulation @debug @keyring @mount @obsolete @privileged @setuid" "setrlimit" ];
         # cfg.secrets
         LoadCredential = map (e: "${e.env}:${e.path}") secrets;
+      } // lib.optionalAttrs (cfg.settings.server.SSH_PORT < 1024) {
+        AmbientCapabilities = "CAP_NET_BIND_SERVICE";
+        CapabilityBoundingSet = "CAP_NET_BIND_SERVICE";
+        PrivateUsers = false;
       };
 
       environment = {
@@ -665,7 +681,7 @@ in
       };
     };
 
-    services.openssh.settings.AcceptEnv = mkIf (!cfg.settings.START_SSH_SERVER or false) "GIT_PROTOCOL";
+    services.openssh.settings.AcceptEnv = mkIf (!cfg.settings.server.START_SSH_SERVER or false) "GIT_PROTOCOL";
 
     users.users = mkIf (cfg.user == "forgejo") {
       forgejo = {
--- a/pkgs/by-name/fo/forgejo/package.nix
+++ b/pkgs/by-name/fo/forgejo/package.nix
@@ -119,7 +119,7 @@ buildGoModule rec {
     homepage = "https://forgejo.org";
     changelog = "https://codeberg.org/forgejo/forgejo/releases/tag/${src.rev}";
     license = lib.licenses.mit;
-    maintainers = with lib.maintainers; [ emilylange urandom bendlas adamcstephens ];
+    maintainers = with lib.maintainers; [ emilylange urandom bendlas adamcstephens pyrox0 ];
     broken = stdenv.isDarwin;
     mainProgram = "gitea";
   };


nixos/modules/services/misc/forgejo.nix
```diff
@@ -621,14 +627,10 @@ in
       environment = {
         USER = cfg.user;
         HOME = cfg.stateDir;
-        # `GITEA_` prefix until https://codeberg.org/forgejo/forgejo/issues/497
-        # is resolved.
-        GITEA_WORK_DIR = cfg.stateDir;
-        GITEA_CUSTOM = cfg.customDir;
```
emilylange, line 627: We still need those.

The comment was referencing the `GITEA_` prefix instead of `FORGEJO_`.
You are right that https://codeberg.org/forgejo/forgejo/issues/497 has been resolved for quite some months now.

But this just means we can now use
```nix
        FORGEJO_WORK_DIR = cfg.stateDir;
        FORGEJO_CUSTOM = cfg.customDir;
```

nixos/modules/services/misc/forgejo.nix
```diff
@@ -651,10 +653,6 @@ in
...
```
emilylange, line 657: Same as above^

nixos/modules/services/misc/forgejo.nix
```diff
@@ -621,14 +627,10 @@ in
...
       };
     };
 
-    services.openssh.settings.AcceptEnv = mkIf (!cfg.settings.START_SSH_SERVER or false) "GIT_PROTOCOL";
+    services.openssh.settings.AcceptEnv = mkIf (!cfg.settings.server.START_SSH_SERVER or false) "GIT_PROTOCOL";
```
emilylange, line 633: I think I would prefer it, if you could more this into another commit.
So two commits in total.

One for the `GITEA_`/`FORGEJO_` prefix thingy, and one for fixing #306205.

nixos/modules/services/misc/forgejo.nix
```diff
@@ -322,6 +322,12 @@ in
                 description = "Disable external SSH feature.";
               };
 
+              START_SSH_SERVER = mkOption {
+                type = types.bool;
+                default = false;
+                description = "When enabled, use the built-in SSH server.";
+              };
```
emilylange, line 329: I hesitate whether we need that option right now.
`services.forgejo.settings` is already a free-form option, so we technically don't need it.

The only benefit would be self-documentation. Making users aware that one does not have to use the hosts' openssh server.

What is your rationale behind adding it? :)

For the integrated SSH server to be able to bind to port < 1024, we need to add a few lines to `systemd.services.forgejo.serviceConfig`.

Namely:
```diff
diff --git a/nixos/modules/services/misc/forgejo.nix b/nixos/modules/services/misc/forgejo.nix
index babed2d5acd4..c8aada68e4a5 100644
--- a/nixos/modules/services/misc/forgejo.nix
+++ b/nixos/modules/services/misc/forgejo.nix
@@ -616,6 +616,10 @@ in
         # System Call Filtering
         SystemCallArchitectures = "native";
         SystemCallFilter = [ "~@cpu-emulation @debug @keyring @mount @obsolete @privileged @setuid" "setrlimit" ];
+      } // lib.optionalAttrs (cfg.settings.server.SSH_PORT < 1024) {
+        AmbientCapabilities = "CAP_NET_BIND_SERVICE";
+        CapabilityBoundingSet = "CAP_NET_BIND_SERVICE";
+        PrivateUsers = false;
       };
 
       environment = {
```

emilylange (changes requested):

emilylange:
Regarding https://github.com/NixOS/nixpkgs/pull/306457#discussion_r1578085673:

See result of `nix-build -A nixosTests.forgejo.sqlite3` (or CI)
@ofborg build nixosTests.forgejo.sqlite3

nixos/modules/services/misc/forgejo.nix
pyrox0: Fixed this!

nixos/modules/services/misc/forgejo.nix
pyrox0: My rationale behind adding the option to the documentation is that it's utilized inside the module itself. Since it's part of that, I feel that it's a good idea to also have it in the documentation, in case someone doesn't look at the upstream's configuration documentation. While that's not something you should do, it's bound to happen, and so I feel that the module should have documentation for at least every option that it uses internally.

nixos/modules/services/misc/forgejo.nix
pyrox0: These have been split out

nixos/modules/services/misc/forgejo.nix
```diff
@@ -322,6 +322,12 @@ in
                 description = "Disable external SSH feature.";
               };
 
+              START_SSH_SERVER = mkOption {
+                type = types.bool;
+                default = false;
+                description = "When enabled, use the built-in SSH server.";
```
emilylange, line 328: Nit, non-blocking, but how about
```suggestion
                description = ''
                  When enabled, use Forgejo's built-in SSH server instead of the
                  hosts' existing SSH server ({option}`services.openssh.enable`).

                  ::: {.note}
                  This likely requires manually changing the value of either
                  {option}`services.openssh.ports` or
                  {option}`services.forgejo.settings.server.SSH_LISTEN_PORT` so they don't collide.
                  :::
                '';
```

fyi you can build the manual locally using `nix-build nixos/release.nix -A manual.x86_64-linux`

nixos/modules/services/misc/forgejo.nix
```diff
@@ -616,19 +622,21 @@ in
         # System Call Filtering
         SystemCallArchitectures = "native";
         SystemCallFilter = [ "~@cpu-emulation @debug @keyring @mount @obsolete @privileged @setuid" "setrlimit" ];
+      } // lib.optionalAttrs (cfg.settings.server.SSH_PORT < 1024) {
+        AmbientCapabilities = "CAP_NET_BIND_SERVICE";
+        CapabilityBoundingSet = "CAP_NET_BIND_SERVICE";
+        PrivateUsers = false;
```
emilylange, line 628: Hmm there are actually a lot of config options that can trigger a bind to a port below 1024.

For SSH, `server.SSH_LISTEN_PORT` defaults to whatever `server.SSH_PORT` is (which is UI-only) and only applies when `server.START_SSH_SERVER` is true and `server.DISABLE_SSH` is false (I think).

And then there is also `server.ENABLE_ACME`, `server.PORT_TO_REDIRECT`, `server.REDIRECT_OTHER_PORT` and `server.HTTP_PORT` and whatnot with their own dependency chain.

I changed my mind and believe we should simply set `CAP_NET_BIND_SERVICE` unconditionally.
The security implications are tiny, and saves us from re-implementing upstream's config logic for all those special cases (which, to add insult to insure, may very well change with an update any time).

```diff
diff --git a/nixos/modules/services/misc/forgejo.nix b/nixos/modules/services/misc/forgejo.nix
index 34668cb2e16c..1f85a3865a8e 100644
--- a/nixos/modules/services/misc/forgejo.nix
+++ b/nixos/modules/services/misc/forgejo.nix
@@ -596,7 +596,8 @@ in
         ReadWritePaths = [ cfg.customDir cfg.dump.backupDir cfg.repositoryRoot cfg.stateDir cfg.lfs.contentDir ];
         UMask = "0027";
         # Capabilities
-        CapabilityBoundingSet = "";
+        AmbientCapabilities = [ "CAP_NET_BIND_SERVICE" ];
+        CapabilityBoundingSet = [ "CAP_NET_BIND_SERVICE" ];
         # Security
         NoNewPrivileges = true;
         # Sandboxing
@@ -604,7 +605,7 @@ in
         ProtectHome = true;
         PrivateTmp = true;
         PrivateDevices = true;
-        PrivateUsers = true;
+        PrivateUsers = false; # true is incompatible with our set of AmbientCapabilities
         ProtectHostname = true;
         ProtectClock = true;
         ProtectKernelTunables = true;
@@ -622,10 +623,6 @@ in
         # System Call Filtering
         SystemCallArchitectures = "native";
         SystemCallFilter = [ "~@cpu-emulation @debug @keyring @mount @obsolete @privileged @setuid" "setrlimit" ];
-      } // lib.optionalAttrs (cfg.settings.server.SSH_PORT < 1024) {
-        AmbientCapabilities = "CAP_NET_BIND_SERVICE";
-        CapabilityBoundingSet = "CAP_NET_BIND_SERVICE";
-        PrivateUsers = false;
       };
 
       environment = {
```

nixos/modules/services/misc/forgejo.nix
```diff
@@ -674,5 +680,5 @@ in
   };
 
   meta.doc = ./forgejo.md;
-  meta.maintainers = with lib.maintainers; [ bendlas emilylange ];
+  meta.maintainers = with lib.maintainers; [ bendlas emilylange pyrox0 ];
```
emilylange, line 683: Feel free to add yourself to `pkgs.forgejo` as well if you want (you asked in https://github.com/NixOS/nixpkgs/pull/306341#issuecomment-2073922593).

But please do so in another commit.
Atomic git commits and all.
You can use the same commit for `nixos/forgejo` and `forgejo` though, imho :)

emilylange (changes requested):

nixos/modules/services/misc/forgejo.nix
pyrox0: Updated this!

nixos/modules/services/misc/forgejo.nix
pyrox0: Updated this description as well

nixos/modules/services/misc/forgejo.nix
pyrox0: Removed from the first commit and made another commit adding myself to both.

pyrox0:
Had to rebase, tests should pass now.

nixos/modules/services/misc/forgejo.nix
```diff
@@ -590,15 +604,16 @@ in
         ReadWritePaths = [ cfg.customDir cfg.dump.backupDir cfg.repositoryRoot cfg.stateDir cfg.lfs.contentDir ];
         UMask = "0027";
         # Capabilities
-        CapabilityBoundingSet = "";
+        AmbientCapabilities = [ "CAP_NET_BIND_SERVICE" ];
+        CapabilityBoundingSet = [ "CAP_NET_BIND_SERVICE" ];
```
bendlas, line 608: Should we maybe only  get this, when `port < 1024`?

Often deployments will be behind nginx and on unprivileged ports.

bendlas (dismissed):
Has the capabilities change been mixed in accidentally?

Feels like this should be a separate PR, because it's security-relevant and hasn't been mentioned in this PR before.

nixos/modules/services/misc/forgejo.nix
pyrox0: This was discussed previously, see https://github.com/NixOS/nixpkgs/pull/306457#discussion_r1592984395. I think it's a reasonable change, and from reading the documention on capabilities(https://www.man7.org/linux/man-pages/man7/capabilities.7.html) I don't see any risk in giving this permission, as the most that a malicious actor could potentially do with it is bind to a low port. I don't see this as a security issue to have with ports > 1024.

nixos/modules/services/misc/forgejo.nix
```diff
@@ -322,6 +322,20 @@ in
                 description = "Disable external SSH feature.";
               };
 
+              START_SSH_SERVER = mkOption {
+                type = types.bool;
+                default = false;
```
adamcstephens, line 327: This should be `mkEnableOption` since it's a simple default false boolean.

pyrox0: Updated

emilylange: Last force-push/change broke eval.

The call signature of `mkOption` is different from `mkEnableOption`.

If you need help with that, let me know :)

nixos/modules/services/misc/forgejo.nix
bendlas: > This was discussed previously

Oh, sorry, I was out of the loop for that.

> What do you think

Note that I'm not necessarily against just granting the CAP, as previously asked by @emilylange. I agree that the security implications are minimal (though not zero), since CI runners are part of a different service def.

What I don't like is tacking a change like this on a PR of unrelated description. But since all the current maintainers are here now, I won't block it.

I like the separate option. It has the advantage of being explicit over an automatic detection of `port < 1024` and breaking a security boundary should always be as explicit as conveniently possible. Maybe `services.forgejo.allowBindPrivilegedPort`?

bendlas: With the option, we could even implement a helpful error message, to set the option if the bind port is privileged.

nixos/modules/services/misc/forgejo.nix
pyrox0: Yup, forgot to run a test build of the manual. Just resolved that error.

nixos/modules/services/misc/forgejo.nix
pyrox0: > > This was discussed previously
> 
> Oh, sorry, I was out of the loop for that.
> 
> > What do you think
> 
> Note that I'm not necessarily against just granting the CAP, as previously asked by @emilylange. I agree that the security implications are minimal (though not zero), since CI runners are part of a different service def.
> 
> What I don't like is tacking a change like this on a PR of unrelated description. But since all the current maintainers are here now, I won't block it.
> 
> I like the separate option. It has the advantage of being explicit over an automatic detection of `port < 1024` and breaking a security boundary should always be as explicit as conveniently possible. Maybe `services.forgejo.allowBindPrivilegedPort`?

So again, as mentioned, I think that a separate option would be good, but we would need to always make sure that we check all options taking a port to ensure all of them are above 1024 before surfacing this warning. As also discussed, this is a very small security decrease, and going from my experience, it doesn't really change the attack surface much at all. Therefore, I think unconditionally enabling this is fine, especially if you are running the service behind a reverse proxy, as is generally suggested.

bendlas: Yeah, so as I said, I was out of the loop for the initial round of feedback and I'm fine going with what @emilylange already asked of you and I don't think this is worth endlessly nitpicking. Since I already expressed my preference (the capability change in a separate PR), but indicated that I won't block, I'll resolve this thread now.

For posterity, I'll leave the lengthier response in, since I've already typed it out. Beware: Severe nitpicking ahead

---

> very small security decrease

The reason I've flagged this, is less about the absolute amount of security change (in fact, I think linux could do without privileged network ports), but more about default expectations when dealing with a linux system. For example, I could easily see a firewall accepting all privileged ports by default, which could easily "amplify" the small attack surface.

But yeah, much of this is hypothetical, because the capability would only be given to forgejo, and not the CI runner, which seems much more vulnerable than the main forgejo process.

Still, as I said before, the thing that bugs me here is PR hygiene: The title here is still "Cleanup environment settings", where it's not just a cleanup, but also a capability change.

> Therefore, I think unconditionally enabling this is fine, especially if you are running the service behind a reverse proxy

I'm having a little trouble tracking this implication, sorry: Are we talking about a reverse proxy on a different machine?

Even when running a server without a reverse proxy, I prefer binding to high ports and then just redirecting the appropriate low ports via firewall.

And with a reverse proxy (on the same machine), I try to not bind network ports at all, but `+unix` domain sockets, because those can also resist local attackers.

bendlas (commented):
Retracting my change request, since I'm realizing that this is mainly about using the built-in ssh server. That's a use case that I simply don't have, but happen to know that @emilylange does. So I'll trust her expertise on this.

I'd still like to leave one self-quote from the resolved thread for posterity, because I feel like it meaningfully adds to the conversation.

> Even when running a server without a reverse proxy, I prefer binding to high ports and then just redirecting the appropriate low ports via firewall.

bendlas (commented):
.

nixos/modules/services/misc/forgejo.nix
pyrox0: I think it's a fair nit that the PR title wasn't appropriate, so I've updated it. Thanks for suggesting that, since the old title definitely didn't include the full scope of this PR.

mweinelt:
I'm slightly worried about unconditionally handing out capabilities, that are not strictly required. My deployment currently only binds to a UNIX socket, and I honestly don't want it to be able bind to privileged ports by default.

```
# ss -lpn | grep forgejo
u_str LISTEN 0      128                     /run/forgejo/forgejo.sock 22892                  * 0    users:((".gitea-wrapped",pid=1603,fd=11))    
```

emilylange:
It's a bit unfortunate the CAP discussion/suggestion has been marked as resolved in the meantime.

From what I can tell the majority of Forgejo maintainers leaned towards something like `allowBindPrivilegedPort` (defaulting to `false`) in the end.

I get that it was my suggestion to make it unconditional in the first place.
But I changed my mind in https://github.com/NixOS/nixpkgs/pull/306457#discussion_r1594155106, voicing my support for such option (suggesting the inferior name `allowPortBindBelow1024`).

We could even backport that variant to `24.05` whenever this PR is ready and merged, long after the branch-off tomorrow, since it wouldn't be a breaking change.

Or, slightly more controversial, we could continue to not provide an option for this at all.

While we do not have concrete guidelines for this in nixpkgs, most folks I speak with expect users of unusual, edge-case-y and advanced configurations to be comfortable doing such override in question themselves via `systemd.services.<name>`.

```nix
{ lib, ... }: {
  systemd.services.forgejo.serviceConfig = {
    AmbientCapabilities = lib.mkForce [ "CAP_NET_BIND_SERVICE" ];
    CapabilityBoundingSet = lib.mkForce [ "CAP_NET_BIND_SERVICE" ];
    PrivateUsers = lib.mkForce false;
  };
}
```

---

What if we, as suggested somewhere in one of the many collapsed comments, split the CAP thingy from this PR as it currently stands.

That way, this whole CAP discussion does not hinder the uncontroversial changes from getting merged.

Specifically, the `services.openssh` conditional bug fix (#306205):
```diff
-    services.openssh.settings.AcceptEnv = mkIf (!cfg.settings.START_SSH_SERVER or false) "GIT_PROTOCOL";
+    services.openssh.settings.AcceptEnv = mkIf (!cfg.settings.server.START_SSH_SERVER or false) "GIT_PROTOCOL";
```

@pyrox0 getting added as maintainer and the `GITEA_` -> `FORGEJO_` thingy (noop).

PS: I genuinely haven't thought about port forwarding on the machine itself (what @bendlas suggested). I genuinely like that approach. Might copy that for my stuff.

emilylange:
Are you still interested in working on this, @pyrox0? :)

pyrox0:
I am still interested in working on this. I'll rebase and remove the CAP override so that the rest of this can get merged.

pyrox0:
This should be ready to merge.

bendlas (approved):

emilylange:
@ofborg test forgejo

PR #289750, opened 2024-02-18
filebrowser: 2.23.0 -> 2.28.0, add service module

christoph-heiss:
## Description of changes

Updates the `filebrowser` package from 2.23.0 to 2.28.0 and adds an appropriate service module for it.

Changelog: https://github.com/filebrowser/filebrowser/compare/v2.23.0...v2.28.0

cc @nielsegberts As this package has not been updated in nearly a year, are you still interested?

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
- For non-Linux: Is sandboxing enabled in `nix.conf`? (See [Nix manual](https://nixos.org/manual/nix/stable/command-ref/conf-file.html))
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [x] Tested, as applicable:
  - [NixOS test(s)](https://nixos.org/manual/nixos/unstable/index.html#sec-nixos-tests) (look inside [nixos/tests](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests))
  - and/or [package tests](https://nixos.org/manual/nixpkgs/unstable/#sec-package-tests)
  - or, for functions and "core" functionality, tests in [lib/tests](https://github.com/NixOS/nixpkgs/blob/master/lib/tests) or [pkgs/test](https://github.com/NixOS/nixpkgs/blob/master/pkgs/test)
  - made sure NixOS tests are [linked](https://nixos.org/manual/nixpkgs/unstable/#ssec-nixos-tests-linking) to the relevant packages
- [x] Tested compilation of all packages that depend on this change using `nix-shell -p nixpkgs-review --run "nixpkgs-review rev HEAD"`. Note: all changes have to be committed, also see [nixpkgs-review usage](https://github.com/Mic92/nixpkgs-review#usage)
- [x] Tested basic functionality of all binary files (usually in `./result/bin/`)
- [24.05 Release Notes](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2405.section.md) (or backporting [23.05](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2305.section.md) and [23.11](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2311.section.md) Release notes)
  - [ ] (Package updates) Added a release notes entry if the change is major or breaking
[...]

PR #306615, opened 2024-04-24
Filebrowser service

sdellysse:
## Description of changes

Added service configuration for filebrowser. I'm open to suggestions on how to better structure the options since this application is primarily configured through a mutable single-file database.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [X] Tested, as applicable:

---


maintainers/maintainer-list.nix
```diff
@@ -19038,6 +19038,12 @@
     githubId = 8904314;
     name = "Konstantin Nick";
   };
+  sqlazer = {
+    email = "me@sqlazer.com";
+    github = "sdellysse";
+    githubId = "293035";
```
Shawn8901, line 19044:
```suggestion
    githubId = 293035;
```

sdellysse: thanks!

nixos/modules/services/misc/filebrowser.nix
```diff
@@ -0,0 +1,89 @@
...
+        type = lib.types.bool;
+        default = false;
+        description = "Open ports in firewall for Filebrowser web interface";
+      };
+      package = lib.mkPackageOption pkgs "filebrowser" { };
+      user = lib.mkOption {
+        type = lib.types.str;
+        default = "filebrowser";
+        description = "User under which Filebrowser runs";
+      };
```
MinerSebas, line 37: I would reorder the optins like this.
This way `enable` and `package` are at the top and `user` and `group` are next to each other.
```suggestion
      enable = lib.mkEnableOption "Filebrowser";
      package = lib.mkPackageOption pkgs "filebrowser" { };
      dataDir = lib.mkOption {
        type = lib.types.str;
        default = "/var/lib/filebrowser";
        description = "Directory where Filebrowser stores its data files";
      };
      user = lib.mkOption {
        type = lib.types.str;
        default = "filebrowser";
        description = "User under which Filebrowser runs";
      };
      group = lib.mkOption {
        type = lib.types.str;
        default = "filebrowser";
        description = "Group under which Filebrowser runs";
      };
      listenAddress = lib.mkOption {
        type = lib.types.str;
        default = "0.0.0.0";
        description = "Address that Filebrowser will bind to";
      };
      listenPort = lib.mkOption {
        type = lib.types.port;
        default = 5983;
        description = "Port that Filebrowser will bind to";
      };
      openFirewall = lib.mkOption {
        type = lib.types.bool;
        default = false;
        description = "Open ports in firewall for Filebrowser web interface";
      };
```

sdellysse: sounds good. I usually default to trying to put stuff in lexical order when possible but I'll take the advice :+1:

MinerSebas (changes requested):
Please add the following to the filebrowser package

At the top:
```diff
-{ buildGoModule, buildNpmPackage, fetchFromGitHub, lib }:
+{ buildGoModule, buildNpmPackage, fetchFromGitHub, lib, nixosTests }:
```
And in the lower `buildGoModule`
```diff
   passthru = {
     inherit frontend;
+    tests = {
+      filebrowser = nixosTests.filebrowser;
+    };
   };
```

This would have let the CI catch the other mistakes.

MinerSebas:
After the above changes i can access the the webpage, but not login.
Presumebly the non-standard install doesnt add the bootstrap admin account.
No Problem, i can just confirm this with the `users ls` command

```
$ sudo filebrowser-service-tool.sh users ls
2024/05/11 23:12:29 timeout
```

:-(

h7x4:
Ref #289750, not sure about the differences but maybe the changes can be synced up?

nixos/modules/services/misc/filebrowser.nix
```diff
@@ -0,0 +1,152 @@
...
+        type = lib.types.str;
+      };
+    };
+  };
+
+  config = let
+    cfg = config.services.filebrowser;
+
+    cmd =
+      "${cfg.package}/bin/filebrowser --config='${cfg.dataDir}/config.json' --database='${cfg.dataDir}/database.boltdb'";
```
SuperSandro2000, line 73:
```suggestion
      "${lib.getExe cfg.package} --config='${cfg.dataDir}/config.json' --database='${cfg.dataDir}/database.boltdb'";
```

nixos/modules/services/misc/filebrowser.nix
```diff
@@ -0,0 +1,152 @@
...
+        fi
+
+        ${preamble}
+
+        exec ${cmd} "$@"
+      '')
+    ];
+
+    networking.firewall =
+      lib.mkIf cfg.openFirewall { allowedTCPPorts = [ cfg.settings.port ]; };
```
SuperSandro2000, line 113:
```suggestion
    networking.firewall = lib.mkIf cfg.openFirewall {
      allowedTCPPorts = [ cfg.settings.port ];
    };
```

nixos/modules/services/misc/filebrowser.nix
```diff
@@ -0,0 +1,152 @@
...
+
+          ${cmd} config set --address=${cfg.settings.address} || exit $?
+          ${cmd} config set --port=${
+            builtins.toString cfg.settings.port
+          } || exit $?
+
+          exec ${cmd}
+        '';
+      };
+
```
SuperSandro2000, line 137:
```suggestion
      };
      script = ''
        ${preamble}

        ${cmd} config set --address=${cfg.settings.address} || exit $?
        ${cmd} config set --port=${toString cfg.settings.port} || exit $?

        exec ${cmd}
      '';
```
we don't need the extra wrapper here

undefined-landmark:
Hi, thanks for taking the first steps in settings up this service! I'd like to see a module for this package included in NixOS. The configuration of this application is a bit complex. Some points that I gathered from looking at the docs:
- A basic `config.json` can be used to set a [number of options](https://filebrowser.org/cli/filebrowser). This should be relatively simple to generate from an nix attribute set.
- More complex configuration is available via the `filebrowser config` command. The config can be imported, however the config is not flat and the options in [`filebrowser config set`](https://filebrowser.org/cli/filebrowser-config-set) cannot be translated 1:1 to a json. If we want to allow for all these options I think it's necessary to generate all the relevant `filebrowser config set` commands from an attribute set.
- Users can be imported. Luckily in this case the json is flat and can contain all the options listed [in filebrowser users add](https://filebrowser.org/cli/filebrowser-users-add). We could use the `--overwrite` options of `filebrowser users import`. If the users is already created it will be updated, if not it will be created. The json can be easily generated from an attribute set, with the only remark that it should also include a unique ID. As the `--overwrite` options expect and ID and name.
- We could use the api in our tests to see if signing into an account works. The api is poorly documented, see https://github.com/filebrowser/docs/issues/1, but a simple login can be tested with `curl -Ss localhost:8001/api/login --data-raw '{"username":"admin","password":"admin"}'`.

I could try to implement some of the above points so that we can try to setup a way to configure filebrowser from this module. Any suggestions would be appreciated, as you have probably looked at the configurations possibilities for a lot longer than I did :).

Thanks!

undefined-landmark:
I tried to implement all points, except for the second one, in this PR: https://github.com/sdellysse/nixpkgs/pull/1.

RumanKhelif:
Hello, thank you for your work. I look forward to using this service. What is the status of this PR? Is it possible to merge it?

undefined-landmark:
> Hello, thank you for your work. I look forward to using this service. What is the status of this PR? Is it possible to merge it?

I believe the author is not actively working on this module anymore. 

I have a somewhat finished version in the PR mentioned above. Might create a new PR sometime in the future.

PR #306616, opened 2024-04-24
openjdk22: add updateScript

jlesquembre:
## Description of changes

Adds an update script for openjdk22. I tried to do make the script generic enough so it can be re-used for any openjdk version, but I'll handle that on a following PR if this one gets merged. 
I had to replace the `version` attribute set with a string. The version is updated and validated in the updateScript with Java using  [Runtime.Version](https://docs.oracle.com/en/java/javase/22/docs/api/java.base/java/lang/Runtime.Version.html) 

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/development/compilers/openjdk/22.nix
+++ b/pkgs/development/compilers/openjdk/22.nix
@@ -3,7 +3,6 @@
 , fetchurl
 , fetchpatch
 , fetchFromGitHub
-, bash
 , pkg-config
 , autoconf
 , cpio
@@ -44,27 +43,32 @@
 , gnome_vfs
 , glib
 , GConf
+, writeShellScript
 }:
 
 let
-  version = {
-    feature = "22";
-    interim = "";
-    build = "36";
-  };
+
+  # Java version format:
+  # $FEATURE.$INTERIM.$UPDATE.$PATCH
+  # See
+  # https://openjdk.org/jeps/223
+  # https://docs.oracle.com/en/java/javase/22/docs/api/java.base/java/lang/Runtime.Version.html
+  featureVersion = "22";
+  info = builtins.getAttr featureVersion (lib.importJSON ./info.json);
+  version = info.version;
 
   # when building a headless jdk, also bootstrap it with a headless jdk
   openjdk-bootstrap = openjdk22-bootstrap.override { gtkSupport = !headless; };
 
-  openjdk = stdenv.mkDerivation {
+  openjdk = stdenv.mkDerivation (finalAttrs: {
     pname = "openjdk" + lib.optionalString headless "-headless";
-    version = "${version.feature}${version.interim}+${version.build}";
+    inherit version;
 
     src = fetchFromGitHub {
       owner = "openjdk";
-      repo = "jdk${version.feature}u";
-      rev = "jdk-${version.feature}${version.interim}+${version.build}";
-      hash = "sha256-itjvIedPwJl/l3a2gIVpNMs1zkbrjioVqbCj1Z1nCJE=";
+      repo = info.repo;
+      rev = "jdk-${version}";
+      hash = info.hash;
     };
 
     nativeBuildInputs = [ pkg-config autoconf unzip ensureNewerSourcesForZipFilesHook ];
@@ -144,9 +148,8 @@ let
     # https://openjdk.org/groups/build/doc/building.html
     configureFlags = [
       "--with-boot-jdk=${openjdk-bootstrap.home}"
-      "--with-version-build=${version.build}"
-      "--with-version-opt=nixos"
-      "--with-version-pre="
+      "--with-version-string=${version}"
+      "--with-vendor-version-string=(nix)"
       "--enable-unlimited-crypto"
       "--with-native-debug-symbols=internal"
       "--with-libjpeg=system"
@@ -247,14 +250,28 @@ let
 
     disallowedReferences = [ openjdk-bootstrap ];
 
-    pos = builtins.unsafeGetAttrPos "feature" version;
-    meta = import ./meta.nix lib version.feature;
+    pos = __curPos;
+    meta = import ./meta.nix lib featureVersion;
 
     passthru = {
-      architecture = "";
-      home = "${openjdk}/lib/openjdk";
+      updateScript =
+        let
+          java-json = fetchurl {
+            url = "https://search.maven.org/remotecontent?filepath=org/json/json/20240303/json-20240303.jar";
+            hash = "sha256-PPbNaJLjLitMHDng9S9SSKL1s3ZG/fu3mma0a2GEFO0=";
+          };
+        in
+        writeShellScript "update-java" ''
+          ${finalAttrs.finalPackage}/bin/java \
+            -cp ${java-json} \
+            ${./JavaUpdater.java} \
+            22 pkgs/development/compilers/openjdk/info.json
+        '';
+
+      home = "${finalAttrs.finalPackage}/lib/openjdk";
+
       inherit gtk3;
     };
-  };
+  });
 in
 openjdk
--- a/pkgs/development/compilers/openjdk/JavaUpdater.java
+++ b/pkgs/development/compilers/openjdk/JavaUpdater.java
@@ -1,0 +1,181 @@
+import org.json.JSONArray;
+import org.json.JSONObject;
+
+import java.io.IOException;
+import java.io.PrintWriter;
+import java.net.URI;
+import java.net.http.*;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.*;
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+import java.util.stream.StreamSupport;
+
+public class JavaUpdater {
+
+  record GitHubResult(Optional<String> latestVersion, Optional<String> next) {
+  }
+
+  record JsonInfo(String repo, String version, String hash) {
+    public JsonInfo(JSONObject json) {
+      this(json.getString("repo"), json.getString("version"), json.getString("hash"));
+    }
+
+    public String toJsonString(String featureVersion) {
+      return """
+        \s "%s": {
+        \s   "version": "%s",
+        \s   "repo":    "%s",
+        \s   "hash":    "%s"
+        \s }\
+        """.formatted(featureVersion, version, repo, hash);
+    }
+  }
+
+  // Parses the GitHub Link header
+  public static Optional<String> getNextLink(HttpHeaders headers) {
+    var linkHeader = headers.map().get("Link");
+    if (linkHeader == null || linkHeader.isEmpty()) return null;
+
+    var links = linkHeader.getFirst();
+    var linksRegex = Pattern.compile("<(.+)>;\\s*rel=\"next\"");
+    return Pattern.compile(",")
+      .splitAsStream(links)
+      .map(x -> linksRegex.matcher(x).results()
+        .map(g -> g.group(1))
+        .findFirst()
+      )
+      .filter(Optional::isPresent)
+      .map(Optional::orElseThrow)
+      .findFirst();
+  }
+
+  // HTTP request helper, sets GITHUB_TOKEN if present
+  private static HttpRequest NewGithubRequest(String url) {
+    var token = System.getenv().get("GITHUB_TOKEN");
+    var builder = HttpRequest.newBuilder()
+      .uri(URI.create(url));
+    if (token != null)
+      builder.setHeader("Authorization", "Bearer " + token);
+    return builder.build();
+  }
+
+  private static GitHubResult getLatestTag(String url) {
+    var request = NewGithubRequest(url);
+
+    var response =
+      HttpClient.newHttpClient().sendAsync(request, HttpResponse.BodyHandlers.ofString())
+        .join();
+
+    var json = new JSONArray(response.body());
+
+    Optional<String> version = StreamSupport.stream(json.spliterator(), false)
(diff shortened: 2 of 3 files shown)


Princemachiavelli (changes requested):
The updater script works for me. A quick nix-diff of the PR shows the only change being the update from "22+36" to "22-ga" which is correct. Running the updater today causes an update from "22-ga" to "22.0.1-ga" which appears correct.

If the latest release always ends in `-ga`, perhaps it's better to strip that from the version? eg. `openjdk-22.0.1` instead of `openjdk-22.0.1-ga`.

pkgs/development/compilers/openjdk/22.nix
```diff
@@ -44,217 +43,234 @@
...
-    '';
+    done
+  '';
+
+  disallowedReferences = [ jdk-bootstrap' ];
 
-    disallowedReferences = [ openjdk-bootstrap ];
+  meta = import ./meta.nix lib featureVersion;
 
-    pos = builtins.unsafeGetAttrPos "feature" version;
```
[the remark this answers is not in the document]
jlesquembre: TIL `__curPos`.

Good point, I missed it, re-added :)

jlesquembre:
@Princemachiavelli Thanks for your review. 
Added the `pos` back.
But I'd prefer to keep the `-ga` suffix to clarify which version we're using. Looking at https://github.com/openjdk/jdk22, there are many jdk-22+XX builds. In my opinion, the -ga suffix makes it clearer which specific version we're referring to.

Princemachiavelli (approved):

pkgs/top-level/java-packages.nix
```diff
@@ -226,7 +226,7 @@ in {
       ../development/compilers/openjdk/22.nix
       ../development/compilers/zulu/22.nix
       {
-        openjdk22-bootstrap = temurin-bin.jdk-21;
+        jdk-bootstrap = temurin-bin.jdk-21;
```
Infinidoge, line 229: Is there a reason why this was changed? It breaks the pattern with all of the other JDK builders, so it seems a bit out of place to change it in just 1 place. Same with the other references to the bootstrap JDK.

jlesquembre: Probably I should have done it on a following PR. 
The main reason is to unify the arguments to `mkOpenJdk` and create a more generic function in the future. There is a lot of duplication between the 22.nix, 21.nix, etc. files. 

It also makes it more consistent with the openjfx argument. In my opinion, we should do this:

```nix
    openjdk22 = mkOpenjdk
      {
        jdk-bootstrap = temurin-bin.jdk-21;
        openjfx = openjfx22;
      };

```

or this:

```nix
    openjdk22 = mkOpenjdk
      {
        openjdk22-bootstrap = temurin-bin.jdk-21;
        openjdk22-openjfx = openjfx22;
      };
```

liff:
Would it make sense to add some `passthru.tests` to this? Like `testVersion` and perhaps something that builds and runs a trivial application, just to make sure `javac` and `java` do what they should.

jlesquembre:
@liff My plan was to add a test on a following PR, but since getting this PR merged is taking longer than expected, I went ahead and added a test. I also reformatted the nix file with `nixfmt` (RFC 166)

Infinidoge:
Please split the commit to separate reformatting, adding the update script, and adding the test.

jlesquembre:
@Infinidoge I split it into 2 commits. For the review, please ignore the first one since it only contains formatting changes. I hope this helps reviewers and we can get this merged.
I didn't add an extra commit for the version tests for pragmatic reasons. The test changes are minimal, and having only one commit makes it easier to modify the commit again if more changes are requested.

pkgs/development/compilers/openjdk/JavaGetVersion.java
```diff
@@ -0,0 +1,8 @@
+
```
SuperSandro2000, line 1: Is there a particular reason why the updater is not written in bash or python other than dog fooding?

It makes it a lot harder for general contributions to use it.

pkgs/development/compilers/openjdk/22.nix
```diff
@@ -1,74 +1,89 @@
...
-  };
+
+  # Java version format:
+  # $FEATURE.$INTERIM.$UPDATE.$PATCH
+  # See
+  # https://openjdk.org/jeps/223
+  # https://docs.oracle.com/en/java/javase/22/docs/api/java.base/java/lang/Runtime.Version.html
+
+  featureVersion = "22";
+  info = builtins.getAttr featureVersion (builtins.fromJSON (builtins.readFile ./info.json));
```
SuperSandro2000, line 60: There should be a libs function to read json in one function

SuperSandro2000 (commented):
Formatting of single packages is discouragrd as it creates lots of noise. Please wait for an organized effort.

pkgs/development/compilers/openjdk/JavaGetVersion.java
jlesquembre: Two reasons:
- The update script uses https://docs.oracle.com/en/java/javase/22/docs/api/java.base/java/lang/Runtime.Version.html to parse and compare versions. It makes more sense to reuse that instead of re-implementing that logic on other languages, particularly since the Java version schema could be more complex than expected: https://openjdk.org/jeps/223
- I expect Java developers to be more interested in maintaining the updater script. I think it's fair to write the script in a language more familiar to them.

jlesquembre:
@SuperSandro2000 
> Formatting of single packages is discouragrd as it creates lots of noise. Please wait for an organized effort.

I agree, but in this particular case I think it makes sense. The nix file was like this:

```
let 
  # ...
  openjdk = mkDerivation {...};
in
openjdk
```
and this PR updates it to 

```
let 
  # ...
in
openjdk = mkDerivation (finalAttrs: {...})
```
only with that change, the diff is already quite noisy, so I took that opportunity to reformat it. Even without reformatting it with `nixfmt`, it makes sense to split the PR into two commits to make the real changes clear. That said, we could still meld both commits into one before merging into the master branch.

pkgs/development/compilers/openjdk/22.nix
jlesquembre: right, there is `lib.importJSON`, updated to use that function

jlesquembre:
What can I do to move forward this PR? I'm open to removing the formatting, remove the test, or other elements if it would help with the review process.

My only sticking point is that I believe the update script should remain in Java (or any JDK language) to leverage the [Runtime.Version](https://docs.oracle.com/en/java/javase/22/docs/api/java.base/java/lang/Runtime.Version.html) utilities.

cc @SuperSandro2000 @raboof

pkgs/top-level/java-packages.nix
doronbehar: @jlesquembre regarding this change `jdk-bootstrap` rename, a quick `git grep openjdk22-bootstrap` showed me this:

```
pkgs/development/compilers/openjdk/22.nix:, openjdk22-bootstrap
pkgs/development/compilers/openjdk/22.nix:  openjdk-bootstrap = openjdk22-bootstrap.override { gtkSupport = !headless; };
```

So, won't it break what's in `22.nix` and `21.nix`?

doronbehar (commented):
2 more general question:

- Could we split perhaps the refactoring and the updateScript addition to separate commits?
- Does this commit apply only to `openjdk22`? What about the other `opdnjdk` attributes?

pkgs/top-level/java-packages.nix
jlesquembre: I just undid that change

jlesquembre:
@SuperSandro2000 @Infinidoge @doronbehar I've undone all the unrelated changes (refactoring included) to focus solely on adding the update script. Hopefully, this helps moving forward the PR. I'll address the remaining changes in future PRs.

pkgs/development/compilers/openjdk/22.nix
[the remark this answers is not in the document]
doronbehar: For reference: https://github.com/NixOS/nix/issues/10036

doronbehar (approved):
Diff LGTM. I trust the author to have tested the update script and that he will maintain it properly. I will merge in a week or so if no body else has any comments.

jlesquembre:
thanks @doronbehar 
I did indeed test the script and plan to maintain it. Also, I intentionally didn't bump openjdk22 to the latest version. This way, once it gets merged, the update bot should create a PR.

For future reference, if anyone wants to test it locally, in the root directory of your `nixpkgs` copy run 
```
nix build .#openjdk22.updateScript --print-out-paths
```
 (or `nix-build . --attr openjdk22.updateScript --no-out-link` if you don't use flakes). 

Then you can inspect the generated update script and run it if you want.

pkgs/development/compilers/openjdk/22.nix
```diff
@@ -144,9 +148,7 @@ let
     # https://openjdk.org/groups/build/doc/building.html
     configureFlags = [
       "--with-boot-jdk=${openjdk-bootstrap.home}"
-      "--with-version-build=${version.build}"
-      "--with-version-opt=nixos"
-      "--with-version-pre="
```
Infinidoge, line 149: These should be kept, as they are used for marking the version. 
From [The Temurin build wiki, about versions in build configuration](https://github.com/adoptium/temurin-build/wiki/A-Conversation-about-versions-and-setting-them-in-build-configuration):
> But when you build something that is fit for others to consume, you will still want to provide a build number with --with-build-number, use -with-version-pre="" if you want to say it is an actual release (use e.g. "ea", or just leave blank to make it "internal" otherwise), and in addition set --with-version-opt to control that thing that defaults to "adhoc.$USERNAME.$basedirname".

Infinidoge (changes requested):
Small problem with the versions specified in the build arguments, but outside of that LGTM

pkgs/development/compilers/openjdk/22.nix
jlesquembre: I replaced it with `--with-version-string=${version}`, where the `version` value is the git tag.
Notice that `--with-version-XX=` overwrites the value set by `--with-version-string`.

Currently `java --version` returns:
```
openjdk 22-ga 2024-03-19
OpenJDK Runtime Environment (build 22-ga)
OpenJDK 64-Bit Server VM (build 22-ga, mixed mode, sharing)
```

If we add those 2 flags again (`version-opt` and `version-pre`) the output becomes::
```
openjdk 22 2024-03-19
OpenJDK Runtime Environment (build 22+-nixos)
OpenJDK 64-Bit Server VM (build 22+-nixos, mixed mode, sharing)
```

In my opinion, the first one makes the JDK version much clearer.

Looking at the link you posted, seems that what we should use is `--with-vendor-version-string` , but that option isn't mentioned here: https://openjdk.org/groups/build/doc/building.html

Infinidoge: Hmm, ideally we would still include the vendor information if possible, especially since we do patch OpenJDK for building on Nixpkgs. I do agree that it is much more clear though. Have you tested if `--with-vendor-version-string` works? (Finishing up a course so I haven't done so myself yet.)

jlesquembre: with `--with-vendor-version-string=${version}_nix-build`:

```
openjdk 22-ga 2024-03-19
OpenJDK Runtime Environment 22-ga_nix-build (build 22-ga)
OpenJDK 64-Bit Server VM 22-ga_nix-build (build 22-ga, mixed mode, sharing)
```

Probably it is a good idea to add `--with-vendor-version-string`. What value makes sense to you? I like `${version}_nix-build`, but I don't have a strong preference.

`vendor-version-string` is defined here:
https://github.com/openjdk/jdk/blob/6968770b1e918c74fc009e3562a827bb4acbe2d7/make/autoconf/jdk-version.m4#L488-L502

As a side note regarding the OpenJDK patches on nixpkgs, I think that some are not necessary for newer JDK versions. I'll open new PRs to remove them.

doronbehar: > Probably it is a good idea to add `--with-vendor-version-string`. What value makes sense to you? I like `${version}_nix-build`, but I don't have a strong preference.

I too don't have a strong preference, but I think the `22-ga` is mentioned too many times if you'd use `${version}` there too, so I'd vote for `--with-veondor-version-string="(nix)"`.

jlesquembre: `"--with-vendor-version-string=(nix)"`:
```
openjdk 22-ga 2024-03-19
OpenJDK Runtime Environment (nix) (build 22-ga)
OpenJDK 64-Bit Server VM (nix) (build 22-ga, mixed mode, sharing)
```

@doronbehar makes sense, changing my vote to `(nix)`, I like it more :) 
Let's wait for @Infinidoge vote

Infinidoge: It feels like it displays a little bit weirdly, with `(nix)` before the version, etc. But that's less of our problem and more of OpenJDK's problem. Looks good to me!

jlesquembre: done, added `"--with-vendor-version-string=(nix)"`

doronbehar:
Feel free to cc me for a review of other changes regarding java ecosystem @jlesquembre .

PR #306625, opened 2024-04-24
pkgsForCross: add a way to create custom cross system package set

Cynerd:
Package set pkgsCross provides a lot of platforms but still that is not enough. My use case is for armv7e-m but instead of adding that this might be more generic approach. The less used platforms can be just simply declared on demand.

## Description of changes

This adds option to declare custom cross system.

The advantage over doing `import final.path` is that overlays (and other configs) are the same as in the original set without explicitly passing them.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/top-level/stage.nix
+++ b/pkgs/top-level/stage.nix
@@ -190,6 +190,9 @@ let
                               nixpkgsFun { inherit crossSystem; })
                               lib.systems.examples;
 
+    # Like pkgsCross but allows caller to specify custom crossSystem.
+    pkgsForCross = crossSystem: nixpkgsFun { inherit crossSystem; };
+
     pkgsLLVM = nixpkgsFun {
       overlays = [
         (self': super': {


Artturin (commented):
-1 on adding a function for this when `import pkgs.path { inherit (pkgs) config overlays; }` works fine.

Cynerd:
> -1 on adding a function for this when `import pkgs.path { inherit (pkgs) config overlays; }` works fine.

Don't forget that there are also `crossOverlays` and `stdenvStages` that, as far as I see, are not as easilly available. I agree that in the end, it is the same, but this ensures that you really use the same setting except those inherited even in the future if some argument gets added or removed. There might also be an option to include all initialization arguments to some attribute set and could be just used to reinitialize the same package set (such as `import pkgs.path (pkgs.args // {crossSystem= ...;})`).

Just for reference this is what I am using right now and this function would be improvement in my eyes:
```
          pkgsBuildNuttx =
            (import final.path {
              localSystem = final.buildPlatform.system;
              crossSystem = {
                config = "arm-none-eabihf";
                libc = "newlib";
                gcc = {
                  arch = "armv7e-m";
                  fpu = "vfpv3-d16";
                };
                inherit (final) overlays;
              };
            })
            .buildPackages;
```

My primary argument is that it just makes sense to me to have something like this available when `pkgCross` for example already is. You can use the same argument for that as well: you can take `lib.systems.examples`, and that works fine, so why have that `map` in the code?

Ericson2314 (changes requested):
`pkgsCross` is already a mistake. This is doubling down on it. Agree with @Artturin

Ericson2314:
You don't need to import Nixpkgs twice like in your example. Just import it once for cross, and use `pkgsBuildBuild` for native things.

Cynerd:
> You don't need to import Nixpkgs twice like in your example. Just import it once for cross, and use `pkgsBuildBuild` for native things.

I am using flakes, I have it already imported technically :wink: . This is also an example from overlay that builds firmware for embedded device alongside other tools, and thus, having a native package set with neighboring for different architectures as needed is a better pattern for this.

PR #306697, opened 2024-04-25
apple-source-releases: always use python3Minimal

reckenrode:
## Description of changes

Using python3Minimal avoids infinite recursions in the stdenv bootstrap on x86_64-darwin when building Python with configd support.

Build was tested on my ld64 branch. Primary issue is ensuring evaluation, which can be tested with `nix-instantiate`.

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [x] x86_64-darwin
  - [x] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/os-specific/darwin/apple-source-releases/default.nix
+++ b/pkgs/os-specific/darwin/apple-source-releases/default.nix
@@ -229,7 +229,7 @@ let
       pname = builtins.head (lib.splitString "/" namePath);
       appleDerivation' = stdenv: appleDerivation'' stdenv pname version sdkName sha256;
       appleDerivation = appleDerivation' stdenv;
-      callPackage = self.newScope { inherit appleDerivation' appleDerivation; };
+      callPackage = self.newScope { inherit appleDerivation' appleDerivation; python3 = pkgs.buildPackages.python3Minimal; };
     in callPackage (./. + "/${namePath}");
 
   applePackage = namePath: sdkName: sha256: let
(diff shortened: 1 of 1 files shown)


reckenrode:
The force push uses `buildPackages.python3Minimal`, which is needed for Darwin cross-compilation.

toonn:
This was previously achieved by passing in python3Minimal from the bootstrap, no? What's the motivation for changing it around? In principle it's nice for the packages in the final package set to all depend on the same Python rather than have some depend on the minimal version.

reckenrode:
> This was previously achieved by passing in python3Minimal from the bootstrap, no? What's the motivation for changing it around? In principle it's nice for the packages in the final package set to all depend on the same Python rather than have some depend on the minimal version.

Meson needs a fairly functional Python, and it simplifies things not having to do a ton of overrides to do enable that. Otherwise, I’d have to be doing overrides of configd, xnu, etc to use python3Minimal.

toonn:
I'm actually fine with overriding it for all the source releases rather than the ones that strictly need it but it would be slightly nicer not to end up with pretty much always needing two copies of Python in the store.

Would it be hard to switch to non-minimal Python after bootstrapping?

reckenrode:
> I'm actually fine with overriding it for all the source releases rather than the ones that strictly need it but it would be slightly nicer not to end up with pretty much always needing two copies of Python in the store.

Wouldn’t it only be pulled into the store as a build-time dependency?

> Would it be hard to switch to non-minimal Python after bootstrapping?

The way the source releases are implemented in `darwin-packages.nix` doesn’t expose a way to override Python for all of the packages. They’re only exposed individually. It may be something to consider for the SDK refactor. There’s going to need to be a way to override the source-release packages in bulk anyway. Maybe the bootstrap could reuse that.

toonn (approved):
LGTM, tested a couple things still build fine on both Darwins.

Not my favorite solution but it shouldn't be blocked on my account if it'll allow further progress.

Issue #295809, opened 2024-03-14
`singularity` fails to run images


ShamrockLee:
### Describe the bug

Nixpkgs's package of Sylabs SingularityCE (`singularity`) now fails to execute image.

After initial examination, the known cause is the failure to substitute default FHS `PATH`s for some files, but the fix is WIP.

### Steps To Reproduce
1. `cd NixOS/nixpkgs`
2. `nix run .#singularity exec -- "$(nix build .#singularity.tests.image-hello-cowsay --print-out-paths)" hello | nix run .#apptainer exec -- "$(nix build .#apptainer.tests.image-hello-cowsay --print-out-paths)" cowsay`
3. See error.

### Expected behavior
`singularity exec <image> <commands>` should succeed.

```
 _______________
< Hello, world! >
 ---------------
        \   ^__^
         \  (oo)\_______
            (__)\       )\/\
                ||----w |
                ||     ||
```

### Notify maintainers
@jbedo @ShamrockLee

### Metadata
Please run `nix-shell -p nix-info --run "nix-info -m"` and paste the result.

```console
[user@system:~]$ nix-shell -p nix-info --run "nix-info -m"
output here
```

---

Add a :+1: [reaction] to [issues you find important].

[reaction]: https://github.blog/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/
[issues you find important]: https://github.com/NixOS/nixpkgs/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions-%2B1-desc

SomeoneSerge:
> nix run .#singularity exec 

singularity isn't supposed to work without setuid?

ShamrockLee:
> > nix run .#singularity exec
> 
> singularity isn't supposed to work without setuid?

Yup. The error seems to result from failure to replace all the default paths after source file hierarchy changes.

PR #306716, opened 2024-04-25
apptainer, singularity: refactor defaultPath substitution

ShamrockLee:
## Description of changes

- Add argument `sourceFilesWithDefaultPaths` to specify the `defaultPath`s to substitute.

- Passthru `sourceFilesWithDefaultPaths` for easier overriding.

- Use `--replace-fail` during `defalutPath` substitution.

- Add `mount` to the `defaultPathInputs` of `apptainer` and `singularity`.

#### How to verify that the fix takes effect:

When executing images using `singularity` from the master branch, it will complain that the `mount` command is not found in PATH. The error persists even after applying the firsts commit of this PR that adds `mount` into `defaultPathInputs`. After applying the second commit of this PR that fix the `defaultPath` substitution, the error will disappear.

Note, `singularity` still fails to execute images, complaining that "`fusermount3` not found" and "permission denied" when unmounting `/tmp/rootfs-.../root`. That won't be fixed simply by adding `fuse3` into `defaultPathInputs`, as unprivileged users need the SUID-bit set `fusermount3` to unmount with `fusermount3 -r`. That one will be solved in #306730.

This PR also increase the default `memSize` of `singularity-tools.buildImage` from 512 to 1024, in the hope that `singularity.tests.image-hello-cowsay` could build successfully on CI without memory allocation error.

This is an initial effort toward fixing #295809, and it needs to be backported to Nixpkgs 24.05 if merged after branch-off.

Cc: @GaetanLepage @jbedo @SomeoneSerge @FynnFreyer

## Things done

- Built on platform(s)
  - [X] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [X] Tested, as applicable:
- [ ] Tested compilation of all packages that depend on this change using `nix-shell -p nixpkgs-review --run "nixpkgs-review rev HEAD"`. Note: all changes have to be committed, also see [nixpkgs-review usage](https://github.com/Mic92/nixpkgs-review#usage) (My own laptop is under repair, and the one I borrow doesn't have enough physical RAM to run `nixpkgs-review` within sensible amount of time. Hope someone could help me run this test.)

---

### DIFF
--- a/pkgs/applications/virtualization/singularity/generic.nix
+++ b/pkgs/applications/virtualization/singularity/generic.nix
@@ -44,6 +44,7 @@ in
   gpgme,
   libseccomp,
   libuuid,
+  mount,
   # This is for nvidia-container-cli
   nvidia-docker,
   openssl,
@@ -185,6 +186,7 @@ in
     fakeroot
     fuse2fs # Mount ext3 filesystems
     go
+    mount # mount
     privileged-un-utils
     squashfsTools # mksquashfs unsquashfs # Make / unpack squashfs image
     squashfuse # squashfuse_ll squashfuse # Mount (without unpacking) a squashfs image without privileges
--- a/pkgs/applications/virtualization/singularity/generic.nix
+++ b/pkgs/applications/virtualization/singularity/generic.nix
@@ -80,6 +80,20 @@ in
   externalLocalStateDir ? null,
   # Remove the symlinks to `singularity*` when projectName != "singularity"
   removeCompat ? false,
+  # The defaultPath values to substitute in each source files.
+  #
+  # `defaultPath` are PATH variables hard-coded inside Apptainer/Singularity
+  # binaries to search for third-party utilities, as a hardening for
+  # `$out/bin/starter-suid`.
+  #
+  # The upstream provided values are suitable for FHS-conformant environment.
+  # We substitute them and insert Nixpkgs-specific values.
+  #
+  # Example:
+  # {
+  #   "path/to/source/file1" = [ "<originalDefaultPath11>" "<originalDefaultPath12>" ... ];
+  # }
+  sourceFilesWithDefaultPaths ? { },
   # Workaround #86349
   # should be removed when the issue is resolved
   vendorHash ? _defaultGoVendorArgs.vendorHash,
@@ -88,7 +102,6 @@ in
 }:
 
 let
-  defaultPathOriginal = "/bin:/usr/bin:/sbin:/usr/sbin:/usr/local/bin:/usr/local/sbin";
   privileged-un-utils =
     if ((newuidmapPath == null) && (newgidmapPath == null)) then
       null
@@ -98,6 +111,12 @@ let
         ln -s ${lib.escapeShellArg newuidmapPath} "$out/bin/newuidmap"
         ln -s ${lib.escapeShellArg newgidmapPath} "$out/bin/newgidmap"
       '');
+
+  concatMapStringAttrsSep =
+    sep: f: attrs:
+    lib.concatMapStringsSep sep (name: f name attrs.${name}) (lib.attrNames attrs);
+
+  addShellDoubleQuotes = s: lib.escapeShellArg ''"'' + s + lib.escapeShellArg ''"'';
 in
 (buildGoModule {
   inherit pname version src;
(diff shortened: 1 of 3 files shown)


GaetanLepage (approved):
LGTM, although I lack knowledge to properly understand those changes.

GaetanLepage:
> Thank you.
> 
> > although I lack knowledge to properly understand those changes.
> 
> It goes quite deep into the implementation detail of Apptainer/Singularity.
> 
> IIUC, Apptainer/Singularity relies heavily on third-party commands instead library functions. One can think of it as a shell application (although written in Go) that builds and runs SquashFS images.
> 
> Inside the upstream sources, there exist hard-coded `PATH` variables known as `defaultPath` or `DefaultPath`. The values are FHS system-level bin paths. The purpose is to get command executables from places only sysadmins can touch, ensuring that Apptainer/Singularity won't accidentally run malicious third-parly executables placed by unprivileged users (as root or with the SUID-bit-set runner).
> 
> For obvious reasons, we need to patch those `defaultPath`s, or Apptainer/Singularity won't work as expected.
> 
> Apptainer:
> 
>     * defaultPath: https://github.com/search?q=repo%3Aapptainer%2Fapptainer%20defaultPath&type=code
> 
>     * findBin: https://github.com/apptainer/apptainer/blob/a5eb75435672593edb3a4691fd310b99796453c4/internal/pkg/util/bin/bin.go#L20
> 
> 
> SingularityCE:
> 
>     * https://github.com/search?q=repo%3Asylabs%2Fsingularity%20defaultPath&type=code
> 
>     * https://github.com/sylabs/singularity/blob/e0050f087c6e0cfc984ba67b8aa6fafbfa3e53a6/internal/pkg/util/bin/bin.go#L14

Thanks for this detailed explanation !

ShamrockLee:
@jbedo @SomeoneSerge Are you available to take a look at this? I hope to merge it before Nixpkgs 24.05 branch-off, in order not to break `singularity` further on the stable channel.

BTW, I received an e-mail from @FynnFreyer reporting that the issue is affected the GPU access of `apptainer` on Nixpkgs 23.11. I guess the same issue would also occur to `singularity`.

SomeoneSerge:
@ShamrockLee I'll take a look in the next 3 days. My apologies for the late response

pkgs/applications/virtualization/singularity/generic.nix
```diff
@@ -44,6 +44,7 @@ in
   gpgme,
   libseccomp,
   libuuid,
+  mount,
```
SomeoneSerge, line 47: Hmm, I just checked my local system and apparently I've been using `mount` from a security wrapper rather than from the store:

```
❯ which mount | xargs readlink -f
/run/wrappers/wrappers.VdfFiNgLp3/mount
```

pkgs/applications/virtualization/singularity/generic.nix
```diff
@@ -199,8 +220,19 @@ in
...
+          lib.concatMapStringsSep " \\\n  " (
+            originalDefaultPath:
+            lib.concatStringsSep " " [
+              "--replace-fail"
+              (addShellDoubleQuotes (lib.escapeShellArg originalDefaultPath))
+              (addShellDoubleQuotes ''$inputsDefaultPath''${inputsDefaultPath:+:}${lib.escapeShellArg originalDefaultPath}'')
+            ]
+          ) originalDefaultPaths
+        }
+    '') sourceFilesWithDefaultPaths}
```
SomeoneSerge, line 235: This is a bit involved and will eventually require refactoring... I think at this point we really should consider switching from templating to `__structuredAttrs`.

SomeoneSerge: This will assign our new "default paths" a higher priority than the FHS (`/usr/bin`) paths?

pkgs/applications/virtualization/singularity/generic.nix
SomeoneSerge, line 1: Whoos, the branch-off...

ShamrockLee: I appreciate your patience, and thank you for reviewing.

pkgs/applications/virtualization/singularity/generic.nix
ShamrockLee: > This is a bit involved and will eventually require refactoring...

I agree.

It would be much simpler if we only had one default path in one file to patch. This could be achieved by either
1. Improve the `defaultPath` specification by submitting a PR upstream. That sounds like a nontrivial fix, especially as I'm unfamiliar with Go.
2. Pick a file whose `defaultPath` most influences the choice of core privileged utilities. This requires further inspection of the impact of each `defaultPath` and would be prone to further fixes.



> This will assign our new "default paths" a higher priority than the FHS (`/usr/bin`) paths?

Yes, and this is the current behavior. It will change in #306730.

pkgs/applications/virtualization/singularity/generic.nix
ShamrockLee: According to [a StackExchange answer](https://unix.stackexchange.com/questions/107500/why-does-mount-have-the-suid-bit-set/107504#107504), the SUID bit set on the `mount` binary allows unprivileged users to mount/unmount some `/etc/fstab` entries with `user`/`group` options specified. However, I think that's not how it is used in Apptainer and SingularityCE.

SomeoneSerge:
Let's battle-test this

PR #306814, opened 2024-04-25
nixos/sd-card: enable Raspberry Pi's mini UART console

Prince213:
## Description of changes

Enable Raspberry Pi's mini UART console by default. This allows for headless configuration. The change has been tested on a Raspberry Pi 3 Model B.

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [X] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/nixos/modules/installer/sd-card/sd-image-aarch64.nix
+++ b/nixos/modules/installer/sd-card/sd-image-aarch64.nix
@@ -15,8 +15,14 @@
 
   # The serial ports listed here are:
   # - ttyS0: for Tegra (Jetson TX1)
+  # - ttyS1: for Raspberry Pi's mini UART
   # - ttyAMA0: for QEMU's -machine virt
-  boot.kernelParams = ["console=ttyS0,115200n8" "console=ttyAMA0,115200n8" "console=tty0"];
+  boot.kernelParams = [
+    "console=ttyS1,115200n8"
+    "console=ttyS0,115200n8"
+    "console=ttyAMA0,115200n8"
+    "console=tty0"
+  ];
 
   sdImage = {
     populateFirmwareCommands = let


Artturin (approved):
The same parameter is mentioned at https://wiki.nixos.org/wiki/NixOS_on_ARM/Raspberry_Pi#Raspberry_Pi_3_/_3B+ and https://wiki.nixos.org/wiki/NixOS_on_ARM/Raspberry_Pi#Serial_console

Prince213:
> The same parameter is mentioned at https://wiki.nixos.org/wiki/NixOS_on_ARM/Raspberry_Pi#Raspberry_Pi_3_/_3B+ and https://wiki.nixos.org/wiki/NixOS_on_ARM/Raspberry_Pi#Serial_console

Thanks for pointing that out! Actually it's these pages that make me feel to create a PR and 'upstream' the changes. After this PR is merged the wiki should be updated as well.

Prince213:
I've changed the order to also allow kernel messages to be printed to the serial console.

samueldr:
> I've changed the order to also allow kernel messages to be printed to the serial console.

This may break for other systems where `tty0` is the serial console. All in all, it is more common than `tty1`.

Prince213:
> > I've changed the order to also allow kernel messages to be printed to the serial console.
> 
> This may break for other systems where `tty0` is the serial console. All in all, it is more common than `tty1`.

Maybe you mean `ttyS0` and `ttyS1`? By the [kernel docs](https://www.kernel.org/doc/html/latest/admin-guide/serial-console.html) the only affected system would be Tegra. And IIRC for some reasons, if I put `ttyS1` after `ttyS0` neither kernel messages or login prompts would show up (but u-boot messages do show up). Not sure about Tegra though.

I think it may be better to have per-hardware modules for Raspberry Pi 3/4 aarch64 like `nixos/modules/installer/sd-card/sd-image-raspberrypi.nix`.

samueldr:
Oops yes, sorry, I did mean tty**S**#.

The kernel documentation linked here, and other parts elsewhere about serial are **heavily** outdated, and actually wrong in some places. For example, the maximum is erroneous, many devices use a higher baud rate (i.e. all/most rockchip).

> The behavior is more complicated when the same device type is defined more times. In this case, there are the following two rules:
>
>   1. The output will appear only on the first device of each defined type.

Also noting that, same example SoC, most rockchip hardware will use `ttyS2` as the preferred console. The expectation is that if the console isn't ttyS0, the `extlinux.conf` file can be edited manually to set an appropriate console temporarily (think of it as if you hit `e` to edit the cmdline in GRUB).

(From the linked documentation)

> If no console device is specified, the first device found capable of acting as a system console will be used. At this time, the system first looks for a VGA card and then for a serial port. So if you don’t have a VGA card in your system the first serial port will automatically become the console.

Outdated! The answer is: *it depends*.

Sadly, there is no real good answer to that conundrum in DT-land. Setting no `console=` values uses `/chosen/stdout` in the device tree, but this breaks display output. Only setting `tty0` doesn't make the kernel use the serial console otherwisely. So to *improve* defaults, something needs to be picked. `ttyS0` is the least worst option, by default.

samueldr:
Forgot to anser this:

> I think it may be better to have per-hardware modules for Raspberry Pi 3/4 aarch64 like `nixos/modules/installer/sd-card/sd-image-raspberrypi.nix`. Or we could enable serial-getty for all these chardevs.

Nope. The answer is to remove anything device-specific. Not much has been moving precisely within NixOS, but I made notes about [planning for a better NixOS on ARM](https://discourse.nixos.org/t/planning-for-a-better-nixos-on-arm-and-other-non-x86-64-systems/15346) a while back.

Prince213:
> Forgot to answer this:
> 
> > I think it may be better to have per-hardware modules for Raspberry Pi 3/4 aarch64 like `nixos/modules/installer/sd-card/sd-image-raspberrypi.nix`. Or we could enable serial-getty for all these chardevs.
> 
> Nope. The answer is to remove anything device-specific. Not much has been moving precisely within NixOS, but I made notes about [planning for a better NixOS on ARM](https://discourse.nixos.org/t/planning-for-a-better-nixos-on-arm-and-other-non-x86-64-systems/15346) a while back.

Thanks for the great writeup! It feels right for a unified and platform-agnostic approach. I've investigated on installing NixOS with UEFI and so far I haven't met any problems. I think the best way forward would be not investing into a to-be-deprecated feature.

Prince213:
Closing since a UEFI-based approach should be used instead.

PR #306956, opened 2024-04-26
hotspot: 1.4.1 -> 1.5.0

K900:
## Description of changes

Finally!

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [x] Tested, as applicable:

---

### DIFF
--- a/pkgs/development/tools/analysis/hotspot/default.nix
+++ b/pkgs/development/tools/analysis/hotspot/default.nix
@@ -1,10 +1,15 @@
 { lib
-, mkDerivation
+, stdenv
+, binutils
 , cmake
-, elfutils
 , extra-cmake-modules
+, patchelfUnstable
+, wrapQtAppsHook
+, elfutils
 , fetchFromGitHub
+, fetchpatch
 , kconfigwidgets
+, kddockwidgets
 , ki18n
 , kio
 , kitemmodels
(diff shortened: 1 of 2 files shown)


pkgs/development/tools/analysis/hotspot/default.nix
```diff
@@ -64,6 +70,7 @@ mkDerivation rec {
 
   qtWrapperArgs = [
     "--prefix LD_LIBRARY_PATH : ${lib.makeLibraryPath [ rustc-demangle ]}"
```
SuperSandro2000, line 72: Do we still need LD_LIBRARY_PATH here?

K900: It does dlopen rustc_demangle. We could probably replace it with a patchelf call.

pkgs/development/tools/analysis/hotspot/default.nix
```diff
@@ -64,6 +70,7 @@ mkDerivation rec {
...
+    "--prefix PATH : ${lib.makeBinPath [ linuxPackages.perf ]}"
```
alois31, line 73: While we are at adding stuff to `PATH`, maybe we should add binutils too so that the disassembly view works?

K900: Wait it doesn't? I guess I tested it in a nix-shell...

alois31: nix-shell includes stdenv, which includes binutils, so it works there.

K900:
Bunch of updates:
- removed the now-unnecessary .git hack
- pulled in patchelf-unstable to add rustc-demangle instead of LD_LIBRARY_PATH shenanigans
- added binutils
- moved PATH changes to go last instead of first, so user-added perf/binutils win

alois31:
There is a very annoying regression that makes it inconvenient to use the "attach to process" or "profile system" modes, which I tracked down in https://github.com/KDAB/hotspot/pull/625. I recommend waiting for the situation to be resolved upstream and then pull an appropriate patch.

SuperSandro2000 (approved):
nix wise looks pretty solid to me

PR #290598, opened 2024-02-22
zrythm: 1.0.0-beta.4.9.1 -> 1.0.0-beta.6.4.1

PowerUser64:
## Description of changes

Update zrythm to [`v1.0.0-beta.6.4.1`](https://gitlab.zrythm.org/zrythm/zrythm/-/tags/v1.0.0-beta.6.4.1). Please note that I am using `overrideAttrs` in two places because nixpkgs' gtk and rtaudio are not on new enough versions for zrythm to compile. I was told that using `overrideAttrs` in buildInputs *should* be allowed [here](https://github.com/NixOS/nixpkgs/pull/253765#issuecomment-1889355182), but please let me know if this is not true.

Closes https://github.com/NixOS/nixpkgs/issues/252561
Closes #253765

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
- For non-Linux: Is sandboxing enabled in `nix.conf`? (See [Nix manual](https://nixos.org/manual/nix/stable/command-ref/conf-file.html))
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:
  - [NixOS test(s)](https://nixos.org/manual/nixos/unstable/index.html#sec-nixos-tests) (look inside [nixos/tests](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests))
  - and/or [package tests](https://nixos.org/manual/nixpkgs/unstable/#sec-package-tests)
  - or, for functions and "core" functionality, tests in [lib/tests](https://github.com/NixOS/nixpkgs/blob/master/lib/tests) or [pkgs/test](https://github.com/NixOS/nixpkgs/blob/master/pkgs/test)
  - made sure NixOS tests are [linked](https://nixos.org/manual/nixpkgs/unstable/#ssec-nixos-tests-linking) to the relevant packages
- [ ] Tested compilation of all packages that depend on this change using `nix-shell -p nixpkgs-review --run "nixpkgs-review rev HEAD"`. Note: all changes have to be committed, also see [nixpkgs-review usage](https://github.com/Mic92/nixpkgs-review#usage)
- [x] Tested basic functionality of all binary files (usually in `./result/bin/`)
[...]

PR #253765, opened 2023-09-07
zrythm: 1.0.0-beta.4.9.1 -> 1.0.0-beta.4.12.5

davidak:
## Description of changes

Related https://github.com/NixOS/nixpkgs/issues/252561

can't add @yuuyins as reviewer for some reason

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
- [ ] For non-Linux: Is `sandbox = true` set in `nix.conf`? (See [Nix manual](https://nixos.org/manual/nix/stable/command-ref/conf-file.html))
- [ ] Tested, as applicable:
  - [NixOS test(s)](https://nixos.org/manual/nixos/unstable/index.html#sec-nixos-tests) (look inside [nixos/tests](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests))
  - and/or [package tests](https://nixos.org/manual/nixpkgs/unstable/#sec-package-tests)
  - or, for functions and "core" functionality, tests in [lib/tests](https://github.com/NixOS/nixpkgs/blob/master/lib/tests) or [pkgs/test](https://github.com/NixOS/nixpkgs/blob/master/pkgs/test)
  - made sure NixOS tests are [linked](https://nixos.org/manual/nixpkgs/unstable/#ssec-nixos-tests-linking) to the relevant packages
- [ ] Tested compilation of all packages that depend on this change using `nix-shell -p nixpkgs-review --run "nixpkgs-review rev HEAD"`. Note: all changes have to be committed, also see [nixpkgs-review usage](https://github.com/Mic92/nixpkgs-review#usage)
- [ ] Tested basic functionality of all binary files (usually in `./result/bin/`)
- [23.11 Release Notes](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2311.section.md) (or backporting [23.05 Release notes](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2305.section.md))
  - [ ] (Package updates) Added a release notes entry if the change is major or breaking
  - [ ] (Module updates) Added a release notes entry if the change is significant
  - [ ] (Module addition) Added a release notes entry if adding a new NixOS module
- [x] Fits [CONTRIBUTING.md](https://github.com/NixOS/nixpkgs/blob/master/CONTRIBUTING.md).

Issue #252561, opened 2023-08-31
Update request: zrythm 1.0.0-beta.4.9.1 → 1.0.0-beta.4.12.34


nyabinary:
- Package name: zrythm
- Latest released version: v1.0.0-beta.4.12.34

- Current version on the unstable channel:  1.0.0-beta.4.9.1
- Current version on the stable/release channel: 1.0.0-beta.4.6.3

- [ X] Checked the [nixpkgs pull requests](https://github.com/NixOS/nixpkgs/pulls)

**Notify maintainers**
@yuuyins 
@magnetophon 
@tshaynik 

-----

Note for maintainers: Please tag this issue in your PR.

ShalokShalom:
1.0 is about to release 🥳

ShalokShalom:
They are now packaging it, see the last entry in this issue :)

You can close it

PowerUser64:
Superseded by #306959 (Merged)

PR #306959, opened 2024-04-26
zrythm: 1.0.0-beta.4.9.1 -> 1.0.0-rc.1

Astavie:
## Description of changes

Update zrythm to [v1.0.0-rc.1](https://gitlab.zrythm.org/zrythm/zrythm/-/tags/v1.0.0-rc.1).

This depends on https://github.com/NixOS/nixpkgs/pull/291339 as zrythm depends on gtk 4.14. This is why this PR targets `staging-next`.

Closes https://github.com/NixOS/nixpkgs/issues/252561
Closes https://github.com/NixOS/nixpkgs/pull/253765
Closes https://github.com/NixOS/nixpkgs/pull/290598

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/applications/audio/zrythm/default.nix
+++ b/pkgs/applications/audio/zrythm/default.nix
@@ -1,12 +1,9 @@
 { stdenv
 , lib
 , fetchFromGitHub
-, fetchFromSourcehut
-, fetchpatch
-, SDL2
+, fetchzip
 , alsa-lib
 , appstream
-, appstream-glib
 , bash-completion
 , boost
 , breeze-icons
@@ -16,7 +13,6 @@
 , curl
 , dbus
 , dconf
-, faust2lv2
 , fftw
 , fftwFloat
 , flex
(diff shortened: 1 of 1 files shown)


Astavie:
(people involved in previous PRs mentioned)

pkgs/applications/audio/zrythm/default.nix
```diff
@@ -61,48 +58,46 @@
 , serd
```
PowerUser64, line 58: `lilv`, `reproc`, `libcyaml`, and `libyaml` have been dropped as dependencies since the last version, so they can be removed here too.

pkgs/applications/audio/zrythm/default.nix
```diff
@@ -220,16 +216,16 @@ stdenv.mkDerivation rec {
...
+      --prefix GSETTINGS_SCHEMA_DIR : "$out/share/gsettings-schemas/${finalAttrs.pname}-${finalAttrs.version}/glib-2.0/schemas/"
       --prefix XDG_DATA_DIRS : "$XDG_ICON_DIRS:${breeze-icons}/share"
     )
   '';
 
   meta = with lib; {
     homepage = "https://www.zrythm.org";
     description = "Automated and intuitive digital audio workstation";
-    maintainers = with maintainers; [ tshaynik magnetophon yuu ];
+    maintainers = with maintainers; [ tshaynik magnetophon yuu astavie ];
```
PowerUser64, line 227: Can you add PowerUser64 (me) as a maintainer too?

PowerUser64 (changes requested):
Just a few changes requested. Also, since there were so many versions of this PR, it would probably make sense to add them all as Co-authors so it's documented.

pkgs/applications/audio/zrythm/default.nix
```diff
@@ -220,16 +216,16 @@ stdenv.mkDerivation rec {
...
     platforms = platforms.linux;
```
PowerUser64, line 228: Could add `broken = stdenv.isDarwin;` and change this line to `platforms.unix` (like what's done [here](https://github.com/NixOS/nixpkgs/blob/7bb2ccd8cdc44c91edba16c48d2c8f331fb3d856/pkgs/applications/audio/sonobus/default.nix#L83-L84)) because zrythm runs on macos but we don't have the framework to build it written.

PowerUser64 (changes requested):

pkgs/applications/audio/zrythm/default.nix
```diff
@@ -61,48 +58,46 @@
...
+      rev = "948991d7b5104280c03960925908e589c77b169a";
+      hash = "sha256-uGAuKheoMfP9hZXsw29ec+58dJM8wMuowe95QutzKBY=";
     };
   });
 in
-stdenv.mkDerivation rec {
+stdenv.mkDerivation (finalAttrs: {
   pname = "zrythm";
-  version = "1.0.0-beta.4.9.1";
+  version = "1.0.0-rc.0";
```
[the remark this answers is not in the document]
Astavie: It still doesn't have a tar.gz on https://www.zrythm.org/releases/ so I can't switch to there just yet

Astavie:
thanks for the review, will make these changes once the tar.gz arrives for rc.1

pkgs/applications/audio/zrythm/default.nix
PowerUser64: Edit - `libcyaml` is still required to upgrade old projects that are still in the yaml format.

Astavie: according to the cmake `lilv` is also still required to compile

alex-tee:
Upstream maintainer here, thanks for packaging this. Some feedback below.

These can/should be removed:
sdl2, appstream-glib, breeze-icons, faust2lv2 (plugins are already compiled into C++), libgtop, libsass (only sassc is needed as a native build input), libsoundio, libxml2, pandoc, pcre (only pcre2 is needed), serd, sord, sratom

These should be moved to native build input ("make" dependencies):
guile, chromaprint, help2man, flex, jq (why is this needed?), lilv

Astavie:
@alex-tee libxml2 is still required as a native input for xmllint, and the cmake still lists sdl2, serd, sord, and sratom as runtime dependencies

alex-tee:
@Astavie 
> and the cmake still lists sdl2, serd, sord, and sratom as runtime dependencies

"the cmake"?

Astavie:
@alex-tee ah apologies, i misinterpreted the logs. it's the meson build file: here is a screenshot
image

alex-tee:
@Astavie 

serd/sord/sratom/lilv are only needed at build time (lilv is used in tests), regardless if that says "runtime dependencies". They are used as "runtime dependencies" by a tool that gets compiled and used during tests, but they are not linked into the main Zrythm binary.

Regarding SDL, it gets marked as required because of `-Dsdl=enabled` in the meson configure flags: https://github.com/NixOS/nixpkgs/pull/306959/files#diff-1adcc62edbc935bfbb269175905b4cc5b0c4a4fd83a2bca9c3ce2a30b1844a40R194

Please remove that as the SDL backend is [broken and experimental](https://manual.zrythm.org/en/getting-started/system-requirements.html#backend-support).

Astavie:
superfluous dependencies have been removed!

alex-tee:
LGTM, thanks!

PowerUser64:
Thanks for taking a look at this @alex-tee!

pkgs/applications/audio/zrythm/default.nix
```diff
@@ -197,7 +174,6 @@ stdenv.mkDerivation rec {
     "-Dmanpage=true"
     "-Drtaudio=enabled"
     "-Drtmidi=enabled"
-    "-Dsdl=enabled"
     # "-Duser_manual=true" # needs sphinx-intl
```
PowerUser64, line 177: I thought this comment was interesting, so I decided to look into it.

[`git blame` shows](https://github.com/NixOS/nixpkgs/blame/e1a4c08ae81a1e4c99b7496bd232cc4e47708957/pkgs/applications/audio/zrythm/default.nix#L201) that it's over 3 years old! Since then, `python312Packages.sphinx-intl` has been added to nixpkgs. I don't know how important this is to people, but given that we have this package now I think this _could_ be enabled. Not sure how important this is to people. 

Thoughts?

cc @tshaynik

PowerUser64 (commented):
Looks good! Just one comment. Not sure how the manual should be added, if it's something we want to add to the output. Could be worth it for completeness given that v1 is on its way, but also I don't know how the output is supposed to be used. I assume it's equivalent to the live version at https://docs.zrythm.org/.

pkgs/applications/audio/zrythm/default.nix
alex-tee: I don't know what the nix packager way to do is, but, for reference, the python requirements to build the user manual are in `requirements.txt`. This is how you can build the user manual in HTML format in English:
```
python3 -m venv venv
source ./venv/bin/activate
python3 -m pip install -r requirements.txt
# setup meson (meson setup builddir ...)
ninja -C builddir html-manual-en
```

Then the built manual will be available in the builddir under doc/user

Astavie:
I've marked this as ready for review now as rtaudio_6 has been merged. Adding the user manual could be added in another PR

PowerUser64:
Nice work @Astavie! Glad to finally get this merged!

jtojnar:
> These should be moved to native build input ("make" dependencies):
> guile, chromaprint, help2man, flex, jq (why is this needed?), lilv

`nativeBuildInputs` are orthogonal concept to “make” dependencies (used e.g. in Arch pkgbuild). Actually, Nix does not distinguish between build and runtime dependencies explicitly – every store path referenced in the build output is considered a runtime dependency.

The distinction between `nativeBuildInputs` and `buildInputs` is about what platform the packages should execute on, and the distinction is only important in the context of cross-compilation.

> serd/sord/sratom/lilv are only needed at build time (lilv is used in tests), regardless if that says "runtime dependencies". They are used as "runtime dependencies" by a tool that gets compiled and used during tests, but they are not linked into the main Zrythm binary.

Then `checkInputs` would be suitable. But that requires `doCheck = true` to be included so `buildInputs` might be best option for now.

PR #307035, opened 2024-04-26
nixos/initrd-network: run DHCP in the background to increase its resiliency

### LINKED ISSUE #306858: initrd fails to get a DHCP address if DHCP server is not available at boot time
### Describe the bug
If boot.initrd.network.enable is true, initrd will start the network and try and acquire DHCP addresses for any network interfaces that are set for DHCP.

This works fine if the DHCP server is up and running, but consider a situation where power has been lost to the entire network. When the power is restored, a system may try and acquire a DHCP address before the DHCP server has started.

initrd only tries once to acquire a DHCP address. If it is not able to get one, then it will be impossible to access the initrd boot environment over the network, for example via ssh to perform pre-boot tasks.

### Steps To Reproduce
1. Disable DHCP server (or unplug Ethernet cable)
2. Boot system. The system will try (and fail) to get a DHCP address and then keep going.
3. Restore the DHCP server.
4. System does not try to get a DHCP address any more even though the DHCP server is available.

### Expected behavior
The DHCP client should run in the background so that it can acquire a DHCP address at any time.

### Metadata

```console
> nix-shell -p nix-info --run "nix-info -m"
 - system: `"x86_64-linux"`
 - host os: `Linux 6.1.87, NixOS, 23.11 (Tapir), 23.11.20240424.dirty`
 - multi-user?: `yes`
 - sandbox: `yes`
 - channels(phunt): `"home-manager-22.11.tar.gz, nixos-unstable"`
 - channels(root): `"nixos-23.11"`
```

presto8:
## Description of changes

This changeset changes the behavior of the DHCP client during the initrd boot sequence.

Previously, the DHCP client (udhcpc) would try a single time to acquire a DHCP address on each interface for which DHCP was configured. It would also block until the DHCP request either succeeded or failed.

If the DHCP server on the network is not available when udhcpc tries to get an address, then there will be no further attempts to obtain an address, even if the DHCP server becomes available again. A common situation in which this can happen is during boot up after a power loss: the DHCP server may be slower to come up than the system.

If ssh has been enabled in initrd, the result is that the system is not accessible if DHCP did not succeed. (Local console use is unaffected).

This changeset removes the --now option and puts udhcpc in the background for all interfaces. This means that udhcpc will continue to attempt to get a DHCP address in the background until one is obtained.

Tested by booting up a system with the network cable unplugged; waiting for the first DHCP request to time out; plugging the network cable back in; and observing that a DHCP address was obtained successfully. Then the boot procedure finished normally.

Resolves #306858

## Things done

- Built on platform(s)
  - [X] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

### DIFF
--- a/nixos/modules/system/boot/initrd-network.nix
+++ b/nixos/modules/system/boot/initrd-network.nix
@@ -152,7 +152,8 @@ in
           # Acquire DHCP leases.
           for iface in ${dhcpIfShellExpr}; do
             echo "acquiring IP address via DHCP on $iface..."
-            udhcpc --quit --now -i $iface -O staticroutes --script ${udhcpcScript} ${udhcpcArgs}
+            # run DHCP in the background until it acquires an address
+            udhcpc --quit -i $iface -O staticroutes --script ${udhcpcScript} ${udhcpcArgs} &
           done
         ''
 


Kreyren (commented):
test

nixos/modules/system/boot/initrd-network.nix
```diff
@@ -144,13 +144,15 @@ in
...
-          udhcpc --quit --now -i $iface -O staticroutes --script ${udhcpcScript} ${udhcpcArgs}
+          # run DHCP in the background until it acquires an address
+          udhcpc --quit -i $iface -O staticroutes --script ${udhcpcScript} ${udhcpcArgs} &
         done
       ''
 
       + cfg.postCommands));
 
     boot.initrd.postMountCommands = mkIf (cfg.flushBeforeStage2 && !config.boot.initrd.systemd.enable) ''
+      pkill udhcpc
```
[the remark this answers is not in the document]
presto8: Thanks for the comment! I think pkill will only match "udhcpc" exactly here (it needs the -f option to match substrings) but let me look into possibly collecting the PIDs and storing them in a variable. It would be more elegant.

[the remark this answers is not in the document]
presto8: I was wrong: pill needs --exact and then it will require an exact match on the process name. I'll still look at collecting the PIDs though.

nixos/modules/system/boot/initrd-network.nix
```diff
@@ -144,13 +144,15 @@ in
         # Acquire DHCP leases.
         for iface in ${dhcpIfShellExpr}; do
           echo "acquiring IP address via DHCP on $iface..."
-          udhcpc --quit --now -i $iface -O staticroutes --script ${udhcpcScript} ${udhcpcArgs}
+          # run DHCP in the background until it acquires an address
+          udhcpc --quit -i $iface -O staticroutes --script ${udhcpcScript} ${udhcpcArgs} &
```
Kreyren, line 148:
```suggestion
          # run DHCP in the background until it acquires an address
          udhcpc --quit -i $iface -O staticroutes --script ${udhcpcScript} ${udhcpcArgs} &
          udhcpc_pid="$!" # Collect pid
```

nixos/modules/system/boot/initrd-network.nix
Kreyren, line 155:
```suggestion
          # run DHCP in the background until it acquires an address
          udhcpc --quit -i $iface -O staticroutes --script ${udhcpcScript} ${udhcpcArgs} &
          udhcpc_pid="$!" # Collect PID
        done
      ''

      + cfg.postCommands));

    boot.initrd.postMountCommands = mkIf (cfg.flushBeforeStage2 && !config.boot.initrd.systemd.enable) ''
      kill -s SIGTERM $udhcpc_pid
```

presto8: Thanks for the research and thorough comment!

Yes, I agree we would need something like this:
```
pids=()
for i in {1..5}; do
    sleep 60 &
    pids+=($!)
done

kill "${pids[@]}"
```

I was travelling but am back now. I will try it out and see if it works this weekend.

presto8: I finally had some time to test it. It turns out that the boot process must kill any processes that are still running as I didn't need any kill statement at all. When the normal system finished booting, there was no udhcpc process.

presto8:
> Interested in this ... I have a headless raspi that could use this change

It should be ready to test, would you be able to try it and report back how it goes?

presto8:
Thanks for the feedback! I run it on all my systems too, and it seems ok. The only minor issue is that the DHCP messages may show up on the screen while the user is entering the password. It doesn't cause any issue, but it may be disconcerting to the user. I am not aware of any solution though, other than suppressing all DHCP output to the screen, which seems worse than doing nothing.

nixos/modules/system/boot/initrd-network.nix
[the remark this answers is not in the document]
presto8: I don't think this pkill is even needed. The interfaces will all be taken down before initrd ends. If udhcpc is still running, it will exit when the underlying interface goes away.

ElvishJerricco:
How is `udhcpc` exiting with this change? Is it just relying on the process reaper in `switch_root`? That *might* work but I'm not sure, since `/dev`, `/proc`, `/run`, and `/sys` have all been moved by then. I would think we would want to make sure `udhcpc` is shut down in `postMountCommands` with this change.

presto8:
> How is `udhcpc` exiting with this change?

Originally I did a pkill to kill all the udhcp processes. But then someone pointed out that was probably overkill. Then I noticed that, as you pointed out, the processes seem to get killed automatically so I just left it out. You are right, this is not the best. I'll add some code to properly remove them and test it out (in postMountCommands as you suggest).

presto8:
I think this approach may work, the other idea is to store the pids of the started processes in a file and then kill them.
```
# Stop any running udhcpc                                                                                                                                         
for iface in ${dhcpIfShellExpr}; do                                                                                                                               
  pkill -f "udhcpc.*$iface"                                                                                                                                       
done                                                                                                                                                              
```

PR #307056, opened 2024-04-26
k3s: skip pre-releases in update script

superherointj:
k3s: skip pre-releases in update script

Avoids bot offering updates when in pre-release.
Currently there are packages in pre-release in: https://github.com/k3s-io/k3s/releases
Soon the bot will open a pre-release PR...

CC @euank @mic92 @yajo

### DIFF
--- a/pkgs/applications/networking/cluster/k3s/update-script.sh
+++ b/pkgs/applications/networking/cluster/k3s/update-script.sh
@@ -19,7 +19,9 @@ LATEST_TAG_RAWFILE=${WORKDIR}/latest_tag.json
 curl --silent -f ${GITHUB_TOKEN:+-u ":$GITHUB_TOKEN"} \
     https://api.github.com/repos/k3s-io/k3s/releases > ${LATEST_TAG_RAWFILE}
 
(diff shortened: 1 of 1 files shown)


pkgs/applications/networking/cluster/k3s/update-script.sh
```diff
@@ -19,7 +19,9 @@ LATEST_TAG_RAWFILE=${WORKDIR}/latest_tag.json
 curl --silent -f ${GITHUB_TOKEN:+-u ":$GITHUB_TOKEN"} \
     https://api.github.com/repos/k3s-io/k3s/releases > ${LATEST_TAG_RAWFILE}
 
-LATEST_TAG_NAME=$(jq 'map(.tag_name)' ${LATEST_TAG_RAWFILE} | \
+LATEST_TAG_NAME=$(cat ${LATEST_TAG_RAWFILE} | \
```
superherointj, line 22: I have used `cat` on purpose (instead of last argument of `jq`). I want the input file to be first because it's naturally easier to understand this way. I'm optimizing for being easy to read/understand instead of performance.

PR #307068, opened 2024-04-26
systemd: 255.9 -> 256.2

### LINKED ISSUE #319328: Update request: systemd 255.6 → 256.2
- Package name: [systemd](https://systemd.io/)
- Latest released version: [256.2](https://github.com/systemd/systemd/releases/tag/v256.2)

- Current version on the unstable channel: [255.6](https://github.com/NixOS/nixpkgs/blob/nixos-unstable/pkgs/os-specific/linux/systemd/default.nix#L887)
- Current version on the stable/release channel: [255.6](https://github.com/NixOS/nixpkgs/blob/nixos-24.05/pkgs/os-specific/linux/systemd/default.nix#L887)

**Notify maintainers**

-----

---

nikstur:
## Description of changes

Closes https://github.com/NixOS/nixpkgs/issues/319328

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/nixos/modules/system/boot/networkd.nix
+++ b/nixos/modules/system/boot/networkd.nix
@@ -18,12 +18,16 @@ let
           "ManageForeignRoutes"
           "RouteTable"
           "IPv6PrivacyExtensions"
+          "IPv4Forwarding"
+          "IPv6Forwarding"
         ])
         (assertValueOneOf "SpeedMeter" boolValues)
         (assertInt "SpeedMeterIntervalSec")
         (assertValueOneOf "ManageForeignRoutingPolicyRules" boolValues)
         (assertValueOneOf "ManageForeignRoutes" boolValues)
         (assertValueOneOf "IPv6PrivacyExtensions" (boolValues ++ ["prefer-public" "kernel"]))
+        (assertValueOneOf "IPv4Forwarding" boolValues)
+        (assertValueOneOf "IPv6Forwarding" boolValues)
       ];
 
       sectionDHCPv4 = checkUnitConfig "DHCPv4" [
@@ -652,6 +656,8 @@ let
           "DNSDefaultRoute"
           "NTP"
           "IPForward"
+          "IPv4Forwarding"
+          "IPv6Forwarding"
           "IPMasquerade"
           "IPv6PrivacyExtensions"
           "IPv6AcceptRA"
@@ -700,7 +706,9 @@ let
         (assertValueOneOf "LLDP" (boolValues ++ ["routers-only"]))
         (assertValueOneOf "EmitLLDP" (boolValues ++ ["nearest-bridge" "non-tpmr-bridge" "customer-bridge"]))
         (assertValueOneOf "DNSDefaultRoute" boolValues)
-        (assertValueOneOf "IPForward" (boolValues ++ ["ipv4" "ipv6"]))
+        (assertRemoved "IPForward" "IPv4Forwarding and IPv6Forwarding in systemd.network(5) and networkd.conf(5)")
+        (assertValueOneOf "IPv4Forwarding" boolValues)
+        (assertValueOneOf "IPv6Forwarding" boolValues)
         (assertValueOneOf "IPMasquerade" (boolValues ++ ["ipv4" "ipv6" "both"]))
         (assertValueOneOf "IPv6PrivacyExtensions" (boolValues ++ ["prefer-public" "kernel"]))
         (assertValueOneOf "IPv6AcceptRA" boolValues)
@@ -2835,6 +2843,7 @@ let
         "systemd-networkd-wait-online.service"
         "systemd-networkd.service"
         "systemd-networkd.socket"
+        "systemd-networkd-persistent-storage.service"
       ];
 
       environment.etc."systemd/networkd.conf" = renderConfig cfg.config;
--- a/nixos/modules/system/boot/stage-1.nix
+++ b/nixos/modules/system/boot/stage-1.nix
@@ -131,6 +131,7 @@ let
 
       # Copy udev.
       copy_bin_and_libs ${udev}/bin/udevadm
+      cp ${lib.getLib udev.kmod}/lib/libkmod.so* $out/lib
       copy_bin_and_libs ${udev}/lib/systemd/systemd-sysctl
       for BIN in ${udev}/lib/udev/*_id; do
         copy_bin_and_libs $BIN
--- a/nixos/modules/system/boot/systemd.nix
+++ b/nixos/modules/system/boot/systemd.nix
@@ -37,6 +37,8 @@ let
       "cryptsetup.target"
       "cryptsetup-pre.target"
       "remote-cryptsetup.target"
+    ] ++ optionals cfg.package.withTpm2Tss [
+      "tpm2.target"
     ] ++ [
       "sigpwr.target"
       "timers.target"
@@ -112,6 +114,7 @@ let
       "sleep.target"
       "hybrid-sleep.target"
       "systemd-hibernate.service"
+      "systemd-hibernate-clear.service"
       "systemd-hybrid-sleep.service"
       "systemd-suspend.service"
       "systemd-suspend-then-hibernate.service"
@@ -136,6 +139,16 @@ let
       "systemd-ask-password-wall.path"
       "systemd-ask-password-wall.service"
 
+      # Varlink APIs
+      "systemd-bootctl@.service"
+      "systemd-bootctl.socket"
+      "systemd-creds@.service"
+      "systemd-creds.socket"
+    ] ++ lib.optional cfg.package.withTpm2Tss [
+      "systemd-pcrlock@.service"
+      "systemd-pcrlock.socket"
+    ] ++ [
+
       # Slices / containers.
       "slices.target"
     ] ++ optionals cfg.package.withImportd [
@@ -158,6 +171,7 @@ let
     ] ++ optionals cfg.package.withHostnamed [
       "dbus-org.freedesktop.hostname1.service"
       "systemd-hostnamed.service"
+      "systemd-hostnamed.socket"
     ] ++ optionals cfg.package.withPortabled [
       "dbus-org.freedesktop.portable1.service"
       "systemd-portabled.service"
--- a/nixos/modules/system/boot/systemd/initrd.nix
+++ b/nixos/modules/system/boot/systemd/initrd.nix
@@ -70,6 +70,7 @@ let
     "systemd-tmpfiles-setup-dev.service"
     "systemd-tmpfiles-setup.service"
     "timers.target"
+    "tpm2.target"
     "umount.target"
     "systemd-bsod.service"
   ] ++ cfg.additionalUpstreamUnits;
@@ -344,7 +345,8 @@ in {
     };
 
     enableTpm2 = mkOption {
-      default = true;
+      default = cfg.package.withTpm2Tss;
+      defaultText = "boot.initrd.systemd.package.withTpm2Tss";
       type = types.bool;
       description = ''
         Whether to enable TPM2 support in the initrd.
@@ -460,6 +462,7 @@ in {
         "${cfg.package}/lib/systemd/systemd-sulogin-shell"
         "${cfg.package}/lib/systemd/systemd-sysctl"
         "${cfg.package}/lib/systemd/systemd-bsod"
+        "${cfg.package}/lib/systemd/systemd-sysroot-fstab-check"
 
         # generators
         "${cfg.package}/lib/systemd/system-generators/systemd-debug-generator"
@@ -486,6 +489,8 @@ in {
         # fido2 support
         "${cfg.package}/lib/cryptsetup/libcryptsetup-token-systemd-fido2.so"
         "${pkgs.libfido2}/lib/libfido2.so.1"
+      ] ++ optionals cfg.package.withKmod [
+        "${pkgs.kmod.lib}/lib/libkmod.so.2"
       ] ++ jobScripts;
 
       targets.initrd.aliases = ["default.target"];
--- a/nixos/modules/system/boot/systemd/initrd.nix
+++ b/nixos/modules/system/boot/systemd/initrd.nix
@@ -488,8 +488,6 @@ in {
         # fido2 support
         "${cfg.package}/lib/cryptsetup/libcryptsetup-token-systemd-fido2.so"
         "${pkgs.libfido2}/lib/libfido2.so.1"
-      ] ++ optionals cfg.package.withKmod [
-        "${pkgs.kmod.lib}/lib/libkmod.so.2"
       ] ++ jobScripts
       ++ map (c: builtins.removeAttrs c ["text"]) (builtins.attrValues cfg.contents);
 
--- a/nixos/modules/system/boot/systemd/initrd.nix
+++ b/nixos/modules/system/boot/systemd/initrd.nix
@@ -112,8 +112,7 @@ let
     inherit (config.boot.initrd) compressor compressorArgs prepend;
     inherit (cfg) strip;
 
-    contents = map (path: { object = path; symlink = ""; }) (subtractLists cfg.suppressedStorePaths cfg.storePaths)
-      ++ mapAttrsToList (_: v: { object = v.source; symlink = v.target; }) (filterAttrs (_: v: v.enable) cfg.contents);
+    contents = lib.filter ({ source, ... }: !lib.elem source cfg.suppressedStorePaths) cfg.storePaths;
   };
 
 in {
@@ -172,7 +171,7 @@ in {
       description = ''
         Store paths to copy into the initrd as well.
       '';
-      type = with types; listOf (oneOf [ singleLineStr package ]);
+      type = utils.systemdUtils.types.initrdStorePath;
       default = [];
     };
 
@@ -491,7 +490,8 @@ in {
         "${pkgs.libfido2}/lib/libfido2.so.1"
       ] ++ optionals cfg.package.withKmod [
         "${pkgs.kmod.lib}/lib/libkmod.so.2"
-      ] ++ jobScripts;
+      ] ++ jobScripts
+      ++ map (c: builtins.removeAttrs c ["text"]) (builtins.attrValues cfg.contents);
 
       targets.initrd.aliases = ["default.target"];
       units =
--- a/nixos/modules/system/boot/systemd/journald.nix
+++ b/nixos/modules/system/boot/systemd/journald.nix
@@ -96,6 +96,7 @@ in {
       "systemd-journald@.service"
       "systemd-journal-flush.service"
       "systemd-journal-catalog-update.service"
+      "systemd-journald-sync@.service"
       ] ++ (optional (!config.boot.isContainer) "systemd-journald-audit.socket") ++ [
       "systemd-journald-dev-log.socket"
       "syslog.socket"
--- a/nixos/tests/systemd-networkd-ipv6-prefix-delegation.nix
+++ b/nixos/tests/systemd-networkd-ipv6-prefix-delegation.nix
@@ -40,7 +40,8 @@ import ./make-test-python.nix ({ pkgs, lib, ... }: {
             address = [
               "2001:DB8::1/64"
             ];
-            networkConfig.IPForward = true;
+            networkConfig.IPv4Forwarding = true;
+            networkConfig.IPv6Forwarding = true;
           };
         };
       };
--- a/nixos/tests/systemd-networkd-vrf.nix
+++ b/nixos/tests/systemd-networkd-vrf.nix
@@ -16,7 +16,8 @@ import ./make-test-python.nix ({ pkgs, lib, ... }: let
         linkConfig.RequiredForOnline = "no";
         networkConfig = {
           Address = "192.168.${toString vlan}.${toString id}/24";
-          IPForward = "yes";
+          IPv4Forwarding = "yes";
+          IPv6Forwarding = "yes";
         };
       };
     };
@@ -57,14 +58,16 @@ in {
 
         networks."10-vrf1" = {
           matchConfig.Name = "vrf1";
-          networkConfig.IPForward = "yes";
+          networkConfig.IPv4Forwarding = "yes";
+          networkConfig.IPv6Forwarding = "yes";
           routes = [
             { Destination = "192.168.1.2"; Metric = 100; }
           ];
         };
         networks."10-vrf2" = {
           matchConfig.Name = "vrf2";
-          networkConfig.IPForward = "yes";
+          networkConfig.IPv4Forwarding = "yes";
+          networkConfig.IPv6Forwarding = "yes";
           routes = [
             { Destination = "192.168.2.3"; Metric = 100; }
           ];
@@ -76,7 +79,8 @@ in {
           networkConfig = {
             VRF = "vrf1";
             Address = "192.168.1.1/24";
-            IPForward = "yes";
+            IPv4Forwarding = "yes";
+            IPv6Forwarding = "yes";
           };
         };
         networks."10-eth2" = {
@@ -85,7 +89,8 @@ in {
           networkConfig = {
             VRF = "vrf2";
             Address = "192.168.2.1/24";
-            IPForward = "yes";
+            IPv4Forwarding = "yes";
+            IPv6Forwarding = "yes";
           };
         };
       };
--- a/pkgs/build-support/kernel/make-initrd-ng/src/main.rs
+++ b/pkgs/build-support/kernel/make-initrd-ng/src/main.rs
@@ -1,4 +1,4 @@
-use std::collections::{HashSet, VecDeque};
+use std::collections::{BTreeSet, HashSet, VecDeque};
 use std::env;
 use std::ffi::{OsStr, OsString};
 use std::fs;
@@ -12,10 +12,41 @@ use eyre::Context;
 use goblin::{elf::Elf, Object};
 use serde::Deserialize;
 
+#[derive(PartialEq, Eq, PartialOrd, Ord, Clone, Copy, Debug, Deserialize, Hash)]
+#[serde(rename_all = "lowercase")]
+enum DLOpenPriority {
+    Required,
+    Recommended,
+    Suggested,
+}
+
+#[derive(PartialEq, Eq, Debug, Deserialize, Clone, Hash)]
+#[serde(rename_all = "camelCase")]
+struct DLOpenConfig {
+    use_priority: DLOpenPriority,
+    features: BTreeSet<String>,
+}
+
 #[derive(Deserialize, Debug)]
+struct DLOpenNote {
+    soname: Vec<String>,
+    feature: Option<String>,
+    // description is in the spec, but we don't need it here.
+    // description: Option<String>,
+    priority: Option<DLOpenPriority>,
+}
+
+#[derive(Deserialize, Debug, PartialEq, Eq, Hash, Clone)]
 struct StoreInput {
     source: String,
     target: Option<String>,
+    dlopen: Option<DLOpenConfig>,
+}
+
+#[derive(PartialEq, Eq, Hash, Clone)]
+struct StorePath {
+    path: Box<Path>,
+    dlopen: Option<DLOpenConfig>,
 }
 
 struct NonRepeatingQueue<T> {
@@ -48,13 +79,47 @@ impl<T: Clone + Eq + Hash> NonRepeatingQueue<T> {
     }
 }
 
-fn add_dependencies<P: AsRef<Path> + AsRef<OsStr>>(
+fn add_dependencies<P: AsRef<Path> + AsRef<OsStr> + std::fmt::Debug>(
     source: P,
     elf: Elf,
-    queue: &mut NonRepeatingQueue<Box<Path>>,
-) {
+    contents: &[u8],
+    dlopen: &Option<DLOpenConfig>,
+    queue: &mut NonRepeatingQueue<StorePath>,
+) -> eyre::Result<()> {
     if let Some(interp) = elf.interpreter {
-        queue.push_back(Box::from(Path::new(interp)));
+        queue.push_back(StorePath {
+            path: Box::from(Path::new(interp)),
+            dlopen: dlopen.clone(),
+        });
+    }
+
+    let mut dlopen_libraries = vec![];
+    if let Some(dlopen) = dlopen {
+        for n in elf
+            .iter_note_sections(&contents, Some(".note.dlopen"))
+            .into_iter()
+            .flatten()
+        {
+            let note = n.wrap_err_with(|| format!("bad note in {:?}", source))?;
+            // Payload is padded and zero terminated
+            let payload = &note.desc[..note
+                .desc
+                .iter()
+                .position(|x| *x == 0)
+                .unwrap_or(note.desc.len())];
+            let parsed = serde_json::from_slice::<Vec<DLOpenNote>>(payload)?;
+            for mut parsed_note in parsed {
+                if dlopen.use_priority
+                    >= parsed_note.priority.unwrap_or(DLOpenPriority::Recommended)
+                    || parsed_note
+                        .feature
+                        .map(|f| dlopen.features.contains(&f))
+                        .unwrap_or(false)
+                {
+                    dlopen_libraries.append(&mut parsed_note.soname);
+                }
+            }
+        }
     }
 
     let rpaths = if elf.runpaths.len() > 0 {
@@ -71,13 +136,21 @@ fn add_dependencies<P: AsRef<Path> + AsRef<OsStr>>(
         .map(|p| Box::<Path>::from(Path::new(p)))
         .collect::<Vec<_>>();
 
-    for line in elf.libraries {
+    for line in elf
+        .libraries
+        .into_iter()
+        .map(|s| s.to_string())
+        .chain(dlopen_libraries)
+    {
         let mut found = false;
         for path in &rpaths_as_path {
-            let lib = path.join(line);
+            let lib = path.join(&line);
             if lib.exists() {
                 // No need to recurse. The queue will bring it back round.
-                queue.push_back(Box::from(lib.as_path()));
+                queue.push_back(StorePath {
+                    path: Box::from(lib.as_path()),
+                    dlopen: dlopen.clone(),
+                });
                 found = true;
                 break;
             }
@@ -92,6 +165,8 @@ fn add_dependencies<P: AsRef<Path> + AsRef<OsStr>>(
             );
         }
     }
+
+    Ok(())
 }
 
 fn copy_file<
@@ -100,7 +175,8 @@ fn copy_file<
 >(
     source: P,
     target: S,
-    queue: &mut NonRepeatingQueue<Box<Path>>,
+    dlopen: &Option<DLOpenConfig>,
+    queue: &mut NonRepeatingQueue<StorePath>,
 ) -> eyre::Result<()> {
     fs::copy(&source, &target)
         .wrap_err_with(|| format!("failed to copy {:?} to {:?}", source, target))?;
@@ -109,7 +185,7 @@ fn copy_file<
         fs::read(&source).wrap_err_with(|| format!("failed to read from {:?}", source))?;
 
     if let Ok(Object::Elf(e)) = Object::parse(&contents) {
-        add_dependencies(source, e, queue);
+        add_dependencies(source, e, &contents, &dlopen, queue)?;
 
         // Make file writable to strip it
         let mut permissions = fs::metadata(&target)
@@ -138,14 +214,18 @@ fn copy_file<
 
 fn queue_dir<P: AsRef<Path> + std::fmt::Debug>(
     source: P,
-    queue: &mut NonRepeatingQueue<Box<Path>>,
+    dlopen: &Option<DLOpenConfig>,
+    queue: &mut NonRepeatingQueue<StorePath>,
 ) -> eyre::Result<()> {
     for entry in
         fs::read_dir(&source).wrap_err_with(|| format!("failed to read dir {:?}", source))?
     {
         let entry = entry?;
         // No need to recurse. The queue will bring us back round here on its own.
-        queue.push_back(Box::from(entry.path().as_path()));
+        queue.push_back(StorePath {
+            path: Box::from(entry.path().as_path()),
+            dlopen: dlopen.clone(),
+        });
     }
 
     Ok(())
@@ -153,12 +233,12 @@ fn queue_dir<P: AsRef<Path> + std::fmt::Debug>(
 
 fn handle_path(
     root: &Path,
-    p: &Path,
-    queue: &mut NonRepeatingQueue<Box<Path>>,
+    p: StorePath,
+    queue: &mut NonRepeatingQueue<StorePath>,
 ) -> eyre::Result<()> {
     let mut source = PathBuf::new();
     let mut target = Path::new(root).to_path_buf();
-    let mut iter = p.components().peekable();
+    let mut iter = p.path.components().peekable();
     while let Some(comp) = iter.next() {
         match comp {
             Component::Prefix(_) => panic!("This tool is not meant for Windows"),
@@ -182,7 +262,7 @@ fn handle_path(
                     .wrap_err_with(|| format!("failed to get symlink metadata for {:?}", source))?
                     .file_type();
                 if typ.is_file() && !target.exists() {
-                    copy_file(&source, &target, queue)?;
+                    copy_file(&source, &target, &p.dlopen, queue)?;
 
                     if let Some(filename) = source.file_name() {
                         source.set_file_name(OsString::from_iter([
@@ -193,7 +273,10 @@ fn handle_path(
 
                         let wrapped_path = source.as_path();
                         if wrapped_path.exists() {
-                            queue.push_back(Box::from(wrapped_path));
+                            queue.push_back(StorePath {
+                                path: Box::from(wrapped_path),
+                                dlopen: p.dlopen.clone(),
+                            });
                         }
                     }
                 } else if typ.is_symlink() {
@@ -213,7 +296,10 @@ fn handle_path(
                     }
                     let link_target_path = source.as_path();
                     if link_target_path.exists() {
-                        queue.push_back(Box::from(link_target_path));
+                        queue.push_back(StorePath {
+                            path: Box::from(link_target_path),
+                            dlopen: p.dlopen.clone(),
+                        });
                     }
                     break;
                 } else if typ.is_dir() {
@@ -224,7 +310,7 @@ fn handle_path(
 
                     // Only recursively copy if the directory is the target object
                     if iter.peek().is_none() {
-                        queue_dir(&source, queue)
+                        queue_dir(&source, &p.dlopen, queue)
                             .wrap_err_with(|| format!("failed to queue dir {:?}", source))?;
                     }
                 }
@@ -244,11 +330,14 @@ fn main() -> eyre::Result<()> {
     let output = &args[2];
     let out_path = Path::new(output);
 
-    let mut queue = NonRepeatingQueue::<Box<Path>>::new();
+    let mut queue = NonRepeatingQueue::<StorePath>::new();
 
     for sp in input {
         let obj_path = Path::new(&sp.source);
-        queue.push_back(Box::from(obj_path));
+        queue.push_back(StorePath {
+            path: Box::from(obj_path),
+            dlopen: sp.dlopen,
+        });
         if let Some(target) = sp.target {
             println!("{} -> {}", &target, &sp.source);
             // We don't care about preserving symlink structure here
@@ -264,7 +353,7 @@ fn main() -> eyre::Result<()> {
         }
     }
     while let Some(obj) = queue.pop_front() {
-        handle_path(out_path, &*obj, &mut queue)?;
+        handle_path(out_path, obj, &mut queue)?;
     }
 
     Ok(())
--- a/pkgs/build-support/kernel/make-initrd-ng/src/main.rs
+++ b/pkgs/build-support/kernel/make-initrd-ng/src/main.rs
@@ -3,7 +3,6 @@ use std::env;
 use std::ffi::{OsStr, OsString};
 use std::fs;
 use std::hash::Hash;
-use std::io::{BufRead, BufReader};
 use std::iter::FromIterator;
 use std::os::unix;
 use std::path::{Component, Path, PathBuf};
@@ -11,6 +10,13 @@ use std::process::Command;
 
 use eyre::Context;
 use goblin::{elf::Elf, Object};
+use serde::Deserialize;
+
+#[derive(Deserialize, Debug)]
+struct StoreInput {
+    source: String,
+    target: Option<String>,
+}
 
 struct NonRepeatingQueue<T> {
     queue: VecDeque<T>,
@@ -231,26 +237,23 @@ fn handle_path(
 
 fn main() -> eyre::Result<()> {
     let args: Vec<String> = env::args().collect();
-    let input =
-        fs::File::open(&args[1]).wrap_err_with(|| format!("failed to open file {:?}", &args[1]))?;
+    let contents =
+        fs::read(&args[1]).wrap_err_with(|| format!("failed to open file {:?}", &args[1]))?;
+    let input = serde_json::from_slice::<Vec<StoreInput>>(&contents)
+        .wrap_err_with(|| format!("failed to parse JSON in {:?}", &args[1]))?;
     let output = &args[2];
     let out_path = Path::new(output);
 
     let mut queue = NonRepeatingQueue::<Box<Path>>::new();
 
-    let mut lines = BufReader::new(input).lines();
-    while let Some(obj) = lines.next() {
-        // Lines should always come in pairs
-        let obj = obj?;
-        let sym = lines.next().unwrap()?;
-
-        let obj_path = Path::new(&obj);
+    for sp in input {
+        let obj_path = Path::new(&sp.source);
         queue.push_back(Box::from(obj_path));
-        if !sym.is_empty() {
-            println!("{} -> {}", &sym, &obj);
+        if let Some(target) = sp.target {
+            println!("{} -> {}", &target, &sp.source);
             // We don't care about preserving symlink structure here
             // nearly as much as for the actual objects.
-            let link_string = format!("{}/{}", output, sym);
+            let link_string = format!("{}/{}", output, target);
             let link_path = Path::new(&link_string);
             let mut link_parent = link_path.to_path_buf();
             link_parent.pop();
--- a/pkgs/build-support/setup-hooks/auto-patchelf.py
+++ b/pkgs/build-support/setup-hooks/auto-patchelf.py
@@ -11,7 +11,7 @@ from contextlib import contextmanager
 from dataclasses import dataclass
 from itertools import chain
 from pathlib import Path, PurePath
-from typing import DefaultDict, Iterator, List, Optional, Set, Tuple
+from typing import DefaultDict, Iterator, Optional
 
 from elftools.common.exceptions import ELFError  # type: ignore
 from elftools.elf.dynamic import DynamicSection  # type: ignore
@@ -38,7 +38,7 @@ def is_dynamic_executable(elf: ELFFile) -> bool:
     return bool(elf.get_section_by_name(".interp"))
 
 
-def get_dependencies(elf: ELFFile) -> List[str]:
+def get_dependencies(elf: ELFFile) -> list[str]:
     dependencies = []
     # This convoluted code is here on purpose. For some reason, using
     # elf.get_section_by_name(".dynamic") does not always return an
@@ -52,7 +52,7 @@ def get_dependencies(elf: ELFFile) -> List[str]:
     return dependencies
 
 
-def get_rpath(elf: ELFFile) -> List[str]:
+def get_rpath(elf: ELFFile) -> list[str]:
     # This convoluted code is here on purpose. For some reason, using
     # elf.get_section_by_name(".dynamic") does not always return an
     # instance of DynamicSection, but that is required to call iter_tags
@@ -119,11 +119,11 @@ def glob(path: Path, pattern: str, recursive: bool) -> Iterator[Path]:
         return [path] if path.match(pattern) else []
 
 
-cached_paths: Set[Path] = set()
-soname_cache: DefaultDict[Tuple[str, str], List[Tuple[Path, str]]] = defaultdict(list)
+cached_paths: set[Path] = set()
+soname_cache: DefaultDict[tuple[str, str], list[tuple[Path, str]]] = defaultdict(list)
 
 
-def populate_cache(initial: List[Path], recursive: bool =False) -> None:
+def populate_cache(initial: list[Path], recursive: bool =False) -> None:
     lib_dirs = list(initial)
 
     while lib_dirs:
@@ -174,7 +174,7 @@ class Dependency:
     found: bool = False     # Whether it was found somewhere
 
 
-def auto_patchelf_file(path: Path, runtime_deps: list[Path], append_rpaths: List[Path] = [], extra_args: List[str] = []) -> list[Dependency]:
+def auto_patchelf_file(path: Path, runtime_deps: list[Path], append_rpaths: list[Path] = [], extra_args: list[str] = []) -> list[Dependency]:
     try:
         with open_elf(path) as elf:
 
@@ -257,13 +257,13 @@ def auto_patchelf_file(path: Path, runtime_deps: list[Path], append_rpaths: List
 
 
 def auto_patchelf(
-        paths_to_patch: List[Path],
-        lib_dirs: List[Path],
-        runtime_deps: List[Path],
+        paths_to_patch: list[Path],
+        lib_dirs: list[Path],
+        runtime_deps: list[Path],
         recursive: bool = True,
-        ignore_missing: List[str] = [],
-        append_rpaths: List[Path] = [],
-        extra_args: List[str] = []) -> None:
+        ignore_missing: list[str] = [],
+        append_rpaths: list[Path] = [],
+        extra_args: list[str] = []) -> None:
 
     if not paths_to_patch:
         sys.exit("No paths to patch, stopping.")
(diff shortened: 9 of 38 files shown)


lukaslihotzki:
With this PR, the generated initrd contains a `libsystemd-shared-256.so` that contains the string `/nix/store/eeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee-kmod-31-lib/lib/libkmod.so.2`. The cause is probably that systemd uses dlopen instead of dynamic linking for `libkmod.so.2`.

This is an invalid path, so libkmod cannot be loaded, so the initrd fails to load any kernel module.

lukaslihotzki:
This PR works for me when using a kernel that has everything built-in which is needed to boot, so no modules need to be loaded in stage 1. Therefore, kmod seems to be the only major thing that is currently broken.

nikstur:
Should be fixed now for the systemd initrd. We needed to explicitly include libkmod in the initrd. 

The scripted initrd, however, is not working with the updated systemd and I have no idea how to fix it since I never use the scripted initrd.

MinerSebas:
The changelog contains this entry:
```
    * The behavior of systemd-sleep and systemd-homed has been updated to
      freeze user sessions when entering the various sleep modes or when
      locking a homed-managed home area. This is known to cause issues with
      the proprietary NVIDIA drivers. Packagers of the NVIDIA proprietary
      drivers may want to add drop-in configuration files that set
      SYSTEMD_SLEEP_FREEZE_USER_SESSION=false for systemd-suspend.service
      and related services, and SYSTEMD_HOME_LOCK_FREEZE_SESSION=false for
      systemd-homed.service.
```
but this PR does not set the mentioned configuration values when using the proprietary NVIDIA drivers.

lukaslihotzki:
Indeed, this PR works with mandatory kernel modules when setting `boot.initrd.systemd.enable = true;`. :tada:

For merging this, either the scripted initrd needs to be fixed, or https://github.com/NixOS/nixpkgs/issues/287308 needs to be completed. Both options seem to be difficult. For fixing scripted initrd, this is a start:
```
diff --git a/nixos/modules/system/boot/stage-1.nix b/nixos/modules/system/boot/stage-1.nix
index ae05bc5ae88c..20630e686580 100644
--- a/nixos/modules/system/boot/stage-1.nix
+++ b/nixos/modules/system/boot/stage-1.nix
@@ -171,6 +171,9 @@ let
       # Copy ld manually since it isn't detected correctly
       cp -pv ${pkgs.stdenv.cc.libc.out}/lib/ld*.so.? $out/lib

+      # Copy libkmod because it is dlopened by systemd >=256
+      cp -pv ${pkgs.kmod.lib}/lib/libkmod.so.? $out/lib
+
       # Copy all of the needed libraries in a consistent order so
       # duplicates are resolved the same way.
       find $out/bin $out/lib -type f | sort | while read BIN; do
```

This includes libkmod.so.2 in the initrd, but still fails at runtime.

nikstur:
The way the scripted initrd patchelfs binaries is incompatible with the way we replace the `.so` references with absolute paths in our systemd derivation. This didn't matter in the past because `libkmod` (needed by `udevadm`) wasn't a dlopen dependency in previous systemd versions.

flokli:
@nikstur can you rebase this?

tulilirockz:
Systemd 256 officially released in systemd-stable, should be able to use that now

jmbaur:
Looks like stable releases will be made from one repo now: https://github.com/systemd/systemd/blob/2af17b5e4c1aa67ed5bcaa105a2a36d4fac9061a/NEWS#L117

codebam:
It builds in-tree but not when overlayed in my nixos configuration.

```
✦ ❯ doas nixos-rebuild switch                                                                                                                                                                                                                                                                                      
doas (codebam@nixos-desktop) password:                                                                                                                                                                                                                                                                             
warning: Git tree '/home/codebam/git/nixos' is dirty                                                                                                                                                                                                                                                               
building the system configuration...                                                                                                                                                                                                                                                                               
warning: Git tree '/home/codebam/git/nixos' is dirty                                                                                                                                                                                                                                                               
error: builder for '/nix/store/6c2nir0bnmlc3zfcqfnl4impdmd9ijx1-libusb-1.0.27.drv' failed with exit code 1;                                                                                                                                                                                                        
       last 10 log lines:                                                                                                                                                                                                                                                                                          
       > checking for nfds_t... yes                                                                                                                                                                                                                                                                                
       > checking for pipe2... yes                                                                                                                                                                                                                                                                                 
       > checking if gcc recognizes -pthread... yes                                                                                                                                                                                                                                                                
       > checking for library containing pthread_create... none required
       > checking whether __atomic_load_n is supported... yes
       > checking for library containing clock_gettime... none required
       > checking for pthread_setname_np... yes
       > checking for libudev.h... yes
       > checking for udev_new in -ludev... no
       > configure: error: udev support requested but libudev not installed
       For full logs, run 'nix log /nix/store/6c2nir0bnmlc3zfcqfnl4impdmd9ijx1-libusb-1.0.27.drv'.
error (ignored): error: cannot unlink '/tmp/nix-build-dbus-1.14.10.drv-1': Directory not empty
error: 1 dependencies of derivation '/nix/store/wwc9gq8225z6bpzmf090v4zkivif4fqj-pipewire-1.0.7.drv' failed to build
error: 1 dependencies of derivation '/nix/store/b2g8h0g7sxmhg7nff75acp56miyclvsf-etc-alsa-conf.d-49-pipewire-modules.conf.drv' failed to build
error (ignored): error: cannot unlink '/tmp/nix-build-emacs-29.3.drv-1/source': Directory not empty
error: 1 dependencies of derivation '/nix/store/kx6sr5lbsr4qx4c1p42dclnnl9rq3d95-etc.drv' failed to build
error (ignored): error: cannot unlink '/tmp/nix-build-libgphoto2-2.5.31.drv-0/source': Directory not empty
error (ignored): error: cannot unlink '/tmp/nix-build-libnvme-1.9.drv-0/source/build/doc': Directory not empty
error (ignored): error: cannot unlink '/tmp/nix-build-texinfo-interactive-7.0.3.drv-0/texinfo-7.0.3': Directory not empty
error (ignored): error: cannot unlink '/tmp/nix-build-avahi-0.8.drv-0/avahi-0.8': Directory not empty
error (ignored): error: cannot unlink '/tmp/nix-build-openldap-2.6.7.drv-0/openldap-2.6.7': Directory not empty
error: 1 dependencies of derivation '/nix/store/g0c9af2wkv8d8i7lpjp86j69dqy3kl59-nixos-system-nixos-desktop-24.11.20240614.1740b0f.drv' failed to build
```

Naxdy:
FYI, there's a pretty nasty bug / non-obvious behavior in upstream systemd involving the invocation of `systemd-tmpfiles --purge`: https://github.com/systemd/systemd/issues/33349

sparky8251:
@Naxdy that issue is pretty clearly one thats been around for a long time and just now reported reading the comments. No behavior has changed, its just that someone got bit by bad documentation when using an option to delete things and now they are discussing how to make it so it doesnt happen to others in the future.

Naxdy:
@sparky8251 define "long time", `--purge` was added during the 256 dev cycle as far as I'm aware

overall agree though, just posted it here for awareness mainly

oxalica:
> YI, there's a pretty nasty bug / non-obvious behavior in upstream systemd 256 involving the invocation of `systemd-tmpfiles --purge`: [systemd/systemd#33349](https://github.com/systemd/systemd/issues/33349)

It's resolved as 256.1 is out. So should this be unblocked then?

arianvp:
> It builds in-tree but not when overlayed in my nixos configuration.

What does this mean exactly?  You have to provide more details.   Did you rebase the PR And do a full rebuild? did you do a `systemd.package` ?  Did you use `nixpkgs.overlays`? If so; what nixpkgs version are you using?

FWIW I rebased the PR on `master` and `libusb` builds fine for me. Can't reproduce.

```
[arian@utm:/mnt/Projects/nixpkgs]$ nix-build -A libusb
/nix/store/s7xnjq2dz7znp0q0kp5p1b9b7ch2jz1y-libusb-1.0.27
```

benaryorg:
Just quickly throwing in here that this new version of systemd ships with [a fix for systemd-vconsole-setup throwing fatal errors](https://github.com/systemd/systemd/commit/e9bdbb6bbcd4725171ca1fd0bb4fff83e5338acf); in case anyone else stumbles upon that service err'ing out with `Failed to lock /dev/console: Input/output error`, then it's most likely that you have a `console=` cmdline argument that points to the wrong device, however fixing that across a diverse fleet of machines may be annoying so at least this update should silence that and at least not cause a degraded system state.

(just throwing this in here so that people searching for the error message will actually find something)

codebam:
> What does this mean exactly?

I used `nixpkgs.overlays` using nixos-unstable-small 24.11

arianvp:
I ran into another issue:

```
warning: Git tree '/mnt/Projects/personal/nixos-stuff' is dirty
updating systemd-boot from 255.6 to 256
Copied "/nix/store/9j7qf00y1936vwynxgq37r7zh2b01w6v-systemd-256/lib/systemd/boot/efi/systemd-bootaa64.efi" to "/boot/EFI/systemd/systemd-bootaa64.efi".
Copied "/nix/store/9j7qf00y1936vwynxgq37r7zh2b01w6v-systemd-256/lib/systemd/boot/efi/systemd-bootaa64.efi" to "/boot/EFI/BOOT/BOOTAA64.EFI".
Skipping "/boot/EFI/BOOT/BOOTAA64.EFI", same boot loader version in place already.
  File "/nix/store/kma88rn7vwxv3x3lc6x8fqygz1xnm8pq-p17ampilnd20xlbq991qxkycf8gfqgbv-systemd-boot", line 393, in <module>
    main()
  File "/nix/store/kma88rn7vwxv3x3lc6x8fqygz1xnm8pq-p17ampilnd20xlbq991qxkycf8gfqgbv-systemd-boot", line 376, in main
    install_bootloader(args)
  File "/nix/store/kma88rn7vwxv3x3lc6x8fqygz1xnm8pq-p17ampilnd20xlbq991qxkycf8gfqgbv-systemd-boot", line 314, in install_bootloader
    subprocess.check_call([f"{SYSTEMD}/bin/bootctl", f"--esp-path={EFI_SYS_MOUNT_POINT}"] + bootctl_flags + ["update"])
  File "/nix/store/q3x28mimkawkdjlvd78jxv3s0fk25vz8-python3-3.11.9/lib/python3.11/subprocess.py", line 413, in check_call
    raise CalledProcessError(retcode, cmd)
subprocess.CalledProcessError: Command '['/nix/store/9j7qf00y1936vwynxgq37r7zh2b01w6v-systemd-256/bin/bootctl', '--esp-path=/boot', 'update']' returned non-zero exit status 1.
```

codebam:
It builds and installs with `systemd.package` but doesn't find my boot device

flokli:
> It builds and installs with `systemd.package` but doesn't find my boot device

During the switch? Or during bootup? Logs, or a picture would help...

arianvp:
> It builds and installs with systemd.package but doesn't find my boot device

Please try with `boot.initrd.systemd.enable = true;`.  As you can see in the rest of the thread;  scripted initrd is currently broken due to it not finding `kmod`. We're trying to come up with a fix.

arianvp:
The simplest albeit hacky fix I can think of to get scripted initrd working again is to add

`-e ${pkgs.kmod}`

here: https://github.com/NixOS/nixpkgs/blob/0ed51f7f33bcc4cfcb00ecc501da6fbd57c8f686/nixos/modules/system/boot/stage-1.nix#L193

so that the ref doesn't get nuked. 
/
It doesn't generically solve the problem but maybe it's good enough for now to get things working again?

codebam:
> Please try with `boot.initrd.systemd.enable = true;`

photo1718810108

r-vdp:
@codebam add the following so that you get an initrd shell and check the logs to know what went wrong: 

```
boot = {
    kernelParams = [
        # Boot into a shell in initrd
        #"rd.systemd.unit=rescue.target"
    ];
    initrd.systemd.emergencyAccess = true;
};
```

arianvp:
Specifically the logs of `systemd-modules-load.service` would be helpful

codebam:
Keyboard isn't working, so I can't press ctrl-d or enter.

arianvp:
You could try setting `systemd.journald.forward_to_console=1` on the kernel command line.

codebam:
> You could try setting `systemd.journald.forward_to_console` on the kernel command line and see if you can get the logs that way

photo1718829370

arianvp:
Just double-checking: Are you using the NixOS module changes to systemd-initrd that are present in this PR? As this is exactly the error that is fixed by that.

codebam:
> Just double-checking: Are you using the NixOS module changes to systemd-initrd that are present in this PR?

I'm not sure. I just set systemd.package

philiptaron:
Per the [above notification](https://github.com/NixOS/nixpkgs/pull/307068#ref-commit-92d9db6), I'm running this on my machine and it's going swimmingly. I do use `boot.initrd.systemd.root = "gpt-auto"` in order to get `systemd-gpt-auto-root` to detect the root filesystem partition, so consider that particular setting tested and working as well.

pkgs/os-specific/linux/systemd/0017-meson.build-do-not-create-systemdstatedir.patch
```diff
@@ -8,14 +8,14 @@ Subject: [PATCH] meson.build: do not create systemdstatedir
...
 +++ b/meson.build
-@@ -2497,7 +2497,6 @@ install_data('LICENSE.GPL2',
+@@ -2652,7 +2652,6 @@ install_data('LICENSE.GPL2',
  install_subdir('LICENSES',
                 install_dir : docdir)
  
 -install_emptydir(systemdstatedir)
  
- ############################################################
+ #####################################################################
```
philiptaron, line 20: It got longer somehow.

pkgs/os-specific/linux/systemd/0018-timesyncd-disable-NSCD-when-DNSSEC-validation-is-dis.patch
```diff
@@ -1,4 +1,4 @@
-From 7a27556920fe1feefd17096841c8f3ca1294a1b3 Mon Sep 17 00:00:00 2001
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
```
philiptaron, line 1: What's going on with this change?

pkgs/os-specific/linux/systemd/default.nix
```diff
@@ -543,6 +552,11 @@ stdenv.mkDerivation (finalAttrs: {
     (lib.mesonOption "mount-path" "${lib.getOutput "mount" util-linux}/bin/mount")
     (lib.mesonOption "umount-path" "${lib.getOutput "mount" util-linux}/bin/umount")
 
+    # SSH
+    # Disabled for now until someone makes this work.
+    (lib.mesonOption "sshconfdir" "no")
+    (lib.mesonOption "sshdconfdir" "no")
```
philiptaron, line 558: # 👀 
[Interesting feature](https://www.freedesktop.org/software/systemd/man/devel/systemd-ssh-proxy.html).

philiptaron (approved):
Lots of patch management! Thanks for keeping track of code movement @nikstur.

pkgs/os-specific/linux/systemd/0018-timesyncd-disable-NSCD-when-DNSSEC-validation-is-dis.patch
arianvp: I think the instructions were not followed at some point.

We have

```
`git -c format.signoff=false format-patch v${version} --no-numbered --zero-commit --no-signature`.
```

which zeros out the commit numbers.   I guess the instructions were not followed in 255

deviantsemicolon:
systemd 256.1 is out. I would recommend updating this pr accordingly

pkgs/os-specific/linux/systemd/default.nix
astro: You can have the files with a small patch but I am not sure that they make sense without the accompanying VM setup. They ought to be linked into the system's /etc.

```patch
diff --git pkgs/os-specific/linux/systemd/default.nix pkgs/os-specific/linux/systemd/default.nix
index 09739968e55d..60be69acb602 100644
--- pkgs/os-specific/linux/systemd/default.nix
+++ pkgs/os-specific/linux/systemd/default.nix
@@ -554,8 +554,8 @@ stdenv.mkDerivation (finalAttrs: {

     # SSH
     # Disabled for now until someone makes this work.
-    (lib.mesonOption "sshconfdir" "no")
-    (lib.mesonOption "sshdconfdir" "no")
+    (lib.mesonOption "sshconfdir" "${placeholder "out"}/etc/ssh")
+    (lib.mesonOption "sshdconfdir" "${placeholder "out"}/etc/ssh")

```

philiptaron:
https://github.com/philiptaron/nixpkgs/tree/nikstur/systemd-256

That branch has v256.1 and rebased on current master. It's a bit of a compile-fest but I'm running it on my actual hardware machine (with systemd stage 1!) and I haven't regretted it yet.

deviantsemicolon:
Is there an ETA on merging this?

arianvp:
We're working on trying to fix the initrd still. We've got a way forward we
all agreed upon and have a working branch that still needs some cleanup.

ElvishJerricco:
@layus The changes to `auto-patchelf.py` might be a little clumsy; I'm not great with python. So some review on that would be appreciated. The intention is to add support for adding `.note.dlopen`-based library dependencies to rpath.

nixos/modules/system/boot/stage-1.nix
```diff
@@ -131,6 +131,7 @@ let
 
       # Copy udev.
       copy_bin_and_libs ${udev}/bin/udevadm
+      cp ${lib.getLib udev.kmod}/lib/libkmod.so* $out/lib
```
philiptaron, line 134: Fun fact: `CONFIG_MODPROBE_PATH`, when the kernel is built, controls what this path is.

pkgs/build-support/setup-hooks/auto-patchelf.py
```diff
@@ -40,21 +40,21 @@ def is_dynamic_executable(elf: ELFFile) -> bool:
     return bool(elf.get_section_by_name(".interp"))
 
 
-def get_dependencies(elf: ELFFile) -> List[str]:
+def get_dependencies(elf: ELFFile) -> List[List[str]]:
```
philiptaron, line 43:
```suggestion
def get_dependencies(elf: ELFFile) -> list[list[Path]]:
```

pkgs/build-support/setup-hooks/auto-patchelf.py
```diff
@@ -40,21 +40,21 @@ def is_dynamic_executable(elf: ELFFile) -> bool:
...
             for tag in section.iter_tags('DT_NEEDED'):
-                dependencies.append(tag.needed)
+                dependencies.append([ tag.needed ])
             break # There is only one dynamic section
 
     return dependencies
 
 
-def get_dlopen_dependencies(elf: ELFFile) -> List[str]:
+def get_dlopen_dependencies(elf: ELFFile) -> List[List[str]]:
```
philiptaron, line 57:
```suggestion
def get_dlopen_dependencies(elf: ELFFile) -> list[list[Path]]:
```

pkgs/build-support/setup-hooks/auto-patchelf.py
```diff
@@ -40,21 +40,21 @@ def is_dynamic_executable(elf: ELFFile) -> bool:
...
+def get_dependencies(elf: ELFFile) -> List[List[str]]:
     dependencies = []
     # This convoluted code is here on purpose. For some reason, using
     # elf.get_section_by_name(".dynamic") does not always return an
     # instance of DynamicSection, but that is required to call iter_tags
     for section in elf.iter_sections():
         if isinstance(section, DynamicSection):
             for tag in section.iter_tags('DT_NEEDED'):
-                dependencies.append(tag.needed)
+                dependencies.append([ tag.needed ])
```
philiptaron, line 51:
```suggestion
                dependencies.append([Path(tag.needed)])
```

pkgs/build-support/setup-hooks/auto-patchelf.py
```diff
@@ -71,8 +71,7 @@ def get_dlopen_dependencies(elf: ELFFile) -> List[str]:
             text = note_desc.decode("utf-8").rstrip("\0")
             j = json.loads(text)
             for d in j:
-                for soname in d["soname"]:
-                    dependencies.append(soname)
+                dependencies.append(d["soname"])
```
philiptaron, line 74:
```suggestion
                dependencies.append([Path(soname) for soname in d["soname"]])
```

pkgs/build-support/setup-hooks/auto-patchelf.py
```diff
@@ -228,7 +227,9 @@ def auto_patchelf_file(path: Path, runtime_deps: list[Path], append_rpaths: List
 
             file_is_dynamic_executable = is_dynamic_executable(elf)
 
-            file_dependencies = map(Path, get_dependencies(elf) + get_dlopen_dependencies(elf))
+            # Each dependency is represented as a list of permissible sonames.
+            # .note.dlopen allows for one of several candidate sonames.
+            file_dependencies = [[Path(soname) for soname in dep] for dep in get_dependencies(elf) + get_dlopen_dependencies(elf)]
```
philiptaron, line 232:
```suggestion
            file_dependencies = get_dependencies(elf) + get_dlopen_dependencies(elf)
```

pkgs/build-support/setup-hooks/auto-patchelf.py
```diff
@@ -247,20 +248,25 @@ def auto_patchelf_file(path: Path, runtime_deps: list[Path], append_rpaths: List
...
-            # checked (because just updating the rpath won't satisfy
-            # it).
-            continue
+        found_dependency = None
+        for candidate in dep:
+            if candidate.is_absolute() and candidate.is_file():
+                # This is an absolute path. If it exists, just use it.
+                # Otherwise, we probably want this to produce an error when
+                # checked (because just updating the rpath won't satisfy
+                # it).
```
philiptaron, line 257: I think this file is formatted at 100 columns wide.
```suggestion
                # checked (because just updating the rpath won't satisfy it).
```

pkgs/build-support/setup-hooks/auto-patchelf.py
```diff
@@ -247,20 +248,25 @@ def auto_patchelf_file(path: Path, runtime_deps: list[Path], append_rpaths: List
...
-        if found_dependency := find_dependency(dep.name, file_arch, file_osabi):
-            rpath.append(found_dependency)
-            dependencies.append(Dependency(path, dep, True))
-            print(f"    {dep} -> found: {found_dependency}")
-        else:
-            dependencies.append(Dependency(path, dep, False))
-            print(f"    {dep} -> not found!")
+            if found_dependency := find_dependency(candidate.name, file_arch, file_osabi):
+                rpath.append(found_dependency)
+                dependencies.append(Dependency(path, candidate, True))
```
philiptaron, line 262: Best practice for readability: name the boolean parameters with keyword args at the call site.
```suggestion
                dependencies.append(Dependency(path, candidate, found=True))
```

pkgs/build-support/setup-hooks/auto-patchelf.py
```diff
@@ -247,20 +248,25 @@ def auto_patchelf_file(path: Path, runtime_deps: list[Path], append_rpaths: List
...
+                print(f"    {candidate} -> found: {found_dependency}")
+                break
+
+        if found_dependency == None:
```
philiptaron, line 266: Typically, you want `is None` not `== None` because that's a tighter bound.
```suggestion
        if found_dependency is None:
```

pkgs/build-support/setup-hooks/auto-patchelf.py
```diff
@@ -247,20 +248,25 @@ def auto_patchelf_file(path: Path, runtime_deps: list[Path], append_rpaths: List
...
+            dep_name = dep[0] if len(dep) == 1 else "any(%s)" % ", ".join([str(soname) for soname in dep])
```
philiptaron, line 267: Little code golf:
```suggestion
            dep_name = dep[0] if len(dep) == 1 else f"any({', '.join(map(str, dep))})"
```

pkgs/build-support/setup-hooks/auto-patchelf.py
```diff
@@ -247,20 +248,25 @@ def auto_patchelf_file(path: Path, runtime_deps: list[Path], append_rpaths: List
...
+            dependencies.append(Dependency(path, dep_name, False))
```
philiptaron, line 268:
```suggestion
            dependencies.append(Dependency(path, dep_name, found=False))
```

pkgs/build-support/kernel/make-initrd-ng/src/main.rs
```diff
@@ -231,26 +237,20 @@ fn handle_path(
 
 fn main() -> eyre::Result<()> {
     let args: Vec<String> = env::args().collect();
-    let input =
-        fs::File::open(&args[1]).wrap_err_with(|| format!("failed to open file {:?}", &args[1]))?;
+    let input = serde_json::from_slice::<Vec<StoreInput>>(&fs::read(&args[1])?)?;
```
philiptaron, line 307: The exact file that failed to parse looks interesting (as it did in the previous version)
```suggestion
    let contents = fs::read(&args[1]).wrap_err_with(|| format!("failed to open file {:?}", &args[1]))?;
    let input = serde_json::from_slice::<Vec<StoreInput>>(&contents).wrap_err_with(|| format!("failed to parse JSON in {:?}", &args[1])?;

File: pkgs/build-support/kernel/make-initrd-ng/src/main.rs [hunk 16]
```diff
@@ -1,4 +1,4 @@
-use std::collections::{HashSet, VecDeque};
+use std::collections::{HashSet, VecDeque, BTreeSet};
```
philiptaron, line 1: Please run `rustfmt` on this source; it would sort this first.

pkgs/build-support/kernel/make-initrd-ng/src/main.rs
```diff
@@ -12,10 +12,40 @@ use eyre::Context;
...
+struct DLOpenConfig {
+    use_priority: DLOpenPriority,
+    features: BTreeSet<String>,
+}
+
 #[derive(Deserialize, Debug)]
+struct DLOpenNote {
+    soname: Vec<String>,
+    feature: Option<String>,
+    // description: Option<String>,
```
philiptaron, line 34: Remove?

pkgs/build-support/kernel/make-initrd-ng/src/main.rs
```diff
@@ -48,13 +78,34 @@ impl<T: Clone + Eq + Hash> NonRepeatingQueue<T> {
...
-fn add_dependencies<P: AsRef<Path> + AsRef<OsStr>>(
+fn add_dependencies<P: AsRef<Path> + AsRef<OsStr> + std::fmt::Debug>(
     source: P,
     elf: Elf,
-    queue: &mut NonRepeatingQueue<Box<Path>>,
-) {
+    contents: &[u8],
+    dlopen: &Option<DLOpenConfig>,
+    queue: &mut NonRepeatingQueue<StorePath>,
+) -> eyre::Result<()> {
```
philiptaron, line 87: The use of mutation and error returns makes me queasy, since this function now can both return an error and mutate its inputs. Since I see that's the pattern with most of the other functions in this file, it's fine, but as a general rule, it's better to do one validation pass (which returns _all_ the errors that could occur, rather than just the first!), then have the rest of the code assert and assume.

philiptaron (commented):
I did a patch-by-patch review. No blocking comments. I'm doing a full build of my machine with these patches applied and we'll see what the result is in a bit.

pkgs/build-support/kernel/make-initrd-ng/src/main.rs
ElvishJerricco: This is mainly a nod to the spec, which does include a description field that we don't need. I'd rather leave a comment explaining that than leave it out.

pkgs/build-support/kernel/make-initrd-ng/src/main.rs
ElvishJerricco: Such a refactor would be welcome :) I don't think it's strictly necessary for this PR though.

nixos/modules/system/boot/stage-1.nix
ElvishJerricco: That's for the `modprobe` binary.

ElvishJerricco:
@philiptaron I *believe* I have addressed / applied all your suggestions. Please double check me.

nixos/modules/system/boot/stage-1.nix
philiptaron: Right you are.

philiptaron (approved):
I'm still building it, but all the code changes look fine to me.

pkgs/build-support/setup-hooks/auto-patchelf.py
```diff
@@ -223,24 +246,24 @@ def auto_patchelf_file(path: Path, runtime_deps: list[Path], append_rpaths: List
...
+            if found_dependency := find_dependency(candidate.name, file_arch, file_osabi):
+                rpath.append(found_dependency)
+                dependencies.append(Dependency(path, candidate, found=True))
+                print(f"    {candidate} -> found: {found_dependency}")
+                break
+
+        if found_dependency is None:
+            dep_name = dep[0] if len(dep) == 1 else f"any({', '.join(map(str, dep))})"
+            dependencies.append(Dependency(path, dep_name, found=False))
+            print(f"    {dep_name} -> not found!")
```
layus, line 266: While I do understand the need to move to lists of strings (as the FreeDesktop spec uses lists), this rewrite seems incomplete and bogus to me.

1. The logic around libc_lib has been completely removed. Please re-add it.
2. At this stage, I think it is necessary to introduce two data types for dependencies. Or to split the logic looking for these dependencies. While libc_lib only applies to DT_NEEDED, absolute paths in .note.dlopen are probably not a `continue` but a break.
3. absolute paths should be considered as found. The continue statement doe snot work inside the nested loop over candidates.

Splitting this into subfunctions or methods of the class of each type of dependency may help.
I am not too much into style ultimately. I care more about the logic and it seems this code broke it.

layus (changes requested):
I think this is a real improvement for autoPatchelfHook. But in this form it is incomplete and broke the logic for the legacy use cases.

layus:
Would you consider splitting autoPatchelfHook changes in a separate PR ? Having both merged is going to make this one hard to merge and to test. I could easily iterate on a separate PR and contribute to the autoPatchelf changes.

pkgs/build-support/setup-hooks/auto-patchelf.py
ElvishJerricco: @layus So, the reason for the removal of the glibc thing was because `glibc` was being removed from the rpath of `systemd` binaries altogether, even when they had `DT_NEEDED` dependencies on it. This resulted in e.g. `/nix/store/yggl0850psl4anbavqqfg5hgpdsc2196-extra-utils/bin/udevadm: error while loading shared libraries: libacl.so.1: cannot open shared object file: No such file or directory`

It seemed odd to me, and I wasn't sure why `glibc` was being excluded in the first place. Running builds for my systems on hydra, however, I am now seeing at least a symptom of why:

```
auto-patchelf: 9 dependencies could not be satisfied
error: auto-patchelf could not satisfy dependency libc.so.6 wanted by /nix/store/rb3vfzranc26i7fk24ikm6v6lzg7v9cz-go-1.21.0-linux-amd64-bootstrap/share/go/src/debug/elf/testdata/gcc-amd64-linux-exec
error: auto-patchelf could not satisfy dependency libc.so.6 wanted by /nix/store/rb3vfzranc26i7fk24ikm6v6lzg7v9cz-go-1.21.0-linux-amd64-bootstrap/share/go/src/debug/dwarf/testdata/line-gcc-zstd.elf
error: auto-patchelf could not satisfy dependency libc.so.6 wanted by /nix/store/rb3vfzranc26i7fk24ikm6v6lzg7v9cz-go-1.21.0-linux-amd64-bootstrap/share/go/src/debug/dwarf/testdata/line-gcc-dwarf5.elf
error: auto-patchelf could not satisfy dependency libc.so.6 wanted by /nix/store/rb3vfzranc26i7fk24ikm6v6lzg7v9cz-go-1.21.0-linux-amd64-bootstrap/share/go/src/debug/dwarf/testdata/ranges.elf
error: auto-patchelf could not satisfy dependency libc.so.6 wanted by /nix/store/rb3vfzranc26i7fk24ikm6v6lzg7v9cz-go-1.21.0-linux-amd64-bootstrap/share/go/src/debug/dwarf/testdata/typedef.elf
error: auto-patchelf could not satisfy dependency libc.so.6 wanted by /nix/store/rb3vfzranc26i7fk24ikm6v6lzg7v9cz-go-1.21.0-linux-amd64-bootstrap/share/go/src/debug/dwarf/testdata/line-clang-dwarf5.elf
error: auto-patchelf could not satisfy dependency libc.so.6 wanted by /nix/store/rb3vfzranc26i7fk24ikm6v6lzg7v9cz-go-1.21.0-linux-amd64-bootstrap/share/go/src/debug/dwarf/testdata/line-gcc.elf
error: auto-patchelf could not satisfy dependency libc.so.6 wanted by /nix/store/rb3vfzranc26i7fk24ikm6v6lzg7v9cz-go-1.21.0-linux-amd64-bootstrap/share/go/src/debug/dwarf/testdata/line-clang.elf
error: auto-patchelf could not satisfy dependency libc.so.6 wanted by /nix/store/rb3vfzranc26i7fk24ikm6v6lzg7v9cz-go-1.21.0-linux-amd64-bootstrap/share/go/src/debug/dwarf/testdata/split.elf
auto-patchelf failed to find all the required dependencies.
Add the missing dependencies to --libs or use `--ignore-missing="foo.so.1 bar.so etc.so"`.
```

The odd bit is that this seems only relevant to the debug files; which is extra odd because systemd uses `separateDebugInfo` and the debug files for it don't have this problem.

Can you explain the logic for this `glibc` exclusion? It's very confusing to me.  And the consequence of failing to find `libacl.so.1` (and probably others) for systemd binaries is rather critical.

ElvishJerricco:
@layus 

> Would you consider splitting autoPatchelfHook changes in a separate PR ? Having both merged is going to make this one hard to merge and to test. I could easily iterate on a separate PR and contribute to the autoPatchelf changes.

This is a little tricky. Those changes are a prerequisite for systemd 256. I would be happy to make it a new PR, but it *needs* to receive fairly high priority attention, as we're already behind on updating to systemd 256.

ElvishJerricco:
@layus Should I make a PR for `autoPatchelfHook` targeting staging I guess?

Saiv46:
As initrd is broken, is there a way to build and use systemd-cryptenroll from this PR directly?

ElvishJerricco:
@Saiv46 please don't use this PR as a place to troubleshoot. If you're saying initrd isn't working on `master`, please file an issue (and tag me).

peigongdsd:
Hey guys, just here to remind that the IPForward= switch in systemd.network is renamed and splitted into IPv4Forwarding= and IPv6Forwarding=. See release note for systemd 256.1. This may cause trouble in the template of systemd.network.\<name\> module. See also [the documentation](https://www.freedesktop.org/software/systemd/man/latest/systemd.network.html#IPv4Forwarding=).

ElvishJerricco:
Ok, separate PR for the `autoPatchelfHook` bits has been made. I've got that branch rebased into this PR, but we're definitely approaching the point that we need to retarget this PR at staging.

ElvishJerricco:
Ok, here's the networking fixes @peigongdsd

ElvishJerricco:
Looks like there's quite a number of new units we should probably look into as well:

```
system/capsule@.service
system/capsule.slice
system/ssh-access.target
system/systemd-bootctl@.service
system/systemd-bootctl.socket
system/systemd-creds@.service
system/systemd-creds.socket
system/systemd-hibernate-clear.service
system/systemd-homed-firstboot.service
system/systemd-hostnamed.socket
system/systemd-journald-sync@.service
system/systemd-mountfsd.service
system/systemd-mountfsd.socket
system/systemd-networkd-persistent-storage.service
system/systemd-nsresourced.service
system/systemd-nsresourced.socket
system/systemd-pcrlock@.service
system/systemd-pcrlock.socket
system/systemd-udev-load-credentials.service
system/systemd-vmspawn@.service
system/tpm2.target
user/capsule@.target
```

JohnRTitor:
> but we're definitely approaching the point that we need to retarget this PR at staging.

yeah, let's do that.

JohnRTitor:
Looks like the base branch `staging` is still broken.

JohnRTitor:
@ofborg eval

nixos/modules/system/boot/networkd.nix
```diff
@@ -2835,6 +2843,7 @@ let
         "systemd-networkd-wait-online.service"
         "systemd-networkd.service"
         "systemd-networkd.socket"
+        "systemd-networkd-persistent-storage.service"
```
philiptaron, line 2846: [networkctl](https://www.freedesktop.org/software/systemd/man/latest/networkctl.html#%0A%20%20%20%20%20%20%20%20%20%20persistent-storage%0A%20%20%20%20%20%20%20%20%20%20BOOL%0A%20%20%20%20%20%20%20%20) has the slim documents. Source is https://raw.githubusercontent.com/systemd/systemd/main/units/systemd-networkd-persistent-storage.service.

nixos/modules/system/boot/systemd.nix
```diff
@@ -37,6 +37,8 @@ let
       "cryptsetup.target"
       "cryptsetup-pre.target"
       "remote-cryptsetup.target"
+    ] ++ optionals cfg.package.withTpm2Tss [
+      "tpm2.target"
```
philiptaron, line 41: [tpm2.target](https://www.freedesktop.org/software/systemd/man/latest/systemd.special.html#tpm2.target)

> This target is started automatically if a TPM2 device is discovered, either by the OS or by the firmware. It acts as synchronization point for services that require TPM2 device access. The target unit is enqueued by systemd-tpm2-generator(8) if it detects that the firmware has discovered a TPM2 device but the OS kernel has not activated a driver for it yet. It is also pulled in whenever systemd-udevd.service(8) discovers a TPM2 device. The target unit is ordered after the /dev/tpmrm0 device node, so that it only becomes active once the TPM2 device is actually accessible. Early boot programs that intend to access the TPM2 device should hence order themselves after this target unit, but not pull it in.
> 
> Added in version 256.

nixos/modules/system/boot/systemd.nix
```diff
@@ -112,6 +114,7 @@ let
       "sleep.target"
       "hybrid-sleep.target"
       "systemd-hibernate.service"
+      "systemd-hibernate-clear.service"
```
philiptaron, line 117: From the release notes:

> A tiny new service `systemd-hibernate-clear.service` has been added which clears hibernation information from the HibernateLocation EFI variable, in case the resume device is gone. Normally, this variable is supposed to be cleaned up by the code that initiates the resume from hibernation image. But when the device is missing and that code doesn't run, this service will now do the necessary work, ensuring that no outdated hibernation image information remains on subsequent boots.

See also https://www.freedesktop.org/software/systemd/man/latest/systemd-hibernate-resume.service.html

nixos/modules/system/boot/systemd.nix
```diff
@@ -136,6 +139,16 @@ let
       "systemd-ask-password-wall.path"
       "systemd-ask-password-wall.service"
 
+      # Varlink APIs
+      "systemd-bootctl@.service"
```
philiptaron, line 143: No docs; https://raw.githubusercontent.com/systemd/systemd/main/units/systemd-bootctl%40.service is the service itself. Release notes mention:

> bootctl now provides a basic Varlink interface and can be run as a daemon via a template unit.

nixos/modules/system/boot/systemd.nix
```diff
@@ -136,6 +139,16 @@ let
...
+      "systemd-bootctl.socket"
```
philiptaron, line 144: No specific docs: https://www.freedesktop.org/software/systemd/man/latest/bootctl.html is the main docs.

Source is https://raw.githubusercontent.com/systemd/systemd/main/units/systemd-bootctl.socket

nixos/modules/system/boot/systemd.nix
```diff
@@ -136,6 +139,16 @@ let
...
+      "systemd-creds@.service"
```
philiptaron, line 145: No specific docs: https://www.freedesktop.org/software/systemd/man/latest/systemd-creds.html is the main docs.

Source is https://raw.githubusercontent.com/systemd/systemd/main/units/systemd-creds%40.service

nixos/modules/system/boot/systemd.nix
```diff
@@ -136,6 +139,16 @@ let
...
+      "systemd-creds.socket"
```
philiptaron, line 146: Source is https://raw.githubusercontent.com/systemd/systemd/main/units/systemd-creds.socket.

nixos/modules/system/boot/systemd.nix
```diff
@@ -136,6 +139,16 @@ let
...
+    ] ++ lib.optional cfg.package.withTpm2Tss [
+      "systemd-pcrlock@.service"
+      "systemd-pcrlock.socket"
```
philiptaron, line 149: Main docs are https://www.freedesktop.org/software/systemd/man/latest/systemd-pcrlock.html

Source is [systemd-pcrlock@.service](https://raw.githubusercontent.com/systemd/systemd/main/units/systemd-pcrlock%40.service.in) and [systemd-pcrlock.socket](https://raw.githubusercontent.com/systemd/systemd/main/units/systemd-pcrlock.socket).

nixos/modules/system/boot/systemd.nix
```diff
@@ -158,6 +171,7 @@ let
     ] ++ optionals cfg.package.withHostnamed [
       "dbus-org.freedesktop.hostname1.service"
       "systemd-hostnamed.service"
+      "systemd-hostnamed.socket"
```
philiptaron, line 174: From the release notes:

> systemd-hostnamed now exposes the machine ID and boot ID via D-Bus. It also exposes the hosts AF_VSOCK CID, if available.
>
> systemd-hostnamed now provides a basic Varlink interface.
>
> systemd-hostnamed exports the full data in os-release(5) and machine-info(5) via D-Bus and Varlink.
> 
> hostnamectl now shows the system's product UUID and hardware serial number if known.

Source is https://raw.githubusercontent.com/systemd/systemd/main/units/systemd-hostnamed.socket

nixos/modules/system/boot/systemd/journald.nix
```diff
@@ -96,6 +96,7 @@ in {
       "systemd-journald@.service"
       "systemd-journal-flush.service"
       "systemd-journal-catalog-update.service"
+      "systemd-journald-sync@.service"
```
philiptaron, line 99: No specific docs; main docs are https://www.freedesktop.org/software/systemd/man/latest/systemd-journald.service.html

Source is https://github.com/systemd/systemd/blob/main/units/systemd-journald-sync%40.service

nixos/tests/systemd-networkd-ipv6-prefix-delegation.nix
```diff
@@ -40,7 +40,8 @@ import ./make-test-python.nix ({ pkgs, lib, ... }: {
             address = [
               "2001:DB8::1/64"
             ];
-            networkConfig.IPForward = true;
+            networkConfig.IPv4Forwarding = true;
```
philiptaron, line 43: I wonder if this setting (`IPv4Forwarding`) is actually needed for this IPv6 test.

nixos/tests/systemd-networkd-vrf.nix
```diff
@@ -16,7 +16,8 @@ import ./make-test-python.nix ({ pkgs, lib, ... }: let
         linkConfig.RequiredForOnline = "no";
         networkConfig = {
           Address = "192.168.${toString vlan}.${toString id}/24";
-          IPForward = "yes";
+          IPv4Forwarding = "yes";
+          IPv6Forwarding = "yes";
```
philiptaron, line 20: Unrelated to this PR: `"yes"` and `"no"` in this file would be better as `true` and `false`, as that's more Nix-ish. But it doesn't matter.

pkgs/build-support/setup-hooks/auto-patchelf.py
```diff
@@ -38,21 +40,42 @@ def is_dynamic_executable(elf: ELFFile) -> bool:
...
     return dependencies
 
 
-def get_rpath(elf: ELFFile) -> List[str]:
+def get_dlopen_dependencies(elf: ELFFile) -> list[list[Path]]:
+    """
+    Extracts dependencies from the .note.dlopen section, which is used by
+    systemd to store the list of libraries that are opened with dlopen.
+    https://systemd.io/ELF_DLOPEN_METADATA/
+    """
```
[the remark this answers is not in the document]
philiptaron: Agree with this edit from @layus.

pkgs/os-specific/linux/systemd/default.nix
```diff
@@ -226,6 +230,19 @@ stdenv.mkDerivation (finalAttrs: {
     ./0015-tpm2_context_init-fix-driver-name-checking.patch
     ./0016-systemctl-edit-suggest-systemdctl-edit-runtime-on-sy.patch
     ./0017-meson.build-do-not-create-systemdstatedir.patch
+
+    # https://github.com/systemd/systemd/pull/33258
```
philiptaron, line 234: - https://github.com/systemd/systemd/pull/33258

pkgs/os-specific/linux/systemd/default.nix
```diff
@@ -226,6 +230,19 @@ stdenv.mkDerivation (finalAttrs: {
...
+    # Remove after 256.3
+    (fetchpatch {
+      url = "https://github.com/systemd/systemd/compare/b268a71069786a45460807967e669d505ba3c5a2..f26b2ec46118a4493608618da2253bb9dfc6b517.patch";
+      hash = "sha256-OmuPDm3NykrDeNTA3NcYt9iTXEUFwKJ5apPP4KqtABg=";
+    })
+
+    # https://github.com/systemd/systemd/pull/33400
```
philiptaron, line 241: - https://github.com/systemd/systemd/pull/33400

pkgs/os-specific/linux/systemd/default.nix
```diff
@@ -226,6 +230,19 @@ stdenv.mkDerivation (finalAttrs: {
     ./0015-tpm2_context_init-fix-driver-name-checking.patch
     ./0016-systemctl-edit-suggest-systemdctl-edit-runtime-on-sy.patch
     ./0017-meson.build-do-not-create-systemdstatedir.patch
+
+    # https://github.com/systemd/systemd/pull/33258
+    # Remove after 256.3
```
philiptaron, line 235: I don't see a backport PR.

```suggestion
    # Remove after these changes are included in a systemd release
```

pkgs/os-specific/linux/systemd/default.nix
```diff
@@ -873,7 +773,8 @@ stdenv.mkDerivation (finalAttrs: {
     interfaceVersion = 2;
 
     inherit withBootloader withCryptsetup withEfi withHostnamed withImportd withKmod
-      withLocaled withMachined withPortabled withTimedated withUtmp util-linux kmod kbd;
+      withLocaled withMachined withPortabled withTimedated withTpm2Tss withUtmp
+      util-linux kmod kbd;
```
philiptaron, line 777: The styling from `nixfmt-rfc-style` is going to make this so much more readable.

philiptaron (approved):
No actionable comments, just a lot of metadata and notes from searching for documentation, sources, and release notes. Please feel free to resolve. I didn't re-review the Python auto-patchelf changes in detail.

I'll get my machine building this over night. I haven't actually run the most recent 256.2 bits yet.

pkgs/os-specific/linux/systemd/default.nix
ElvishJerricco: IIUC, they don't do backports the way we do with nixpkgs. The PR does have the `needs-stable-backport` label, which means it *should* just end up pushed to the 256 branch before the next release (although it didn't before 256.2, but those were only days apart)

philiptaron:
I get a build failure (`/nix/store/v2c43js97ks0z4afpg44m2li8d39h082-lazy-options.json.drv`) that I think is related to this PR. 

```
error:
       … while evaluating attribute 'boot.initrd.systemd.enableTpm2'

         at /nix/store/r2b760s2ks0xcjiv0rnwbcpr6fcn5i08-nixos/lib/make-options-doc/default.nix:167:63:

          166|
          167|   optionsNix = builtins.listToAttrs (map (o: { name = o.name; value = removeAttrs o ["name" "visible" "internal"]; }) optionsList);
             |                                                               ^
          168|

       … while evaluating attribute 'default'

         at /nix/store/ish82x3mrhxiajq4h10rd1q4kfbqfj1l-lib/options.nix:344:11:

          343|         // optionalAttrs (opt ? defaultText || opt ? default) {
          344|           default =
             |           ^
          345|             builtins.addErrorContext "while evaluating the ${if opt?defaultText then "defaultText" else "default value"} of option `${name}`" (

       … while calling the 'addErrorContext' builtin

         at /nix/store/ish82x3mrhxiajq4h10rd1q4kfbqfj1l-lib/options.nix:345:13:

          344|           default =
          345|             builtins.addErrorContext "while evaluating the ${if opt?defaultText then "defaultText" else "default value"} of option `${name}`" (
             |             ^
          346|               renderOptionValue (opt.defaultText or opt.default)

       … while evaluating the default value of option `boot.initrd.systemd.enableTpm2`

       … from call site

         at /nix/store/ish82x3mrhxiajq4h10rd1q4kfbqfj1l-lib/options.nix:346:15:

          345|             builtins.addErrorContext "while evaluating the ${if opt?defaultText then "defaultText" else "default value"} of option `${name}`" (
          346|               renderOptionValue (opt.defaultText or opt.default)
             |               ^
          347|             );

       … while calling 'renderOptionValue'

         at /nix/store/ish82x3mrhxiajq4h10rd1q4kfbqfj1l-lib/options.nix:383:23:

          382|   */
          383|   renderOptionValue = v:
             |                       ^
          384|     if v ? _type && v ? text then v

       error: attribute 'boot' missing

       at /nix/store/r2b760s2ks0xcjiv0rnwbcpr6fcn5i08-nixos/modules/system/boot/systemd/initrd.nix:19:9:

           18|
           19|   cfg = config.boot.initrd.systemd;
             |         ^
           20|
Cacheable portion of option doc build failed.
Usually this means that an option attribute that ends up in documentation (eg `default` or `description`) depends on the restricted module arguments `config` or `pkgs`.
```

nixos/modules/system/boot/systemd/initrd.nix
```diff
@@ -344,7 +343,7 @@ in {
     };
 
     enableTpm2 = mkOption {
-      default = true;
+      default = cfg.package.withTpm2Tss;
```
philiptaron, line 346: This causes a build failure in lazy-options.json, sadly.

philiptaron:
I was able to complete the build with a fix like https://github.com/NixOS/nixpkgs/commit/cee4ca8fa112f7c9bbee6f30be7fb8ec1e840893.

nixos/modules/system/boot/systemd/initrd.nix
CertainLach: Shouldn't it just be `mkPackageOption`? (I.e using defaultText for docs)

ElvishJerricco: Yea, it's fixed with `defaultText`.

philiptaron (approved):
This all builds and runs.

pkgs/build-support/setup-hooks/auto-patchelf.py
```diff
@@ -38,21 +40,42 @@ def is_dynamic_executable(elf: ELFFile) -> bool:
...
 
     return dependencies
 
 
-def get_rpath(elf: ELFFile) -> List[str]:
+def get_dlopen_dependencies(elf: ELFFile) -> list[list[Path]]:
+    """
+    Extracts dependencies from the `.note.dlopen` section.
+    This is a FreeDesktop standard to annotate binaries with libraries that it may `dlopen`.
+    See https://systemd.io/ELF_DLOPEN_METADATA/
```
benaryorg, line 61: I've seen that phrasing twice now, and so far all I could find was systemd documenting this and referring to the owner of the section header(?) as "FreeDesktop.org". Does anyone have any information on whether this is actually a FreeDesktop standard, because if I'm not mistaken systemd is neither under the freedesktop umbrella (they have moved elsewhere AFAIK), nor is any of this [listed in the FDO specs](https://www.freedesktop.org/wiki/Specifications/) (unless I just fail to see it right now).

This isn't a blocker or anything, I'm just curious. And if it happens to not be an actual FreeDesktop standard, could we please not call it that, and otherwise referring to the actual standard in question (rather than to a systemd related website) would be cool I guess?

ElvishJerricco: Hm, interesting. systemd calls it a FDO owned thing. In their documentation on their site, the documentation for the section header has this:

```
Owner: `FDO` (FreeDesktop.org)
```

arianvp (approved):

tulilirockz:
oh my god! great job everyone that contributed to this!

Mic92:
This broke the netboot image. Fix in: https://github.com/NixOS/nixpkgs/pull/331712
This is relevant for nixos-anywhere.

squalus:
Commit 80be92660 broke the autoPatchelf step in `osquery.toolchain`. Reversing the order of steps 2 and 3 in the new `for candidate in dep` loop (thus restoring the previous behavior) fixes the problem.

SuperSandro2000:
I think since this update plymouth is no longer working for me but I wasn't be able to get a useful log yet other than it existed.

arianvp:
Can you create a new issue for this?  We can start collecting logs there. I hope it's not too cursed :(

SuperSandro2000:
> Can you create a new issue for this? We can start collecting logs there. I hope it's not too cursed :(

#332812

JohnRTitor:
I am seeing a issue where systemd-oomd would activate in initrd and randomly kill processes like `systemd-journald` and `plymouth`.
The system does boot successfully most of the time, sometimes it fails. I am on 16Gigs of RAM, and this has started occurring after this reached unstable.

```log
$  sudo dmesg | grep oom
[  102.330084] kswapd0 invoked oom-killer: gfp_mask=0xcc0(GFP_KERNEL), order=0, oom_score_adj=0
[  102.330104]  oom_kill_process+0x203/0x2f0
[  102.330297] [  pid  ]   uid  tgid total_vm      rss rss_anon rss_file rss_shmem pgtables_bytes swapents oom_score_adj name
[  102.330309] [    963]   993   963     3645     1395        0     1395         0    73728      192          -900 systemd-oomd
[  102.330327] oom-kill:constraint=CONSTRAINT_NONE,nodemask=(null),cpuset=/,mems_allowed=0,global_oom,task_memcg=/system.slice/plymouth-start.service,task=plymouthd,pid=265,uid=0
[  102.330337] Out of memory: Killed process 265 (plymouthd) total-vm:115232kB, anon-rss:520kB, file-rss:11700kB, shmem-rss:0kB, UID:0 pgtables:128kB oom_score_adj:0
[  102.333497] kswapd0 invoked oom-killer: gfp_mask=0xcc0(GFP_KERNEL), order=0, oom_score_adj=0
[  102.333512]  oom_kill_process+0x203/0x2f0
[  102.333674] [  pid  ]   uid  tgid total_vm      rss rss_anon rss_file rss_shmem pgtables_bytes swapents oom_score_adj name
[  102.333685] [    963]   993   963     3645     1395        0     1395         0    73728      192          -900 systemd-oomd
[  102.333702] oom-kill:constraint=CONSTRAINT_NONE,nodemask=(null),cpuset=/,mems_allowed=0,global_oom,task_memcg=/system.slice/NetworkManager.service,task=NetworkManager,pid=980,uid=0
[  102.333723] Out of memory: Killed process 980 (NetworkManager) total-vm:329680kB, anon-rss:0kB, file-rss:11084kB, shmem-rss:0kB, UID:0 pgtables:152kB oom_score_adj:0
[  102.335620] kswapd0 invoked oom-killer: gfp_mask=0xcc0(GFP_KERNEL), order=0, oom_score_adj=0
[  102.335631]  oom_kill_process+0x203/0x2f0
[  102.335784] [  pid  ]   uid  tgid total_vm      rss rss_anon rss_file rss_shmem pgtables_bytes swapents oom_score_adj name
[  102.335794] [    963]   993   963     3645     1395        0     1395         0    73728      192          -900 systemd-oomd
[  102.335810] oom-kill:constraint=CONSTRAINT_NONE,nodemask=(null),cpuset=/,mems_allowed=0,global_oom,task_memcg=/system.slice/systemd-resolved.service,task=systemd-resolve,pid=964,uid=153
[  102.335818] Out of memory: Killed process 964 (systemd-resolve) total-vm:21596kB, anon-rss:3328kB, file-rss:9240kB, shmem-rss:0kB, UID:153 pgtables:84kB oom_score_adj:0
[  102.355076] kswapd0 invoked oom-killer: gfp_mask=0xcc0(GFP_KERNEL), order=0, oom_score_adj=0
[  102.355093]  oom_kill_process+0x203/0x2f0
[  102.355285] [  pid  ]   uid  tgid total_vm      rss rss_anon rss_file rss_shmem pgtables_bytes swapents oom_score_adj name
[  102.355296] [    963]   993   963     3645     1235        0     1235         0    73728      192          -900 systemd-oomd
[  102.355310] oom-kill:constraint=CONSTRAINT_NONE,nodemask=(null),cpuset=/,mems_allowed=0,global_oom,task_memcg=/system.slice/home-manager-masum.service,task=nix-build,pid=1005,uid=1001
[  102.355319] Out of memory: Killed process 1005 (nix-build) total-vm:449524kB, anon-rss:0kB, file-rss:4740kB, shmem-rss:0kB, UID:1001 pgtables:136kB oom_score_adj:0
[  104.467223] oom_reaper: reaped process 1005 (nix-build), now anon-rss:0kB, file-rss:304kB, shmem-rss:0kB
```

arianvp:
Those logs are from the kernel OOM killer not systemd-oomd

arianvp:
Which is even more weird with 16GB RAM...

PR #217536, opened 2023-02-21
nixos/caddy: add support for reload

sephii:
###### Description of changes

Add support for a `services.caddy.enableReload` option to the Caddy module. This is similar to what is available in the nginx module and allows reloading the configuration without stopping the service.

Maybe `enableReload` should be enabled by default? I’m not sure why it’s not in the nginx module, maybe because of the additional SSL setup?

###### Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
- [ ] For non-Linux: Is `sandbox = true` set in `nix.conf`? (See [Nix manual](https://nixos.org/manual/nix/stable/command-ref/conf-file.html))
- [ ] Tested, as applicable:
  - [NixOS test(s)](https://nixos.org/manual/nixos/unstable/index.html#sec-nixos-tests) (look inside [nixos/tests](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests))
  - and/or [package tests](https://nixos.org/manual/nixpkgs/unstable/#sec-package-tests)
  - or, for functions and "core" functionality, tests in [lib/tests](https://github.com/NixOS/nixpkgs/blob/master/lib/tests) or [pkgs/test](https://github.com/NixOS/nixpkgs/blob/master/pkgs/test)
  - made sure NixOS tests are [linked](https://nixos.org/manual/nixpkgs/unstable/#ssec-nixos-tests-linking) to the relevant packages
- [ ] Tested compilation of all packages that depend on this change using `nix-shell -p nixpkgs-review --run "nixpkgs-review rev HEAD"`. Note: all changes have to be committed, also see [nixpkgs-review usage](https://github.com/Mic92/nixpkgs-review#usage)
- [x] Tested basic functionality of all binary files (usually in `./result/bin/`)
- [23.05 Release Notes (or backporting 22.11 Release notes)](https://github.com/NixOS/nixpkgs/blob/master/CONTRIBUTING.md#generating-2305-release-notes)
  - [ ] (Package updates) Added a release notes entry if the change is major or breaking
  - [ ] (Module updates) Added a release notes entry if the change is significant
  - [ ] (Module addition) Added a release notes entry if adding a new NixOS module
- [x] Fits [CONTRIBUTING.md](https://github.com/NixOS/nixpkgs/blob/master/CONTRIBUTING.md).

PR #307076, opened 2024-04-26
nixos/caddy: don't set ExecReload if enableReload is disabled

flokli:
Otherwise, setting services.caddy.enableReload to false fails in a very bad fashion:

The reload command still gets executed, but fails:

```
Apr 26 21:23:01 n1-rk1 systemd[1]: Reloading Caddy...
Apr 26 21:23:01 n1-rk1 caddy[70793]: {"level":"info","ts":1714166581.733018,"msg":"using provided configuration","config_file":"/etc/caddy/caddy_config","config_adapter":"caddyfile"}
Apr 26 21:23:01 n1-rk1 caddy[70793]: {"level":"warn","ts":1714166581.7353032,"msg":"Caddyfile input is not formatted; run 'caddy fmt --overwrite' to fix inconsistencies","adapter":"caddyfile","file":"/etc/caddy/caddy_config","line":3}
Apr 26 21:23:01 n1-rk1 caddy[70793]: Error: sending configuration to instance: performing request: Post "http://localhost:2019/load": dial tcp [::1]:2019: connect: connection refused
Apr 26 21:23:01 n1-rk1 systemd[1]: caddy.service: Control process exited, code=exited, status=1/FAILURE
Apr 26 21:23:01 n1-rk1 systemd[1]: Reload failed for Caddy.
```

… and the server is not restarted either, as a ExecReload= command is specified.

Fix this, by only setting ExecReload if the reload exists.

The first empty string is still necessary to reset the old option.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/nixos/modules/services/web-servers/caddy/default.nix
+++ b/nixos/modules/services/web-servers/caddy/default.nix
@@ -365,7 +365,7 @@ in
         # If the empty string is assigned to this option, the list of commands to start is reset, prior assignments of this option will have no effect.
         ExecStart = [ "" ''${cfg.package}/bin/caddy run ${runOptions} ${optionalString cfg.resume "--resume"}'' ];
         # Validating the configuration before applying it ensures we’ll get a proper error that will be reported when switching to the configuration
-        ExecReload = [ "" ''${cfg.package}/bin/caddy reload ${runOptions} --force'' ];
+        ExecReload = [ "" ] ++ lib.optional cfg.enableReload "${lib.getExe cfg.package} reload ${runOptions} --force";
         User = cfg.user;
         Group = cfg.group;
         ReadWritePaths = [ cfg.dataDir ];


nixos/modules/services/web-servers/caddy/default.nix
```diff
@@ -364,7 +364,7 @@ in
         # If the empty string is assigned to this option, the list of commands to start is reset, prior assignments of this option will have no effect.
         ExecStart = [ "" ''${cfg.package}/bin/caddy run ${runOptions} ${optionalString cfg.resume "--resume"}'' ];
         # Validating the configuration before applying it ensures we’ll get a proper error that will be reported when switching to the configuration
-        ExecReload = [ "" ''${cfg.package}/bin/caddy reload ${runOptions} --force'' ];
+        ExecReload = [ "" ] ++ lib.optional cfg.enableReload ''${cfg.package}/bin/caddy reload ${runOptions} --force'';
```
drupol, line 367:
```suggestion
        ExecReload = lib.optionals cfg.enableReload [ ''${lib.getExe cfg.package} reload ${runOptions} --force'' ];
```

drupol (dismissed):

nixos/modules/services/web-servers/caddy/default.nix
flokli: No, the empty list item is explicitly there to clear the list from any previous entries (see systemd man pages).

If we apply the change you proposed, enableDefault set to false would make the config in the upstream unit file shine through.

drupol, line 367:
```suggestion
        ExecReload = [ "" ] ++ lib.optionals cfg.enableReload [ ''${lib.getExe cfg.package} reload ${runOptions} --force'' ];
```

emilylange (approved):
Makes sense to me.

Thanks :)

Ref: #217536

drupol:
> Makes sense to me.
> 
> Thanks :)
> 
> Ref: #217536

Please don't merge immediatelly, there's still one comment of mine that needs to be addressed.

emilylange:
> Please don't merge immediatelly, there's still one comment of mine that needs to be addressed.

Your review is blocking, no worries :)

nixos/modules/services/web-servers/caddy/default.nix
```diff
@@ -364,7 +364,7 @@ in
         # If the empty string is assigned to this option, the list of commands to start is reset, prior assignments of this option will have no effect.
         ExecStart = [ "" ''${cfg.package}/bin/caddy run ${runOptions} ${optionalString cfg.resume "--resume"}'' ];
         # Validating the configuration before applying it ensures we’ll get a proper error that will be reported when switching to the configuration
```
jian-lin, line 366: It is a mistake to add this comment in https://github.com/NixOS/nixpkgs/pull/217536. See https://github.com/NixOS/nixpkgs/pull/217536#discussion_r1175279412.

While it is not relevant to this PR, it would be good to remove this comment.

emilylange: Yes, I considered pointing this out here as well (as part of my review).

But I eventually decided against doing so, because of an extremely fuzzy and far fetched interpretation of that comment:

> Validating the configuration before applying it ensures we’ll get a proper error that will be reported when switching to the configuration

being something like

> [Caddy] validat[es] the configuration before applying [, which] ensures we'll get a proper error that will be reported when switching to the configuration.

Caddy will never accept a broken config.

But yeah, that's really far fetched.

If you feel like updating the comment, please do :)

flokli: Let's deal with this in a followup.

nixos/modules/services/web-servers/caddy/default.nix
flokli: Still don't think we should use `lib.optionals` if it's just a single string (but `lib.optional`), but happy to do the `lib.getExe` suggestion.

drupol: Technically both solutions are valid and this is mostly an opinionated change.

I see that many people are making mistakes in `nixpkgs`, mixing the use of `lib.optional` and `lib.optionals`. For all these reasons, I think using `lib.optionals` when dealing with an attributes which is a list makes more sense.

Also, using `lib.optionals` reduce the cognitive load for newcomers since the brackets needs to be explicitly added.

In any case, this won't prevent me from sleeping if you don't do the change, it's just based on my thin experience in the Nix world.

flokli: Hmmh, I misinterpreted as a "request changes" type of review, because the old red review didn't get cleared.

I personally prefer using `lib.optional`. These are kinds of personal preferences, I usually leave it up to the authors of PRs.

flokli: Updated.

flokli:
@drupol is there anything left, or can this be merged?

flokli:
Thanks!

PR #307123, opened 2024-04-27
shiori: 1.5.5 -> 1.7.0

CaptainJawZ:
## Description of changes

Added myself as a maintainer and updated to 1.6.2 (coincidentally 1.6.3 came out some hours ago, but it requires a go version not yet on the nixpkgs)

A lot of changes have happened between 1.5.5 and this newer release. Mainly the ability to setup environment variables to connect to postgres / mysql.

as version 1.6.0 is the one with the most changes, I link to that changelog as well:
https://github.com/go-shiori/shiori/releases/tag/v1.6.0
https://github.com/go-shiori/shiori/releases/tag/v1.6.2

This change required to modify the service module as well key changes include:
- changed the exec line so it uses the new "server" mode, "serve" is now deprecated
- created an option to include httpSecretKey, which setups the environment variable that keeps the session from invalidating with every restart. 
- created an option for databaseUrl, which allows the user to connect shiori with an existing postgres/mysql database (tested with postgres, but I don't see why it wouldn't work with mysql). This change required one of the hardening functions to be disabled, otherwise unix sockets were disabled, the socket is only exposed depending on if this setting was set, and on the contents of the connection string.

Noting that I thought it would be better to let the user decide how they form their connection string, instead of forcing that last option to be an enum and connect using the recommended string, this way the user can connect to remote databases, docker databases, using passwords, etc.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [x] Tested, as applicable:

---

### DIFF
--- a/nixos/modules/services/web-apps/shiori.nix
+++ b/nixos/modules/services/web-apps/shiori.nix
@@ -80,10 +80,14 @@ in {
 
           # For SSL certificates, and the resolv.conf
           "/etc"
-        ] ++ lib.optional (lib.strings.hasInfix "postgres" cfg.databaseUrl
-          && config.services.postgresql.enable) "/run/postgresql"
-          ++ lib.optional (lib.strings.hasInfix "mysql" cfg.databaseUrl
-            && config.services.mysql.enable) "/var/run/mysqld";
+        ] ++ lib.optional (config.services.postgresql.enable &&
+                           cfg.databaseUrl != null &&
+                           lib.strings.hasPrefix "postgres://" cfg.databaseUrl)
+            "/run/postgresql"
+          ++ lib.optional (config.services.mysql.enable &&
+                           cfg.databaseUrl != null &&
+                           lib.strings.hasPrefix "mysql://" cfg.databaseUrl)
+            "/var/run/mysqld";
 
         CapabilityBoundingSet = "";
         AmbientCapabilities = "CAP_NET_BIND_SERVICE";
--- a/nixos/modules/services/web-apps/shiori.nix
+++ b/nixos/modules/services/web-apps/shiori.nix
@@ -1,17 +1,15 @@
 { config, lib, pkgs, ... }:
 
-with lib;
-let
-  cfg = config.services.shiori;
+let cfg = config.services.shiori;
 in {
   options = {
     services.shiori = {
-      enable = mkEnableOption "Shiori simple bookmarks manager";
+      enable = lib.mkEnableOption "Shiori simple bookmarks manager";
 
-      package = mkPackageOption pkgs "shiori" { };
+      package = lib.mkPackageOption pkgs "shiori" { };
 
-      address = mkOption {
-        type = types.str;
+      address = lib.mkOption {
+        type = lib.types.str;
         default = "";
         description = ''
           The IP address on which Shiori will listen.
@@ -19,30 +17,55 @@ in {
         '';
       };
 
-      port = mkOption {
-        type = types.port;
+      port = lib.mkOption {
+        type = lib.types.port;
         default = 8080;
         description = "The port of the Shiori web application";
       };
 
-      webRoot = mkOption {
-        type = types.str;
+      webRoot = lib.mkOption {
+        type = lib.types.str;
         default = "/";
         example = "/shiori";
         description = "The root of the Shiori web application";
       };
+
+      environmentFile = lib.mkOption {
+        type = lib.types.null or lib.types.path;
+        default = null;
+        example = "/path/to/environmentFile";
+        description = ''
+          Path to file containing environment variables.
+          Useful for passing down secrets.
+          <https://github.com/go-shiori/shiori/blob/master/docs/Configuration.md#overall-configuration>
+        '';
+      };
+
+      databaseUrl = lib.mkOption {
+        type = lib.types.null or lib.types.str;
+        default = null;
+        example = "postgresql:///shiori?host=/run/postgresql";
+        description = "The connection URL to connect to MySQL or PostgreSQL";
+      };
     };
   };
 
-  config = mkIf cfg.enable {
-    systemd.services.shiori = with cfg; {
+  config = lib.mkIf cfg.enable {
+    systemd.services.shiori = {
       description = "Shiori simple bookmarks manager";
       wantedBy = [ "multi-user.target" ];
-
-      environment.SHIORI_DIR = "/var/lib/shiori";
+      after = [ "postgresql.service" "mysql.service" ];
+      environment = {
+        SHIORI_DIR = "/var/lib/shiori";
+      } // lib.optionalAttrs (cfg.databaseUrl != null) {
+        SHIORI_DATABASE_URL = cfg.databaseUrl;
+      };
 
       serviceConfig = {
-        ExecStart = "${package}/bin/shiori serve --address '${address}' --port '${toString port}' --webroot '${webRoot}'";
+        ExecStart =
+          "${cfg.package}/bin/shiori server --address '${cfg.address}' --port '${
+            toString cfg.port
+          }' --webroot '${cfg.webRoot}'";
 
         DynamicUser = true;
         StateDirectory = "shiori";
@@ -50,15 +73,20 @@ in {
         RuntimeDirectory = "shiori";
 
         # Security options
-
+        EnvironmentFile =
+          lib.optional (cfg.environmentFile != null) cfg.environmentFile;
         BindReadOnlyPaths = [
           "/nix/store"
 
           # For SSL certificates, and the resolv.conf
           "/etc"
-        ];
+        ] ++ lib.optional (lib.strings.hasInfix "postgres" cfg.databaseUrl
+          && config.services.postgresql.enable) "/run/postgresql"
+          ++ lib.optional (lib.strings.hasInfix "mysql" cfg.databaseUrl
+            && config.services.mysql.enable) "/var/run/mysqld";
 
         CapabilityBoundingSet = "";
+        AmbientCapabilities = "CAP_NET_BIND_SERVICE";
 
         DeviceAllow = "";
 
@@ -78,7 +106,7 @@ in {
         ProtectKernelTunables = true;
 
         RestrictNamespaces = true;
-        RestrictAddressFamilies = [ "AF_INET" "AF_INET6" ];
+        RestrictAddressFamilies = [ "AF_INET" "AF_INET6" "AF_UNIX" ];
         RestrictRealtime = true;
         RestrictSUIDSGID = true;
 
@@ -88,11 +116,17 @@ in {
         SystemCallErrorNumber = "EPERM";
         SystemCallFilter = [
           "@system-service"
-          "~@cpu-emulation" "~@debug" "~@keyring" "~@memlock" "~@obsolete" "~@privileged" "~@setuid"
+          "~@cpu-emulation"
+          "~@debug"
+          "~@keyring"
+          "~@memlock"
+          "~@obsolete"
+          "~@privileged"
+          "~@setuid"
         ];
       };
     };
   };
 
-  meta.maintainers = with maintainers; [ minijackson ];
+  meta.maintainers = with lib.maintainers; [ minijackson CaptainJawZ ];
 }
--- a/nixos/modules/services/web-apps/shiori.nix
+++ b/nixos/modules/services/web-apps/shiori.nix
@@ -44,7 +44,7 @@ in {
       databaseUrl = lib.mkOption {
         type = lib.types.nullOr lib.types.str;
         default = null;
-        example = "postgresql:///shiori?host=/run/postgresql";
+        example = "postgres:///shiori?host=/run/postgresql";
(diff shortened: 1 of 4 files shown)


nixos/modules/services/web-apps/shiori.nix
```diff
@@ -78,7 +108,8 @@ in {
         ProtectKernelTunables = true;
 
         RestrictNamespaces = true;
-        RestrictAddressFamilies = [ "AF_INET" "AF_INET6" ];
+        # Prevents shiori from connecting from UNIX socket.
+        # RestrictAddressFamilies = [ "AF_INET" "AF_INET6" ];
```
Mic92, line 112:
```suggestion
        RestrictAddressFamilies = [ "AF_INET" "AF_INET6" "AF_UNIX" ];
```

nixos/modules/services/web-apps/shiori.nix
```diff
@@ -78,7 +108,8 @@ in {
         ProtectKernelTunables = true;
 
         RestrictNamespaces = true;
-        RestrictAddressFamilies = [ "AF_INET" "AF_INET6" ];
+        # Prevents shiori from connecting from UNIX socket.
```
Mic92, line 111:
```suggestion
```

nixos/modules/services/web-apps/shiori.nix
```diff
@@ -8,7 +8,12 @@ in {
     services.shiori = {
       enable = mkEnableOption "Shiori simple bookmarks manager";
 
-      package = mkPackageOption pkgs "shiori" { };
+      package = mkOption {
+        type = types.package;
+        default = pkgs.shiori;
+        defaultText = literalExpression "pkgs.shiori";
+        description = "The Shiori package to use.";
```
Mic92, line 15: Is this not the same as mkPackageOption?

CaptainJawZ: Oh, I based this chance on the most resent commit on unstable, I figured there was a reason for it, but I read the docu and indeed that's the better option!

nixos/modules/services/web-apps/shiori.nix
CaptainJawZ: There is always something new to learn! made a rebase with these changes :)

qbit:
There is also https://github.com/NixOS/nixpkgs/pull/296657 which mentions some breakage with the module (I haven't investigated it myself).

CaptainJawZ:
> There is also #296657 which mentions some breakage with the module (I haven't investigated it myself).

all the things mentioned there are fixed here
- serve changed to server
- SHIORI_HTTP_SECRET_KEY added as an option in the module
- And the database support, which is the main change.

nixos/modules/services/web-apps/shiori.nix
```diff
@@ -31,6 +31,20 @@ in {
         example = "/shiori";
         description = "The root of the Shiori web application";
       };
+
+      httpSecretKey = mkOption {
+        type = types.str;
+        default = "When empty all sessions will be invalidated on server restart";
```
qbit, line 37: This seems like it should be the `description` and not the `default`?

CaptainJawZ: OOOOOOOOPS, fixed

CaptainJawZ:
@qbit @Mic92 

I noticed there is a bug, caused by line 110 and maybe 50?:
RootDirectory = "/run/shiori";

(in my defense one of the OG lines not my addition teehee)

Upon rebooting the computer, shiori can't find anything from BindReadOnlyPaths.

But! if you run a sudo systemctl restart shiori, it reads them.

Meaning that everytime you reboot your computer, it will stop connecting to postgres, but it will do so if you reload the systemd unit. 

I was wondering if there is something you guys would advice me do about it, I thought of simply removing the lines but that would mess with the original intent of having this module hardened.

qbit:
I am not sure. I don't know much about systemd and what not :D

l0b0:
> @qbit @Mic92
> 
> I noticed there is a bug, caused by line 110 and maybe 50?: RootDirectory = "/run/shiori";
> 
> (in my defense one of the OG lines not my addition teehee)
> 
> Upon rebooting the computer, shiori can't find anything from BindReadOnlyPaths.

Sounds like implementing a NixOS test would be useful to make sure this doesn't happen.

CaptainJawZ:
> > @qbit @Mic92
> > I noticed there is a bug, caused by line 110 and maybe 50?: RootDirectory = "/run/shiori";
> > (in my defense one of the OG lines not my addition teehee)
> > Upon rebooting the computer, shiori can't find anything from BindReadOnlyPaths.
> 
> Sounds like implementing a [NixOS test](https://github.com/NixOS/nixpkgs/tree/master/nixos/tests) would be useful to make sure this doesn't happen.

can i upload the test on this same PR?

l0b0:
> can i upload the test on this same PR?

Absolutely, that's the preferred way.

aktaboot:
@CaptainJawZ new major version dropped :)

CaptainJawZ:
> @CaptainJawZ new major version dropped :)

Oh I'll update it this week! thanks for letting me know

nixos/modules/services/web-apps/shiori.nix
```diff
@@ -31,6 +31,20 @@ in {
         example = "/shiori";
         description = "The root of the Shiori web application";
       };
+
+      httpSecretKey = mkOption {
+        type = types.str;
+        example = "SuperSecretPassword";
+        description =
+          "When empty all sessions will be invalidated on server restart";
+      };
```
h7x4, line 40: Do you think you could make this take a file instead, and inject the contents into the environment variable on startup? It's not much more secure, but at least you won't have to leave the key in the nix store.

Also, it's probably better to represent *unset* as `null` instead of empty string.

nixos/modules/services/web-apps/shiori.nix
```diff
@@ -31,6 +31,20 @@ in {
...
+
+      databaseUrl = mkOption {
+       type = types.str;
+       default = "";
+       example = "postgresql:///shiori?host=/run/postgresql";
+       description = "The connection URL to connect to MySQL or PostgreSQL";
+      };
```
h7x4, line 47: Is this a required connection? In that case, I think you should drop the `default`, so that the user will be warned when it is left unset.

nixos/modules/services/web-apps/shiori.nix
```diff
@@ -56,9 +76,14 @@ in {
 
           # For SSL certificates, and the resolv.conf
           "/etc"
-        ];
+        ] ++
+        lib.optional (lib.strings.hasInfix "postgres" cfg.databaseUrl)
+          "/run/postgresql" ++
+        lib.optional (lib.strings.hasInfix "mysql" cfg.databaseUrl)
+          "/var/run/mysqld";
```
h7x4, line 83: I think there is a case to be made for instances that use external databases here, but I don't have any clever solutions at the moment. Maybe append `&& config.services.{postgresql,mysql}.enable`? Just to be sure - this does not fail if these paths does not exist, right?

pkgs/servers/web-apps/shiori/default.nix
```diff
@@ -2,17 +2,17 @@
...
 
   src = fetchFromGitHub {
     owner = "go-shiori";
     repo = pname;
     rev = "v${version}";
-    sha256 = "sha256-kGPvCYvLLixEH9qih/F3StUyGPqlKukTWLSw41+Mq8E=";
+    sha256 = "sha256-1ZZpQXlAHs5MTowCv3sWS3L7X5FTnU/b4trvHPiz+uE=";
   };
 
   passthru.tests = {
```
donovanglover, line 18:
```suggestion
  nativeBuildInputs = [ installShellFiles ];

  postInstall = ''
    installShellCompletion --cmd shiori \
      --bash <($out/bin/shiori completion bash) \
      --fish <($out/bin/shiori completion fish) \
      --zsh <($out/bin/shiori completion zsh)
  '';

  passthru.tests = {
```
Make sure to add `installShellFiles` to the top.

donovanglover (commented):
We should install the shell completions as well.

CaptainJawZ:
> > @qbit @Mic92
> > I noticed there is a bug, caused by line 110 and maybe 50?: RootDirectory = "/run/shiori";
> > (in my defense one of the OG lines not my addition teehee)
> > Upon rebooting the computer, shiori can't find anything from BindReadOnlyPaths.
> 
> Sounds like implementing a [NixOS test](https://github.com/NixOS/nixpkgs/tree/master/nixos/tests) would be useful to make sure this doesn't happen.

I fixed it by using the line, I tested it on two computers, one with postgres and another with mysql, both with multiple reboots, before I was complicated my life too much by trying to make this after optional, but it doesn't matter if the respective database service is not enabled, there is other safeguards implemented through latter suggestions on the thread.

`after = [ "postgresql.service" "mysql.service" ];`

CaptainJawZ:
> @CaptainJawZ new major version dropped :)

updated!

nixos/modules/services/web-apps/shiori.nix
CaptainJawZ: I changed httpSecretKey in favor for environmentFile, this way the user can not only set up this environment key variable themselves (which is optional) but also, whichever other keys they decide to, linked to the documentation on the description, and implemented your null as unset suggestion! for everything except listen address, that one has to be "" for the default to trigger.

nixos/modules/services/web-apps/shiori.nix
CaptainJawZ: It's not a required connection, if it's unset, shiori will create a sqlite database.

nixos/modules/services/web-apps/shiori.nix
CaptainJawZ: Good suggestion! implemented that check and worked like a charm! it's good to be extra secure.

pkgs/servers/web-apps/shiori/default.nix
CaptainJawZ: Thank you so much! I didn't know about this, and I implemented it as you suggested, and it works!

CaptainJawZ:
The module feels pretty much complete to me, all the feedback has been implemented, sorry guys it took me a bit to get to it.  If there is any revisions to be had I'll be more than happy to implement them!

pkgs/servers/web-apps/shiori/default.nix
```diff
@@ -1,29 +1,35 @@
...
     rev = "v${version}";
-    sha256 = "sha256-kGPvCYvLLixEH9qih/F3StUyGPqlKukTWLSw41+Mq8E=";
+    sha256 = "sha256-5+hTtvBnj3Nh5HitReVkLift9LTiMYVuuYx5EirN0SA=";
   };
 
-  passthru.tests = {
-    smoke-test = nixosTests.shiori;
-  };
+  nativeBuildInputs = [ installShellFiles ];
+  postInstall = ''
```
Mic92, line 19: This doesn't cross compile otherwise:

```suggestion
  postInstall = lib.optionalString (stdenv.hostPlatform != stdenv.buildPlatform) ''
```

nixos/tests/shiori.nix
```diff
@@ -1,80 +1,79 @@
-import ./make-test-python.nix ({ pkgs, lib, ...}:
+import ./make-test-python.nix ({ pkgs, lib, ... }:
```
Mic92, line 1: In future please separate formatting commits from commits that changes stuff to speed up the review process.

Mic92:
@ofborg build shiori

nixos/modules/services/web-apps/shiori.nix
```diff
@@ -1,64 +1,96 @@
...
+          Path to file containing environment variables.
+          Useful for passing down secrets.
+          <https://github.com/go-shiori/shiori/blob/master/docs/Configuration.md#overall-configuration>
+        '';
+      };
+
+      databaseUrl = lib.mkOption {
+        type = lib.types.nullOr lib.types.str;
+        default = null;
+        example = "postgresql:///shiori?host=/run/postgresql";
```
Mic92, line 47: │[eve] Jun 30 11:39:14 eve shiori[1233953]: time="2024-06-30T11:39:14Z" level=fatal msg="error opening database" error="unsupported database scheme: postgresql"

aktaboot:
Might want to update the PR title accordingly :)

Mic92:
@CaptainJawZ do these changes look good to you?

nixos/tests/shiori.nix
CaptainJawZ: ah sorry! I restored this file and didn't intended to bring it in the commit, this sneaked through me.

CaptainJawZ (commented):
all good!

CaptainJawZ:
> @CaptainJawZ do these changes look good to you?

yeah, awesome! they're already implemented in the PR right? thank you very much!

PR #307132, opened 2024-04-27
Add Intel VTune

xzfc:
## Description of changes

[website]: https://www.intel.com/content/www/us/en/developer/tools/oneapi/vtune-profiler.html
[wiki]: https://en.wikipedia.org/wiki/VTune

This PR brings Intel VTune Profiler v2025.0.1 ([website], [wiki]), a proprietary performance analysis tool for x86-based machines.
This PR consists of the following additions:

### 1. Application Package
1. Name: `vtune-oneapi-vtune`.
2. Ernomous installed size of 1.6 GiB.
3. Provides a desktop entry for `vtune-gui`, a CLI utility `vtune`, and a lot more.
4. Uses the same package name as the official deb/rpm packages made by Intel.
5. Most of the files are installed to `$out/opt/intel/oneapi/vtune`, following the same structure as the official packages.
6. The `opt/…/bin64` directory contains a lot of garbage, so I've added only the bare minimum to `$out/bin`.

### 2. Sampling Drivers Package
1.  Name: `linuxPackages.vtune-sepdk`.
2.  Kernel drivers to enable hardware event-based sampling for VTune Profiler. ([website](https://www.intel.com/content/www/us/en/docs/vtune-profiler/user-guide/2024-1/sep-driver.html))
3.  It's possible to use these drivers without installing the application package, e.g., if [running VTune Profiler in a Docker container](https://www.intel.com/content/www/us/en/docs/vtune-profiler/user-guide/2023-0/run-from-container.html).
4.  The package provides `*.ko` files as well as `insmod-*` and `rmmod-*` scripts (upstream) to load/unload the modules.
5.  In order to use these modules, a user should be in the `vtune` group.
6.  These modules are optional, and basic profiling could be performed by configuring `/proc/sys/kernel` parameters instead.
7.  Caveat: loading the modules directly using `modprobe` could lead to kernel panics, so using the provided scripts is recommended.
8.  Caveat: these scripts might want to create `vtune` group if it doesn't exist. The NixOS module handles this.

9.  The source of these modules is available from either:
    - The same installer package as the application, i.e., a 547.8MiB archive.
    - An 612KiB `*.tar.gz` file [at the Intel website](https://www.intel.com/content/www/us/en/developer/articles/code-sample/vtune-profiler-sampling-driver-downloads.html). It's a bit outdated and belongs to VTune v2023.1.

    I've used the former to keep versions of both packages in sync.
    We might add the later if there's a demand for it.

10. Modules self-report their licenses as "Dual BSD/GPL" (run `modinfo sep5 pax socperf3`), but:
    - The source package is a part of large proprietary archive. (see above)
    - A few header files contain an unclear license notice. Might be an oversight.

    So, I've marked the package as `unfree bsd3 gpl2Only`.

### 3. NixOS module
1.  Option name: `programs.vtune`.
2.  Enable it to install both the application and the kernel module, or just one of them.
3.  It provides a systemd service to load/unload the modules, similar to the official packages.
4.  Allocates a new group id `vtune` in `nixos/modules/misc/ids.nix`.

### 4. Dependencies
1.  `safeclib` v3.3.0. Some `*.so` in the application package is linked to this particular version of `safeclib`.
2.  Some `*.so` dependencies for the vendored python are missing.
    I'm not sure if they are worth adding as I haven't stumbled upon any issues yet.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/nixos/modules/programs/vtune.nix
+++ b/nixos/modules/programs/vtune.nix
@@ -1,0 +1,74 @@
+{
+  config,
+  lib,
+  pkgs,
+  ...
+}:
+
+let
+  inherit (lib)
+    mkEnableOption
+    mkIf
+    mkMerge
+    mkOption
+    mkPackageOption
+    types
+    ;
+
+  cfg = config.programs.vtune;
+in
+{
+  options.programs.vtune = {
+    enable = mkEnableOption "VTune Profiler";
+
+    package = mkPackageOption pkgs "intel-oneapi-vtune" {
+      nullable = true;
+      extraDescription = ''
+        Set to `null` to install only the kernel module.
+      '';
+    };
+
+    modulePackage = mkOption {
+      type = types.nullOr types.package;
+      default = config.boot.kernelPackages.intel-vtune-sepdk.override {
+        intel-oneapi-vtune = config.programs.package or pkgs.intel-oneapi-vtune;
+      };
+      defaultText = ''
+        config.boot.kernelPackages.intel-vtune-sepdk.override {
+          intel-oneapi-vtune = config.programs.vtune.package or pkgs.intel-oneapi-vtune;
+        }
+      '';
+      example = "config.boot.kernelPackages.intel-vtune-sepdk";
+      description = ''
+        The package to use for the VTune Profiler kernel module.
+        Set to `null` to disable the kernel module.
+        An user should be in the `vtune` group to use the module.
+      '';
+    };
+  };
+
+  config = mkIf cfg.enable (mkMerge [
+    (mkIf (cfg.package != null) { environment.systemPackages = [ cfg.package ]; })
+
+    (mkIf (cfg.modulePackage != null) {
+      boot.extraModulePackages = [ cfg.modulePackage ];
+
+      environment.systemPackages = [ cfg.modulePackage ];
+
+      users.groups.vtune = { };
+
+      systemd.services.vtune-sep5 = {
+        description = "Load VTune SEP5 driver";
+        wantedBy = [ "multi-user.target" ];
+        serviceConfig = {
+          Type = "oneshot";
+          ExecStart = "${cfg.modulePackage}/bin/insmod-sep --load-in-tree-driver";
+          ExecStop = "${cfg.modulePackage}/bin/rmmod-sep";
+          RemainAfterExit = true;
+        };
+      };
+    })
+  ]);
+
+  meta.maintainers = [ lib.maintainers.xzfc ];
+}
--- a/pkgs/by-name/in/intel-oneapi-vtune/package.nix
+++ b/pkgs/by-name/in/intel-oneapi-vtune/package.nix
@@ -1,0 +1,142 @@
+{
+  alsa-lib,
+  atk,
+  autoPatchelfHook,
+  cairo,
+  cups,
+  dbus,
+  elfutils,
+  expat,
+  fetchurl,
+  file-rename,
+  glib,
+  gtk3,
+  kmod,
+  lib,
+  libdrm,
+  libndctl,
+  libsafec,
+  libxcrypt-legacy,
+  libxkbcommon,
+  mesa,
+  ncurses5,
+  nspr,
+  nss,
+  opencl-clang,
+  p7zip,
+  pango,
+  stdenv,
+  systemd,
+  wrapGAppsHook3,
+  xorg,
+}:
+
+stdenv.mkDerivation (
+  finalAttrs:
+  let
+    versionMajorMinor = lib.versions.majorMinor finalAttrs.version;
+  in
+  {
+    pname = "intel-oneapi-vtune";
+    version = "2025.0.1";
+
+    src = fetchurl {
+      url = "https://installer.repos.intel.com/oneapi/vtune/lin/intel.oneapi.lin.vtune,v=2025.0.1%2B14/cupPayload.cup";
+      sha256 = "1pzh0kx7wxwr48rbv7wfkpkixqxsaass3xkpwhvvkdm09v5fhm7h";
+    };
+
+    nativeBuildInputs = [
+      autoPatchelfHook
+      file-rename
+      p7zip
+      wrapGAppsHook3
+    ];
+
+    buildInputs = [
+      alsa-lib
+      atk
+      cairo
+      cups
+      dbus
+      elfutils
+      expat
+      glib
+      gtk3
+      kmod
+      libdrm
+      libndctl
+      libsafec
+      libxcrypt-legacy
+      libxkbcommon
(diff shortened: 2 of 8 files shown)


nixos/modules/misc/ids.nix
```diff
@@ -666,6 +667,7 @@ in
       rstudio-server = 324;
       localtimed = 325;
       automatic-timezoned = 326;
+      vtune = 327;
```
Mic92, line 670: Do we actually need fixed uids here? We only do this for services that acquire a lot of state and where chown on startup to fixup ids would be too slow.

nixos/modules/programs/vtune.nix
```diff
@@ -0,0 +1,76 @@
...
+    in
+    mkIf cfg.enable (mkMerge [
+      (mkIf (cfg.package != null) { environment.systemPackages = [ cfg.package ]; })
+
+      (mkIf (cfg.modulePackage != null) {
+        boot.extraModulePackages = [ cfg.modulePackage ];
+
+        environment.systemPackages = [ cfg.modulePackage ];
+
+        users.groups.vtune.gid = config.ids.gids.vtune;
```
Mic92, line 60: This group is used for accessing some devices in /dev?

Mic92: Usually this would be enough for simple groups like this:

```suggestion
        users.groups.vtune = {};
```

xzfc: Yes, it's for `/dev/{pax,sep5,socperf3}`.
Pushed an update.

nixos/modules/misc/ids.nix
xzfc: No. I've removed this.

gfaster:
I was testing this out and I kept getting a crash with the message `No GSettings schemas are installed` - To the best of my knowledge that means `wrapGAppsHook` should be added to `nativeBuildInputs` (at least according to [this](https://discourse.nixos.org/t/custom-gsettings-schema-being-ignored-in-xdg-data-dirs/3449))

xzfc:
@gfaster
Thanks, for pointing it out. I've missed it because it works fine for my setup.
Could you please check if the crash is fixed in the updated version?

---

For the reference, I did these steps to reproduce the crash:
1. `killall xsettingsd`
2. `env -i DISPLAY=$DISPLAY HOME=$HOME ./result/bin/vtune-gui`
3. Press `^O` to invoke open file dialog.
4. ```console
   (vtune-gui:828176): GLib-GIO-ERROR **: 02:40:46.745: Settings schema 'org.gtk.Settings.FileChooser' is not installed
   zsh: trace trap (core dumped)  env -i DISPLAY=$DISPLAY HOME=$HOME ./result/bin/vtune-gui
   ```

gfaster:
I'm not in a great position to test the full PR, but the `wrapGAppsHook` change matches my personal fix.

pkgs/os-specific/linux/vtune-sepdk/default.nix
```diff
@@ -0,0 +1,106 @@
...
+    procps
+    python3
+    shadow
+    su
+    which
+  ];
+  libexecDir = "\${out}/libexec/vtune-sepdk";
+  modDestDir = "$out/lib/modules/${kernel.modDirVersion}/kernel/drivers/platform/x86";
+  version_majmin = builtins.head (
+    builtins.match "([0-9]+\.[0-9]+)\.[0-9]+" intel-oneapi-vtune.version
```
SuperSandro2000, line 36: Can't this use lib.versions.majorMinor? Same question for other similar usages.

pkgs/os-specific/linux/vtune-sepdk/default.nix
```diff
@@ -0,0 +1,106 @@
...
+  binPath = lib.makeBinPath [
+    bash
+    coreutils
+    gnugrep
+    gnused
+    kmod
+    procps
+    python3
+    shadow
+    su
```
SuperSandro2000, line 30: that is useless here as it does not have the setuid bit in the nix store. We nee to rely on the wrapper.

pkgs/by-name/in/intel-oneapi-vtune/package.nix
```diff
@@ -0,0 +1,142 @@
...
+  stdenv,
+  systemd,
+  wrapGAppsHook3,
+  xorg,
+}:
+
+stdenv.mkDerivation (
+  finalAttrs:
+  let
+    version_majmin = builtins.head (builtins.match "([0-9]+\.[0-9]+)\.[0-9]+" finalAttrs.version);
```
SuperSandro2000, line 37: same here

lf-:
I have tested this on some Intel hardware and it seems to work properly for me, minus some bugs that are definitely upstream problems, like difficulties interrupting builds, and issues with showing source in comparison view when the sources are different (busted lol, but not surprising). Thank you for packaging this; I had made some abortive attempts to package it earlier involving FHS envs (in which their executable would `ud2` SIGABRT for unknown reasons) and their installer being very broken and FHS envs being kind of busted inside builders that I could throw away after finding your PR.

Maybe the `vtune-sepdk` package should be called `intel-oneapi-vtune-sepdk` or idk. It's confusing that it's not at least prefixed `intel`.

pkgs/os-specific/linux/vtune-sepdk/default.nix
xzfc: Fixed.

---
For the reference, I've took this (anti-)pattern from `libcutl`.
https://github.com/NixOS/nixpkgs/blob/79723dbb2ad738817e2a1a52d2954def4bb1f004/pkgs/development/libraries/libcutl/default.nix#L24
I've fixed it for `libcutl` as well: #332257.

NthTensor:
What's the status of this? Anything I can do to help move this along?

xzfc:
> What's the status of this? Anything I can do to help move this along?

I still want it to get merged, but currently have no capacity to push it forward. (particularly, finding/pinging reviewers) 

In a week, I'll rebase this branch to the latest `master`, and bump VTune version.

I'd be grateful if you or someone else will bring someone willing to review and merge it.

NthTensor:
I will attempt a review, but I'm not sure what the merge process is for this repo.

robertodr (approved):

LockedThread:
Any updates on this? Is there anything that I can do to help?

peigongdsd:
Any updates?

PR #307184, opened 2024-04-27
llvmPackages_*.libcxx: include libcxxabi within libcxx

pwaller:
Test case: nixpkgs#pkgsStatic.pkgsLLVM.ncurses

Prior to this patch, this fails with errors such as:

```
error: undefined symbol: __cxa_throw
```

I think this is a reasonable solution because in #292043, libcxxabi was
'merged into libcxx', however, the commit message suggests that only
dynamic linking was accounted for, because it says:

```
* linux/freebsd `libc++.so` is a linker script `LINK(libc++.so.1, -lc++abi)` making `-lc++` sufficient.
```

Whereas, I found that if I tried linking a "hello world" C++ program
with a static hostPlatform, it failed unless -lc++abi was passed.

Signed-off-by: Peter Waller <p@pwaller.net>

## Description of changes

Set `-DLIBCXX_STATICALLY_LINK_ABI_IN_STATIC_LIBRARY=On` for hostPlatform.isStatic so that we get the equivalent behaviour is for dynamic linking that -lc++ is enough to pull in all the needed code.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [x] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [x] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/development/compilers/llvm/common/libcxx/default.nix
+++ b/pkgs/development/compilers/llvm/common/libcxx/default.nix
@@ -67,6 +67,12 @@ let
 
   cxxCMakeFlags = [
     "-DLIBCXX_CXX_ABI=${cxxabiName}"
+  ] ++ lib.optionals (cxxabi == null && lib.versionAtLeast release_version "16") [
+    # Note: llvm < 16 doesn't support this flag (or it's broken); handled in postInstall instead.
+    # Include libc++abi symbols within libc++.a for static linking libc++;
+    # dynamic linking includes them through libc++.so being a linker script
+    # which includes both shared objects.
+    "-DLIBCXX_STATICALLY_LINK_ABI_IN_STATIC_LIBRARY=ON"
   ] ++ lib.optionals (cxxabi != null) [
     "-DLIBCXX_CXX_ABI_INCLUDE_PATHS=${lib.getDev cxxabi}/include"
   ] ++ lib.optionals (stdenv.hostPlatform.isMusl || stdenv.hostPlatform.isWasi) [
@@ -126,6 +132,31 @@ stdenv.mkDerivation (rec {
   postInstall = lib.optionalString (cxxabi != null) ''
     lndir ${lib.getDev cxxabi}/include $dev/include/c++/v1
     lndir ${lib.getLib cxxabi}/lib $out/lib
+    libcxxabi=$out/lib/lib${cxxabi.libName}.a
+  ''
+  # LIBCXX_STATICALLY_LINK_ABI_IN_STATIC_LIBRARY=ON doesn't work for LLVM < 16 or
+  # external cxxabi libraries so merge libc++abi.a into libc++.a ourselves.
+
+  # GNU binutils emits objects in LIFO order in MRI scripts so after the merge
+  # the objects are in reversed order so a second MRI script is required so the
+  # objects in the archive are listed in proper order (libc++.a, libc++abi.a)
+  + lib.optionalString (cxxabi != null || lib.versionOlder release_version "16") ''
+    libcxxabi=''${libcxxabi-$out/lib/libc++abi.a}
+    if [[ -f $out/lib/libc++.a && -e $libcxxabi ]]; then
+      $AR -M <<MRI
+        create $out/lib/libc++.a
+        addlib $out/lib/libc++.a
+        addlib $libcxxabi
+        save
+        end
+    MRI
+      $AR -M <<MRI
+        create $out/lib/libc++.a
+        addlib $out/lib/libc++.a
+        save
+        end
+    MRI
+    fi
   '';
 
   passthru = {
--- a/pkgs/test/cc-wrapper/default.nix
+++ b/pkgs/test/cc-wrapper/default.nix
@@ -46,6 +46,17 @@ in stdenv.mkDerivation {
       $READELF -d ./atomics.so | grep libatomic.so && echo "ok" >&2 || echo "failed" >&2
     ''}
 
+    # Test that linking libc++ works, and statically.
+    ${lib.optionalString isCxx ''
+      echo "checking whether can link with libc++... " >&2
+      NIX_DEBUG=1 $CXX ${./cxx-main.cc} -c -o cxx-main.o
+      NIX_DEBUG=1 $CC cxx-main.o -lc++ -o cxx-main
+      NIX_DEBUG=1 $CC cxx-main.o ${lib.getLib stdenv.cc.libcxx}/lib/libc++.a -o cxx-main-static
+      ${emulator} ./cxx-main
+      ${emulator} ./cxx-main-static
+      rm cxx-main{,-static,.o}
+    ''}
+
     ${lib.optionalString (stdenv.isDarwin && stdenv.cc.isClang) ''
       echo "checking whether compiler can build with CoreFoundation.framework... " >&2
       mkdir -p foo/lib


pwaller:
I sent this to master rather than staging on the understanding that it should be reasonable to do this for `pkgsStatic.pkgsLLVM`, which is likely broken without this change anyway.

annaleeleaves:
seems reasonable. see also https://github.com/NixOS/nixpkgs/pull/305876 (that was my PR as was the libcxx / libcxxabi merge but i deleted my github account).

pwaller:
Thanks for the reference to the closed PR, I had missed that and the additional context. I would like to see this fix landed so I'll leave this open for comments; I take it that this won't fix things for older LLVMs. Advice welcomed.

I'd be interested if the community might also accept this since it fixes the more recent toolchains at least.

I've pulled across the test you wrote for #305876.

pwaller:
/cc potentially interested reviewers from the other thread:
@Ericson2314 @rrbutani @RossComputerGuy @Mindavi

pkgs/test/cc-wrapper/default.nix
```diff
@@ -46,6 +46,17 @@ in stdenv.mkDerivation {
       $READELF -d ./atomics.so | grep libatomic.so && echo "ok" >&2 || echo "failed" >&2
     ''}
 
+    # Test that linking libc++ works, and statically.
+    ${lib.optionalString isCxx ''
```
Ericson2314, line 50: Is this the right condition? we need to make sure we're using `libc++` not `libstdc++` or whatever.

pwaller: Good spot, but I think the condition is actually the correct one from how isCxx is currently defined; it actually is asking if it's libcxx. For libstdc++ environments, `stdenv.cc.libcxx` is null:

https://github.com/NixOS/nixpkgs/blob/32f0167ef1422bfee3f1ee4576d5d992a489d24e/pkgs/test/cc-wrapper/default.nix#L12

pkgs/development/compilers/llvm/common/libcxx/default.nix
```diff
@@ -63,6 +63,8 @@ let
     "-DLIBCXXABI_ENABLE_EXCEPTIONS=OFF"
   ] ++ lib.optionals (!enableShared) [
     "-DLIBCXXABI_ENABLE_SHARED=OFF"
+  ] ++ lib.optionals stdenv.hostPlatform.isStatic [
+    "-DLIBCXX_STATICALLY_LINK_ABI_IN_STATIC_LIBRARY=On"
```
rrbutani, line 67: What do you think about _not_ gating this on `hostPlatform.isStatic`?

I know using `cc` from `!isStatic` stdenvs for static linking isn't a supported configuration but I think this still ends up happening for some out-of-nixpkgs use cases (i.e. `devShell`s) and we _do_ provide `libc++.a` even when `!hostPlatform.isStatic`...

rrbutani (approved):
This looks good, thank you!

Just one nit/question:

pkgs/development/compilers/llvm/common/libcxx/default.nix
pwaller: Seems reasonable to me. I'll try tomorrow and see what hapoens.

pkgs/test/cc-wrapper/default.nix
```diff
@@ -46,6 +46,17 @@ in stdenv.mkDerivation {
...
+      echo "checking whether can link with libc++... " >&2
+      $CXX ${./cxx-main.cc} -c -o cxx-main.o
+      $CC cxx-main.o -lc++ -o cxx-main
+      $CC cxx-main.o -l:libc++.a -o cxx-main-static
+      ${emulator} ./cxx-main
+      ${emulator} ./cxx-main-static
+      rm cxx-main{,-static,.o}
+    ''}
+
```
annaleeleaves, line 59: this test fails `nix-build -A tests.cc-wrapper` because it tries to run on vanilla llvm[12-18] and this change only modifies a subset of the llvm16+ packages. eg:

@ofborg build tests.cc-wrapper

(note aarch64-linux fails for other reasons, #285016)

annaleeleaves:
@ofborg build tests.cc-wrapper

pkgs/test/cc-wrapper/default.nix
annaleeleaves: `nix-build -A tests.cc-wrapper` also fails on darwin due to the linker not supporting `-l:libc++.a` format flags.

>  checking whether can link with libc++...
> ld: library not found for -l:libc++.a
 > clang-16: error: linker command failed with exit code 1 (use -v to see invocation)

but just verified using the full library path works on x64 linux & x64 darwin.
```diff
- $CC cxx-main.o -l:libc++.a -o cxx-main-static
+ $CC cxx-main.o ${lib.getLib stdenv.cc.libcxx}/lib/libc++.a -o cxx-main-static
```

(PR https://github.com/NixOS/nixpkgs/pull/305876 needed to rebuild darwin stdenv so test timed out building before able to run).

pwaller: I think I've accounted for these two comments now, thank you (please shout if I seem to have missed anything).

pkgs/development/compilers/llvm/common/libcxx/default.nix
pwaller: Done.

pwaller:
@ofborg build tests.cc-wrapper

pwaller:
I think I've accounted for everything that was known (and mentioned in the other thread https://github.com/NixOS/nixpkgs/pull/305876).

Most recent update:
* For versions before 16, merge the two archive files with `ar`.
* Fix the cc-wrapper tests.
* Update PR title / commit message.

sternenseemann (commented):
Can you add an extra comment to the two conditionals (`!lib.versionAtLeast release_version "16"` and `lib.versionAtLeast release_version "16"`) noting that the other versions have a different way of achieving the same, so that readers of the code will be able to connect the dots?

pwaller:
@sternenseemann wrote:
> Can you add an extra comment to the two conditionals
Done. Latest patch also removes `touch $out/lib/workaround-present` from postInstall which I never intended to push.

pkgs/development/compilers/llvm/common/libcxx/default.nix
```diff
@@ -123,9 +129,28 @@ stdenv.mkDerivation (rec {
...
   # libc++.so.1 and libc++abi.so or the external cxxabi. ld-wrapper doesn't
   # support linker scripts so the external cxxabi needs to be symlinked in
-  postInstall = lib.optionalString (cxxabi != null) ''
+  postInstall = ''
+    libcxxabi=$out/lib/libc++abi.a
+  '' + lib.optionalString (cxxabi != null) ''
     lndir ${lib.getDev cxxabi}/include $dev/include/c++/v1
     lndir ${lib.getLib cxxabi}/lib $out/lib
+    libcxxabi=$out/lib/lib${cxxabi.libName}.a
+  '' + lib.optionalString (!lib.versionAtLeast release_version "16") ''
```
sternenseemann, line 138: Why doesn't this also check for `cxxabi == null`?

annaleeleaves: `cxxabi != null` is true for freebsd. `nix-build -A pkgsCross.x86_64-freebsd.llvmPackages_17.libcxx` tho need to modify libcxxrt to something like PR https://github.com/NixOS/nixpkgs/pull/306238 so libcxxrt.a exists in the non pkgsStatic case (no idea if freebsd works under pkgsStatic).

however, it seems like that is causing an infinite recursion so some change was committed between now and  https://github.com/NixOS/nixpkgs/pull/305876 which broke freebsd eval.

to drop into a freebsd builder to try linking / building hello.cxx by hand do something like.

`nix-shell -I nixpkgs=. --expr 'with import <nixpkgs> {}; pkgsCross.x86_64-freebsd.stdenv.mkDerivation { name = "abc"; }'`

AFAIK even if you get freebsd to build the libc it generates segfaults when used on actual freebsd, but maybe that will be fixed soon.

pkgs/development/compilers/llvm/common/libcxx/default.nix
```diff
@@ -123,9 +129,28 @@ stdenv.mkDerivation (rec {
...
+      cp -d $libcxxabi $scratch/libcxxabi.a
+      $AR -M <<MRI
+      create libcxx.a
+      addlib $scratch/libcxx.a
+      addlib $scratch/libcxxabi.a
+      save
+      end
+    MRI
+      mv libcxx.a $out/lib/libc++.a
+      rm -rf $scratch
```
annaleeleaves, line 152: this can be simplified to:
```suggestion
     $AR -M <<MRI
        create $out/lib/libc++.a
        addlib $out/lib/libc++.a
        addlib $libcxxabi
        save
        end
    MRI
```

`ar` in newer versions of binutils can handle `+`.  (i initially was testing with ar from binutils 2.38 which can't handle `+`).

also. note that gnu binutils adds objects to the archive in LIFO order, so when you look at the result `ar t result/lib/libc++.a` the objects will be reversed. this can be fixed, if desired, by adding an additional mri script.
```bash
      $AR -M <<MRI
        create $out/lib/libc++.a
        addlib $out/lib/libc++.a
        save
        end
    MRI
```

this is what is done in the diff embedded in  https://github.com/NixOS/nixpkgs/pull/307184#issuecomment-2080457122

pkgs/development/compilers/llvm/common/libcxx/default.nix
annaleeleaves: i agree that this needs the `cxxabi != null` check. if you're going to leave it out for some additional PR the hackery with `libcxxabi=xxx` can be removed and just `$out/lib/libc++abi.a` used instead.

ghost (commented):
this will now need need to target staging.

to support `cxxabi != null` and all LLVM versions my preference is to use the diff embedded in  https://github.com/NixOS/nixpkgs/pull/307184#issuecomment-2080457122 since that is the minimal change. I don't see the value in using both `LIBCXX_STATICALLY_LINK_ABI_IN_STATIC_LIBRARY` and `ar` given the extra complexity and the fact that the `ar` script is 4 lines long.

also note that using `LIBCXX_STATICALLY_LINK_ABI_IN_STATIC_LIBRARY` prepends libc++abi.a to libc++.a which is the reverse order of what the shared libs do in the linker script. it probably doesn't really matter except if there are objects that in libc++abi and libc++ which contain the same symbol.

nonetheless, some minor comments inline.

pwaller:
I've reinstated @annaleeleaves's diff for generating the AR and retested on x86_64-linux. I wasn't able to get a freebsd test to work.

pwaller:
Also retargeted staging.

annaleeleaves:
@pwaller thanks for working on this but unfortunately the change still is not correct.
`nix-build -A libcxx  && ar t result/lib/libc++.a` shows duplicate objects in libc++.a

pwaller:
Thanks @annaleeleaves - Note that I didn't get a notification for your edit which added suggested changes, so I was unaware of it until now.

Also ambiguous to me is whether the suggestions address your comment that the symbols are duplicated. I had a quick look for duplicated symbols with nm building `pkgsStatic.pkgsLLVM.llvmPackages_17.libcxx`, but I don't see them. If the problem is still present please would you mind sharing what test I can use as to whether it's right?

pwaller:
Forgot to mention in the above comment: I believe I've applied your suggestions.

annaleeleaves:
> Also ambiguous to me is whether the suggestions address your comment that the symbols are duplicated. I had a quick look for duplicated symbols with nm building `pkgsStatic.pkgsLLVM.llvmPackages_17.libcxx`, but I don't see them. If the problem is still present please would you mind sharing what test I can use as to whether it's right?

it is fixed with the suggested changes. i listed a test in https://github.com/NixOS/nixpkgs/pull/307184#issuecomment-2084096802 `nix-build -A libcxx  && ar t result/lib/libc++.a`. `ar t result/lib/libc++.a` lists the objects in the archive. before this change the objects from `libc++abi.a` were added twice for llvm16+, with the change only once. `nm` also shows the duplicates. before this change there were 8024 lines listed for llvm17 libcxx, after this change there are 6364.

ghost (approved):
change looks good. my preference is still the diff in https://github.com/NixOS/nixpkgs/pull/307184#issuecomment-2080457122 but this works too.

tested linux x64 `nix-build -A tests.cc-wrapper` and also tested the changes work when libcxxrt is used via slight modifications to libcxxrt and libcxx.

also tested `nix-build -A tests.cc-wrapper.llvmTests` on darwin x64 with minor changes to skip llvm16 so to not require building stdenv.

annaleeleaves:
@ofborg build tests.cc-wrapper.llvmTests

RossComputerGuy (approved):

rrbutani (approved):
@annaleeleaves, @pwaller: Thanks for your work on this.

annaleeleaves:
@wegank
would be good to get this in for 24.05 https://github.com/NixOS/nixpkgs/issues/303285 as there are only one or two staging cycles left. don't think this would be considered a breaking change as it fixes a bug introduced in https://github.com/NixOS/nixpkgs/pull/292043 (which removed `-lc++abi` from `cc-wrapper` for `clang++` and thus changed the behaviour for linking static libs).

this change not only rectifies that bug but also fixes long standing bug `clang -x c++ -lc++ hello.cxx` for static linking ( https://github.com/NixOS/nixpkgs/pull/292043 fixed it for shared libs).

pwaller:
I don't have commit rights, please can someone merge?

Thanks for the review input, all!

kvtb:
Does it have sense to have two blocks with "create $out/lib/libc++.a" commands in sequence?
The second overwrites what did he first

pwaller:
> The second overwrites what did the first
This comment discusses the rationale; https://github.com/NixOS/nixpkgs/pull/307184#discussion_r1582419756

AFAIU it has the effect of reversing the order of objects in the archive, so the second invocation sets the order straight.

PR #307268, opened 2024-04-27
treewide: pin zig package in top-level instead of the function

AndersonTorres:
## Description of changes

In order to obtain a more `by-name` -compatible override interface.

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---


wegank:
Merge conflict, please rebase.

Also, should we move `zig = zig_0_11;` to aliases, so that every `zig` in Nixpkgs is guaranteed to be overridden?

AndersonTorres:
> Also, should we move `zig = zig_0_11;` to aliases, so that every `zig` in Nixpkgs is guaranteed to be overridden?

Since Zig is still too unstable, the rationale we adopted is:

- each package that uses zig should unconditionally set the suitable version for this package at top-level (as shown in this PR);
  - before the by-name hierarchy, we did it inside the expression. Now with by-name, we should do it outside.
- as a convenience (for e.g. `nix-shell` users), the `zig` handler should be set to the latest version we packaged.

I believe setting zig in `aliases.nix` will not be a good move, given the second point above.

wegank (approved):

Aleksanaa:
I personally don't get the need of pinning it this way, and it makes the version used less obvious. As you have stated it is often not the case that the latest zig version works for all packages (right?), so we can just leave it as-is.

AndersonTorres:
> I personally don't get the need of pinning it this way

This is to keep the [`.override` interface](https://nixos.org/manual/nixpkgs/stable/#sec-pkg-override) uniform and predicatble.

"Why a uniform `.override` interface is useful?"
Because a uniform `.override` interface is easier to override, facilitating the use of [overlays](https://nixos.org/manual/nixpkgs/stable/#sec-overlays-definition) and other methods of customization.

You can look at the Nixpkgs Manual for some interesting examples and concepts:

https://nixos.org/manual/nixpkgs/stable/#sec-overlays-alternatives
https://nixos.org/manual/nixpkgs/stable/#sec-overlays-alternatives-mpi
https://nixos.org/manual/nixpkgs/stable/#usage-1-pkgs.zlib.override

And a code that overrides `zig = prev.zig_0_11;` is way cleaner than `zig_0_9 = prev.zig_0_11;`

> , and it makes the version used less obvious.

How exactly? The version is clearly stated at all-packages.nix.

> As you have stated it is often not the case that the latest zig version works for all packages (right?)

Yes. Zig makes some breaking changes in versions before 1.0, and sometimes the upstream developer did not update his code yet.

Because of it, we had that unwritten rule [I wrote above](https://github.com/NixOS/nixpkgs/pull/307268#issuecomment-2081442441).

> Also people are less likely to override it (for what?), unlike other runtime dependencies

I don't think it is a sufficient reason to reject this patch.
The "less likely" people that want to override it will have an annoyance to deal, and well, _rejecting minorities' concerns_ is not a nice thing.

Aleksanaa:
Short question: will we have things like `callPackage ... { gtk = gtk4; }`?

-----

I find you to be much more adventurous than most of us in dealing with such issues. At present, I even doubt if there is a complete consensus on "keep overriding interface", especially in practice. Can we open an issue first to discuss whether this type of modifications are valuable and whether we should enforce them? If you see an example today, change it, and give your reasons, and tomorrow you see another example, things can get even more confusing if the cycle continues.

AndersonTorres:
> Short question: will we have things like `callPackage ... { gtk = gtk4; }`?

I don't think so. GTK3 and GTK4 are basically inbred forks - different softwares that share a common repo, like Python 2 (abandonware) and 3.

> 
> -----
> 
> I find you to be much more adventurous than most of us in dealing with such issues. At present, I even doubt if there is a complete consensus on "keep overriding interface", especially in practice. 

With Emacs we used some guards in order to warn users about deprecated and renamed parameters.
Hey, the NixOS module system is full of backward-compatibility layers.

So I believe that consistency in the names of inputs is a desirable outcome. The whole system is directed towards it.

> Can we open an issue first

I will open the issue, but I will not condition any PR to it.

iFreilicht (approved):
I agree with this change. It makes the packages fit the [`by-name` guidelines specific mention of this case](https://github.com/NixOS/nixpkgs/tree/master/pkgs/by-name#changing-implicit-attribute-defaults). As I understand, these were part of the RFC introducing by-name, so I don't think this is the place for re-negotiating that.


I also ran nixpkgs-review on this PR for both x86_64-linux and aarch64-darwin, and no packages were built, proving that no inputs were changed.

Additionally, this blocks https://github.com/NixOS/nixpkgs/pull/308248, which is part of the 24.05 milestone.

AndersonTorres:
Merge conflict. Fuck this shit, I am too tired.

PR #307320, opened 2024-04-27
stduuid: fix cmake version

OlivierLDff:
stduuid reported the wrong version to cmake, causing
`find_package(stduuid 1.2.3)` to fail (but `find_package(stduuid 1.0)` was working which is NOT expected).
Patch is from this [PR](https://github.com/mariusbancila/stduuid/pull/79).

This will need to be updated if `stduuid` is getting maintained again one day.

```console
$ cat /nix/store/srxjm722bl6n012q0s2xx4w7scwbrv7j-stduuid-1.2.3/lib/cmake/stduuid/stduuid-config-version.cmake
# This is a basic version file for the Config-mode of find_package().
# It is used by write_basic_package_version_file() as input file for configure_file()
# to create a version-file which can be installed along a config.cmake file.
#
# The created file sets PACKAGE_VERSION_EXACT if the current version string and
# the requested version string are exactly the same and it sets
# PACKAGE_VERSION_COMPATIBLE if the current version is >= requested version.
# The variable CVF_VERSION must be set before calling configure_file().

set(PACKAGE_VERSION "1.2.3")
```

Before it was `set(PACKAGE_VERSION "1.0")`


pkgs/development/libraries/stduuid/fix-version.patch
```diff
@@ -0,0 +1,60 @@
+From c889d9f8b419ced3254ec552c81993bab81fa841 Mon Sep 17 00:00:00 2001
```
AndersonTorres, line 1: Use `fetchpatch` instead of vendoring it.

AndersonTorres (changes requested):

pkgs/development/libraries/stduuid/fix-version.patch
OlivierLDff: Done:
- I'm not sure of how reliable is the url I used as this is from a PR. Maybe there is a better way to get a `.patch` from a PR in github? This seems like it could break any time if the author update his `master` branch? Or am I missing something?
- I squashed my amended my old commit, this is the convention I understood from the contributing guide is ok?

AndersonTorres: Use the long format with git hash.

OlivierLDff: Because the fixed commits are in the `master` branch of not the upstream repo, I copied the patch to my own fork on its dedicated branch and referenced that: https://github.com/OlivierLDff/stduuid/commits/fix-version-1.2.3-cmake/

I'm unsure what the github retention policy, so if the pull request author force pushed his master branch at least everything should still be there. It seems github would have kept the orphan commit accessible by sha, but this is unclear. Tell me if you want me to reference the 2 original commit instead the one in the branch of my fork.

AndersonTorres: Hum, the things are a bit more complicated.
The upstream have seen no updates in two years, and there are useful pull requests from December that weren't merged.

For now your solution is fine.

pkgs/development/libraries/stduuid/default.nix
```diff
@@ -1,4 +1,4 @@
-{ stdenv, cmake, fetchFromGitHub, lib }: let
+{ stdenv, cmake, fetchFromGitHub, fetchpatch, lib }: let
   version = "1.2.3";
 in stdenv.mkDerivation {
   name = "stduuid-${version}";
```
AndersonTorres, line 4: Well, this expression is certainly very old.

```suggestion
  pname = "stduuid";
```

AndersonTorres (commented):
Migrate this expression to by-name hierarchy:

1. Move this `default.nix` file to `pkgs/by-name/st/stduuid/package.nix`;
2. Remove the line `stduuid = callPackage ../development/libraries/stduuid { };` from `pkgs/top-level/all-packages.nix`.
3. Make a new commit with the two changes above (no need for a new PR).

pkgs/development/libraries/stduuid/default.nix
OlivierLDff: Sure, I should also add `inherit version;` right?

OlivierLDff: Done

AndersonTorres (approved):
LGTM

OlivierLDff:
Migration to `by-name` done. Thanks for the review and the time spend guided a newcomer.

AndersonTorres (approved):

shlevy (approved):

PR #307323, opened 2024-04-27
maintainers: remove endocrimes

endocrimes:
My original commit was somewhat more tongue in cheek, but I think this time I'm going to be honest, rather than hiding behind humor.

I'm angry that we as a community have reached a point where we're falling apart because people would rather accept sealioning and bad faith arguments that belong in a 2002 mailing list, not a 2024 software community.

Every time I dare come back to the Nix community as a more active contributor it feels like we have a fresh wave of controversy that we don't ever heal from, everyone just moves on, more tired, and more broken than before. And now it's crossed a line.

I'm out.

I hope the community decides to actually try and heal. Tolerating the intolerant for too long has consequences - and the [paradox
of tolerance](https://en.wikipedia.org/wiki/Paradox_of_tolerance) has come to its natural conclusion here.

If it does heal, I'll be back with a vengeance, because it turns out replacing Nix is hard - emotionally and technically, and I'll miss it.

~Danielle

follow up to #307119.

giphy

### DIFF
--- a/maintainers/maintainer-list.nix
+++ b/maintainers/maintainer-list.nix
@@ -5945,12 +5945,6 @@
     name = "Zhenbo Li";
     matrix = "@zhenbo:matrix.org";
   };
-  endocrimes = {
-    email = "dani@builds.terrible.systems";
-    github = "endocrimes";
-    githubId = 1330683;
-    name = "Danielle Lancashire";
-  };
   enorris = {
     name = "Eric Norris";
     email = "erictnorris@gmail.com";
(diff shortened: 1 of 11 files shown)


ryantm (approved):
Thank you for being a maintainer and sorry about my mixup earlier.

Artek52:
"Paradox of intolerance" aka we cant ban wrong think.  From the wikipedia article that you linked: "it seems contradictory to extend freedom of speech to extremists who ... if successful, ruthlessly suppress the speech of those with whom they disagree".  Perfectly describes left wing extremists. The problem is that they always fail, because the majority of people dont agree with them. They know that their ideas can only be sustained in an echo chamber, so they need to make the nix community one.

hraban:
Surely this is not the place to have that debate :( it's done, she's leaving, let's say goodbye like human beings and hope for better days.

Thank you for your contributions and godspeed, @endocrimes.

PR #307528, opened 2024-04-28
nixos/systemd-tmpfiles: add initrd support

WilliButz:
## Description of changes

This adds support for declaring tmpfiles rules exclusively for the
systemd initrd. Configuration is possible through the new option
`boot.initrd.systemd.tmpfiles.settings` that shares the same interface as
`systemd.tmpfiles.settings`.

I did intentionally not replicate the `rules` interface here, given that
the settings attribute set is more versatile than the list of strings
used for `rules`. This should also make it unnecessary to implement the
workaround from https://github.com/WilliButz/nixpkgs/commit/1a68e21d474c5d6005812459c9bce28168625384 again.

A self-contained `tmpfiles.d` directory is generated from the new initrd
settings and it is added to the initrd as a content path at
`/etc/tmpfiles.d`.

The stage-1 `systemd-tmpfiles-setup.service` is now altered to no longer
operate under the `/sysroot` prefix, because the `/sysroot` hierarchy
cannot be expected to be available when the default upstream service is
started.

To handle files under `/sysroot` a slightly altered version of the
upstream default service is introduced. This new unit
`systemd-tmpfiles-setup-sysroot.service` operates only under the
`/sysroot` prefix and it is ordered between `initrd-fs.target` and the
nixos activation.

Config related to tmpfiles was moved from initrd.nix to tmpfiles.nix.

## Motivation

I'd like to be able to prepare files under sysroot before switch-root in a declarative manner, without resorting to custom scripts. We already pull `systemd-tmpfiles-setup.service` into the initrd so I thought it would make sense to make it configurable.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

cc: @Lassulus (#262731)

### DIFF
--- a/nixos/modules/system/boot/systemd/tmpfiles.nix
+++ b/nixos/modules/system/boot/systemd/tmpfiles.nix
@@ -1,10 +1,121 @@
-{ config, lib, pkgs, utils, ... }:
+{ config, lib, pkgs, ... }:
 
 with lib;
 
 let
   cfg = config.systemd.tmpfiles;
+  initrdCfg = config.boot.initrd.systemd.tmpfiles;
   systemd = config.systemd.package;
+
+  settingsOption = {
+    description = ''
+      Declare systemd-tmpfiles rules to create, delete, and clean up volatile
+      and temporary files and directories.
+
+      Even though the service is called `*tmp*files` you can also create
+      persistent files.
+    '';
+    example = {
+      "10-mypackage" = {
+        "/var/lib/my-service/statefolder".d = {
+          mode = "0755";
+          user = "root";
+          group = "root";
+        };
+      };
+    };
+    default = {};
+    type = types.attrsOf (types.attrsOf (types.attrsOf (types.submodule ({ name, config, ... }: {
+      options.type = mkOption {
+        type = types.str;
+        default = name;
+        example = "d";
+        description = ''
+          The type of operation to perform on the file.
+
+          The type consists of a single letter and optionally one or more
+          modifier characters.
+
+          Please see the upstream documentation for the available types and
+          more details:
+          <https://www.freedesktop.org/software/systemd/man/tmpfiles.d>
+        '';
+      };
+      options.mode = mkOption {
+        type = types.str;
+        default = "-";
+        example = "0755";
+        description = ''
+          The file access mode to use when creating this file or directory.
+        '';
+      };
+      options.user = mkOption {
+        type = types.str;
+        default = "-";
+        example = "root";
+        description = ''
+          The user of the file.
+
+          This may either be a numeric ID or a user/group name.
+
+          If omitted or when set to `"-"`, the user and group of the user who
+          invokes systemd-tmpfiles is used.
+        '';
+      };
+      options.group = mkOption {
+        type = types.str;
+        default = "-";
+        example = "root";
+        description = ''
+          The group of the file.
+
+          This may either be a numeric ID or a user/group name.
+
+          If omitted or when set to `"-"`, the user and group of the user who
+          invokes systemd-tmpfiles is used.
+        '';
+      };
+      options.age = mkOption {
+        type = types.str;
+        default = "-";
+        example = "10d";
+        description = ''
+          Delete a file when it reaches a certain age.
+
+          If a file or directory is older than the current time minus the age
+          field, it is deleted.
+
+          If set to `"-"` no automatic clean-up is done.
+        '';
+      };
+      options.argument = mkOption {
+        type = types.str;
+        default = "";
+        example = "";
+        description = ''
+          An argument whose meaning depends on the type of operation.
+
+          Please see the upstream documentation for the meaning of this
+          parameter in different situations:
+          <https://www.freedesktop.org/software/systemd/man/tmpfiles.d>
+        '';
+      };
+    }))));
+  };
+
+  # generates a single entry for a tmpfiles.d rule
+  settingsEntryToRule = path: entry: ''
+    '${entry.type}' '${path}' '${entry.mode}' '${entry.user}' '${entry.group}' '${entry.age}' ${entry.argument}
+  '';
+
+  # generates a list of tmpfiles.d rules from the attrs (paths) under tmpfiles.settings.<name>
+  pathsToRules = mapAttrsToList (path: types:
+    concatStrings (
+      mapAttrsToList (_type: settingsEntryToRule path) types
+    )
+  );
+
+  mkRuleFileContent = paths: concatStrings (pathsToRules paths);
 in
 {
   options = {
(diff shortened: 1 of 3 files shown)


WilliButz:
I don't think this warrants its own vm test but I'm not sure where to put a test case. Should I just extend `systemd-initrd-simple`? :)

WilliButz:
While working on [a use-case for this](https://github.com/WilliButz/preservation) it became obvious that ordering `systemd-tmpfiles-setup.service` in the initrd after `initrd-root-fs.target` is not sufficient. When the /sysroot hierarchy consists of more than just a single mount for the rootfs, those other mounts are not considered when tmpfiles is started and so it might happen that tmpfiles starts before or after everything is mounted. This is not helpful when tmpfiles is used to create files inside the sysroot hierarchy.
Because of this I updated the PR to order tmpfiles after `initrd-fs.target`.

Another issue arises from using the default upstream service file in the initrd: [`upstreamWants` from initrd.nix](https://github.com/WilliButz/nixpkgs/blob/systemd-initrd/tmpfiles-settings/nixos/modules/system/boot/systemd/initrd.nix#L75-L77) causes `sysinit.target` to have `systemd-tmpfiles-setup.service` as a (soft) dependency. Also the upstream tmpfiles service contains the ordering `Before=sysinit.target`. While this works fine in stage-2, the issue is that there isn't really anything prepared for tmpfiles to work on when `sysinit.target` is reached in the initrd.
While I'm able to override the ordering (done with the last push), I'm not sure how to remove the dependency introduced by `upstreamWants` for this unit without writing a custom service for systemd-tmpfiles in the initrd.

msanft:
> there isn't really anything prepared for tmpfiles to work on when sysinit.target is reached in the initrd

Maybe we could  use the upstream unit, but rename it to something like `systemd-tmpfiles-initrd-setup.service`.

Otherwise, we might want to patch the upstream wants for this one. I don't know how much sense it makes to use the "normal" `sysinit.target`'s dependencies in the initrd.

WilliButz:
@msanft Now I think the best approach would be to create a new unit from scratch because we'd still need to override the upstream unit even if we renamed it. I'll update the PR later

ElvishJerricco:
So, I think this is going to want two separate units, `systemd-tmpfiles-setup.service` as normal, and something like `systemd-tmpfiles-setup-sysroot.service` which could be ordered after `initrd-fs.target` and before `initrd-nixos-activation.service` or something. I assume the need for this arises from needing to setup tmpfiles before activation, right? Otherwise I'm not sure what purpose this serves.

Anyway, the reason we want to keep the normal `systemd-tmpfiles-setup.service` around is because `sysinit` is still very much a thing in initrd, and we can't just be removing tmpfiles from it. For instance, there are files that `systemd-stub` will place in the initrd that are expected to moved into a more useful and correct place by `tmpfiles`, for use by other initrd services like cryptsetup; e.g. it can place TPM2 PCR signature files in initrd that need to be copied. Or, in my case, [I copy SSH host keys from a TPM2-locked volume](https://github.com/ElvishJerricco/stage1-tpm-tailscale/blob/2c07f2a531e1557965a0d483ea694fabf9a6d5bb/boot.nix#L19-L22) so I can login and unlock sysroot.

So I think the `systemd-tmpfiles-setup-sysroot.service` idea would need `--prefix=/sysroot`, and we would need to *add* `--exclude-prefix=/sysroot` to the normal `systemd-tmpfiles-setup.service`. Heck, these might be reasonable things to take upstream, potentially.

WilliButz:
@ElvishJerricco thanks for your input, I wasn't aware of that use case for tmpfiles so early in initrd.
The updated changes should reflect your proposed approach with two complementing services:
* the upstream default, now altered to exclude `/sysroot`
* a new service that is a slight variation of the default, only operating under `/sysroot` and ordered between `initrd-fs.target` and activation
 
with both services consuming the same configuration, rendered from `boot.initrd.systemd.tmpfiles.settings`.

nikstur (approved):

msanft (approved):
Future improvement might be to actually share the options with the "normal" tmpfiles service, but I think what's currently there is good enough for now.

nixos/modules/system/boot/systemd/tmpfiles.nix
```diff
@@ -254,5 +276,62 @@ in
...
+      services.systemd-tmpfiles-setup-sysroot = {
+        description = "Create Volatile Files and Directories in the Real Root";
+        after = [ "initrd-fs.target" ];
+        before = [
+          "initrd-nixos-activation.service"
+          "shutdown.target" "initrd-switch-root.target"
+        ];
+        conflicts = [ "shutdown.target" "initrd-switch-root.target" ];
+        wantedBy = [ "initrd.target" ];
+        serviceConfig = {
```
ElvishJerricco, line 306: I think you're missing `RemainAfterExit=true`, not that I think it matters much.

nixos/modules/system/boot/systemd/tmpfiles.nix
```diff
@@ -254,5 +276,62 @@ in
...
       "R! /nix/var/nix/temproots             -    -    -    - -"
     ];
+
+    boot.initrd.systemd = {
+      additionalUpstreamUnits = [
+        "systemd-tmpfiles-setup-dev.service"
+        "systemd-tmpfiles-setup.service"
+      ];
+      # make sure all the /dev nodes are set up
+      services.systemd-tmpfiles-setup-dev.wantedBy = ["sysinit.target"];
```
ElvishJerricco, line 286: I *think* this is actually handled by `sysinit.target.wants`: https://github.com/NixOS/nixpkgs/blob/66ea67f394126fce803fb9ee03f9097637655eb0/nixos/modules/system/boot/systemd/initrd.nix#L76-L78

nixos/modules/system/boot/systemd/tmpfiles.nix
```diff
@@ -254,5 +276,62 @@ in
       "R! /nix/var/nix/gcroots/tmp           -    -    -    - -"
       "R! /nix/var/nix/temproots             -    -    -    - -"
     ];
+
+    boot.initrd.systemd = {
+      additionalUpstreamUnits = [
```
ElvishJerricco, line 281: Something we've neglected to add so far has been `systemd-tmpfiles-setup-dev-early.service`, so now would be a good time to do that.

ElvishJerricco (changes requested):
Overall, this is great! I do think it could make sense to take some of this upstream. Something that would be fairly typical of systemd's code would be having `systemd-tmpfiles-setup` recognize when it's in initrd and automatically exclude `/sysroot`, unless `--prefix=/sysroot` was given (implying it's actually `systemd-tmpfiles-setup-sysroot.service`). But we don't have to upstream it for this PR.

nixos/modules/system/boot/systemd/tmpfiles.nix
WilliButz: Good catch, I'll add it

nixos/modules/system/boot/systemd/tmpfiles.nix
WilliButz: Hmm, I can't see how. Am I missing something?
https://github.com/systemd/systemd/blob/v256.4/units/sysinit.target

nixos/modules/system/boot/systemd/tmpfiles.nix
WilliButz: We pull that in already, but in [boot/systemd.nix](https://github.com/NixOS/nixpkgs/blob/102b1152a6ea69f4be9d66bde3cb3141d5642b6c/nixos/modules/system/boot/systemd.nix#L56). I'll move it here to tmpfiles.nix, seems appropriate.

ElvishJerricco (approved):
It all looks ok to me. I *kinda* want there to be a nixos test, but that's not super important.

PR #307609, opened 2024-04-29
add a gdb wrapper capable of lazy loading pwndbg's gdbinit.py

fidgetingbits:
## Description of changes

Adds a gdb wrapper whose environment is capable of lazy loading pwndbg via gdbinit.py. The use case is that I very often have to run gdb without pwndbg in order to setup initial debugging and attach to remote targets, to avoid crashes/errors, but once I hit the state that I actually want to proper analysis I load pwndbg. Some of the things that caused this to be required can be (and have been) fixed upstream in pwndbg, but the reality is day-to-day I still need use this approach and have been for years.

So the idea is from your dev shell (or wherever) you would use gdb-with-pwndbg-env (via aliasing gdb cmd or manually running), then when you actually need pwndbg you source ~/.nix-profile/share/pwndbg/gdbinit.py. The way I've been dealing with this in my devshells is using something like this: 
```bash
sed -e 's/^exec \(.*\)-x.*/export PWNDBG_VENV_PATH=PWNDBG_PLEASE_SKIP_VENV\nexec \1 "\$@"/g' ${pkgs.lib.getBin pkgs.pwndbg}/bin/pwndbg > debug/gdb-with-pwndbg-env
``` 

But, I'd rather have a solution in nixpkgs itself if possible, as it will benefit other pwndbg users.

I'm open to a different name for `gdb-with-pwndbg-env`; I'm not sure if there's precedent for this type of thing elsewhere in nixpkgs, so maybe there is a standard.

@mic92 @patryk4815 @msanft 

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/by-name/pw/pwndbg/package.nix
+++ b/pkgs/by-name/pw/pwndbg/package.nix
@@ -4,8 +4,8 @@
 , fetchFromGitHub
 , makeWrapper
 , gdb
+,
 }:
-
 let
   pwndbg-py = python3.pkgs.pwndbg;
 
@@ -13,7 +13,8 @@ let
 
   binPath = lib.makeBinPath ([
     python3.pkgs.pwntools # ref: https://github.com/pwndbg/pwndbg/blob/2022.12.19/pwndbg/wrappers/checksec.py#L8
-  ] ++ lib.optionals stdenv.isLinux [
+  ]
+  ++ lib.optionals stdenv.isLinux [
     python3.pkgs.ropper # ref: https://github.com/pwndbg/pwndbg/blob/2022.12.19/pwndbg/commands/ropper.py#L30
     python3.pkgs.ropgadget # ref: https://github.com/pwndbg/pwndbg/blob/2022.12.19/pwndbg/commands/rop.py#L32
   ]);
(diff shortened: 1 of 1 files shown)


Mic92:
Quiet a long name but I can also not think of a better one.

Mic92:
I think the standard is usually to have a gdb.withPlugins wrapper. But potentially the gdb ecosystem is not big enough for that.

patryk4815:
maybe better option would be something like `pwndbg --lazy-init`

Mic92:
What do you think about this solution? https://github.com/NixOS/nixpkgs/pull/307655
It should still allow you to load `~/.nix-profile/share/pwndbg/gdbinit.py` but now you can just use the normal gdb executable.

fidgetingbits:
That's much better imo, tested and works fine afaict. Happy to go that route unless other maintainers see any issue. I'll close this PR.

PR #305273, opened 2024-04-19
Revert "dbus: Switch default implementation to broker"

peterhoeg:
## Description of changes

This reverts commit e827697fd3435b458a759ef0b92de7f26624b910.

We need a more elegant way to handle this as it currently breaks upgrades on some systems.

Ref:
 - Fixes #302771
 - #299812

## Things done

- Built on platform(s)
  - [X] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
- For non-Linux: Is sandboxing enabled in `nix.conf`? (See [Nix manual](https://nixos.org/manual/nix/stable/command-ref/conf-file.html))
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:
  - [NixOS test(s)](https://nixos.org/manual/nixos/unstable/index.html#sec-nixos-tests) (look inside [nixos/tests](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests))
  - and/or [package tests](https://github.com/NixOS/nixpkgs/blob/master/pkgs/README.md#package-tests)
  - or, for functions and "core" functionality, tests in [lib/tests](https://github.com/NixOS/nixpkgs/blob/master/lib/tests) or [pkgs/test](https://github.com/NixOS/nixpkgs/blob/master/pkgs/test)
  - made sure NixOS tests are [linked](https://nixos.org/manual/nixpkgs/unstable/#ssec-nixos-tests-linking) to the relevant packages
- [ ] Tested compilation of all packages that depend on this change using `nix-shell -p nixpkgs-review --run "nixpkgs-review rev HEAD"`. Note: all changes have to be committed, also see [nixpkgs-review usage](https://github.com/Mic92/nixpkgs-review#usage)
- [ ] Tested basic functionality of all binary files (usually in `./result/bin/`)
- [24.05 Release Notes](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2405.section.md) (or backporting [23.05](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2305.section.md) and [23.11](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2311.section.md) Release notes)
  - [ ] (Package updates) Added a release notes entry if the change is major or breaking
  - [ ] (Module updates) Added a release notes entry if the change is significant
[...]

Issue #302771, opened 2024-04-09
dbus-broker fails to start


SFrijters:
### Describe the bug

Presumably a result of https://github.com/NixOS/nixpkgs/pull/299812

I noticed that one of my servers that has `system.autoUpgrade.enable = true;` failed to perform its update/switch.

I have rolled back to the previous working configuration and followed the instructions in https://discourse.nixos.org/t/would-it-be-feasible-to-make-dbus-broker-the-default-dbus-implementation/41760/20 to not switch the live system, but after boot the new `dbus-broker` service fails anyway:

```
$ sudo systemctl status --failed
× dbus-broker.service - D-Bus System Message Bus
     Loaded: loaded (/etc/systemd/system/dbus-broker.service; enabled; preset: enabled)
    Drop-In: /nix/store/9z7ykk0lf37nzy0l1qdaqib1dpxp6j2f-system-units/dbus-broker.service.d
             └─overrides.conf
     Active: failed (Result: exit-code) since Wed 2023-12-06 21:38:40 CET; 4 months 2 days ago
   Duration: 8ms
TriggeredBy: × dbus.socket
       Docs: man:dbus-broker-launch(1)
    Process: 1349 ExecStart=/nix/store/xp5gvw55ga4cs2vkj45vlyks75ppdbkl-dbus-broker-35/bin/dbus-broker-launch --scope system --audit (code=exited, status=1/FAILURE)
   Main PID: 1349 (code=exited, status=1/FAILURE)
         IP: 0B in, 0B out
        CPU: 35ms

Notice: journal has been rotated since unit was started, output may be incomplete.

× dbus.socket - D-Bus System Message Bus Socket
     Loaded: loaded (/etc/systemd/system/dbus.socket; linked; preset: enabled)
     Active: failed (Result: service-start-limit-hit) since Wed 2023-12-06 21:38:40 CET; 4 months 2 days ago
   Duration: 1.847s
   Triggers: ● dbus-broker.service
     Listen: /run/dbus/system_bus_socket (Stream)

Notice: journal has been rotated since unit was started, output may be incomplete.
```

While in this state, I cannot switch configurations (and do some other things) because of `Failed to connect to bus: Connection refused`.

An obvious workaround is to set `services.dbus.implementation = "dbus";` but I'd rather not have to confront this problem again in the future when I've forgotten all about it.

[...]

Issue #297756, opened 2024-03-21
DynamicUser services can not access dbus since systemd 255


yu-re-ka:
there is an issue with DynamicUser services accessing dbus since systemd 255 was merged

old:

```
$ init --version
systemd 254 (254.6)
+PAM +AUDIT -SELINUX +APPARMOR +IMA +SMACK +SECCOMP +GCRYPT -GNUTLS +OPENSSL +ACL +BLKID +CURL +ELFUTILS +FIDO2 +IDN2 -IDN +IPTC +KMOD +LIBCRYPTSETUP +LIBFDISK +PCRE2 -PWQUALITY +P11KIT -QRENCODE +TPM2 +BZIP2 +LZ4 +XZ +ZLIB +ZSTD +BPF_FRAMEWORK -XKBCOMMON +UTMP -SYSVINIT default-hierarchy=unified

$ sudo systemd-run -p DynamicUser=1 --collect --pty systemctl is-system-running
Running as unit: run-u57.service
Press ^] three times within 1s to disconnect TTY.
degraded

$
```

new:

```
$ init --version
systemd 255 (255.2)
+PAM +AUDIT -SELINUX +APPARMOR +IMA +SMACK +SECCOMP +GCRYPT -GNUTLS +OPENSSL +ACL +BLKID +CURL +ELFUTILS +FIDO2 +IDN2 -IDN +IPTC +KMOD +LIBCRYPTSETUP +LIBFDISK +PCRE2 +PWQUALITY +P11KIT +QRENCODE +TPM2 +BZIP2 +LZ4 +XZ +ZLIB +ZSTD +BPF_FRAMEWORK -XKBCOMMON +UTMP -SYSVINIT default-hierarchy=unified

$ sudo systemd-run -p DynamicUser=1 --collect --pty systemctl is-system-running
Running as unit: run-u3393.service
Press ^] three times within 1s to disconnect TTY.
Failed to query system state: Transport endpoint is not connected
unknown

$ 
```

@NixOS/systemd

arianvp:
Could you run with log level set to debug?

r-vdp:
FWIW, this seems to work for me. I'm using dbus-broker, which might make a difference.

```sh
$ sudo systemd-run -E SYSTEMD_LOG_LEVEL=debug -p DynamicUser=1 --collect --pty systemctl is-system-running
Running as unit: run-u47.service
Press ^] three times within 1s to disconnect TTY.
[...]

PR #307731, opened 2024-04-29
Revert "dbus: Switch default implementation to broker"

flokli:
This reverts commit e827697fd3435b458a759ef0b92de7f26624b910.

Supersedes #305273, which is drafted and with merge conflicts.

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/nixos/modules/services/system/dbus.nix
+++ b/nixos/modules/services/system/dbus.nix
@@ -39,7 +39,7 @@ in
 
       implementation = mkOption {
         type = types.enum [ "dbus" "broker" ];
-        default = "broker";
+        default = "dbus";
         description = ''
           The implementation to use for the message bus defined by the D-Bus specification.
           Can be either the classic dbus daemon or dbus-broker, which aims to provide high
--- a/nixos/doc/manual/release-notes/rl-2405.section.md
+++ b/nixos/doc/manual/release-notes/rl-2405.section.md
@@ -44,8 +44,6 @@ Use `services.pipewire.extraConfig` or `services.pipewire.configPackages` for Pi
 - A new option `systemd.sysusers.enable` was added. If enabled, users and
   groups are created with systemd-sysusers instead of with a custom perl script.
 
-- The default dbus implementation has transitioned to dbus-broker from the classic dbus daemon for better performance and reliability. Users can revert to the classic dbus daemon by setting `services.dbus.implementation = "dbus";`. For detailed deviations, refer to [dbus-broker's deviations page](https://github.com/bus1/dbus-broker/wiki/Deviations).
-
 - `virtualisation.docker.enableNvidia` and `virtualisation.podman.enableNvidia` options are deprecated. `hardware.nvidia-container-toolkit.enable` should be used instead. This option will expose GPUs on containers with the `--device` CLI option. This is supported by Docker 25, Podman 3.2.0 and Singularity 4. Any container runtime that supports the CDI specification will take advantage of this feature.
 
 - `system.etc.overlay.enable` option was added. If enabled, `/etc` is


Aleksanaa:
Also the systemd team should be added to maintainers in dbus module later.

nixos/modules/services/system/dbus.nix
```diff
@@ -39,12 +39,14 @@ in
...
         type = types.enum [ "dbus" "broker" ];
-        default = "broker";
+        default = "dbus";
         description = ''
           The implementation to use for the message bus defined by the D-Bus specification.
           Can be either the classic dbus daemon or dbus-broker, which aims to provide high
           performance and reliability, while keeping compatibility to the D-Bus
           reference implementation.
+          dbus-broker currently has some issues on NixOS, which is why it's
+          considered experimental and not the default for now.
```
SuperSandro2000, line 49: What issues exactly? This vague comment isn't helpful for the reader.

I've been running this since months without any issues on all my systems, so it cannot be that experimental.

SuperSandro2000 (commented):
We need a motivation here. 

I am guessing https://github.com/NixOS/nixpkgs/issues/297756#issuecomment-2082402329 ?

But there exists a comment that the issue is fixed with dbus-broker. nsspath is also not properly updated and there is another issue, which we cannot reproduce and maybe had something to do with nsspath.

flokli:
The motivation is https://github.com/NixOS/nixpkgs/issues/302771, which is probably also the reason why @peterhoeg prepared the revert in https://github.com/NixOS/nixpkgs/pull/305273. @Aleksanaa also seems to be in agreement with that reasoning.

nixos/modules/services/system/dbus.nix
flokli: I can also entirely drop this line (it wasn't part of the revert per se). We can create a tracking issue to re-enable it past the next branchoff.

flokli: Done.

alyssais:
Please include motivation for the revert in the commit message.

flokli:
Done.

AndersonTorres (approved):

wegank (approved):

magneticflux-:
> I fully support the temporary reversal of this change. However, unstable users who rely on automatic updates and remote updates may be a little unlucky 😋

Hi, it's me, unlucky :sweat_smile:
Activating a new NixOS profile (via `deploy-rs`) freezes after stopping `dbus-broker.service` and I eventually get dropped into a blank terminal.

For future reference, setting `services.dbus.implementation = "broker";` lets me activate a profile. The `--boot` flag (to update the boot loader without activating a profile right now) in `deploy-rs` also works after a reboot.

flokli:
> > I fully support the temporary reversal of this change. However, unstable users who rely on automatic updates and remote updates may be a little unlucky 😋
> 
> Hi, it's me, unlucky 😅 Activating a new NixOS profile (via `deploy-rs`) freezes after stopping `dbus-broker.service` and I eventually get dropped into a blank terminal.

Did you encounter the same issue while the first switch (from dbus to dbus-broker) happened, too? Do you encounter it today, when switching from dbus to dbus-broker?

arianvp:
Both dbus and dbus-broker are unrestartable. Normally  our switch-to-configuration takes care of not restarting them. But that logic is missed if you swap out the backend. 


Switching into a configuration that switches from one to the other is not supported and can not be supported.   We can not make it work.  There isn't much debugging we can do as it's not a feature that dbus supports.

If you still have problems after rebooting let us now

SuperSandro2000:
So by reverting we are forcing everyone to reboot again, if I understand this correct?

SuperSandro2000:
> The motivation is #302771

That's just a single person and no one could reproduce the issue and we have no logs. If it really boils down to some impermenance issue, I would not put the burden to fix highly experimental setups onto nixos.

wegank:
> So by reverting we are forcing everyone to reboot again, if I understand this correct?

Yes. I guess there is no other way, since the systemd team has the final say on whether https://github.com/NixOS/nixpkgs/commit/e827697fd3435b458a759ef0b92de7f26624b910 will be in 24.05, and the proposal to [revert on release-24.05 instead of master](https://github.com/NixOS/nixpkgs/pull/305273#issuecomment-2067158830) was rejected.

flokli:
Well, there is an alias `dbus.service` in place, to theoretically prevent the exact symptoms here from happening, delaying the dbus implementation switch until the next reboot. Unfortunately our switching script didn't account for this and still restarted it...

This is all stuff that might have ended up being surfaced if this feature would have gotten some more review :shrug: 

In subsequent PRs we can ensure the switching experience gets less disruptive, to be ready when we change the default again, but IMHO a slight bit of pain for unstable users to do this revert now is better than shipping a not-so-much-reviewed feature to the upcoming short release.

SuperSandro2000:
So if I want to repropose the change, I would need to take a look into nixos-rebuild and exclude dbus-broker like it is already done for dbus? Maybe I find time going that.

flokli:
As I said, not opposed to the change in general, just feel like we should do a lot more testing and make sure the switch is not the fire hazard it is right now.

oxalica:
> > > I fully support the temporary reversal of this change. However, unstable users who rely on automatic updates and remote updates may be a little unlucky 😋
> > 
> > 
> > Hi, it's me, unlucky 😅 Activating a new NixOS profile (via `deploy-rs`) freezes after stopping `dbus-broker.service` and I eventually get dropped into a blank terminal.
> 
> Did you encounter the same issue while the first switch (from dbus to dbus-broker) happened, too? Do you encounter it today, when switching from dbus to dbus-broker?

I also got a blackscreen after hitting `switch` now. By login in a new tty and `tmux attach`, I can confirm the activation script tried to stop `dbus-broker.service`. A reboot can fix it, but with the cost of losing several minutes of data in GUI apps. I did not run into any issue when upgrading from dbus to dbus-broker, which was landed in nixos-unstable for quite a while. I don't remember if I did a `nixos-rebuild switch` or a `boot` for that upgrade. ~~Ironically, the reversion gives another chance for previously lucky ones to become unlucky, since the motivation of this PR is to fix upgrading failures, should we revert this PR again for the same reason?~~

Aleksanaa:
> since the motivation of this PR is to fix upgrading failures

No, their problems also appear after rebooting. Next time maybe we'll go with a `stateVersion`.

SuperSandro2000:
On a friends machine loginctl broke which made the nixos-rebuild switch already fail

```
dbus-daemon[2845527]: [system] Activating via systemd: service name='org.freedesktop.login1' unit='dbus-org.freedesktop.login1.service' requested by ':1.22' (uid=0 pid=2847559 comm="reboot" label="kernel")
dbus-daemon[2845527]: [system] Failed to activate service 'org.freedesktop.login1': timed out (service_start_timeout=25000ms)
```

SuperSandro2000:
I think we have a massive problem. Machines in the state switched from dbus-broker to dbus cannot be rebooted with `reboot` or `systemctl reboot` and need to be hard-reseted 💀

arianvp:
Yes that's bad. I'm happy we caught it now though? I'd be pretty miffed if rolling back from 24.05 to 23.11 would break rebooting without any mention of that in the release notes. 

Do we have a good understanding of why switching from dbus to dbus-broker doesn't cause issues but the other way around does? Did we skip restarting dbus in the one direction?

arianvp:
My suggestion would be:  we use the exact same unit file name (dbus.service) instead of using the aliasing.  IThat way we're sure nixos-rebuild `systemctl restart`s (or ignores) it instead of `systemctl stop dbus-broker; systemctl start dbus`.   

Maybe the alias stuff only works one direction?
So it saved us for dbus->dbus-broker  but not the other way around?

flokli:
I'd like to understand what's going on. Ideally, the activation script should understand what aliases are, so we don't need to deviate from how this is called on other distros to workaround bugs in our activation script.

As initially wrote, I propose we write a VM test with 2 specializations, testing switching back and forth, and ensure it keeps the old implementation running in both cases. That should allow testing out the behavior, ideally fixing the activation script.

K900:
Looking into it.

SuperSandro2000:
please review #309068

Lyndeno:
> > since the motivation of this PR is to fix upgrading failures
> 
> No, their problems also appear after rebooting. Next time maybe we'll go with a `stateVersion`.

> anything relying on polkit breaks

I was reverted back to dbus, polkit operations like pkexec and using polkit authentication with 1password do not work in Gnome, the authorization window does not pop up. Is this related?

I switched back to dbus-broker and am still experiencing the same issues

arianvp:
Most likely yes. Do your problems persist across reboot? If so then this is probably another issue.

Lyndeno:
> Most likely yes. Do your problems persist across reboot? If so then this is probably another issue.

Persists across reboot, I will see if #308884 fixes it. I see the same errors when running sudo or in the journal whenever polkit is attempted.

PR #308138, opened 2024-05-01
freebsd: Separate package set contents from construction

Ericson2314:
## Description of changes

~~This creates multiple package sets for different versions. The other versions don't yet work, but that will be fixed in subsequent PRs.~~

This refactors to create the beginnings of what could become support for multiple package sets / multiple versions of FreeBSD later.

CC @rhelmot @artemist. I am pushing from @rhelmot's repo (thanks for the perms!) so all 3 of us can collaborate.

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---


rhelmot (approved):

rhelmot:
I've taken a few passes through this and the only thing I can see that's even a little bit objectionable was some dead code, which I removed. Thanks!!

rhelmot:
Upon yet further inspection and noticing the ofborg failures I found that the addition of the "patch" version attr was half baked. Not all branches had this attribute, so I filled it in with 0 in the default case. You may wish to fix this in the update.py script instead. Moreover, to distinguish the various branches (release/14.0.0 vs releng/14.0) I added the branch disposition to the end of the version string (14.0.0-RELEASE vs 14.0.0-RELEASE-p6).

alyssais:
Can you explain why we want multiple versions?

Ericson2314:
> Upon yet further inspection and noticing the ofborg failures I found that the addition of the "patch" version attr was half baked. Not all branches had this attribute, so I filled it in with 0 in the default case.

Actually, I purposefully wanted not all versions to have it it. The problem was I did the `toString` *after* the non-`null` check.

Ericson2314:
I removed `branch` from the version for now, because I don't want any hashes in the default version to change. But we can add that back later. Also rebased now that the other PR is merged.

Ericson2314:
I also see the `RELEASE-p<patch>` thing @artemist tried to explain before, which I missed. I've updated the script to also extract the patch version from it.

`main` and the `stable/*` branches still lack a `patch` version, however.

Ericson2314:
> Can you explain why we want multiple versions?

@artemist and @rhelmot this one is for you.

I assume it is because you want to both support the last release, and work on the bleeding edge / dogfood things recently upstreamed. I am broadly deferential to that. We have multiple `linuxPackages` for the same reason.

But, all things equal, Nixpkgs does strive to have fewer not more versions of the same package. So there should be a clear answer here if we're going to do this :).

Ericson2314:
(I have one more nix code reworking to do. I want the splicing to take place in `versions.nix` so `package-set.nix` just returns a function and "doesn't know what its own name is.

rhelmot:
> Can you explain why we want multiple versions?

Several enterprise appliances (TrueNAS and PFSense come to mind) are based off of relatively shallow forks of FreeBSD. TrueNAS in particular is known for lagging a version behind the latest FreeBSD offering, so I wanted to not leave those users in the dust. The application for configuring a FreeBSD appliance with nix is, of course, being able to build app jails (not a feature we have right now but I've been slowly building up to it).

And, of course, I would like to be able to work on the bleeding edge for other purposes. My vision was that the only things we would 100% make sure worked would be `freebsd14` and `freebsd13` (pending the release of 15 of course), and `freebsd` would be the latest of these two.

I'm noticing suddenly the `patchesRoot` variable is missing. I'm not sure whether that was an explicit rejection or just considered to not be part of this PR - it is in fact essential for being able to support multiple versions, since different versions will need different sets of patches. I've just added it.

I was thinking a little bit about the feasibility of being able to override the entire splice with your own source information and `patchesRoot` so that you could theoretically work off of a fork if you're willing to play debugger for a while, and then hiding both `freebsdGit` and `freebsdBranches` from the consumer of nixpkgs. That might better signal our support goal, though I am genuinely clueless about how it might interact with John's beautiful splices.

Adjacent to this goal is the goal of moving the freebsd ports tree. I think if compatibility for freebsd-ports is at all present in nixpkgs, it should live outside the freebsd package set due to the difficulties in switching its version in tandem with the freebsd-base version.

rhelmot:
With the most recent commit I was in fact able to start building freebsd14 code on this branch, with only overriding the freebsd scope with versionData, source, and patchesRoot!

I think this opens the door for the more limited interface I described 30min ago that more accurately describes our support capabilities.

rhelmot:
I think the ofborg failure is due to it recursing infinitely into buildFreebsd. You can try fixing this with dontRecursIntoAttrs (?) or by simply removing buildFreebsd from the scope, though that would make it un-overridable. I have to log off for the day so good luck to whoever gets to this first :)

Ericson2314:
> `dontRecursIntoAttrs`

Good call, that is what `pkgs/top-level/stage.nix` does.

Ericson2314:
> That might better signal our support goal, though I am genuinely clueless about how it might interact with John's beautiful splices.

I think it should be fine.

If you are saying `pkgs/os-specific/bsd/freebsd/versions.nix` should define the whitelisted shorthands (like `freebsd13`, `freebsd14`, `freebsdGit`, etc.) that's fine. The only requirement is that `generateSplicesForMkScope` be given the attribute path to where the package set ends up being bound, whatever that may be.

rhelmot:
From inspecting the lib-tests code, the only way I could find to squash the recursion error (dontRecurseIntoAttrs was not working) was to add a manual exception. Not sure if this is intended.

rhelmot:
The PR title is now inaccurate, but I'm ready to merge if ofborg blesses us. we now only export `freebsd{13,14,}{Cross,}`.

pkgs/os-specific/bsd/freebsd/versions.nix
Ericson2314, line 1: Some of the indentation in here looks a little bit funky?

Ericson2314: Can also rename this to `default.nix`, signifying it's the "entry point" file for this directory.

pkgs/os-specific/bsd/freebsd/versions.nix
```diff
@@ -0,0 +1,56 @@
...
+      * `release/<major>.<minor>.0` for tagged releases without security updates
+      * `releng/<major>.<minor>` for release update branches with security updates
+      * `stable/<major>` for stable versions working towards the next minor release
+      * `main` for the latest development version
+
+      Set one with the NIXPKGS_FREEBSD_BRANCH environment variable or by setting `nixpkgs.config.freebsdBranch`.
+  '';
+  branches = callPackage ./default-branch.nix {};
+  exportedAttrSetsNative = lib.flip lib.mapAttrs branches (_: branch: freebsdBranches.${branch} or (badBranchError branch));
+  exportedAttrSetsCross = lib.flip lib.mapAttrs' branches (
```
Ericson2314, line 50: Actually the first ones are not necessarily "native". Rather the whole `*Cross` convention is a stupid relic that only exists because the weird bad GCC bootstrapping. Hopefully these new names, while cryptic, are at least accurate. 
```suggestion
  exportedAttrSetsRegular = lib.flip lib.mapAttrs branches (_: branch: freebsdBranches.${branch} or (badBranchError branch));
  exportedAttrSetsPreLibc = lib.flip lib.mapAttrs' branches (
```

Ericson2314:
@alyssais does the explanation of the motivation that @rhelmot gave sound good to you?

alyssais:
> We have multiple `linuxPackages` for the same reason.

This isn't totally analogous, because most packages (apart from kernel modules) don't care about linuxPackages.  (They might use linuxHeaders, but we don't provide multiple versions of that because it's backwards-compatible.)  But here, _all_ packages would have to be built separately for each FreeBSD release.

How often is building the "bleeding edge" something normal FreeBSD users are likely to want to do?  Wouldn't it generally be fine to build packages for the oldest version of FreeBSD we want to support, since we've established that FreeBSD is backwards compatible?

We don't provide any special mechanism to allow Linux developers to build against new versions of the kernel headers, but it's simple enough to do in an overlay without needing a global config option (which are generally difficult to work with in Nixpkgs due to us not being able to specify global defaults, among other things):

```nix
final: super: {
  linuxHeaders = linuxHeaders.overrideAttrs ({ ... }: {
    # …
  })
}
```

Could we make FreeBSD similar?

```nix
final: super: {
  freebsd = freebsd.override {
    version = "…";
    hashes = { /* … */ };
  };
}
```

rhelmot:
Okay, that's fine. Are you opposed to freebsd13 and freebsd14 to signal our support capabilities?

alyssais:
Can we start with just one version in Nixpkgs, and discuss expanding it separately once we have that version working well?

rhelmot:
I have performed the requested operation. Again, I don't know if I did it right with respect to splicing, but I'm a little more confident this time!

Ericson2314:
I fixed the formatting issue by running the RFC formatter on the new files.

Ericson2314:
Merged since this is no longer changing public interfaces.

sternenseemann:
Fix for eval regression https://github.com/NixOS/nixpkgs/pull/313110

PR #308271, opened 2024-05-01
vmagent: build from victoriametrics derivation

### LINKED ISSUE #306253: Merge request: vmagent → victoriametrics
## Issue description

Initially, I opened this issue to get _vmagent_ removed in favour of _victoriametrics_.
Instead, we propose that _vmagent_ be based on the _victoriametrics_ package.

---

## Original issue description for reference

The _[vmagent](https://github.com/NixOS/nixpkgs/blob/6143fc5eeb9c4f00163267708e26191d1e918932/pkgs/servers/monitoring/vmagent/default.nix)_ package provides a subset of the programs provided by the _[victoriametrics](https://github.com/NixOS/nixpkgs/blob/6143fc5eeb9c4f00163267708e26191d1e918932/pkgs/by-name/vi/victoriametrics/package.nix)_ package.
This package should be removed (and aliased to _victoriametrics_).

Shawn8901:
## Description of changes

fixes #306253 

Main goal is to avoid having 2 definitions for building for the same repository (see linked issue).

This PR migrates the building of the client tool vmagent to the victoriametrics  package.
This deduplicates the code base on our side as upstream is maintaining and building everything from the same code base (so its a mono repo with different subpackages).

Beside just to split out vmagent, i tried to build up all logical packages that make sense to me, to that people can override the package easily to optimize the derivation size if needed.

One has now
1. the timeseries server
2. vmagent as pushing service
3. alterting tools
4. vmauth as HTTP proxy in case one wants to use authentication for the timeseries server
5. the backup tools to backup the timeseries
6. vmctl which is used to migrate between timeseries (vm to vm, but also from other servers to vm)
7. logging server

A pure minimal server then could just enable the `withServer` and disable everything else.
I personally switched my VM server for testing to just use `withServer` and `withVmAuth` (and `withVmAgent` to be able to push metrics to itself), beside that every other thing is disabled as i am not using it and it seems to run fine.
The said package is ~76MiB large.
The full package victoriametrics is about 162 MiB. 
Vmagent by itself is be about 53 MiB (which we are migrating here and is about the same like the old).

I did check the functionality of the reduces package by looking at logs and verifying that my Grafana (with vm-datasource installed) is still able to query the data. 
Vmagent is deployed to all my machines, checked the log on those also.

As the vmagent package had different maintainers listed, i migrated everyone to the victoriametrics package.
Please let me know if you disagree with that action @NULLx76 @leona-ya

~~As the action should be transparent for user of the vmagent package, i dont see a requirement to create a release note here.~~
I added a release note entry as Victoria metrics no longer provides vmagent by default as suggested in the PR comments.

In addition, i also added myself as a maintainer.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [x] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---


leona-ya:
I'm not quite sure why `vmagent` needs to be included in the normal `victoriametrics` package. Does it needs it for basic functioning (i.e. without the vmagent module probably?)?

Shawn8901:
It could be excluded, due sending metrics to the server itself was at least how I did configure it and then you don't need two derivations to install (as one could set the package in vmagent module then), but if it's preferred I can change the default

SuperSandro2000:
If it was included before by accident, we might as well disabled it.

Shawn8901:
> If it was included before by accident, we might as well disabled it.

I am actually not sure if its by accident. I personally find it absolute useful as one might want to scrape the vmserver  as well.
But as it seems that others are not in favor of this use case (as its was pointed out that it could be an accident) and the freeze is over, i have pushed an update that removes vmagent from the victoriametrics package and added an entry for 24.11 release notes as it might break users setup otherwise.

Shawn8901:
did force push a migration to `pkgs/by-name` for vmagent.

Aleksanaa:
@ofborg vmagent

Aleksanaa:
@ofborg build vmagent

aanderse (approved):
i think the disabling of the agent isn't the best choice personally, but it really doesn't matter and is strictly an opinion based choice

---

as a user of this package (and the associated module) i approve these changes and will merge in the next day or two of no one disagrees 

thanks for pushing this through

Shawn8901:
> i think the disabling of the agent isn't the best choice personally, but it really doesn't matter and is strictly an opinion based choice

I am fine with both options, we have just to agree on one, initially I kept it enabled (as it's also my personal preference as pointed out in prior comments), I just switched as it was suggested.

@leona-ya , @SuperSandro2000 you both still prefer to exclude vmagent from the full VictoriaMectrics package, am I correct on this?

SuperSandro2000:
I don't really have a strong opinion right now. Can you two decide this?

aanderse:
it is very normal for stat gathering and monitoring tools to keep information on themselves 🤷‍♂️

Shawn8901:
As I think the same (see previous comments) that it's a quite normal use case that a monitoring system also monitors itself I would flit the default in the next days unless @leona-ya reports a strong opinion against it.

Shawn8901:
Pushed a change that flips the `vmagent` args default on the package again to `true` and removed the release-note entry.

Also rebased on top of recent master.

Shawn8901:
@ofborg build vmagent victoriametrics

PR #308291, opened 2024-05-01
nixos/nextcloud: add nextcloud-update-db.service, nextcloud-cron isn't oneshot

Ma27:
## Description of changes

__Note:__ before merging this, a bit of testing would be appreciated.

This service performs operations that significantly increase the performance of Nextcloud, can take a while. These are designed however to not require maintenance mode and can be executed during normal operation[1].

Make nextcloud-cron a simple unit instead of oneshot: otherwise we risk that it'll be stopped by the startup timeout (oneshot executes ExecStart while "activating") which can be an issue for very long running tasks or if Nextcloud needs to catch up if one task was broken for a while.

[1] https://docs.nextcloud.com/server/29/admin_manual/maintenance/upgrade.html#long-running-migration-steps

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [x] Tested, as applicable:

---

### DIFF
--- a/nixos/modules/services/web-apps/nextcloud.nix
+++ b/nixos/modules/services/web-apps/nextcloud.nix
@@ -939,6 +939,7 @@ in {
 
         in {
           wantedBy = [ "multi-user.target" ];
+          wants = [ "nextcloud-update-db.service" ];
           before = [ "phpfpm-nextcloud.service" ];
           after = optional mysqlLocal "mysql.service" ++ optional pgsqlLocal "postgresql.service";
           requires = optional mysqlLocal "mysql.service" ++ optional pgsqlLocal "postgresql.service";
@@ -997,7 +998,7 @@ in {
           after = [ "nextcloud-setup.service" ];
           environment.NEXTCLOUD_CONFIG_DIR = "${datadir}/config";
           serviceConfig = {
-            Type = "oneshot";
+            Type = "exec";
             User = "nextcloud";
             ExecCondition = "${lib.getExe phpPackage} -f ${webroot}/occ status -e";
             ExecStart = "${lib.getExe phpPackage} -f ${webroot}/cron.php";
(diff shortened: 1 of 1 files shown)


nixos/modules/services/web-apps/nextcloud.nix
```diff
@@ -997,7 +998,6 @@ in {
           after = [ "nextcloud-setup.service" ];
           environment.NEXTCLOUD_CONFIG_DIR = "${datadir}/config";
           serviceConfig = {
-            Type = "oneshot";
```
Tom-Hubrecht, line 1000: Shouldn't the type be set to `exec` instead of `simple` ?

>The exec type is similar to simple, but the service manager will consider the unit started immediately after the main service binary has been executed. The service manager will delay starting of follow-up units until that point. (Or in other words: simple proceeds with further jobs right after fork() returns, while exec will not proceed before both fork() and execve() in the service process succeeded.) Note that this means systemctl start command lines for exec services will report failure when the service's binary cannot be invoked successfully (for example because the selected User= doesn't exist, or the service binary is missing).

Ma27: Not sure if that makes that much of a difference, but done :)

SuperSandro2000:
I can't fully test this, as I already have executed those occ commands some time ago and also my instance is rather small and "long running maintenance tasks" usually take less than a minute unless some recognize processing is happening.

onny (approved):
Really cool improvement! Works on my Nextcloud instance

SuperSandro2000 (approved):
Deployed it to my instance and worked

Issue #308247, opened 2024-05-01
Openssh password reset broken


wlcx:
### Describe the bug
When logging in over ssh using key-based auth to a user account with an expired passwd, openssh cannot access `/usr/bin/passwd` (because it doesn't exist - nixos keeps it at `/run/wrappers/bin/passwd`) to allow the user to change their password, and the session terminates.
Presumably this would be the same for password-based authentication too!

### Steps To Reproduce
1. Expire a user's password `# passwd -e someuser`
2. try to login via SSH with an existing key `ssh someuser@nixos`
3. observe the error:
```WARNING: Your password has expired.
You must change your password now and login again!
passwd: No such file or directory
Connection to nixos closed.
```

### Expected behavior
Opensshd should be able to get at `passwd` and allow the user to change their password.

### Additional context
You can verify that this is the issue by symlinking `/run/wrappers/bin/passwd` to `/usr/bin/passwd` and retrying the above, which will now work.

### Metadata
Please run `nix-shell -p nix-info --run "nix-info -m"` and paste the result.

```console
[user@system:~]$ nix-shell -p nix-info --run "nix-info -m"
 - system: `"x86_64-linux"`
 - host os: `Linux 6.1.87, NixOS, 23.11 (Tapir), 23.11.20240421.a5e4bbc`
 - multi-user?: `yes`
 - sandbox: `yes`
 - version: `nix-env (Nix) 2.18.1`
 - nixpkgs: `/nix/var/nix/profiles/per-user/root/channels/nixos`

```

---

Add a :+1: [reaction] to [issues you find important].

[reaction]: https://github.blog/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/
[issues you find important]: https://github.com/NixOS/nixpkgs/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions-%2B1-desc

wlcx:
Fixed by #308325

PR #308325, opened 2024-05-01
openssh: build with correct passwd path

wlcx:
## Description of changes
Tentative fix for #308247

Points openssh at nixos' `passwd` so it can allow users to change expired passwords on login

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/tools/networking/openssh/common.nix
+++ b/pkgs/tools/networking/openssh/common.nix
@@ -32,6 +32,7 @@
 , withPAM ? stdenv.hostPlatform.isLinux
 , dsaKeysSupport ? false
 , linkOpenssl ? true
+, isNixos ? stdenv.hostPlatform.isLinux
 }:
 
 stdenv.mkDerivation {
@@ -74,6 +75,12 @@ stdenv.mkDerivation {
     # to use: `configure' wants `gcc', but `make' wants `ld'.
     unset LD
   '';
+
+  env = lib.optionalAttrs isNixos {
+    # openssh calls passwd to allow the user to reset an expired password, but nixos
+    # doesn't ship it at /usr/bin/passwd.
+    PATH_PASSWD_PROG = "/run/wrappers/bin/passwd";
+  };
 
   # I set --disable-strip because later we strip anyway. And it fails to strip
   # properly when cross building.


eclairevoyant:
As this is a change to openssh which requires massive rebuilds, the PR should go to staging.
Please ensure to _rebase_ (do not simply change the branch target!) as per the contribution guidelines here: https://github.com/NixOS/nixpkgs/blob/master/CONTRIBUTING.md#rebasing-between-branches-ie-from-master-to-staging

Marking as draft in the meantime.

wlcx:
Just realised I did exactly what you said not to do. Will sort shortly.

SuperSandro2000:
> As this is a change to openssh which requires massive rebuilds

That was true in the past but we are now below 500 rebuilds which is generally fine to go master. The general rule of thumb is, that when the rebuild counts are red, it needs to to go to staging.

wlcx:
Ok... so back against master then?

pkgs/tools/networking/openssh/common.nix
```diff
@@ -96,6 +96,10 @@ stdenv.mkDerivation {
 
   ${if stdenv.hostPlatform.isStatic then "NIX_LDFLAGS" else null}= [ "-laudit" ] ++ lib.optionals withKerberos [ "-lkeyutils" ];
 
+  # openssh calls passwd to allow the user to reset an expired password, but nixos
+  # doesn't ship it at /usr/bin/passwd.
+  PATH_PASSWD_PROG = "/run/wrappers/bin/passwd";
```
LeSuisse, line 101: This should probably be only set when `stdenv.isLinux` (even if it is unlikely that macOS users are going to run nixpkgs sshd :) )

wlcx: Ah, I was wondering about that. Is there any way to determine that we're on nixos specifically, not just Linux? Again - unlikely that a non-nixos Linux user would be running openssh; but...

dasJ: Unforunately no, there is also a patch in glibc iirc that just goes with "every linux is nixos"

eclairevoyant: Most straightforward way I can think of would be adding an overridable arg to the top of the file, e.g. `isNixOS ? true`

wlcx: How about `isNixOS ? false` and then override it to true in `nixos/modules/programs/ssh.nix`?

eclairevoyant: Then it wouldn't be cached by hydra. IMO the priority should be to cache the version of openssh used by NixOS users.

wlcx: yeah, fair enough

pkgs/tools/networking/openssh/common.nix
```diff
@@ -182,4 +183,9 @@ stdenv.mkDerivation {
     maintainers = (extraMeta.maintainers or []) ++ (with maintainers; [ eelco aneeshusa ]);
     mainProgram = "ssh";
   } // extraMeta;
-}
+} // lib.optionalAttrs isNixos {
+  # openssh calls passwd to allow the user to reset an expired password, but nixos
+  # doesn't ship it at /usr/bin/passwd.
+  PATH_PASSWD_PROG = "/run/wrappers/bin/passwd";
```
SuperSandro2000, line 189: I always find it a bit cleaner to not use optionalAttrs for such things. We can probably place this inside of env and use optionalAttrs there. What do you think?

wlcx: Yeah, that's much nicer. Wasn't aware of `env`!

PR #155056, opened 2022-01-14
notmuch: install ruby bindings

arcnmx:
###### Motivation for this change

Previously, the ruby bindings couldn't be used because they weren't actually installed into `$out`

###### Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
- [ ] For non-Linux: Is `sandbox = true` set in `nix.conf`? (See [Nix manual](https://nixos.org/manual/nix/stable/command-ref/conf-file.html))
- [ ] Tested, as applicable:
  - [NixOS test(s)](https://nixos.org/manual/nixos/unstable/index.html#sec-nixos-tests) (look inside [nixos/tests](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests))
  - and/or [package tests](https://nixos.org/manual/nixpkgs/unstable/#sec-package-tests)
  - or, for functions and "core" functionality, tests in [lib/tests](https://github.com/NixOS/nixpkgs/blob/master/lib/tests) or [pkgs/test](https://github.com/NixOS/nixpkgs/blob/master/pkgs/test)
  - made sure NixOS tests are [linked](https://nixos.org/manual/nixpkgs/unstable/#ssec-nixos-tests-linking) to the relevant packages
- [ ] Tested compilation of all packages that depend on this change using `nix-shell -p nixpkgs-review --run "nixpkgs-review rev HEAD"`. Note: all changes have to be committed, also see [nixpkgs-review usage](https://github.com/Mic92/nixpkgs-review#usage)
- [x] Tested basic functionality of all binary files (usually in `./result/bin/`)
- [22.05 Release Notes (or backporting 21.11 Release notes)](https://github.com/NixOS/nixpkgs/blob/master/CONTRIBUTING.md#generating-2205-release-notes)
  - [ ] (Package updates) Added a release notes entry if the change is major or breaking
  - [ ] (Module updates) Added a release notes entry if the change is significant
  - [ ] (Module addition) Added a release notes entry if adding a new NixOS module
  - [ ] (Release notes changes) Ran `nixos/doc/manual/md-to-db.sh` to update generated release notes
- [ ] Fits [CONTRIBUTING.md](https://github.com/NixOS/nixpkgs/blob/master/CONTRIBUTING.md).

PR #308348, opened 2024-05-01
notmuch: install the vim plugin

RyanGibb:
Seems I messed up squashing https://github.com/NixOS/nixpkgs/pull/308318

> Changes look good, but I haven't tested them. In any case, could you squash the commits please into 1 notmuch: install the vim plugin?

That's fixed now @doronbehar.

I'm still thinking about how to pick up the Ruby bindings. It would make it easier if they were in the same output. It doesn't look like there were any strong reasons against this https://github.com/NixOS/nixpkgs/pull/155056.

### DIFF
--- a/pkgs/applications/networking/mailreaders/notmuch/default.nix
+++ b/pkgs/applications/networking/mailreaders/notmuch/default.nix
@@ -15,12 +15,12 @@
 , withVim ? true
 }:
 
-stdenv.mkDerivation rec {
+stdenv.mkDerivation (finalAttrs: {
   pname = "notmuch";
   version = "0.38.3";
 
   src = fetchurl {
-    url = "https://notmuchmail.org/releases/notmuch-${version}.tar.xz";
+    url = "https://notmuchmail.org/releases/notmuch-${finalAttrs.version}.tar.xz";
     hash = "sha256-mvRsyA2li0MByiuu/MJaQNES0DFVB+YywPPw8IMo0FQ=";
   };
 
@@ -141,9 +141,8 @@ stdenv.mkDerivation rec {
   '';
 
   passthru = {
-    pythonSourceRoot = "notmuch-${version}/bindings/python";
+    pythonSourceRoot = "notmuch-${finalAttrs.version}/bindings/python";
     tests.version = testers.testVersion { package = notmuch; };
-    inherit version;
 
     updateScript = gitUpdater {
       url = "https://git.notmuchmail.org/git/notmuch";
@@ -160,4 +159,4 @@ stdenv.mkDerivation rec {
     platforms   = platforms.unix;
     mainProgram = "notmuch";
   };
-}
+})
--- a/pkgs/applications/networking/mailreaders/notmuch/default.nix
+++ b/pkgs/applications/networking/mailreaders/notmuch/default.nix
@@ -1,4 +1,4 @@
-{ fetchurl, lib, stdenv, makeWrapper
+{ fetchurl, lib, stdenv, makeWrapper, buildEnv
 , pkg-config, gnupg
 , xapian, gmime3, sfsexp, talloc, zlib
 , doxygen, perl, texinfo
@@ -77,8 +77,7 @@ stdenv.mkDerivation (finalAttrs: {
   '';
 
   outputs = [ "out" "man" "info" "bindingconfig" ]
-    ++ lib.optional withEmacs "emacs"
-    ++ lib.optional withRuby "ruby";
+    ++ lib.optional withEmacs "emacs";
 
   # if notmuch is built with s-expression support, the testsuite (T-850.sh) only
   # passes if notmuch-git can be executed, so we need to patch its shebang.
@@ -124,7 +123,7 @@ stdenv.mkDerivation (finalAttrs: {
     moveToOutput bin/notmuch-emacs-mua $emacs
   '' + lib.optionalString withRuby ''
     make -C bindings/ruby install \
-      vendordir=$ruby/lib/ruby \
+      vendordir=$out/lib/ruby \
       SHELL=$SHELL \
       $makeFlags "''${makeFlagsArray[@]}" \
       $installFlags "''${installFlagsArray[@]}"
@@ -138,10 +137,26 @@ stdenv.mkDerivation (finalAttrs: {
     make -C vim DESTDIR="$out/share/vim-plugins/notmuch" prefix="" install
     mkdir -p $out/share/nvim
     ln -s $out/share/vim-plugins/notmuch $out/share/nvim/site
+  '' + lib.optionalString (withVim && withRuby) ''
+    PLUG=$out/share/vim-plugins/notmuch/plugin/notmuch.vim
+    cat >> $PLUG << EOF
+      let \$GEM_PATH=\$GEM_PATH . ":${finalAttrs.passthru.gemEnv}/${ruby.gemPath}"
+      let \$RUBYLIB=\$RUBYLIB . ":$out/${ruby.libPath}/${ruby.system}"
+      if has('nvim')
+    EOF
+    for gem in ${finalAttrs.passthru.gemEnv}/${ruby.gemPath}/gems/*/lib; do
+      echo "ruby \$LOAD_PATH.unshift('$gem')" >> $PLUG
+    done
+    echo 'endif' >> $PLUG
   '';
 
   passthru = {
     pythonSourceRoot = "notmuch-${finalAttrs.version}/bindings/python";
+    gemEnv = buildEnv {
+      name = "notmuch-vim-gems";
+      paths = with ruby.gems; [ mail ];
+      pathsToLink = [ "/lib" "/nix-support" ];
+    };
     tests.version = testers.testVersion { package = notmuch; };
 
     updateScript = gitUpdater {
--- a/pkgs/applications/networking/mailreaders/notmuch/default.nix
+++ b/pkgs/applications/networking/mailreaders/notmuch/default.nix
@@ -12,6 +12,7 @@
 , withEmacs ? true
 , withRuby ? true
 , withSfsexp ? true # also installs notmuch-git, which requires sexp-support
+, withVim ? true
 }:
 
 stdenv.mkDerivation rec {
@@ -133,6 +134,10 @@ stdenv.mkDerivation rec {
   + lib.optionalString withSfsexp ''
     cp notmuch-git $out/bin/notmuch-git
     wrapProgram $out/bin/notmuch-git --prefix PATH : $out/bin:${lib.getBin git}/bin
+  '' + lib.optionalString withVim ''
+    make -C vim DESTDIR="$out/share/vim-plugins/notmuch" prefix="" install
+    mkdir -p $out/share/nvim
+    ln -s $out/share/vim-plugins/notmuch $out/share/nvim/site
   '';
 
   passthru = {


RyanGibb:
Okay, I've moved the Ruby bindings to `$out` along with the vim plugins. I've also added `rubyPackages.mail` to `RUBYLIBS`. This works!

There's two issues I'm having with separate outputs: I'm not picking up vim plugins from the separate vim output for whatever reason, but more importantly Nix complains about cyclic dependencies if we reference the ruby output from the vim output, so we'd require duplicating them.

pkgs/applications/networking/mailreaders/notmuch/default.nix
```diff
@@ -133,6 +134,14 @@ stdenv.mkDerivation rec {
   + lib.optionalString withSfsexp ''
     cp notmuch-git $out/bin/notmuch-git
     wrapProgram $out/bin/notmuch-git --prefix PATH : $out/bin:${lib.getBin git}/bin
+  '' + lib.optionalString withVim ''
+    make -C vim DESTDIR="$out/share/vim-plugins/notmuch" prefix="" install
+    mkdir -p $out/share/nvim
+    ln -s $out/share/vim-plugins/notmuch $out/share/nvim/site
+  '' + lib.optionalString (withVim && withRuby) ''
+    cat >> $out/share/vim-plugins/notmuch/plugin/notmuch.vim << EOF
+      let \$RUBYLIB=\$RUBYLIB . ":$out/${ruby.libPath}/${ruby.system}:${rubyPackages.mail}/lib/ruby/gems/${ruby.version.libDir}/gems/mail-${rubyPackages.mail.version}/lib"
```
doronbehar, line 143: Hmm I'm surprised there is no shortcut for:

```nix
${rubyPackages.mail}/lib/ruby/gems/${ruby.version.libDir}/gems/mail-${rubyPackages.mail.version}/lib"
```

The python equivalent would have been simply:

```nix
${pythonPackages.mail}/${pythonPackages.python.sitePackages}
```

It would have been nice if `ruby.gems` had provided such an equivalent, but according to my poor `git grep` searches it doesn't seem like a common thing to do - to set `RUBYLIB` to such a search path of both ruby and a few gems.

Either way, it'd be nice to be able to define this variable a bit more abstractly, and generally in the `passthru` like this (untested):

```nix
passthru = {
  # ...
  # Perhaps this should be part of the `ruby.passthru`, or an attribute
  # for all `ruby.gems`.
  getGemRubyLibPath = gem: ${gem.ruby.gemPath}/gems/${gem.gemName}-${gem.version}/lib";
  rubyLibPath = lib.concatStringsSep ":" [
    "$out/${ruby.libPath}/${ruby.system}"
  ] ++ lib.map finalAttrs.passthru.getGemRubyLibPath [
    ruby.gems.mail
  ];
};
```

Note how `getGemRubyLibPath` does what you wanted to do much more elegantly. With the above, you can access the `RUBYLIB` path you want with:

```suggestion
      let \$RUBYLIB=\$RUBYLIB . ":${finalAttrs.passthru.rubyLibPath}"
```

You'd also have to transition to `finalAttrs` (like [this commit for instance](https://github.com/NixOS/nixpkgs/commit/c1f5835e6ba8b2df709d51fa0e81b1f51fd7105c)) in a separate commit, before this commit.

doronbehar:
> There's two issues I'm having with separate outputs: I'm not picking up vim plugins from the separate vim output for whatever reason, but more importantly Nix complains about cyclic dependencies if we reference the ruby output from the vim output, so we'd require duplicating them.

I'm not sure I understood this, but I trust you you made an effort and reached the correct conclusion. Removing the `ruby` output is definitely shouldn't be a big deal.

RyanGibb:
> I'm not sure I understood this, but I trust you you made an effort and reached the correct conclusion. Removing the ruby output is definitely shouldn't be a big deal.

Re-reading this I don't think I made much sense actually. We wouldn't *require* deduplicating them, we'd just be duplicating the ruby bindings across both outputs. But yeah, that sounds good.

pkgs/applications/networking/mailreaders/notmuch/default.nix
RyanGibb: Thanks for the suggestions!

> You'd also have to transition to finalAttrs (like [this commit for instance](https://github.com/NixOS/nixpkgs/commit/c1f5835e6ba8b2df709d51fa0e81b1f51fd7105c)) in a separate commit, before this commit.

Can I ask why does this need to be in a separate commit?

doronbehar:
> Can I ask why does this need to be in a separate commit?

IMO it makes the changes to the expression laid out better, and easier to review, both in the far and the near future. It really is a change independent from the other changes.

RyanGibb:
It turns out there was issues with the mail gem, as we weren't setting RUBYLIB for it's dependencies. I've squash pushed a fix adapted from https://github.com/arcnmx/nixexprs/blob/master/pkgs/vimPlugins.nix that seems to work.

pkgs/applications/networking/mailreaders/notmuch/default.nix
```diff
@@ -6,12 +6,14 @@
 , pythonPackages
 , emacs
 , ruby
+, rubyPackages
```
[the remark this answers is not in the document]
RyanGibb: Done!

pkgs/applications/networking/mailreaders/notmuch/default.nix
RyanGibb: Okay, I've changed the approach a bit to use `buildEnv`.

pkgs/applications/networking/mailreaders/notmuch/default.nix
```diff
@@ -133,7 +133,28 @@ stdenv.mkDerivation rec {
   + lib.optionalString withSfsexp ''
     cp notmuch-git $out/bin/notmuch-git
     wrapProgram $out/bin/notmuch-git --prefix PATH : $out/bin:${lib.getBin git}/bin
-  '';
+  '' + lib.optionalString withVim ''
+    make -C vim DESTDIR="$out/share/vim-plugins/notmuch" prefix="" install
+    mkdir -p $out/share/nvim
+    ln -s $out/share/vim-plugins/notmuch $out/share/nvim/site
+  '' + lib.optionalString (withVim && withRuby) (let
+    gemEnv = buildEnv {
```
doronbehar, line 141: Interesting! Now I'm surprised how `ruby.withPackages` doesn't fit here, because it doesn't add `"/lib"` to `pathsToLink` (found this out locally).

In anycase, it'd still be great if you'd transition into `finalAttrs`, as it would allow someone to override this derivation if they need from some reason more gems installed.

Here are changes I tried, based upon yours:

```diff
diff --git i/pkgs/applications/networking/mailreaders/notmuch/default.nix w/pkgs/applications/networking/mailreaders/notmuch/default.nix
index c9b0e158acf2..8348919d1769 100644
--- i/pkgs/applications/networking/mailreaders/notmuch/default.nix
+++ w/pkgs/applications/networking/mailreaders/notmuch/default.nix
@@ -15,12 +15,12 @@
 , withVim ? true
 }:
 
-stdenv.mkDerivation rec {
+stdenv.mkDerivation (finalAttrs: {
   pname = "notmuch";
   version = "0.38.3";
 
   src = fetchurl {
-    url = "https://notmuchmail.org/releases/notmuch-${version}.tar.xz";
+    url = "https://notmuchmail.org/releases/notmuch-${finalAttrs.version}.tar.xz";
     hash = "sha256-mvRsyA2li0MByiuu/MJaQNES0DFVB+YywPPw8IMo0FQ=";
   };
 
@@ -137,29 +137,27 @@ stdenv.mkDerivation rec {
     make -C vim DESTDIR="$out/share/vim-plugins/notmuch" prefix="" install
     mkdir -p $out/share/nvim
     ln -s $out/share/vim-plugins/notmuch $out/share/nvim/site
-  '' + lib.optionalString (withVim && withRuby) (let
-    gemEnv = buildEnv {
-      name = "notmuch-vim-gems";
-      paths = with ruby.gems; [ mail ];
-      pathsToLink = [ "/lib" "/nix-support" ];
-    };
-  in ''
+  '' + lib.optionalString (withVim && withRuby) ''
     PLUG=$out/share/vim-plugins/notmuch/plugin/notmuch.vim
     cat >> $PLUG << EOF
-      let \$GEM_PATH=\$GEM_PATH . ":${gemEnv}/${ruby.gemPath}"
+      let \$GEM_PATH=\$GEM_PATH . ":${finalAttrs.passthru.gemEnv}/${ruby.gemPath}"
       let \$RUBYLIB=\$RUBYLIB . ":$out/${ruby.libPath}/${ruby.system}"
       if has('nvim')
     EOF
-    for gem in ${gemEnv}/${ruby.gemPath}/gems/*/lib; do
+    for gem in ${finalAttrs.passthru.gemEnv}/${ruby.gemPath}/gems/*/lib; do
       echo "ruby \$LOAD_PATH.unshift('$gem')" >> $PLUG
     done
     echo 'endif' >> $PLUG
-  '');
+  '';
 
   passthru = {
-    pythonSourceRoot = "notmuch-${version}/bindings/python";
+    pythonSourceRoot = "notmuch-${finalAttrs.version}/bindings/python";
     tests.version = testers.testVersion { package = notmuch; };
-    inherit version;
+    gemEnv = buildEnv {
+      name = "notmuch-vim-gems";
+      paths = with ruby.gems; [ mail ];
+      pathsToLink = [ "/lib" "/nix-support" ];
+    };
 
     updateScript = gitUpdater {
       url = "https://git.notmuchmail.org/git/notmuch";
@@ -176,4 +174,4 @@ stdenv.mkDerivation rec {
     platforms   = platforms.unix;
     mainProgram = "notmuch";
   };
-}
+})
```

And here is how I tested the overriding mechanism:

Without overriding:

```
$ nix show-derivation --impure --expr 'with (builtins.getFlake "'$PWD'").legacyPackages.${builtins.currentSystem}; notmuch.overrideAttrs(old: {
})' | jq --raw-output '.[].env.postInstall'
moveToOutput bin/notmuch-emacs-mua $emacs
make -C bindings/ruby install \
  vendordir=$out/lib/ruby \
  SHELL=$SHELL \
  $makeFlags "${makeFlagsArray[@]}" \
  $installFlags "${installFlagsArray[@]}"
cp notmuch-git $out/bin/notmuch-git
wrapProgram $out/bin/notmuch-git --prefix PATH : $out/bin:/nix/store/apl2ibxqrxbbj6z79nfgv892999z1zyv-git-2.44.0/bin
make -C vim DESTDIR="$out/share/vim-plugins/notmuch" prefix="" install
mkdir -p $out/share/nvim
ln -s $out/share/vim-plugins/notmuch $out/share/nvim/site
PLUG=$out/share/vim-plugins/notmuch/plugin/notmuch.vim
cat >> $PLUG << EOF
  let \$GEM_PATH=\$GEM_PATH . ":/nix/store/a4s4ymn7k90yh9bld5zic587n1kg0k1v-notmuch-vim-gems/lib/ruby/gems/3.1.0"
  let \$RUBYLIB=\$RUBYLIB . ":$out/lib/ruby/3.1.0/x86_64-linux"
  if has('nvim')
EOF
for gem in /nix/store/a4s4ymn7k90yh9bld5zic587n1kg0k1v-notmuch-vim-gems/lib/ruby/gems/3.1.0/gems/*/lib; do
  echo "ruby \$LOAD_PATH.unshift('$gem')" >> $PLUG
done
echo 'endif' >> $PLUG
```

With overriding:

```
$ nix show-derivation --impure --expr 'with (builtins.getFlake "'$PWD'").legacyPackages.${builtins.currentSystem}; notmuch.overrideAttrs(old: {
  passthru = old.passthru // {
    gemEnv = buildEnv {
      name = "notmuch-more-gems";
      paths = with ruby.gems; [
        mail
        clamp
      ];
    };
  };
})' | jq --raw-output '.[].env.postInstall'
moveToOutput bin/notmuch-emacs-mua $emacs
make -C bindings/ruby install \
  vendordir=$out/lib/ruby \
  SHELL=$SHELL \
  $makeFlags "${makeFlagsArray[@]}" \
  $installFlags "${installFlagsArray[@]}"
cp notmuch-git $out/bin/notmuch-git
wrapProgram $out/bin/notmuch-git --prefix PATH : $out/bin:/nix/store/apl2ibxqrxbbj6z79nfgv892999z1zyv-git-2.44.0/bin
make -C vim DESTDIR="$out/share/vim-plugins/notmuch" prefix="" install
mkdir -p $out/share/nvim
ln -s $out/share/vim-plugins/notmuch $out/share/nvim/site
PLUG=$out/share/vim-plugins/notmuch/plugin/notmuch.vim
cat >> $PLUG << EOF
  let \$GEM_PATH=\$GEM_PATH . ":/nix/store/hadibs48n5cympz47k3ypag89yvbd6x1-notmuch-more-gems/lib/ruby/gems/3.1.0"
  let \$RUBYLIB=\$RUBYLIB . ":$out/lib/ruby/3.1.0/x86_64-linux"
  if has('nvim')
EOF
for gem in /nix/store/hadibs48n5cympz47k3ypag89yvbd6x1-notmuch-more-gems/lib/ruby/gems/3.1.0/gems/*/lib; do
  echo "ruby \$LOAD_PATH.unshift('$gem')" >> $PLUG
done
echo 'endif' >> $PLUG
```

Note how the path to the `gemEnv` changed - as requested. It won't be possible to perform such a change so elegantly without `finalAttrs`.

doronbehar (changes requested):
P.S it would be nice to explain in the commit message why the `ruby` output was discarded.

pkgs/applications/networking/mailreaders/notmuch/default.nix
RyanGibb: Okay, makes sense!

RyanGibb:
I've transitioned to finalAttrs and added a commit message to explain why the ruby output was discarded.

doronbehar (changes requested):
@RyanGibb it looks better overall, but it doesn't make sense to put the `finalAttrs` commit at the end. It makes much more sense to put it before the `withRuby && withVim` commit, as it makes the overall complexity of the diffs much cleaner.

Also, the explanation in the commit message about the removal of the `ruby` output is not phrased well. I'd write it as follows (format it of course to Git commit messages' 80 characters format:

> We removed the ruby output so we will be able to reference the ruby files from the vim plugin, which is in `$out`. If we try to reference $ruby from $out we get a cyclic dependency.

No need to mention the idea of duplicating `$ruby` and `$out`, if you don't explain why you don't do that :).

doronbehar:
BTW, how can one test this new feature? What exactly I need to run in Vim / Neovim?

doronbehar:
Also, naturally if the `finalAttrs` commit will be 2nd, it would simply be:

```
notmuch: use finalAttrs
```

RyanGibb:
> @RyanGibb it looks better overall, but it doesn't make sense to put the `finalAttrs` commit at the end. It makes much more sense to put it before the `withRuby && withVim` commit, as it makes the overall complexity of the diffs much cleaner.

Sure

> No need to mention the idea of duplicating $ruby and $out, if you don't explain why you don't do that :).

But I did explain that we don't do that to avoid duplicating the bindings? I'm not too bothered about this though

> Also, naturally if the finalAttrs commit will be 2nd, it would simply be:

I've rebased and changed the order

> BTW, how can one test this new feature? What exactly I need to run in Vim / Neovim?

Try `vim -c NotMuch`. It might not show anything much without a notmuch index though.

RyanGibb:
Agree! Thanks for the suggestion :-)

doronbehar:
> Agreed! Thanks for the suggestion :-)

Great :) (I'm waiting for you to `git push`).

RyanGibb:
Whoops, looks like I pushed the wrong commit before! I've pushed the right one now

pkgs/applications/networking/mailreaders/notmuch/default.nix
doronbehar: But @RyanGibb you didn't define `gemEnv` in `passthru`, and then used `finalAttrs.passthru`...

doronbehar (approved):
Works great!

doronbehar:
From some reason I didn't get a notification for the last force push, sorry it took a while :). Thanks for your cooperation and for insisting on contributing this!

RyanGibb:
No worries! Thanks for the help reviewing @doronbehar :-)

PR #308366, opened 2024-05-01
darwin.stdenv: propagate sdkRoot via extraBuildInputs

reckenrode:
## Description of changes

This ensures it is properly dropped when cross-compiling.

I tested by building the Darwin bootstrap on aarch64-darwin and building psutil in the bootstrap to confirm the sdk root is being propagated correctly. It should have an SDK version of 11.0 (and did).

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [x] x86_64-darwin
  - [x] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/os-specific/darwin/apple-sdk/sdkRoot.nix
+++ b/pkgs/os-specific/darwin/apple-sdk/sdkRoot.nix
@@ -41,18 +41,14 @@ runCommand "sdkroot-${sdkVersion}" { } ''
 
   install -D '${../../../build-support/setup-hooks/role.bash}' "$out/nix-support/setup-hook"
   cat >> "$out/nix-support/setup-hook" <<-hook
-  #
-  # See comments in cc-wrapper's setup hook. This works exactly the same way.
-  #
-  [[ -z \''${strictDeps-} ]] || (( "\$hostOffset" < 0 )) || return 0
-
   sdkRootHook() {
     # See ../../../build-support/setup-hooks/role.bash
     local role_post
     getHostRoleEnvHook
 
     # Only set the SDK root if one has not been set via this hook or some other means.
-    if [[ ! \$NIX_CFLAGS_COMPILE =~ isysroot ]]; then
+    local cflagsVar=NIX_CFLAGS_COMPILE\''${role_post}
+    if [[ ! \''${!cflagsVar} =~ isysroot ]]; then
       export NIX_CFLAGS_COMPILE\''${role_post}+=' -isysroot $out/${sdkName}.sdk'
     fi
   }
@@ -60,9 +56,9 @@ runCommand "sdkroot-${sdkVersion}" { } ''
   # See ../../../build-support/setup-hooks/role.bash
   getTargetRole
 
-  addEnvHooks "\$targetOffset" sdkRootHook
+  addEnvHooks "\$hostOffset" sdkRootHook
 
   # No local scope in sourced file
-  unset -v role_post
+  unset -v cflagsVar role_post
   hook
 ''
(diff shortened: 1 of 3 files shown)


pkgs/os-specific/darwin/apple-sdk/sdkRoot.nix
```diff
@@ -41,28 +41,24 @@ runCommand "sdkroot-${sdkVersion}" { } ''
 
   install -D '${../../../build-support/setup-hooks/role.bash}' "$out/nix-support/setup-hook"
   cat >> "$out/nix-support/setup-hook" <<-hook
-  #
```
toonn, line 44: So we're dropping the exception when the hook isn't a build-time dependency and we're not doing a native compile?

Seems like a good thing to have it apply consistently.

pkgs/os-specific/darwin/apple-sdk/sdkRoot.nix
```diff
@@ -41,28 +41,24 @@ runCommand "sdkroot-${sdkVersion}" { } ''
...
   sdkRootHook() {
     # See ../../../build-support/setup-hooks/role.bash
     local role_post
     getHostRoleEnvHook
 
     # Only set the SDK root if one has not been set via this hook or some other means.
-    if [[ ! \$NIX_CFLAGS_COMPILE =~ isysroot ]]; then
+    local cflagsVar=NIX_CFLAGS_COMPILE\''${role_post}
+    if [[ ! \''${!cflagsVar} =~ isysroot ]]; then
       export NIX_CFLAGS_COMPILE\''${role_post}+=' -isysroot $out/${sdkName}.sdk'
```
toonn, line 52: Don't we want to reuse the `cflagsVar` here or can you not use an indirect reference with export?

pkgs/os-specific/darwin/apple-sdk/sdkRoot.nix
```diff
@@ -41,28 +41,24 @@ runCommand "sdkroot-${sdkVersion}" { } ''
...
     fi
   }
 
   # See ../../../build-support/setup-hooks/role.bash
   getTargetRole
```
toonn, line 58:
```suggestion
```

If the below is correct I believe this no longer serves a purpose. Actually seems to be redundant even without the change from target to host offset because this sets role_post, which is unset right after.

pkgs/os-specific/darwin/apple-sdk/sdkRoot.nix
```diff
@@ -41,28 +41,24 @@ runCommand "sdkroot-${sdkVersion}" { } ''
...
 
-  addEnvHooks "\$targetOffset" sdkRootHook
+  addEnvHooks "\$hostOffset" sdkRootHook
```
toonn, line 59: This means we run the hooks in the context of the host rather than the target? Because the SDK root is necessary during the build rather than at run-time?

toonn (commented):
Couple questions because I've really not grokked all the cross setup yet.

pkgs/os-specific/darwin/apple-sdk/sdkRoot.nix
reckenrode: I think so? The SDK root is used by the host compiler for the target system.

pkgs/os-specific/darwin/apple-sdk/sdkRoot.nix
reckenrode: It needs to be `depsHostTarget` (i.e., a `buildInput`), but the hook wasn’t being run unless I took this out. The doc seems to suggest it should work, but maybe I was doing something wrong.

aviallon:
Closed by https://github.com/NixOS/nixpkgs/pull/346043

PR #307838, opened 2024-04-29
tree-wide: Replace unsupported dontCheck with doCheck attribute

afh:
## Description of changes

Remove unsupported `dontCheck` attribute (for context see [this comment on #293498](https://github.com/NixOS/nixpkgs/pull/293498#discussion_r1531436723)).

ℹ️ Since `python3Module.gpt-2-simple` is broken on `aarch64-darwin` I was not able to test it locally; yet the GitHub tests indicate that the changes proposed in this PR do not affect it negatively.

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [x] aarch64-darwin
- For non-Linux: Is sandboxing enabled in `nix.conf`? (See [Nix manual](https://nixos.org/manual/nix/stable/command-ref/conf-file.html))
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:
  - [NixOS test(s)](https://nixos.org/manual/nixos/unstable/index.html#sec-nixos-tests) (look inside [nixos/tests](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests))
  - and/or [package tests](https://github.com/NixOS/nixpkgs/blob/master/pkgs/README.md#package-tests)
  - or, for functions and "core" functionality, tests in [lib/tests](https://github.com/NixOS/nixpkgs/blob/master/lib/tests) or [pkgs/test](https://github.com/NixOS/nixpkgs/blob/master/pkgs/test)
  - made sure NixOS tests are [linked](https://nixos.org/manual/nixpkgs/unstable/#ssec-nixos-tests-linking) to the relevant packages
- [ ] Tested compilation of all packages that depend on this change using `nix-shell -p nixpkgs-review --run "nixpkgs-review rev HEAD"`. Note: all changes have to be committed, also see [nixpkgs-review usage](https://github.com/Mic92/nixpkgs-review#usage)
- [ ] Tested basic functionality of all binary files (usually in `./result/bin/`)
[...]

PR #308377, opened 2024-05-01
buildPythonPackage, buildPythonApplication: make doCheck default less confusing

ShamrockLee:
## Description of changes

Make it clear that argument doCheck of buildPythonPackage and buildPythonApplication defaults to true instead of false.

(For context, see https://github.com/NixOS/nixpkgs/pull/307838#discussion_r1585915343)

The current implementation of `mk-python-derivation.nix` appears to default `doCheck` `config.doCheckByDefault or false`, but effectively default it `true`.

```nix
{
  stdenv,
  ...
}:
{
  doCheck ? config.doCheckByDefault or false,
  ...
}@attrs:
stdenv.mkDerivation (removeAttrs attrs [ ... ]) {
  doCheck = false;
  doInstallCheck = attrs.doCheck or true; # ???
}
```

This PR resolves such confusion:
```nix
{
  stdenv,
  ...
}:
{
  doCheck ? true,
  ...
}@attrs:
stdenv.mkDerivation (removeAttrs attrs [ ... ]) {
  doCheck = false;
  # The doCheck argument determines the doInstallCheck attribute.
  doInstallCheck = doCheck;
}
```

Another confusion caused by argument `doCheck` determining attribute `doInstallCheck` should be resolved by instructing people to specify `doInstallCheck` instead of `doCheck` and `installCheckPhase` instead of `checkPhase`. That's a change of interface that needs more discussion, and is excluded from the scope of this PR.

This PR is a back-portable fix that brings no interface changes and no rebuilds.

Cc: @fridh @afh

## Things done

- Built on platform(s)
  - [ ] x86_64-linux (no rebuild)
  - [X] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/development/interpreters/python/mk-python-derivation.nix
+++ b/pkgs/development/interpreters/python/mk-python-derivation.nix
@@ -144,7 +144,7 @@ in
 
 , meta ? {}
 
-, doCheck ? config.doCheckByDefault or false
+, doCheck ? true
 
 , disabledTestPaths ? []
 
--- a/pkgs/development/interpreters/python/python2/mk-python-derivation.nix
+++ b/pkgs/development/interpreters/python/python2/mk-python-derivation.nix
@@ -92,7 +92,7 @@
 
 , passthru ? {}
 
-, doCheck ? config.doCheckByDefault or false
+, doCheck ? true
 
 , disabledTestPaths ? []
 


afh:
@ShamrockLee, thanks for this. Apologies if my confusion about around the `doCheck` default for regular and python packages still persist. Maybe you could help clear it up. What is the motivation for `doCheck` to be `true` for python packages / applications opposed to it being `false` for the (majority?) of regular packages?

Also: should this information about what (`doCheck` and/or `doInstallCheck`) to use for python packages / applications find its way to the [Python-specific section of the Nixpkgs manual](https://nixos.org/manual/nixpkgs/stable/#python)?

TomaSajt:
I think this is the classic mistake of expecting the `@attrs` pattern to respect the `?` pattern.

Example:
```nix
# a.nix
{ a ? "foo" }@attrs:
attrs.a or "bar"
```
When doing `import ./a.nix {}` this will evaluate to `"bar"` even though we told it that if `a` is unset set it to `foo`.
This is because `attrs` is actually just `{}`, exactly what we passed in.

TomaSajt:
Here's the change that broke/changed the old default of `false` to `true`:  https://github.com/NixOS/nixpkgs/commit/f7e28bf5d8181926e600a222cb70180519d09726

TomaSajt:
We could also set `config.doCheckByDefault or true` as the default, so that it starts respecting the config value again.
This should not cause rebuilds, unless you have the config set to `false`, but then you're probably always expecting massive rebuilds anyway.

ShamrockLee:
@afh

> What is the motivation for `doCheck` to be `true` for python packages / applications opposed to it being `false` for the (majority?) of regular packages?

I haven't inspect into the reason why it is enabled by default. The design choice has been made years ago. What I'm doing in this PR is to uncover it.

Besides, defaulting `doCheck`/`doInstallCheck` `true` isn't uncommon for build helpers. Searching the codebase with regular expression `do(Install)?Check = .* or true`, we can find similar defaults in language-specific build helpers of Crystal, Go, Haskell, PHP, and Rust, aside from Python. Such defaults are sensible when there are some default checks or some arguments that, if set, would enable certain pre-defined checks.

> Also: should this information about what (`doCheck` and/or `doInstallCheck`) to use for python packages / applications find its way to the [Python-specific section of the Nixpkgs manual](https://nixos.org/manual/nixpkgs/stable/#python)?

It would be great to explain them  in the Nixpkgs Manual. However, I plan to propose setting it directly via `doInstallCheck` and `installCheckPhase` to make it  less ambiguous. The documentation will be added there.

ShamrockLee:
@TomaSajt

> Here's the change that broke/changed the old default of `false` to `true`: [f7e28bf](https://github.com/NixOS/nixpkgs/commit/f7e28bf5d8181926e600a222cb70180519d09726)

Thanks a lot for finding it out!

> We could also set `config.doCheckByDefault or true` as the default, so that it starts respecting the config value again. This should not cause rebuilds, unless you have the config set to `false`, but then you're probably always expecting massive rebuilds anyway.

According to [the documentation](https://nixos.org/manual/nixpkgs/unstable/#opt-doCheckByDefault), `config.doCheckByDefault` is "default to `false`", and "changing the default may cause a mass rebuild." AFAIK, it is unexpected for `doCheckByDefault = false` to cause rebuild.

TomaSajt:
> @TomaSajt
> 
> > Here's the change that broke/changed the old default of `false` to `true`: [f7e28bf](https://github.com/NixOS/nixpkgs/commit/f7e28bf5d8181926e600a222cb70180519d09726)
> 
> Thanks a lot for finding it out! (#68244)
> 
> > We could also set `config.doCheckByDefault or true` as the default, so that it starts respecting the config value again. This should not cause rebuilds, unless you have the config set to `false`, but then you're probably always expecting massive rebuilds anyway.
> 
> According to [the documentation](https://nixos.org/manual/nixpkgs/unstable/#opt-doCheckByDefault), `config.doCheckByDefault` is "default to `false`", and "changing the default may cause a mass rebuild." AFAIK, it is unexpected for `doCheckByDefault = false` to cause rebuild.

Oh, I see. I didn't see that it was set by default.
I thought the docs would be something like: "either set this to true/false to force it to be true/false, or don't set it to leave the choice to the builder helpers", but I guess it's just defaulted.

FRidh:
Python tests have always been enabled because it is otherwise very hard to
judge during build time whether the package will work at all.

afh (approved):
With the information provided and my current understanding, I support this change. Someone more knowledgable may disagree and hopefully they decide to speak up and share their insight.

Thank you for putting this together, @ShamrockLee, much appreciated! And thank you everyone else for chiming in with your expertise.

TomaSajt (approved):

afh:
@JohnRTitor would you be willing to lend your expertise for review and if acceptable merge this PR?

pkgs/development/interpreters/python/mk-python-derivation.nix
```diff
@@ -309,7 +309,8 @@ let
 
     # Python packages don't have a checkPhase, only an installCheckPhase
     doCheck = false;
-    doInstallCheck = attrs.doCheck or true;
+    # The doCheck argument determines the doInstallCheck attribute.
+    doInstallCheck = doCheck;
```
natsukium, line 313:
```suggestion
    doInstallCheck = attrs.doCheck;
```
Could you clarify this with `attrs`?
I think it is confusing since there is a `doCheck = false` right above it.

pkgs/development/interpreters/python/mk-python-derivation.nix
```diff
@@ -309,7 +309,8 @@ let
 
     # Python packages don't have a checkPhase, only an installCheckPhase
     doCheck = false;
-    doInstallCheck = attrs.doCheck or true;
+    # The doCheck argument determines the doInstallCheck attribute.
```
natsukium, line 312: This comment explains what it does, but I don't think it's necessary because it's just what the code says.

pkgs/development/interpreters/python/mk-python-derivation.nix
TomaSajt: Can't test it right now, but IIRC if we use the `@attrs` pattern, the defaults set with `?` will not be applied when accessing through `attrs.`

ShamrockLee: > Could you clarify this with `attrs`?

`attrs` does not consider the default value given by the set pattern of the build helper. It would be confusing to set a default value in the set pattern (`{ doCheck ? true, ... }`) but shadow it with another default when accessing (`doInstallCheck = attrs.doCheck or true`).

natsukium: Hmmm, indeed, `attrs` seems to be empty, as @TomaSajt pointed out at the beginning.
Then I think `attrs.doCheck or true` would be easier to understand.

The confusion was originally caused by the fact that the argument, `doCheck`, actually evaluates to true even though it defaults to false, right?
Do we need to change this here?

ShamrockLee: > Hmmm, indeed, `attrs` seems empty, as @TomaSajt pointed out at the beginning. Then I think `attrs.doCheck or true` would be easier to understand.

It still shadows the default set by the set pattern, but it sounds like an acceptable trade-off, considering the increased readability.

IMO, the root cause of the readability issue is the choice to specify the `doInstallCheck` attribute with the `doCheck` argument. I'll propose unifying the name (as `doInstallCheck`) in subsequent PR.

pkgs/development/interpreters/python/mk-python-derivation.nix
ShamrockLee: Addressed.

pkgs/development/interpreters/python/mk-python-derivation.nix
ShamrockLee: Addressed.

ShamrockLee:
I just rebased it onto the tip of the `master` branch.

natsukium (approved):
Thanks for improving this. LGTM

ShamrockLee:
Thanks for merging.

I created another PR #324124 proposing directly taking the `installCheck`-related arguments. Please take a look and leave some feedback if you are interested.

pkgs/development/interpreters/python/mk-python-derivation.nix
```diff
@@ -144,7 +144,7 @@ in
 
 , meta ? {}
 
-, doCheck ? config.doCheckByDefault or false
+, doCheck ? true
```
TomaSajt, line 147: It's a bit late now, but I think this fix was not the one that needed to happen, since `doCheck` never gets used.
Only `attrs.doCheck` is ever referenced, which doesn't need this line.
```suggestion
```
Should we open another PR that just removes this line?

ShamrockLee: Ah, yes. Removing this line from the set pattern is a better solution to such confusion.

PR #308395, opened 2024-05-01
simutrans: 121.0 -> 124.3.1; adopt, refactor

sodiboo:
## Description of changes

This package hasn't been updated for 3 years, and all commits since are either treewide changes or people removing themselves as a maintainer.

In that time, there have been 4 major releases. This PR catches up on the following updates:

- [Simutrans 122.0](https://forum.simutrans.com/index.php/topic,20389.0.html)
- [Simutrans 123.0](https://forum.simutrans.com/index.php/topic,21289.0.html)
- [Simutrans 123.0.1](https://forum.simutrans.com/index.php/topic,21377.0.html)
- [Simutrans 124.0](https://forum.simutrans.com/index.php/topic,22946.0.html) (current version)

These updates bring internal engine changes, the most important of which for my use cases is a simplified backend choice. There are only a couple now, and we're using `sdl2` (the previous `sdl` backend is no longer supported)

I've also updated PakSets to their latest versions and briefly tested that the game works. For the reviewer who may not be experienced with Simutrans, a PakSet is basically a collection of assets. The `simutrans` executable is just a game engine, really. You need a PakSet to play the game, that's why they are packaged with it. Along with updating PakSets, i've also added several new PakSets, most importantly the ones from the in-game installer which has been introduced in 123.0.1, from my understanding. I'll get to that in a moment, but first some other changes:

I've cleaned up a lot of the package, among other things now fetching the source code from Subversion instead of SourceForge. I've talked to the Simutrans team on Discord, where they told me that Subversion is preferred over all other methods (it is also the source of truth; the actual dev repo).

Additionally, i'm passing `-DREVISION` into the build config, which previously was just unset in the game. From my understanding, this needs to be set for online play to work. So, i guess all previous versions of this package in nixpkgs just don't work online? I haven't been able to test them, because i use Wayland without XWayland, and SDL2 is necessary for Wayland support. By the way, it works basically flawlessly on Wayland. At least, from a quick glance without having played it, it looks fine.

I feel it may be relevant that i'm not exactly a long-time Simutrans player or anything. I just saw the trailer this week on YouTube, and figured "what the heck", and thought i'd try it. Then, i saw that the nixpkgs version is horribly out of date and doesn't even work on my system (Wayland moment), so i figured i'd update it. At a certain point, i became familiar enough with the codebase that i thought i might as well adopt the package entirely.

I am yet to actually play the game at all. So i may turn out to be somewhat uninterested in the game itself. Only time will tell. But i've added myself as a maintainer for now because i plan to, at the very least, update it to 124.0 and then we'll see if i still care or not.

Despite this, i'm somewhat familiar with at least some parts of the codebase. ~~I wrote the patch i've applied to the package, because i realized that the Install button can't possibly work on NixOS~~ (now obsolete in 124.0 which has much better support for it to begin with). I've documented a lot of this in code comments, along with other processes like how to find the revision number, where to source the paksets, how to name them, etc. ~~As such, i won't cover why in this PR body, but given that the Install button doesn't work, i've opted to remove it entirely (actually just hiding it; along with a more subtle change to not affect the layout). I've deemed this to be okay, given that in the same move,~~ i've also added all of the PakSets that the game would try to download anyways. All of them are included.

And finally, i've tried to mostly avoid changing any outwards-facing interfaces. I have, however, taken the liberty to revamp how the paksets are specified. The final `simutrans` package is now overridable, and you can pass it a list of PakSets. This allows a user to easily package their own, and add them, without needing to merge those changes into nixpkgs. This is a breaking change, but i figured it's fine since [i found exactly 0 uses of `config.simutrans.paksets` across GitHub](https://github.com/search?q=%2Fsimutrans%5C.paksets%2F&type=code). Addiitonally, the default set of PakSets is now all of them, rather than the seemingly arbitrary choices we previously had.

I'd like to see input on the following matters before this PR is merged, though:
- What is `simutrans_binaries` doing? Why is it named so differently from everything else? \
  [No other packages are named `_binaries`](https://search.nixos.org/packages?channel=unstable&query=_binaries), but it's [common to see `-unwrapped`](https://search.nixos.org/packages?channel=unstable&query=-unwrapped). I want to rename it, but i'm not sure that's a great idea, or necessary?
- ~~Do we add additional compile flags? I'm not sure, but i notice e.g. [`USE_UPNP`](https://github.com/simutrans/simutrans/blob/ef1c33d800952bafaedbbf2012fba2decbc51513/config.template#L42-L43) should probably be enabled? I haven't enabled anything besides `FLAGS := -DREVISON` because the previous package didn't.~~ (now  resolved by myself; the Simutrans team referred me to the wiki which clarifies these matters)

---

## Things done

- Built on platform(s)
  - [X] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---


sodiboo:
Additional notes:

---

`simutrans_binaries` is defined here:

https://github.com/NixOS/nixpkgs/blob/53f114f694a0b97b156cdc60fd2448fb5133b91b/pkgs/top-level/all-packages.nix#L37447-L37449

It could easily be changed to `simutrans-unwrapped`, which i believe is the convention other packages generally follow? I haven't, mainly because i'm not sure if i should.

---

Should we add a `.desktop` file? [Arch does it](https://gitlab.archlinux.org/archlinux/packaging/packages/simutrans/-/blob/main/simutrans.desktop?ref_type=heads) and it makes sense, since Simutrans is a GUI application. Currently, it doesn't appear in my fuzzel or similar, because it's lacking a desktop file.

---

Should PakSets maybe be a package set? That's kinda what it is, but i'm not familiar with how to create them or if/when i should. 

Previously, they were not exposed at all, and could be customized only by changing a magic string `config.simutrans.paksets`, but i've changed the package to be overridable instead, so now to achieve the same thing you can do `pkgs.simutrans.override { paks = [ ... ]; }` where `paks` is an array of PakSets, which you can take from [`pkgs.simutrans.pakSpec`](https://github.com/NixOS/nixpkgs/blob/53f114f694a0b97b156cdc60fd2448fb5133b91b/pkgs/games/simutrans/default.nix#L181). The main benefit of this is that a user can add their own PakSets, as i mentioned. But, this attribute feels like maybe an awkward place to expose `pakSpec`? This is mainly because it's not necessarily part of `simutrans` itself (there exists forks of Simutrans, e.g. Simutrans Extended, which also accept these PakSets, so it's weird to reference the "wrong fork" if you wanna customize them). PakSets would nautrally lend themselves to being a package set, sorta like gnome extensions and vim plugins. but that feels overkill maybe?

An alternative way is to expose `pakSpec` as the `paks` attribute in the override, instead of it being an array. Then, consuming code that wants to "pick" just some paksets could do something like `pkgs.simutrans.override ({ paks }: { paks = { inherit (paks) pak64 pak128 "pak192.comic"; }; })`, which still lets them add their own alongside by putting `"pak92.my_fancy_one"` as another key.

They could also be exposed as top-level packages flat alongside Simutrans, but this has a major downside of it being nontrivial to select "all paksets please". The main advantage of doing it as a custom package set is that it allows for overlays. But again, is it necessary at all?

sodiboo:
I've converted this to a draft to prevent it merging immediately because i noticed something peculiar: When i tried actually playing the game just now, the tutorial doesn't work. It throws an error. I've asked in the Simutrans Discord and will resolve this issue before marking it as ready again. What i had tested previously is just getting to the splash screen, making sure everything looks right, ensuring the revision is properly set, and disabling the broken Install button.

sodiboo:
wait what? this has got me confused. multiple people tell me that the tutorials might be too new; and that does seem to be the case! other scenarios, not called `tutorial`, from `pak64` seem to work just fine??

for reference, the reason i marked as draft is because the tutorials from `pak128`, `pak64` and `pak192.comic` all were broken. i figured something must have gone wrong in my package, since how could they all be broken at the same time, in the same way? (all tabs on the tutorial menu showed `(null)` and then reloading caused an error in the scripts).

but apparently, it really is the tutorials? because i just confirmed that other scenarios in `pak64` work just fine. and the other scenarios in `pak128` work too.

so... i guess i didn't do anything wrong with `pak64` and `pak128`?

but `pak192.comic` was genuinely on a too recent version:

from its [`README.md`](https://github.com/Flemmbrav/Pak192.Comic/blob/711a6fa4ca993752ec219c9b1e4702ca9d442324/README.md?plain=1#L21-L22):

> Since this is a nightly pakset, it may not run with the Simutrans version you have already installed.

and the version i was shipping was `0.7.1`, released this february. the game version is about 2 years old. so, let's... not do that. i'm gonna revert it to an older version. game version `123.0.1` knows about v0.6-rc2 of `pak192.comic`, so i'm gonna use the actual release version of `v0.6`. that version doesn't actually have a tutorial, so no tutorial woes in there i guess?

sodiboo:
Here's a response from the developer of `pak192.comic` confirming that `v0.6` is good

image

As for other paksets? Most of them are ancient, and don't receive updates, so there's no worry about requiring simutrans nightly on that end.

sodiboo:
Simutrans has a new [release 124.0](https://forum.simutrans.com/index.php/topic,22946.0.html), so this PR is already out of date. Will update.

sodiboo:
I believe this is ready for review again. It's up-to-date with 124.0 and recent paksets, everything works (including the tutorial), and there aren't really any issues remaining. Several patches have been applied.

I'd like to see reviewers comment on the following:

- What is the purpose of `simutrans_binaries`? I think it should be renamed to `simutrans_unwrapped` or removed entirely. Do we need it? See also the initial PR body.
- Should we add additional XDG resources, such as a `.desktop` file? [Arch does this](https://gitlab.archlinux.org/archlinux/packaging/packages/simutrans/-/blob/main/simutrans.desktop?ref_type=heads).
- Should we just override the path behaviour and use proper XDG basedirs? The current behaviour is a compromise between the previous behaviour of this package (which overrode the default behaviour of the game; placing everything in`~/ .simutrans`) and the current default behaviour of the game (almost following XDG basedirs but not actually if the env vars are unset). It's unintuitive, but my best shot at "not changing behaviour". I strongly think it would be best to just obsolete the previous behaviour. Are there *any* users that depend on the nixpkgs package for Simutrans and would even care about or be affected by such a change?

sodiboo:
I noticed just now that building my system failed, because the hashes for pak64.german and pak128.german changed. I don't know why this happened; it should not be happening. Those are fully qualified versions i'm requesting.

fpletz:
Just found this PR. First of all, thank you very much for your detailed notes here. I've last played simutrans over 12 years ago and would love to see this getting more love. 

> * What is the purpose of `simutrans_binaries`? I think it should be renamed to `simutrans_unwrapped` or removed entirely. Do we need it? See also the initial PR body.

The convention here is that unwrapped versions are only provided if there are any use case where the user would want to use the unwrapped version. Furthermore, if the wrapper doesn't change any behavior and just fixes running the program, just proved the wrapped version. I think in simutrans' case, we can just remove that weird binaries passthru business.

> * Should we add additional XDG resources, such as a `.desktop` file? [Arch does this](https://gitlab.archlinux.org/archlinux/packaging/packages/simutrans/-/blob/main/simutrans.desktop?ref_type=heads).

Yup, that sounds useful. Lots of launchers use those to discover installed packages.

> * Should we just override the path behaviour and use proper XDG basedirs? The current behaviour is a compromise between the previous behaviour of this package (which overrode the default behaviour of the game; placing everything in`~/ .simutrans`) and the current default behaviour of the game (almost following XDG basedirs but not actually if the env vars are unset). It's unintuitive, but my best shot at "not changing behaviour". I strongly think it would be best to just obsolete the previous behaviour. Are there _any_ users that depend on the nixpkgs package for Simutrans and would even care about or be affected by such a change?

Well, I guess users will generally expect what upstream does here. On the other hand, since there are no open issues for simutrans and it is indeed quite outdated, it would be fair to say that not a lot of people use it. I would personally go with upstream but you as maintainer may decide. :smile:

fpletz:
Another thing, it would be nice if we could use `fetchpatch` to get those patches instead of putting them into nixpkgs. Not a requirement and I saw that you needed to adjust some tings. Maybe just a thing you could look into if you haven't already.

sodiboo:
> Furthermore, if the wrapper doesn't change any behavior and just fixes running the program, just provide the wrapped version. I think in simutrans' case, we can just remove that weird binaries passthru business.

I would say that it *does* change behaviour in its current state, and thus "should" have an unwrapped variant, but looking more closely at the command line arguments behaviour, i see that we could definitely modify it to work without "modifying behaviour". In particular, what the wrapper does is three things:

- It changes the default "user dir".  This can always be overridden by setting `SIMUTRANS_USERDIR`
- It changes the default "install dir". This can always be overridden by setting `SIMUTRANS_INSTALLDIR`
- It changes the default "base dir". This is normally overridden by passing `-set_basedir`, but since we pass it already, that one takes priority

The change necessary would be to put this as a suffix argument rather than prefix, and then i can see the unwrapped version being wholly unnecessary. (i tested the behaviour of the parser, and it accepts multiple `-set_basedir` args with the first one always taking precedence)

> Well, I guess users will generally expect what upstream does here.

As a user, i do not like what upstream is doing (prefers `~/simutrans`). In particular, i don't expect any programs to write non-dotfiles to my home directory. In testing this package, the game would write to my locally checked out repository anytime i passed the wrong arguments to override it. I do, however, like how upstream handles it within their Flatpak package (writes to `$XDG_DATA_HOME/simutrans`) and as a user i do not expect it to violate the XDG basedirs spec. However, it is not my choice to make just because of how i prefer it be done. Upstream thinks users expect it to be in `~/simutrans` because "that's how it has been for more than twenty years" and that's fair enough, but in nixpkgs we have likely no users that have been using it for twenty years, and even so, the package didn't follow this to begin with.

> it would be nice if we could use `fetchpatch` to get those patches instead of putting them into nixpkgs

~~I agree, but there isn't an obviously good place to fetch these from? The subversion repository has no primary web interface, so we'd have to fetch it from e.g. a git mirror. There are two github mirrors and they do not provide identical patches. Here's a commit from yesterday in [aburch/simutrans](https://github.com/aburch/simutrans/commit/502de2842e4def6e7694a5f1cc8060fb4a5d7aa4.patch) and [simutrans/simutrans](https://github.com/simutrans/simutrans/commit/a004ef2b132224d677cbd439a07e08af72676fd0.patch). That's the main reason i didn't want to fetch from github; it's not the canonical source.

I suppose one way to do it (lol) is to actually fetch that revision and the one before, and creating a diff. This is the worst way to do it, but would be "from the canonical repo". Obviously we're not doing that. If there is a real way to fetch SVN revisions as patches, i'd be happy to use that! But as far as i can tell, there isn't, so that's why i've been fetching them "manually".~~

In writing this all, before hitting the big green button to publish this reply, i figured i'd take a look at subversion's command line. Apparently a command like `svn diff --git svn://servers.simutrans.org@r11272 svn://servers.simutrans.org@r11273` would probably be good for all we need. So, uh, i'll be seeing if i can implement that soon! `miniupnpc.patch` is staying though, since that one is *actually* modified from the source. A relevant question is if such a helper belongs in this pull request, or if i should add it to nixpkgs in a separate PR? Since i could see this being useful outside of simutrans.

jj-marr:
Anything this is blocked on? Would pick this PR up if it would otherwise be abandoned.

sodiboo:
blocked on procrastination and lack of object permanence

sodiboo:
rebased this just now. some of the pakset sources are dead now, so it no longer builds :(

sodiboo:
I've made an executive decision on the XDG basedir thing and this is how it's going to work:

- If there is a `~/.simutrans` directory already, from an older version of this package we will default to using that.
- Else, we follow XDG basedir (which is **NOT** how the game "wants" to work by default)

This can be overridden by setting `SIMUTRANS_USERDIR=` and `SIMUTRANS_INSTALLDIR=` to empty strings. This will force the default behaviour of the executable, and cause it to prefer `~/simutrans` (notice the lack of a dot)

sodiboo:
@makuru-org you wanted to be a maintainer here. should i add you now?

makuru-org:
> @makuru-org you wanted to be a maintainer here. should i add you now?

Thanks!

makuru-org (approved):
Update runs and file looks way better.

pkgs/by-name/si/simutrans/package.nix
```diff
@@ -202,15 +445,19 @@ let
...
       license = with lib.licenses; [
         artistic1
         gpl1Plus
       ];
-      maintainers = [ ];
+      maintainers = with lib.maintainers; [
+        sodiboo
+        Makuru
+      ];
       platforms = lib.platforms.linux; # TODO: ++ darwin;
```
SuperSandro2000, line 479: We should set sourceProvenance

pkgs/by-name/si/simutrans/package.nix
```diff
@@ -2,196 +2,439 @@
...
+        '';
 
-      passthru.meta = binaries.meta // {
+      strip = false;
+      passthru.meta = simutrans-bin.meta // {
         hydraPlatforms = [ ];
       };
-      passthru.binaries = binaries;
+      passthru.bin = simutrans-bin;
+      passthru.pakSpec = pakSpec;
```
SuperSandro2000:
```suggestion
     passthru = {
       binaries = binaries;
       bin = simutrans-bin;
       pakSpec = pakSpec;
       meta = simutrans-bin.meta // {
         hydraPlatforms = [ ];
       };
     };
```
I am a bit confused by having bin and binaries here. They seem duplicated. Also placing meta in passthru is a bit unusual. I would expect it on the same level as passthru

pkgs/by-name/si/simutrans/package.nix
```diff
@@ -2,196 +2,439 @@
...
-        cd "$out"/share/simutrans
-        exec "${binaries}/bin/simutrans" -use_workdir "\$@"
-        EOF
-        chmod +x "$out/bin/simutrans"
-      '';
+          mkdir -p $out/share/{applications,icons/hicolor/scalable/apps}
+
+          ln -s ${src}/trunk/src/simutrans/simutrans.svg $out/share/icons/hicolor/scalable/apps/simutrans.svg
+          substitute ${src}/trunk/src/linux/simutrans.desktop $out/share/applications/simutrans.desktop \
+            --replace-fail "Exec=simutrans" "Exec=$out/bin/simutrans"
```
SuperSandro2000: We should not do this as many desktops like KDE do not properly update the paths to the binaries. So just leaving simutrans here is actually preferred.

pkgs/by-name/si/simutrans/package.nix
```diff
@@ -2,196 +2,439 @@
...
-        cat > "$out/bin/simutrans" <<EOF
-        #!${runtimeShell}
-        cd "$out"/share/simutrans
-        exec "${binaries}/bin/simutrans" -use_workdir "\$@"
-        EOF
-        chmod +x "$out/bin/simutrans"
-      '';
+          mkdir -p $out/share/{applications,icons/hicolor/scalable/apps}
+
+          ln -s ${src}/trunk/src/simutrans/simutrans.svg $out/share/icons/hicolor/scalable/apps/simutrans.svg
```
SuperSandro2000: We should cp here to not keep a refernece on src

pkgs/by-name/si/simutrans/package.nix
```diff
@@ -2,196 +2,439 @@
...
+          # Then, fix the hash, and keep fixing `prefix` until it shows `.pak` files.
+          # Then you can remove `inspect` and it should work.
+          inspect ? false,
+        }:
+        mkPak {
+          inherit pakName prefix inspect;
+          src = fetcher { inherit url hash; };
+        }
+      )
+      {
```
SuperSandro2000, line 116: Would it be easier to manage this if we place this in a separate file?

pkgs/by-name/si/simutrans/package.nix
```diff
@@ -2,196 +2,439 @@
...
+  fetchcab =
+    args:
+    let
+      src = fetchurl args;
+    in
+    runCommand "source-cab" { nativeBuildInputs = [ cabextract ]; } ''
+      cabextract ${src} -d $out
+    '';
+
+  fetchzip' = defaults: args: fetchzip (defaults // args);
```
SuperSandro2000, line 29: Don't we need to use override to apply the args to fetchzip?

pkgs/by-name/si/simutrans/package.nix
```diff
@@ -2,196 +2,439 @@
...
-    url = "mirror://sourceforge/simutrans/simutrans/${ver_dash}/simutrans-src-${ver_dash}.zip";
-    sha256 = "1f463r6kr5ig0zd3mncc74k93xbjywsq3d06j5r17831jyc9bzb9";
+  fetchcab =
+    args:
+    let
+      src = fetchurl args;
+    in
+    runCommand "source-cab" { nativeBuildInputs = [ cabextract ]; } ''
+      cabextract ${src} -d $out
+    '';
```
SuperSandro2000: Could we use postFetch inside of fetchurl to extract the files?

SuperSandro2000 (commented):
Does this still work after the recent SDL2 changes? Also I am not sure how well the commit history is managed, I didn't look at it but I would be fine with squashing some of the commits together to make things easier when doing another rebase or so.

pkgs/by-name/si/simutrans/package.nix
sodiboo: > I am a bit confused by having bin and binaries here. They seem duplicated.

I tried to rename `simurans_binaries` -> `simutrans-bin`, under the impression that a `-bin` suffix is more conventional in nixpkgs ([this is the only one with a `_binaries` suffix](https://search.nixos.org/packages?channel=25.05&query=_binaries) vs [200-something with `-bin`](https://search.nixos.org/packages?channel=25.05&query=-bin)). I think i fucked up if both are visible in the end result.

> Also placing meta in passthru is a bit unusual. I would expect it on the same level as passthru

Yeah, idk, [it was like that before i touched it](https://github.com/NixOS/nixpkgs/blob/4792576cb003c994bd7cc1edada3129def20b27d/pkgs/by-name/si/simutrans/package.nix#L129-L131) and i never questioned it. I'll change this.

pkgs/by-name/si/simutrans/package.nix
sodiboo: hm.. i'm actually not sure if the paksets contain source code or bytecode. will find out

pkgs/by-name/si/simutrans/package.nix
sodiboo: need? no, this works fine.

could? almost certainly. i actually never thought about the fact that `fetchzip` is just `stdenv.mkDerivation` and that it will naturally support overrides.

pkgs/by-name/si/simutrans/package.nix
sodiboo: probably! never knew that existed. i'll see what i can do

pkgs/by-name/si/simutrans/package.nix
sodiboo: probably.

sodiboo:
> Does this still work after the recent SDL2 changes?

I think so? Game launched fine when i tested the executable last night. What changes are you thinking of?

> I am not sure how well the commit history is managed, I didn't look at it but I would be fine with squashing some of the commits together to make things easier when doing another rebase or so.

I tried to keep it fairly clean. like, it contains three separate major version upgrades and there were a bunch of patches that are no longer necessary. so all that could probably be squashed, but i still think just about every commit makes sense in isolation.

pkgs/by-name/si/simutrans/package.nix
```diff
@@ -2,196 +2,439 @@
...
+            "nightly builds"
+            "pak128.CS-r2096.zip"
+          ];
+          hash = "sha256-vQ8HZqwICB3CUGQW8NwEWGVA3anUbCOdj3BEV3lLLpk=";
+        };
+        "pak128.german" = {
+          url = "https://pak128-german.de/PAK128.german_2.3_beta.zip";
+          prefix = "/PAK128.german";
+          hash = "sha256-anWImSrS5ayjYzqVV7MY2VZ4dD/jLkhaavcZaufre+k=";
+        };
```
makuru-org: Apparently version `2.3` has a bug where the scenarios of the German map don't load right, update to:
`https://www.pak128-german.de/PAK128.german_2.4_beta.zip`

sodiboo: ooh. good catch. [their README](https://www.pak128-german.de/inc/readme.html) also recommends `2.4` for Simutrans `124.3.1`; i used 2.3 here because [that's what's included in the game by default](https://github.com/simutrans/simutrans/blob/923ecf7256474d58df6643e2a12e5a557a9dc2a2/tools/get_pak.sh#L218)

pkgs/by-name/si/simutrans/package.nix
sodiboo: wait no, i can't do that at all.

this part is overridable:

https://github.com/NixOS/nixpkgs/blob/0a81508bd9a411b1c08eb241f990912a3ff6f874/pkgs/build-support/fetchzip/default.nix#L8-L14

but this part is just a normal function, not `mkOverridable`:

https://github.com/NixOS/nixpkgs/blob/0a81508bd9a411b1c08eb241f990912a3ff6f874/pkgs/build-support/fetchzip/default.nix#L16-L33

so, what i do here with my `fetchzip'` is the only way to do it i think? `fetchurl` is basically the same, and `fetchzip` is essentially implemented as `fetchurl (defaults // args)` as well.

https://github.com/NixOS/nixpkgs/blob/0a81508bd9a411b1c08eb241f990912a3ff6f874/pkgs/build-support/fetchzip/default.nix#L102

pkgs/by-name/si/simutrans/package.nix
sodiboo: something like this would be the "most proper" way to do it, using just args of `fetchurl`:

```nix
fetchcab =
  {
    nativeBuildInputs ? [ ],
    postFetch ? "",
    ...
  }@args:
  fetchurl (
    {
      recursiveHash = true;
    }
    // args
    // {
      nativeBuildInputs = [ cabextract ] ++ nativeBuildInputs;
      postFetch = ''
        tmp=$(mktemp)
        mv $out $tmp
        cabextract $tmp -d $out
        ${postFetch}
      '';
    }
  );
```

Or something like that, which is slightly dirtier and shorter but is basically equivalent:

```nix
fetchcab =
  args:
  fetchurl (
    {
      recursiveHash = true;
      nativeBuildInputs = [ cabextract ];
      postFetch = ''
        tmp=$(mktemp)
        mv $out $tmp
        cabextract $tmp -d $out
      '';
    }
    // args
  );
```

But i'm not convinced these are any better than:

```nix
fetchcab =
  args:
  runCommand "source-cab" { nativeBuildInputs = [ cabextract ]; } ''
    cabextract ${fetchurl args} -d $out
  '';
```

It's so beautiful and simple as-is. I like what i came up with. I guess it adds +1 derivation? But it's way simpler.

pkgs/by-name/si/simutrans/package.nix
SuperSandro2000: I think the -bin suffix is for real packages and as far as I have understood it, simutrans-binaries is not a real packages but just a collection of binary files for simutrans.

pkgs/by-name/si/simutrans/package.nix
SuperSandro2000: > need? no, this works fine.

Oh, I missed the (). With them it works, without it doesn't. Probably fine to keep. If we had deep attrs we should probably use lib.recursiveUpdate.

pkgs/by-name/si/simutrans/package.nix
SuperSandro2000: Keeping them packed makes FOD also easier. Whatever suites you.

pkgs/by-name/si/simutrans/package.nix
sodiboo: It's kind of a real package? We do expose it in the toplevel package attrset?

pkgs/by-name/si/simutrans/package.nix
sodiboo: source code. the Squirrel language does have a JIT compiler and from what i gather it supports an intermediate bytecode representation (like a jar file), but the game seems to expect *source code* in the paksets.

`fd '\.nut$' $(nix build --print-out-paths --no-link .#simutrans)`

pkgs/by-name/si/simutrans/package.nix
sodiboo: i think for now, though, it's fine? we might want to expand the `pakSpec` into its own packageset (cc @makuru-org you suggested this to me?), and at *that* point it makes a lot of sense to split them all into separate files. but with how it currently works, i don't think there's much of a need to separate them. not yet. maybe in a followup PR that splits them as suggested.

PR #308504, opened 2024-05-02
circt: fix darwin linking issues

Liamolucko:
## Description of changes

This PR gets CIRCT to build successfully on Darwin.

While `circt-llvm` technically already builds, the output is unusable because the binaries try to reference libraries in ../lib, which doesn't work because the libraries are in another derivation. This causes `circt` itself to fail to build.

This PR fixes that by adding a `postFixup` step that changes all the libraries to be referenced by their absolute paths.

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin (I tried to via. Rosetta, and it seemed to build but then Rosetta started throwing errors during check phase (which is a first!))
  - [x] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [x] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/development/compilers/circt/circt-llvm.nix
+++ b/pkgs/development/compilers/circt/circt-llvm.nix
@@ -1,4 +1,5 @@
-{ stdenv
+{ lib
+, stdenv
 , cmake
 , ninja
 , circt
@@ -31,6 +32,14 @@
 
   outputs = [ "out" "lib" "dev" ];
 
+  # Get rid of ${extra_libdir} (which ends up containing a path to circt-llvm.dev
+  # in circt) so that we only have to remove the one fixed rpath.
+  postPatch = lib.optionalString stdenv.isDarwin ''
+    substituteInPlace llvm/llvm/cmake/modules/AddLLVM.cmake \
+      --replace-fail 'set(_install_rpath "@loader_path/../lib''${LLVM_LIBDIR_SUFFIX}" ''${extra_libdir})' \
+        'set(_install_rpath "@loader_path/../lib''${LLVM_LIBDIR_SUFFIX}")'
+  '';
+
   postInstall = ''
     # move llvm-config to $dev to resolve a circular dependency
     moveToOutput "bin/llvm-config*" "$dev"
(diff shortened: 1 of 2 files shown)


pkgs/development/compilers/circt/circt-llvm.nix
```diff
@@ -31,6 +32,14 @@
 
   outputs = [ "out" "lib" "dev" ];
 
+  # Get rid of ${extra_libdir} (which ends up containing a path to circt-llvm.dev
+  # in circt) so that we only have to remove the one fixed rpath.
+  postPatch = lib.optionalString stdenv.isDarwin ''
+    substituteInPlace llvm/llvm/cmake/modules/AddLLVM.cmake \
+      --replace-fail 'set(_install_rpath "@loader_path/../lib''${LLVM_LIBDIR_SUFFIX}" ''${extra_libdir})' \
+        'set(_install_rpath "@loader_path/../lib''${LLVM_LIBDIR_SUFFIX}")'
+  '';
```
Liamolucko, line 41: This (and the `-delete_rpath @loader_path/../lib` in postFixup) isn't needed for things to work, it just produces cleaner output (makes it so that there aren't any unused rpaths). Linux actually has the same problem, there's an unused $ORIGIN/../lib in the Linux version of circt's RUNPATH.

pkgs/development/compilers/circt/circt-llvm.nix
```diff
@@ -50,6 +59,30 @@
...
   '';
 
+  # Replace all references to @rpath with absolute paths and remove the rpaths.
+  #
+  # This is different from what the regular LLVM package does, which is to make
+  # everything absolute from the start: however, that doesn't work for us because
+  # we have `-DBUILD_SHARED_LIBS=ON`, meaning that many more things are
+  # dynamically rather than statically linked. This includes TableGen, which then
+  # fails to run halfway through the build because it tries to reference $lib when
+  # it hasn't been populated yet.
```
Liamolucko, line 69: A different way of solving this would to do what LLVM does for its tests and set DYLD_LIBRARY_PATH while building, but I don't know how to smuggle an environment variable through CMake/Ninja like that.

Fugoes:
I encounter the same problem recently and I come up with another way of fixing this problem using `install_name_tool` and does not need to modify cmake files, see: https://github.com/NixOS/nixpkgs/pull/308669.

SharzyL:
Both solutions seem to work. This solution makes cleaner outputs while Fugoes' solution is much simpler. @pineapplehunter how do you like?

pineapplehunter:
I got my hands on a Mac and tried it both.
Although I like the solution from @Fugoes for maintainability, it seems to increase the closure size. (it includes the -dev output)
I think keeping the closure size small is important, so I think I prefer this PR.

pineapplehunter (approved):
Checked on aarch64-darwin and x86_64-linux.
LGTM

PR #308515, opened 2024-05-02
lyra-cursors: init at 0-unstable-2021-12-04

LordMZTE:
## Description of changes

This adds the [Lyra-Cursors cursor theme](https://github.com/yeyushengfan258/Lyra-Cursors).

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:
- [x] ~~Tested basic functionality of all binary files (usually in `./result/bin/`)~~ - No binaries

---

### DIFF
--- a/pkgs/by-name/ly/lyra-cursors/package.nix
+++ b/pkgs/by-name/ly/lyra-cursors/package.nix
@@ -1,0 +1,82 @@
+{
+  lib,
+  stdenvNoCC,
+  inkscape,
+  xcursorgen,
+  fetchFromGitHub,
+  fetchpatch2,
+}:
+let
+  styles = [
+    "LyraB"
+    "LyraF"
+    "LyraG"
+    "LyraP"
+    "LyraQ"
+    "LyraR"
+    "LyraS"
+    "LyraX"
+    "LyraY"
+  ];
+
+  # This is a patch from a fork of the upstream repository which addresses several issues with the
+  # build script such as the fact that the style to build isn't hardcoded. We don't simply use this
+  # fork as source, as the upstream repository is what we want to track.
+  buildScriptPatch = fetchpatch2 {
+    name = "use-more-flexible-build-script.patch";
+    url = "https://github.com/KiranWells/Lyra-Cursors/commit/2735acb37a51792388497c666cc28370660217cb.patch?full_index=1";
+    hash = "sha256-KCT4zNdep1TB7Oa4qrPw374ahT30o9/QrNTEgobp8zM=";
+  };
+in
+stdenvNoCC.mkDerivation {
+  pname = "lyra-cursors";
+  version = "0-unstable-2021-12-04";
+
+  src = fetchFromGitHub {
+    owner = "yeyushengfan258";
+    repo = "Lyra-Cursors";
+    rev = "c096c54034f95bd35699b3226250e5c5ec015d9a";
+    hash = "sha256-lfaX8ouE0JaQwVBpAGsrLIExQZ2rCSFKPs3cch17eYg=";
+  };
+
+  nativeBuildInputs = [
+    inkscape
+    xcursorgen
+  ];
+
+  patches = [ buildScriptPatch ];
+
+  dontConfigure = true;
+
+  postPatch = ''
+    patchShebangs build.sh
+  '';
+
+  buildPhase = ''
+    runHook preBuild
+
+    rm -r dist
+    for THEME in ${lib.escapeShellArgs styles}; do
+      ./build.sh "$THEME"
+    done
+
+    runHook postBuild
+  '';
+
+  installPhase = ''
+    runHook preInstall
+
+    mkdir -p $out/share/icons
+    mv dist/*-cursors $out/share/icons
+
+    runHook postInstall
+  '';
+
+  meta = {
+    description = "A cursor theme inspired by macOS and based on capitaine-cursors";
+    homepage = "https://github.com/yeyushengfan258/Lyra-Cursors";
+    license = lib.licenses.gpl3Only;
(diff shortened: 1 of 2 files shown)


pkgs/by-name/ly/lyra-cursors/package.nix
```diff
@@ -0,0 +1,99 @@
...
+
+    installPhase = ''
+      runHook preInstall
+
+      cp "$src" "$out"
+      chmod +x "$out"
+      patchShebangs "$out"
+
+      runHook postInstall
+    '';
```
emilazy, line 54: I believe the fixup phase will handle the shebang if you don’t disable it.

pkgs/by-name/ly/lyra-cursors/package.nix
```diff
@@ -0,0 +1,99 @@
...
+    mkdir -p $out/share/icons
+    mv dist/*-cursors $out/share/icons
+
+    runHook postInstall
+  '';
+
+  meta = with lib; {
+    description = "A cursor theme inspired by macOS and based on capitaine-cursors";
+    homepage = "https://github.com/yeyushengfan258/Lyra-Cursors";
+    license = licenses.gpl3;
```
emilazy, line 95: Please use `gpl3Only` or `gpl3Plus` as appropriate.

emilazy (changes requested):

pkgs/by-name/ly/lyra-cursors/package.nix
```diff
@@ -0,0 +1,97 @@
...
+      validStyles
+    else
+      let unknown = lib.subtractLists validStyles styles; in
+      if (unknown != [ ]) then
+        throw "Unknown style(s): ${lib.concatStringsSep " " unknown}"
+      else styles;
+
+  # This is a buildscript from a fork of the upstream repository which addresses several issues
+  # such as the fact that the style to build isn't hardcoded. We don't simply use this fork as
+  # source, as the upstream repository is what we want to track.
```
emilazy, line 33: Could we simply `substituteInPlace` the build script ourselves? It seems like the main change we need here is just to use `$1` for the theme name, and it would feel better to me than relying on a downstream that makes a bunch of changes to the build script. However, if those other changes are important then this may be okay. In that case, it may be better to `fetchpatch2` the commit URL (`https://github.com/KiranWells/Lyra-Cursors/commit/2735acb37a51792388497c666cc28370660217cb.patch?full_index=1`) and apply it directly to the source, rather than using a separate derivation like this.

pkgs/by-name/ly/lyra-cursors/package.nix
```diff
@@ -0,0 +1,97 @@
...
+  ];
+
+  selectedStyles =
+    if (styles == [ ]) then
+      validStyles
+    else
+      let unknown = lib.subtractLists validStyles styles; in
+      if (unknown != [ ]) then
+        throw "Unknown style(s): ${lib.concatStringsSep " " unknown}"
+      else styles;
```
emilazy, line 29: Is the complexity of the customization point here worthwhile? The AUR package in the fork simply builds all these styles, which seems like a simpler option to me. Hydra will handle doing the build for users, anyway, so it would make things slower for most users to customize this.

LordMZTE: Since the build script already expects the names of the styles to build, I don't think this is a lot of extra effort. I don't see a reason not to give the user the ability to remove styles they don't want for whatever reason, especially since this doesn't pose any drawbacks for users who want the full theme.

pkgs/by-name/ly/lyra-cursors/package.nix
LordMZTE: It would be a little much for a simple `substituteInPlace`, as that build script makes several other important changes, namely:
- Not dumping the build artifacts into a fixed output directory, but one that includes the theme name
- Adding a "Cursors" suffix to the theme name declared in `index.theme`, which appears to be convention with cursor themes to make them look distinct from icon themes
- Replacing the overcomplicated `find`-based solution to rendering the SVGs. This means we have a lot less Inkscape invocations and thus faster builds.

I see 3 options here:
- We include a custom patch that includes a boiled down version of these changes in the package
- We write a completely custom build script directly in the derivation, but this would get somewhat complicated.
- We fetch the patch from the fork as you suggested.

What would you suggest?

pkgs/by-name/ly/lyra-cursors/package.nix
emilazy: It definitely shouldn’t use the name `styles` since that’s a plausible package name that might end up in Nixpkgs in future and break the default here. Generally speaking if people are going to want only a subset of styles I think it makes more sense to split this into multiple outputs or multiple packages and offer a package that combines them all.

emilazy (approved):
Looks good; sorry this took a while! I’ve confirmed the build on Darwin. Do you want this backporting to 24.11?

Edit: Ah, you need to run the formatter though :) And while you’re at it, we prefer to avoid `with lib;` these days.

LordMZTE:
Thanks! I've removed the `with lib;` and run the formatter. I'd be fine with a backport!

emilazy (approved):

emilazy:
Thanks!

## `nixpkgs-review` result

Command: `nixpkgs-review pr 308515`

---
### `aarch64-darwin`
<details>
  <ul>
    <li>lyra-cursors</li>
  </ul>
</details>

PR #308059, opened 2024-04-30
build-support/php: init `buildComposerProject2` and `mkComposerRepository2`

drupol:
This PR:

- Add a new `buildComposerProject2` and `mkComposerRepository2` builders
- Do not use `composer-local-repo-plugin`
- Enable `--optimize-autoloader` by default
- Enable `--apcu-autoloader` by default
- Performance improvements
- Simplified workflow
- Require:
  - #309605
  - #312231

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
- For non-Linux: Is sandboxing enabled in `nix.conf`? (See [Nix manual](https://nixos.org/manual/nix/stable/command-ref/conf-file.html))
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:
  - [NixOS test(s)](https://nixos.org/manual/nixos/unstable/index.html#sec-nixos-tests) (look inside [nixos/tests](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests))
  - and/or [package tests](https://github.com/NixOS/nixpkgs/blob/master/pkgs/README.md#package-tests)
  - or, for functions and "core" functionality, tests in [lib/tests](https://github.com/NixOS/nixpkgs/blob/master/lib/tests) or [pkgs/test](https://github.com/NixOS/nixpkgs/blob/master/pkgs/test)
  - made sure NixOS tests are [linked](https://nixos.org/manual/nixpkgs/unstable/#ssec-nixos-tests-linking) to the relevant packages
- [x] Tested compilation of all packages that depend on this change using `nix-shell -p nixpkgs-review --run "nixpkgs-review rev HEAD"`. Note: all changes have to be committed, also see [nixpkgs-review usage](https://github.com/Mic92/nixpkgs-review#usage)
- [x] Tested basic functionality of all binary files (usually in `./result/bin/`)
- [24.05 Release Notes](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2405.section.md) (or backporting [23.05](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2305.section.md) and [23.11](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2311.section.md) Release notes)
[...]

PR #308608, opened 2024-05-02
psysh: 0.11.21 -> 0.12.3 + build-support/php: update

drupol:
Some projects in PHP are running `composer` commands on their CI before building a release.

Even if the project's `composer.lock` gets shipped along the PHAR release, running `composer validate` returns and error (https://github.com/bobthecow/psysh/issues/796).

This PR allows users to run a `preBuild` hook so that extra composer commands can be triggered just before building the composer repository (https://github.com/NixOS/nixpkgs/pull/308608/commits/017b29d988f4919321e82d4e9ca0bae5542f7806)

The downside of this is that all the PHP hashes needs to be regenerated. I don't know if there's a better way to do this yet.

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/build-support/php/build-composer-project.nix
+++ b/pkgs/build-support/php/build-composer-project.nix
@@ -62,7 +62,7 @@ let
         runHook postInstallCheck
       '';
 
-      composerRepository = phpDrv.mkComposerRepository {
+      composerRepository = phpDrv.mkComposerRepository ({
         inherit composer composer-local-repo-plugin;
         inherit (finalAttrs) patches pname src vendorHash version;
 
@@ -71,7 +71,7 @@ let
         composerNoPlugins = previousAttrs.composerNoPlugins or true;
         composerNoScripts = previousAttrs.composerNoScripts or true;
         composerStrictValidation = previousAttrs.composerStrictValidation or true;
-      };
+      } // (previousAttrs.composerRepository or {}));
 
       COMPOSER_CACHE_DIR="/dev/null";
       COMPOSER_DISABLE_NETWORK="1";
--- a/pkgs/build-support/php/hooks/composer-install-hook.sh
+++ b/pkgs/build-support/php/hooks/composer-install-hook.sh
@@ -23,6 +23,10 @@ composerInstallConfigureHook() {
         cp "$composerLock" composer.lock
     fi
 
+    chmod +w composer.{json,lock}
+    cp ${composerRepository}/composer.{json,lock} .
+    chmod +w composer.{json,lock}
+
     if [[ ! -f "composer.lock" ]]; then
         setComposeRootVersion
 
@@ -70,8 +74,6 @@ composerInstallConfigureHook() {
 
         exit 1
     fi
-
-    chmod +w composer.json composer.lock
 
     echo "Finished composerInstallConfigureHook"
 }
--- a/pkgs/development/php-packages/composer/default.nix
+++ b/pkgs/development/php-packages/composer/default.nix
@@ -40,7 +40,7 @@ php.buildComposerProject (finalAttrs: {
       --prefix PATH : ${lib.makeBinPath [ _7zz cacert curl git unzip xz ]}
   '';
 
-  vendorHash = "sha256-dNNV9fTyGyRoGeDV/vBjn0aMgkaUMsrKQv5AOoiYokQ=";
+  vendorHash = "sha256-4s912b04PTPbK9+suni9uJ2Cx4qDb8CkMLy6y25a1eQ=";
 
   meta = {
     changelog = "https://github.com/composer/composer/releases/tag/${finalAttrs.version}";
--- a/pkgs/development/php-packages/psysh/default.nix
+++ b/pkgs/development/php-packages/psysh/default.nix
@@ -1,18 +1,22 @@
-{ fetchFromGitHub, lib, php }:
+{ fetchFromGitHub, fetchurl, lib, php }:
 
 php.buildComposerProject (finalAttrs: {
   pname = "psysh";
-  version = "0.11.21";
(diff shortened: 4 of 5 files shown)


pkgs/build-support/php/build-composer-project.nix
```diff
@@ -62,7 +62,7 @@ let
         runHook postInstallCheck
       '';
 
-      composerRepository = phpDrv.mkComposerRepository {
+      composerRepository = phpDrv.mkComposerRepository ({
```
jtojnar, line 65: I would use `previousAttrs.composerRepository or phpDrv.mkComposerRepository …`. This looks more like `composerRepositoryArgs` and having different behaviour when passing it to `buildComposerProject` and to `overrideAttrs` would be confusing.

pkgs/build-support/php/hooks/composer-install-hook.sh
```diff
@@ -23,6 +23,10 @@ composerInstallConfigureHook() {
         cp "$composerLock" composer.lock
     fi
 
+    chmod +w composer.{json,lock}
+    cp ${composerRepository}/composer.{json,lock} .
```
jtojnar, line 27: This makes the two lock checks below redundant.

pkgs/development/php-packages/composer/default.nix
```diff
@@ -40,7 +40,7 @@ php.buildComposerProject (finalAttrs: {
       --prefix PATH : ${lib.makeBinPath [ _7zz cacert curl git unzip xz ]}
   '';
 
-  vendorHash = "sha256-dNNV9fTyGyRoGeDV/vBjn0aMgkaUMsrKQv5AOoiYokQ=";
+  vendorHash = "sha256-4s912b04PTPbK9+suni9uJ2Cx4qDb8CkMLy6y25a1eQ=";
```
jtojnar, line 43: Yeah, this is bad. If we want to not break BC for outside consumers, we would need to introduce `mkComposerRepository2` and `buildComposerProject2`, and deprecate the old one.

pkgs/development/php-packages/psysh/default.nix
```diff
@@ -16,7 +16,18 @@ php.buildComposerProject (finalAttrs: {
...
     hash = "sha256-ur6mzla3uXeFL6aEHAPdpxGdvcgzOgTLW/CKPbNqeCg=";
   };
-  vendorHash = "";
+
+  composerRepository = {
+    preBuild = ''
+      setComposeRootVersion
+      composer config platform.php 7.4
+      composer require --no-update symfony/polyfill-iconv symfony/polyfill-mbstring
+      composer require --no-update --dev roave/security-advisories:dev-latest
```
jtojnar, line 25: Actually, this is not reproducible, is it? The hash would change over time since the lockfile is included in the output.

pkgs/development/php-packages/psysh/default.nix
```diff
@@ -16,7 +16,18 @@ php.buildComposerProject (finalAttrs: {
     url = "https://github.com/bobthecow/psysh/releases/download/v${finalAttrs.version}/composer-v${finalAttrs.version}.lock";
     hash = "sha256-ur6mzla3uXeFL6aEHAPdpxGdvcgzOgTLW/CKPbNqeCg=";
   };
-  vendorHash = "";
+
+  composerRepository = {
+    preBuild = ''
```
jtojnar, line 21: Hmm, maybe we should use `make build/psysh` then?

pkgs/development/php-packages/psysh/default.nix
```diff
@@ -16,7 +16,18 @@ php.buildComposerProject (finalAttrs: {
     url = "https://github.com/bobthecow/psysh/releases/download/v${finalAttrs.version}/composer-v${finalAttrs.version}.lock";
     hash = "sha256-ur6mzla3uXeFL6aEHAPdpxGdvcgzOgTLW/CKPbNqeCg=";
   };
-  vendorHash = "";
+
+  composerRepository = {
```
jtojnar, line 20: Yeah, this is confusing. I would expect this attribute to be passed the result of `mkcomposerRepository`.

If you apply the suggested change to `buildComposerProjectOverride`, you should be able to use `composerRepository = previousAttrs.composerRepository.overrideAttrs { preBuild = …; };`

pkgs/development/php-packages/psysh/default.nix
drupol: No, this is fully reproducible actually, tested with `--rebuild`.

pkgs/development/php-packages/psysh/default.nix
drupol: `make` ? What does make has anything to do in here?

pkgs/development/php-packages/psysh/default.nix
drupol: Yeah I'm not yet at ease with this.... I will do the change.

pkgs/development/php-packages/psysh/default.nix
jtojnar: What if `roave/security-advisories` releases a new version?

pkgs/development/php-packages/psysh/default.nix
jtojnar: You are mentioning it in the commit message.

drupol:
superseded by https://github.com/NixOS/nixpkgs/pull/308059

PR #308700, opened 2024-05-03
nixos/pgbouncer: add services.pgbouncer.settings option

tie:
## Description of changes

This change adds services.pgbouncer.settings option as per [RFC 0042] and removes other options that were previously used to generate configuration file.

In addition to that, we also place the configuration file under environment.etc to allow reloading configuration without service restart.

[RFC 0042]: https://github.com/NixOS/rfcs/blob/master/rfcs/0042-config-option.md

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [x] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [x] Tested, as applicable:

---

### DIFF
--- a/nixos/modules/services/databases/pgbouncer.nix
+++ b/nixos/modules/services/databases/pgbouncer.nix
@@ -1,543 +1,138 @@
-{ lib, pkgs, config, ... } :
-
-with lib;
-
+{ config, lib, utils, pkgs, ... }:
 let
   cfg = config.services.pgbouncer;
 
-  confFile = pkgs.writeTextFile {
-    name = "pgbouncer.ini";
-    text =  ''
-      [databases]
-      ${concatStringsSep "\n"
-      (mapAttrsToList (dbname : settings : "${dbname} = ${settings}") cfg.databases)}
-
-      [users]
-      ${concatStringsSep "\n"
-      (mapAttrsToList (username : settings : "${username} = ${settings}") cfg.users)}
-
-      [peers]
-      ${concatStringsSep "\n"
-      (mapAttrsToList (peerid : settings : "${peerid} = ${settings}") cfg.peers)}
-
-      [pgbouncer]
-      # general
-      ${optionalString (cfg.ignoreStartupParameters != null) "ignore_startup_parameters = ${cfg.ignoreStartupParameters}"}
-      listen_port = ${toString cfg.listenPort}
-      ${optionalString (cfg.listenAddress != null) "listen_addr = ${cfg.listenAddress}"}
-      pool_mode = ${cfg.poolMode}
-      max_client_conn = ${toString cfg.maxClientConn}
-      default_pool_size = ${toString cfg.defaultPoolSize}
-      max_user_connections = ${toString cfg.maxUserConnections}
-      max_db_connections = ${toString cfg.maxDbConnections}
-
-      #auth
-      auth_type = ${cfg.authType}
-      ${optionalString (cfg.authHbaFile != null) "auth_hba_file = ${cfg.authHbaFile}"}
-      ${optionalString (cfg.authFile != null) "auth_file = ${cfg.authFile}"}
-      ${optionalString (cfg.authUser != null) "auth_user = ${cfg.authUser}"}
-      ${optionalString (cfg.authQuery != null) "auth_query = ${cfg.authQuery}"}
-      ${optionalString (cfg.authDbname != null) "auth_dbname = ${cfg.authDbname}"}
-
-      # TLS
-      ${optionalString (cfg.tls.client != null) ''
-      client_tls_sslmode = ${cfg.tls.client.sslmode}
-      client_tls_key_file = ${cfg.tls.client.keyFile}
-      client_tls_cert_file = ${cfg.tls.client.certFile}
-      client_tls_ca_file = ${cfg.tls.client.caFile}
-      ''}
-      ${optionalString (cfg.tls.server != null) ''
-      server_tls_sslmode = ${cfg.tls.server.sslmode}
-      server_tls_key_file = ${cfg.tls.server.keyFile}
-      server_tls_cert_file = ${cfg.tls.server.certFile}
-      server_tls_ca_file = ${cfg.tls.server.caFile}
-      ''}
-
-      # log
-      ${optionalString (cfg.logFile != null) "logfile = ${cfg.homeDir}/${cfg.logFile}"}
-      ${optionalString (cfg.syslog != null) ''
-      syslog = ${if cfg.syslog.enable then "1" else "0"}
-      syslog_ident = ${cfg.syslog.syslogIdent}
-      syslog_facility = ${cfg.syslog.syslogFacility}
-      ''}
-      ${optionalString (cfg.verbose != null) "verbose = ${toString cfg.verbose}"}
-
-      # console access
-      ${optionalString (cfg.adminUsers != null) "admin_users = ${cfg.adminUsers}"}
-      ${optionalString (cfg.statsUsers != null) "stats_users = ${cfg.statsUsers}"}
-
-      # extra
-      ${cfg.extraConfig}
-    '';
-  };
-
-in {
+  settingsFormat = pkgs.formats.ini { };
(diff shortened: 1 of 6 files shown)


tie:
W.r.t. `notify-reload` vs `notify`:

Systemd docs: https://www.freedesktop.org/software/systemd/man/latest/systemd.service.html#:~:text=Behavior%20of%20notify-reload,reloading%20of%20the%20service's%20configuration.

Relevant pgbouncer code: https://github.com/pgbouncer/pgbouncer/blob/6c629ab98e8c06aec05feda49019908693121755/src/main.c#L546-L554

tie:
@wegank, sorry for pinging you directly, but since we already have some potentially breaking changes in `pgbouncer` module configuration for 24.05 release, is it possible to get this PR reviewed and hopefully merged before release? Or it’s already too late in 24.05 release schedule?

wegank:
Sounds too late for me. Also @Mic92.

tie:
OK, that’s fine, thanks, just wanted to confirm.

nixos/doc/manual/release-notes/rl-2405.section.md
```diff
@@ -412,7 +412,7 @@ The pre-existing [services.ankisyncd](#opt-services.ankisyncd.enable) has been m
   `wants`), because the dependency that `multi-user.target` has on
   `network-online.target` is planned for removal.
 
-- `services.pgbouncer` now has systemd support enabled and will log to journald. The default setting for `services.pgbouncer.logFile` is now `null` to disable logging to a separate log file.
+- `services.pgbouncer` now has systemd support enabled and will log to journald. The systemd service is configured with `Type=notify-reload` and allows reloading configuration without process restart. PgBouncer configuration options were moved to the free-form type option named [`services.pgbouncer.settings`](#opt-services.pgbouncer.settings) according to the NixOS RFC 0042.
```
SuperSandro2000, line 415:
```suggestion
- `services.pgbouncer` now has systemd support enabled and will log to journald.
  The systemd service is configured with `Type=notify-reload` and allows reloading configuration without process restart.
  PgBouncer configuration options were moved to the free-form type option named [`services.pgbouncer.settings`](#opt-services.pgbouncer.settings) according to the NixOS RFC 0042.
```

nixos/modules/services/monitoring/prometheus/exporters/pgbouncer.nix
```diff
@@ -116,35 +120,24 @@ in
...
-            ++ optionals (cfg.webConfigFile != null) [
-            "--web.config.file ${escapeShellArg cfg.webConfigFile}"
-          ]
-            ++ cfg.extraFlags)}";
-      in
-      {
-        ExecStart = "${startScript}/bin/pgbouncer-start";
-      };
+    script = concatStringsSep " " ([
+      "${pkgs.prometheus-pgbouncer-exporter}/bin/pgbouncer_exporter"
```
SuperSandro2000, line 124:
```suggestion
      "${lib.getExe pkgs.prometheus-pgbouncer-exporter}"
```

nixos/doc/manual/release-notes/rl-2405.section.md
tie: 👍
Also we should probably wait for 24.11 release notes file to be initialized, see above w.r.t. release schedule.

nixos/modules/services/monitoring/prometheus/exporters/pgbouncer.nix
tie: Done. Also added package option.

nixos/modules/services/databases/pgbouncer.nix
```diff
@@ -1,543 +1,58 @@
...
-    };
-
-    # Authentication settings
-    authType = mkOption {
-      type = types.enum [ "cert" "md5" "scram-sha-256" "plain" "trust" "any" "hba" "pam" ];
-      default = "md5";
+    settings = lib.mkOption {
+      type = settingsFormat.type;
+      default = { };
       description = ''
```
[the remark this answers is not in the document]
tie: How do we decide which settings deserve to show up in search? In addition to that, I assume that people that need PgBouncer are actually familiar with all the options, and non-NixOS-specific options add a lot of noise to the search results (at least that is the case for me in current pgbouncer module implementation).

nixos/modules/services/databases/pgbouncer.nix
```diff
@@ -1,543 +1,58 @@
...
-  };
-
-in {
+  settingsFormat = pkgs.formats.ini { };
+  configFile = settingsFormat.generate "pgbouncer.ini" cfg.settings;
+  configPath = "pgbouncer/pgbouncer.ini";
+in
+{
+  imports = [
+    (lib.mkRemovedOptionModule [ "services" "pgbouncer" "logFile" ] "Use services.pgbouncer.settings instead.")
```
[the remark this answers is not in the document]
tie: Done.

nixos/doc/manual/release-notes/rl-2405.section.md
tie: Moved to 24.11 release notes.

Ma27 (commented):
I must say, I don't like this change very much.

RFC42 encourages to use freeform modules, i.e. predefined options for type-checking and defaults in e.g. the manual.

Even worse, two pitfalls when configuring pgbouncer that were previously caught with assertions are now gone and people will find out at runtime.

All of taht can be added in an RFC42-compliant module, but for me, the current state makes the module worse than it was before.

tie:
>Even worse, two pitfalls when configuring pgbouncer that were previously caught with assertions are now gone and people will find out at runtime.

If you are talking about pgbouncer-exporter assertions in prometheus module, I think these are a bad design. The evaluated configuration is not necessarily the one being monitored. It’s OK to delay some issues until run time if we can’t check for them at evaluation time (or it’s not possible to do that efficiently).

>RFC42 encourages to use freeform modules, i.e. predefined options for type-checking and defaults in e.g. the manual.

That’s not how I understand [RFC42 Part 2](https://github.com/NixOS/rfcs/blob/master/rfcs/0042-config-option.md#part-2-1), but I may be wrong here. I agree that we should add common predefined settings for the pgbouncer module. I didn’t do that initially (https://github.com/NixOS/nixpkgs/pull/308700#discussion_r1600285914), but I’ll add these in a subsequent PR soon, in particular, `listen_addr`, `listen_port` and `unix_socket_dir` (since we are using these in the configuration and these are the most basic options for pgbouncer setup).

nixos/modules/services/databases/pgbouncer.nix
```diff
@@ -1,543 +1,138 @@
...
+  imports = [
+    (lib.mkRenamedOptionModule
+      [ "services" "pgbouncer" "logFile" ]
+      [ "services" "pgbouncer" "settings" "pgbouncer" "log_file" ])
+    (lib.mkRenamedOptionModule
+      [ "services" "pgbouncer" "listenAddress" ]
+      [ "services" "pgbouncer" "settings" "pgbouncer" "listen_addr" ])
+    (lib.mkRenamedOptionModule
+      [ "services" "pgbouncer" "listenPort" ]
+      [ "services" "pgbouncer" "settings" "pgbouncer" "listen_port" ])
```
Ma27, line 19: This gives me

```
error: evaluation aborted with the following error message: 'Renaming error: option `services.pgbouncer.settings.pgbouncer.listen_port' does not exist.'
```

btw.
Keeping options around is also useful for a painless upgrade path.

Ma27:
Opened #356965.

omares:
Hey, it seems that pgbouncer handles reloads differently. This causes the daemon to hang indefinitely. You can see more details in issue #507 on GitHub: https://github.com/pgbouncer/pgbouncer/issues/507.

Re-adding the following code to my `systemd.services.pgbouncer` definition resolves the issue:

``` 
serviceConfig = {
    Type = "notify-reload";
    ExecReload = "${pkgs.coreutils}/bin/kill -HUP $MAINPID";
}; 
```

Maybe the removal of the change should be reverted?

PR #308703, opened 2024-05-03
nixos/nomad: change dropPrivileges default to false

tylerjl:
## Description of changes

Rootless nomad is more appealing from an operational security perspective, but upstream indicates that rootless nomad is _not_ yet a supported mode of operation an may exhibit unexpected behavior. This change keeps the option available, but changes the default to [what upstream recommends][1].

I came across this because after updating my nomad deployment, the Docker driver suddenly stopped working, and this was the only way I was able to resolve it (despite ensuring appropriate group membership, correct cgroups, etc.). My deployment kept working after making the change in-place to an existing deployment, so it appears like the systemd `StateDirectory=` behavior (ensuring ownership is consistent with the `.service` unit's designed owner) is working as intended to make sure that the change will work with existing installations. (Annoyingly, upstream source code actually forcibly _disables_ the Docker driver entirely unless EUID is 0).

I haven't documented a related release note for this, since I wanted to consult with the folks marked as maintainers for the nomad package before committing with the associated release note changelog. But I definitely think it deserves a release note if we do make the change since it changes a default.

[1]: https://github.com/hashicorp/nomad/issues/13669

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
- ~For non-Linux: Is sandboxing enabled in `nix.conf`? (See [Nix manual](https://nixos.org/manual/nix/stable/command-ref/conf-file.html))~
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:
  - ~(Package updates) Added a release notes entry if the change is major or breaking~
  - ~(Module addition) Added a release notes entry if adding a new NixOS module~

---

### DIFF
--- a/nixos/modules/services/networking/nomad.nix
+++ b/nixos/modules/services/networking/nomad.nix
@@ -25,7 +25,7 @@ in
 
       dropPrivileges = mkOption {
         type = types.bool;
-        default = true;
+        default = false;
         description = ''
           Whether the nomad agent should be run as a non-root nomad user.
         '';


tylerjl:
Rebased onto latest `master` since this was lagging behind.

I wanted to let this sit before direct mentioning other maintainers but this is probably the right time to loop them in, so /cc @Cottand @rushmorem @techknowlogick 

For those folks, the [original PR comment](https://github.com/NixOS/nixpkgs/pull/308703#issue-2276846075) explains the reasoning here.

Cottand:
Hi @tylerjl

I understand where this comes from, I had a similar experience when I first deployed Nomad on NixOS. I am not sure about changing this, my reasoning follows:

You mention running as root is what upstream recommends. This is only partially true: Nomad clients should be run as root, but [Nomad servers should not](https://developer.hashicorp.com/nomad/docs/operations/nomad-agent#permissions).
> Nomad servers should be run with the lowest possible permissions. They need access to their own data directory and the ability to bind to their ports. You should create a nomad user with the minimal set of required privileges.

So when deploying a server, you would actually prefer `dropPrivileges = false`. This makes either setting 'wrong' in some scenario! So I do not think there is a single right choice here.

I did not create this option myself (I became a maintainer later) but if faced with this trade-off for the first time, I think I would make the same decision and choose least-privileges to be the default.

On top of this, changing the option would make existing Nomad servers that rely on this default less secure relatively silently.

Let me know what you think! Happy to see if the other maintainers have different opinions too.

tylerjl:
@Cottand ah, I had no idea upstream guidance was to drop privileges for server instances. I truly have no other interest here other than to make it less frustrating for operators so I'm not overly-partial to either option, just whichever is best for everyone.

I'd also agree that more-secure-by-default is probably the better option. How about if we add an assertion and warning if _both_ `dropPrivileges` **and** `enableDocker` are set? (or, possibly, if `client.enabled` is set?) It's a fairly benign change that won't force a correct configuration by construction but would point a user in the right direction.

I'm not 100% sure about what the right predicate logic for an assertion and warning might be. Possibly:

- If `enableDocker` **and** `dropPrivileges` are true, maybe we hard-fail? As my previous comment mentioned, nomad has a hardcoded Docker driver failure if EUID=0, so it's going to fail anyway.
- If `client.enabled=true` **and** `dropPrivileges`, maybe we just issue a warning given upstream's guidance about client operation. The nomad module adheres to the `settings` as attribute set pattern so we can reliably read out attributes from there.

Cottand:
I am happy to issue a warning if both `services.nomad.enableDocker = true` and `services.nomad.dropPrivileges = false`. I could go either way on an assertion.

I would prefer not to rely on `client.enableDocker` because it is also possible to set this option outside of NixOS via [`services.nomad.extraSettingsPaths`](https://search.nixos.org/options?channel=unstable&show=services.nomad.enableDocker&from=0&size=50&sort=relevance&type=packages&query=nomad), making any warnings potential false positives.

PR #307562, opened 2024-04-28
switch-to-configuration.pl: Fix timers resulting in old unit running

nh2:
Fixes #49415.

See the added comment for a detail description.

See also the related #49528.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
- For non-Linux: Is sandboxing enabled in `nix.conf`? (See [Nix manual](https://nixos.org/manual/nix/stable/command-ref/conf-file.html))
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:
  - [NixOS test(s)](https://nixos.org/manual/nixos/unstable/index.html#sec-nixos-tests) (look inside [nixos/tests](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests))
  - and/or [package tests](https://github.com/NixOS/nixpkgs/blob/master/pkgs/README.md#package-tests)
  - or, for functions and "core" functionality, tests in [lib/tests](https://github.com/NixOS/nixpkgs/blob/master/lib/tests) or [pkgs/test](https://github.com/NixOS/nixpkgs/blob/master/pkgs/test)
  - made sure NixOS tests are [linked](https://nixos.org/manual/nixpkgs/unstable/#ssec-nixos-tests-linking) to the relevant packages
- [ ] Tested compilation of all packages that depend on this change using `nix-shell -p nixpkgs-review --run "nixpkgs-review rev HEAD"`. Note: all changes have to be committed, also see [nixpkgs-review usage](https://github.com/Mic92/nixpkgs-review#usage)
- [ ] Tested basic functionality of all binary files (usually in `./result/bin/`)
- [24.05 Release Notes](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2405.section.md) (or backporting [23.05](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2305.section.md) and [23.11](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2311.section.md) Release notes)
  - [ ] (Package updates) Added a release notes entry if the change is major or breaking
  - [ ] (Module updates) Added a release notes entry if the change is significant
  - [ ] (Module addition) Added a release notes entry if adding a new NixOS module
[...]

Issue #301014, opened 2024-04-02
[Tracking issue] remove all `src = ./.` entries that unintentionally include their own `.nix` files


philiptaron:
## Issue description

Derivations that use `src = ./.` will be rebuilt on any change to the `.nix` file that includes them, including formatting changes caused by `nixfmt-rfc-style` for [RFC 166](https://github.com/NixOS/rfcs/blob/master/rfcs/0166-nix-formatting.md).

### List of files where this may be the case

Generated with `rg -F 'src = ./.' --sort=path -l -tnix` and not very thoroughly audited. Unchecked entries need work or attention.

- [x] [doc/default.nix](https://github.com/NixOS/nixpkgs/blob/master/doc/default.nix#L102): https://github.com/NixOS/nixpkgs/pull/319875
- [x] [maintainers/scripts/convert-to-import-cargo-lock/default.nix](https://github.com/NixOS/nixpkgs/blob/master/maintainers/scripts/convert-to-import-cargo-lock/default.nix#L7) (uses `cleanSourceWith`)
- [x] [nixos/lib/test-driver/default.nix](https://github.com/NixOS/nixpkgs/blob/master/nixos/lib/test-driver/default.nix#L20): https://github.com/NixOS/nixpkgs/pull/319878
- [x] [pkgs/build-support/node/fetch-npm-deps/default.nix](https://github.com/NixOS/nixpkgs/blob/master/pkgs/build-support/node/fetch-npm-deps/default.nix#L9) (uses `cleanSourceWith`)
- [x] [pkgs/build-support/release/default.nix](https://github.com/NixOS/nixpkgs/blob/master/pkgs/build-support/release/default.nix#L99) (only in a comment)
- [x] [pkgs/build-support/rust/test/import-cargo-lock/basic/default.nix](https://github.com/NixOS/nixpkgs/blob/master/pkgs/build-support/rust/test/import-cargo-lock/basic/default.nix#L7) #303146
- [x] [pkgs/build-support/rust/test/import-cargo-lock/basic-dynamic/default.nix](https://github.com/NixOS/nixpkgs/blob/master/pkgs/build-support/rust/test/import-cargo-lock/basic-dynamic/default.nix#L7) #303146
- [x] [pkgs/build-support/rust/test/import-cargo-lock/git-dependency/default.nix](https://github.com/NixOS/nixpkgs/blob/master/pkgs/build-support/rust/test/import-cargo-lock/git-dependency/default.nix#L7) #303146
[...]

PR #270727, opened 2023-11-28
Perlless Activation

nikstur:
Introduces the bits and pieces necessary to build a fully perlless system (i.e.  a system without the perl interpreter.) 

## Acknowledgement

Even if this is not necessarily reflected in the commits or code, this was a team effort. Thank you for your help!

Started at OceanSprint 2023 with bootstrapping help from @lheckemann and @blitz.

Mounting `/etc` via an overlayfs was @arianvp's original idea. @RaitoBezarius pointed out that we can use composefs to get modes and permissions into this overlay.

This is also part of the broader [Boot Security Work](https://github.com/NixOS/nixpkgs/issues/265640)

## Description of changes

Design Document: https://pad.lassul.us/nixos-perlless-activation

This change consists of three parts:

1. Introduce the ability to create users and groups with systemd-sysusers instead of via `update-users-groups.pl`
2. Introduce the ability to mount `/etc` via an overlay (using `mkcomposefs` and other ideas stolen from composefs)
3. Add the perlless profile, tying this all together and proving that you can have a fully perlless system.

Furthermore it depends on two more PRs: 
- https://github.com/NixOS/nixpkgs/pull/267983
- https://github.com/NixOS/nixpkgs/pull/271067

The commits from these PRs are also contained in this PR and are marked with `rebase:`.

This change doesn't remove any of the current mechanisms to create users and `/etc` but only adds opt-in mechanism do it all without perl. I see this as the only way to introduce the new functionality. Straight up replacing the existing ones is too risky as test coverage is generally poor (or very depedent on the implementation) and this is a critical path. This way we can start live testing these features and incrementally improving them until they are mature enough to fully replace the existing mechanisms.

I'm pretty confident that you can use the new mechanisms even for a normal switchable system (but then you will have Perl because of switch-to-configuration.pl!). 

Closes #267982

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
- For non-Linux: Is sandboxing enabled in `nix.conf`? (See [Nix manual](https://nixos.org/manual/nix/stable/command-ref/conf-file.html))
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [x] Tested, as applicable:
[...]

PR #308801, opened 2024-05-03
nixos/switch-to-configuration: add new implementation

jmbaur:
## Description of changes

This adds an implementation of switch-to-configuration that allows for closer interaction with the lifecycle of systemd units by using DBus APIs directly instead of using systemctl. It is disabled by default, but can be enabled by specifying `{ system.switch = { enable = false; enableNg = true; }; }`.

The goal of this change is to provide an implementation of switch-to-configuration that is compatible (bug-for-bug, feature-for-feature) with the current perl script. This means that it is expected that all current NixOS VM tests under `./nixos/tests` will pass without any of the test scripts being modified. Assuming the changes here land, hopefully this implementation is found to be easier to maintain and more testable, allowing for easier modification of it in the future and eventually deprecation of the perl script. 

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [x] Tested, as applicable:

---


jmbaur:
@ofborg test switchTestNg

dasJ:
Stellar work, thank you! I haven't looked at the Rust code in detail yet but scrolling over it quickly looked great.

One thing to keep in mind is the existence of #307562.

Also, pretty much the last thing I wanted to do with Perl STC (apart from getting the sockets to work) was to port the fstab handling to the same logic the .mount unit uses.
The reason is that the .mount handling is newer and uses more sophisticated (if you can call it that) logic to decide on whether to reload or restart.
While this would make both implementations easier (probably only the Rust one since I don't think I will find the energy to factor out a new subroutine in Perl), this would also get my unfinished test cases for fstab handling to nixpkgs. This way, we could make sure that this implementation doesn't introduce regressions and also that it works the exact same as the Perl one.
What do you think about that? The only obvious downside I can see is that this would change the existing stc and I don't think we should rush this before 24.05 branch-off.

jmbaur:
> One thing to keep in mind is the existence of #307562.
I did see that issue. I haven't tested against https://github.com/NixOS/nixpkgs/pull/307562#issuecomment-2087015893 yet, but I'll do so shortly and paste the results. I started this PR as a 1-to-1 port of the perl code, so I'm "hoping" that this issue is also present in this PR, then they can be addressed together.

> port the fstab handling to the same logic the .mount unit uses.
That sounds great, I believe the logic was if any of the mount options change, reload the unit, otherwise restart it?

> I don't think we should rush this before 24.05 branch-off.
Definitely no rush

nixos/modules/system/activation/switch-to-configuration-ng/Cargo.toml
SuperSandro2000, line 1: Packages should be under pkgs. Also then it can be easily overlayed.

nixos/modules/system/activation/switchable-system.nix
```diff
@@ -4,52 +4,95 @@ let
...
-        fi
-      ''}
-    '';
+    enableNg = lib.mkOption {
+      type = lib.types.bool;
+      default = false;
+      description = description ''
+        `enableNg` enables an experimental implementation of
+        `switch-to-configuration` with the goal of replacing the original
+        implementation.
```
SuperSandro2000, line 36:
```suggestion
        Wether to use `switch-to-cofniguration-ng`, an experimental re-implementation of
        `switch-to-configuration`, with the goal of replacing the original.
```

nixos/modules/system/activation/switchable-system.nix
```diff
@@ -4,52 +4,95 @@ let
...
+          mkdir $out/bin
+          ln -sf ${lib.getExe config.system.build.switchToConfigurationNg} $out/bin/switch-to-configuration
+          wrapProgram $out/bin/switch-to-configuration \
+            --set OUT $out \
+            --set TOPLEVEL ''${!toplevelVar} \
+            --set DISTRO_ID ${lib.escapeShellArg config.system.nixos.distroId} \
+            --set INSTALL_BOOTLOADER ${lib.escapeShellArg config.system.build.installBootLoader} \
+            --set LOCALE_ARCHIVE ${config.i18n.glibcLocales}/lib/locale/locale-archive \
+            --set SYSTEMD ${config.systemd.package} \
+            --set UTILLINUX ${pkgs.util-linux} \
```
SuperSandro2000, line 92:
```suggestion
            --set UTILLINUX ${pkgs.util-linux}
```
also we can probably hardcode util-linux into the package

nixos/modules/system/activation/switch-to-configuration-ng/default.nix
```diff
@@ -0,0 +1,27 @@
...
+  dbus,
+  pkg-config,
+  systemd,
+}:
+
+rustPlatform.buildRustPackage {
+  pname = "switch-to-configuration";
+  version = "0.1.0";
+
+  src = ./.;
```
SuperSandro2000, line 13: please see #301014

nixos/modules/system/activation/switch-to-configuration-ng/src/main.rs
SuperSandro2000, line 1: We should definitely split this file into multiple smaller ones.

nixos/modules/system/activation/switch-to-configuration-ng/default.nix
jmbaur: Thanks! Using lib.fileset now

nixos/modules/system/activation/switchable-system.nix
jmbaur: Implemented to not need util-linux at all anymore, thanks!

nixos/modules/system/activation/switch-to-configuration-ng/Cargo.toml
jmbaur: Moved to be under `pkgs/by-name/sw/switch-to-configuration-ng`

jmbaur:
@ofborg test switchTestNg

alyssais:
We need to be very careful about rewriting fundamental parts of NixOS in Rust, because we're going to break NixOS on platforms that Rust doesn't support, but do have some committed NixOS users, like MIPS, which had a lot of work put into it over the last couple of years.  (Most scripting languages don't have this problem, so something like Python might be a better fit, but compiled languages, especially those using LLVM, will.)

jmbaur:
@alyssais totally agree, though I would assume LLVM would give us great platform support? After fixing the derivation (using dbus interface definitions from `buildPackages.systemd` instead of `systemd`), I was able to cross-compile this implementation using `pkgsCross.mips64el-linux-gnuabi64.switch-to-configuration-ng`, were there other platforms of concern?

alyssais:
Rust supports cross-compilation to more platforms than it supports running the compiler on, but the latter is what would be important for switch-to-configuration.  Otherwise, we'd be completely removing the ability to switch configurations on those systems, since that system wouldn't be cross-compiled.

So systems supported by Nixpkgs that would lose out on being able to self-host NixOS on this change would be:

```
nix-repl> lib.subtractLists rustc.meta.platforms lib.platforms.linux
[
  "armv5tel-linux"
  "armv7a-linux"
  "m68k-linux"
  "microblaze-linux"
  "microblazeel-linux"
  "mips-linux"
  "mips64-linux"
  "mips64el-linux"
  "mipsel-linux"
  "riscv32-linux"
  "s390-linux"
]
```

Some of these, like microblaze and s390, likely can't even build all of NixOS anyway.  MIPS definitely has actual users (one use case I think is NixOS on a router, but I don't really know) and I'm pretty sure armv7 does too, although that might be armv7l and it wouldn't surprise me if most of those systems were cross-compiled.

jmbaur:
>  MIPS definitely has actual users (one use case I think is NixOS on a router, but I don't really know) and I'm pretty sure armv7 does too, although that might be armv7l and it wouldn't surprise me if most of those systems were cross-compiled.

I run nixos on my armv7 router :) though I will admit I cross compile to it, as just doing an eval on it is too expensive.

I guess it's important to narrow down the focus to the list of users that run on the platforms listed above that _don't_ cross compile their systems, since users that do cross-compile would be fine. While I would personally be surprised if users were doing native builds on many of those platforms listed above, I don't have much knowledge outside of the arm and x86 space and I definitely don't want to make any arbitrary limits on the platforms nixos can be built from.

alyssais:
I think MIPS64 is the main one to worry about.

dasJ:
Is this even something to worry about? Afaik, NixOS has no official MIPS64 support and for these platforms I honestly think it would be okay to require cross-compilation.

Also, while the new implementation is there now, it's still opt-in and I don't see why we wouldn't keep the old variant around for a bit after it has be come the default. This would also give the LLVM people more time (years probably) to fix the MIPS targets if they even intend doing so.

alyssais:
It depends whether this would break anything for anybody in practice.  I've drawn attention to this PR in places where users using those platforms are likely to be reachable (#exotic:nixos.org, #networking:nixos.org), and nobody has come forward, so it seems like that's a no, and so that's fine.  It's just important to make sure of that when making this kind of decision.

flokli:
As written a bunch of times above, it's probably fine to introduce this additionally, even flip the default at some point. rustc platform support is gonna catch up at some point, and we can still keep the perl script around for these usecases - if people for these architectures really need it, they at some point need to step up with maintainership of the perl flavour.

Having the activation script in something not-perl is such a long-term quality of life improvement, and empowers more people to fix it, so I'm heavily in favor of this. Thanks :heart:

flokli (approved):

pluiedev (approved):
Less Perl in NixOS is definitely the way to go forward — following the footsteps of #270727 — and I'm excited to see how this will turn out!

flokli:
Let's get this in - this is behind an explicit feature flag, so a no-op for anyone not explicitly enabling it.

nyabinary:
Using ``networking.useNetworkd`` with this makes the switch hang up on ``systemd-networkd-wait-online.service``.
```
restarting sysinit-reactivation.target
the following new units were started: sysinit-reactivation.target, systemd-tmpfiles-resetup.service
warning: the following units failed: systemd-networkd-wait-online.service
× systemd-networkd-wait-online.service - Wait for Network to be Configured
     Loaded: loaded (/etc/systemd/system/systemd-networkd-wait-online.service; enabled; preset: enabled)
    Drop-In: /nix/store/nlj3x6lpwgvkmnpj5zd8k7z6ynrgr5q1-system-units/systemd-networkd-wait-online.service.d
             └─overrides.conf
     Active: failed (Result: exit-code) since Thu 2024-05-16 09:35:39 EDT; 266ms ago
       Docs: man:systemd-networkd-wait-online.service(8)
    Process: 5664 ExecStart=/nix/store/9cxd17xnmw0bi8n4nf722ysqj2bjlh8s-systemd-255.4/lib/systemd/systemd-networkd-wait-online --timeout=120 (code=exited, status=1/FAILURE)
   Main PID: 5664 (code=exited, status=1/FAILURE)
         IP: 0B in, 0B out
        CPU: 4ms

May 16 09:33:39 nyan systemd[1]: Starting Wait for Network to be Configured...
May 16 09:35:39 nyan systemd-networkd-wait-online[5664]: Timeout occurred while waiting for network connectivity.
May 16 09:35:39 nyan systemd[1]: systemd-networkd-wait-online.service: Main process exited, code=exited, status=1/FAILURE
May 16 09:35:39 nyan systemd[1]: systemd-networkd-wait-online.service: Failed with result 'exit-code'.
May 16 09:35:39 nyan systemd[1]: Failed to start Wait for Network to be Configured.
warning: error(s) occurred while switching to the new configuration
```

samhug:
> Using `networking.useNetworkd` with this makes the switch hang up on `systemd-networkd-wait-online.service`.

I think the perl implementation also hangs if it has to wait for `systemd-networkd-wait-online.service` to timeout?

I include the following in my config to make the service log which interface it's waiting on:
```
systemd.services.systemd-networkd-wait-online.environment.SYSTEMD_LOG_LEVEL = "debug";
```

nyabinary:
> > Using `networking.useNetworkd` with this makes the switch hang up on `systemd-networkd-wait-online.service`.
> 
> I think the perl implementation also hangs if it has to wait for `systemd-networkd-wait-online.service` to timeout?
> 
> I include the following in my config to make the service log which interface it's waiting on:
> 
> ```
> systemd.services.systemd-networkd-wait-online.environment.SYSTEMD_LOG_LEVEL = "debug";
> ```

Not for me, I tested it on both and the switch happened on the Perl implementation but hangs on this implementation.

jmbaur:
Made an issue [here](https://github.com/NixOS/nixpkgs/issues/312297)

nh2:
@jmbaur 

First: Great. I think the Perl-written key parts of NixOS are some of the weakest parts of it: Very difficult to read/understand, very easy to footgun.

Now just [`install-grub.pl`](https://github.com/NixOS/nixpkgs/blob/b1eff03c35aa7c90ab3a4d9f6ef297dae5fba37b/nixos/modules/system/boot/loader/grub/install-grub.pl) has to go!

Next:

I recommend to make the following improvements to make this whole thing more likely to succeed:

* Add a `README.md` to [`pkgs/by-name/sw/switch-to-configuration-ng/`](https://github.com/NixOS/nixpkgs/tree/e1c79563c7a864b86135b77c2d96b598021ea6df/pkgs/by-name/sw/switch-to-configuration-ng) that explains the project's goals, design, and usage instructions. I had to dig up this PR to figure out what the hell `switch-to-configuration-ng` is that I found mentioned in some newer issue.
* Create a NixOS maintainers team for people who are interested in this, which can be pinged on issues.
* Add to the beginning of [`switch-to-configuration.pl`](https://github.com/NixOS/nixpkgs/blob/b1eff03c35aa7c90ab3a4d9f6ef297dae5fba37b/nixos/modules/system/activation/switch-to-configuration.pl) a comment:

  ```perl
  # Note there is a new alternative implementation "switch-to-configuration-ng"
  # that aims to be compatible with this one for now.
  # If you make changes here, consider making them there as well or
  # let the maintainers (INSERT TEAM NAME CREATED ABOVE), so they can mirror them.
  # While that's not required, it would certainly be helpful.
  ```

pluiedev:
> Now just [`install-grub.pl`](https://github.com/NixOS/nixpkgs/blob/b1eff03c35aa7c90ab3a4d9f6ef297dae5fba37b/nixos/modules/system/boot/loader/grub/install-grub.pl) has to go!

Well, I just made a WIP port of the script (also in Rust) in one (very productive) Saturday: https://github.com/pluiedev/install-grub
I'll try to figure out if I can get some inputs to test this script with tomorrow, which could be a bit challenging since I don't use GRUB myself ;)

jmbaur:
@nh2 Thanks for the guidance! Made a follow up PR [here](https://github.com/NixOS/nixpkgs/pull/317623)

nh2:
> > One thing to keep in mind is the existence of #307562.
> 
> I did see that issue. I haven't tested against [#307562 (comment)](https://github.com/NixOS/nixpkgs/pull/307562#issuecomment-2087015893) yet, but I'll do so shortly and paste the results. I started this PR as a 1-to-1 port of the perl code, so I'm "hoping" that this issue is also present in this PR, then they can be addressed together.

@jmbaur Did you ever get around to this?

I am quite certain this issue is still present in `switch-to-configuration-ng`.

I filed a draft PR:

* https://github.com/NixOS/nixpkgs/pull/525813

CC @dasJ

pkgs/by-name/sw/switch-to-configuration-ng/src/main.rs
```diff
@@ -0,0 +1,2077 @@
...
+    let current_system_bin = std::path::PathBuf::from("/run/current-system/sw/bin")
+        .canonicalize()
+        .context("/run/current-system/sw/bin is missing")?;
+
+    let os_release = parse_os_release().context("Failed to parse os-release")?;
+
+    let distro_id_re = Regex::new(format!("^\"?{}\"?$", distro_id).as_str())
+        .context("Invalid regex for distro ID")?;
+
+    // This is a NixOS installation if it has /etc/NIXOS or a proper /etc/os-release.
```
nh2, line 964: > if it has /etc/NIXOS **or** a proper /etc/os-release

@jmbaur This says `OR`. If it has only one of them, is it still a NixOS installation?

If yes, the code is wrong:

`&& !os_release....` short-circuiting `&&` is implemented below, but

```rust
let os_release = parse_os_release().context("Failed to parse os-release")?;
```

unconditonally fails if `/etc/os-release` doesn't exist.

The original Perl implemented a proper `OR`:

```perl
die "This is not a NixOS installation!\n" unless
    -f "/etc/NIXOS" || (read_file("/etc/os-release", err_mode => 'quiet') // "") =~ /ID=nixos/s;
```

https://github.com/NixOS/nixpkgs/commit/f9e2af1e8b62809e9af9acc92826c29f63db8043

---

This change broke deploying with NixOps (in my case, 1.7) to Hetzner, which only creates `/etc/NIXOS` in the target machine's `/mnt` `chroot`, before calling `switch-to-configuration`.

nh2: PR:

* #543312

PR #308815, opened 2024-05-03
the-powder-toy: 97.0.352 -> 98.2.365

LordGrimmauld:
## Description of changes
Changelog: https://github.com/The-Powder-Toy/The-Powder-Toy/compare/v97.0.352...v98.2.365

Additional (nix) changes:
- moved package from pkgs/games to pkgs/by-name
  - for that purpose, restructured the parameter list to be backwards compatible with what was overrideable from all-packages.nix
- drop python dependency, it is just a dependency of meson and does not need to be specified explicitly
- make use of more modern nix tooling to install the desktop file
- add missing preInstall and postInstall run hooks
- drop unnecessary patches that were previously needed to compile on gcc13: the patch was merged upstream
- bump lua dependency from 5.1 to the current lua (5.2)
- ordered parameters and build inputs to be alphabetical
- ran nixpkgs-fmt

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---


LordGrimmauld:
Note: I do not own any darwin machines, i am waiting for ofborg to build this but can't actually run the binaries either. Do *not* merge if darwin tests didn't succeed yet. And feedback from darwin users is welcome.

LordGrimmauld:
Ok, ofborg logs look fine. Quick heads-up by a darwin user would still be appreciated but it seems to work as intended.

pkgs/by-name/th/the-powder-toy/package.nix
```diff
@@ -0,0 +1,75 @@
+{ bzip2
+, Cocoa ? darwin.apple_sdk.frameworks.Cocoa # all-packages.nix used to allow overriding Cocoa, this is fully backwards-compatible
+, copyDesktopItems
+, curl
+, darwin
+, fetchFromGitHub
+, fetchpatch
```
eclairevoyant, line 7:
```suggestion
```

pkgs/by-name/th/the-powder-toy/package.nix
```diff
@@ -0,0 +1,75 @@
...
+}:
+stdenv.mkDerivation rec {
+  pname = "the-powder-toy";
+  version = "98.2.365";
+
+  src = fetchFromGitHub {
+    owner = "The-Powder-Toy";
+    repo = "The-Powder-Toy";
+    rev = "v${version}";
+    sha256 = "sha256-S2aUa25EnUfX6ShW6D+wHrsTLxTcCFcZ/uLE9EWGu4Q=";
```
eclairevoyant, line 29:
```suggestion
    rev = "refs/tags/v${version}";
    hash = "sha256-S2aUa25EnUfX6ShW6D+wHrsTLxTcCFcZ/uLE9EWGu4Q=";
```

eclairevoyant (changes requested):
Other changes are okay, though note that we don't really need to order the input args, nor do we use `nixpkgs-fmt` in nixpkgs (the official formatter is `nixfmt-rfc-style`). If reformatting, that should be a separate commit ideally.

pkgs/top-level/all-packages.nix
```diff
@@ -37516,11 +37516,6 @@ with pkgs;
 
   tcl2048 = callPackage ../games/tcl2048 { };
 
-  the-powder-toy = callPackage ../games/the-powder-toy {
-    lua = lua5_1;
-    inherit (darwin.apple_sdk.frameworks) Cocoa;
-  };
-
```
eclairevoyant, line 37523: This should not be removed, rather the path should be changed to `../by-name/th/the-powder-toy/package.nix`

(Though it now uses lua 5.2, so should likely be pinned to that.)

pkgs/by-name/th/the-powder-toy/package.nix
eclairevoyant, line 7:
```suggestion
, fetchFromGitHub
```

pkgs/top-level/all-packages.nix
LordGrimmauld: Why is this? The line at https://github.com/NixOS/nixpkgs/blob/9863d8c74a959d73df0c495a55097f804f7b379d/pkgs/by-name/th/the-powder-toy/package.nix#L2 means the package behaves identical after the update as it did before (in regards to .override which as far as i understood is the motivation) even without listing in all-packages.nix, so why should it still be listed in all-packages.nix?

eclairevoyant: lua versions aren't backwards-compatible
and in that case what's the point of having Cocoa as a function arg?

LordGrimmauld: I have no idea why Cocoa is a function arg, it was there before i touched the module and figured getting rid of it would break the existing override behaviour. I'd be more than happy getting rid of it.

LordGrimmauld:
Addressed the feedback. Though i can't say i quite agree with it staying in all-packages.nix seeing as i tried keeping the override interface consistent, it is probably easier to just leave it in all-packages instead of arguing over it.

eclairevoyant (approved):

PR #308816, opened 2024-05-03
julia.withPackages: Disable pretty terminal output during nix build

pwaller:
Prior to this patch, during precompilation julia would log output destined for a terminal
which can support cursor movements, which we don't have while building
under nix.

The only way I could find to turn this off is to set CI=true during
precompilation, which substantially reduces the amount of log output you
end up with, to only the salient output (one line per package, instead
of many identical lines being printed).

Signed-off-by: Peter Waller <p@pwaller.net>

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [x] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/development/julia-modules/depot.nix
+++ b/pkgs/development/julia-modules/depot.nix
@@ -55,6 +55,11 @@ runCommand "julia-depot" {
   # git config --global --add safe.directory '/nix'
   export JULIA_PKG_USE_CLI_GIT="true"
 
+  # At time of writing, this appears to be the only way to turn precompiling's
+  # terminal output into standard logging, so opportunistically do that.
+  # (Note this is different from JULIA_CI).
+  export CI=true
+
   julia -e ' \
     import Pkg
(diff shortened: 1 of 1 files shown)


pwaller:
Reference:

https://github.com/JuliaLang/julia/blob/2d87ce3a8d5aff9a67e5294bffe69ccb93cf1217/base/precompilation.jl#L343

Slightly unfortunate that it seems the tty check is failing, which feels like a bug, but I could also imagine that nix is actually allocating a tty? Not sure if there is some other way of achieving the same effect.

Note that inhibiting `canfancyprint` doesn't currently appear to eliminate colours, so those remain, which seems to suit the uses in nix from what I've seen

/cc @NickCao @thomasjm

thomasjm (approved):
This look great, thanks! The output is much improved.

PR #308820, opened 2024-05-03
thunderbirdPackages.thunderbird-102: convert to a `throw`

vcunat:
## Things done

- Build-related checkboxes: N/A

---

### DIFF
--- a/pkgs/applications/networking/mailreaders/thunderbird/packages.nix
+++ b/pkgs/applications/networking/mailreaders/thunderbird/packages.nix
@@ -1,9 +1,7 @@
-{ stdenv, lib, buildMozillaMach, callPackage, fetchurl, fetchpatch, nixosTests, icu, fetchpatch2 }:
+{ stdenv, lib, buildMozillaMach, callPackage, fetchurl, fetchpatch, nixosTests, icu, fetchpatch2, config }:
 
 rec {
   thunderbird = thunderbird-115;
-
-  thunderbird-102 = throw "Thunderbird 102 support ended in September 2023";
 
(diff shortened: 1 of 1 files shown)


vcunat:
As pointed out, maybe someone could consider it a breaking change, even if the package was marked as insecure.  In that case we'd need to wait with the merge.

drupol (approved):

pkgs/applications/networking/mailreaders/thunderbird/packages.nix
```diff
@@ -3,44 +3,7 @@
...
-      attrPath = "thunderbird-unwrapped";
-      versionPrefix = "102";
-    };
-  }).override {
-    geolocationSupport = false;
-    webrtcSupport = false;
-
-    pgoSupport = false; # console.warn: feeds: "downloadFeed: network connection unavailable"
-  };
+  thunderbird-102 = throw "Thunderbird 102 support ended in September 2023";
```
alyssais, line 6: Throws like this should be gated on `config.allowAliases`.

vcunat: Can you reference why?

alyssais: 1. It's just what the convention in Nixpkgs is.  top-level throws go in aliases.nix, which isn't included if allowAliases is set to false, and it's nice to not have that be surprisingly different for non-top-level ones.
2. It makes it so that there's a way to bulid a whole package set at once, even if it contains throws — `nix-build -A thunderbirdPackages --arg config '{allowAliases = false;}'`.

PR #308822, opened 2024-05-03
dockerTools.streamLayeredImage: add includeNixDB argument, expose conf and streamScript

yorickvP:
## Description of changes

This change allows me to override `streamscript` and `conf` without resorting to sed hacks. It looks like this module could use some more overridability love.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [x] Tested, as applicable:

---

### DIFF
--- a/pkgs/build-support/docker/default.nix
+++ b/pkgs/build-support/docker/default.nix
@@ -570,6 +570,8 @@ rec {
       created ? "1970-01-01T00:00:01Z"
     , # Compressor to use. One of: none, gz, zstd.
       compressor ? "gz"
+      # Populate the nix database in the image with the dependencies of `copyToRoot`.
+    , includeNixDB ? false
     , # Deprecated.
       contents ? null
     ,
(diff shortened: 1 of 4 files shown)


pkgs/build-support/docker/default.nix
```diff
@@ -911,6 +911,8 @@ rec {
     , fakeRootCommands ? ""
     , enableFakechroot ? false
     , includeStorePaths ? true
+      # Generate a Nix DB inside the image. The same caveats as `buildImageWithNixDb` apply.
```
roberth, line 914:
```suggestion
```

Please move this to user documentation at `doc/build-helpers/images/dockertools.section.md`.

roberth (changes requested):
Implementation lgtm, but please
- move docs to user manual
- add tests in `nixos/tests/docker-tools.nix`

yorickvP:
@ofborg test docker-tools

yorickvP:
Tests pass locally, not sure what's up with ofborg but it seems unrelated to my changes

pkgs/build-support/docker/default.nix
```diff
@@ -1105,7 +1111,7 @@ rec {
             };
             nativeBuildInputs = [ makeWrapper ];
           } ''
-          makeWrapper ${streamScript} $out --add-flags ${conf}
+          makeWrapper $streamScript $out --add-flags $conf
```
the-sun-will-rise-tomorrow, line 1112: What does this change do?

roberth: It uses the environment variable passed to the derivation environment using the attrset above, instead of interpolating it into this string that runs as bash commands.

the-sun-will-rise-tomorrow: Indeed, but to what end? Why was this change made?

If I had to pick between the two, isn't interpolation slightly better because it avoids polluting the environment and any possible clashes with `mkDerivation` internals?

yorickvP: Having `streamScript` and `conf` in the environment instead of interpolated means we can use `.overrideAttrs (o: { streamScript = ... })` to override the stream script, and `.conf` to extract the conf to use for something else.
In practice, this means I won't have to parse it out of the buildCommand anymore.

the-sun-will-rise-tomorrow (approved):

pkgs/build-support/docker/default.nix
roberth: We may switch `conf` to the `nix2container` format in the future.

roberth (approved):

roberth:
Thank you Yorick!

PR #309028, opened 2024-05-04
hostapd: add "wpa2" to authentication.mode enum

tomfitzhenry:
This is required for some Kindles, and printers (e.g. Brother MFC-J4440DW).

See https://w1.fi/cgit/hostap/plain/hostapd/hostapd.conf for valid `wpa_key_mgmt=` values.

OpenWRT typically adds "wpa_key_mgmt = WPA-PSK", per https://github.com/openwrt/openwrt/blob/3f28c422ba7ca06efd41686fd2f9e664f7e8a12e/package/network/config/wifi-scripts/files/lib/netifd/hostapd.sh#L44-L71

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/nixos/modules/services/networking/hostapd.nix
+++ b/nixos/modules/services/networking/hostapd.nix
@@ -687,7 +687,7 @@ in {
                   authentication = {
                     mode = mkOption {
                       default = "wpa3-sae";
-                      type = types.enum ["none" "wpa2-sha256" "wpa3-sae-transition" "wpa3-sae"];
+                      type = types.enum ["none" "wpa2-sha1" "wpa2-sha256" "wpa3-sae-transition" "wpa3-sae"];
                       description = ''
                         Selects the authentication mode for this AP.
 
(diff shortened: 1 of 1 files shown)


oddlama:
To be honest, I'm not sure whether it would be a good idea to add this new mode to the upstream module for two reasons:

And secondly, if you need this older mode you can use use the `wpa-sha256` mode and add a single configuration override

```nix
```

oddlama:
> This PR doesn't change the default. It just adds a new value to the enum. I think the default should be kept as `wpa3-sae`.

I know, I was referring to the authentication modes as a sort of predefined defaults that users can choose from. Since this module uses settings-style options it doesn't need to be exhaustive, one can just add or override stuff if necessary.
 
> Supporting such devices (which aren't necessarily old, my printer is a 2021 model) is an important part of making hostapd work well for users out-the-box.

Yeah the device isn't old by itself, but it is probably running ancient software versions (better don't look :P)

> That said, it sounds like WPA-PSK-SHA256 is preferable, so we should be encouraging users to use that.

I agree, the SHA1 based method isn't insecure per-se, but it has been superseded by the sha256 one which is why I added only sha256 as an authentication mode to choose from.
 
> How about I rename "wpa2" to "wpa2-sha1", and document it as being less preferred than "wpa2-sha256"?

Sure that'd be good. Just FYI I'm not against adding this I just wasn't sure whether it would be worth the confusion since the the same can be achieved by setting one other option. But I guess it isn't bad for discoverability at least.

> Separately, we could consider changing `authentication.mode` to be a list of strings, rather than a string, to support multiple algorithms, as hostapd supports, and as openwrt configures.

The authentication modes do more than just defining the mgmt algorithms - they are used to determine which password options are necessary and some other related stuff. In hindsight the better thing would probably have been to split the modes just by their logical implications and allow multiple modes to be enabled (ie sae-based and wpa2-based) while allowing a separate option to set the mgmt modes to common tuples. But this would be a bigger (and breaking) change unfortunately.

oddlama:
There are some other occurrences of `wpa2-sha256` in option docstrings which might need to be updated to say "wpa2-sha1" or "wpa2-sha256"

oddlama (approved):

PR #309032, opened 2024-05-04
workflows/backport: Add checkboxes for acceptable and unacceptable changes

pbsds:
## Description of changes

Our backport guidelines are vague, too permissive, and often ignored/forgotten. 
This PR addresses primarily the latter.

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/.github/workflows/backport.yml
+++ b/.github/workflows/backport.yml
@@ -31,5 +31,26 @@ jobs:
           pull_description: |-
             Bot-based backport to `${target_branch}`, triggered by a label in #${pull_number}.
 
-            * [ ] Before merging, ensure that this backport is [acceptable for the release](https://github.com/NixOS/nixpkgs/blob/master/CONTRIBUTING.md#changes-acceptable-for-releases).
-              * Even as a non-commiter, if you find that it is not acceptable, leave a comment.
+            Before merging, ensure that this backport is [acceptable for the stable release](https://github.com/NixOS/nixpkgs/blob/master/CONTRIBUTING.md#changes-acceptable-for-releases), and tick all boxes that apply.
+            As a non-committer unable to tick the checkboxes: if you find that this backport is not acceptable, do leave a comment and consider closing this PR.
+
+            Backwards-compatible changes that are generally okay to backport:
+
+            * [ ] The PR adds new packages, modules and functions
+            * [ ] The PR adds security fixes
+            * [ ] The PR contains patch version updates
+            * [ ] The PR contains minor version updates, which have been verified to include no breaking changes
+
+            Breaking changes that are _not_ okay to backport:
+
+            * [ ] The PR contains major version updates
+            * [ ] The PR changes or removes existing functionality
+            * [ ] The PR contains backward-incompatible changes to the `.override*` interface of a derivation
+            * [ ] The PR deserves an addition to the release notes
+            * [ ] The PR will cause a state migration which risks the loss of user data
+
+            Reasons why a breaking change still needs to be backported:
+
+            * [ ] The PR updates security critical applications, such as `firefox` and `chromium`
+            * [ ] The PR is required to continue interoperating with external services, such as `spotify`, `steam`, and `discord`
+            * [ ] Other, detailed in a comment below
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -484,13 +484,15 @@ The oldest supported release (`YYMM`) can be found using
 nix-instantiate --eval -A lib.trivial.oldestSupportedRelease
 ```
 
-The release branches should generally only receive backwards-compatible changes, both for the Nix expressions and derivations.
+Users should be confident tracking the stable release channels for updates worry-free from breaking changes that will affect them.
+The release branches should only receive backwards-compatible changes, both for the Nix expressions and derivations.
+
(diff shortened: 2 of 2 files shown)


.github/workflows/backport.yml
```diff
@@ -31,5 +31,26 @@ jobs:
...
+            As a non-committer unable to tick the checkboxes: if you find that this backport is not acceptable, do leave a comment and consider closing this PR.
+
+            Backwards-compatible changes that are generally okay to backport:
+
+            * [ ] The PR adds new packages, modules and functions
+            * [ ] The PR adds security fixes
+            * [ ] The PR contains patch version updates
+            * [ ] The PR contains minor version updates, which have been verified to include no breaking changes
+
+            Breaking changes that are _not_ okay to backport:
```
pbsds, line 44: I'd love some input on this section. Is it effective as-is, or should we invert it? I.e. `The PR does not include these backward-incompatible changes:`

.github/workflows/backport.yml
```diff
@@ -31,5 +31,26 @@ jobs:
...
+            * [ ] The PR changes or removes existing functionality
+            * [ ] The PR contains backward-incompatible changes to the `.override*` interface of a derivation
+            * [ ] The PR deserves an addition to the release notes
+            * [ ] The PR will cause a state migration which risks the loss of user data
+
+            Reasons why a breaking change still needs to be backported:
+
+            * [ ] The PR updates security critical applications, such as `firefox` and `chromium`
+            * [ ] The PR is required to continue interoperating with external services, such as `spotify`, `steam`, and `discord`
+            * [ ] Other, detailed in a comment below
```
risicle, line 56: One issue is that this will not be shown for manual backports, of which there are quite a few. So a lot of people won't see this detail - which is why I was suggesting putting information in CONTRIBUTING.

CONTRIBUTING.md
```diff
@@ -484,13 +484,15 @@ The oldest supported release (`YYMM`) can be found using
...
+Users should be confident tracking the stable release channels for updates worry-free from breaking changes that will affect them.
+The release branches should only receive backwards-compatible changes, both for the Nix expressions and derivations.
+
 Here are some examples of backwards-compatible changes that are okay to backport:
 - ✔️ New packages, modules and functions
 - ✔️ Security fixes
 - ✔️ Package version updates
   - ✔️ Patch versions with fixes
-  - ✔️ Minor versions with new functionality, but no breaking changes
+  - ✔️ Minor versions with new functionality, verified to include no breaking changes
```
risicle, line 495: I've still seen plenty of cases where people think they have verified there are no breaking changes when there are. I think people have a very limited idea of what a "breaking change" is. Perhaps we should talk more about "behaviour changes"?

Though I'd personally remove this line entirely, because people often use this as a green light for backporting x.y releases, citing "semantic versioning", even though x.y releases can be quite significant for a lot of projects.

.github/workflows/backport.yml
pbsds: Can we extend the workflow to have the bot comment on all PRs targeting `release-xx.yy` and `staging-xx.yy` with a similar checklist?

CONTRIBUTING.md
pbsds: > I think people have a very limited idea of what a "breaking change" is. Perhaps we should talk more about "behaviour changes"?

I'm up for either term. I think the current failure is that we don't really define it.

> Though I'd personally remove this line entirely, ...

I agree, we're kinda asking them to prove a negative.
Whether to drop the line or not I believe is a `label:significant` change however, as it would mean the release channels to stray further from Fedora and more towards Debian

.github/workflows/backport.yml
Mic92, line 44: I find this actually creates a bit too many checkboxes. Can we shorten this down?

.github/workflows/backport.yml
```diff
@@ -31,5 +31,26 @@ jobs:
...
+
+            Breaking changes that are _not_ okay to backport:
+
+            * [ ] The PR contains major version updates
+            * [ ] The PR changes or removes existing functionality
+            * [ ] The PR contains backward-incompatible changes to the `.override*` interface of a derivation
+            * [ ] The PR deserves an addition to the release notes
+            * [ ] The PR will cause a state migration which risks the loss of user data
+
+            Reasons why a breaking change still needs to be backported:
```
Mic92, line 52: We don't need the section above if we have this section, I would say. It's either an non-breaking backport or it falls into the second category.

.github/workflows/backport.yml
Mic92: Manual backports might be an issue, but those people hopefully eventually see these examples and learn from them.

CONTRIBUTING.md
Mic92: Instead of version, we just say "update" to disconnect it from semantic versioning.

.github/workflows/backport.yml
pbsds: I think of it as a "yes i'm aware we break expectation on x, y, z"

PR #299578, opened 2024-03-27
treewide: CUDA reformat with `nixfmt-rfc-style`

ConnorBaker:
## Description of changes

Reformats the following paths with `nixfmt-rfc-style`:

- `pkgs/development/cuda-modules`
- `pkgs/test/cuda`
- `pkgs/top-level/cuda-packages.nix`

Additionally, adds the commit in which the reformatting happens to `.git-blame-ignore-revs`.

Ideally this will not cause any rebuilds, but it may if whitespace in raw text blocks change.

This PR also introduces a workflow to ensure that the changed files continue to stay formatted.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [x] aarch64-linux
  - [x] x86_64-darwin
  - [x] aarch64-darwin
- For non-Linux: Is sandboxing enabled in `nix.conf`? (See [Nix manual](https://nixos.org/manual/nix/stable/command-ref/conf-file.html))
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:
  - [NixOS test(s)](https://nixos.org/manual/nixos/unstable/index.html#sec-nixos-tests) (look inside [nixos/tests](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests))
  - and/or [package tests](https://nixos.org/manual/nixpkgs/unstable/#sec-package-tests)
  - or, for functions and "core" functionality, tests in [lib/tests](https://github.com/NixOS/nixpkgs/blob/master/lib/tests) or [pkgs/test](https://github.com/NixOS/nixpkgs/blob/master/pkgs/test)
  - made sure NixOS tests are [linked](https://nixos.org/manual/nixpkgs/unstable/#ssec-nixos-tests-linking) to the relevant packages
- [x] Tested compilation of all packages that depend on this change using `nix-shell -p nixpkgs-review --run "nixpkgs-review rev HEAD"`. Note: all changes have to be committed, also see [nixpkgs-review usage](https://github.com/Mic92/nixpkgs-review#usage)
- [ ] Tested basic functionality of all binary files (usually in `./result/bin/`)
[...]

PR #287169, opened 2024-02-08
doc/qt: refresh

K900:
Make examples don't require a custom callPackage, remove note on multiversioning, reword a few things.

Fixes #287015

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
- For non-Linux: Is sandboxing enabled in `nix.conf`? (See [Nix manual](https://nixos.org/manual/nix/stable/command-ref/conf-file.html))
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:
  - [NixOS test(s)](https://nixos.org/manual/nixos/unstable/index.html#sec-nixos-tests) (look inside [nixos/tests](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests))
  - and/or [package tests](https://nixos.org/manual/nixpkgs/unstable/#sec-package-tests)
  - or, for functions and "core" functionality, tests in [lib/tests](https://github.com/NixOS/nixpkgs/blob/master/lib/tests) or [pkgs/test](https://github.com/NixOS/nixpkgs/blob/master/pkgs/test)
  - made sure NixOS tests are [linked](https://nixos.org/manual/nixpkgs/unstable/#ssec-nixos-tests-linking) to the relevant packages
- [ ] Tested compilation of all packages that depend on this change using `nix-shell -p nixpkgs-review --run "nixpkgs-review rev HEAD"`. Note: all changes have to be committed, also see [nixpkgs-review usage](https://github.com/Mic92/nixpkgs-review#usage)
- [x] Tested basic functionality of all binary files (usually in `./result/bin/`)
- [24.05 Release Notes](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2405.section.md) (or backporting [23.05](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2305.section.md) and [23.11](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2311.section.md) Release notes)
  - [ ] (Package updates) Added a release notes entry if the change is major or breaking
  - [ ] (Module updates) Added a release notes entry if the change is significant
  - [ ] (Module addition) Added a release notes entry if adding a new NixOS module
[...]

PR #265524, opened 2023-11-04
qt6: use makeScopeWithSplicing

Artturin:
## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
- For non-Linux: Is sandboxing enabled in `nix.conf`? (See [Nix manual](https://nixos.org/manual/nix/stable/command-ref/conf-file.html))
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:
  - [NixOS test(s)](https://nixos.org/manual/nixos/unstable/index.html#sec-nixos-tests) (look inside [nixos/tests](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests))
  - and/or [package tests](https://nixos.org/manual/nixpkgs/unstable/#sec-package-tests)
  - or, for functions and "core" functionality, tests in [lib/tests](https://github.com/NixOS/nixpkgs/blob/master/lib/tests) or [pkgs/test](https://github.com/NixOS/nixpkgs/blob/master/pkgs/test)
  - made sure NixOS tests are [linked](https://nixos.org/manual/nixpkgs/unstable/#ssec-nixos-tests-linking) to the relevant packages
- [ ] Tested compilation of all packages that depend on this change using `nix-shell -p nixpkgs-review --run "nixpkgs-review rev HEAD"`. Note: all changes have to be committed, also see [nixpkgs-review usage](https://github.com/Mic92/nixpkgs-review#usage)
- [ ] Tested basic functionality of all binary files (usually in `./result/bin/`)
- [23.11 Release Notes](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2311.section.md) (or backporting [23.05 Release notes](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2305.section.md))
  - [ ] (Package updates) Added a release notes entry if the change is major or breaking
  - [ ] (Module updates) Added a release notes entry if the change is significant
  - [ ] (Module addition) Added a release notes entry if adding a new NixOS module
- [ ] Fits [CONTRIBUTING.md](https://github.com/NixOS/nixpkgs/blob/master/CONTRIBUTING.md).

PR #309064, opened 2024-05-04
duckstation: 0.1-6292 -> 0.1-6658 

AndersonTorres:
## Description of changes

Closes https://github.com/NixOS/nixpkgs/pull/308974 (by assimilation)
## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---


qubitnano:
Result of `nixpkgs-review pr 309064` run on x86_64-linux [1](https://github.com/Mic92/nixpkgs-review)
<details>
  <ul>
    <li>duckstation</li>
  </ul>
</details>

Thank you

marcusramberg (approved):
lgtm.

getchoo:
Result of `nixpkgs-review pr 309064` run on aarch64-linux [1](https://github.com/Mic92/nixpkgs-review)
<details>
  <ul>
    <li>duckstation</li>
  </ul>
</details>

getchoo (approved):

pkgs/top-level/all-packages.nix
```diff
@@ -41186,4 +41184,22 @@ with pkgs;
   insulator2 = callPackage ../applications/misc/insulator2 {};
 
   animdl = python3Packages.callPackage ../applications/video/animdl { };
+
+  ### BY-NAME
+
+  duckstation = callPackage ../by-name/du/duckstation/package.nix {
```
AndersonTorres, line 41190: Oh man, I need to think in some `import` hack for this...

SuperSamus, line 41190: Is there any particular reason why it's been moved to the bottom of the file?
(Asking because `shaderc-changes.patch` is also a thing on PCSX2 (#305509), and I didn't know about `excludes`, so I'm wondering whether to do the exact same thing with PCSX2.)

SuperSandro2000: This shouldn't be moved to the bottom and the overwrite should also really be in the package file

pkgs/by-name/du/duckstation/package.nix
```diff
@@ -1,35 +1,42 @@
...
+  qt6,
+  shaderc,
+  stdenv,
+  substituteAll,
+  vulkan-loader,
+  wayland,
 }:
-
+let
+  inherit (qt6)
```
SuperSandro2000, line 23: I think we should use qt6.callPackage here instead

SuperSandro2000 (changes requested):

pkgs/by-name/du/duckstation/package.nix
AndersonTorres: This is not compatible with by-name hierarchy.

pkgs/top-level/all-packages.nix
AndersonTorres: > Is there any particular reason why it's been moved to the bottom of the file?

By personal choice.
Since the old hierarchy is deprecated, but it can still be needed in things like these, it is cleaner to segregate the old hierarchy from the new.

Also, we know well that `all-packages.nix` is too prone to merge conflicts sometimes, and this reduces conflicts a little bit.

> (Asking because `shaderc-changes.patch` is also a thing on PCSX2 (#305509), and I didn't know about `excludes`, so I'm wondering whether to do the exact same thing with PCSX2.)

This is way better than vendoring a modified patch ourselves, given that this patch can be fetched instead.

> and the overwrite should also really be in the package file

I  believe it betrays the intention of the input parameters in the package file. If I am passing the unpatched `shaderc` as an input library, it makes little sense to modify it internally and recompile it.
It is more natural to pass the patched `shaderc` as input instead.

(We can call this "one package, one build step".)

SuperSamus: There's a pro to modify `shaderc` in the package.

Let's say someone wants to override Duckstation's input packages (including `shaderc`), maybe because they want to compile them with some optimizations or something.

If the `shaderc` override is done in `all-packages`, then the user must also manually re-add the patches.

If instead it's done in the package, there's no such problem.
(The problem is in case they wanted to remove the patches.)

AndersonTorres: > If the shaderc override is done in all-packages, then the user must also manually re-add the patches.

You are arguing that we should override _inside_ because of a hypothetical someone that wants to override _outside_. It looks self-defeating.

Also, if the person is interested in override shaderc as you suggest, what are the chances that this overridden shaderc would not be compatible with the patches we did inside the expression?

SuperSandro2000: > Also, we know well that all-packages.nix is too prone to merge conflicts sometimes, and this reduces conflicts a little bit.

Actually appending to the end of the file is super prone to merge conflicts and it was probably done one purpose to not create a new section. Please just leave the callPackage statement where it is.

pkgs/by-name/du/duckstation/package.nix
SuperSandro2000: Avoiding the wrapper by copying what it does into the derivation is definitely not the solution to this.

by-name can have calls from all-packages.nix for exactly those use cases and it is allowed by design.

AndersonTorres: by-name does not accept `qt6.callPackage`, only `callPackage` itself.

Further, to be honest I have problems with those magic `hugePackageSet.callPackage` things, they look too mysterious.
Further, what does this wrapper do that can't be done without it?

Further (I like that word!), I remember very well that some packages had `libsForQt5.callPackage` when it was otherwise not needed. `cmake` was a clear example.

pkgs/top-level/all-packages.nix
AndersonTorres: OK, let's rework this.

pkgs/top-level/all-packages.nix
```diff
@@ -2664,7 +2664,21 @@ with pkgs;
     inherit (darwin.apple_sdk_11_0.frameworks) AudioUnit Carbon Cocoa;
   };
 
-  duckstation = qt6Packages.callPackage ../applications/emulators/duckstation { };
+  duckstation = callPackage ../by-name/du/duckstation/package.nix {
+    shaderc = shaderc.overrideAttrs (old: {
+      patches = (old.patches or [ ]) ++ [
+        (fetchpatch {
+          url = "file://${duckstation.src}/scripts/shaderc-changes.patch";
```
doronbehar, line 2671: This is a bit weird. Is there an IFD limitation to use this:

```nix
patches = [
  ${duckstation.src}/scripts/shaderc-changes.patch
];
```

? I was also thinking, since you promote the `pkgs/by-name` usage, how about putting these details inside the expression, inside the `let` `in`? Then you could leave `all-packages.nix` completely.

AndersonTorres: I am not aware of IFD limitations.
I used this fetchpatch because 
1. fetchpatch is preferrred to vendoring
2. the patch can't be applied without the exclusions listed in the code
3. fetchpatch provides a clean and understandable interface to patch manipulation

i will answer the second part later.

AndersonTorres: > I was also thinking, since you promote the `pkgs/by-name` usage, how about putting these details inside the expression, inside the `let` `in`? Then you could leave `all-packages.nix` completely.

The current documentation about `by-name` puts some effort in preserving the `.override` interface:

  - https://github.com/NixOS/nixpkgs/tree/master/pkgs/by-name#changing-implicit-attribute-defaults
  - https://github.com/NixOS/nixpkgs/tree/master/pkgs/by-name#manual-migration-guidelines

That being said, we have two possibilities:

1. Expect an unpatched `shaderc`, patch and compile it inside the expression, and after that patch duckstation itself
2. Expect an already patched `shaderc` and use it as is.

The second approach looks more natural, since it does not subject the user to a bad surprise.

If by any reason the user of `duckstation` expression wants to override `shaderc`, theey is free to do so.
In the first case, the user would be forced to undo the expression, probably using `overrideAttrs`.

In a certain sense it is an application of [POLA](https://docs.freebsd.org/en/books/handbook/glossary/#pola-glossary).

doronbehar: Hmm it is not clear from what you are saying, why does `shaderc` needs to get this patch.

- If without the patch `duckstation` is not usable, then I don't think it is too bad if it is not trivial to override the `shaderc` that is eventually used.
- Eitherway, when a user is attempting to override a package, usually they need to read the source code anyway, so I don't think the POLA argument applies here, because if anyway you have to read the source code you won't be "astonished".
- You can still provide the user the ability to override the `shaderc` that is actually used in the package, by defining it in the `passthru`, and use it as so:

```nix
stdenv.mkDerivation(finalAttrs: {
  # ...
  buildInputs = [
    # ...
    finalAttrs.passthru.shaderc
    # ...
  ];
  # ...
  passthru = {
    shaderc = shaderc.overrideAttrs(old: {
      # ...
    });
  };
  # ...
}
```

This way, the user will be able to override it by overriding the `passthru`. Also, if they will use `patches = old.patches + [ (their patches) ];`, they should retain the patches provided by `duckstation`, and the overriding experience will be pleasant IMO.

SuperSamus: Without the patch, Duckstation fail to compile.

AndersonTorres: It has an extra problem: if the shaderc is patched and built inside duckstation, then duckstation requires a nested building, and an expression should build only one thing, not two.

AndersonTorres: > Hmm it is not clear from what you are saying, why does `shaderc` needs to get this patch.

Basically it is a vendoring, but made via patches instead of using git modules or scripts.

> 
>     * If without the patch `duckstation` is not usable, then I don't think it is too bad if it is not trivial to override the `shaderc` that is eventually used.

It has an extra problem: if the shaderc is patched and built inside duckstation, then duckstation requires a nested building, and an expression should build only one thing, not two.

> 
>     * Eitherway, when a user is attempting to override a package, usually they need to read the source code anyway, so I don't think the POLA argument applies here, because if anyway you have to read the source code you won't be "astonished".

The expression should be encapsulated, acting like a blackbox. For all effects and purposes, Nixpkgs is an API,  sorta. 
It is not transparent to declare `shaderc` as an input when it is an implementation detail of that expression.

> 
>     * You can still provide the user the ability to override the `shaderc` that is actually used in the package, by defining it in the `passthru`, and use it as so:

This is too confusing.

First, usually passthru is used to emit informations about the expression, not as an input.
Second, your suggestion looks like a convoluted form of doing the same: patching an input.

doronbehar: > > Hmm it is not clear from what you are saying, why does `shaderc` needs to get this patch.
> 
> Basically it is a vendoring, but made via patches instead of using git modules or scripts.

This is not an answer to the question. @SuperSamus' answered this.

> > If without the patch `duckstation` is not usable, then I don't think it is too bad if it is not trivial to override the `shaderc` that is eventually used.
> 
> It has an extra problem: if the shaderc is patched and built inside duckstation, then duckstation requires a nested building, and an expression should build only one thing, not two.

I didn't mean to suggest to not patch `shaderc` at all, or to build it inside `duckstation`, I was only talking about how to define, and in what scope, the patched `shaderc`.

> > Eitherway, when a user is attempting to override a package, usually they need to read the source code anyway, so I don't think the POLA argument applies here, because if anyway you have to read the source code you won't be "astonished".
> 
> The expression should be encapsulated, acting like a blackbox.

I disagree with that. It sounds like a philosophy you developed personally, and I never heard it before. Moreover, we had discussions at [RFC 166](https://github.com/NixOS/rfcs/pull/166) inventing and using Nix "docstring" format, to document near the inputs of an expression, exactly the kind of information about the role of inputs.

> For all effects and purposes, Nixpkgs is an API, sorta. It is not transparent to declare `shaderc` as an input when it is an implementation detail of that expression.

Sounds like this argument was derived from the derived from the blackbox philosophy, and I disagree with it as well.

> > You can still provide the user the ability to override the `shaderc` that is actually used in the package, by defining it in the `passthru`, and use it as so:
> 
> This is too confusing.

If you explain this in comments, both near the initial inputs, and near the `finalAttrs.passthru.shaderc` line in `buildInputs`. It is not confusing IMO.

> First, usually passthru is used to emit informations about the expression, not as an input.

The fact that `shaderc` is overridden, and it is not the same `shaderc` from `pkgs`, is in a way an information about the expression, that you help expose, by putting it in the `passthru`.

Secondly, this practice is very useful in cases such as this:

- `pkg1` requires a patched `pkg2`, named `pkg2'`.
- `pkg3` requires both `pkg1`, and `pkg2'`.

Defining `pkg1.passthru.pkg2 = pkg2';` helps accessing the overriden `pkg2` in the expression of `pkg3`. [This method is used in octave](https://github.com/NixOS/nixpkgs/blob/52493b0dfa0031b9b5f3edff7a96aea74d5a741b/pkgs/development/interpreters/octave/default.nix#L186-L202), and the `passthru` packages are later [used in `octavePackages`](https://github.com/NixOS/nixpkgs/blob/52493b0dfa0031b9b5f3edff7a96aea74d5a741b/pkgs/top-level/octave-packages.nix#L133-L136)

> Second, your suggestion looks like a convoluted form of doing the same: patching an input.

Correct, I am not against patching `shaderc`, I am only discussing where and how the patch should be applied.

AndersonTorres: > This is not an answer to the question. @SuperSamus' answered this.

SuperSamus said what happens without the patch. 
I explained the origin and role of the patch.
Both are acceptable answers.

> I didn't mean to suggest to not patch `shaderc` at all, or to build it inside `duckstation`, I was only talking about how to define, and in what scope, the patched `shaderc`.

I am talking about this all along.
Is there anything confusing about what I am talking about?

> > The expression should be encapsulated, acting like a blackbox.
> 
> I disagree with that. It sounds like a philosophy you developed personally, and I never heard it before. 

You are shifting the burden here.

First, do you have a more technical reason besides your personal disagreement and ignorance? 
Personal disagreement and ignorance are not sufficient reasons to block a PR.

Second, there is nothing hindering me, or any contributor for that matter, of following philosophies developed personally while writing Nix expressions for Nixpkgs.

> Moreover, we had discussions at [RFC 166](https://github.com/NixOS/rfcs/pull/166) inventing and using Nix "docstring" format, to document near the inputs of an expression, exactly the kind of information about the role of inputs.

This is completely irrelevant to this thread.
The docstring standard is, or should be, completely capable to document any of the approaches discussed here.

> > For all effects and purposes, Nixpkgs is an API, sorta. It is not transparent to declare `shaderc` as an input when it is an implementation detail of that expression.
> 
> Sounds like this argument was derived from the derived from the blackbox philosophy, and I disagree with it as well.

Do you have a more technical reason besides your personal disagreement?

> If you explain this in comments, both near the initial inputs, and near the `finalAttrs.passthru.shaderc` line in `buildInputs`. It is not confusing IMO.

If it needs careful commenting, then it is confusing, or at least it is unexpected.

>  [This method is used in octave](https://github.com/NixOS/nixpkgs/blob/52493b0dfa0031b9b5f3edff7a96aea74d5a741b/pkgs/development/interpreters/octave/default.nix#L186-L202), and the `passthru` packages are later [used in `octavePackages`](https://github.com/NixOS/nixpkgs/blob/52493b0dfa0031b9b5f3edff7a96aea74d5a741b/pkgs/top-level/octave-packages.nix#L133-L136)

If you are to cite precedents, then I can cite occurrences similar to mine:

https://github.com/NixOS/nixpkgs/blob/52493b0dfa0031b9b5f3edff7a96aea74d5a741b/pkgs/top-level/all-packages.nix#L24296-L24312

> Correct, I am not against patching `shaderc`, I am only discussing where and how the patch should be applied.

Nice.
Do you have a more technical reason besides your personal preference?

AndersonTorres: > The only technical consideration I will stand upon, is to not put so many details in `all-packages.nix`, and avoid editing it as much as possible (only removing lines is OK of course).

I do not like to edit `all-packages` too.
Indeed when I was doing tree-wide migration Infinisil theemself told me to stop.

However there is almost no other way to override inputs.
Many of the ideas were rejected in the [RFC 140 itself](https://github.com/NixOS/rfcs/blob/f15c1cf8a73013d32092148147abaaddbf1b5e75/rfcs/0140-simple-package-paths.md?plain=1#L357)

> I feel you prefer misunderstanding me, instead of acknowledge the subtleties I am trying to layout for you, to help you with your PR.

I am trying to see your point, but I feel that hiding the modifications over shaderc is not correct.
I will try another idea.

pkgs/by-name/du/duckstation/package.nix
```diff
@@ -1,35 +1,43 @@
...
-, ninja
-, pkg-config
-, qtbase
-, qtsvg
-, qttools
-, qtwayland
-, substituteAll
-, vulkan-loader
-, wayland
-, wrapQtAppsHook
```
doronbehar, line 21: Personally, I think this is not a required change, and it enlarges the diff for a weak reason, but it is your right to format the expression how you wish, since you are a maintainer.

pkgs/by-name/du/duckstation/package.nix
```diff
@@ -57,27 +65,34 @@ stdenv.mkDerivation (finalAttrs: {
...
-    runHook postCheck
-  '';
+  qtWrapperArgs =
+    let
+      libPath = lib.makeLibraryPath ([
+        vulkan-loader
+      ] ++ cubeb.passthru.backendLibs);
+    in [
+      "--prefix LD_LIBRARY_PATH : ${libPath}"
+    ];
```
doronbehar, line 91: Same as this change - you just moved it, and you enlarge the diff without no real change to the derivation.. I usually make an effort to make the amount of diffs as small as possible.

pkgs/by-name/du/duckstation/package.nix
```diff
@@ -1,35 +1,43 @@
...
+  pkg-config,
+  qt6,
+  stdenv,
+  substituteAll,
+  vulkan-loader,
+  wayland,
 }:
-
+let
+  shaderc-patched = callPackage ./shaderc-patched.nix { };
```
doronbehar, line 23: Now no one can override `shaderc` without using `overrideAttrs` and modifying the `buildInputs` - something that would have been possible if you had used `finalAttrs.passthru.shaderc` as I suggested back then. However, that's not mandatory, and many packages work like this and probably barely anyone complains, so it's not too bad.

pkgs/top-level/all-packages.nix
```diff
@@ -2646,8 +2646,6 @@ with pkgs;
     inherit (darwin.apple_sdk_11_0.frameworks) AudioUnit Carbon Cocoa;
   };
 
-  duckstation = qt6Packages.callPackage ../applications/emulators/duckstation { };
-
```
doronbehar, line 2650: Excellent!

pkgs/by-name/du/duckstation/package.nix
AndersonTorres: I split this in a separate commit.

pkgs/by-name/du/duckstation/package.nix
AndersonTorres: As I said before, shaderc works like a vendored lib - but instead of pulling the whole code, they just distribute a patch.

pkgs/by-name/du/duckstation/package.nix
doronbehar: > I split this in a separate commit.

That's good idea, but still, now in the 3rd commit, the `qtWrapperArgs` and the `installCheckPhase` switched places after the reformat commit for no real reason.

doronbehar (commented):
Looks much better now!

pkgs/by-name/du/duckstation/package.nix
AndersonTorres: Hammering:

https://github.com/jtojnar/nixpkgs-hammering

AndersonTorres (commented):
.

pkgs/by-name/du/duckstation/package.nix
```diff
@@ -57,26 +63,27 @@ stdenv.mkDerivation (finalAttrs: {
...
+    extra-cmake-modules
     libXrandr
     libbacktrace
     qtbase
     qtsvg
     qtwayland
     wayland
-  ]
-  ++ cubeb.passthru.backendLibs;
+  ] ++ cubeb.passthru.backendLibs;
```
doronbehar, line 77: Is this diff related to nixpkgs-hammering?

pkgs/by-name/du/duckstation/package.nix
```diff
@@ -57,26 +63,27 @@ stdenv.mkDerivation (finalAttrs: {
...
+
+  cmakeFlags = [ (lib.cmakeBool "BUILD_TESTS" true) ];
```
doronbehar, line 79: This attribute also switched places, and you removed the line break and indentation, with no justification.

pkgs/by-name/du/duckstation/package.nix
```diff
@@ -95,12 +102,9 @@ stdenv.mkDerivation (finalAttrs: {
...
     let
-      libPath = lib.makeLibraryPath ([
-        vulkan-loader
-      ] ++ cubeb.passthru.backendLibs);
-    in [
-      "--prefix LD_LIBRARY_PATH : ${libPath}"
-    ];
+      libPath = lib.makeLibraryPath ([ vulkan-loader ] ++ cubeb.passthru.backendLibs);
+    in
+    [ "--prefix LD_LIBRARY_PATH : ${libPath}" ];
```
doronbehar: I see no reason for this change as well.

doronbehar (changes requested):
Please minimize your diffs.

pkgs/by-name/du/duckstation/package.nix
AndersonTorres: Nope, this is nixfmt.

pkgs/by-name/du/duckstation/package.nix
AndersonTorres: The "justification" is nixfmt.

pkgs/by-name/du/duckstation/package.nix
AndersonTorres: I will remove nixfmt from this code.

AndersonTorres:
Who added that "Check that Nix files are formatted" without telling it to everyone???

doronbehar:
CI error:

```
/nix/store/xfhkjnpqjwlf6hlk1ysmq3aaq80f3bjj-stdenv-linux/setup: line 1581: out/share/duckstation/common-tests: No such file or directory
```

doronbehar:
> Who added that "Check that Nix files are formatted" without telling it to everyone???

Where?

AndersonTorres:
Now it should work.

pkgs/by-name/du/duckstation/package.nix
```diff
@@ -1,35 +1,43 @@
 { lib
-, stdenv
-, fetchFromGitHub
```
doronbehar, line 3: Unnecessary change - moving inputs enlarges the diff for no good reason.

pkgs/by-name/du/duckstation/package.nix
```diff
@@ -71,21 +71,22 @@ stdenv.mkDerivation (finalAttrs: {
...
-
   cmakeFlags = [
     (lib.cmakeBool "BUILD_TESTS" true)
   ];
 
-  doInstallCheck = true;
-
-  installCheckPhase = ''
-    runHook preCheck
+  qtWrapperArgs =
```
doronbehar, line 83: There is no concensus even in https://github.com/jtojnar/nixpkgs-hammering/blob/main/explanations/attribute-ordering.md about where to put this argument. I think this commit should be dropped.

doronbehar (changes requested):
I'm beginning to lose patience due to your insistence on enlarging the diff for no real good reason. I wanted to spare us the review-change ping pong, by rebasing your branch myself, but you are not [allowing maintainers to edit the PR](https://github.blog/changelog/2023-02-10-see-when-a-pull-request-is-editable-by-maintainers/) .

Your next trial is your last.

pkgs/by-name/du/duckstation/package.nix
doronbehar: > Unnecessary change - moving inputs enlarges the diff for no good reason.

Still unresolved.

marcusramberg: Sorting them seems reasonable to me.

AndersonTorres: > Sorting them seems reasonable to me.

It does seem stupid to doronbehar, let's solve this.

AndersonTorres:
> but you are not [allowing maintainers to edit the PR](https://github.blog/changelog/2023-02-10-see-when-a-pull-request-is-editable-by-maintainers/) .

I am on that web interface of GitHub, and the "allow edits by maintainers" checkbox does not exist.
I don't know why.

I have opened another PR [right now](https://github.com/NixOS/nixpkgs/pull/314714)  and can confirm: there is no checkbox.

edits

AndersonTorres:
Oh, anything else? [2]

eclairevoyant:
> the "allow edits by maintainers" checkbox does not exist.

https://github.com/orgs/community/discussions/5634

> Currently, if a PR is created from a fork _living under an organisation_ (rather than an individual) it's not possible to allow edits by maintainers.

doronbehar (approved):

pkgs/by-name/du/duckstation/package.nix
doronbehar: Merging is blocked due to this @SuperSandro2000 's review, so I marked this conversation as unresolved. I think it is OK to not use `qt6.callPackage` here, especially if it allows us to transition to `pkgs/by-name`, and I think it should be the choice of the maintainer to use it or not, but that's only my opinion.

SuperSandro2000: qt6.callPackage does more than bringing the qt packages into scope. It also does splicing which is complicated (#265524, #269654) and if it is slightly off, it breaks cross compilation. It also uses the correct stdenv for darwin but that is not relevant for this package. Instead of suggesting this pattern to other contributors and luring them into broken darwin and cross builds, please just use the qt6.callPackage.

If other ecosystems have a patterns on how to consume them, we should really use them, especially if it is easily possible, to avoid issues for our packages and to avoid extra work for the ecosystem maintainers and especially avoid extra work debugging hard to debug topics like cross compilation which is especially hairy inside of qt.

pkgs by-name is not a good reason to avoid the wrapper. A good known reason is when qt is used inside another wrapper like buildPythonPackage.

libsForQt5 also generated the package for every minor qt5 version which is right now not relevant and cmake probably doesn't need it but duckstation is I think a qt application.

SuperSandro2000 (dismissed):

pkgs/by-name/du/duckstation/package.nix
infinisil: Oh yeah the cross compilation is a good argument. Recently the qt docs [have been updated](https://github.com/NixOS/nixpkgs/pull/287169) by @K900 (also ping @Aleksanaa and @eclairevoyant) to avoid recommending `qt6.callPackage`, but the cross compilation argument never came up.

There is in fact still a way to use `qt6.callPackage` with `pkgs/by-name`, but you need to nest it:

```nix
# pkgs/by-name/fo/foo/package.nix
{ qt6 }:
qt6.callPackage (
  { qtbase, stdenv }:
  stdenv.mkDerivation { ... }
) { }
```

infinisil:
For easily finding it, a discussion is happening in [this thread](https://github.com/NixOS/nixpkgs/pull/309064#discussion_r1615019940).

> Bringing back qt6.callPackage is not compatible to get rid of all-packages.nix.

https://github.com/NixOS/nixpkgs/pull/309064#discussion_r1617804002 :)

> Who added that "Check that Nix files are formatted" without telling it to everyone???

It's [only for a small subset of files](https://github.com/NixOS/nixpkgs/blob/7af44693dff0a56523149fa2abdff85d17318fe1/.github/workflows/check-nix-format.yml#L35-L48) for now, people are very eager to use the standard formatter already :). For reference, https://github.com/NixOS/nixpkgs/pull/299578 introduced it, also https://github.com/NixOS/nixpkgs/pull/313628 is relevant. Also see https://github.com/NixOS/nixfmt/issues/204.

pkgs/by-name/du/duckstation/package.nix
eclairevoyant: > it breaks cross compilation.

Cross-compilation is already broken in this package.

<details>
<summary><code>nix-build -A pkgsCross.aarch64-multiplatform.duckstation --option keep-going false</code></summary>

```
error:
         at <nix/derivation-internal.nix>:9:12:
            8|
            9|   strict = derivationStrict drvAttrs;
             |            ^
           10|

       … while evaluating derivation 'duckstation-aarch64-unknown-linux-gnu-0.1-6658'
         whose name attribute is located at /mnt/scratch/nixpkgs/pkgs/stdenv/generic/make-derivation.nix:331:7

       … while evaluating attribute 'postFixup' of derivation 'duckstation-aarch64-unknown-linux-gnu-0.1-6658'
         at /mnt/scratch/nixpkgs/pkgs/by-name/du/duckstation/package.nix:119:3:
          118|   # Can't avoid the double wrapping, the binary wrapper from qtWrapperArgs doesn't support --run
          119|   postFixup = ''
             |   ^
          120|     source "${makeWrapper}/nix-support/setup-hook"

       error: makeWrapper/makeShellWrapper must be in nativeBuildInputs
```

</details>

After fixing with the following diff:

<details>

```diff
diff --git a/pkgs/by-name/du/duckstation/package.nix b/pkgs/by-name/du/duckstation/package.nix
index 37d2d01534b1..3c29c8f0daec 100644
--- a/pkgs/by-name/du/duckstation/package.nix
+++ b/pkgs/by-name/du/duckstation/package.nix
@@ -16,18 +16,16 @@
 , qt6
 , substituteAll
 , vulkan-loader
+, qtbase
+, qtsvg
+, qttools
+, qtwayland
+, wrapQtAppsHook
 , wayland
 }:
 
 let
   shaderc-patched = callPackage ./shaderc-patched.nix { };
-  inherit (qt6)
-    qtbase
-    qtsvg
-    qttools
-    qtwayland
-    wrapQtAppsHook
-  ;
 in
 stdenv.mkDerivation (finalAttrs: {
   pname = "duckstation";
@@ -60,6 +58,7 @@ stdenv.mkDerivation (finalAttrs: {
     pkg-config
     qttools
     wrapQtAppsHook
+    makeWrapper
   ];
 
   buildInputs = [
@@ -118,7 +117,6 @@ stdenv.mkDerivation (finalAttrs: {
   # https://github.com/stenzek/duckstation/blob/master/scripts/appimage/apprun-hooks/default-to-x11.sh
   # Can't avoid the double wrapping, the binary wrapper from qtWrapperArgs doesn't support --run
   postFixup = ''
-    source "${makeWrapper}/nix-support/setup-hook"
     wrapProgram $out/bin/duckstation-qt \
       --run 'if [[ -z $I_WANT_A_BROKEN_WAYLAND_UI ]]; then export QT_QPA_PLATFORM=xcb; fi'
   '';
diff --git a/pkgs/top-level/all-packages.nix b/pkgs/top-level/all-packages.nix
index a4d7e00bd616..a485289fb4bb 100644
--- a/pkgs/top-level/all-packages.nix
+++ b/pkgs/top-level/all-packages.nix
@@ -41039,4 +41039,6 @@ with pkgs;
   dillo = callPackage ../by-name/di/dillo/package.nix {
     fltk = fltk13;
   };
+
+  duckstation = qt6.callPackage ../by-name/du/duckstation/package.nix { };
 }
```
</details>

 we get

<details>

```
these 26 derivations will be built:
  /nix/store/c0sndkqwvr8v15qbg7nl6in4pkxdp8xz-psqlodbc-aarch64-unknown-linux-gnu-10.01.0000.drv
  /nix/store/p7axjbgyy75k6971ccg9lyyl7swggx0y-qtbase-aarch64-unknown-linux-gnu-6.7.1.drv
  /nix/store/28mzzsxwsnzagg1phy7wzk747fmqz0b5-qtlanguageserver-aarch64-unknown-linux-gnu-6.7.1.drv
  /nix/store/49bf87db3pxd61523jhv9wzv1pwxx0ih-qtsvg-aarch64-unknown-linux-gnu-6.7.1.drv
  /nix/store/g89cpcfa3dq4pwk0wp6vq5fh9lx22pcx-qttranslations-6.7.1.drv
  /nix/store/8mwdxyj4k01m9bskcqw5bd0iydn280vp-qtbase-6.7.1.drv
  /nix/store/4nr0ia7yg5f5iq6bjwnvj3jxpwn7qfk1-qtlanguageserver-6.7.1.drv
  /nix/store/8lpsajmwdvwq0cdmwaj2aw2l2rznz4z4-qtsvg-aarch64-unknown-linux-gnu-5.15.12.drv
  /nix/store/b9rvq2zw7qzfifqqwc82rsf7s7jvvq2y-qtdeclarative-aarch64-unknown-linux-gnu-5.15.12.drv
  /nix/store/5v5ha9yn3jqi9r9x2w0psxbnqdljjldb-qtwebchannel-aarch64-unknown-linux-gnu-5.15.12.drv
  /nix/store/krgycar52dh123x4c483bfw1rvsj7bm5-python3.11-pyqt5-5.15.9-aarch64-unknown-linux-gnu.drv
  /nix/store/68ryybbx4z7fyd1kzszjz2skciw19p9x-python3-aarch64-unknown-linux-gnu-3.11.9-env.drv
  /nix/store/mspls9azizz05d4rcvhisg1yz7brq9jk-qtsvg-5.15.12.drv
  /nix/store/7ar0rgvx6hx6ng5z7iw2d45vi61m1qj1-qtdeclarative-5.15.12.drv
  /nix/store/xn1970m3bw410lxiicbf9125grhmhkzj-qtshadertools-aarch64-unknown-linux-gnu-6.7.1.drv
  /nix/store/n2xiqaya1y12hx6riif63pgdn2bvv55p-qtdeclarative-aarch64-unknown-linux-gnu-6.7.1.drv
  /nix/store/b68lsdbi5wny6zgbd31baj08xbx6qrfx-qtwayland-aarch64-unknown-linux-gnu-6.7.1.drv
  /nix/store/d5dgjms7mkc44y8qyynbrcaqcgij9v8i-qtquickcontrols-5.15.12.drv
  /nix/store/8zsw31dyizv9bp5xg3a0qq47kf1sxc2f-qtwayland-5.15.12.drv
  /nix/store/mk4q210p59fwp99ralhdhg5kjd2mn7pn-wrap-qt5-apps-hook.drv
  /nix/store/dbzcs11w17z9pr5hqn0ikk29qrsfv63r-ffado-aarch64-unknown-linux-gnu-2.4.8.drv
  /nix/store/isgcnw2rx2i6ygk2g99lywn0r7k1gq69-jack2-aarch64-unknown-linux-gnu-1.9.22.drv
  /nix/store/ncb6vp6ybwx7igp7sp8ibcmsnvrmxr1a-qtshadertools-6.7.1.drv
  /nix/store/i9r3nchvhsbcjvjm5z9rx7cfczm2q9fh-qtdeclarative-6.7.1.drv
  /nix/store/lihl5w0zl5nrw3kk25pxgl7kafl6nd6l-qttools-6.7.1.drv
  /nix/store/8qd2jxh1v6wf2vr3bhr91m49drpb32i0-duckstation-aarch64-unknown-linux-gnu-0.1-6658.drv
building '/nix/store/c0sndkqwvr8v15qbg7nl6in4pkxdp8xz-psqlodbc-aarch64-unknown-linux-gnu-10.01.0000.drv'...
building '/nix/store/g89cpcfa3dq4pwk0wp6vq5fh9lx22pcx-qttranslations-6.7.1.drv'...
building '/nix/store/8lpsajmwdvwq0cdmwaj2aw2l2rznz4z4-qtsvg-aarch64-unknown-linux-gnu-5.15.12.drv'...
building '/nix/store/mspls9azizz05d4rcvhisg1yz7brq9jk-qtsvg-5.15.12.drv'...
Running phase: unpackPhase
unpacking source archive /nix/store/gq9jk11lkim0fvy504593nd4rkwzv0kr-psqlodbc-10.01.0000.tar.gz
source root is psqlodbc-10.01.0000
setting SOURCE_DATE_EPOCH to timestamp 1514380278 of file psqlodbc-10.01.0000/test/Makefile
Running phase: patchPhase
Running phase: updateAutotoolsGnuConfigScriptsPhase
Updating Autotools / GNU config script to a newer upstream version: ./config/config.sub
Updating Autotools / GNU config script to a newer upstream version: ./config/config.guess
Running phase: updateAutotoolsGnuConfigScriptsPhase
Updating Autotools / GNU config script to a newer upstream version: ./config/config.sub
Updating Autotools / GNU config script to a newer upstream version: ./config/config.guess
Running phase: configurePhase
fixing libtool script ./config/ltmain.sh
patching script interpreter paths in ./configure
./configure: interpreter directive changed from "#! /bin/sh" to "/nix/store/306znyj77fv49kwnkpxmb0j2znqpa8bj-bash-5.2p26/bin/sh"
configure flags: --disable-static --disable-dependency-tracking --prefix=/nix/store/r8bx0c4whyz4ki1bx3ng3l7cnahg7ig6-psqlodbc-aarch64-unknown-linux-gnu-10.01.0000 --build=x86_64-unknown-linux-gnu --host=aarch64-unknown-linux-gnu
checking for a BSD-compatible install... /nix/store/php4qidg2bxzmm79vpri025bqi0fa889-coreutils-9.5/bin/install -c
checking whether build environment is sane... yes
checking for aarch64-unknown-linux-gnu-strip... aarch64-unknown-linux-gnu-strip
checking for a thread-safe mkdir -p... /nix/store/php4qidg2bxzmm79vpri025bqi0fa889-coreutils-9.5/bin/mkdir -p
checking for gawk... gawk
checking whether make sets $(MAKE)... yes
checking whether make supports nested variables... yes
checking whether to enable maintainer-specific portions of Makefiles... no
checking for aarch64-unknown-linux-gnu-gcc... aarch64-unknown-linux-gnu-gcc
Error: detected mismatched Qt dependencies:
    /nix/store/lrz3zjch3xyihm08a3q2p7fin13a169l-qtbase-aarch64-unknown-linux-gnu-5.15.12-dev
    /nix/store/6fr1msliinblzypw9zz51ffib1nzrkd4-qtbase-5.15.12-dev
Error: detected mismatched Qt dependencies:
    /nix/store/6fr1msliinblzypw9zz51ffib1nzrkd4-qtbase-5.15.12-dev
    /nix/store/p8cr0k5g2jbjy6k80sdhfyr7vp4zzl8w-qtbase-5.15.12-dev
error: builder for '/nix/store/8lpsajmwdvwq0cdmwaj2aw2l2rznz4z4-qtsvg-aarch64-unknown-linux-gnu-5.15.12.drv' failed with exit code 1;
       last 3 log lines:
       > Error: detected mismatched Qt dependencies:
       >     /nix/store/lrz3zjch3xyihm08a3q2p7fin13a169l-qtbase-aarch64-unknown-linux-gnu-5.15.12-dev
       >     /nix/store/6fr1msliinblzypw9zz51ffib1nzrkd4-qtbase-5.15.12-dev
       For full logs, run 'nix log /nix/store/8lpsajmwdvwq0cdmwaj2aw2l2rznz4z4-qtsvg-aarch64-unknown-linux-gnu-5.15.12.drv'.
error: 1 dependencies of derivation '/nix/store/krgycar52dh123x4c483bfw1rvsj7bm5-python3.11-pyqt5-5.15.9-aarch64-unknown-linux-gnu.drv' failed to build
error: 1 dependencies of derivation '/nix/store/dbzcs11w17z9pr5hqn0ikk29qrsfv63r-ffado-aarch64-unknown-linux-gnu-2.4.8.drv' failed to build
error: 1 dependencies of derivation '/nix/store/isgcnw2rx2i6ygk2g99lywn0r7k1gq69-jack2-aarch64-unknown-linux-gnu-1.9.22.drv' failed to build
error: 1 dependencies of derivation '/nix/store/8qd2jxh1v6wf2vr3bhr91m49drpb32i0-duckstation-aarch64-unknown-linux-gnu-0.1-6658.drv' failed to build
```

</details>

This error happens whether you switch back to `qt6.callPackage` and switch the expression args or not... so no one is bothering to cross-compile `duckstation` anyway, this argument is pointless. Fixing cross-compilation should be done in another PR.

K900: Qt cross is woefully broken, and if it is ever fixed, it should be handled by makeScopeWithSplicing and friends, not whatever this is.

doronbehar: > There is in fact still a way to use `qt6.callPackage` with `pkgs/by-name`, but you need to nest it:
> 
> ```nix
> # pkgs/by-name/fo/foo/package.nix
> { qt6 }:
> qt6.callPackage (
>   { qtbase, stdenv }:
>   stdenv.mkDerivation { ... }
> ) { }
> ```

@AndersonTorres this looks like the best option to me - both satisfying @SuperSandro2000's request and the desire to use `pkgs/by-name`.

eclairevoyant: That's as bad as doing `{ pkgs }: pkgs.stdenv.mkDerivation { ... }`, except worse, since you can't even override anything under `pkgs` here.

Rather than tortured solutions to theoretical concerns that haven't even been theoretically demonstrated, let's concretely decide on either `by-name` or `qt6.callPackage`, not both.

eclairevoyant: By the way, using `by-name` does not break splicing; _inheriting_ from `qt6` when `callPackage`-ing _would_ break splicing (see #204303).

To adapt the example shown in that issue:
```nix
# test.nix
let
  pkgs = import <nixpkgs> { };
  pkgsCross = pkgs.pkgsCross.aarch64-multiplatform;
  test1 = pkgsCross.callPackage
    ({ qt6, xorg }: {
      q = qt6.qtbase.__spliced;
      x = xorg.libX11.__spliced;
    })
    { };

  test2 = pkgsCross.callPackage
    ({ qtbase, libX11 }: {
      d = qtbase.__spliced;
      x = libX11.__spliced;
    })
    {
      inherit (pkgsCross.qt6) qtbase;
      inherit (pkgsCross.xorg) libX11;
    };
in
{
  # for --json to work
  workingSplices =
    removeAttrs test1 [ "override" "overrideDerivation" ];
  notWorkingSplices =
    removeAttrs test2 [ "override" "overrideDerivation" ];
}
```


<details>
<summary><code>NIXPKGS_ALLOW_UNSUPPORTED_SYSTEM=1 nix eval -f test.nix workingSplices --json | jq</code></summary>

```json
{
  "q": {
    "buildBuild": "/nix/store/ikqvl8jq0d7vlp25d9zlbh44m4hljahc-qtbase-6.7.1",
    "buildHost": "/nix/store/lfqjxhf4vf6nwz75sjzg1ixqaszzz1ay-qtbase-6.7.1",
    "buildTarget": "/nix/store/lfqjxhf4vf6nwz75sjzg1ixqaszzz1ay-qtbase-6.7.1",
    "hostHost": "/nix/store/b1547ipq4rbzxwccs2x8fh52bw07h8rw-qtbase-aarch64-unknown-linux-gnu-6.7.1",
    "hostTarget": "/nix/store/b1547ipq4rbzxwccs2x8fh52bw07h8rw-qtbase-aarch64-unknown-linux-gnu-6.7.1"
  },
  "x": {
    "buildBuild": "/nix/store/x9fw7rbdb34gq0f8q750kw344lbv9nk1-libX11-1.8.9",
    "buildHost": "/nix/store/x9fw7rbdb34gq0f8q750kw344lbv9nk1-libX11-1.8.9",
    "buildTarget": "/nix/store/x9fw7rbdb34gq0f8q750kw344lbv9nk1-libX11-1.8.9",
    "hostHost": "/nix/store/0ds953302knjk042h2j78pyz851zip5r-libX11-aarch64-unknown-linux-gnu-1.8.9",
    "hostTarget": "/nix/store/0ds953302knjk042h2j78pyz851zip5r-libX11-aarch64-unknown-linux-gnu-1.8.9"
  }
}
```

</details>

<details>
<summary><code>NIXPKGS_ALLOW_UNSUPPORTED_SYSTEM=1 nix eval -f test.nix notWorkingSplices --json | jq</code></summary>

```
error:
       … while evaluating attribute 'd'
         at /home/jelly3781/Desktop/test.nix:13:7:
           12|     ({ qtbase, libX11 }: {
           13|       d = qtbase.__spliced;
             |       ^
           14|       x = libX11.__spliced;

       error: attribute '__spliced' missing
       at /home/jelly3781/Desktop/test.nix:13:11:
           12|     ({ qtbase, libX11 }: {
           13|       d = qtbase.__spliced;
             |           ^
           14|       x = libX11.__spliced;
```

</details>

So I consider this is a non-issue, splicing works.

eclairevoyant (approved):
I didn't test properly running this in a game, but builds and starts fine.

pkgs/by-name/du/duckstation/package.nix
AndersonTorres: Closing as solved.

PR #309068, opened 2024-05-04
nixos/dbus: fix switching from dbus-broker to dbus

SuperSandro2000:
## Description of changes

When switching ``services.dbus.implementation = "dbus-broker";`` back to "dbus", we don't want to start/stop/restart any dbus related service.

see #307731

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [x] Tested, as applicable:

---

### DIFF
--- a/nixos/modules/services/system/dbus.nix
+++ b/nixos/modules/services/system/dbus.nix
@@ -147,6 +147,10 @@ in
       };
 
       systemd.services.dbus = {
+        aliases = [
+          # hack aiding to prevent dbus from restarting when switching from dbus-broker back to dbus
+          "dbus-broker.service"
+        ];
         # Don't restart dbus-daemon. Bad things tend to happen if we do.
         reloadIfChanged = true;
         restartTriggers = [
@@ -158,6 +162,10 @@ in
       };
 
       systemd.user.services.dbus = {
+        aliases = [
+          # hack aiding to prevent dbus from restarting when switching from dbus-broker back to dbus
+          "dbus-broker.service"
+        ];
         # Don't restart dbus-daemon. Bad things tend to happen if we do.
         reloadIfChanged = true;
         restartTriggers = [
@@ -184,6 +192,8 @@ in
       # https://github.com/NixOS/nixpkgs/issues/108643
       systemd.services.dbus-broker = {
         aliases = [
+          # allow other services to just depend on dbus,
+          # but also a hack aiding to prevent dbus from restarting when switching from dbus-broker back to dbus
           "dbus.service"
         ];
         unitConfig = {
@@ -203,6 +213,8 @@ in
 
       systemd.user.services.dbus-broker = {
         aliases = [
+          # allow other services to just depend on dbus,
+          # but also a hack aiding to prevent dbus from restarting when switching from dbus-broker back to dbus
           "dbus.service"
         ];
         # Don't restart dbus. Bad things tend to happen if we do.
--- a/nixos/modules/system/activation/switch-to-configuration.pl
+++ b/nixos/modules/system/activation/switch-to-configuration.pl
@@ -472,6 +472,9 @@ sub handle_modified_unit { ## no critic(Subroutines::ProhibitManyArgs, Subroutin
             $units_to_reload->{$unit} = 1;
             record_unit($reload_list_file, $unit);
         }
+        elsif ($unit eq "dbus.service" || $unit eq "dbus-broker.service") {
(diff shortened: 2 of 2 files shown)


nixos/modules/services/system/dbus.nix
```diff
@@ -147,6 +147,9 @@ in
       };
 
       systemd.services.dbus = {
+        aliases = [
+          "dbus-broker.service"
```
flokli, line 151: This is for the dbus implementation, not the dbus-broker implementation. Please add here (and further down) a comment stating this is a hack due to our activation script misbehaving when switching back and forth.

flokli (changes requested):

nixos/modules/services/system/dbus.nix
SuperSandro2000: Its actually in both cases helping to prevent the issue but for dbus-broker is has a real reason, that other units can just depend on dbus.

flokli:
@SuperSandro2000 please add the NixOS VM test that's been circulating to this PR.

flokli:
@MatthewCroughan did test-drive this PR:

> I tried that PR on my systems, and it seemingly fixed it, though subsequent rebuild switches misbehaved

MatthewCroughan:
FWIW even if this PR did work, one of the instabilities on my systems was such that dns stopped resolving, and so it was only possible to `nix copy` a toplevel to it, and activate that.

SuperSandro2000:
> @SuperSandro2000 please add the NixOS VM test that's been circulating to this PR.

The one that K900 posted yesterday is failing because only the special case for dbus is fixed, not the general issue with aliases.

It could probably be adopted and changed to test dbus specifically though.

MatthewCroughan:
More generally, I think a VM test like `activationTest rev1 rev2`, where `revX` are the ranges of nixpkgs to test activation between is possible, and could catch activation bugs like this earlier.

flokli:
So @MatthewCroughan's situation is another one - he's on 23.11 and explicitly set dbus-broker trying to workaround an issue he wasn't affected with, and switching back from there was his issue. Once tested, this PR should be cherry-picked to stable too, to fix switching behaviours for people opting in and out of dbus-broker on 23.11.

Regarding testing this PR, I'll try to test it this evening or tomorrow with some of my systems that already got migrated to dbus-broker, seeing if with this PR they get moved back to dbus for the next reboot, without breaking during activation.

MatthewCroughan:
No, let me clarify accurately.

I have a mixture of machines, some 23.11, some 23.05 and some unstable. In a desparation to try and get the systems working again, I set `services.dbus.implementation = "broker"`, believing that reverting the revert would fix things. But setting the dbus implementation to broker on 23.11 and earlier, simply introduces this bug to all of them, with no way out.

Now,  I just reintroduced the issue into a machine, and this PR did not fix the issue. The below log is what happens when getting an effected machine to use this PR

```
 '/nix/store/ibjsvc6fj9mlwss5x7fxb3b43spnmhm4-system-path/bin/busctl --json=short call org.freedesktop.systemd1 /org/freedesktop/systemd1 org.freedesktop.systemd1.Manager ListUnitsByPatterns asas 0 0' exited with value 1 at /nix/var/nix/profiles/system/bin/switch-to-configuration line 145.
```

MatthewCroughan:
However, this PR, after effected machines are rebooted, does allow me to switch back and forth between dbus and dbus-broker.

flokli (approved):

flokli:
In case someone gets confused by the output from the activation script: https://github.com/NixOS/nixpkgs/pull/309352#issuecomment-2094914385

SuperSandro2000:
Do we still want to write a test for this?

flokli:
There's rarely a reason to not write a test, to ensure it keeps running. Testing it in a test is much better than testing it on a live system only.

arianvp:
This did not fix the issue for me. At least not when going from a nixpkgs-unstable  checkout with `dbus-broker` still enabled by default to a nixpkgs-unstable checkout with `dbus` enabled by default + this fix. `dbus-broker.service` still got stopped.


this is because the `dbus-broker.service` alias doesn't exist in that commit.

I think we can't fix this properly without telling people to `nixos-rebuild boot`.

PR #309084, opened 2024-05-04
nixos/consul: add options for registering services

mjm:
## Description of changes

This adds options under `services.consul.services` to allow declaratively registering Consul services with the agent. The repetition in the name is a little awkward, but "services" is what they're called, so I think anything else would be confusing.

The config file for services will trigger a reload of Consul when it's changed. This allows services to be registered/deregistered without disrupting others.

For testing, I've been running a version of this for months on my home servers.

Example:

```nix
{
  services.consul.services.miniflux = {
    port = 9999;

    meta.metrics_path = "/metrics";

    checks = [
      {
        name = "miniflux is ready";
        http = "http://localhost:9999/healthcheck";
        interval = "15s";
        timeout = "5s";
      }
    ];
  };
}
```

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/nixos/modules/services/networking/consul.nix
+++ b/nixos/modules/services/networking/consul.nix
@@ -13,7 +13,14 @@ let
     };
   } // cfg.extraConfig;
 
+  jsonFormat = pkgs.formats.json { };
+  enabledServices = attrValues (filterAttrs (_: svc: svc.enable) cfg.services);
+  servicesCfg = jsonFormat.generate "consul-services.json" {
+    services = map (filterAttrs (k: _: k != "enable")) enabledServices;
+  };
+
   configFiles = [ "/etc/consul.json" "/etc/consul-addrs.json" ]
+    ++ (optional (enabledServices != [ ]) "/etc/consul-services.json")
     ++ cfg.extraConfigFiles;
 
   devices = attrValues (filterAttrs (_: i: i != null) cfg.interface);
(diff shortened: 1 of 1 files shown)


emilylange:
How about making this more generic?

There is nothing special about the `service`/`services` config option upstream, so why not make it a single option, a variant of `cfg.extraConfig`, that's placed in `/etc/consul.d/`, but will not trigger a service restart on change?

mjm:
There's a small amount of special behavior here. This option allows specifying services as an attrset, which is a little bit nicer and avoids having to explicitly write `name = "foo"`. And the `id` for each service has a default value based on the service name and the hostname. It's not huge but I find it very nice to use.

I wouldn't mind an option for config that triggers a reload instead of a restart. Any thoughts on what to call such an option?

The `services` option could be built on top of that. But if you don't think that's worth having in NixOS, I can keep that part in my own configs.

emilylange:
Hmm I see.

I have an internal/private abstraction/module that also has static consul service registration and additionally spawns envoy sidecars if needed, so I might be a bit out of touch here.

But I am also in the progress of migrating away from consul, so uhmm

I am afraid I shouldn't review this because of that.

Pinging the consul (package) maintainers, in hope they have time for the consul module:
@adamcstephens @pradeepchhetri @vdemeester @nh2 @techknowlogick

nixos/modules/services/networking/consul.nix
```diff
@@ -13,7 +13,14 @@ let
     };
   } // cfg.extraConfig;
 
+  jsonFormat = pkgs.formats.json { };
+  enabledServices = attrValues (filterAttrs (_: svc: svc.enable) cfg.services);
+  servicesCfg = jsonFormat.generate "consul-services.json" {
+    services = map (filterAttrs (k: _: k != "enable")) enabledServices;
```
adamcstephens, line 19: I'll note that this can be a bit of a footgun. Consul will fail to start if there's a service config that doesn't match its expectations. There's maybe no way around this, but anybody coming later to add an option that consul doesn't like will need to exclude it here too.

adamcstephens (commented):
Overall I think this is a simple way to start building this out. Similar to Emily, I have my own consul connect module. You can see it [here](https://codeberg.org/adamcstephens/stop-export/src/commit/3ee47b0efa462751336bc8b9f0af0eac30af997a/services/consul/default.nix), but for multiple reasons I wouldn't submit it as-is.

I would prefer this to be tested before being merged. Can you extend the consul test suite to also test that a service is correctly registered with this?

nixos/modules/services/networking/consul.nix
mjm: I'm not deeply attached to having the `enable` property if we're worried that sets a precedent for adding Nix-only options here, but I'm not sure how likely that is tbh.

mjm:
Yeah, might take me a few days to get around to it, but I'll try to add a test. Haven't written a NixOS test before but now's as good a time as any.

Issue #98090, opened 2020-09-16
k3s: package k3s-killall.sh script


theduke:
**Describe the bug**

`k3s` provides a a `k3s-killall.sh` script that is required for orderly shutdown. (remove iptables rules, stop running containers, ...). Also `k3s-uninstall.sh` which cleans up all related directories.

Those is currently not included in the `k3s` package.

While not very complicated scripts, it would be nice to have them. (And track upstream changes)

**Notify maintainers**

@euank

theduke:
Note that the scripts are created inside install.sh: https://github.com/rancher/k3s/blob/7ca50efb8b6c5218690e3f1a0a8df4b9843bb3e1/install.sh#L351

euank:
I agree those are handy scripts to have around.
I think we may end up just wanting to fork em since the `systemd` related stuff won't work on nixos I think.

In a normal k3s installation, the systemd unit files are dropped in etc by `install.sh` and can be deleted / stopped reasonably.
On nixos, those systemd unit files are created / started / managed by the nixos module and nixos-rebuild, so having those scripts stop/remove them doesn't seem right.

For the other state nixos doesn't manage (the stuff in /var/lib*, running containers, etc), having those makes sense.

euank:
There's a few additional choices here which add complexity.

It's sometimes desirable for resources to keep running when k3s is stopped, most notably it's usually desirable for kubernetes pods and networking to keep working during a k3s upgrade, during which the service is stopped and started.

The killall script expresses the intent that it's desired for all that stuff to go away, but `systemctl stop k3s` doesn't by itself.

euank:
Still an issue, we just haven't gotten to it yet. Un-stale, stalebot

jceb:
A partial workaround is to change k3s' service configuration to use `KillMode=mixed` as suggested here: https://github.com/k3s-io/k3s/issues/2400#issuecomment-711065914

```nix
systemd.services.k3s.serviceConfig.KillMode = mkForce "mixed";
```

Running `k3s-killall.sh` is still required + even after running it I still find a ton of active containerd units:

```bash
systemctl list-units|grep -i containerd
  cri-containerd-a65c2668f36e2a1eaf27f86440d3fcbede3c408ce7148b735f0818d670c3e6ff.scope
loaded active running   libcontainer container a65c2668f36e2a1eaf27f86440d3fcbede3c408ce7148b735f0818d670c3e6ff
[...]

PR #309099, opened 2024-05-04
k3s: package k3s-killall script

rorosen:
Provide the k3s-killall.sh script for orderly shutdown of k3s.

## Description of changes

This is an attempt to package [k3s-killall.sh](https://docs.k3s.io/upgrades/killall?_highlight=killall#killall-script), which is kinda weird as the killall script is generated by (and contained in) the k3s installer script. I copy the installer script and modify it so I can call only the `create_killall` function which generates the killall script. I'm not completely convinced by this approach tbh, but given the circumstances it's the "best" way I can think about. However, afaict the packaged script behaves the same way as it would in a regular k3s installation.

Additionally, I included the script in the single node test to verify that it's working. 

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [x] Tested, as applicable:

@Mic92 @euank @yajo @superherointj 

## Related Issue

- #98090 

---

### DIFF
--- a/nixos/tests/k3s/single-node.nix
+++ b/nixos/tests/k3s/single-node.nix
@@ -78,6 +78,18 @@ import ../make-test-python.nix ({ pkgs, lib, k3s, ... }:
       # regression test for #176445
       machine.fail("journalctl -o cat -u k3s.service | grep 'ipset utility not found'")
 
+      with subtest("Run k3s-killall"):
+          # Call the killall script with a clean path to assert that
+          # all required commands are wrapped
+          output = machine.succeed("PATH= ${k3s}/bin/k3s-killall.sh 2>&1 | tee /dev/stderr")
+          assert "command not found" not in output, "killall script contains unknown command"
+
+          # Check that killall cleaned up properly
+          machine.fail("systemctl is-active k3s.service")
+          machine.fail("systemctl list-units | grep containerd")
+          machine.fail("ip link show | awk -F': ' '{print $2}' | grep -e flannel -e cni0")
+          machine.fail("ip netns show | grep cni-")
+
       machine.shutdown()
     '';
   })
--- a/pkgs/applications/networking/cluster/k3s/builder.nix
+++ b/pkgs/applications/networking/cluster/k3s/builder.nix
@@ -56,6 +56,14 @@ lib:
 , nixosTests
 , pkgsBuildBuild
 , go
+, runCommand
+, bash
+, procps
+, coreutils
+, gnugrep
+, findutils
+, gnused
+, systemd
 }:
 
 # k3s is a kinda weird derivation. One of the main points of k3s is the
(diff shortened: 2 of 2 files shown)


Mic92:
This could than replace this whole section in the wiki right? https://wiki.nixos.org/wiki/K3s#Cluster_Reset

rorosen:
> This could than replace this whole section in the wiki right? https://wiki.nixos.org/wiki/K3s#Cluster_Reset

Not exactly, the killall script cleans up containers with associated data, networking components, iptables chains etc. The cluster data on server nodes (e.g. cluster datastore and token) will not be deleted.

I think the equivalent of what's described in the wiki would be [k3s-uninstall.sh](https://docs.k3s.io/installation/uninstall?_highlight=uninstall) but that's harder to package than the killall script.

rorosen:
Just noticed that there are leftover containerd units by running `systemctl list-units|grep containerd` after running the killall script. This is different to a regular installation on ubuntu. I'll look into this and convert this PR to draft in the meantime.

rorosen:
There are no more leftover units after killall and the test checks this now as well.

thomasjm:
Thanks for this! Could we make it so that this script runs when you change `services.k3s.enable` from `true` to `false`?

IMO that would be the most Nix-y way for this to work. Hopefully there's an appropriate systemd hook that can run when the k3s service is removed.

rorosen:
The idea is to just package killall so that people have it available. I don't think this should be run automatically if k3s is stopped, as k3s intentionally [keeps containers running when the K3s service is stopped](https://docs.k3s.io/upgrades/killall). However, you could hook it yourself in your config.

```nix
systemd.services = {
  k3s.onSuccess = [ "killall-k3s.service" ];
  killall-k3s.serviceConfig = {
    Type = "oneshot";
    ExecStart = "${k3s}/bin/k3s-killall.sh";
  };
};
``` 

Adding only a `ExecStop` or `ExecStopPost` hook to the k3s service won't work as killall also stops the k3s service. In fact, it stops every service file in `/etc/systemd/system/` that starts with `k3s` so take care that the killall service doesn't start with k3s.

thomasjm:
Hmm, the k3s docs say that containers continue running when the service is stopped "to allow high availability during upgrades."

For the NixOS scenario, is there some way we can keep the containers running while you're updating to a new version of the k3s service, but also kill everything when the system is transitioning to a `services.k3s.enable=false` state? 

I feel it's important for a declarative system like NixOS that if I enable k3s, and then go back to disabling it, my system shouldn't end up in a different state with lots of extra containerd processes running (and these might even persist after a reboot?).

> won't work as killall also stops the k3s service

FWIW when I was experimenting with purging a k3s cluster from my system, I found that I needed to tweak it. It seems likely that we'd want to check in a slightly modified version of k3s-killall.sh in any case.

rorosen:
The scope of this PR is to just package the upstream killall script, only making the required changes so that it works on NixOS the same way it would work in a regular k3s installation.

I don't think it's a big problem to require a manual cleanup after disabling k3s, but maybe this should be discussed in a separate issue.

euank (approved):
This lgtm! I appreciate the testing for it too!

The `sed`ing of the install script is both clever, and seems like it could be fragile if upstream changes that at lot, but we can also just wait and see if it does end up being a maintenance burden

Thank you!

euank:
I'll throw in my own comments about trying to hook this into systemd lifecycle etc.

I think in one sense, it's very nice to make this all fully declarative and make it so disabling the k3s service cleans things up fully.

On the other hand, I think the killall script, as it's offered by upstream, is mainly suitable for manual human usage. It doesn't handle alternate interface names (if you did configure that), it removes all cni network namespaces, even if they're from a different piece of software, it deletes kubernetes iptables rules that likewise could be from another k8s tool.

Basically, it's a pretty blunt tool, and so it feels slightly scary to automatically run it.

It doesn't make it any better, but there are other units that don't fully cleanup, like for example `virtualisation.docker.enable = false;` also leaves behind running containers, iptables rules, etc.

thomasjm (approved):
Thanks for doing this!

One optional piece of feedback: I agree with @euank that the `sed` stuff looks fragile. I think it would be cleaner to just inline the script with the necessary changes for NixOS, for two reasons: 1) it's not very long, and 2) as previously mentioned, it makes some sweeping changes to the system so it might be better for maintainers to have the contents front and center so they can see what they're dealing with.

superherointj:
> I'm happy to file a follow-up about hooking this into the systemd lifecycle.

@thomasjm @euank A compromise could be making it optional in the NixOS module.

yajo:
In case it helps, I think [there are some packages or options that mark a NixOS upgrade as "it needs to reboot"](https://nixos.org/manual/nixos/stable/#sec-upgrading-automatic).

If that makes sense and is possible, you can set k3s to trigger the "needs reboot" state when disabled. Rebooting will also kill all, and after the reboot no new systemd units should start, so IMHO that would fix the non-declarativity problem.

I think that a module could enable k3s-killall on system shutdown, similar to what is explained in https://github.com/NixOS/nixpkgs/issues/98090#issuecomment-1980273467. That module should be separate from the k3s module itself, because otherwise, disabling k3s would disable the killall module and it wouldn't kill after disabled.

rorosen:
Thanks for the reviews!

I'm also not super happy with the `sed`ing but copying the script would make it very possible to miss upstream changes. The script hasn't been changed upstream that often recently, but once in a while there are smaller adjustments. I also thought about patching the script, but I'm afraid that this may requires several diffs for different versions.

I tried to make the `sed` stuff less fragile by checking whether the expected patterns are really found and erroring if not. Still not stable but I hope that major bugs will be detected together with the tests.

yajo (approved):

superherointj (approved):
LGTM.

@rorosen Thanks for the contribution.

PR #309298, opened 2024-05-05
whitesur-cursors: fix version, do small cleanup

TomaSajt:
## Description of changes

The main reason for this PR is to correct the version of the `whitesur-cursors` package to begin with a number.

Other small changes
- format with `nixfmt`
- use `hash` instead of `sha256`
- use `mkdir -p` instead of `install -d`

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/data/icons/whitesur-cursors/default.nix
+++ b/pkgs/data/icons/whitesur-cursors/default.nix
@@ -1,17 +1,18 @@
-{ lib
-, stdenvNoCC
-, fetchFromGitHub
+{
+  lib,
+  stdenvNoCC,
+  fetchFromGitHub,
 }:
 
 stdenvNoCC.mkDerivation {
   pname = "whitesur-cursors";
-  version = "unstable-2022-06-17";
+  version = "0-unstable-2022-06-17";
 
   src = fetchFromGitHub {
     owner = "vinceliuice";
     repo = "WhiteSur-cursors";
     rev = "5c94e8c22de067282f4cf6d782afd7b75cdd08c8";
-    sha256 = "sha256-CFse0XZzJu+PWDcqmvIXvue+3cKX47oavZU9HYRDAg0=";
+    hash = "sha256-CFse0XZzJu+PWDcqmvIXvue+3cKX47oavZU9HYRDAg0=";
   };
 
   installPhase = ''


pkgs/data/icons/whitesur-cursors/default.nix
```diff
@@ -1,22 +1,23 @@
...
     repo = "WhiteSur-cursors";
     rev = "5c94e8c22de067282f4cf6d782afd7b75cdd08c8";
-    sha256 = "sha256-CFse0XZzJu+PWDcqmvIXvue+3cKX47oavZU9HYRDAg0=";
+    hash = "sha256-CFse0XZzJu+PWDcqmvIXvue+3cKX47oavZU9HYRDAg0=";
   };
 
   installPhase = ''
     runHook preInstall
-    install -dm 755 $out/share/icons/WhiteSur-cursors
+    mkdir -p $out/share/icons/WhiteSur-cursors
```
Aleksanaa, line 20: Why is this?

Aleksanaa: We may be more inclined to write `install` (instead of `mkdir` and `cp`) at present, although I may not be able to find the tracking issue now

TomaSajt: Note that this is not `install -D` but `install -d` which is AFAIK just directory creation.
Sadly, `install -D` doesn't support moving nested directories, that's why I had to use `cp`.

Aleksanaa: So since there is nothing wrong with it, why do you want to change it?

Aleksanaa: I mean, since it works normally and looks normally, there is no need to change it. What if another person comes and changes it back to `install` tomorrow?

TomaSajt: Okay, I think we won't really agree on this, so I'll revert this change, it's not that important.

PR #309318, opened 2024-05-05
nixos/plasma6: add CAP_SYS_NICE for kwin_wayland

oxalica:
## Description of changes

This makes kwin_wayland successfully gain SCHED_RR priority, which significantly improves the rendering lag, ie. cursor movement and desktop animations, under heavy CPU load like on compilation. This was already enabled for plasma5 but is not for plasma6 somehow.

The capability lowering patch is copied unchanged [from plasma5](https://github.com/NixOS/nixpkgs/blob/def8d7e2549f2b2eb54d1115833cd814ffb0ba14/pkgs/desktops/plasma-5/kwin/0001-Lower-CAP_SYS_NICE-from-the-ambient-set.patch). With the patch, it seems that applications launched by global shortcuts having this:
```
$ getpcaps $(pgrep alacritty)
2607: cap_setpcap,cap_sys_nice=i
```

So these capabilities are inheritable but not effective nor permitted. I guess it means nothing? cc the original patch author @alois31

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [x] `sandbox = true`
- [x] Tested, as applicable:
  - Tested with plasma6 desktop environment.

---

### DIFF
--- a/nixos/modules/services/desktop-managers/plasma6.nix
+++ b/nixos/modules/services/desktop-managers/plasma6.nix
@@ -286,6 +286,15 @@ in {
       kde-smartcard = lib.mkIf config.security.pam.p11.enable { p11Auth = true; };
     };
 
+    security.wrappers = {
+      kwin_wayland = {
+        owner = "root";
+        group = "root";
+        capabilities = "cap_sys_nice+ep";
+        source = "${lib.getBin pkgs.kdePackages.kwin}/bin/kwin_wayland";
+      };
+    };
+
     programs.dconf.enable = true;
 
     programs.firefox.nativeMessagingHosts.packages = [kdePackages.plasma-browser-integration];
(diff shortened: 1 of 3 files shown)


K900:
I don't think the wrapper is enough, it needs to also be added to the systemd units to actually run the right thing.

oxalica:
> I don't think the wrapper is enough, it needs to also be added to the systemd units to actually run the right thing.

The description in the PR description is what I observed in my plasma6 session, thus I'm sure it's working.

> This makes kwin_wayland successfully gain SCHED_RR priority, which significantly improves the rendering lag, ie. cursor movement and desktop animations, under heavy CPU load like on compilation.

Currently the systemd unit `plasma-kwin_wayland.service` points to `${kwin}/bin/kwin_wayland_wrapper --xwayland`. The wrapper is not privileged but it calls `kwin_wayland --wayland-fd [..] --socket [..] [..more complex flags]` [here](https://invent.kde.org/plasma/kwin/-/blob/v6.0.4/src/helpers/wayland_wrapper/kwin_wrapper.cpp#L94) using a non-absolute path. The main process is found in the wrapper and gets capabilities correctly.

Do you think it's unexpected to work in this way, and we should patch the path in kwin_wayland_wrapper to call `/run/wrapper/bin/kwin_wayland`? But that will make `kwin` not working without the plasma NixOS module enabled or outside NixOS.

K900:
OK, no, the added wrapper makes it fine.

PR #309424, opened 2024-05-06
treewide: ReadWriteDirectories -> ReadWritePaths

nh2:
## Description of changes

These were renamed in systemd v231:

https://github.com/systemd/systemd/commit/2a624c36e646e9ef8d204a506b12e7dbd380e111

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [x] Tested, as applicable:
  - [x] [NixOS test(s)](https://nixos.org/manual/nixos/unstable/index.html#sec-nixos-tests) (look inside [nixos/tests](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests))

---

CC maintainers:

* `caddy`: @Br1ght0ne @emilylange @techknowlogick
* `trafic`: @vdemeester
* `zoneminder` has no maintainers

### DIFF
--- a/nixos/modules/services/security/step-ca.nix
+++ b/nixos/modules/services/security/step-ca.nix
@@ -107,7 +107,7 @@ in
           UMask = "0077";
           Environment = "HOME=%S/step-ca";
           WorkingDirectory = ""; # override upstream
-          ReadWriteDirectories = ""; # override upstream
+          ReadWritePaths = ""; # override upstream
 
           # LocalCredential handles file permission problems arising from the use of DynamicUser.
           LoadCredential = "intermediate_password:${cfg.intermediatePasswordFile}";
--- a/nixos/modules/services/web-servers/caddy/default.nix
+++ b/nixos/modules/services/web-servers/caddy/default.nix
@@ -368,7 +368,7 @@ in
         ExecReload = [ "" ''${cfg.package}/bin/caddy reload ${runOptions} --force'' ];
         User = cfg.user;
         Group = cfg.group;
-        ReadWriteDirectories = cfg.dataDir;
+        ReadWritePaths = [ cfg.dataDir ];
         StateDirectory = mkIf (cfg.dataDir == "/var/lib/caddy") [ "caddy" ];
         LogsDirectory = mkIf (cfg.logDir == "/var/log/caddy") [ "caddy" ];
         Restart = "on-failure";
--- a/nixos/modules/services/web-servers/caddy/default.nix
+++ b/nixos/modules/services/web-servers/caddy/default.nix
@@ -360,6 +360,7 @@ in
       serviceConfig = let
         runOptions = ''--config ${configPath} ${optionalString (cfg.adapter != null) "--adapter ${cfg.adapter}"}'';
       in {
+        # Override the `ExecStart` line from upstream's systemd unit file by our own:
         # https://www.freedesktop.org/software/systemd/man/systemd.service.html#ExecStart=
         # If the empty string is assigned to this option, the list of commands to start is reset, prior assignments of this option will have no effect.
         ExecStart = [ "" ''${cfg.package}/bin/caddy run ${runOptions} ${optionalString cfg.resume "--resume"}'' ];
--- a/nixos/modules/services/web-servers/traefik.nix
+++ b/nixos/modules/services/web-servers/traefik.nix
@@ -170,7 +170,7 @@ in {
         PrivateDevices = true;
         ProtectHome = true;
         ProtectSystem = "full";
-        ReadWriteDirectories = cfg.dataDir;
+        ReadWritePaths = [ cfg.dataDir ];
         RuntimeDirectory = "traefik";
       };
     };
--- a/nixos/modules/services/misc/zoneminder.nix
+++ b/nixos/modules/services/misc/zoneminder.nix
@@ -350,7 +350,7 @@ in {
           RestartSec = "10s";
           CacheDirectory = dirs cacheDirs;
           RuntimeDirectory = dirName;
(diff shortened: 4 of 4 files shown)


techknowlogick (approved):
LGTM. Does this need to be added to the changelog?

nixos/modules/services/security/step-ca.nix
```diff
@@ -108,6 +108,9 @@ in
           Environment = "HOME=%S/step-ca";
           WorkingDirectory = ""; # override upstream
           ReadWriteDirectories = ""; # override upstream
+          # systemd renamed the above to `ReadWritePaths`, but upstream
+          # still uses `ReadWriteDirectories`, and it is unclear if the
+          # override works when the alias is used.
```
emilylange, line 113:
```suggestion
          ReadWritePaths = ""; # override upstream
```

systemd handles aliases properly.

This works as expected.

You can confirm this interactively in the NixOS VM test using `systemctl cat step-ca` and `systemctl show step-ca --all`.

nixos/modules/services/web-servers/caddy/default.nix
```diff
@@ -360,14 +360,15 @@ in
...
+        # Override the `ExecStart` line from upstream's systemd unit file by our own:
         # https://www.freedesktop.org/software/systemd/man/systemd.service.html#ExecStart=
         # If the empty string is assigned to this option, the list of commands to start is reset, prior assignments of this option will have no effect.
         ExecStart = [ "" ''${cfg.package}/bin/caddy run ${runOptions} ${optionalString cfg.resume "--resume"}'' ];
         # Validating the configuration before applying it ensures we’ll get a proper error that will be reported when switching to the configuration
         ExecReload = [ "" ''${cfg.package}/bin/caddy reload ${runOptions} --force'' ];
         User = cfg.user;
         Group = cfg.group;
-        ReadWriteDirectories = cfg.dataDir;
+        ReadWritePaths = [ cfg.dataDir ];
```
emilylange, line 371:
```suggestion
        ReadWritePaths = cfg.dataDir;
```

Why did you make this a list?

nixos/modules/services/web-servers/traefik.nix
```diff
@@ -170,7 +170,7 @@ in {
         PrivateDevices = true;
         ProtectHome = true;
         ProtectSystem = "full";
-        ReadWriteDirectories = cfg.dataDir;
+        ReadWritePaths = [ cfg.dataDir ];
```
emilylange, line 173:
```suggestion
        ReadWritePaths = cfg.dataDir;
```

Same as above.

emilylange (changes requested):
Please also adjust the PR title, thanks :)

nixos/modules/services/web-servers/caddy/default.nix
```diff
@@ -360,14 +360,15 @@ in
       serviceConfig = let
         runOptions = ''--config ${configPath} ${optionalString (cfg.adapter != null) "--adapter ${cfg.adapter}"}'';
       in {
+        # Override the `ExecStart` line from upstream's systemd unit file by our own:
```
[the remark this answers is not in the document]
nh2: @emilylange Because the code and existing comment are unclear to anybody who didn't write them. They explain how one can reset a unit's `ExecStart`, but not _why_ we want to do that.

The info the reader needs is that we re-use upstream's unit file. That is not obvious. It is implicit by `systemd.packages = [ cfg.package ];`. It should be obvious on first read.

nixos/modules/services/security/step-ca.nix
nh2: Thanks, I'll use that then!

nixos/modules/services/web-servers/caddy/default.nix
nh2: Because it is a list of paths, and if I'm not mistaken, merges as such. Most usages of `ReadWritePaths` in nixpkgs use the proper type, only a few use the fact that a single element is also assignable for systemd units.

nh2:
> Does this need to be added to the changelog?

@techknowlogick I don't think so, because it is an implementation detail of these units. 

As @emilylange confirmed in https://github.com/NixOS/nixpkgs/pull/309424#discussion_r1591383735, it everything still works as expected even if a user overrides a setting to its old name, so this change is only a cleanup to not refer to legacy names should be completely transparent to the users.

nh2:
> Please also adjust the PR title, thanks :)

@emilylange Ah right, thanks!

nixos/modules/services/web-servers/caddy/default.nix
emilylange: Makes sense. I am just usually a bit cautious when changing a type, but this should be perfectly fine here.

Thanks for responding :)

emilylange (approved):

emilylange:
Ah, one last thing:

Your commit message isn't mentioning the `nixos/step-ca` change.
You can either extend your list, or simply write `treewide: ReadWriteDirectories -> ReadWritePaths`

techknowlogick:
Awesome. Thanks @nh2 for this PR, and @emilylange for your review of it :)

nh2:
> You can either extend your list, or simply write `treewide: ReadWriteDirectories -> ReadWritePaths`

Good point, now indeed it is treewide.

emilylange:
One just noticing:
Please create the branch in your fork instead of NixOS/nixpkgs next time.

PR #308500, opened 2024-05-02
nixos/ebusd: clean up module

bobrippling:
## Description of changes

This is a rebase of #264893 with conflicts with other changes to `ebusd.nix` resolved.

### Original PR description

This PR only cleans up the code of the `ebusd` module, with the intention of not changing the outcome.

In general, I've cleaned up a bunch of descriptions, inlined a few single element lists, removed some extra space. The two big changes are the automatic generation of the `services.ebusd.logs` options, and the use of `cli.toGNUCommandLineShell` for constructing the string of arguments. The latter has also been moved from the top of the file, closer to the `ExecStart` where it is used.

I've tested that the code evalutates manually, but it would be nice if this had a basic test. Not sure how easy that would be to make though, considering it's a hardware related program.

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
- For non-Linux: Is sandboxing enabled in `nix.conf`? (See [Nix manual](https://nixos.org/manual/nix/stable/command-ref/conf-file.html))
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:
  - [NixOS test(s)](https://nixos.org/manual/nixos/unstable/index.html#sec-nixos-tests) (look inside [nixos/tests](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests))
  - and/or [package tests](https://github.com/NixOS/nixpkgs/blob/master/pkgs/README.md#package-tests)
  - or, for functions and "core" functionality, tests in [lib/tests](https://github.com/NixOS/nixpkgs/blob/master/lib/tests) or [pkgs/test](https://github.com/NixOS/nixpkgs/blob/master/pkgs/test)
  - made sure NixOS tests are [linked](https://nixos.org/manual/nixpkgs/unstable/#ssec-nixos-tests-linking) to the relevant packages
- [ ] Tested compilation of all packages that depend on this change using `nix-shell -p nixpkgs-review --run "nixpkgs-review rev HEAD"`. Note: all changes have to be committed, also see [nixpkgs-review usage](https://github.com/Mic92/nixpkgs-review#usage)
- [ ] Tested basic functionality of all binary files (usually in `./result/bin/`)
[...]

PR #264893, opened 2023-11-01
nixos/ebusd: clean up module

h7x4:
## Description of changes

This PR only cleans up the code of the `ebusd` module, with the intention of not changing the outcome.

In general, I've cleaned up a bunch of descriptions, inlined a few single element lists, removed some extra space. The two big changes are the automatic generation of the `services.ebusd.logs` options, and the use of `cli.toGNUCommandLineShell` for constructing the string of arguments. The latter has also been moved from the top of the file, closer to the `ExecStart` where it is used.

I've tested that the code evalutates manually, but it would be nice if this had a basic test. Not sure how easy that would be to make though, considering it's a hardware related program.

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
- For non-Linux: Is sandboxing enabled in `nix.conf`? (See [Nix manual](https://nixos.org/manual/nix/stable/command-ref/conf-file.html))
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:
  - [NixOS test(s)](https://nixos.org/manual/nixos/unstable/index.html#sec-nixos-tests) (look inside [nixos/tests](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests))
  - and/or [package tests](https://nixos.org/manual/nixpkgs/unstable/#sec-package-tests)
  - or, for functions and "core" functionality, tests in [lib/tests](https://github.com/NixOS/nixpkgs/blob/master/lib/tests) or [pkgs/test](https://github.com/NixOS/nixpkgs/blob/master/pkgs/test)
  - made sure NixOS tests are [linked](https://nixos.org/manual/nixpkgs/unstable/#ssec-nixos-tests-linking) to the relevant packages
- [ ] Tested compilation of all packages that depend on this change using `nix-shell -p nixpkgs-review --run "nixpkgs-review rev HEAD"`. Note: all changes have to be committed, also see [nixpkgs-review usage](https://github.com/Mic92/nixpkgs-review#usage)
- [ ] Tested basic functionality of all binary files (usually in `./result/bin/`)
- [23.11 Release Notes](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2311.section.md) (or backporting [23.05 Release notes](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2305.section.md))
[...]

PR #309517, opened 2024-05-06
Fix ebusd service argument passing

bobrippling:
## Description of changes

This resolves a bug introduced in #308500 / #264893, where ebusd arguments were changed from `--scanconfig=all` to `--scanconfig all` with the use of `cli.toGNUCommandLineShell`.

The bug is an unfortunate result of `ebusd`'s argument parsing - the arguments must be [separated by equals signs](https://github.com/john30/ebusd/blob/c084dd2cbb3411883bd1b048585724a93e1aff83/src/ebusd/main_args.cpp#L658).

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

### DIFF
--- a/lib/cli.nix
+++ b/lib/cli.nix
@@ -90,7 +90,16 @@ rec {
     mkOption ?
       k: v: if v == null
             then []
-            else [ (mkOptionName k) (lib.generators.mkValueStringDefault {} v) ]
+            else if optionValueSeparator == null then
+              [ (mkOptionName k) (lib.generators.mkValueStringDefault {} v) ]
+            else
+              [ "${mkOptionName k}${optionValueSeparator}${lib.generators.mkValueStringDefault {} v}" ],
+
+    # how to separate an option from its flag;
+    # by default, there is no separator, so option `-c` and value `5`
+    # would become ["-c" "5"].
+    # This is useful if the command requires equals, for example, `-c=5`.
+    optionValueSeparator ? null
     }:
     options:
       let
--- a/lib/tests/misc.nix
+++ b/lib/tests/misc.nix
@@ -1639,6 +1639,27 @@ runTests {
     ];
   };
 
+  testToGNUCommandLineSeparator = {
+    expr = cli.toGNUCommandLine { optionValueSeparator = "="; } {
+      data = builtins.toJSON { id = 0; };
+      X = "PUT";
+      retry = 3;
+      retry-delay = null;
+      url = [ "https://example.com/foo" "https://example.com/bar" ];
+      silent = false;
+      verbose = true;
+    };
+
+    expected = [
+      "-X=PUT"
+      "--data={\"id\":0}"
+      "--retry=3"
+      "--url=https://example.com/foo"
+      "--url=https://example.com/bar"
+      "--verbose"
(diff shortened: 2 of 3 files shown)


bobrippling:
I'm not familiar with the nix language and suspect this fix may not be the best approach, open to suggestions.

@nathan-gs are you able to recommend anyone that might help with improving my fix?

nathan-gs:
For me it looks good! However let's also have a review from some experts of `cli.toGNUCommandLineShell`, just to be sure (@Gabriella439 / @Profpatsch)?

nathan-gs (approved):

Gabriella439 (approved):

lib/cli.nix
```diff
@@ -90,15 +90,33 @@ rec {
     mkOption ?
       k: v: if v == null
             then []
-            else [ (mkOptionName k) (lib.generators.mkValueStringDefault {} v) ]
+            else [ (mkOptionName k) (lib.generators.mkValueStringDefault {} v) ],
```
infinisil, line 93: The current implementation looks rather backwards. Shouldn't something simple like

```suggestion
            else if optionValueSeparator == null then
              [ (mkOptionName k) (lib.generators.mkValueStringDefault {} v) ]
            else
              [ "${mkOptionName k}${optionValueSeparator}${lib.generators.mkValueStringDefault {} v}" ],
```

work?

infinisil (changes requested):

lib/cli.nix
bobrippling: Thank you, yes that's much better

nathan-gs (approved):

lib/cli.nix
```diff
@@ -90,7 +90,16 @@ rec {
...
-            else [ (mkOptionName k) (lib.generators.mkValueStringDefault {} v) ]
+            else if optionValueSeparator == null then
+              [ (mkOptionName k) (lib.generators.mkValueStringDefault {} v) ]
+            else
+              [ "${mkOptionName k}${optionValueSeparator}${lib.generators.mkValueStringDefault {} v}" ],
+
+    # how to separate an option from its flag;
+    # by default, there is no separator, so option `-c` and value `5`
+    # would become ["-c" "5"].
+    # This is useful if the command requires equals, for example, `-c=5`.
```
infinisil, line 101: Unfortunately this is currently not rendered in the reference manual. If you're interested in improving that, feel free to make a PR to move these into the doc comment above the function! Check out the part on 'function arguments' [here](https://github.com/NixOS/nixpkgs/tree/master/doc#documentation-conventions) to see the convention for that :)

infinisil (approved):
Looking good!

lib/cli.nix
bobrippling: Thanks for the tip - would you mind taking a look at #315820? Also, do you have a link to the reference doc? I can't find where `cli.nix` appears when I've been searching

PR #309551, opened 2024-05-06
nixos/tailscale: add extraSetFlags to configure daemon

### LINKED ISSUE #276912: tailscale: package is not using option extraUpFlags
### Describe the bug
Tailscale package is not using the `extraUpFlags` unless `authKeyFile` is set.

### Steps To Reproduce
1. Enable tailscale and set a flag
``` 
  services.tailscale.enable = true;
  services.tailscale.extraUpFlags = ["--ssh" ];
```
2. `nixos-rebuild switch`
3. `ps -ef | grep tailscaled`  ==> no flags are used

### Expected behavior
Flags are used

### Additional context
Code reference: https://github.com/NixOS/nixpkgs/blob/a2c4338855f665358aabd3a307d640d3ce6abd56/nixos/modules/services/networking/tailscale.nix#L112

extraUpFlags is only used inside the if of auth key not being null check.

### Notify maintainers
@evandeaubl @jwygoda @Enzime 

### Metadata

```console
 - system: `"x86_64-linux"`
 - host os: `Linux 6.1.68, NixOS, 23.11 (Tapir), 23.11.2130.d65bceaee0fb`
 - multi-user?: `yes`
 - sandbox: `yes`
```

---

aca:
## Description of changes

`services.tailscale.extraUpFlags` is valid only when `services.tailscale.authKeyFile` is set.
This change makes `extraUpFlags` work without `authKeyFile`.
`extraUpFlags` can be used to configure tailscale daemon and `authKey` is not an required option.

Fix https://github.com/nixos/nixpkgs/issues/276912

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/nixos/modules/services/networking/tailscale.nix
+++ b/nixos/modules/services/networking/tailscale.nix
@@ -61,10 +61,19 @@ in {
     };
 
     extraUpFlags = mkOption {
-      description = "Extra flags to pass to {command}`tailscale up`.";
+      description = ''
+        Extra flags to pass to {command}`tailscale up`. Only applied if `authKeyFile` is specified.";
+      '';
       type = types.listOf types.str;
       default = [];
       example = ["--ssh"];
+    };
+
+    extraSetFlags = mkOption {
+      description = "Extra flags to pass to {command}`tailscale set`.";
+      type = types.listOf types.str;
+      default = [];
+      example = ["--advertise-exit-node"];
     };
 
     extraDaemonFlags = mkOption {
@@ -117,6 +126,18 @@ in {
         if [[ $status != Connected* ]]; then
           ${cfg.package}/bin/tailscale up --auth-key 'file:${cfg.authKeyFile}' ${escapeShellArgs cfg.extraUpFlags}
         fi
+      '';
+    };
+
+    systemd.services.tailscaled-set = mkIf (cfg.extraSetFlags != []) {
+      after = ["tailscaled.service"];
+      wants = ["tailscaled.service"];
+      wantedBy = [ "multi-user.target" ];
+      serviceConfig = {
+        Type = "oneshot";
+      };
+      script = ''
+        ${cfg.package}/bin/tailscale set ${escapeShellArgs cfg.extraSetFlags}
       '';
     };
 


nixos/modules/services/networking/tailscale.nix
```diff
@@ -115,7 +115,9 @@ in {
       script = ''
         status=$(${config.systemd.package}/bin/systemctl show -P StatusText tailscaled.service)
         if [[ $status != Connected* ]]; then
```
misuzu, line 117: What is the usecase though? This will only work if tailscale is not connected yet, but in this case it won't work without auth-key

aca: I used it to set option like this.
```
services.tailscale.enable = true;
services.tailscale.useRoutingFeatures = "both";
services.tailscale.extraUpFlags = [ "--ssh" "--advertise-exit-node" ];
```

I think you are correct. I misunderstood the action. Didn't know there was set command.

aca: Created new option `extraSetFlags`.

aca:
@misuzu I'm sorry to bother you but is there a way to get this reviewed?

misuzu (approved):

nixos/modules/services/networking/tailscale.nix
```diff
@@ -61,12 +61,21 @@ in {
...
       default = [];
       example = ["--ssh"];
     };
 
+    extraSetFlags = mkOption {
+      description = "Extra flags to pass to {command}`tailscale set`.";
+      type = types.listOf types.str;
+      default = [];
+      example = ["--advertise-exit-node"];
+    };
```
SuperSandro2000, line 77: I am not convinced that we need the second option. If you once logged into tailscale via headscale, the up command also works as expected.

SuperSandro2000: I've took a 2nd look at the condition inside of extraUpFlags and I think this is correct now.

nixos/modules/services/networking/tailscale.nix
```diff
@@ -120,6 +129,18 @@ in {
...
     };
 
+    systemd.services.tailscaled-set = mkIf (cfg.extraSetFlags != []) {
+      after = ["tailscaled.service"];
+      wants = ["tailscaled.service"];
+      wantedBy = [ "multi-user.target" ];
+      serviceConfig = {
+        Type = "oneshot";
+      };
+      script = ''
```
SuperSandro2000, line 139: Do we need a condition here like above or not? I don't know.

SuperSandro2000 (approved):

nixos/modules/services/networking/tailscale.nix
DerRockWolf: I still don' 100% see the usecase.
If we think about a scenario where we deploy a new server with our NixOS configuration and use `extraSetFlags` we would expect that once the server is deployed we only need to SSH into it run `tailscale up`, login and authorize the new machine and after that the server should e.g., advertise itself as an exit node.

This, at least in my tests, didn't work. The changes made by `tailscale set` are simply reverted when doing the first `tailscale up`. After doing `tailscale up` we could obviously do `systemctl restart tailscaled-set.service` and it would work again, but I don't think this is the desired behavior.

It would be great if someone could also try to reproduce this. I used `tailscale debug prefs` to print the currently used settings.

aca: I wrote this PR because I was not using "authkey" option and there was no way to configure daemon in nix. #276912 
Rebuild / Reboot or restarting oneshot service as you mentioned makes this option work. 
I admit that this option is tricky.. but couldn't find better solution..

```
services.tailscale.enable = true;
services.tailscale.useRoutingFeatures = "both";
services.tailscale.extraSetFlags = ["--ssh" "--advertise-exit-node=true"];
```

DerRockWolf: Ok, then we should definitely document that a restart of the oneshot service is required **after** running `tailscale up`.
Because currently it's not obvious when/how this option works and it's sadly also not at all declarative 😔

PR #307540, opened 2024-04-28
zlib: link with --undefined-version on lld

rhelmot:
## Description of changes

zlib is one of many packages impacted by changes to lld's default flags, wherein it will throw errors unless linked with --undefined-verison or fixed to set its version info correctly. This does the former. This is necessary to build on freebsd.

## Things done

- Built on platform(s)
  - [X] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
- For non-Linux: Is sandboxing enabled in `nix.conf`? (See [Nix manual](https://nixos.org/manual/nix/stable/command-ref/conf-file.html))
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:
  - [NixOS test(s)](https://nixos.org/manual/nixos/unstable/index.html#sec-nixos-tests) (look inside [nixos/tests](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests))
  - and/or [package tests](https://github.com/NixOS/nixpkgs/blob/master/pkgs/README.md#package-tests)
  - or, for functions and "core" functionality, tests in [lib/tests](https://github.com/NixOS/nixpkgs/blob/master/lib/tests) or [pkgs/test](https://github.com/NixOS/nixpkgs/blob/master/pkgs/test)
  - made sure NixOS tests are [linked](https://nixos.org/manual/nixpkgs/unstable/#ssec-nixos-tests-linking) to the relevant packages
- [ ] Tested compilation of all packages that depend on this change using `nix-shell -p nixpkgs-review --run "nixpkgs-review rev HEAD"`. Note: all changes have to be committed, also see [nixpkgs-review usage](https://github.com/Mic92/nixpkgs-review#usage)
- [ ] Tested basic functionality of all binary files (usually in `./result/bin/`)
- [24.05 Release Notes](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2405.section.md) (or backporting [23.05](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2305.section.md) and [23.11](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2311.section.md) Release notes)
  - [ ] (Package updates) Added a release notes entry if the change is major or breaking
  - [ ] (Module updates) Added a release notes entry if the change is significant
[...]

PR #309564, opened 2024-05-06
sysbench: Use vendored statically linked libck

avnik:
Motivation: use shared version of libck in sysbench affect on measurement results too much

## Description of changes

This partially revert and redo 4af13554542ef24b20acb3a023c3f7a34f845cff 
Vendored version of concurrency kit (aka libck) require little more patching to support cross-compilation

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [x] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/development/tools/misc/sysbench/default.nix
+++ b/pkgs/development/tools/misc/sysbench/default.nix
@@ -5,7 +5,6 @@
 , pkg-config
 , libmysqlclient
 , libaio
-, libck
 , luajit
 # For testing:
 , testers
@@ -17,7 +16,7 @@ stdenv.mkDerivation rec {
   version = "1.0.20";
 
   nativeBuildInputs = [ autoreconfHook pkg-config ];
-  buildInputs = [ libmysqlclient luajit libck ] ++ lib.optionals stdenv.isLinux [ libaio ];
+  buildInputs = [ libmysqlclient luajit ] ++ lib.optionals stdenv.isLinux [ libaio ];
   depsBuildBuild = [ pkg-config ];
 
   src = fetchFromGitHub {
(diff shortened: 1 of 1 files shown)


pkgs/development/tools/misc/sysbench/default.nix
```diff
@@ -33,11 +32,28 @@ stdenv.mkDerivation rec {
...
   ];
 
+  # Vendored libck version require more hacks than regular nixpkgs' one.
+  # Sysbench related on statically linked vendored libck.
+  postPatch = ''
+    substituteInPlace \
+      third_party/concurrency_kit/ck/configure \
+        --replace \
+          'COMPILER=`./.1 2> /dev/null`' \
+          "COMPILER=gcc" \
```
rrbutani, line 46: I'm maybe missing something here; is there a reason `COMPILER` is being set to `gcc` instead of `$CC` or `${stdenv.cc.targetPrefix}cc`?

rrbutani:
Just for posterity: `pkgs.libck` actually does have static lib outputs:
```console
❯ ls $(nix build nixpkgs#libck --print-out-paths --no-link)/lib
libck.a  libck.so  libck.so.0  libck.so.0.7.1  pkgconfig
```
But its pkg-config file (`pkgconfig/ck.pc`) isn't configured so that `pkg-config --libs ck --static` returns linker flags that use `libck.a`. 

I don't see a good way to get `sysbench`'s build to use the "system" provided `libck.a` — we could [override `CK_LIBS`](https://github.com/akopytov/sysbench/blob/9ffd5873cf217307557e20e8a4360c4994630146/m4/sb_concurrency_kit.m4#L40-L41) manually but that seems like a worse solution than using the bundled `libck` (as you've done in this PR).

rrbutani:
@ofborg build pkgsCross.aarch64-multiplatform.sysbench pkgsCross.aarch64-multiplatform.pkgsLLVM.sysbench

rrbutani (approved):
When stacked on #307540, #309884, and #309887 I can confirm that `pkgsLLVM.sysbench`, `pkgsCross.aarch64-multiplatform.sysbench`, and `pkgsCross.aarch64-multiplatform.pkgsLLVM.sysbench` build successfully (on `x86_64`).

pkgs/development/tools/misc/sysbench/default.nix
```diff
@@ -33,11 +32,28 @@ stdenv.mkDerivation rec {
...
+        --replace \
+          'PLATFORM=`uname -m 2> /dev/null`' \
+          "PLATFORM=${stdenv.hostPlatform.parsed.cpu.name}"
+    substituteInPlace \
+      third_party/concurrency_kit/ck/src/Makefile.in \
+        --replace \
```
Mic92, line 52:
```suggestion
        --replace-fail \
```

pkgs/development/tools/misc/sysbench/default.nix
```diff
@@ -33,11 +32,28 @@ stdenv.mkDerivation rec {
...
     "--with-mysql-includes=${lib.getDev libmysqlclient}/include/mysql"
     "--with-mysql-libs=${libmysqlclient}/lib/mysql"
   ];
 
+  # Vendored libck version require more hacks than regular nixpkgs' one.
+  # Sysbench related on statically linked vendored libck.
+  postPatch = ''
+    substituteInPlace \
+      third_party/concurrency_kit/ck/configure \
+        --replace \
```
Mic92, line 44:
```suggestion
        --replace-fail \
```

pkgs/development/tools/misc/sysbench/default.nix
Mic92: Fixed.

Mic92: We support mingw64 as well, but it's very experimental, so I don't think we need to fix sysbench for it just now.

Mic92: Mhm. Actually this doesn't work for cross compiling.

brianmcgillion (approved):
Verified on Nvidia jetson agx/nx cross build

PR #189398, opened 2022-09-02
nixos/steam: support adding additional compatibility packages

romatthe:
###### Description of changes

Valve allows users to use other compatibility tools with Steam aside from the versions of Proton which they ship themselves. This is especially useful for using versions of Proton, Wine (proton-ge, wine-staging) or other tools (boxtron, luxtorpeda, roberta, etc.) which you've built yourself of pulled from packages not maintained by Valve.

More details can be found here: https://github.com/ValveSoftware/steam-for-linux/issues/6310

This PR allows users to set which additional packages they want Steam to search when listing additional compatibility tools. 

**NOTE**: This is my first modules PR and could probably use some help, any feedback is appreciated. I've indicated where I had my doubts.

Please let me know if this needs to target `staging` instead of `master`.

###### Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
- [ ] For non-Linux: Is `sandbox = true` set in `nix.conf`? (See [Nix manual](https://nixos.org/manual/nix/stable/command-ref/conf-file.html))
- [ ] Tested, as applicable:
  - [NixOS test(s)](https://nixos.org/manual/nixos/unstable/index.html#sec-nixos-tests) (look inside [nixos/tests](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests))
  - and/or [package tests](https://nixos.org/manual/nixpkgs/unstable/#sec-package-tests)
  - or, for functions and "core" functionality, tests in [lib/tests](https://github.com/NixOS/nixpkgs/blob/master/lib/tests) or [pkgs/test](https://github.com/NixOS/nixpkgs/blob/master/pkgs/test)
  - made sure NixOS tests are [linked](https://nixos.org/manual/nixpkgs/unstable/#ssec-nixos-tests-linking) to the relevant packages
- [x] Tested compilation of all packages that depend on this change using `nix-shell -p nixpkgs-review --run "nixpkgs-review rev HEAD"`. Note: all changes have to be committed, also see [nixpkgs-review usage](https://github.com/Mic92/nixpkgs-review#usage)
- [ ] Tested basic functionality of all binary files (usually in `./result/bin/`)
- [22.11 Release Notes (or backporting 22.05 Release notes)](https://github.com/NixOS/nixpkgs/blob/master/CONTRIBUTING.md#generating-2211-release-notes)
  - [ ] (Package updates) Added a release notes entry if the change is major or breaking
[...]

PR #309585, opened 2024-05-06
nixos/steam: add protontricks submodule

diniamo:
## Description of changes

The title is self-explanatory, as for the reason: other programs also rely on this variable to find proton versions, for example protontricks.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [x] Tested, as applicable:

---

### DIFF
--- a/nixos/modules/programs/steam.nix
+++ b/nixos/modules/programs/steam.nix
@@ -4,6 +4,8 @@ let
   cfg = config.programs.steam;
   gamescopeCfg = config.programs.gamescope;
 
+  extraCompatPaths = lib.makeSearchPathOutput "steamcompattool" "" cfg.extraCompatPackages;
+
   steam-gamescope = let
     exports = builtins.attrValues (builtins.mapAttrs (n: v: "export ${n}=${v}") cfg.gamescopeSession.env);
   in
@@ -42,7 +44,7 @@ in {
       '';
       apply = steam: steam.override (prev: {
         extraEnv = (lib.optionalAttrs (cfg.extraCompatPackages != [ ]) {
-          STEAM_EXTRA_COMPAT_TOOLS_PATHS = lib.makeSearchPathOutput "steamcompattool" "" cfg.extraCompatPackages;
+          STEAM_EXTRA_COMPAT_TOOLS_PATHS = extraCompatPaths;
         }) // (lib.optionalAttrs cfg.extest.enable {
           LD_PRELOAD = "${pkgs.pkgsi686Linux.extest}/lib/libextest.so";
         }) // (prev.extraEnv or {});
@@ -139,6 +141,11 @@ in {
       Load the extest library into Steam, to translate X11 input events to
       uinput events (e.g. for using Steam Input on Wayland)
     '';
+
+    protontricks = {
+      enable = lib.mkEnableOption "protontricks, a simple wrapper for running Winetricks commands for Proton-enabled games";
+      package = lib.mkPackageOption pkgs "protontricks" { };
+    };
   };
 
   config = lib.mkIf cfg.enable {
@@ -169,7 +176,8 @@ in {
     environment.systemPackages = [
       cfg.package
       cfg.package.run
-    ] ++ lib.optional cfg.gamescopeSession.enable steam-gamescope;
+    ] ++ lib.optional cfg.gamescopeSession.enable steam-gamescope
+    ++ lib.optional cfg.protontricks.enable (cfg.protontricks.package.override { inherit extraCompatPaths; });
 
     networking.firewall = lib.mkMerge [
       (lib.mkIf (cfg.remotePlay.openFirewall || cfg.localNetworkGameTransfers.openFirewall) {
--- a/pkgs/tools/package-management/protontricks/default.nix
+++ b/pkgs/tools/package-management/protontricks/default.nix
@@ -12,6 +12,7 @@
 , yad
 , pytestCheckHook
 , nix-update-script
+, extraCompatPaths ? ""
 }:
 
 buildPythonApplication rec {
@@ -51,7 +52,7 @@ buildPythonApplication rec {
     ]}"
     # Steam Runtime does not work outside of steam-run, so don't use it
     "--set STEAM_RUNTIME 0"
-  ];
+  ] ++ lib.optional (extraCompatPaths != "") "--set STEAM_EXTRA_COMPAT_TOOLS_PATHS ${extraCompatPaths}";
 
   nativeCheckInputs = [ pytestCheckHook ];
 


Shawn8901:
FYI in the PR adding that it was explicitly asked to not expose that globally: see https://github.com/NixOS/nixpkgs/pull/189398#discussion_r1437993188

Atemu:
I don't think there's all that many tools which require steam-specific things. I'd be in favour of adding options for the one or two tools which actually need this. I honestly can't think of much else than `protontricks`.

As an escape hatch, you can also always run the tool inside steam-run.

diniamo:
Tested, how is this?

nixos/modules/programs/steam.nix
```diff
@@ -141,6 +143,28 @@ in {
       Load the extest library into Steam, to translate X11 input events to
       uinput events (e.g. for using Steam Input on Wayland)
     '';
+
+    protontricks = mkOption {
+      description = "Protontricks Steam integration";
+      default = {};
+      type = types.submodule {
+        options = {
```
Atemu, line 151: This does not need to be a submodule.

```suggestion
    protontricks = {
```

nixos/modules/programs/steam.nix
```diff
@@ -141,6 +143,28 @@ in {
...
+          enable = mkEnableOption "Protontricks";
+          package = mkOption {
```
Atemu, line 153:
```suggestion
          package = mkPackageOption pkgs "protontricks";
```

nixos/modules/programs/steam.nix
```diff
@@ -141,6 +143,28 @@ in {
...
+            type = types.package;
+            default = pkgs.protontricks;
+            defaultText = literalExpression "pkgs.protontricks";
+            apply = protontricks: protontricks.overrideAttrs (prev: {
+              makeWrapperArgs = prev.makeWrapperArgs ++ (lib.optional (cfg.extraCompatPackages != [ ])
+                "--set STEAM_EXTRA_COMPAT_TOOLS_PATHS ${extraCompatPath}");
+            });
```
Atemu, line 160: Move this to the implementation. `apply` is for when you to do some change so that it's visible to Nix code reading the same config option.

Atemu (changes requested):

nixos/modules/programs/steam.nix
diniamo: Why? What's the point of having gamescopeSession as a submodule then?

nixos/modules/programs/steam.nix
diniamo: I'm not sure I understand, could you maybe give me an example?

nixos/modules/programs/steam.nix
Atemu: I don't see a reason for it to be a submodule either.

cc @nrdxp

nixos/modules/programs/steam.nix
diniamo: Then again, what's the reason it's used for the steam package?

Atemu: For it I could understand it slightly more as I could imagine usage by users outside of the steam module but it too should probably be moved to the implementation.

Note that this is just a minor issue.

diniamo: I see, so it's only if you want to expose it for external usage.

That's all my questions then, I'll try to implement the suggestions tomorrow.

Atemu:
I'll take a closer look later but I don't see any obvious formatting issues. Unless you're doing something utterly weird that doesn't fit into the rest of Nixpkgs, there aren't any strict rules yet until https://github.com/NixOS/rfcs/pull/166 is implemented.

Shawn8901 (approved):
lgtm

diniamo:
What is that ofborg error and where is the request change that's blocking the merge?

Atemu:
It's an internal error that someone who maintains ofBorg will have to take a look at and manually remove at some point. It doesn't block the merge though, that's just my review. It'll stop blocking it as soon as I approve after I've taken a proper look or another committer comes around and dismisses it.

diniamo:
Resolved conflicts, tested.

diniamo:
@Atemu have you had a chance to look? It would be great if we could get this in before the next unstable merge.

nixos/modules/programs/steam.nix
```diff
@@ -20,6 +22,11 @@ let
       Exec=${steam-gamescope}/bin/steam-gamescope
       Type=Application
     '').overrideAttrs (_: { passthru.providedSessions = [ "steam" ]; });
+
+  protontricks = cfg.protontricks.package.overrideAttrs (prev: {
+    makeWrapperArgs = prev.makeWrapperArgs ++
+      (lib.optional (cfg.extraCompatPackages != [ ]) "--set STEAM_EXTRA_COMPAT_TOOLS_PATHS ${extraCompatPaths}");
+  });
```
Atemu, line 29: It'd be better if protontricks provided a proper override option for this. `overrideAttrs` isn't really a stable interface.

nixos/modules/programs/steam.nix
```diff
@@ -139,6 +146,11 @@ in {
       Load the extest library into Steam, to translate X11 input events to
       uinput events (e.g. for using Steam Input on Wayland)
     '';
+
+    protontricks = {
+      enable = lib.mkEnableOption "Protontricks";
```
Atemu, line 151: Please consider users who may not know what protontricks is. The option to enable it should offer at least a little information on it.

We could also pull it from the package's meta like this: 

```suggestion
      enable = lib.mkEnableOption "protontricks, ${cfg.protontricks.package.meta.description}";
```

Atemu (commented):
Two minor things, otherwise LGTM.

nixos/modules/programs/steam.nix
diniamo: I'm not sure what you mean by that / how it should be implemented. Do you mean the derivation should contain an argument for extraSteamCompatPaths, and we should use override here instead?

Atemu: Yup that's precisely it. You'd basically define an API that's quite a bit more stable than overrideAttrs.

diniamo:
Looks like your suggestion fails the documentation CI, apparently option descriptions can't rely on config.

diniamo:
Welp, adding it manually is the best I can come up with.

Atemu:
> Looks like your suggestion fails the documentation CI (and not just the CI, my system config errors too), apparently option descriptions can't rely on config/pkgs.

Ah, indeed it makes sense; we don't have access to `config` when defining `options`. You could probably dig it out of `options..package.default` but hard-coding like this is fine.

Atemu (commented):
Diff LGTM and the NixOS part appears to be working but I'm not sure protontricks is as it's giving me:

```
protontricks (ERROR): Could not find configured Proton installation!
protontricks (ERROR): Active Proton installation could not be found automatically.
Proton installation could not be found!
```

Though I'm also getting the same error when I `steam-run` it which worked before?

Is it working on your end @diniamo?

Atemu (approved):

diniamo:
Nice, thanks

Issue #96006, opened 2020-08-22
mkRenamedOptionModule doesn't work in submodules


jD91mZM2:
**Describe the bug**

I'm working on a home-manager module, where I now want to rename

```
programs.firefox.profiles.<name>.privacy.enableSettings
```

to

```
programs.firefox.profiles.<name>.privacy.settings.enable
```

The problem is, that list of submodules.

I tried to add

```nix
imports = [
  (mkRenamedOptionModule
    [ "privacy" "enableSettings" ]
    [ "privacy" "settings" "enable" ])
];
```

to the `programs.firefox.profiles.<name>` module, but that causes the error

```
warning: Git tree '/home/user/dotfiles' is dirty
warning: Git tree '/home/user/dotfiles' is dirty
error: --- ThrownError ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ nix-build
The option `programs.firefox.profiles.main.warnings' defined in `/home/user/dotfiles/nur-packages/hm-modules/programs/firefox-privacy.nix' does not exist.
(use '--show-trace' to show detailed location information)

There are 135 unread and relevant news items.
Read them by running the command 'home-manager news'.
```

because each submodule doesn't have a separate warning system.

Obviously doing

```nix
imports = [
  (mkRenamedOptionModule
    [ "programs" "firefox" "profiles" "privacy" "enableSettings" ]
    [ "programs" "firefox" "profiles" "privacy" "settings" "enable" ])
];
```

doesn't work either, because it's missing the profile name.

**To Reproduce**

Try to rename a suboption in an option of type `types.attrsOf (types.submodule ({ config, ... }: {`.

The error I got can be seen [on my nur-packages' GitLab](https://gitlab.com/jD91mZM2/nur-packages/-/blob/nixpkgs-issue-96006/hm-modules/programs/firefox-privacy.nix#L92-96) (on tag `nixpkgs-issue-96006` which I created for this occasion).

**Expected behavior**

Unclear. I guess `warnings` would be propagated upwards, somehow, using magic?

**Metadata**
Please run `nix-shell -p nix-info --run "nix-info -m"` and paste the result.

 - system: `"x86_64-linux"`
 - host os: `Linux 5.7.16, NixOS, 20.09.20200821.4a87da1 (Nightingale)`
 - multi-user?: `yes`
 - sandbox: `yes`
 - version: `nix-env (Nix) 2.4pre20200721_ff314f1`
 - channels(user): `""`
[...]

PR #309643, opened 2024-05-06
nixos/garage: drop replication_mode setting

flokli:
This got broken up into separate `replication_factor` and `consistency_mode` settings with Garage 1.x, and due to the the "none" default kicking in, Garage fails to startup with

> : Error: Either the legacy replication_mode or replication_level and consistency_mode can be set, not both.

if we actually make the migration as documented in the migration guide.

Drop this explicit setting, so users can set replication_mode or replication_factor/consistency_mode, depending on the version they're using.

With this PR, redeploying with the old key is a no-op (as expected), and I'm able to switch to `replication_factor` / `consistency_mode`.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [x] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [x] Tested, as applicable:

---

### DIFF
--- a/nixos/modules/services/web-servers/garage.nix
+++ b/nixos/modules/services/web-servers/garage.nix
@@ -52,13 +52,6 @@ in
             type = types.path;
             description = "The main data storage, put this on your large storage (e.g. high capacity HDD)";
           };
-
-          replication_mode = mkOption {
-            default = "none";
-            type = types.enum ([ "none" "1" "2" "3" "2-dangerous" "3-dangerous" "3-degraded" 1 2 3 ]);
-            apply = v: toString v;
-            description = "Garage replication mode, defaults to none, see: <https://garagehq.deuxfleurs.fr/documentation/reference-manual/configuration/#replication-mode> for reference.";
-          };
         };
       };
       description = "Garage configuration, see <https://garagehq.deuxfleurs.fr/documentation/reference-manual/configuration/> for reference.";
--- a/nixos/modules/services/web-servers/garage.nix
+++ b/nixos/modules/services/web-servers/garage.nix
@@ -64,6 +64,24 @@ in
   };
 
   config = mkIf cfg.enable {
+
+    assertions = [
+      # We removed our module-level default for replication_mode. If a user upgraded
+      # to garage 1.0.0 while relying on the module-level default, they would be left
+      # with a config which evaluates and builds, but then garage refuses to start
+      # because either replication_factor or replication_mode is required.
+      # This assertion can be removed in NixOS 24.11, when all users have been warned once.
+      {
+        assertion = (cfg.settings ? replication_factor || cfg.settings ? replication_mode) || lib.versionOlder cfg.package "1.0.0";
+        message = ''
+          Garage 1.0.0 requires an explicit replication factor to be set.
+          Please set replication_factor to 1 explicitly to preserve the previous behavior.
+          https://git.deuxfleurs.fr/Deuxfleurs/garage/src/tag/v1.0.0/doc/book/reference-manual/configuration.md#replication_factor
+
+        '';
+      }
+    ];
+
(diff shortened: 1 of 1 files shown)


flokli:
@yu-re-ka thanks for the assertion :heart:

nixos/modules/services/web-servers/garage.nix
```diff
@@ -71,6 +64,19 @@ in
   };
```
pluiedev, line 65: Try using `lib.mkRemovedOptionModule` instead, though I'm not sure if that works with submodules (`mkRenamedOptionModule` doesn't, see #96006)

```nix
{
  imports = [
    (lib.removedOptionModule ["services" "garage" "settings" "replication_mode"] ''
      Garage requires an explicit replication factor to be set since version 1.0.0.
        
      Please set `services.garage.settings.replication_factor` to 1 explicitly to preserve the previous behavior.
      See https://git.deuxfleurs.fr/Deuxfleurs/garage/src/tag/v1.0.0/doc/book/reference-manual/configuration.md#replication_factor for more info.
    '')
  ];
    
  config = mkIf cfg.enable {
    # ...
  }
}
```

pluiedev (changes requested):

nixos/modules/services/web-servers/garage.nix
```diff
@@ -71,6 +64,19 @@ in
...
+
+    assertions = [
+      # We removed our module-level default for replication_mode. If a user upgraded
+      # to garage 1.0.0 while relying on the module-level default, they would be left
+      # with a config which evaluates and builds, but then garage refuses to start
+      # because either replication_factor or replication_mode is required.
+      # This assertion can be removed in NixOS 24.11, when all users have been warned once.
+      {
+        assertion = (cfg.settings ? replication_factor || cfg.settings ? replication_mode) || lib.versionOlder cfg.package "1.0.0";
+        message = "Garage 1.0.0 requires an explicit replication factor to be set.\nPlease set replication_factor to 1 explicitly to preserve the previous behavior.\nhttps://git.deuxfleurs.fr/Deuxfleurs/garage/src/tag/v1.0.0/doc/book/reference-manual/configuration.md#replication_factor";
```
pluiedev, line 76: Use multiline strings
```suggestion
        message = ''
          Garage requires an explicit replication factor to be set since version 1.0.0.
        
          Please set `services.garage.settings.replication_factor` to 1 explicitly to preserve the previous behavior.
          See https://git.deuxfleurs.fr/Deuxfleurs/garage/src/tag/v1.0.0/doc/book/reference-manual/configuration.md#replication_factor for more info.
        '';
```

pluiedev (changes requested):

nixos/modules/services/web-servers/garage.nix
```diff
@@ -71,6 +64,24 @@ in
...
   config = mkIf cfg.enable {
+
+    assertions = [
+      # We removed our module-level default for replication_mode. If a user upgraded
+      # to garage 1.0.0 while relying on the module-level default, they would be left
+      # with a config which evaluates and builds, but then garage refuses to start
+      # because either replication_factor or replication_mode is required.
+      # This assertion can be removed in NixOS 24.11, when all users have been warned once.
+      {
+        assertion = (cfg.settings ? replication_factor || cfg.settings ? replication_mode) || lib.versionOlder cfg.package "1.0.0";
```
teutat3s, line 75: This should be
```suggestion
        assertion = (cfg.settings ? replication_factor || cfg.settings ? replication_mode) || lib.versionOlder cfg.package.version "1.0.0";
```

teutat3s: Created #311776

PR #309650, opened 2024-05-06
haskellPackages.ghcjs-dom: build on js backend of ghc 9.8

alexfmpe:
## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [x] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---


alexfmpe:
I assume on 9.6 the FFI support wasn't there yet as the build fails with 
```
Preprocessing library for ghcjs-base-0.8.0.0..
Building library for ghcjs-base-0.8.0.0..
[ 1 of 73] Compiling Data.JSString.Internal ( Data/JSString/Internal.hs, dist/build/Data/JSString/Internal.o )
[ 2 of 73] Compiling Data.JSString.Internal.Fusion.Types ( Data/JSString/Internal/Fusion/Types.hs, dist/build/Data/JSString/Internal/Fusion/Types.o )
[ 3 of 73] Compiling Data.JSString.Internal.Fusion.CaseMapping ( Data/JSString/Internal/Fusion/CaseMapping.hs, dist/build/Data/JSString/Internal/Fusion/CaseMapping.o )
[ 4 of 73] Compiling Data.JSString.Internal.Fusion.Common ( Data/JSString/Internal/Fusion/Common.hs, dist/build/Data/JSString/Internal/Fusion/Common.o )
[ 5 of 73] Compiling GHCJS.Concurrent ( GHCJS/Concurrent.hs, dist/build/GHCJS/Concurrent.o )

GHCJS/Concurrent.hs:37:1: error:
    Could not find module ‘GHC.JS.Foreign.Callback’
    Use -v (or `:set -v` in ghci) to see a list of the files searched for.
   |
37 | import           GHC.JS.Foreign.Callback (OnBlocked(..))
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[ 6 of 73] Compiling GHCJS.Internal.Types ( GHCJS/Internal/Types.hs, dist/build/GHCJS/Internal/Types.o )

GHCJS/Internal/Types.hs:17:1: error:
    Could not find module ‘GHC.JS.Foreign.Callback’
    Use -v (or `:set -v` in ghci) to see a list of the files searched for.
   |
17 | import GHC.JS.Foreign.Callback (Callback)
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[25 of 73] Compiling JavaScript.Cast  ( JavaScript/Cast.hs, dist/build/JavaScript/Cast.o )
[32 of 73] Compiling JavaScript.Number ( JavaScript/Number.hs, dist/build/JavaScript/Number.o )
```

pkgs/development/haskell-modules/configuration-ghcjs-9.x.nix
```diff
@@ -0,0 +1,47 @@
...
+
+pkgs.lib.composeManyExtensions [
+
+  (self: super: {
+    ghcjs-base = super.ghcjs-base_0_8_0_0;
+  })
+
+  (self: super: {
+    ghcjs-dom-javascript = unmarkBroken super.ghcjs-dom-javascript;
+    ghcjs-dom-jsffi = unmarkBroken super.ghcjs-dom-jsffi;
```
alexfmpe, line 13: I think I shouldn't need these, but `cabal2nix` gets [confused](https://github.com/NixOS/cabal2nix/blob/f8e6bf749a158a5ed866c57deee907b5f16c38e5/distribution-nixpkgs/CHANGELOG.md#unreleased) when I try to set the supported platform. When given "javascript-ghcjs" it outputs "js-ghcjs" which then leads to
```
 error: Package ‘ghcjs-dom-javascript-0.9.9.0’ in /home/alexfmpe/profiles/alexfmpe/repos/nixpkgs/pkgs/development/haskell-modules/hackage-packages.nix:120114 is not available on the requested hostPlatform:
         hostPlatform.config = "javascript-unknown-ghcjs"
         package.meta.platforms = [
           "js-ghcjs"
         ]
         package.meta.badPlatforms = [ ]
       , refusing to evaluate.
```

This is what I tried, including generated `hackage-packages.nix` : 

```diff
diff --git a/pkgs/development/haskell-modules/configuration-hackage2nix/broken.yaml b/pkgs/development/haskell-modules/configuration-hackage2nix/broken.yaml
index 66895c194519..e44096f7a52f 100644
--- a/pkgs/development/haskell-modules/configuration-hackage2nix/broken.yaml
+++ b/pkgs/development/haskell-modules/configuration-hackage2nix/broken.yaml
@@ -1989,8 +1989,6 @@ broken-packages:
   - ghci-ng # failure in job https://hydra.nixos.org/build/233229533 at 2023-09-02
   - ghcitui # failure in job https://hydra.nixos.org/build/252737339 at 2024-03-16
   - ghcjs-base-stub # timeout
-  - ghcjs-dom-javascript # failure in job https://hydra.nixos.org/build/255688382 at 2024-04-16
-  - ghcjs-dom-jsffi # failure in job https://hydra.nixos.org/build/233215225 at 2023-09-02
   - ghcjs-fetch # timeout
   - ghcjs-promise # failure in job https://hydra.nixos.org/build/233243985 at 2023-09-02
   - ghcjs-xhr # failure in job https://hydra.nixos.org/build/233235693 at 2023-09-02
diff --git a/pkgs/development/haskell-modules/configuration-hackage2nix/main.yaml b/pkgs/development/haskell-modules/configuration-hackage2nix/main.yaml
index 0f0b3520b2e0..125a0774d3fa 100644
--- a/pkgs/development/haskell-modules/configuration-hackage2nix/main.yaml
+++ b/pkgs/development/haskell-modules/configuration-hackage2nix/main.yaml
@@ -721,6 +721,8 @@ supported-platforms:
   geomancy:                                     [ platforms.x86 ] # x86 intrinsics
   geomancy-layout:                              [ platforms.x86 ] # x86 intrinsics
   gi-gtkosxapplication:                         [ platforms.darwin ]
+  ghcjs-dom-javascript:                         [ javascript-ghcjs ]
+  ghcjs-dom-jsffi:                              [ javascript-ghcjs ]
   gtk-mac-integration:                          [ platforms.darwin ]
   gtk3-mac-integration:                         [ platforms.darwin ]
   halide-haskell:                               [ platforms.linux ]
diff --git a/pkgs/development/haskell-modules/hackage-packages.nix b/pkgs/development/haskell-modules/hackage-packages.nix
index 4ceb1c7be6ae..3952740cfa54 100644
--- a/pkgs/development/haskell-modules/hackage-packages.nix
+++ b/pkgs/development/haskell-modules/hackage-packages.nix
@@ -120116,8 +120116,7 @@ self: {
        sha256 = "0vhk1gjr5g5cp6rzkr9p0km4mrdi61kfwk8jc0w97wkwmi030qcb";
        description = "DOM library using JSFFI and GHCJS";
        license = lib.licenses.mit;
-       hydraPlatforms = lib.platforms.none;
-       broken = true;
+       platforms = [ "js-ghcjs" ];
      }) {};
 
   "ghcjs-dom-jsaddle" = callPackage
@@ -120140,8 +120139,7 @@ self: {
        sha256 = "0fhcs89x180kw75qgbfh28wjyn55frwdfj4rqlqa1smx0fwhzyy1";
        description = "DOM library using JSFFI and GHCJS";
        license = lib.licenses.mit;
-       hydraPlatforms = lib.platforms.none;
-       broken = true;
+       platforms = [ "js-ghcjs" ];
      }) {};
 
   "ghcjs-dom-webkit" = callPackage
```

pkgs/development/haskell-modules/configuration-ghcjs-9.x.nix
```diff
@@ -0,0 +1,47 @@
...
+  (self: super: {
+    QuickCheck = disableLibraryProfiling super.QuickCheck;
+
+    aeson = disableLibraryProfiling super.aeson;
+    ghcjs-base = disableLibraryProfiling super.ghcjs-base;
+    ghcjs-dom = disableLibraryProfiling super.ghcjs-dom;
+    ghcjs-dom-javascript = disableLibraryProfiling super.ghcjs-dom-javascript;
+    jsaddle = disableLibraryProfiling super.jsaddle;
+    jsaddle-dom = disableLibraryProfiling super.jsaddle-dom;
+  })
```
alexfmpe, line 45: Is there a way to apply `disableLibraryProfiling` to everything that depends on QuickCheck? 
Do we even want that?

pkgs/development/haskell-modules/default.nix
```diff
@@ -36,6 +37,8 @@ let
     compilerConfig
     packageSetConfig
     overrides
+  ] ++ lib.optionals stdenv.targetPlatform.isGhcjs [
+    (configurationJS { inherit pkgs haskellLib; })
```
alexfmpe, line 41: I tried to add this before, alongside the other `platformConfigurations`, but the overrides would get lost as `configuration-ghc-9.8.nix`  (passed in `compilerConfig`) would then reset to `null` or overwrite due to picking the most-recent hackage version.

pkgs/development/haskell-modules/configuration-ghcjs-9.x.nix
sternenseemann: This “just” needs a fix in distribution-nixpkgs. The relevant code there was written before GHC upstream/Cabal forced us to adopt the javascript arch prefix and I forgot to change it there, too.

pkgs/development/haskell-modules/configuration-ghcjs-9.x.nix
sternenseemann: Globally disabling library profiling for the js backend would also be a possibility…

sternenseemann: Alternatively we could try applying https://gitlab.haskell.org/ghc/ghc/-/merge_requests/12463…

sternenseemann:
8c61a345 should simplify this PR a lot! I'm wondering if we even need a special file for the js backend at all. Most of these packages wouldn't build with normal GHCs anyways, so we can apply the overrides in the normal configuration-nix.nix? Or do you think there's an advantage in having them conditional? The disadvantage of this is that eval regressions won't be caught since we can't test the entire set on Hydra…

pkgs/development/haskell-modules/configuration-ghcjs-9.x.nix
sternenseemann: Seems reasonable. Maybe backporting the patch is too much investment already. 9.10 is around the corner and we probably shouldn't go above and beyond for the dev preview backend necessarily?

alexfmpe:
> [8c61a34](https://github.com/NixOS/nixpkgs/commit/8c61a345004cd0eb42a1cf57698bae895b806450) should simplify this PR a lot! 

Huh, how? DId you mean the `addBuildDepends self.ghcjs-base` shouldn't be needed anymore? 
Looking at that diff it doesn't seem `hackage-packages.nix` changed anything in the dependent packages generation.

With this change, the `ghcjs-base = super.ghcjs-base_0_8_0_0;` override fails but if I remove it I get

```
$ nix-build -A pkgsCross.ghcjs.haskell.packages.ghc98.ghcjs-base                                                                                                                        
error:
(...)
           12|   (self: super: {
           13|     ghcjs-base = addBuildDepends (with self; [
             |     ^
           14|       aeson

       error: value is null while a set was expected
```

I think the change also requires removing

https://github.com/NixOS/nixpkgs/blob/8c61a345004cd0eb42a1cf57698bae895b806450/pkgs/development/haskell-modules/configuration-common.nix#L270

so that it builds on `pkgsCross.ghcjs`. 

That in turn makes `nix-build -A haskell.packages.ghc98.ghcjs-base` switch from 
```
error: expression does not evaluate to a derivation (or a set or list of those)
```  
to 
```
(...)
Running phase: buildPhase
Building ghcjs-base-0.8.0.0...
Running phase: checkPhase
Package has no test suites.
Running phase: haddockPhase
Warning: No documentation was generated as this package does not contain a
library. Perhaps you want to use the --executables, --tests, --benchmarks or
--foreign-libraries flags.
Running phase: installPhase
Error: Setup: No executables and no library found. Nothing to do.

error: builder for '/nix/store/nw2fyq4r6vv34nv8jlf9g3dvrkh5i0yv-ghcjs-base-0.8.0.0.drv' failed with exit code 1;
       last 10 log lines:
```
so we probably should set its supported platforms, but that needs the call2nix release.

pkgs/development/haskell-modules/configuration-ghcjs-9.x.nix
alexfmpe: How does one disable profiling globally? 
Or did you mean piping that entire package set through `builtins.mapAttrs (_: disableLibraryProfiling)` ?

pkgs/development/haskell-modules/configuration-ghcjs-9.x.nix
alexfmpe: Fixed by the newly merged release of distribution-nixpkgs

alexfmpe:
Anything left to do here?

pkgs/development/haskell-modules/default.nix
maralorn: That’s unfortunate.

maralorn: I don’t get it. there do not seem to be conflicting overrides for the packages in ghcjs-9.x.nix after the changes you are making. Are you sure this doesn’t work?

maralorn:
Thank you very much!

Any explanation regarding the white space changes in hackage-packages.nix?

I left one other comment. Apart from that this looks good to me.

Ah, do we already have enough hydra jobs for this?

pkgs/development/haskell-modules/default.nix
```diff
@@ -36,6 +37,8 @@ let
     compilerConfig
     packageSetConfig
     overrides
+  ] ++ lib.optionals stdenv.hostPlatform.isGhcjs [
+    (configurationJS { inherit pkgs haskellLib; })
```
sternenseemann, line 41: I'd prefer if we could just use `platformConfigurations` for this instead.

pkgs/development/haskell-modules/default.nix
alexfmpe: > Are you sure this doesn’t work?

Confirmed to be working now. I think the problem is that I originally had
```
    aeson = disableLibraryProfiling super.aeson;
```
which was being replaced by the flat `aeson` override in the 9.8 bit but now we disable profiling globally for JS flavors so this is no longer a problem.

pkgs/development/haskell-modules/default.nix
alexfmpe: Agreed. This now works, as per: https://github.com/NixOS/nixpkgs/pull/309650#discussion_r1640253924

alexfmpe:
> Any explanation regarding the white space changes in hackage-packages.nix?

Oops, fixed. 
My editor automatically trimmed the trailing whitespace after I manually edited the file to handle a conflict with #316272

sternenseemann (approved):

sternenseemann:
> Ah, do we already have enough hydra jobs for this?

Probably not, we should expand that, but that can probably be done in a follow up change.

pkgs/development/haskell-modules/configuration-ghcjs-9.x.nix
alexfmpe: @sternenseemann this has been fixed/backported for 9.8.3, 9.10.2, 9.12.1. 
I already confirmed it on ghc984 and ghc9121. 
Should we simply re-enable profiling across the board once we have ghc9102 ?

PR #309681, opened 2024-05-07
platformsh: 4.17.0 -> 5.0.13

trsteel88:
## Description of changes

Remove the deprecated [legacy cli](https://github.com/platformsh/legacy-cli) in favor of the [new version](https://github.com/platformsh/cli).

## Things done

This script is a copy of the [Upsun CLI package](https://github.com/NixOS/nixpkgs/commit/6805bd7bb0f15118dd44ad7ece298153ef2392bd) which is released under the same repository.

I have modified the binary to use **platform** instead of **upsun**

### DIFF
--- a/pkgs/by-name/pl/platformsh/package.nix
+++ b/pkgs/by-name/pl/platformsh/package.nix
@@ -52,10 +52,10 @@ stdenvNoCC.mkDerivation (finalAttrs: {
   };
 
   meta = {
-    description = "The unified tool for managing your Platform.sh services from the command line";
+    description = "The unified tool for managing your Platform.sh services from the command line.";
     homepage = "https://github.com/platformsh/cli";
     license = lib.licenses.mit;
-    mainProgram = "platformsh";
+    mainProgram = "platform";
     maintainers = with lib.maintainers; [ shyim spk ];
     platforms = [ "x86_64-linux" "aarch64-linux" "x86_64-darwin" "aarch64-darwin" ];
     sourceProvenance = with lib.sourceTypes; [ binaryNativeCode ];
--- a/pkgs/by-name/pl/platformsh/package.nix
+++ b/pkgs/by-name/pl/platformsh/package.nix
@@ -14,19 +14,19 @@ stdenvNoCC.mkDerivation (finalAttrs: {
     {
       x86_64-darwin = fetchurl {
         url = "https://github.com/platformsh/cli/releases/download/${finalAttrs.version}/platformsh-cli_${finalAttrs.version}_darwin_all.tar.gz";
-        hash = "sha256-5JKXtAUnqrlufyNE05uZjEDfJv557auYPriTxvUbMJI=";
+        hash = "sha256-ABnfxLMtY8E5KqJkrtIlPB4ML7CSFvjizCabv7i7SbU=";
       };
       aarch64-darwin = fetchurl {
         url = "https://github.com/platformsh/cli/releases/download/${finalAttrs.version}/platformsh-cli_${finalAttrs.version}_darwin_all.tar.gz";
-        hash = "sha256-5JKXtAUnqrlufyNE05uZjEDfJv557auYPriTxvUbMJI=";
+        hash = "sha256-ABnfxLMtY8E5KqJkrtIlPB4ML7CSFvjizCabv7i7SbU=";
       };
       x86_64-linux = fetchurl {
         url = "https://github.com/platformsh/cli/releases/download/${finalAttrs.version}/platformsh-cli_${finalAttrs.version}_linux_amd64.tar.gz";
-        hash = "sha256-fjVL/sbO1wmaJ4qZpUMV/4Q4Jzf0p6qx0ElRdY5EUJU=";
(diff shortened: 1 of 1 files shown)


Aleksanaa (commented):
Diff lgtm. Please reword to `platformsh: 4.17.0 -> 5.0.12`.

Also is it not possible to be built from source now? I prefer to wait longer to see if there is a solution.

spk:
> Also is it not possible to be built from source now? I prefer to wait longer to see if there is a solution.

its still complicated build at the moment using docker buildx. If possible I think we should try to have platformsh cli updated because we are migrating new features there

spk (approved):

Aleksanaa:
Also why is @shyim removed from maintainers?

pkgs/by-name/pl/platformsh/package.nix
```diff
@@ -1,42 +1,34 @@
...
-
-    if [[ "v${finalAttrs.version}" = "$NEW_VERSION" ]]; then
-      echo "The new version same as the old version."
-      exit 0
-    fi
-
-    update-source-version "platformsh" "$NEW_VERSION"
+  version = "5.0.12";
+
+  src = {
```
shyim, line 7: can you also add the macOS releases?

trsteel88: I'm relatively new to nix and I'm unsure what you need here. Do you have an example or docs I could follow. This was originally a copy of the upsun package which is released under the same repository (different cli though)

spk: I was new to nixpkgs too when I did the upsun package, I think something like this https://github.com/NixOS/nixpkgs/pull/311643 should do it

spk: File has been updated to latest version with darwin too in https://github.com/nixos/nixpkgs/blob/51f5b20cdba6d80024fa5c76147146442eb8137c/pkgs/by-name/up/upsun/package.nix

trsteel88:
> Also why is @shyim removed from maintainers?

Sorry, fixed this.

pkgs/by-name/pl/platformsh/package.nix
trsteel88: Thanks, I've replicated this.

pkgs/by-name/pl/platformsh/package.nix
```diff
@@ -1,34 +1,63 @@
...
 
+  passthru = {
+    tests.version = testers.testVersion {
+      inherit (finalAttrs) version;
+      package = platformsh;
+    };
+  };
+
   meta = {
+    description = "The unified tool for managing your Platform.sh services from the command line.";
```
Aleksanaa, line 55:
```suggestion
    description = "The unified tool for managing your Platform.sh services from the command line";
```

spk (approved):

spk:
maybe commits can be squashed to `platformsh: 4.17.0 -> 5.0.13`

Aleksanaa:
Okay

vcunat:
It doesn't build on Hydra now:
https://hydra.nixos.org/eval/1806362?filter=platformsh#tabs-now-fail

Aleksanaa:
@trsteel88 are you fixing it? Otherwise I'll fix it

trsteel88:
> @trsteel88 are you fixing it? Otherwise I'll fix it

Was just looking at it and it is because the binary is called "platform" rather than "platformsh". I'm not sure what needs to be adjusted to fix this.

Is it line [42](https://github.com/NixOS/nixpkgs/blob/619f889674f841be1ce4a18e9a35afdc095b912c/pkgs/by-name/pl/platformsh/package.nix#L42) and line [50](https://github.com/NixOS/nixpkgs/blob/619f889674f841be1ce4a18e9a35afdc095b912c/pkgs/by-name/pl/platformsh/package.nix#L50)?

shyim:
I am creating a PR to fix it

trsteel88:
@shyim thanks for sorting that. I am new to nix but wondering how updates get delivered. e.g. [5.0.14](https://github.com/platformsh/cli/releases) has been released. Does someone have to come into this repository and update the version and hashes on https://github.com/NixOS/nixpkgs/blob/master/pkgs/by-name/pl/platformsh/package.nix each time a new release comes out?

If that's correct it is very manual - is there a way to automate this?

vcunat:
There's [a bot](https://nix-community.org/update-bot/) that can detect updates for some kinds of packages and open pull requests: https://github.com/NixOS/nixpkgs/pulls/r-ryantm

shyim:
The better way would we if we provide a update script, so we can run it also locally 🤔

vcunat:
The bot uses those update scripts, too.

spk:
Look like the bot is using https://repology.org/project/platformsh/versions but since the package is not on any distributions (beside nixos) it fails to fetch the latest one?

vcunat:
Yes, repology is used, but if you provide an update script, it should be used instead AFAIK:
https://nix-community.github.io/nixpkgs-update/batch-updates/#batch-updates

PR #309805, opened 2024-05-07
util-linux: 2.39.3 -> 2.40.1

alyssais:
## Description of changes

CVE-2024-28085 (low priority since NixOS doesn't make wall setgid by default).

### Release notes

- [2.40](https://lore.kernel.org/util-linux/20240327151052.5kjyfcy5w7wnkqf3@ws.net.home/)
- [2.40.1](https://lore.kernel.org/util-linux/20240506085127.ywva7jnimovnnrlm@ws.net.home/)

Don't see anything looking like a breaking change.

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [x] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/os-specific/linux/util-linux/default.nix
+++ b/pkgs/os-specific/linux/util-linux/default.nix
@@ -2,6 +2,7 @@
 , capabilitiesSupport ? stdenv.isLinux
 , libcap_ng
 , libxcrypt
+, sqlite
 , ncursesSupport ? true
 , ncurses
 , pamSupport ? true
@@ -20,11 +21,11 @@
 
 stdenv.mkDerivation rec {
   pname = "util-linux" + lib.optionalString (!nlsSupport && !ncursesSupport && !systemdSupport) "-minimal";
-  version = "2.39.3";
+  version = "2.40.1";
 
   src = fetchurl {
     url = "mirror://kernel/linux/utils/util-linux/v${lib.versions.majorMinor version}/util-linux-${version}.tar.xz";
-    hash = "sha256-e2YF5I0aSfQ8xLTPxZ8xPQ3VQC+kC5aBC9Vy4Wff7Q8=";
+    hash = "sha256-WeZ2qlPMtEtsOfD/4BqPonSJHJG+8UdHUvrZJGHe8k8=";
   };
 
   patches = [
@@ -40,7 +41,7 @@ stdenv.mkDerivation rec {
   separateDebugInfo = true;
 
   postPatch = ''
-    patchShebangs tests/run.sh
+    patchShebangs tests/run.sh tools/all_syscalls
 
     substituteInPlace sys-utils/eject.c \
       --replace "/bin/umount" "$bin/bin/umount"
@@ -59,6 +60,7 @@ stdenv.mkDerivation rec {
     "--enable-fs-paths-default=/run/wrappers/bin:/run/current-system/sw/bin:/sbin"
     "--disable-makeinstall-setuid" "--disable-makeinstall-chown"
     "--disable-su" # provided by shadow
+    "--with-tmpfilesdir=${placeholder "out"}/lib/tmpfiles.d"
     (lib.enableFeature writeSupport "write")
     (lib.enableFeature nlsSupport "nls")
     (lib.withFeature ncursesSupport "ncursesw")
@@ -80,7 +82,7 @@ stdenv.mkDerivation rec {
   nativeBuildInputs = [ pkg-config installShellFiles ]
     ++ lib.optionals translateManpages [ po4a ];
 
-  buildInputs = [ zlib libxcrypt ]
+  buildInputs = [ zlib libxcrypt sqlite ]
     ++ lib.optionals pamSupport [ pam ]
     ++ lib.optionals capabilitiesSupport [ libcap_ng ]
     ++ lib.optionals ncursesSupport [ ncurses ]


alyssais:
Security backport for stable: https://github.com/NixOS/nixpkgs/pull/309808

06kellyjac (approved):
Is the `liblastlog2` and `pam_lastlog2` dependency on `sqlite` always necessary or only for with `pamSupport`?

LGTM

alyssais:
> Is the `liblastlog2` and `pam_lastlog2` dependency on `sqlite` always necessary or only for with `pamSupport`?

If I'm reading correctly, it's possible to build liblastlog2 without building pam_lastlog2.  We could add an option to disable it, but I don't know how to decide whether that's worth it.

vcunat:
For reference, the parallel build is a bit flaky now:
```
/nix/store/bgcaxhhxswzvmxjbbgvvaximm5hwghz1-binutils-2.41/bin/ld: cannot find -llastlog2: No such file or directory
collect2: error: ld returned 1 exit status
libtool:   error: error: relink 'pam_lastlog2.la' with the above command before installing it

**Comment by alyssais** [comment]
Time to switch to Meson?

**Comment by vcunat** [comment]
It feels late for 24.05.  But fortunately adding meson+ninja wouldn't create an eval-time cycle here.

**Comment by alyssais** [comment]
[Reported](https://lore.kernel.org/util-linux/87le4c1zm4.fsf@alyssa.is/), also.

**Comment by vcunat** [comment]
And (aarch64-)darwin build regressed on this update with a different error now: https://hydra.nixos.org/build/260146685
```
libuuid/src/uuid_time.c:89:87: error: aliases are not supported on darwin
```

**Comment by vcunat** [comment]
Oh, that darwin issue [would](https://hydra.nixos.org/build/260152845) even become a blocker for `nixpkgs-unstable` channel :-/

**Comment by alyssais** [comment]
There's also a bug report here: https://github.com/util-linux/util-linux/issues/3011#issuecomment-2119153020

**Comment by vcunat** [comment]
Man, x86_64-darwin sports yet another error:
```
libsmartcols/src/filter.c:194:14: error: call to undeclared function 'fmemopen'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
        fltr->src = fmemopen((void *) str, strlen(str), "r");
                    ^
```
(IIRC older macOS doesn't provide this function yet)

**Comment by vcunat** [comment]
Quite a bad luck with this release, and in the last moments before nixpkgs "freeze".

**Comment by alyssais** [comment]
I really wish OfBorg was capable of getting passed LLVM on staging for Darwin…

**Comment by vcunat** [comment]
If it was so easy, we perhaps wouldn't even need `staging*`.

**Comment by alyssais** [comment]
It gets most of the way through — we just need to up the timeout a bit, and then I think OfBorg might even be less overloaded, because it wouldn't be wasting so much of its time trying the same futile LLVM builds over and over.

**Comment by vcunat** [comment]
Maybe we should just roll back to util-linux 2.39.x for now and keep nixpkgs 24.05 that way?  (darwin now, linux on the next rebuild)

**Comment by alyssais** [comment]
Sounds fine.

**Comment by alyssais** [comment]
We should probably cherry-pick #309808 in that case.

**Comment by vcunat** [comment]
Perhaps 2.39.4 instead, as it sounds minimal and should contain also that fix
https://github.com/util-linux/util-linux/blob/v2.39.4/Documentation/releases/v2.39.4-ReleaseNotes

**Comment by alyssais** [comment]
Ah yes, that sounds good.  I missed that.

**Comment by vcunat** [comment]
For reference, 2.40.1 also broke its static musl build:
https://hydra.nixos.org/build/259722977

**Comment by alyssais** [comment]
That's likely easily fixed by just disabling the PAM plugin on static builds.  But is there any point fixing that, even on staging, or will we revert this there as well?

**Comment by vcunat** [comment]
Right, not now.  I've given up on fixing 2.40.x for nixpkgs 24.05.

**Comment by alyssais** [comment]
I mean what about staging?  i.e. post 24.04

**Comment by alyssais** [comment]
I guess it'd just cause merge conflicts, since the gradual reversion on staging-next will propagate to staging.

So for future reference when we come back to 2.40, here's the diff for static:

```diff
diff --git i/pkgs/os-specific/linux/util-linux/default.nix w/pkgs/os-specific/linux/util-linux/default.nix
index 642480b670c7..169f8293ae3e 100644
--- i/pkgs/os-specific/linux/util-linux/default.nix
+++ w/pkgs/os-specific/linux/util-linux/default.nix
@@ -6,6 +6,7 @@
 , ncursesSupport ? true
 , ncurses
 , pamSupport ? true
+, pamLastlogSupport ? !stdenv.hostPlatform.isStatic
 , pam
 , systemdSupport ? lib.meta.availableOn stdenv.hostPlatform systemd
 , systemd
@@ -61,6 +62,7 @@ stdenv.mkDerivation rec {
     "--disable-makeinstall-setuid" "--disable-makeinstall-chown"
     "--disable-su" # provided by shadow
     "--with-tmpfilesdir=${placeholder "out"}/lib/tmpfiles.d"
+    (lib.enableFeature pamLastlogSupport "pam-lastlog2")
     (lib.enableFeature writeSupport "write")
     (lib.enableFeature nlsSupport "nls")
     (lib.withFeature ncursesSupport "ncursesw")
```

leona-ya:
For context: This update also caused a regression with detecting the UUID of some special LUKS devices (maybe only if the LUKS device contains a LVM). I haven't reported this upstream yet, will do later.
Some more context is in https://discourse.nixos.org/t/nixos-24-05-upgrade-no-boot-device/46335, which is quite analogous to what I experienced.

YellowOnion:
bcachefs-tools on master needs https://github.com/util-linux/util-linux/pull/3001 to detect my disks correctly, not a huge issue, but worth noting for future potential issues if that makes it in to nixpkgs-unstable before this does.

alyssais:
We can backport that patch.

PR #309867, opened 2024-05-07
github: add k3s label

superherointj:
github: add k3s label

* Enables linking issues and PRs to K3s label.
* Automatically adds K3s label to PRs changing K3s files.

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -113,6 +113,14 @@
         - pkgs/applications/editors/jupyter-kernels/**/*
         - pkgs/applications/editors/jupyter/**/*
 
+"6.topic: k3s":
+  - any:
+    - changed-files:
+      - any-glob-to-any-file:
+        - nixos/modules/services/cluster/k3s/**/*
+        - nixos/tests/k3s/**/*
+        - pkgs/applications/networking/cluster/k3s/**/*
+
 "6.topic: kernel":
   - any:
     - changed-files:


teto:
that's too specific, if we set a label for every package in nixpkgs it doesn't scale. Would a `kubernetes` label be ok ?

superherointj:
> that's too specific, if we set a label for every package in nixpkgs it doesn't scale. Would a `kubernetes` label be ok ?

Kubernetes is not the same. The upkeep is different. K3s is heavy on upkeep. There are ~8 maintainers (in and out, some are not listed) and issues pop up often. This eases upkeep because we can link issues and PRs to a label then look up label directly. It's handy.

.github/labeler.yml
```diff
@@ -113,6 +113,13 @@
         - pkgs/applications/editors/jupyter-kernels/**/*
         - pkgs/applications/editors/jupyter/**/*
 
+"6.topic: k3s":
+  - any:
+    - changed-files:
+      - any-glob-to-any-file:
+        - pkgs/applications/networking/cluster/k3s/**/*
+        - nixos/tests/k3s/**/*
```
Aleksanaa, line 121:
```suggestion
        - nixos/tests/k3s/**/*
        - nixos/modules/services/cluster/k3s/**/*
```

Well idk what `/**/*` means, but remember to include this directory

superherointj: Thanks. Totally missed that.

superherointj: I believe `/**/` means to include sub-directories.

wegank (approved):

teto:
> Unfortunately K3s and (NixOS) Kubernetes are different distributions with different cycles, issues, etc. There isn't overlap. So a Kubernetes label for K3s maintainers would be useless.

the label is to filter issues and PRs right ? a "k3s" label would be aimed just at a few maintainers while a "kubernetes" one would be of interest to more maintainers and users (I dont think it would add too much noise ?). The nixpkgs project stretches the limits of label support in github, selecting a label in UI is slow for me for instance which is why I dont propose to add both "kubernetes" and "k3s". 

I dont have any particular authority over labels, just sharing my sentiment. I get how convenient a label is to maintainers,I myself added the "jupyter" label.

superherointj:
@teto Do you use K3s?

teto:
I dont. What do you think of using "kubernetes" tag ?  in practice for now it just tags k3s files as you wanted and if kubernetes-related maintainers want to add their soft they can ?

superherointj:
> I dont.

Makes sense. Thanks for answering.

> What do you think of using "kubernetes" tag ? 

Not a good idea.

K3s is different from Kubernetes. K3s is for edge, SBCs like Raspberry Pi, heavily patched, stripped and adapted for low resource usage. Kubernetes is used in (large) servers. Think of: RabbitMQ and Mosquito. Or Java and JavaScript. Or Application programming vs Embedded programming. Different worlds, the large and the small, implemented differently to fulfill different purposes.

Interface wise might look similar. But implementation wise it's different beasts. Different in source code, packaging, maintainers, patches, process, documentation, tests, modules, etc. It diverges in many ways. It's different communities, I'd say. As I said earlier, the overlap is not that much.

> in practice for now it just tags k3s files as you wanted and if kubernetes-related maintainers want to add their soft they can ?

Upkeep wise I don't have see a good reason to conflate Kubernetes with K3s.

I'm only seeking to add K3s tag, which is what I maintain.

I think, it is best to keep it separate. As they are separate things managed by different people for different purposes.

Aleksanaa (commented):
I think adding this tag is okay. It does have more PR than some other tags

Aleksanaa (approved):

ulrikstrid (approved):
This seems like it makes sense. I would agree with @superherointj that it's not the same thing as Kubernetes when you get into it. On the surface it's the same thing but not really.

teto:
ok I didn't know it was this distinct. I dont want to be in the way of maintainers wanting to do work so feel free to merge.

Issue #50300, opened 2018-11-13
Systemctl kexec always boots /nix/var/nix/profiles/system even if custom profile is selected in nixos-rebuild


arianvp:
## Issue description

When doing `nixos-rebuild boot -p my-profile`  and then `systemctl kexec` I would
expect it to kexec into the just created profile, but it kexecs into my previous nixos configuration
that is in `/var/nix/profiles/system`

### Steps to reproduce
```
sudo nixos-rebuild boot  # make an entry in the system profile
# change /etc/nixos/configuration.nix
sudo nixos-rebuild boot -p hello  # make an entry in /var/nix/profiles/system-profiles/
sudo systemctl kexec
```
And we're rebooted into the old config instead of the new one.

This seems to be because the `kexec` module hardcodes the fact that boots are inside `/var/nix/profiles/system`

## Technical details

Please run `nix-shell -p nix-info --run "nix-info -m"` and paste the
results.

lheckemann:
I'd think that adding a symlink like `/run/next-boot-system` would be a way to fix this.

flokli:
@lheckemann hmmh, that could work, but then it shouldn't be a template unit, and have a condition on that file to exist. Still, `/run/next-boot-system` introduces yet another magic NixOS-specific symlink, where it doesn't necessarily need to be.

The `kexec-load@.service` template unit from the archlinux wiki is just a fancy way of telling systemd to `kexec --load` a new kernel doing the real kexec, by doing this in a unit that's added to `kexec.target`.

We should be able to do the same by just adding a non-templated unit to a fixed location in `/run/systemd/system` (and adding it to `kexec.target`).

We could also try to invoke `kexec --load` (without the  `--exec` argument of course) on `kexec`-able platforms during `nixos-rebuild boot` to load the "next-boot" kernel, initrd and cmdline into memory.

flokli:
Also interesting: https://lists.freedesktop.org/archives/systemd-devel/2012-March/004764.html

flokli:
Turns out, there already is `nixos/modules/system/boot/kexec.nix`, which creates a `prepare-kexec.service with `WantedBy=kexec.target` that loads the kernel from `/nix/var/nix/profiles/system/kernel`.

I built a new kernel with a custom cmdline (by setting `boot.kernelParams = ["foo"]';`), and after a `nixos-rebuild boot`, `/nix/var/nix/profiles/system/kernel-params` contains "foo".

Are you sure this is still a problem?

[...]

PR #309911, opened 2024-05-07
add support for soft-reboot

arianvp:
## Description of changes

Usage:

```
sudo nixos-rebuild boot && sudo systemctl soft-reboot
```

`switch-to-configuration boot` will now prepare a symlink `/run/next-system`
and just before soft-reboot we will automatically call
`/run/next-system/activate` to activate the next system to be booted when
`systemctl soft-reboot` is called.

not only can soft-reboot can be used as an alternative to kexec and reboot, but
it can also be used as an alternative to `nixos-rebuild switch` in many cases.

Unlike kexec and reboot, soft-reboot can keep certain resources around across the reboot.
For example, socket units can be configured to stay around such that reboots can happen
without stopping accepting any connections. Furthermore file descriptors can be passed across
boots using FDSTORE to even keep existing connections intact.

Please see the man-page for more info: https://www.freedesktop.org/software/systemd/man/latest/systemd-soft-reboot.service.html#

from Systemd NEWS:

> A "soft reboot" is similar to a regular reboot, except that it affects
> userspace only: the service manager shuts down any running services and other
> units, then optionally switches into a new root file system (mounted to
> /run/nextroot/), and then passes control to a systemd instance in the new file
> system which then starts the system up again. The kernel is not rebooted and
> neither is the hardware, firmware or boot loader. This provides a fast,
> lightweight mechanism to quickly reset or update userspace, without the latency
> that a full system reset involves. Moreover, open file descriptors may be
> passed across the soft reboot into the new system where they will be passed
> back to the originating services. This allows pinning resources across the
> reboot, thus minimizing grey-out time further

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/nixos/modules/system/activation/switch-to-configuration.pl
+++ b/nixos/modules/system/activation/switch-to-configuration.pl
@@ -115,6 +115,8 @@ if (($ENV{"NIXOS_NO_SYNC"} // "") ne "1") {
 }
 
 if ($action eq "boot") {
+    # /run/next-system/activate is called when systemctl soft-reboot is called
+    system("@coreutils@/bin/ln", "-sfn", "$toplevel", "/run/next-system") == 0 or exit 1;
     exit(0);
 }
 
--- a/nixos/modules/system/boot/systemd.nix
+++ b/nixos/modules/system/boot/systemd.nix
@@ -128,6 +128,8 @@ let
       "final.target"
       "kexec.target"
       "systemd-kexec.service"
+      "systemd-soft-reboot.service"
+      "soft-reboot.target"
     ] ++ lib.optional cfg.package.withUtmp "systemd-update-utmp.service" ++ [
 
       # Password entry.
@@ -695,6 +697,20 @@ in
         create = "0664 root ${config.users.groups.utmp.name}";
         minsize = "1M";
       };
+    };
+
+    # make sure /run/next-system isn't garbage collected before the next boot
+    systemd.tmpfiles.rules = [ "L+ /nix/var/nix/gcroots/next-system - - - - /run/next-system" ];
+
+    # if /run/next-system exists, activate it before  rebooting
+    systemd.services.systemd-soft-reboot-activate-next-system = {
+      wantedBy = [ "soft-reboot.target" ];
+      before = [ "systemd-soft-reboot.service" ];
+      unitConfig = {
+        DefaultDependencies = false;
+        ConditionPathExists = "/run/next-system/activate";
+      };
+      serviceConfig.ExecStart = "/run/next-system/activate";
     };
   };
 


nixos/modules/system/boot/systemd.nix
```diff
@@ -695,6 +697,11 @@ in
         minsize = "1M";
       };
     };
+
+    # make sure /run/next-system isn't garbage collected before the next boot
+    systemd.tmpfiles.rules = [ "L+ /nix/var/nix/gcroots/next-system - - - - /run/next-system" ];
+    # if /run/next-system exists, activate it before  rebooting
+    systemd.services.systemd-soft-reboot.serviceConfig.ExecStartPre = "-/run/next-system/activate";
```
arianvp, line 704: I am getting the following logs:

```
May 07 21:11:02 utm systemd[1]: Reached target System Shutdown.
May 07 21:11:02 utm systemd[1]: Reached target Late Shutdown Services.
May 07 21:11:02 utm systemd[1]: Starting Reboot System Userspace...
May 07 21:11:02 utm (activate)[6377]: systemd-soft-reboot.service: Failed to connect stdout to the journal socket, ignoring: Connection refused
May 07 21:11:02 utm systemd[1]: systemd-soft-reboot.service: Deactivated successfully.
May 07 21:11:02 utm systemd[1]: Finished Reboot System Userspace.
```


`systemd-soft-reboot.service` executes very late (`After=shutdown.target`) .  At this point running `activate` probably doesn't work anymore. Also see the

```
May 07 21:11:02 utm (activate)[6377]: systemd-soft-reboot.service: Failed to connect stdout to the journal socket, ignoring: Connection refused
```


so this probably isn't working as expected

arianvp:
So the `ExecStartPre` here will not log anything to journal as per

https://github.com/systemd/systemd/pull/30220

`systemd-journald` will be stopped before `soft-reboot.target` is reached.   So i will have to do something with unit dependencies instead

arianvp:
Actually I'm confused. `systemd-journald.socket` remains  and will persist across `soft-reboot`

so where is the

```
May 07 21:11:02 utm (activate)[6377]: systemd-soft-reboot.service: Failed to connect stdout to the journal socket, ignoring: Connection refused
```
coming from?

arianvp:
This needs a VM test.  Also this probably can not be merged anymore until the 24.05 cutoff has happened.

nixos/modules/system/activation/switch-to-configuration.pl
```diff
@@ -110,6 +110,7 @@
 }
 
 if ($action eq "boot") {
+    system("@coreutils@/bin/ln", "-sfn", "$toplevel", "/run/next-system") == 0 or exit 1;
```
arianvp, line 113:
```suggestion
    # This symlink is used by soft-reboot
    system("@coreutils@/bin/ln", "-sfn", "$toplevel", "/run/next-system") == 0 or exit 1;
```

arianvp:
Neat follow-up; now that `soft-reboot` can be used as an alternative to `switch-to-configuration switch`

Make a version of `switch-to-configuration.pl` that is:

1. **Not** perl
2. Only does `boot`, `kexec` and `soft-reboot` and nothing else
3.  throws away all the  systemd service start/stopping code as it's not needed anymore

nixos/modules/system/activation/switch-to-configuration.pl
```diff
@@ -110,6 +110,8 @@
 }
 
 if ($action eq "boot") {
+    # /run/next-system/activate is called when systemctl soft-reboot is called
+    system("@coreutils@/bin/ln", "-sfn", "$toplevel", "/run/next-system") == 0 or exit 1;
```
arianvp, line 114: Should we do this on `"switch"` too?

nixos/modules/system/boot/kexec.nix
```diff
@@ -19,7 +20,7 @@
               exit 0
             fi
 
-            p=$(readlink -f /nix/var/nix/profiles/system)
+            p=$(readlink -f /run/next-system)
```
arianvp, line 23: Should this fall back  to `/run/current-system`  if `/run/next-system` doesn't exist?

nixos/modules/system/boot/systemd.nix
```diff
@@ -695,6 +697,20 @@ in
...
+
+    # if /run/next-system exists, activate it before  rebooting
+    systemd.services.systemd-soft-reboot-activate-next-system = {
+      wantedBy = [ "soft-reboot.target" ];
+      before = [ "systemd-soft-reboot.service" ];
+      unitConfig = {
+        DefaultDependencies = false;
+        ConditionPathExists = "/run/next-system/activate";
+      };
+      serviceConfig.ExecStart = "/run/next-system/activate";
```
arianvp, line 712: Should this fall back to `/run/current-system/activate` if `/run/next-system/activate` doesn't exist?

NickCao:
It seems the minimal required part of this PR is the two additional upstream units and the wrapper mount point. It's enough for the `nixos-rebuild boot` then `systemctl soft-reboot` workflow to work as `nixos-rebuild boot` already links the system profile to `/nix/var/nix/profiles/system`. There are not need for the `/run/next-system` shenanigans.

nixos/modules/system/boot/systemd.nix
```diff
@@ -695,6 +697,20 @@ in
         minsize = "1M";
       };
     };
+
+    # make sure /run/next-system isn't garbage collected before the next boot
+    systemd.tmpfiles.rules = [ "L+ /nix/var/nix/gcroots/next-system - - - - /run/next-system" ];
+
+    # if /run/next-system exists, activate it before  rebooting
```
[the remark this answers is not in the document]
arianvp: Activation happens in the initrd  which soft-reboot doesn't transition into. Same reason why we do activation in the initrd before transitioning into stage-2

arianvp:
> It seems the minimal required part of this PR is the two additional upstream units and the wrapper mount point. It's enough for the `nixos-rebuild boot` then `systemctl soft-reboot` workflow to work as `nixos-rebuild boot` already links the system profile to `/nix/var/nix/profiles/system`. There are no need for the `/run/next-system` shenanigans.

Using `/nix/var/nix/profiles/system` is incorrect. See  long-standing issue https://github.com/NixOS/nixpkgs/issues/50300  (which the last commit from this PR fixes; but can maybe be split up into a new PR)

`/nix/var/nix/profiles/system` is not guaranteed to be updated when `nixos-rebuild boot` is run

arianvp:
Moving this back to draft. This needs to cook a bit more.

But just keeping this PR around for notes and feedback

I just realized a pretty serious defect:

soft-reboot re-execs into `SYSTEMD_BINARY_PATH`   (Or into `/run/nextroot/$SYSTEMD_BINARY_PATH`) (https://github.com/systemd/systemd/blob/18303adcd3a9b16339008b5fa909a7c09247072d/src/core/main.c#L1942)

but we set `SYSTEMD_BINARY_PATH=/run/current-system/systemd/lib/systemd` https://github.com/NixOS/nixpkgs/blob/master/pkgs/os-specific/linux/systemd/default.nix#L799

which  means soft-reboot doesn't execute the new systemd version but into the old one.

Maybe we _do_ need to do the `/run/nextroot` population instead

arianvp:
Wait no..   We call `/run/next-sytem/activate` which  creates the  `/run/current-system` symlink! This actually already works. 🥳 

However I want to have a NixOS Test that  asserts this

nixos/modules/system/boot/kexec.nix
```diff
@@ -19,13 +22,12 @@
...
-              exit 1
-            fi
-            echo "Loading NixOS system via kexec."
-            exec kexec --load $p/kernel --initrd=$p/initrd --append="$(cat $p/kernel-params) init=$p/init"
+            for p in /run/next-system /run/current-system; do
+              if [[ -d $p ]]; then
+                echo "Loading NixOS system via kexec."
+                exec kexec --load $p/kernel --initrd=$p/initrd --append="$(cat $p/kernel-params) init=$p/init"
+              fi
+            done
```
ElvishJerricco, line 30: I don't think we should mess with kexec at all in this PR. It's actually got some problems as is and this is only going to make it more confusing.

nixos/modules/system/boot/systemd.nix
```diff
@@ -695,6 +697,20 @@ in
...
+    systemd.services.systemd-soft-reboot-activate-next-system = {
```
ElvishJerricco, line 705: What happened to doing this in `ExecStartPre` in `systemd-soft-reboot.service` with a `-` in the front of the command? That seemed quite elegant to me. With a new unit, I'm less sure that the unit ordering is right.

nixos/modules/system/activation/switch-to-configuration.pl
ElvishJerricco: Doesn't seem necessary. The point of `next-system` is to have a link to activate on soft-reboot, and a `switch` will have already activated the new system. Even on a `soft-reboot`, reactivation shouldn't be necessary.

nixos/modules/system/boot/systemd.nix
ElvishJerricco: Same argument as not creating `next-system` on `switch`; the current system is already activated so it shouldn't be necessary.

nixos/modules/system/boot/systemd.nix
arianvp: I copied the ordering from prepare-kexec. I'm pretty confident this is right.


Using ExecStartPre= It would complain  about it not being able to connect to the journald.

```
May 07 21:11:02 utm (activate)[6377]: systemd-soft-reboot.service: Failed to connect stdout to the journal socket, ignoring: Connection refused
```

soft-reboot.target causes systemd-journald.service to exit before it is reached (has a Conflicts on it) to make sure logs are properly flushed before the switch is made. I think that's what causing this.

nixos/modules/system/boot/kexec.nix
arianvp: Yeh I'll drop the commit (and maybe open a separate or for it)

errnoh:
Is this also the correct issue to follow for support of `systemctl soft-reboot` even in non nixos-rebuild related use cases? I've been using `systemctl switch-root` for switching into different non-nixos installations on the fly but since 24.05 that has been broken. I've been able to continue using them by switching to 23.11 configuration first and then switch-rooting, but would prefer not having to do that. 

Based on the answer from @ElvishJerricco [here](https://www.reddit.com/r/NixOS/comments/184nvl5/systemctl_softreboot_is_not_working/) my assumption is that this PR is the main branch of work for adding that support?

andrevmatos (approved):

SohamG:
Its been around a year - Is this still a candidate for inclusion/merging? I'd love to have this feature.

PR #309969, opened 2024-05-07
mealie: allow writing to the data directory

anoadragon453:
## Description of changes

This allows the mealie process to write to the data directory (specified by the `DATA_DIR` environment variable). Adding this option is required when the `DynamicUser` systemd service option is enabled, which is the case here.

See the systemd documentation for a full explanation: https://www.freedesktop.org/software/systemd/man/latest/systemd.exec.html#DynamicUser=

In my case, this was necessary in order to allow mealie to start up correctly when using a custom `DATA_DIR` value. Therefore I can confirm it works via manual testing.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [x] Tested, as applicable:

---

### DIFF
--- a/nixos/modules/services/web-apps/mealie.nix
+++ b/nixos/modules/services/web-apps/mealie.nix
@@ -33,6 +33,9 @@ in
         permissions and then leave this option as the default value. That user,
         with a stable UID, will then be used and systemd will not dynamically
         generate a new one.
+
+        To reduce confusion, if a custom `DATA_DIR` is configured, then a user
+        with the same name as this option's value MUST exist.
       '';
       example = lib.literalExpression ''
         {
@@ -139,5 +142,29 @@ in
           StateDirectory = null;
         };
     };
+
+    assertions = [
+      {
+        # Assert that a dynamic user is not in use if a custom data directory is
+        # set. This is to help users avoid setting a custom data directory which
+        # the user that mealie is running as cannot write to.
+        assertion = (
+          config.systemd.services.mealie.environment.DATA_DIR == mealieDataDir
+          || config.users.users ? ${cfg.user}
+        );
+        description = ''
+          When a custom data directory is in use (through the `DATA_DIR`
+          environment variable), the `user` option must be set to a system user
+          that actually exists. Otherwise, systemd will dynamically create a user,
+          and it is unlikely that this user will have the appropriate read/write
+          permissions for the custom data directory.
+
+          Please create a user named "${cfg.user}" and give them read/write
+          permissions for the configured mealie data directory.
+
+          See the description of the `user` option for more information.
+        '';
+      }
+    ];
   };
 }
--- a/nixos/modules/services/web-apps/mealie.nix
+++ b/nixos/modules/services/web-apps/mealie.nix
@@ -2,6 +2,9 @@
 let
   cfg = config.services.mealie;
   pkg = cfg.package;
+
+  # The directory where mealie user data is stored.
+  mealieDataDir = "/var/lib/mealie";
 in
 {
   options.services.mealie = {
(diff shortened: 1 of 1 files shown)


nixos/modules/services/web-apps/mealie.nix
```diff
@@ -50,7 +50,7 @@ in
   };
 
   config = lib.mkIf cfg.enable {
-    systemd.services.mealie = {
+    systemd.services.mealie = rec {
```
anoadragon453, line 53: The `rec` is required in order to allow referencing the `environment` key from within the same set.

litchipi (approved):
Good catch !

ambroisie:
I think the simplest/best way would just be to have `DynamicUser = mkForce false;` and add/create the relevant user/group for the service in your own configuration.

This is quite a niche problem. I can't really identify a need to share `DATA_DIR` (unlike, say, Jellyfin, Sonarr, and Radarr which should _probably_ all share the same folders and use the same group/user to access them). IMO the module should be kept simple and use `DynamicUser`.

anoadragon453:
@james-atkins @ambroisie thank you both for your comments. I've changed the way things work here to try and account for the edge cases.

I went with @ambroisie's suggestion of allowing passing a user with the correct permissions into the module, rather than attempting to change the permissions of an external directory on the fly (which users may not expect).

It is now effectively expected for users to create a system user with the appropriate permissions when specifying a custom data directory. Doing so will disable the `DynamicUser` behaviour of the systemd service, which should make things more predictable.

I tried to include a wealth of documentation. Please review and check that it's accurate/useful.

nixos/modules/services/web-apps/mealie.nix
```diff
@@ -59,21 +111,31 @@ in
...
         StandardOutput="journal";
-      };
+      }
+        # If a custom data directory has been set, then allow the user mealie is
+        # running as to write to it. This is only necessary when systemd has
+        # dynamically generated a user to run mealie as. Otherwise, it is a no-op.
+        #
+        # Also, unset the state directory, as mealie won't be writing to it when
+        # a custom data directory is in use.
+        // lib.mkIf (environment.DATA_DIR != "/var/lib/mealie") {
```
litchipi, line 135: Nitpick: I would maybe put the `/var/lib/mealie` inside a variable somewhere so that if its value gets changed, we don't forget this condition

litchipi (approved):
Thanks for taking the time of writing exactly what happens, and why we chose to do things this way ! :+1:

anoadragon453:
@james-atkins @litchipi Thanks! Both suggestions taken into account, and PR rebased on top of latest `master` (to address some merge conflicts).

litchipi (approved):

nixos/modules/services/web-apps/mealie.nix
```diff
@@ -2,13 +2,71 @@
...
+            user = "mealie";
+            settings = {
+              DATA_DIR = "/path/to/data/directory";
+
+              // Additional options...
+            };
+          };
+        }
+      '';
+    };
```
ambroisie, line 68: IMO `example` should be for the option itself, not the module.

nixos/modules/services/web-apps/mealie.nix
ambroisie: Using `config` as a module argument and referencing `config.path.to.the.option` should be preferred -- it deals better with the module system's merging behaviour.

ambroisie (changes requested):
Not a _big_ fan of this option, IMO this is a fringe case. But it is easy to add and doesn't add _much_ complexity 🤷.

ambroisie:
Also, I don't know how `DynamicUser = true;` works when `User` is set to a pre-existing user 🤔 -- that's why I'd suggested it being a downstream user configuration thing through `mkForce` instead.

anoadragon453:
> Also, I don't know how `DynamicUser = true;` works when `User` is set to a pre-existing user 🤔 -- that's why I'd suggested it being a downstream user configuration thing through `mkForce` instead.

In this case, systemd will simply make use of the existing user. A user will not be dynamically created. See the documentation for the option in the PR description.

PR #310018, opened 2024-05-08
Revert "ffmpeg-full: disable avisynth on Darwin"

vcunat:
Reverts NixOS/nixpkgs#309423

The reason has passed.  I expect it's better to be more consistent.


vcunat:
@ofBorg build ffmpeg-full
(I did test it on aarch64-darwin)

Atemu (approved):

PR #293767, opened 2024-03-06
Nix docs: remove `with lib;` from example code

AndersonTorres:
Following [Best Practices](https://nix.dev/guides/best-practices#with-scopes), `with` is a problematic language construction and should be avoided.

Usually it is employed like a "factorization": `[ X.A X.B X.C X.D ]` is written `with X; [ A B C D ]`.

However, as shown in the link above, the syntatical rules of `with` are not so intuitive, and this "distributive rule" is very selective, in the sense that `with X; [ A B C D ]` is not equivalent to `[ X.A X.B X.C X.D ]`.

However, this factorization is still useful to "squeeze" some code, especially in lists like `meta.maintainers`.

On the other hand, it becomes less justifiable in bigger scopes. This is especially true in cases like `with lib;` in the top of expression and in sets like `meta = with lib; { . . . }`.

That being said, this patch removes most of example code in the current documentation.

The exceptions are, for now
- doc/functions/generators.section.md
- doc/languages-frameworks/coq.section.md

because, well, they are way more complicated and I couldn't parse them mentally - another reason why `with` should be avoided!

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
- For non-Linux: Is sandboxing enabled in `nix.conf`? (See [Nix manual](https://nixos.org/manual/nix/stable/command-ref/conf-file.html))
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:
  - [NixOS test(s)](https://nixos.org/manual/nixos/unstable/index.html#sec-nixos-tests) (look inside [nixos/tests](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests))
  - and/or [package tests](https://nixos.org/manual/nixpkgs/unstable/#sec-package-tests)
  - or, for functions and "core" functionality, tests in [lib/tests](https://github.com/NixOS/nixpkgs/blob/master/lib/tests) or [pkgs/test](https://github.com/NixOS/nixpkgs/blob/master/pkgs/test)
  - made sure NixOS tests are [linked](https://nixos.org/manual/nixpkgs/unstable/#ssec-nixos-tests-linking) to the relevant packages
[...]

PR #310158, opened 2024-05-08
tracexec: Packaging improvements

nh2:
Original PR title: `tracexec: init at 0.1.0`

## Description of changes

See https://github.com/kxxt/tracexec

:heart: co-maintainers would be very appreciated!

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/by-name/tr/tracexec/package.nix
+++ b/pkgs/by-name/tr/tracexec/package.nix
@@ -1,37 +1,68 @@
 {
   lib,
-  rustPlatform,
+  stdenv,
   fetchFromGitHub,
+  rustPlatform,
+  cargo-about,
+  nix-update-script,
 }:
-
-rustPlatform.buildRustPackage rec {
+let
   pname = "tracexec";
   version = "0.2.2";
+in
+rustPlatform.buildRustPackage {
+  inherit pname version;
 
   src = fetchFromGitHub {
     owner = "kxxt";
     repo = "tracexec";
-    rev = "v${version}";
+    rev = "refs/tags/v${version}";
     hash = "sha256-X2hLaBndeYLBMnDe2MT4pgZiPj0COHG2uTvAbW+JVd4=";
   };
 
   cargoHash = "sha256-3xANOv+A4soDcKMINy+RnI8l6uS3koZpw3CMIUCmK5A=";
 
-  # Remove test binaries and only retain tracexec
-  postInstall = ''
-    find "$out/bin" -type f \! -name tracexec -print0 | xargs -0 rm -v
+  nativeBuildInputs = [
+    cargo-about
+  ];
+
+  # Remove RiscV64 specialisation when this is fixed:
+  # * https://github.com/NixOS/nixpkgs/pull/310158#pullrequestreview-2046944158
+  # * https://github.com/rust-vmm/seccompiler/pull/72
+  cargoBuildFlags = lib.optional stdenv.hostPlatform.isRiscV64 "--no-default-features";
+
+  preBuild = ''
+    sed -i '1ino-clearly-defined = true' about.toml  # disable network requests
+    cargo about generate --config about.toml -o THIRD_PARTY_LICENSES.HTML about.hbs
   '';
 
-  # ptrace is not allowed in sandbox
-  doCheck = false;
+  # Tests don't work for native non-x86 compilation
+  # because upstream overrides the name of the linker executables,
+  # see https://github.com/NixOS/nixpkgs/pull/310158#issuecomment-2118845043
+  doCheck = stdenv.hostPlatform.isx86_64;
+
+  checkFlags = [
+    "--skip=cli::test::log_mode_without_args_works" # `Permission denied` (needs `CAP_SYS_PTRACE`)
+    "--skip=tracer::test::tracer_emits_exec_event" # needs `/bin/true`
+  ];
+
+  postInstall = ''
+    # Remove test binaries (e.g. `empty-argv`, `corrupted-envp`) and only retain `tracexec`
+    find "$out/bin" -type f \! -name tracexec -print0 | xargs -0 rm -v
+
+    install -Dm644 LICENSE -t "$out/share/licenses/${pname}/"
+    install -Dm644 THIRD_PARTY_LICENSES.HTML -t "$out/share/licenses/${pname}/"
+  '';
+
+  passthru.updateScript = nix-update-script { };
 
   meta = {
+    changelog = "https://github.com/kxxt/tracexec/blob/v${version}/CHANGELOG.md";
     description = "A small utility for tracing execve{,at} and pre-exec behavior";
     homepage = "https://github.com/kxxt/tracexec";
-    changelog = "https://github.com/kxxt/tracexec/blob/${src.rev}/CHANGELOG.md";
-    license = lib.licenses.gpl2;
-    maintainers = with lib.maintainers; [ fpletz ];
+    license = lib.licenses.gpl2Plus;
     mainProgram = "tracexec";
+    maintainers = with lib.maintainers; [ fpletz nh2 ];
     platforms = lib.platforms.linux;
   };
 }


pkgs/by-name/tr/tracexec/package.nix
```diff
@@ -0,0 +1,33 @@
+{ lib
+, stdenv
+, fetchFromGitHub
+, rustPlatform
+}:
```
DontEatOreo, line 5:
```suggestion
{
  lib,
  stdenv,
  fetchFromGitHub,
  rustPlatform,
}:
```

Formatted as per RFC 0166

pkgs/by-name/tr/tracexec/package.nix
```diff
@@ -0,0 +1,33 @@
...
+
+rustPlatform.buildRustPackage rec {
+  pname = "tracexec";
+  version = "0.1.0";
```
DontEatOreo, line 9:
```suggestion
let
  pname = "tracexec";
  version = "0.1.0";
in
rustPlatform.buildRustPackage {
```

Using `rec` is discouraged, please take a look at: https://nix.dev/guides/best-practices#recursive-attribute-set-rec

pkgs/by-name/tr/tracexec/package.nix
```diff
@@ -0,0 +1,33 @@
...
+
+  cargoHash = "sha256-Ry7dwaoEAVWl04ug3MI1UX92+ynnnuwA3jfmA/6X+L8=";
+
+  meta = with lib; {
+    platforms = platforms.linux;
+    description = "A small utility for tracing execve{,at} and pre-exec behavior";
+    mainProgram = "tracexec";
+    homepage = "https://github.com/kxxt/tracexec";
+    license = with licenses; [ gpl2 ];
+    maintainers = with maintainers; [ nh2 ];
```
DontEatOreo, line 31:
```suggestion
  meta = {
    description = "A small utility for tracing execve{,at} and pre-exec behavior";
    homepage = "https://github.com/kxxt/tracexec";
    license = with lib.licenses; [ gpl2 ];
    mainProgram = "tracexec";
    maintainers = with lib.maintainers; [ nh2 ];
    platforms = platforms.linux;
```
- Using `with` is discouraged, please take a look at https://nix.dev/guides/best-practices#with-scopes and https://github.com/NixOS/nixpkgs/pull/293767
- Order alphabetically

4evy (changes requested):

pkgs/by-name/tr/tracexec/package.nix
```diff
@@ -0,0 +1,33 @@
...
+
+  src = fetchFromGitHub {
+    owner = "kxxt";
+    repo = pname;
+    rev = "v${version}";
+    hash = "sha256-HQTNYReJkfiqS05APtLAUf9dnnPYTzF0ratd5QKv+8w=";
+  };
+
+  checkFlags = [
+    "--skip=cli::test::log_mode_without_args_works" # `Permission denied` (maybe needs root?)
```
kxxt, line 19: This test doesn't need root and it only executes `/proc/self/exe` so I think it should be compatible with NixOS. Does the build environment drops some privileges, like  `CAP_SYS_PTRACE`?

pkgs/by-name/tr/tracexec/package.nix
```diff
@@ -0,0 +1,33 @@
...
+  ];
+
+  cargoHash = "sha256-Ry7dwaoEAVWl04ug3MI1UX92+ynnnuwA3jfmA/6X+L8=";
+
+  meta = with lib; {
+    platforms = platforms.linux;
+    description = "A small utility for tracing execve{,at} and pre-exec behavior";
+    mainProgram = "tracexec";
+    homepage = "https://github.com/kxxt/tracexec";
+    license = with licenses; [ gpl2 ];
```
kxxt, line 30: I ship a third party licenses file (generated by cargo-about) alongside tracexec. It would be nice to also include it here if it fits into the workflow of Nix. My PKGBUILD at AUR can be taken as an example: https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=tracexec#n37 .

kxxt (commented):
Thanks for your packaging effort! I have added some review comments and the test and licenses. Apart from that, there's some extra complexity for handling different architectures: `seccomp-bpf` feature isn't available for riscv64 architecture due to seccompiler doesn't support it: https://github.com/rust-vmm/seccompiler/pull/72 .

In general, you can take a look at my PKGBUILD at AUR: https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=tracexec

And a kind reminder that in the future tests should be run single threaded, with `RUST_TEST_THREADS=1`

pkgs/by-name/tr/tracexec/package.nix
nh2: I would be surprised if `CAP_SYS_PTRACE` was available in the build sandbox.

Because `CAP_SYS_PTRACE` allows tracing arbitrary processes; I expect that tracing your own children would be allowed.

pkgs/by-name/tr/tracexec/package.nix
nh2: @kxxt I think that does not work here, because this field is supposed to be a simple string (or list of strings) to facilitate automatic querying.

See https://nixos.org/manual/nixpkgs/stable/#var-meta-license

If there are multiple licenses I should mention in the list, please let me know.

I added now the copying of the license to `share/` like you do in the AUR.

nh2:
Ah, looks like tracexec was already packaged meanwhile:

* #311809

I'll apply the improvements made here onto it.

nh2:
Good to merge from my side; if ofborg passes, please somebody merge.

WilliButz:
@ofborg build tracexec

WilliButz:
Hmm, this seems to introduce a build failure in the check phase on aarch64-linux:

Relevant log:
<details>

```
buildPhase completed in 2 minutes 43 seconds
Running phase: checkPhase
Executing cargoCheckHook
++ cargo test -j 4 --profile release --target aarch64-unknown-linux-gnu --frozen -- --test-threads=4 --skip=cli::test::log_mode_without_args_works --skip=tracer::test::tracer_emits_exec_event
   Compiling proc-macro2 v1.0.81
   Compiling unicode-ident v1.0.12
   Compiling cfg-if v1.0.0
   Compiling libc v0.2.154
   Compiling autocfg v1.3.0
   Compiling memchr v2.7.2
error: linker `aarch64-linux-gnu-gcc` not found
  |
  = note: No such file or directory (os error 2)

error: could not compile `proc-macro2` (build script) due to 1 previous error
warning: build failed, waiting for other jobs to finish...
error: could not compile `libc` (build script) due to 1 previous error
```
</details>

DontEatOreo:
It would be nice to have `updateScript` to automate updates

```nix
passthru.updateScript = gitUpdater {
  url = "https://github.com/kxxt/tracexec";
  rev-prefix = "v";
};
```

pkgs/by-name/tr/tracexec/package.nix
```diff
@@ -1,37 +1,60 @@
...
   pname = "tracexec";
   version = "0.2.2";
+in
+rustPlatform.buildRustPackage {
+  inherit pname version;
 
   src = fetchFromGitHub {
     owner = "kxxt";
-    repo = "tracexec";
+    repo = pname;
```
DontEatOreo, line 17:
```suggestion
    repo = "tracexec";
```

https://github.com/nix-community/nixpkgs-lint/issues/21

pkgs/by-name/tr/tracexec/package.nix
kxxt: > @kxxt I think that does not work here, because this field is supposed to be a simple string (or list of strings) to facilitate automatic querying.
> 
> See https://nixos.org/manual/nixpkgs/stable/#var-meta-license
> 
> If there are multiple licenses I should mention in the list, please let me know.
> 
> I added now the copying of the license to `share/` like you do in the AUR.

Thanks for your explanation! LGTM.

fpletz (commented):
Sorry, didn't see your PR. Thanks! :heart:

pkgs/by-name/tr/tracexec/package.nix
```diff
@@ -1,37 +1,60 @@
...
     rev = "v${version}";
```
DontEatOreo, line 18:
```suggestion
    rev = "refs/tags/v${version}";
```

DontEatOreo: The reason behind this change is that without explicitly referring to either tags or branches, it's unclear what the package is supposed to follow

nh2:
> Hmm, this seems to introduce a build failure in the check phase on aarch64-linux:

```
error: linker `aarch64-linux-gnu-gcc` not found
```

@WilliButz No clue about that one. Should Rust aarch64 have a working GCC out of the box?

kxxt:
> > Hmm, this seems to introduce a build failure in the check phase on aarch64-linux:
> 
> ```
> error: linker `aarch64-linux-gnu-gcc` not found
> ```
> 
> @WilliButz No clue about that one. Should Rust aarch64 have a working GCC out of the box?

I defined it here: https://github.com/kxxt/tracexec/blob/main/.cargo/config.toml . It's because rust defaults to linking with `cc` and that will fail with a recent gcc(e.g. gcc13/14) when cross-compiling from x86_64 host to aarch64 target.

pkgs/by-name/tr/tracexec/package.nix
kxxt: >  I expect that tracing your own children would be allowed.

Ah, I didn't read too much about `CAP_SYS_PTRACE`.

>  I expect that tracing your own children would be allowed.


Yes, that's enough for this test.

nh2:
> passthru.updateScript = gitUpdater { url = "https://github.com/kxxt/tracexec"; rev-prefix = "v"; };

@DontEatOreo I used the more generic

```
passthru.updateScript = nix-update-script { };
```

mentioned in the `Tip` of https://nixos.org/manual/nixpkgs/stable/#var-passthru-updateScript which works for this package and does not copy-pasting the Github URL.

nh2:
> I defined it here: https://github.com/kxxt/tracexec/blob/main/.cargo/config.toml . It's because rust defaults to linking with `cc` and that will fail with a recent gcc(e.g. gcc13/14) when cross-compiling from x86_64 host to aarch64 target.

@kxxt What would be a good solution to that?

(The cross-compilation of the package currently works in nixpkgs btw, I just tested it and edited put the invocation into https://github.com/NixOS/nixpkgs/pull/310158#issuecomment-2118844349, in case that is useful for you).

nh2:
@kxxt

How does this work in the [AUR package](https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=tracexec) (which only has `0.1.0` but that already has the relevant code https://github.com/kxxt/tracexec/commit/b8d29bdc8c6d8c00dfd0d937ae73b54c6f81595b)? Is the compiler just always called/aliased `aarch64-linux-gnu-gcc` there?

nh2:
I suspect `aarch64` worked ([ofborg log](https://github.com/NixOS/nixpkgs/pull/311809/checks?check_run_id=25025004125)) in the initial packaging because that had tests disabled ([`doCheck = false;`](https://github.com/NixOS/nixpkgs/pull/311809/files#diff-5b2ef1684c44de7cc4b71e1faa99bff59cb4d9497e1c207ecd4b3213b4f461fbR26)) and the problem only occurs in `checkPhase`.

I'll disable tests for non-x86_64 archs until a solution is found.

4evy (approved):
LGTM!

kxxt:
> @kxxt
> 
> How does this work in the [AUR package](https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=tracexec) (which only has `0.1.0` but that already has the relevant code [kxxt/tracexec@b8d29bd](https://github.com/kxxt/tracexec/commit/b8d29bdc8c6d8c00dfd0d937ae73b54c6f81595b))? Is the compiler just always called/aliased `aarch64-linux-gnu-gcc` there?

Because when host and target architecture is the same, a compatible symlink is there, at least on Arch Linux. (e.g. aarch64-linux-gnu-gcc symlinked to gcc).

pkgs/by-name/tr/tracexec/package.nix
```diff
@@ -1,37 +1,68 @@
...
+    "--skip=cli::test::log_mode_without_args_works" # `Permission denied` (needs `CAP_SYS_PTRACE`)
+    "--skip=tracer::test::tracer_emits_exec_event" # needs `/bin/true`
+  ];
+
   postInstall = ''
+    # Remove test binaries (e.g. `empty-argv`, `corrupted-envp`) and only retain `tracexec`
     find "$out/bin" -type f \! -name tracexec -print0 | xargs -0 rm -v
+
+    install -Dm644 LICENSE -t "$out/share/licenses/${pname}/"
+    install -Dm644 THIRD_PARTY_LICENSES.HTML -t "$out/share/licenses/${pname}/"
```
Qyriad, line 54: `pname` probably shouldn't be reused for the license directory

Qyriad (changes requested):

nh2:
With disabled tests on aarch64, ofborg passes, merging.

PR #310195, opened 2024-05-08
cc-wrapper: don't use system rm and mktemp

IlyaNiklyaev:
## Description of changes

When response file is enabled (which is default for clang), cc-wrapper uses `rm` and `mktemp` commands from external `$PATH`. It breaks clang wrapper on systems w/o such tools, and using system tools is not a good idea in general. So we just take these tools from coretools explicitly.

On `master`:
```
$ PATH= $(nix build .#clang_16 --print-out-paths)/bin/c++
/nix/store/1n3cfg8hnh6izkjfnw0xwdz7h7wszyv4-clang-wrapper-16.0.6/bin/c++: line 249: mktemp: No such file or directory
```

On this PR:
```
$ PATH= $(nix build .#clang_16 --print-out-paths)/bin/c++
clang-16: error: no input files
```

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [x] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/build-support/cc-wrapper/cc-wrapper.sh
+++ b/pkgs/build-support/cc-wrapper/cc-wrapper.sh
@@ -246,8 +246,8 @@ if [[ -e @out@/nix-support/cc-wrapper-hook ]]; then
 fi
 
 if (( "${NIX_CC_USE_RESPONSE_FILE:-@use_response_file_by_default@}" >= 1 )); then
-    responseFile=$(mktemp "${TMPDIR:-/tmp}/cc-params.XXXXXX")
-    trap 'rm -f -- "$responseFile"' EXIT
+    responseFile=$(@coreutils_bin@/bin/mktemp "${TMPDIR:-/tmp}/cc-params.XXXXXX")
+    trap '@coreutils_bin@/bin/rm -f -- "$responseFile"' EXIT
     printf "%q\n" \
        ${extraBefore+"${extraBefore[@]}"} \
        ${params+"${params[@]}"} \


matshch:
@ofborg build tests.cc-wrapper

pkgs/build-support/cc-wrapper/cc-wrapper.sh
```diff
@@ -246,8 +247,8 @@ if [[ -e @out@/nix-support/cc-wrapper-hook ]]; then
 fi
 
 if (( "${NIX_CC_USE_RESPONSE_FILE:-@use_response_file_by_default@}" >= 1 )); then
-    responseFile=$(mktemp "${TMPDIR:-/tmp}/cc-params.XXXXXX")
-    trap 'rm -f -- "$responseFile"' EXIT
+    responseFile=$(PATH=$path_utils mktemp "${TMPDIR:-/tmp}/cc-params.XXXXXX")
+    trap 'PATH=$path_utils rm -f -- "$responseFile"' EXIT
```
sdht0, line 251: Why does PATH need to be explicitly set here again when it is already set in line 15/16?

matshch: It is being restored to its original value in line 240.

IlyaNiklyaev: Yep, that's why this PR is needed. PATH restoration on line 240 is done for the hook a couple lines after. We could place PATH restoration after the hook and this response file-related code, but I'm not sure it won't break hooks somewhere.

sdht0: Ah right. Maybe we move that line inside the if and else instead of creating another variable?

sdht0: Although `cc-wrapper-hook` probably relies on the original PATH. So nvm.

matshch:
It is pretty usual for ofborg to timeout on stdenv rebuilds, at least we have tried. I'll try running these tests myself, starting with aarch64-darwin, but it will take some time.

IlyaNiklyaev:
> I'll try running these tests myself, starting with aarch64-darwin

Let's share this work :) I could run some tests tomorrow.

matshch:
I've got the same undefined reference errors on aarch64-linux on nixos-unstable with and without this PR, looks like #285016.

IlyaNiklyaev:
> I've got the same undefined reference errors on aarch64-linux on nixos-unstable with and without this PR, looks like #285016.

Oops, I mean `aarch64-linux` here, not `aarch64-darwin` :( https://github.com/NixOS/nixpkgs/pull/310195#issuecomment-2105215370

pkgs/build-support/cc-wrapper/cc-wrapper.sh
IlyaNiklyaev: I think I found a better solution: instead of playing with `$PATH` variable we may just call `rm` and `mktemp` via absolute paths like `@coreutils_bin@/bin/mktemp`. Also it makes our intent to not use system paths much clearer. I'll try this and return :)

IlyaNiklyaev: Yep, that worked.

IlyaNiklyaev:
I think it looks good now. @Ericson2314 could you please take a look on this PR?

matshch (approved):
This patch seems to work, and `tests.cc-wrapper` builds successfully on x86_64-linux.

matshch:
On aarch64-linux I have successfully built:
* tests.cc-wrapper.default
* tests.cc-wrapper.llvmTests (it has recurseIntoAttrs)
* tests.cc-wrapper.gccTests.gccStdenv
* tests.cc-wrapper.gccTests.gcc8Stdenv
* tests.cc-wrapper.gccTests.gcc9Stdenv
* tests.cc-wrapper.gccTests.gcc10Stdenv
* tests.cc-wrapper.gccTests.gcc11Stdenv
* tests.cc-wrapper.gccTests.gcc12Stdenv
* tests.cc-wrapper.gccTests.gcc13Stdenv

Next attributes are broken in nixos-unstable, nixos-23.11 and in this PR:
* tests.cc-wrapper.gccTests.gcc49Stdenv (it has some other issue about atomics, probably too old for them)
* tests.cc-wrapper.gccTests.gcc6Stdenv (due to #285016)
* tests.cc-wrapper.gccTests.gcc7Stdenv (due to #285016)
* tests.cc-wrapper.supported (it depends on both attributes above)

matshch:
tests.cc-wrapper builds just fine on aarch64-darwin and x86_64-darwin.

cormacrelf:
Sorry, I missed this PR when I filed #316674. The fix in #316700 seems better to me because `@coreutils_bin@` is sometimes empty, when the `nativeTools` flag in `cc-wrapper/default.nix` is true. In which case this approach ends up as `/bin/mktemp`. Which is not quite correct, but will limp on long enough that someone will rediscover this issue in 5 years.

IlyaNiklyaev:
@cormacrelf yep, that's unfortunate, I'll try to fix that. The PR you mentioned has it's own issue, as I can see https://github.com/NixOS/nixpkgs/pull/316700/files#r1623952918

Artturin:
https://github.com/NixOS/nixpkgs/pull/349589 (Added you as a co-author in the commit)

https://github.com/NixOS/nixpkgs/pull/349585

PR #310309, opened 2024-05-09
root: 6.30.06 -> 6.32.00

guitargeek:
ROOT 6.32.00 came out today.


nicoonoclaste (commented):
Looks good, aside from the release not being out yet

guitargeek:
Hi @nbraud, you're right I used the wrong hash! It should work now.

@veprbl, FYI

guitargeek:
Oh okay, yes that makes sense. I'll look into it. Root requires C++ 17 or 20, and all libraries that are build with ROOT as a dependency need to be built with the same C++ standard for these dependents to work in the context of the ROOT C++ Interpreter, called Cling.

veprbl:
Result of `nixpkgs-review pr 310309` run on x86_64-linux [1](https://github.com/Mic92/nixpkgs-review)
<details>
  <summary>8 packages built:</summary>
  <ul>
    <li>hepmc3</li>
    <li>ns-3</li>
    <li>python311Packages.hepmc3</li>
    <li>python312Packages.hepmc3</li>
    <li>root</li>
    <li>root.debug</li>
    <li>xfitter</li>
    <li>yoda-with-root</li>
  </ul>
</details>

veprbl (approved):

pkgs/applications/science/misc/root/default.nix
```diff
@@ -90,15 +88,14 @@ stdenv.mkDerivation rec {
...
-    llvm_13
+    llvm_16
     lz4
     xz
     gsl
     gtest
     openblas
     openssl
     xxHash
-    libAfterImage
```
veprbl, line 101: Why is that? https://github.com/root-project/root/tree/master/graf2d/asimage/src/libAfterImage is still in the tree.

guitargeek: Good question! Root is now forcing to statically link against builtin lol afterimage because it contains custom patches. I was not with my laptop today so I have a hard time finding the commit that made this change, but I'll upgrade on this tomorrow

veprbl: If libAfterImage is not using anymore, we need to remove the package (and alias needs to be added to `pkgs/top-level/aliases.nix`).

guitargeek: Ok! Here is the relevant ROOT PR, for the record: https://github.com/root-project/root/pull/14395

PR #310345, opened 2024-05-09
nixos/wireguard: add option preShutdown for commands called before interface deletion

f2k1de:
add option preShutdown for commands called before interface deletion

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/nixos/modules/services/networking/wireguard.nix
+++ b/nixos/modules/services/networking/wireguard.nix
@@ -80,6 +80,15 @@ let
         description = "Commands called at the end of the interface setup.";
       };
 
+      preShutdown = mkOption {
+        example = literalExpression ''"''${pkgs.iproute2}/bin/ip netns del foo"'';
+        default = "";
+        type = with types; coercedTo (listOf str) (concatStringsSep "\n") lines;
+        description = ''
+          Commands called before shutting down the interface.
+        '';
+      };
+
       postShutdown = mkOption {
         example = literalExpression ''"''${pkgs.openresolv}/bin/resolvconf -d wg0"'';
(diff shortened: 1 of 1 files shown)


nixos/modules/services/networking/wireguard.nix
```diff
@@ -80,6 +80,15 @@ let
         description = "Commands called at the end of the interface setup.";
       };
 
+      preShutdown = mkOption {
+        example = literalExpression ''"''${pkgs.iproute2}/bin/ip netns del foo"'';
+        default = "";
+        type = with types; coercedTo (listOf str) (concatStringsSep "\n") lines;
```
pluiedev, line 86: I think `lines` by itself already suffices.

pluiedev (commented):
The prefix for module changes should start with `nixos/`, so the title of this PR and commit should be prefixed with `nixos/wireguard` instead

nixos/modules/services/networking/wireguard.nix
f2k1de: This is copied from the other options which do the same (but at different locations). 
Then this might be changed for all those options and not just this. I would like to do that in a seperate PR then, because it does not just affects the preShutdown option but is more a cleanup for all those options.

pluiedev (approved):

SuperSandro2000 (approved):

Issue #307159, opened 2024-04-27
nixos/systemd-sysusers: silently ignores password and hashedPassword


Princemachiavelli:
### Describe the bug
If `systemd.sysusers.enable` is used with users having passwords set by `hashedPassword`, the defined password is ignored and the users are provisioned as password-less (i.e "disabled").

### Steps To Reproduce
1. Enable `systemd.sysusers.enable`
2. Enable disable `users.mutableUsers` or enable `system.etc.overlay.enable`. (I only tested via `mutableUsers=false`).
3. Test new config (careful!)
4. No user will be able to login via password - including root.

### Expected behavior
Users should retain existing password configured via `hashedPassword`.

The existing `users-groups` module (`./nixos/modules/config/users-groups.nix`) actually sets the default opposite of what the new systemd-sysusers module expects. `users-groups` sets `hashedPassword` and `password` based on the `initialHashedPassword` and `initialPassword` values while `systemd-sysusers` *only* looks in the `initial*` values. `systemd-sysfiles` should just look at `hasedPassword` and `password`.

The assertions that prevent building a config without any login-able user (locally or via SSH) should be reviewed. I believe they were not triggered because while I did have an SSH public key configured via Nix - the config does not have an SSH server enabled.

Princemachiavelli:
Also, the default UID of users does not start at 1000 and often results in an UID < 1000. `systemd-sysusers` doesn't really implement a distinction for `isSystemUser` or `isNormalUser` since it's intended to be used only for system users.

One option is to run systemd-sysusers twice; once for "normal" users and once for "system" users. In each call set the range of allowed UID/GID values via the `r` line option. I *think* just setting the `r` line type once will restrict all (dynamic) UID/GID allocations.

poscat0x04:
@Princemachiavelli Hi do you mind creating a separate issue for the UID issue? These are orthogonal and I think are better off tracked separately.

con-f-use:
https://github.com/NixOS/nixpkgs/issues/318365 for cross-reference.

devurandom:
> Also, the default UID of users does not start at 1000 and often results in an UID < 1000. `systemd-sysusers` doesn't really implement a distinction for `isSystemUser` or `isNormalUser` since it's intended to be used only for system users.

This affects even manually specified uids: They are ignored and instead a random uid <1000 is assigned.

PR #296562, opened 2024-03-17
zsh-zhooks: init at 0-unstable-10-31-2021

fidgetingbits:
## Description of changes

Add myself to maintainers list and add a new simple zsh plugin as my first package.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [x] aarch64-darwin
- For non-Linux: Is sandboxing enabled in `nix.conf`? (See [Nix manual](https://nixos.org/manual/nix/stable/command-ref/conf-file.html))
  - [ ] `sandbox = relaxed`
  - [x] `sandbox = true`
- [ ] Tested, as applicable:
  - [NixOS test(s)](https://nixos.org/manual/nixos/unstable/index.html#sec-nixos-tests) (look inside [nixos/tests](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests))
  - and/or [package tests](https://nixos.org/manual/nixpkgs/unstable/#sec-package-tests)
  - or, for functions and "core" functionality, tests in [lib/tests](https://github.com/NixOS/nixpkgs/blob/master/lib/tests) or [pkgs/test](https://github.com/NixOS/nixpkgs/blob/master/pkgs/test)
  - made sure NixOS tests are [linked](https://nixos.org/manual/nixpkgs/unstable/#ssec-nixos-tests-linking) to the relevant packages
- [ ] Tested compilation of all packages that depend on this change using `nix-shell -p nixpkgs-review --run "nixpkgs-review rev HEAD"`. Note: all changes have to be committed, also see [nixpkgs-review usage](https://github.com/Mic92/nixpkgs-review#usage)
- [x] Tested basic functionality of all binary files (usually in `./result/bin/`)
- [24.05 Release Notes](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2405.section.md) (or backporting [23.05](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2305.section.md) and [23.11](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2311.section.md) Release notes)
  - [ ] (Package updates) Added a release notes entry if the change is major or breaking
  - [ ] (Module updates) Added a release notes entry if the change is significant
  - [ ] (Module addition) Added a release notes entry if adding a new NixOS module
[...]

PR #310484, opened 2024-05-10
Correct password option docs and add related tests

fidgetingbits:
## Description of changes

This attempts to clarify the true behavior of password option overrides, as the documentation seems to be wrong (and appears to have been for at least 10 years). These docs also clarify systemd-sysusers quirks that I coincidentally noticed someone happened to file recently [here](https://github.com/NixOS/nixpkgs/issues/307159) as well.

The relevant option docs have been updated, but more importantly I created two new tests to try to show the true behavior of the overrides in their current form. This way if anything ever changes on this front it will be caught and docs can be updated accordingly.

If you want to run the tests I added, you will have to change the maintainer for now as my being added to maintainers is pending [this PR](https://github.com/NixOS/nixpkgs/pull/296562) being merged.

I've not used Nix for very long, so would appreciate someone who has dealt with this code in the past to validate what I've found. @nikstur and @NeQuissimus, perhaps you could both take a look as I've repurposed some of your tests that touch the same areas, so I suspect you'll both be familiar. 

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [x] Tested, as applicable:

---

### DIFF
--- a/nixos/modules/config/users-groups.nix
+++ b/nixos/modules/config/users-groups.nix
@@ -53,37 +53,75 @@ let
         "*"    # password unset
       ]);
 
+  overrideOrderMutable = ''
+    {option}`initialHashedPassword` -> {option}`initialPassword` -> {option}`hashedPassword` -> {option}`password` -> {option}`hashedPasswordFile`
+  '';
+
+  overrideOrderImmutable = ''
+    {option}`initialHashedPassword` -> {option}`hashedPassword` -> {option}`initialPassword` -> {option}`password` -> {option}`hashedPasswordFile`
+  '';
+
+  overrideOrderText = isMutable: ''
+    If the option {option}`users.mutableUsers` is
+    `${if isMutable then "true" else "false"}`, then the order of precedence is as shown
+    below, where values on the left are overridden by values on the right:
+    ${if isMutable then overrideOrderMutable else overrideOrderImmutable}
+  '';
+
+  multiplePasswordsWarning = ''
+    If multiple of these password options are set at the same time then a
+    specific order of precedence is followed, which can lead to surprising
+    results. The order of precedence differs depending on whether the
+    {option}`users.mutableUsers` option is set.
+  '';
+
+  overrideDescription = ''
+    ${multiplePasswordsWarning}
+
+    ${overrideOrderText false}
+
+    ${overrideOrderText true}
+  '';
+
   passwordDescription = ''
-    The options {option}`hashedPassword`,
-    {option}`password` and {option}`hashedPasswordFile`
-    controls what password is set for the user.
-    {option}`hashedPassword` overrides both
-    {option}`password` and {option}`hashedPasswordFile`.
-    {option}`password` overrides {option}`hashedPasswordFile`.
-    If none of these three options are set, no password is assigned to
-    the user, and the user will not be able to do password logins.
-    If the option {option}`users.mutableUsers` is true, the
-    password defined in one of the three options will only be set when
-    the user is created for the first time. After that, you are free to
-    change the password with the ordinary user management commands. If
-    {option}`users.mutableUsers` is false, you cannot change
-    user passwords, they will always be set according to the password
-    options.
+    The {option}`initialHashedPassword`, {option}`hashedPassword`,
+    {option}`initialPassword`, {option}`password` and
+    {option}`hashedPasswordFile` options all control what password is set for
+    the user.
+
+    In a system where [](#opt-systemd.sysusers.enable) is `false`, typically
+    only one of {option}`hashedPassword`, {option}`password`, or
+    {option}`hashedPasswordFile` will be set.
+
+    In a system where [](#opt-systemd.sysusers.enable) is `true`, typically
+    only one of {option}`initialPassword`, {option}`initialHashedPassword`,
+    or {option}`hashedPasswordFile` will be set.
+
+    If the option {option}`users.mutableUsers` is true, the password defined
+    in one of the above password options will only be set when the user is
+    created for the first time. After that, you are free to change the
+    password with the ordinary user management commands. If
+    {option}`users.mutableUsers` is false, you cannot change user passwords,
+    they will always be set according to the password options.
+
+    If none of the password options are set, then no password is assigned to
+    the user, and the user will not be able to do password-based logins.
+
+    ${overrideDescription}
   '';
 
   hashedPasswordDescription = ''
     To generate a hashed password run `mkpasswd`.
 
-    If set to an empty string (`""`), this user will
-    be able to log in without being asked for a password (but not via remote
-    services such as SSH, or indirectly via {command}`su` or
-    {command}`sudo`). This should only be used for e.g. bootable
-    live systems. Note: this is different from setting an empty password,
-    which can be achieved using {option}`users.users.<name?>.password`.
+    If set to an empty string (`""`), this user will be able to log in without
+    being asked for a password (but not via remote services such as SSH, or
+    indirectly via {command}`su` or {command}`sudo`). This should only be used
+    for e.g. bootable live systems. Note: this is different from setting an
+    empty password, which can be achieved using
+    {option}`users.users.<name?>.password`.
 
-    If set to `null` (default) this user will not
-    be able to log in using a password (i.e. via {command}`login`
-    command).
+    If set to `null` (default) this user will not be able to log in using a
+    password (i.e. via {command}`login` command).
   '';
 
   userOpts = { name, config, ... }: {
@@ -281,6 +319,7 @@ let
         default = null;
         description = ''
           Specifies the hashed password for the user.
+
           ${passwordDescription}
           ${hashedPasswordDescription}
         '';
@@ -294,6 +333,7 @@ let
           Warning: do not set confidential information here
           because it is world-readable in the Nix store. This option
           should only be used for public accounts.
+
           ${passwordDescription}
         '';
       };
@@ -307,6 +347,7 @@ let
           password. The password file is read on each system activation. The
           file should contain exactly one line, which should be the password in
           an encrypted form that is suitable for the `chpasswd -e` command.
+
           ${passwordDescription}
         '';
       };
@@ -329,9 +370,7 @@ let
           {command}`passwd` command. Otherwise, it's
           equivalent to setting the {option}`hashedPassword` option.
 
-          Note that the {option}`hashedPassword` option will override
-          this option if both are set.
-
+          ${passwordDescription}
           ${hashedPasswordDescription}
         '';
       };
@@ -351,8 +390,7 @@ let
           used for guest accounts or passwords that will be changed
           promptly.
 
-          Note that the {option}`password` option will override this
-          option if both are set.
+          ${passwordDescription}
         '';
       };
 
@@ -960,11 +998,12 @@ in {
           (filter (x: x != null) (map (flip getAttr user) passwordOptions));
       in optional (!unambiguousPasswordConfiguration) ''
         The user '${user.name}' has multiple of the options
-        `hashedPassword`, `password`, `hashedPasswordFile`, `initialPassword`
-        & `initialHashedPassword` set to a non-null value.
-        The options silently discard others by the order of precedence
-        given above which can lead to surprising results. To resolve this warning,
-        set at most one of the options above to a non-`null` value.
+        `initialHashedPassword`, `hashedPassword`, `initialPassword`, `password`
+        & `hashedPasswordFile` set to a non-null value.
+
+        ${multiplePasswordsWarning}
+
+        ${overrideOrderText cfg.mutableUsers}
 
         The values of these options are:
         ${concatMapStringsSep
--- a/nixos/modules/config/users-groups.nix
+++ b/nixos/modules/config/users-groups.nix
@@ -53,13 +53,9 @@ let
         "*"    # password unset
       ]);
 
-  overrideOrderMutable = ''
-    {option}`initialHashedPassword` -> {option}`initialPassword` -> {option}`hashedPassword` -> {option}`password` -> {option}`hashedPasswordFile`
-  '';
+  overrideOrderMutable = ''{option}`initialHashedPassword` -> {option}`initialPassword` -> {option}`hashedPassword` -> {option}`password` -> {option}`hashedPasswordFile`'';
 
-  overrideOrderImmutable = ''
-    {option}`initialHashedPassword` -> {option}`hashedPassword` -> {option}`initialPassword` -> {option}`password` -> {option}`hashedPasswordFile`
-  '';
+  overrideOrderImmutable = ''{option}`initialHashedPassword` -> {option}`hashedPassword` -> {option}`initialPassword` -> {option}`password` -> {option}`hashedPasswordFile`'';
 
   overrideOrderText = isMutable: ''
     If the option {option}`users.mutableUsers` is
@@ -1002,9 +998,7 @@ in {
         & `hashedPasswordFile` set to a non-null value.
 
         ${multiplePasswordsWarning}
-
         ${overrideOrderText cfg.mutableUsers}
-
         The values of these options are:
         ${concatMapStringsSep
           "\n"
--- a/nixos/tests/password-option-override-ordering.nix
+++ b/nixos/tests/password-option-override-ordering.nix
@@ -1,0 +1,171 @@
+let
+  password1 = "foobar";
+  password2 = "helloworld";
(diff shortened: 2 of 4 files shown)


fidgetingbits:
@Ma27 any thoughts on this PR or know who should review it? I just noticed today you recently made some recent changes that lead to a merge conflict (which I have yet to fix), so just pinging to see if you have want to take a look or have an idea who to ping to get this reviewed.

Ma27:
Last times I touched this part, it was kinda hard to find anybody who felt responsible.
Given that this seems to be a correctness fix mostly, I'll try to take a look soonish.

fidgetingbits:
Merge conflicts should be fixed now

nixos/modules/config/users-groups.nix
```diff
@@ -352,7 +357,8 @@ let
           promptly.
 
           Note that the {option}`password` option will override this
-          option if both are set.
+          option if both are set. One exception to this rule is if
+          `systemd.sysusers.enable = true;`.
```
Ma27, line 361: Hmm after reading through the code it seems to me as if all non-initial password options would override this.

nixos/modules/config/users-groups.nix
```diff
@@ -69,6 +69,10 @@ let
     {option}`users.mutableUsers` is false, you cannot change
     user passwords, they will always be set according to the password
     options.
+
+    NOTE: one exception is if `systemd.sysusers.enable = true;`. In this case
```
Ma27, line 73: I think you can cross-reference this option with `[](#opt-systemd.sysusers.enable)`.

nixos/modules/config/users-groups.nix
```diff
@@ -960,11 +966,14 @@ in {
...
-        `hashedPassword`, `password`, `hashedPasswordFile`, `initialPassword`
-        & `initialHashedPassword` set to a non-null value.
+        `initialHashedPassword`, `hashedPassword`, `initialPassword`, `password`,
+        & `hashedPasswordFile` set to a non-null value.
         The options silently discard others by the order of precedence
         given above which can lead to surprising results. To resolve this warning,
         set at most one of the options above to a non-`null` value.
+        NOTE: If `systemd.sysusers.enable = true;` then the the options and order
+        of precedence are: `initialPassword`,  `initialHashedPassword`,
+        & `hashedPasswordFile`.
```
Ma27, line 976: What about the other two options?

nixos/modules/config/users-groups.nix
```diff
@@ -960,11 +966,14 @@ in {
           (filter (x: x != null) (map (flip getAttr user) passwordOptions));
       in optional (!unambiguousPasswordConfiguration) ''
         The user '${user.name}' has multiple of the options
-        `hashedPassword`, `password`, `hashedPasswordFile`, `initialPassword`
-        & `initialHashedPassword` set to a non-null value.
+        `initialHashedPassword`, `hashedPassword`, `initialPassword`, `password`,
```
Ma27, line 969: The trailing comma should go here: there's an `&` the next char.

fidgetingbits: TIL

nixos/modules/config/users-groups.nix
fidgetingbits: The other two options aren't supported when using `systemd.sysusers.enable`. I changed that to use some other (hopefully more clear) wording from elsewhere. So should be addressed, but let me know if you still think it's too unclear.

nixos/modules/config/users-groups.nix
fidgetingbits: I might not understand what you mean, but I think the test on line 35 of nixos/tests/password-option-override-ordering.nix that I added shows that initialPassword actually overrides hashedPassword (a non-initial password option), which conflicts with what you are saying?

fidgetingbits:
Just a note that to run the tests I added you can run the following from nixpkgs root folder: `nix-build . -A nixosTests.password-option-override-ordering` and `nix-build . -A nixosTests.systemd-sysusers-password-option-override-ordering`. 

Looks like the latter is failing now due to a new assertion about user types, so I will have to investigate.

fidgetingbits:
Looks like there are more aggressive asserts that helps with some of the systemd-sysusers stuff, so I removed a couple of assertions from that test that no longer will run. Also changed the user type and added groups to ensure other new assertions pass so the test actually runs.

I've left the changes as separate commit now for sake of reference, but can squash them before merge.

nixos/tests/password-option-override-ordering.nix
```diff
@@ -0,0 +1,151 @@
...
+      with subtest("Test initialPassword override"):
+          machine.send_key("alt-f2")
+          machine.wait_until_succeeds("[ $(fgconsole) = 2 ]")
+          machine.wait_for_unit("getty@tty2.service")
+          machine.wait_until_succeeds("pgrep -f 'agetty.*tty2'")
+          machine.wait_until_tty_matches("2", "login: ")
+          machine.send_chars("egon\n")
+          machine.wait_until_tty_matches("2", "login: egon")
+          machine.wait_until_succeeds("pgrep login")
+          machine.sleep(2)
```
Ma27, line 129: This seems a little brittle as it only depends on the timing. After reading through all the checks: is that really necessary?

nixos/tests/password-option-override-ordering.nix
```diff
@@ -0,0 +1,151 @@
...
+        print(machine.succeed("getent passwd cat"))
+        assert "${hashed_bcrypt}" in machine.succeed("getent shadow cat"), "cat user password is not correct"
+
+      with subtest("dan user has correct password"):
+        print(machine.succeed("getent passwd dan"))
+        assert "${hashed_bcrypt}" in machine.succeed("getent shadow dan"), "dan user password is not correct"
+
+      with subtest("greg user has correct password"):
+        print(machine.succeed("getent passwd greg"))
+        assert "${hashed_sha512crypt}" not in machine.succeed("getent shadow greg"), "greg user password is not correct"
```
Ma27, line 115: We could match the output of the hashed password from `getent` against the password set for greg here I think.

Also, it's a little weird that you return `passwd` rather than shadow above?

nixos/modules/config/users-groups.nix
```diff
@@ -57,9 +57,9 @@ let
     The options {option}`hashedPassword`,
     {option}`password` and {option}`hashedPasswordFile`
     controls what password is set for the user.
-    {option}`hashedPassword` overrides both
-    {option}`password` and {option}`hashedPasswordFile`.
-    {option}`password` overrides {option}`hashedPasswordFile`.
+    {option}`hashedPasswordFile` overrides both
+    {option}`password` and {option}`hashedPassword`.
+    {option}`hashedPassword` overrides {option}`password`.
```
Ma27, line 62: Line 57 of the password-option-* test says `hashedPassword -> ... -> password`, so the override is the other way round here, right?

nixos/modules/config/users-groups.nix
Ma27: Only for `mutableUsers = false`; though, see https://github.com/NixOS/nixpkgs/blob/400e3ecf59205c6197e093e571e6518a669c4d55/nixos/modules/config/users-groups.nix#L417-L421

nixos/modules/config/users-groups.nix
Ma27: Also given that this also depends on `mutableUsers` somehow, I guess this is also worth mentioning.

Ma27 (changes requested):
Oof thank you for investing time into this! Using NixOS for almost 8 years now and I had no idea _how_ much of a mess that is...

Ma27:
Also asked in the nixos-systemd Matrix channel for more reviewers for the systemd-sysusers part.

Ma27:
I'd really like to see this in 24.11 when it's good to go: this isn't a behavioral fix, but clarifies the current behavior in both warnings and the manual (and locks it down with a test).

nixos/tests/password-option-override-ordering.nix
Ma27: > Can only check exact match when providing the hash yourself.

Sure, but there are tools that you can use to verify the hash against the plaintext password.

> By return you mean what is printed ya? I switched them all to print getent shadow. Makes sense.

Correct.

nixos/tests/password-option-override-ordering.nix
Ma27: Oof.

But well, shadow is part of the tested set, so it seems to be surprisingly stable :shrug:

nixos/tests/password-option-mutable-override-ordering.nix
```diff
@@ -0,0 +1,150 @@
...
+      maintainers = [ fidgetingbits ];
+    };
+
+    nodes.machine =
+      { pkgs, ... }:
+      {
+        environment.systemPackages = [ pkgs.shadow ];
+
+        users = {
+          mutableUsers = true;
```
Ma27, line 22: Wouldn't it make sense to merge this with the immutable test, i.e. have two nodes in there and deduplicate the config and assertions as much as possible?

nixos/tests/password-option-override-ordering.nix
fidgetingbits: > > Can only check exact match when providing the hash yourself.
> 
> Sure, but there are tools that you can use to verify the hash against the plaintext password.

Ok. I'll add that

nixos/tests/password-option-mutable-override-ordering.nix
fidgetingbits: Done, also deduplicated the login test part.

nixos/tests/password-option-override-ordering.nix
fidgetingbits: Actually I spent some time digging around / testing while I was deduplicating the code and it turns out it's easy to drop the sleep. I'll plan to send some PRs to drop the sleeps from `shadow.nix` and any others I find to speed up tests in the future.

fidgetingbits:
I still need to add the clearer wording for the greg case with mutableUsers difference, but otherwise all other feedback should be addressed now.

nixos/modules/config/users-groups.nix
```diff
@@ -329,8 +333,10 @@ let
           {command}`passwd` command. Otherwise, it's
           equivalent to setting the {option}`hashedPassword` option.
 
-          Note that the {option}`hashedPassword` option will override
-          this option if both are set.
+          Note that, when {option}`users.mutableUsers` is true, the {option}`hashedPassword`
+          option will override both the {option}`user.initialHashedPassword` option and
+          `{option}`user.initialPassword` if set. One exception to this rule is if
+          [](#opt-systemd.sysusers.enable) is true.
```
EmergentMind, line 339: What is the exception?
Perhaps editing to something like: "... is true, in which case only `initialPassword`, `initialHashedPassword`, and `hashedPasswordFile` options will be used."

This also applies to Note one line 360

EmergentMind (changes requested):
Some suggestions to add clarity and improve some grammar. 

I also reviewed the various cases of precedence depending on the values of mutable and/or sysUsers. All the option descriptions and related tests appear to be consistent.

fidgetingbits:
I took a stab at rewording the whole thing based on EmergentMind's feedback and because I found myself getting confused reading it even though I'm working on it. That allowed me to duplicate some text from other options, which makes `passwordDescription` the single source of truth, which will be nice for future tweaks. 

One part I didn't include which I considered putting at the end of `passwordDescription` is:

```
    If [](#opt-systemd.sysusers.enable) is `true`, then the order of password is:
    `initialPassword` -> `initialHashedPassword` -> `hashedPasswordFile`
```

Since technically that order still matches what is shown earlier (just with password, hashedPassword, hashedPasswordFile omitted), I wasn't sure if I should explicitly include it or not. Happy to add it if others thing it's worth.

nixos/modules/config/users-groups.nix
```diff
@@ -960,12 +979,16 @@ in {
           (filter (x: x != null) (map (flip getAttr user) passwordOptions));
       in optional (!unambiguousPasswordConfiguration) ''
         The user '${user.name}' has multiple of the options
-        `hashedPassword`, `password`, `hashedPasswordFile`, `initialPassword`
-        & `initialHashedPassword` set to a non-null value.
+        `initialHashedPassword`, `hashedPassword`, `initialPassword`, `password`
+        & `hashedPasswordFile` set to a non-null value.
         The options silently discard others by the order of precedence
```
fidgetingbits, line 984: I guess this is not true because of the mutableUsers true/false difference, so the order of precedence isn't necessary as shown above. Would be nice to refer to a non-duplicated source of truth here as well, so maybe I'll add a separate variable with the order for mutable false/true text and then include that both in passwordDescription and here.. 

I also wonder if the order of precedence difference between mutableUsers true/false is intentional or accidental. If the former, I wonder who would know, and if the latter, I think it should just be changed for a future release so the orders are consistent?

fidgetingbits:
Good for another review pass I think. In order to fix the assert warning being wrong, and make it harder to mess up the override order text in various places, I tried to deduplicate it all and use some variable expansion to help generate it in various places.

EmergentMind (commented):
The rewording and de-duplication is a big improvement for clarity. Nice work.

nixos/modules/config/users-groups.nix
```diff
@@ -53,37 +53,80 @@ let
         "*"    # password unset
       ]);
 
+  overrideOrderMutable = ''
+    {option}`initialHashedPassword` -> {option}`initialPassword` ->
+    {option}`hashedPassword` -> {option}`password`
```
Ma27, line 58: Isn't there an arrow missing? Same for the ordering below.

Ma27 (commented):
Code LGTM otherwise.
Please reword your commit messages to follow the contribution guidelines.

nixos/modules/config/users-groups.nix
fidgetingbits: Thanks for catching that!

nixos/modules/config/users-groups.nix
```diff
@@ -53,37 +53,80 @@ let
...
+    {option}`initialPassword` -> {option}`password`
+    {option}`hashedPasswordFile`
+  '';
+
+  overrideOrderText = isMutable: ''
+    If the option {option}`users.mutableUsers` is
+    `${builtins.toString isMutable}`, then the order of precedence is as shown
+    below, where values on the left are overridden by values on the right:
+
+    ${if isMutable then overrideOrderMutable else overrideOrderImmutable}
```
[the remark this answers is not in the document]
fidgetingbits: I've rebased with new commits following the contribution guideline and at the same time I _think_ I addressed this suggestion, but please take a look at confirm it's what you wanted. 

I will look into how to actually generate the new docs from nixpkgs, so I can confirm things are formatted sanely.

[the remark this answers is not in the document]
fidgetingbits: Hah.. there is indeed a rendering error I'll fix, but figured I'd note it down here anyway. Built with `nix-build nixos/release.nix -A manual.x86_64-linux` and checked the `options.html`:

> If the option users.mutableUsers is ``, then the order of precedence is as shown below, where values on the left are overridden by values on the right: initialHashedPassword -> hashedPassword -> initialPassword -> password -> hashedPasswordFile

> If the option users.mutableUsers is 1, then the order of precedence is as shown below, where values on the left are overridden by values on the right: initialHashedPassword -> initialPassword -> hashedPassword -> password -> hashedPasswordFile


Is because:
```
nix-repl> builtins.toString true
"1"

nix-repl> builtins.toString false
""
```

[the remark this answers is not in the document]
fidgetingbits: The rendering issue is fixed now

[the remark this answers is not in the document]
Ma27: > Is because:

Heh, didn't even notice it during review.

Fwiw, this is (IIRC) because Nix does `toString` on all derivation attributes before putting those into the env to make them accessible for the builder (almost always sh/bash) and it kinda makes sense to have an empty string for false in such a language. But yeah, it's a surprise the first time for everybody ;-)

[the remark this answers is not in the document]
fidgetingbits: Thanks for the insight! It seemed a bit cursed at first, but in the context of generating shellscript output it does indeed make a lot of sense.

Ma27 (approved):
Pushed two small fixes (one for the whitespace, one for an assertion in the sysusers test - details are in the messages). LGTM now.

fidgetingbits:
Thanks a lot for all your help working through this!

PR #310567, opened 2024-05-10
undocker: 1.0.4 -> 1.2.2

motiejus:
## Description of changes

* bump v1.0.4 -> v1.2.2
* change the origin website
* add myself to the maintainers (I am the upstream author).

Co-Authored-By: John Chadwick <johnwchadwick@gmail.com>

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin

### DIFF
--- a/pkgs/tools/misc/undocker/default.nix
+++ b/pkgs/tools/misc/undocker/default.nix
@@ -1,25 +1,36 @@
 { lib
 , buildGoModule
-, fetchFromSourcehut
+, fetchFromGitea
+, gnumake
 }:
-buildGoModule rec {
-  pname = "undocker";
-  version = "1.0.4";
-
-  src = fetchFromSourcehut {
-    owner = "~motiejus";
-    repo = pname;
+let
+  version = "1.2.2";
+  hash = "sha256-kBqNopcHpldU5oD6zoVjPjP84t12QFcbWBSNNgwImKg=";
+  src = fetchFromGitea {
+    domain = "git.jakstys.lt";
+    owner = "motiejus";
+    repo = "undocker";
     rev = "v${version}";
-    hash = "sha256-I+pTbr1lKELyYlyHrx2gB+aeZ3/PmcePQfXu1ckhKAk=";
+    hash = hash;
   };
+in
+buildGoModule {
+  pname = "undocker";
+  inherit version src;
+
+  nativeBuildInputs = [ gnumake ];
+
+  buildPhase = "make VSN=v${version} VSNHASH=${hash} undocker";
+
+  installPhase = "install -D undocker $out/bin/undocker";
 
   vendorHash = null;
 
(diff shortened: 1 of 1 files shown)


jchv (approved):
LGTM.
- Just as a sanity check, I verified that the repo move seems legit. [The upstream](https://sr.ht/~motiejus/undocker/) links to the new location, and the PR author is the maintainer. So that looks good.
- Tested under x86_64-linux by using `podman image save > redis-all.tar docker.io/redis` followed by `nix run .#undocker -- ./redis-all.tar ./redis.tar`. The resulting tarball looks good.
- The diff itself LGTM from a Nix standpoint.

motiejus:
> I think either of these approaches would be fine. It's not strictly necessary IMO, though.

Thank you, this is awesome. I copy-pasted your derivation and just put it here with a slight adjustment -- I added an explicit `make undocker` Makefile target.

I remain curious on why it needs `VSN=<...>` and why it can't pick it up from git, just like it does with sha1. But that's for another time.

jchv (approved):
LGTM! Now with correct version information :)

motiejus:
I'll leave it at that. Thanks!

pkgs/tools/misc/undocker/default.nix
```diff
@@ -1,25 +1,37 @@
...
+    leaveDotGit = true;
   };
+in
+buildGoModule {
+  pname = "undocker";
+  inherit version src;
+
+  nativeBuildInputs = [ git gnumake ];
+
+  buildPhase = "make VSN=v${version} undocker";
```
Aleksanaa, line 24: So we don't need `leaveDotGit` and `git` in `nativeBuildInputs` anymore?

Aleksanaa: Is it that important? Can we just provide a fake hash? `leaveDotGit` frequently has reproducibility issues.

motiejus: Updated.

motiejus:
I removed all git references and will use the sha256 in place of `VSNHASH`.

Aleksanaa:
What does that VSNHASH mean? Git hash? This sha256 is a hash in SRI format, it may not be suitable here.

motiejus:
I added VSNHASH as a debugging aide to understand where the binary is built from, it is visible in `undocker --version`. It defaults to git's sha1 hash, but, when we don't have it, we use what we have.

Aleksanaa:
You're more likely to have problems by stuffing a value in there that clearly doesn't belong there. We already have a VSN, so let’s put a hash there that is obviously wrong, for example, all 0s.

jordanisaacs (approved):

PR #310616, opened 2024-05-10
nixos/phpfpm: quote string config values

### LINKED ISSUE #79469: `services.phpfpm.pools.<name>.phpEnv` values are not quoted
**Describe the bug**
`services.phpfpm.pools.<name>.phpEnv` values do not require quotation to be parsed correctly by php-fpm, but if the value contains a `=` or is `true`/`false` (and perhaps other values, then the file fails to parse).

**To Reproduce**
1. Use the following NixOS configuration:

```nix
services.phpfpm.pools.foo.phpEnv."bar" = "true";
```

2. Observe that `phpfpm-bar.service` fails to start.
3. Observe that manually quoting `"true"` fixes the issue.

**Expected behavior**
`phpfpm` module appropriately quotes values provided to `phpEnv`.

**Metadata**

 - system: `"x86_64-linux"`
 - host os: `Linux 4.19.88, NixOS, 19.09.1549.45ea6092203 (Loris)`
 - multi-user?: `yes`
 - sandbox: `yes`
 - version: `nix-env (Nix) 2.3`
 - channels(root): `"nixos-19.09.1549.45ea6092203"`
 - channels(david): `""`

Maintainer information:
```yaml
# a list of nixpkgs attributes affected by the problem
attribute:
# a list of nixos modules affected by the problem
module: services.phpfpm
```

arcnmx:
## Description of changes

Avoid treating [strings such as `"yes"` or `"false"` as keywords](https://bugs.php.net/bug.php?id=69851).

Fixes #79469

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [x] Tested, as applicable:

---

### DIFF
--- a/nixos/modules/services/web-servers/phpfpm/default.nix
+++ b/nixos/modules/services/web-servers/phpfpm/default.nix
@@ -10,6 +10,8 @@ let
   toStr = value:
     if true == value then "yes"
     else if false == value then "no"
+    # escape according to https://www.php.net/manual/en/function.parse-ini-file.php
+    else if isString value then ''"${escape [ "\"" "\\" "$" ] value}"''
     else toString value;
 
   fpmCfgFile = pool: poolOpts: pkgs.writeText "phpfpm-${pool}.conf" ''
--- a/nixos/tests/php/fpm.nix
+++ b/nixos/tests/php/fpm.nix
@@ -10,7 +10,15 @@ import ../make-test-python.nix ({ pkgs, lib, php, ... }: {
 
       virtualHosts."phpfpm" =
         let
-          testdir = pkgs.writeTextDir "web/index.php" "<?php phpinfo();";
+          testdir = pkgs.writeTextDir "web/index.php" ''
+            <?php
+              phpinfo();
+              $envNames = [ 'keyword', 'characters', 'quoted' ];
+              foreach ($envNames as $env) {
+                echo "<!-- $env=''${_SERVER[$env]} -->";
+              }
+            ?>
+          '';
         in
         {
           root = "${testdir}/web";
@@ -41,6 +49,12 @@ import ../make-test-python.nix ({ pkgs, lib, php, ... }: {
         "pm.min_spare_servers" = 1;
         "pm.start_servers" = 2;
       };
+      phpEnv = {
+        # check that keywords and special characters are escaped properly
+        keyword = "false";
+        characters = "^ \${hi} \\";
+        quoted = "\"'";
+      };
     };
   };
   testScript = { ... }: ''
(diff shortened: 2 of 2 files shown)


nixos/tests/php/fpm.nix
```diff
@@ -10,7 +10,15 @@ import ../make-test-python.nix ({ pkgs, lib, php, ... }: {
 
       virtualHosts."phpfpm" =
         let
-          testdir = pkgs.writeTextDir "web/index.php" "<?php phpinfo();";
+          testdir = pkgs.writeTextDir "web/index.php" ''
+            <?php
+              phpinfo();
+              $envNames = array('keyword', 'characters', 'quoted');
```
drupol, line 16:
```suggestion
              $envNames = [ 'keyword', 'characters', 'quoted' ];
```

nixos/tests/php/fpm.nix
```diff
@@ -10,7 +10,15 @@ import ../make-test-python.nix ({ pkgs, lib, php, ... }: {
...
+              foreach ($envNames as &$env) {
```
drupol, line 17:
```suggestion
              foreach ($envNames as $env) {
```

drupol (changes requested):

nixos/modules/services/web-servers/phpfpm/default.nix
```diff
@@ -10,6 +10,8 @@ let
   toStr = value:
     if true == value then "yes"
     else if false == value then "no"
+    # escape according to https://www.php.net/manual/en/function.parse-ini-file.php
+    else if isString value then ''"${escape [ "\"" "\\" "$" ] value}"''
```
jtojnar, line 14: Escaping `$` does not work, though – PHP will interpolate it even though you quoted it: https://github.com/NixOS/nixpkgs/issues/79469#issuecomment-631461513

jtojnar: The quotes on the other hand will get stuck when using your patch:

```
test1=
test2="$2y$10$test2"
test3="\$2y\$10\$test3"
test4='$2y$10$test4'
test5='\$2y\$10\$test5'
test6=$2y$10$test6
```

arcnmx: Very strange. Though I believe the escape does prevent `${x}` from being interpolated at least, which can appear anywhere in a string?

Not sure how to handle that case though, there are unfortunately some sequences that can't seem to be escaped... Another example that fails is newlines, despite the php.ini documentation implying that they're valid as-is inside quotes (see the "value concatenation" part for an example)...

diogotcorreia: I believe this is the code responsible for this behavior: https://github.com/php/php-src/blob/631c366f9f58c8ba4078a48d1f56187cfbf8e549/sapi/fpm/fpm/fpm_env.c#L171-L180

Seems like any value that starts with `$` will be replaced by the respective env variable (or empty string). There does not appear to be a way to escape this, so I'd say we can ignore this edge case.

It shows up properly with `php-fpm -y config.ini -tt`, so it's definitely a post-processing step.

diogotcorreia:
Is there any reason for this to not be merged other than the merge conflict?

I have opened a [similar PR](https://github.com/NixOS/nixpkgs/pull/516530) yesterday, but only now found this one.
Tbh I do prefer this PR's approach.

@arcnmx Are you still interested in this PR? Otherwise I can incorporate these changes into mine so we can finally get this fixed.

Also, I figured out we can escape new lines with the following:
```
env[test1] = "Lorem \"ipsum\"" PHP_EOL "test"
```

Unclear if there is a better way, I tried just putting a literal newline inside the quotes, but php-fpm fails to parse that.

PR #310631, opened 2024-05-10
nixos/libinput: Allow 'custom' acceleration profile

ghost:
## Description of changes

Expand the allowed enum values for the `accelProfile` option of `services/hardware/libinput ` to allow for `"custom"`, which is [another possible value](https://wayland.freedesktop.org/libinput/doc/latest/pointer-acceleration.html#pointer-acceleration-profiles) in the newer versions of the library. I have tested to make sure the option gets applied correctly.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [x] Tested, as applicable:

---

### DIFF
--- a/nixos/modules/services/hardware/libinput.nix
+++ b/nixos/modules/services/hardware/libinput.nix
@@ -18,18 +18,19 @@ let cfg = config.services.libinput;
       };
 
       accelProfile = mkOption {
-        type = types.enum [ "flat" "adaptive" ];
+        type = types.enum [ "flat" "adaptive" "custom" ];
         default = "adaptive";
         example = "flat";
         description = ''
             Sets the pointer acceleration profile to the given profile.
-            Permitted values are `adaptive`, `flat`.
+            Permitted values are `adaptive`, `flat`, `custom`.
             Not all devices support this option or all profiles.
             If a profile is unsupported, the default profile for this is used.
             `flat`: Pointer motion is accelerated by a constant
             (device-specific) factor, depending on the current speed.
             `adaptive`: Pointer acceleration depends on the input speed.
             This is the default profile for most devices.
+            `custom`: Allows the user to define a custom acceleration function.
           '';
       };
 
--- a/nixos/modules/services/hardware/libinput.nix
+++ b/nixos/modules/services/hardware/libinput.nix
@@ -47,40 +47,39 @@ let cfg = config.services.libinput;
       };
 
       accelPointsFallback = mkOption {
-        type = types.nullOr types.str;
+        type = types.nullOr (types.listOf types.number);
         default = null;
-        example = "0.0 1.0 2.4 2.5";
+        example = [ 0.0 1.0 2.4 2.5 ];
         description = ''
-            Sets the points of the fallback acceleration function. The string must be a space-separated
-            list of floating point non-negative numbers. This only applies to the custom profile.
+            Sets the points of the fallback acceleration function. The value must be a list of
+            floating point non-negative numbers. This only applies to the custom profile.
           '';
       };
 
       accelPointsMotion = mkOption {
-        type = types.nullOr types.str;
+        type = types.nullOr (types.listOf types.number);
         default = null;
-        example = "0.0 1.0 2.4 2.5";
+        example = [ 0.0 1.0 2.4 2.5 ];
         description = ''
-            Sets the points of the (pointer) motion acceleration function. The string must be a
-            space-separated list of floating point non-negative numbers. This only applies to
-            the custom profile.
-          '';
-      };
-
-      accelPointsScroll = mkOption {
-        type = types.nullOr types.str;
-        default = null;
-        example = "0.0 1.0 2.4 2.5";
-        description = ''
-            Sets the points of the scroll acceleration function. The string must be a space-separated
+            Sets the points of the (pointer) motion acceleration function. The value must be a
             list of floating point non-negative numbers. This only applies to the custom profile.
           '';
       };
 
-      accelStepFallback = mkOption {
-        type = types.nullOr types.str;
+      accelPointsScroll = mkOption {
+        type = types.nullOr (types.listOf types.number);
         default = null;
-        example = "0.1";
+        example = [ 0.0 1.0 2.4 2.5 ];
+        description = ''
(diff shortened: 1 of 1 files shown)


pluiedev (approved):
Looks good, but is there any way for the user to define what the acceleration function *actually* is in Nix?

VSinerva:
> Looks good, but is there any way for the user to define what the acceleration function _actually_ is in Nix?

This can be done with `services.xserver.libinput.(mouse/touchpad).additionalOptions` by using the relevan libinput options like `Option "AccelStepMotion"` and `Option "AccelPointsMotion"`. The problem was that the `AccelProfile` enum is forced to *always* have a value, which couldn't be overridden via `additionalOptions`. So you could define an acceleration function, but couldn't set libinput to use the `custom` profile.

pluiedev:
> > Looks good, but is there any way for the user to define what the acceleration function _actually_ is in Nix?
> 
> 
> 
> This can be done with `services.xserver.libinput.(mouse/touchpad).additionalOptions` by using the relevan libinput options like `Option "AccelStepMotion"` and `Option "AccelPointsMotion"`. The problem was that the `AccelProfile` enum is forced to *always* have a value, which couldn't be overridden via `additionalOptions`. So you could define an acceleration function, but couldn't set libinput to use the `custom` profile.

Hm, I see! I would then suggest to add those options (step motion and points motion) to the module too if feasible, and mention those options in the custom profile's description. It should also be possible then to set the profile to custom by default, if those options were set. 

Worst case, just mention those raw options and tell people to add them to their `additionalOptions` if they want the custom profile to do anything.

VSinerva:
> > > Looks good, but is there any way for the user to define what the acceleration function _actually_ is in Nix?
> > 
> > 
> > This can be done with `services.xserver.libinput.(mouse/touchpad).additionalOptions` by using the relevan libinput options like `Option "AccelStepMotion"` and `Option "AccelPointsMotion"`. The problem was that the `AccelProfile` enum is forced to _always_ have a value, which couldn't be overridden via `additionalOptions`. So you could define an acceleration function, but couldn't set libinput to use the `custom` profile.
> 
> Hm, I see! I would then suggest to add those options (step motion and points motion) to the module too if feasible, and mention those options in the custom profile's description. It should also be possible then to set the profile to custom by default, if those options were set.
> 
> Worst case, just mention those raw options and tell people to add them to their `additionalOptions` if they want the custom profile to do anything.

I added all the options relevant to custom accel functions. I held off on making `accelProfile` automatically change to "custom", because the only way I know of doing that would be to change the default value to some conditional mess with all six options being checcked for a non-null values

nixos/modules/services/hardware/libinput.nix
```diff
@@ -18,26 +18,96 @@ let cfg = config.services.libinput;
...
         example = "-0.5";
-        description = "Cursor acceleration (how fast speed increases from minSpeed to maxSpeed).";
+        description = ''
+            Cursor acceleration (how fast speed increases from minSpeed to maxSpeed).
+            This only applies to the flat or adaptive profile.
+          '';
+      };
+
+      accelPointsFallback = mkOption {
+        type = types.nullOr types.str;
```
pluiedev, line 50:
```suggestion
        type = types.nullOr (types.listOf types.number)
```
Ditto for below

nixos/modules/services/hardware/libinput.nix
```diff
@@ -18,26 +18,96 @@ let cfg = config.services.libinput;
...
+        default = null;
+        example = "0.0 1.0 2.4 2.5";
+        description = ''
+            Sets the points of the scroll acceleration function. The string must be a space-separated
+            list of floating point non-negative numbers. This only applies to the custom profile.
+          '';
+      };
+
+      accelStepFallback = mkOption {
+        type = types.nullOr types.str;
```
pluiedev, line 81:
```suggestion
        type = types.nullOr types.number;
```
Ditto for below

pluiedev (changes requested):
Mmmmmm... I personally tend to model this with a `types.attrTag` that allows options to be specified that are only relevant for the specific enum type, like this:
```nix
{
  custom.stepMotion = [ ... ];
}
# or
{
  adaptive.pointsMotion = [ ... ];
}
```

However I perfectly understand if this couldn't be done due to backwards compatibility reasons, or it's simply too complicated.

nixos/modules/services/hardware/libinput.nix
VSinerva: I can see the reason for changing the types, and had considered it. I prioritized consistency with the pre-existing options, as there are already very similar options in the same module that are typed as `str`. Would it be better to ignore them?

VSinerva:
> Mmmmmm... I personally tend to model this with a `types.attrTag` that allows options to be specified that are only relevant for the specific enum type, like this:
> 
> ```nix
> {
>   custom.stepMotion = [ ... ];
> }
> # or
> {
>   adaptive.pointsMotion = [ ... ];
> }
> ```
> 
> However I perfectly understand if this couldn't be done due to backwards compatibility reasons, or it's simply too complicated.

I think I understand the basic point you're making, but I have no idea how to implement it in the Nix language, and my Googling is coming up short :/

If this is something that's important (and you're able to help a newbie through it) I won't object, but otherwise I'd be inclined to leave it.

pluiedev:
TBH I wouldn't worry too much about it. It's going to be a pain to try to migrate all the old settings to the new format and it's really just not worth it — I would totally try to do it myself if it's a new module though

VSinerva:
Is there something more that needs to be done on my part to get this merged?

No rush ofc, just want to make sure I'm not unknowingly holding anything up :)

pluiedev (approved):
LGTM

VSinerva:
It's been a while and I noticed that this module is not covered by anyone via `CODEOWNERS`, and modules don't have a maintainers list built in, unlike packages. Best I can find is the `freedesktop` team.

@jtojnar Are you able to merge this PR? Sorry if this ping is erroneous! (First time contributing to NixOS)

SuperSandro2000:
Someone wants to take a look at #308777?

PR #310700, opened 2024-05-10
k3s: fix a blank space making update script not match nixfmt-rfc-styles

superherointj:
k3s: fix a blank space making update script not match nixfmt-rfc-styles

Caught issue after update script execution. Unfortunately, a bit late.

CC @wrmilling @yajo @euank @mic92


wrmilling (approved):
LGTM

PR #310708, opened 2024-05-11
nixos/input-method: deprecate .enabled option; add .type and .enable options

### LINKED ISSUE #180654: i18n.inputMethod.enabled has a confusing type
I was expecting the type to be a simple `true` or `false`. Would it make sense splitting this into  `i18n.inputMethod.enabled`, and `i18n.inputMethod.engine`?

https://github.com/NixOS/nixpkgs/blob/316b762afdb9e142a803f29c49a88b4a47db80ee/nixos/modules/i18n/input-method/default.nix#L31-L52

importantblimp:
## Description of changes
This commit introduces two new properties:
`enable` and `type`, to replace the `enabled` property.
`enable` has the same meaning as is common across nixpkgs.
`type` has the same meaning as the existing `enabled` property.
`enabled` property is now deprecated and will be removed in a future release.

Fixes #180654

I updated and ran the existing inputMethod tests:
`nix-build -A nixosTests.fcitx5 -A nixosTests.installed-tests.ibus`

Maintainers:
@ericsagnes

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [x] Tested, as applicable:

---

### DIFF
--- a/doc/packages/ibus.section.md
+++ b/doc/packages/ibus.section.md
@@ -11,7 +11,8 @@ On NixOS, you need to explicitly enable `ibus` with given engines before customi
 ```nix
 { pkgs, ... }: {
   i18n.inputMethod = {
-    enabled = "ibus";
+    enable = true;
+    type = "ibus";
     ibus.engines = with pkgs.ibus-engines; [ typing-booster ];
   };
 }
--- a/nixos/modules/i18n/input-method/default.md
+++ b/nixos/modules/i18n/input-method/default.md
@@ -25,7 +25,8 @@ The following snippet can be used to configure IBus:
 ```nix
 {
   i18n.inputMethod = {
-    enabled = "ibus";
+    enable = true;
+    type = "ibus";
     ibus.engines = with pkgs.ibus-engines; [ anthy hangul mozc ];
   };
 }
@@ -81,7 +82,8 @@ The following snippet can be used to configure Fcitx:
 ```nix
 {
   i18n.inputMethod = {
-    enabled = "fcitx5";
+    enable = true;
+    type = "fcitx5";
     fcitx5.addons = with pkgs; [ fcitx5-mozc fcitx5-hangul fcitx5-m17n ];
   };
 }
@@ -119,7 +121,8 @@ The following snippet can be used to configure Nabi:
 ```nix
 {
   i18n.inputMethod = {
-    enabled = "nabi";
+    enable = true;
+    type = "nabi";
   };
 }
 ```
@@ -134,7 +137,8 @@ The following snippet can be used to configure uim:
 ```nix
 {
   i18n.inputMethod = {
-    enabled = "uim";
+    enable = true;
+    type = "uim";
   };
 }
 ```
@@ -154,7 +158,8 @@ The following snippet can be used to configure Hime:
 ```nix
 {
   i18n.inputMethod = {
-    enabled = "hime";
+    enable = true;
+    type = "hime";
   };
 }
 ```
@@ -168,7 +173,8 @@ The following snippet can be used to configure Kime:
 ```nix
 {
   i18n.inputMethod = {
-    enabled = "kime";
+    enable = true;
+    type = "kime";
   };
 }
 ```
--- a/nixos/modules/i18n/input-method/default.nix
+++ b/nixos/modules/i18n/input-method/default.nix
@@ -4,6 +4,8 @@ with lib;
 let
   cfg = config.i18n.inputMethod;
 
+  allowedTypes = types.enum [ "ibus" "fcitx5" "nabi" "uim" "hime" "kime" ];
+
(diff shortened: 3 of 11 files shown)


importantblimp:
This is a breaking change for 24.05 (hence the documentation update).
I thought it simple enough to make in one go, but if this is the type of change that needs to be split up just let me know.

honnip:
I think you meant `enable` rather than `enabled` for a bool property

nixos/modules/i18n/input-method/default.nix
```diff
@@ -28,8 +30,10 @@ in
 {
   options.i18n = {
     inputMethod = {
-      enabled = mkOption {
-        type    = types.nullOr (types.enum [ "ibus" "fcitx5" "nabi" "uim" "hime" "kime" ]);
+      enabled = mkEnableOption "an additional input method type";
```
eclairevoyant, line 33:
```suggestion
      enable = mkEnableOption "an additional input method type";
```
for consistency with any other module.
Then we also add a "removed option" assertion

eclairevoyant:
Drafting for now as breaking changes are frozen anyway due to the release (see #303285)

nixos/modules/i18n/input-method/default.nix
importantblimp: I've made both of these changes, but they seem to be mutually exclusive, and the two tests are now failing.

Am I just doing it wrong? I'd appreciate your guidance.

Test with new style config (`nix-build -A nixosTests.installed-tests.ibus`), where the error is in the module definition itself:
```
error:
       Failed assertions:
       - The option definition `i18n.inputMethod.enabled' in `/home/gy/src/nixpkgs/nixos/modules/i18n/input-method/default.nix' no longer has any effect; please remove it.
       Use <literal>i18n.inputMethod.type and .enable = true; instead<literal>

```

Test with old style config (`nix-build -A nixosTests.fcitx5`):
```
error: The option `i18n.inputMethod.enabled' can no longer be used since it's been removed. Use <literal>i18n.inputMethod.type and .enable = true; instead<literal>
```

importantblimp:
> Drafting for now as breaking changes are frozen anyway due to the release (see #303285)

I don't feel the need to rush this in, but if it is ready before the 15th: with your suggestions the change is no longer breaking so could be included in 24.05?

nixos/modules/i18n/input-method/default.nix
eclairevoyant: You're right. We should not use mk(Removed|Renamed)OptionModule here.
Removing that fixes the tests.

What we could do (to ensure that users get a warning with `.enabled`) is to set `config.warnings` if `.enabled != null`.

eclairevoyant:
Yes, if we're not breaking anything, I believe it could even be backported per https://github.com/NixOS/nixpkgs/blob/master/CONTRIBUTING.md#changes-acceptable-for-releases:
> The release branches should generally only receive backwards-compatible changes, both for the Nix expressions and derivations.

doc/packages/ibus.section.md
```diff
@@ -11,7 +11,8 @@ On NixOS, you need to explicitly enable `ibus` with given engines before customi
 ```nix
 { pkgs, ... }: {
   i18n.inputMethod = {
-    enabled = "ibus";
```
importantblimp, line 14: I've intentionally left the other test (`nixos/tests/fcitx5/default.nix`, not present in this review) using `enabled` to have a test for the backwards compatibility of this change

nixos/modules/i18n/input-method/default.md
```diff
@@ -81,7 +82,8 @@ The following snippet can be used to configure Fcitx:
 ```nix
 {
   i18n.inputMethod = {
-    enabled = "fcitx5";
```
importantblimp, line 84: Although I've updated the documentation for all of the input methods

doc/packages/ibus.section.md
importantblimp: Whoops, this is the docs file. I meant to add this to nixos/tests/installed-tests/ibus.nix

importantblimp:
@eclairevoyant this is ready to go

importantblimp:
I've rebased on master, removed the 24.05 release note change, and updated the 24.11 release notes

eclairevoyant (approved):
Apologies that this fell off the radar; the change is good. Thank you for your work on this!

eclairevoyant:
@ofborg test installed-tests.ibus fcitx5

nakoo:
Is there any reason behind why keep using `.type` instead of `.package` option? Current implementation makes harder to change proper package version. I had to override it because of the reason.

eclairevoyant:
`.package` options can be added as well, feel free to send a PR.

Issue #290642, opened 2024-02-22
[Tracking issue] Silent retirement of maintainers


AndersonTorres:
There are many maintainers that didn't touch Nixpkgs in a long span of time.
They should be listed, formally retired and their packages put to adoption.

Edit: here are some I have found while reading the PR list:

https://github.com/NixOS/nixpkgs/issues/290642#issuecomment-2147303665

emilylange:
We have [RFC 55](https://github.com/NixOS/rfcs/blob/e4c27fdaba705f8b63caf1ace171b503ffe0b5ff/rfcs/0055-retired-committers.md) (and it's amendment [RFC 71](https://github.com/NixOS/rfcs/blob/e4c27fdaba705f8b63caf1ace171b503ffe0b5ff/rfcs/0071-retired-committers-amendment.md)) for committers already.

It's current tracking issue is #88867 and has been so since 2020.

Is this what you mean or are you specifically trying to target contributors which are not committers?

AndersonTorres:
Yes, independent of commit bit.

Sigmanificient:
Having a dedicated tag for package adoption would be very nice:
- https://github.com/NixOS/nixpkgs/issues/310364

AndersonTorres:
I am cogitating another idea: mark them as inactive at maintainer-list.

AndersonTorres:
Here is a more up-top-date list:

| GitHub user           | Fork                                        | Estimated most recent update                                                                              | Estimated most recent contribution (git log) |
|-----------------------|---------------------------------------------|-----------------------------------------------------------------------------------------------------------|---------------------------------------|
| yanateras             | inexistent                                  | inexistent                                                                                                |                                       |
| lovek323 | https://github.com/lovek323/nixpkgs | |
| max-niederman | https://github.com/max-niederman/nixpkgs | |
| cko     | https://github.com/cko/nixpkgs | |
| meisternu | https://github.com/meisternu/nixpkgs | |
| abbradar | https://github.com/abbradar/nixpkgs | |
| rardiol               | https://github.com/rardiol/nixpkgs          |                                                                                                           |                                       |
| vyp | https://github.com/vyp/nixpkgs ||
| gebner                | https://github.com/gebner/nixpkgs           |                                                                                                           |                                       |
[...]

Issue #88867, opened 2020-05-25
RFC 55 - Remove inactive Nixpkgs


Mic92:
As part of [RFC55](https://github.com/NixOS/rfcs/blob/master/rfcs/0055-retired-committers.md) we move inactive members of the "Nixpkgs Committers team" to the "Nixpkgs Committers Emeritus" (not created yet). The migration will happen on the 25th June.

- @aycanirican          [commits](https://github.com/NixOS/nixpkgs/commits?author=aycanirican)
- @bluescreen303        [commits](https://github.com/NixOS/nixpkgs/commits?author=bluescreen303)
- @dezgeg               [commits](https://github.com/NixOS/nixpkgs/commits?author=dezgeg)
- @GrahamcOfBorg        [commits](https://github.com/NixOS/nixpkgs/commits?author=GrahamcOfBorg)
- @kosmikus             [commits](https://github.com/NixOS/nixpkgs/commits?author=kosmikus)
- @lovek323             [commits](https://github.com/NixOS/nixpkgs/commits?author=lovek323)
- @matejc               [commits](https://github.com/NixOS/nixpkgs/commits?author=matejc)
- @nckx                 [commits](https://github.com/NixOS/nixpkgs/commits?author=nckx)
- @ocharles             [commits](https://github.com/NixOS/nixpkgs/commits?author=ocharles)
- @Phreedom             [commits](https://github.com/NixOS/nixpkgs/commits?author=Phreedom)
- @rushmorem            [commits](https://github.com/NixOS/nixpkgs/commits?author=rushmorem)
- @the-kenny            [commits](https://github.com/NixOS/nixpkgs/commits?author=the-kenny)
- @ts468                [commits](https://github.com/NixOS/nixpkgs/commits?author=ts468)
- @viric                [commits](https://github.com/NixOS/nixpkgs/commits?author=viric)
- @vrthra               [commits](https://github.com/NixOS/nixpkgs/commits?author=vrthra)
- @wkennington          [commits](https://github.com/NixOS/nixpkgs/commits?author=wkennington)

Mic92:
I could not find an email address for @wkennington and @ts468.

Mic92:
I will send out emails to each affected account later this day.

Mic92:
@grahamc Can we move @GrahamcOfBorg to https://github.com/orgs/NixOS/teams/nixpkgs-maintainers As far as I know members of this group can also do labels and request reviews.

[...]

PR #310759, opened 2024-05-11
maintainers-list: create a field `inactivityReason`

AndersonTorres:
Default null; should be a free-form string explaining why the maintainer
became inactive.

The most immediate use of this attribute is to mark maintainers that are
not contributing to Nixpkgs from a long amount of time. Nonetheless it can
be used to mark any *long term inactivity*, including but not limited to:

- Sabbatical leave
- Retirement
- Personal issues
- Force majeure

In principle, a third person can set this attribute. For the sake of a good
etiquette, in such case at least one contact attempt should be issued and
carried out before effectively committing it to the codebase.

This attribute can be employed to many useful activities, including but not
limited to:

- Treewide automation
  - Removal of inactive maintainers from packages
  - Orphaning alerts
- Filter automatic notifications

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/maintainers/maintainer-list.nix
+++ b/maintainers/maintainer-list.nix
@@ -14,6 +14,36 @@
      keys = [{
        fingerprint = "AAAA BBBB CCCC DDDD EEEE  FFFF 0000 1111 2222 3333";
      }];
+
+     # Default null; should be a free-form string explaining why the maintainer
+     # became inactive.
+     #
+     # The most immediate use of this attribute is to mark maintainers that are
+     # not contributing to Nixpkgs from a long amount of time.
+     # Nonetheless it can be used to mark any *long term inactivity*, including
+     # but not limited to:
+     #
+     # - Sabbatical leave
+     # - Retirement
+     # - Personal or force majeure isues
+     #
+     # In principle, a third person can set this attribute. For the sake of a
+     # good etiquette, in such case at least one contact attempt should be
+     # issued and carried out before effectively committing it to the codebase.
+     #
+     # This attribute can be employed to many useful activities, including but
+     # not limited to:
+     #
+     # - Treewide automation
+     #   - Removal of inactive maintainers from packages
+     #   - Orphaning alerts
+     # - Filter automatic notifications
+     #
+     inactivityReason = ''
+       Real life issues: I want to take the Hunter Exam this year.
+
+       Registered at 2024-06-19 by Theresia.
+     '';
    };
    ```
 
@@ -26,6 +56,7 @@
    - `github` is your GitHub handle (as it appears in the URL of your profile page, `https://github.com/<userhandle>`),
    - `githubId` is your GitHub user ID, which can be found at `https://api.github.com/users/<userhandle>`,
    - `keys` is a list of your PGP/GPG key fingerprints.
+   - `inactivityReason` is an optional string attribute, explained above.
 
    Specifying a GitHub account ensures that you automatically:
    - get invited to the @NixOS/nixpkgs-maintainers team ;
--- a/lib/tests/maintainer-module.nix
+++ b/lib/tests/maintainer-module.nix
@@ -28,5 +28,9 @@ in {
       });
       default = [];
     };
+    inactivityReason = lib.mkOption {
+      type = types.nullOr types.str;
+      default = null;
+    };
   };
 }


K900:
Why?

Aleksanaa:
In this way we can indeed trace other people's active status easily. But self mark as inactive and by others/bot have different meanings. If they are marked passively, they may not want to be disturbed, but this behavior may lead to more inactivity. It's good to use this to remind the maintainer or help remove inactive ones, but it also deserves more consideration.

AndersonTorres:
> Why?

Routinely I am bumping on code maintained by inactive people. https://github.com/NixOS/nixpkgs/issues/290642

Recently I am refactoring SDL and Guile libraries, and it is not hard to find people that are at least two years away from Nixpkgs.

It would be easier to have some help from Nixpkgs machinery to detect them.

----------

> If they are marked passively, they may not want to be disturbed, but this behavior may lead to more inactivity. 

The idea is to mark cases of long term inactivity. 
Also they are also free to return, just by updating the field.

> It's good to use this to remind the maintainer or help remove inactive ones, but it also deserves more consideration.

Agreed.

AndersonTorres:
Related issues:
- https://github.com/NixOS/nixpkgs/issues/88867
- https://github.com/NixOS/nixpkgs/issues/290642

AndersonTorres:
Hello, @abbradar!

Apologies for marking you without previous advice, but I have read what you wrote at the [RFC55 comment](https://github.com/NixOS/nixpkgs/issues/88867#issuecomment-2106134035).

What do you think about marking yourself as inactive for the time being?

SuperSandro2000:
> The idea is to mark cases of long term inactivity.
Also they are also free to return, just by updating the field.

They would still receive PR review requests and the like. What would more align with the current tooling and not requiring further changes, would be to just drop those entries and maybe together packages they maintain which are broken for a longer time.

AndersonTorres:
> They would still receive PR review requests and the like. 

the tooling can be updated to reflect this...

> just drop those entries

I am accumulating days cleaning up SDL-related packages is a state of abandonment...

pbsds:
Aleksana said
> But self mark as inactive and by others/bot have different meanings. 

Rather than a boolean field, what if we make it `nullOr str`, where the string states the reason? Any string indicates inactivity. The field would also then need to be renamed.

AndersonTorres:
`whyInactive`? I will test this.

AndersonTorres:
Done. What do you think,  @pbsds

pbsds (commented):
I like it! It could use an entry in the release notes, possibly some notes in the maintainer readme, and more opinions

maintainers/maintainer-list.nix
```diff
@@ -21049,6 +21087,11 @@
     github = "viric";
     githubId = 66664;
     name = "Lluís Batlle i Rossell";
+    inactivityReason = ''
+      No new activity since at least 2018-07-22.
+
+      Registered at 2024-06-19 by AndersonTorres.
```
SuperSandro2000, line 21093: Isn't the git history enough for this?

SuperSandro2000:
> The tooling can be updated to reflect this.

But until that is done, this doesn't change things and from a reviewer perspective, this is completely intransparent. It would be way easier to review things, if the packages didn't have a maintainer, then we also wouldn't start a discussion with every PR that picks up an abandon package.

maintainers/maintainer-list.nix
AndersonTorres: Git metadata is not directly accessible to Nix machinery.

AndersonTorres:
> > The tooling can be updated to reflect this.
> 
> But until that is done, 

This is not the hardest thing on this planet.

> this doesn't change things and from a reviewer perspective, this is completely intransparent.

Asking a dormant reviewer is not what I call transparency.

pbsds:
This pr seems to be stuck in a limbo

* The `[test]` commit makes this un-mergeable as-is
* My [request for some release note entry or documentation](https://github.com/NixOS/nixpkgs/pull/310759#pullrequestreview-2129153331) is unadressed
* We need more people to look at this and determine whether this is the direction we want to go or not since i don't consider myself influential enough to make that call. Should we advertise it on matrix?

AndersonTorres:
> This pr seems to be stuck in a limbo
> 
>     * The `[test]` commit makes this un-mergeable as-is

It was a test and a proof of concept, I did it to catch errors.
I can delete it right now if you like.

> 
>     * My [request for some release note entry or documentation](https://github.com/NixOS/nixpkgs/pull/310759#pullrequestreview-2129153331) is unadressed

Apologies, I did not notice this part.
Where should I  write such documentation?

> 
>     * We need more people to look at this and determine whether this is the direction we want to go or not since i don't consider myself influential enough to make that call. Should we advertise it on matrix?

Matrix and Discourse. I will open a post on Discourse.

maintainers/maintainer-list.nix
```diff
@@ -14,6 +14,37 @@
...
+     # became inactive.
+     #
+     # The most immediate use of this attribute is to mark maintainers that are
+     # not contributing to Nixpkgs from a long amount of time.
+     # Nonetheless it can be used to mark any *long term inactivity*, including
+     # but not limited to:
+     #
+     # - Sabbatical leave
+     # - Retirement
+     # - Intimate issues
```
rhendric, line 28: ... what?

vcunat: I guess they meant "Private issues" or whatever fallback when you don't feel like providing reasons?

rhendric, line 28:
```suggestion
     # - Personal issues
```

AndersonTorres: Bad translation.
The idea is "any personal issue I don't want to talk about". From rehab to mourning.

AndersonTorres: Done

willbush:
For someone to have write access to another repo in the NixOS org they have to be in the org. Adding oneself to the maintainer-list.nix triggers some automatic process that invites the person to the org. I only know this because it's a step I had to do when @infinisil gave me write access to nixpkgs-check-by-name. Any pruning of maintainer-list.nix should take this into account.

willbush:
Some other random thoughts / questions:

- Seems unlikely that anyone would have the foresight or time (when asked) to note a reason.
- Would a bot / GitHub action eventually stomp on the reason anyway?
- How long before the reason is stale?

AndersonTorres:
I will close this. It is not the suitable time to ponder about.

PR #310809, opened 2024-05-11
nwjs, nwjs084: added versioning for the nwjs package

ilya-epifanov:
## Description of changes

Added nwjs084 package, as later versions break some software, including:
```
betaflight-configurator
blheli-configurator
inav-blackbox-explorer
```

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/top-level/all-packages.nix
+++ b/pkgs/top-level/all-packages.nix
@@ -7414,10 +7414,25 @@ with pkgs;
     _7zz = _7zz-rar;
   };
 
-  nwjs = callPackage ../development/tools/nwjs { };
+  nwjs = callPackage ../development/tools/nwjs {
+    version = "0.100.1";
+    hashes = {
+      "sdk-ia32" = "sha256-8T3lyQi5tvXhThapng/l+wEbmXdUQBqPT7HHDBpEk5Y=";
+      "sdk-x64" = "sha256-nD1MYEm+4sq4MHkiOI46IgM6oDWpcyVWl3Yis0beEIg=";
+      "ia32" = "sha256-TfAw7+z/I3eYRn20DN0sR9moqD23UG7nvQbeXQDNDcg=";
+      "x64" = "sha256-EqLESL4zVdBzFvhPgXfcIwSVNNBVIMy1SdCu6eNU96Q=";
+    };
+  };
 
   nwjs-sdk = callPackage ../development/tools/nwjs {
     sdk = true;
+    version = "0.100.1";
+    hashes = {
+      "sdk-ia32" = "sha256-8T3lyQi5tvXhThapng/l+wEbmXdUQBqPT7HHDBpEk5Y=";
+      "sdk-x64" = "sha256-nD1MYEm+4sq4MHkiOI46IgM6oDWpcyVWl3Yis0beEIg=";
+      "ia32" = "sha256-TfAw7+z/I3eYRn20DN0sR9moqD23UG7nvQbeXQDNDcg=";
+      "x64" = "sha256-EqLESL4zVdBzFvhPgXfcIwSVNNBVIMy1SdCu6eNU96Q=";
+    };
   };
 
   obelisk = callPackage ../development/tools/ocaml/obelisk { menhir = ocamlPackages.menhir; };
--- a/pkgs/top-level/all-packages.nix
+++ b/pkgs/top-level/all-packages.nix
@@ -7435,6 +7435,16 @@ with pkgs;
     };
   };
 
+  nwjs084 = callPackage ../development/tools/nwjs {
+    version = "0.84.0";
+    hashes = {
+      "sdk-ia32" = "sha256-uy6WZuA5b79yACSe3wiKiEeMb6K/z84cSeQDrKFUUdE=";
+      "sdk-x64" = "sha256-xI/YMHg5RWYh9XCLskSkuDwemH77U43Fzb8C9+fS9wE=";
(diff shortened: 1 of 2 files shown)


ilya-epifanov:
required for https://github.com/NixOS/nixpkgs/pull/310811 https://github.com/NixOS/nixpkgs/pull/310813 https://github.com/NixOS/nixpkgs/pull/310813

pkgs/top-level/all-packages.nix
```diff
@@ -19221,7 +19221,25 @@ with pkgs;
 
   norminette = callPackage ../development/tools/norminette { };
 
-  nwjs = callPackage ../development/tools/nwjs { };
+  nwjs = callPackage ../development/tools/nwjs {
+    version = "0.85.0";
```
qubitnano, line 19225: Is there a reason nwjs is reverted to 0.85.0? Only betaflight-configurator has issues prompting the need for 0.84.0 correct?

qubitnano: Also all-packages may not be the best place for this, perhaps nwjs can be reworked to allow overrideAttrs?

ilya-epifanov: rolling back to 0.85 was just a mistake, I've fixed it now

ilya-epifanov: A number of packages were broken by an upgrade to 0.85:
```
blheli-configurator (new package, https://github.com/NixOS/nixpkgs/pull/310820)
betaflight-configurator
inav-blackbox-explorer (new package, https://github.com/NixOS/nixpkgs/pull/310813)
```
Probably `betaflight-blackbox-explorer` is affected, too (I haven't packaged it yet)

ilya-epifanov: regarding overrideAttrs, I don't see how it's possible to make it work without code duplication. We're interpolating the version in the url, so we'd have to do it again with a new value for version.
I made version and hashes the same kind of input as `sdk`. Also, python uses a similar approach to allow for multiple versions

qubitnano:
CC @mikaelfangel

MikaelFangel:
I think the general idea is fine, but it also leaves me with some questions. Wouldn't it be better if the package was built from scratch? Because nwjs uses chromium it's susceptible to some of the same vulnerabilities as chrome and by keeping the old binaries as an option we would have to determine which is vulnerable to what CVE and to my knowledge we wouldn't be able to patch those.

qubitnano:
nwjs doesn't appear to backport chromium fixes like electron, so we'd be on our own trying to maintain something as big as that. nwjs from source would be nice, but just looking at the build instructions these are big projects and who knows if hydra may be able to handle it.

Ideally upstream for these packages should be notified new versions of nwjs don't work anymore. It's probably safer to just mark these as broken or let users override an older version with an overlay at their own risk.

MikaelFangel:
Okay kinda had the feeling that it just wouldn't be that easy. 

But, then we agree that leaving this package as is and fixing/mark the packages that may be broken from upgrades of nwjs is the best solution?

ilya-epifanov:
@eclairevoyant what should we do with the `pkgs/by-name` check here?

qubitnano:
Another option could be to add 0.84.0 with it already marked as insecure
which requires the user to override to build. 0.84.0 itself doesn’t seem to
have any specific vulnerabilities except for using older chromium and
nodejs which aren’t getting patched by nwjs. But could these packages using
vulnerable nwjs be exploited and how so?

Just not sure nixpkgs policy regarding this.

ilya-epifanov:
Doesn't marking something as insecure require a CVE or eqv.? Also, the packages I'm fixing by the downgrade don't even access the internet (almost), don't open arbitrary files etc. So most probably not exploitable regardless of the vulnerability.

ilya-epifanov:
I think since nwjs doesn't maintain backwards compatibility we just have to provide multiple versions. It should be easier once they hit 1.0 (that is if they follow semantic versioning)

MikaelFangel:
> Doesn't marking something as insecure require a CVE or eqv.? Also, the packages I'm fixing by the downgrade don't even access the internet (almost), don't open arbitrary files etc. So most probably not exploitable regardless of the vulnerability.

You're correct but because nwjs includes both Chromium and NodeJS it could be susceptible to the same CVE's as both NodeJS and Chromium. Previously we have marked packages including either with their projects' CVE's.

ilya-epifanov:
@qubitnano @MikaelFangel guys, sorry for bumping the conversation but several packages are now broken and several others can't be merged because they're blocked on this change. I'm happy to refactor this but I'd rather do something right now to fix the outstanding issues

ilya-epifanov:
@qubitnano it's not currently compatible with `overrideAttrs`. The `pkgs/by-name` check is optional. I suggest merging a fix so that other packages can benefit from it.

ilya-epifanov:
Let's just roll back nwjs to 0.84 then. A number of packages are broken by a system-wide upgrade.

MikaelFangel:
I have created a pr reverting the breaking change against 23.11.

wucke13:
Do I correctly understand: To get inav-configurator (especially older versions) up and running, we need a (potentially too) old nwjs version? I would like to see the work ([5 open PRs by now and since for quite some time](https://github.com/NixOS/nixpkgs/pulls?q=is%3Apr+is%3Aopen+inav-configurator)) from @ilya-epifanov bear some fruit. Is there any way forward?

PR #310857, opened 2024-05-11
nixos/firewall: fix reverse path check failures with IPsec 

rnhmjoj:
## Description of changes

The endpoint of an IPsec tunnel receives encrypted IPsec packets that
are first decrypted and then forwarded to the intended destination.
The decrypted traffic appears to originate from the same interface it
came in from, so in most cases these packets will fail the reverse path
check even if legitimate.

This change adds an exception to not reject packets that were previously
IPsec-encrypted, meaning the have been accepted, decrypted and are in
the process of being forwarded to their final destinal.

## Things done

- Tested via 
  - [x] `nixosTests.firewall`
  - [x] `nixosTests.firewall-nftables`
  - [x] `nixosTests.libreswan-nat` see PR #310404

### DIFF
--- a/nixos/modules/services/networking/firewall-nftables.nix
+++ b/nixos/modules/services/networking/firewall-nftables.nix
@@ -82,6 +82,11 @@ in
       }
     ];
 
+    networking.nftables.preCheckRuleset = ''
+      # can't validate IPsec rules
+      sed '/meta ipsec/d' -i ruleset.conf
+    '';
+
     networking.nftables.tables."nixos-fw".family = "inet";
     networking.nftables.tables."nixos-fw".content = ''
         ${optionalString (cfg.checkReversePath != false) ''
@@ -89,6 +94,7 @@ in
             type filter hook prerouting priority mangle + 10; policy drop;
 
             meta nfproto ipv4 udp sport . udp dport { 67 . 68, 68 . 67 } accept comment "DHCPv4 client/server"
+            meta ipsec exists accept comment "decrypted packets from an IPsec VPN"
             fib saddr . mark ${optionalString (cfg.checkReversePath != "loose") ". iif"} oif exists accept
 
             jump rpfilter-allow
--- a/nixos/modules/services/networking/firewall-iptables.nix
+++ b/nixos/modules/services/networking/firewall-iptables.nix
@@ -123,6 +123,9 @@ let
       # Allows this host to act as a DHCP4 client without first having to use APIPA
       iptables -t mangle -A nixos-fw-rpfilter -p udp --sport 67 --dport 68 -j RETURN
 
+      # Allows decrypted packets from an IPsec VPN
+      ip46tables -t mangle -A nixos-fw-rpfilter -m policy --dir in --pol ipsec -j RETURN
+
       # Allows this host to act as a DHCPv4 server
       iptables -t mangle -A nixos-fw-rpfilter -s 0.0.0.0 -d 255.255.255.255 -p udp --sport 68 --dport 67 -j RETURN
 
--- a/nixos/tests/firewall.nix
+++ b/nixos/tests/firewall.nix
@@ -36,7 +36,7 @@ import ./make-test-python.nix ( { pkgs, nftables, ... } : {
     };
 
   testScript = { nodes, ... }: let
-    newSystem = nodes.walled2.config.system.build.toplevel;
+    newSystem = nodes.walled2.system.build.toplevel;
     unit = if nftables then "nftables" else "firewall";
   in ''
     start_all()


rnhmjoj:
@grahamcOfBorg test firewall firewall-nftables

rnhmjoj:
I added some sources, see the message in https://github.com/NixOS/nixpkgs/pull/310857/commits/3c12ef3f219c1a0f458d72e7b460782287974bbd.

nixos/modules/services/networking/firewall-nftables.nix
```diff
@@ -82,13 +82,19 @@ in
       }
     ];
 
+    networking.nftables.preCheckRuleset = ''
+      # can't validate IPsec rules
+      sed '/meta ipsec/d' -i ruleset.conf
```
mkg20001, line 87: if it's possible to make this work similarly with some workarrounds like the current nftables validation that would be preferred

rnhmjoj: I don't think so, It depends on which kernel modules the machine building the NixOS system has loaded.

rnhmjoj:
ping

mkg20001 (approved):

mkg20001:
feel free to ping me eariler, you shouldn't have to wait 3 months :sweat_smile:

nixos/modules/services/networking/firewall-nftables.nix
nakato: @rnhmjoj What kernel module does this rely on?

```
shd5j444d36l1sc4wziflq6m2aj8y0k3-nftables-rules[778]: /nix/store/shd5j444d36l1sc4wziflq6m2aj8y0k3-nftables-rules:11:3-12: Error: Could not process rule: Operation not supported
shd5j444d36l1sc4wziflq6m2aj8y0k3-nftables-rules[778]:   meta ipsec exists accept comment "decrypted packets from an IPsec VPN"
shd5j444d36l1sc4wziflq6m2aj8y0k3-nftables-rules[778]:   ^^^^^^^^^^
```

rnhmjoj: I don't remember, but the rule have been excluded from the check. I'm not sure how that error is possible.

nakato: Ah, it's conditional in a module, not a module to itself.
*CONFIG_XFRM* must be set, which gets selected by *CONFIG_NFT_XFRM*, which is hidden behind *CONFIG_XFRM_USER*.  Clear as mud.

https://elixir.bootlin.com/linux/v6.10.7/source/net/netfilter/nft_meta.c#L515

> I don't remember, but the rule have been excluded from the check. I'm not sure how that error is possible.

Error was not in the check, it was on actual rule load.

rnhmjoj: Uhm, I checked that the rule worked on a standard NixOS kernel. Do you have a custom kernel?

nakato: Yes, custom kernel.

The all-module kernel takes 16 hours on the most performant aarch64 hardware I own and much, much longer if anyone attempts it on the target hardware, so I only enable what I reasonably expect will be needed for its use case, and ipsec was not on that list, though it is a router, so maybe it should be; though that wouldn't be the case for other hardware like smaller SBCs.

I am slightly curious, why add this by default instead of having the ipsec configuration add it to *extraReversePathFilterRules* when ipsec is in use?

rnhmjoj: I think the rule should be there by default because the reverse path check is on by default and it's not smart enough to figure IPSec should not be messed with. As you say, we could add it only if (libreswan, strongswan, etc.) are enabled, but there may be other IPSec software not managed by a NixOS module. 

This is important enough because figuring out your tunnel doesn't work because of the reverse path check is far from trivial. Sadly nftables is designed badly enough that it can't parse rules for which the plugin is not loaded...

nakato: > but there may be other IPSec software not managed by a NixOS module.

They should get a nixos module.  They have to run as root or as a service anyways, so they shouldn't be unmanaged.

The extra configuration for ipsec to work in nftables that is extremely opaque shouldn't be required for the system to be functional.  The reverse path filtering doesn't require anything beyond very basic nftables support.

> Sadly nftables is designed badly enough that it can't parse rules for which the plugin is not loaded...

It's actually the kernel throwing an error back to userspace, and that's absolutely not bad design.  It has no idea if a rule being missing will make the setup insecure so it errors out.  To treat that failure as a fail open or fail closed is then up to the systems developer.  In the case of NixOS, it fails open.

rnhmjoj: > The extra configuration for ipsec to work in nftables that is extremely opaque shouldn't be required for the system to be functional. 

What's required for a system to be functional is pretty arbitrary: for example, by default NixOS installs a bunch of fonts, icons and even sound clips. I don't care about any of that, but desktop users do.

Since IPsec support is enabled by default in the kernel, I think it's reasonable that the default firewall rules shouldn't be breaking it in subtle ways. Also, this rule is not "extremely opaque", I'm pretty sure other distributions have similar rules.

> It's actually the kernel throwing an error back to userspace, and that's absolutely not bad design. It has no idea if a rule being missing will make the setup insecure so it errors out.

IMHO It's a bad design becase to check the configuration you have to talk to the kernel as root and validity depends on the kernel configuration. It could have been avoided by defining all syntax directly in the userspace nftables library. If a kernel module is not available the failure should be upon loading the rules.

mkg20001: Is it possible to make this dependent on the kernel config for ipsec being enabled? Is there a way to access it from a module?

rnhmjoj: @mkg20001 It's a good idea, but IIRC inspecting the kernel config at eval time is not possible due to IFD.

nakato: > Also, this rule is not "extremely opaque", I'm pretty sure other distributions have similar rules.

What is required to be enabled in the kernel is extremely opaque.  I never called the rule opaque.

> Since IPsec support is enabled by default in the kernel

It's not.  It's enabled by the default kernel building with support for everything through `autoModules`.

> I think it's reasonable that the default firewall rules shouldn't be breaking it in subtle ways.

Custom kernels should work with base configuration without needing to enable some undocumented config options.  There is no discoverability of why the ipsec rule fails.  No help-text in kconfig, grepping for ipsec across the whole kernel source tree turns up nothing.

And when nftables fails to start, it breaks by subtly deciding to forward all traffic instead of dropping all but the allowed traffic as configured, or dropping everything, because it fails open.

> I think it's reasonable that the default firewall rules shouldn't be breaking it in subtle ways.

Not having ipsec enabled should also not break in a way where you have to roll your own firewall, there is no way to disable this one rule, and it is included here for something else to work that is configured elsewhere.

None of the default configs in `arch/arm64/configs` have XFRM enabled.  Only 4 of the 75 in arm, and those are device specific ones not default.  This means it's very difficult to get this right when configuring the kernel as the NFT toggle is missing.

> If a kernel module is not available the failure should be upon loading the rules.

The kernel module is available, it doesn't have optional features compiled in.

---

I really do not think this should be here by default.  If it does stay though, then the required kernel config need to be explicitly enabled by including it in `pkgs/os-specific/linux/kernel/common-config.nix`.

nakato: Actually, with the implication of the failed load in mind, should this be considered a security issue?

1. If the kernel is built with `autoModules = false`, the required configuration to load the firewall will not exist unless the user has explicitly searched for and enabled XFRM kernel configuration.
  * Previous behaviour, due to common-config, is that the firewall works due to the requirements being pulled in through `common-config.nix`
  * https://github.com/NixOS/nixpkgs/blob/344e3016dc155ac5ca981d628948ef994bec9921/pkgs/os-specific/linux/kernel/common-config.nix#L289-L301

3. Configuration switches
   - Upon a live configuration switch, the new rules fail to load and the old rules stay in place.  Unless you catch the nftables.service error you won't know of the failure.  Upon reboot there will be no firewall, services will be accessible to the internet, if forwarding is enabled all traffic will pass.
   - Configuration switch with "nixos-rebuild boot", there will be no warning, the machine will boot, and there will be no firewall, services will be accessible to the internet, if forwarding is enabled all traffic will pass.


Explicit ping @mkg20001 due to the question at hand here around this being a potential security regression.

I'm not sure how to use the platforms listed in https://github.com/NixOS/nixpkgs/blob/344e3016dc155ac5ca981d628948ef994bec9921/lib/systems/platforms.nix , but as far as I can tell, nftables will be broken on the following now:
* pc_simplekernel
* pogoplug4
* sheevaplug
* zero-gravitas
* zero-sugar
* utilite
* guruplug
* beaglebone
* fuloong2f_n32

I've also found other repositories doing kernel builds with `autoModules = false` via github search.

https://github.com/search?q=autoModules+%3D+false++language%3ANix+&type=code

For example, this repository is active with the last commit being 5 days ago, https://github.com/anduril/jetpack-nixos/blob/17635c8913c2b9e61c0ec384cff5de389bbac4af/kernel/default.nix#L47, has actively enabled nftables modules and does not have XFRM modules enabled.

There's a couple in nixos-hardware too:
https://github.com/search?q=repo%3ANixOS%2Fnixos-hardware%20autoModules&type=code

rnhmjoj: > It's not. It's enabled by the default kernel building with support for everything through autoModules.

Does that matter? As an end-user, If download an ISO from nixos.org and install it I get a kernel that support IPSec. I would also not expect the firewall to break it.

> I really do not think this should be here by default. 

I understand and agree that this is adding a dependency on something that *shouldn't* be required, but unfortunately it is required to not make IPSec fail the reverse path check (which should never happen).

> If it does stay though, then the required kernel config need to be explicitly enabled by including it in pkgs/os-specific/linux/kernel/common-config.nix.

Ok, I didn't know it wasn't explicitely enabled. I guess it should be, given there are at least two IPSec modules in NixOS that depend on it.

Anyway, I don't maintain either NixOS firewall or kernels and I don't mean to impose myself.
So, feel free to revert this commit if you still disagree.

nakato: > Anyway, I don't maintain either NixOS firewall or kernels and I don't mean to impose myself.
> So, feel free to revert this commit if you still disagree.

If we can't work towards a solution that doesn't break non autoModules using systems, okay, PR #339393

PR #310880, opened 2024-05-11
nixos/xss-lock: add warning for startx

presto8:
## Description of changes

The xss-lock service only works if a display manager is configured in nixos. If the window manager is started manually using "startx", the xss-lock service will still run but it won't lock the screen.

This changeset adds a warning message if the configuration has startx enabled and xss-lock enabled.

This was a difficult issue for me to debug, a warning like this would have saved a lot of time :-)

## Things done

- Built on platform(s)
  - [X] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/nixos/modules/programs/xss-lock.nix
+++ b/nixos/modules/programs/xss-lock.nix
@@ -42,5 +42,10 @@ in
         ]);
       serviceConfig.Restart = "always";
     };
+
+    warnings = lib.mkIf (config.services.xserver.displayManager.startx.enable) [
+      "xss-lock service only works if a displayManager is set; it doesn't work when services.xserver.displayManager.startx.enable = true"
+    ];
+
   };
 }


pluiedev (approved):

turboMaCk (approved):
makes sense

rnhmjoj:
Ehm... no? I have been using xss-lock with startx for years.
@presto8 what was your issue?

rnhmjoj:
This should be the relevant portion of my configuration, if you need it:
<details>

```nix
{ config, lib, pkgs, ... }:

let
  # X session scripts
  session = with builtins;
    head (filter (x: x.name == "bspwm")
          config.services.xserver.windowManager.session);

  xinitrc = pkgs.writers.writeDashBin "xinitrc" ''
    # load resources
    xrdb -load "$XDG_CONFIG_HOME/x/resources"

    # start user services
    systemctl --user import-environment DISPLAY SSH_AUTH_SOCK XDG_SESSION_ID
    systemctl --user start nixos-fake-graphical-session.target

    # start programs
    dbus-daemon --session --address="unix:path=/run/user/$(id -u)/bus" &
    ${session.start}
    wait $waitPID

    # stop programs and services
    systemctl --user stop nixos-fake-graphical-session.target
    kill 0
  '';

  custom-startx = pkgs.writers.writeDashBin "startx" ''
    mkdir -p "$XDG_CACHE_HOME/x"
    export XINITRC=${xinitrc}/bin/xinitrc
    export XSERVERRC=/etc/X11/xinit/xserverrc
    exec ${pkgs.xorg.xinit}/bin/xinit "$@" -- $XSERVERRC >/dev/null 2>&1
  '';

in

{
  environment.systemPackages = lib.singleton (lib.hiPrio custom-startx);

  services.xserver =
    { enable = true;
      windowManager.bspwm.enable = true;
      displayManager.startx.enable = true;
      displayManager.xserverArgs = lib.mkAfter [ "vt$XDG_VTNR" ];
    };

  security.polkit.enable = true;

  programs.xss-lock =
    { enable = true;
      lockerCommand = lib.getExe pkgs.i3lock;
      extraOptions = [ "--transfer-sleep-lock" ];
    };
}
```

</details>

presto8:
@rnhmjoj, thanks for sharing your configuration! I tried to recreate my previous .xinitrc/startx environment but I long since moved to using a display manager and it's too much effort to go back.

I take your word for it that it works. In light of this, my earlier commit was a mistake. What do you recommend, should we just revert it, or is there some way to warn if those needed commands are not done (I'm guessing probably not). Probably just removing the warning is the best course of action...

rnhmjoj:
The proper solution would probably be replacing the startx script, which is recommended anyway, with one we generate from NixOS that include all the proper bits (like updating the systemd environment, etc.). I don't have time to work on such a big change, though.

For now I think we can drop the warning and maybe replace it with a runtime error if DISPLAY is missing before starting the service. If that's indeed what happened...

Was xss-lock the only service you couldn't get to work? I think that without importing those variables all user services that need X11 will not start.

turboMaCk:
Is this discussion somehow relevant to the PR?

I'm using this configuration https://github.com/turboMaCk/Dotfiles/blob/cd71147b3a1d184b36450abc53de5cc8f3b2550b/nixos/machines/thinkpad.nix#L168-L174 which does work. Of course there likely are many other ways in which this configuration option could be broken. There are many more configurations of bits which case stuff not to work then there are configurations in which it does work.

rnhmjoj:
I'm not sure I understood what you're asking. You seem to be using sddm: that a display manager works is not being questioned.

turboMaCk:
Re-read the thread to get more context. Yes I do use display manager. My understanding is that this at least at the moment does require manger. That's why warning was added in the PR.

I'm maybe confused if there you believe there is something wrong with the PR or suggesting we improve situation top remove current limitation / issue? I'm just asking because since I did the review and merge I feel some responsibility for any issue it could cause.

If you're just investigating what is going on and if there is a way we can improve this then I agree it would be great to fix this.

presto8:
The warning here doesn't cause any blocking issues, but would be an annoyance for somebody who has it working. As the creator of this PR, I definitely feel bad about that (sorry!).

I didn't reproduce it, but I take @rnhmjoj's word that it does work if doing the necessary import of the DISPLAY and other environment variables. The issue is, I can see a lot of people don't do that. The only place this is documented is on the wiki I think(?). I didn't come across it when trying to get it working.

Maybe the startx approach is for really advanced users anyway, and they should expect to have to figure out things like this on their own. I think detecting it at runtime is probably the only way that would work, not sure how to do that exactly.

turboMaCk:
>  The issue is, I can see a lot of people don't do that.

I share this concern. I think ideally we would find some way we can make setup without DM more user friendly and remove the warning.

> Maybe the startx approach is for really advanced users anyway, and they should expect to have to figure out things like this on their own.

I see what you mean. But I think the promise of nixos should be to deal with these details in the abstraction. I think we should aim to setting users for success. The warning, as I see it, can actually even help people to get to the solution and it doesn't prevent them from finding it. So I personally feel warning might be appropriate unless we find a way to improve the situation.

>  I think detecting it at runtime is probably the only way that would work, not sure how to do that exactly.

I also don't have any idea but frankly I did not have proper time to look into this. But I have a feeling maybe @rnhmjoj might already have some informed opinion?

presto8:
I can take a try at updating xss-lock service to emit a warning if DISPLAY is not set. I'll open a new issue to track this perhaps rather than keep talking on this already-merged PR?

presto8:
Please see #319435 and maybe we can continue the conversation over there?

PR #310978, opened 2024-05-12
nixos/gnome-keyring: add SSH support by exporting SSH_AUTH_SOCK

no-mood:
## Description of changes
Added SSH support inside the gnome-keyring module, similar to [ssh.nix](https://github.com/NixOS/nixpkgs/blob/27c13997bf450a01219899f5a83bd6ffbfc70d3c/nixos/modules/programs/ssh.nix#L347) and [gnupg.nix](https://github.com/NixOS/nixpkgs/blob/71bae31b7dbc335528ca7e96f479ec93462323ff/nixos/modules/programs/gnupg.nix#L232) by exporting `SSH_AUTH_SOCK`.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/nixos/modules/services/desktops/gnome/gnome-keyring.nix
+++ b/nixos/modules/services/desktops/gnome/gnome-keyring.nix
@@ -26,14 +26,7 @@ in
         '';
       };
 
-      enableSSHSupport = lib.mkOption {
-        type = lib.types.bool;
-        default = false;
-        description = ''
-          Enable SSH agent support in Gnome Keyring by setting SSH_AUTH_SOCK
-          environment variable correctly.
-        '';
-      };
+      enableSSHSupport = lib.mkEnableOption "SSH agent support for GNOME Keyring by setting the SSH_AUTH_SOCK environment variable";
 
     };
 
--- a/nixos/modules/services/desktops/gnome/gnome-keyring.nix
+++ b/nixos/modules/services/desktops/gnome/gnome-keyring.nix
@@ -1,7 +1,9 @@
 # GNOME Keyring daemon.
 
 { config, pkgs, lib, ... }:
-
+let
+  cfg = config.services.gnome.gnome-keyring;
+in
 {
 
   meta = {
@@ -24,6 +26,15 @@
         '';
       };
 
+      enableSSHSupport = lib.mkOption {
+        type = lib.types.bool;
+        default = false;
+        description = ''
+          Enable SSH agent support in Gnome Keyring by setting SSH_AUTH_SOCK
+          environment variable correctly.
+        '';
+      };
+
     };
 
   };
@@ -31,7 +42,7 @@
 
   ###### implementation
 
-  config = lib.mkIf config.services.gnome.gnome-keyring.enable {
+  config = lib.mkIf cfg.enable {
 
     environment.systemPackages = [ pkgs.gnome.gnome-keyring ];
 
@@ -47,6 +58,12 @@
       capabilities = "cap_ipc_lock=ep";
       source = "${pkgs.gnome.gnome-keyring}/bin/gnome-keyring-daemon";
     };
+
+    environment.extraInit = lib.mkIf cfg.enableSSHSupport ''
+      if [ -z "$SSH_AUTH_SOCK" -a -n "$XDG_RUNTIME_DIR" ]; then
+        export SSH_AUTH_SOCK="$XDG_RUNTIME_DIR/keyring/ssh"
+      fi
+    '';
 
   };
 
--- a/nixos/modules/services/desktops/gnome/gnome-keyring.nix
+++ b/nixos/modules/services/desktops/gnome/gnome-keyring.nix
@@ -26,7 +26,7 @@ in
         '';
       };
 
-      enableSSHSupport = lib.mkEnableOption "SSH agent support for GNOME Keyring by setting the SSH_AUTH_SOCK environment variable";
+      SSHSupport.enable = lib.mkEnableOption "SSH agent support for GNOME Keyring by setting the SSH_AUTH_SOCK environment variable";
 
     };
 
--- a/nixos/modules/services/desktops/gnome/gnome-keyring.nix
+++ b/nixos/modules/services/desktops/gnome/gnome-keyring.nix
@@ -52,11 +52,7 @@ in
       source = "${pkgs.gnome.gnome-keyring}/bin/gnome-keyring-daemon";
     };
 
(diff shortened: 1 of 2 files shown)


nixos/modules/services/desktops/gnome/gnome-keyring.nix
```diff
@@ -48,6 +59,12 @@
       source = "${pkgs.gnome.gnome-keyring}/bin/gnome-keyring-daemon";
     };
 
+    environment.extraInit = lib.mkIf cfg.enableSSHSupport ''
```
pluiedev, line 62: Hmm. Is there no easy way to express this with `environment.sessionVariables` or `environment.variables`?

nixos/modules/services/desktops/gnome/gnome-keyring.nix
```diff
@@ -24,14 +26,23 @@
         '';
       };
 
+      enableSSHSupport = lib.mkOption {
```
pluiedev, line 29: This can be a `mkEnableOption`. 
```nix
enableSSHSupport = lib.mkEnableOption "SSH agent support for GNOME Keyring by setting the SSH_AUTH_SOCK environment variable";
```

pluiedev (changes requested):

pluiedev (changes requested):
Also, please make sure the title of both the PR and the commit fit existing conventions. They should look something more like `nixos/gnome-keyring: add SSH support by exporting SSH_AUTH_SOCK`

nixos/modules/services/desktops/gnome/gnome-keyring.nix
no-mood: At first I was trying to use something like this:
```
environment.sessionVariables = {
  SSH_AUTH_SOCK = "${builtins.getEnv "XDG_RUNTIME_DIR"}/keyring/ssh";
};
```
Then I relied on what is done with SSH and GnuPG, I thought it was a better practice.

no-mood:
**Update**: I hadn't realized that in my configuration, the gnome-keyring-daemon is manually launched by my window manager (Hyprland):
```
exec-once = [ "${gnome.gnome-keyring}/bin/gnome-keyring-daemon --start --components=secrets,ssh" ];
```
Optionals components should be then passed to the wrapper?

There's a [home-manager gnome-keyring module](https://github.com/nix-community/home-manager/blob/44677a1c96810a8e8c4ffaeaad10c842402647c1/modules/services/gnome-keyring.nix#L17) that does this in a similar way, but we're still missing a nixos module

maanu1234 (commented):
👍

addy419:
@no-mood isn't it fine to use the same option to append `,ssh` at the end of components?

nixos/modules/services/desktops/gnome/gnome-keyring.nix
Aleksanaa: This `${builtins.getEnv "XDG_RUNTIME_DIR"}` is `$XDG_RUNTIME_DIR` in eval, not in activation.

The current way is not that bad though:

https://github.com/NixOS/nixpkgs/blob/8535fb92661f37ff9f0da3007fbc942f7d134b41/nixos/modules/services/security/yubikey-agent.nix#L48-L52

addy419:
The gnome keyring component for setting ssh is moved to gcr https://wiki.archlinux.org/title/GNOME/Keyring#Setup_gcr

Seems like the systemd component is added to gcr_4 https://github.com/NixOS/nixpkgs/blob/nixos-unstable/pkgs/development/libraries/gcr/4.nix#L81

So changing pkgs.gcr to pkgs.gcr_4 should do the trick? In theory

Aleksanaa:
There is another problem. SSH_AUTH_SOCK should be mutually exclusive in various modules to avoid undefined behavior. In the current implementation, they will not conflict with each other during eval.

no-mood:
> There is another problem. SSH_AUTH_SOCK should be mutually exclusive in various modules to avoid undefined behavior. In the current implementation, they will not conflict with each other during eval.

Something like this was added here: https://github.com/NixOS/nixpkgs/blob/71bae31b7dbc335528ca7e96f479ec93462323ff/nixos/modules/programs/gnupg.nix#L238
I feel like we could use add another layer and make an option. Something like `services.keyring = gnome-keyring`

pluiedev:
> I feel like we could use add another layer and make an option. Something like `services.keyring = gnome-keyring` 

An option for setting `SSH_AUTH_SOCK` would definitely be nice — I use 1Password as my SSH signer at the moment and that also requires setting `SSH_AUTH_SOCK`. Having a dedicated option for that would allow both GNOME Keyring and 1Pass to use the same mechanism which should make things much simpler

no-mood:
> > I feel like we could use add another layer and make an option. Something like `services.keyring = gnome-keyring`
> 
> An option for setting `SSH_AUTH_SOCK` would definitely be nice — I use 1Password as my SSH signer at the moment and that also requires setting `SSH_AUTH_SOCK`. Having a dedicated option for that would allow both GNOME Keyring and 1Pass to use the same mechanism which should make things much simpler
I agree. Is this the right place for this or should I open a new issue?
Could I propose a solution with a commit?

pluiedev:
> > An option for setting `SSH_AUTH_SOCK` would definitely be nice — I use 1Password as my SSH signer at the moment and that also requires setting `SSH_AUTH_SOCK`. Having a dedicated option for that would allow both GNOME Keyring and 1Pass to use the same mechanism which should make things much simpler
> 
> I agree. Is this the right place for this or should I open a new issue? Could I propose a solution with a commit?

I think it's reasonable to add it in this PR and then change the title (therefore repurpose) the PR to reflect the change

no-mood:
Before making a commit, the idea would be to add a level of abstraction and create an additional option that sets `$SSH_AUTH_SOCK`. Each module will then offer an option to be set as SSH Agent using that option.

Example: The new option will be `users.users.<name>.SSHAgent`. Each module, like Gnome-Keyring, instead of setting `SSH_AUTH_SOCK` directly, will access the SSHAgent parameter like this:
`users.users.<name>.SSHAgent ="$XDG_RUNTIME_DIR/ssh-agent"`.

By creating this option, checking for other active SSH agents won't be needed.

gabyx:
any progress on this?

PR #311039, opened 2024-05-12
nixos/smokeping: use nginx instead of thttpd

### LINKED ISSUE #265953: Smokeping doesn't (re)start thttpd.service when it should
### Describe the bug
When `smokeping.service` gets stopped and started (such as by `nixos-rebuild switch`) and you have `services.smokeping.webService` enabled, then `thttpd` should (re)start. But it doesn't.

Note, when I do `systemctl restart smokeping.service` then both services do properly restart.

### Steps To Reproduce
1. Enable smokeping
2. Make sure `services.smokeping.webService` is enabled (which is the default)
3. Try to do things like adding targets, via `nixos-rebuild switch`
  3a. Notice `stopping the following units: smokeping.service` which also (silently) stops `thttpd.service`
  3b. Notice `starting the following units: smokeping.service` which doesn't start `thttpd.service`
  3c. Confusingly stare at non-responsive webpage while the service seems running fine, until you realiset that there's a separate webserver service involved.

### Expected behavior
`thttpd.service` should either not stop when not required to stop, or (re)start when required to be running.

### Additional context
I see two possible solutions:
1. remove "requires smokeping" from the thttpd service
2. add "requires thttpd" to the smokeping service

### Notify maintainers

@erictapen @nh2 

### Metadata

```console
peter@niks ~> nix-shell -p nix-info --run "nix-info -m"
 - system: `"x86_64-linux"`
 - host os: `Linux 6.1.60, NixOS, 23.05 (Stoat), 23.05.4726.aeefe2054617`
 - multi-user?: `yes`
 - sandbox: `yes`
 - version: `nix-env (Nix) 2.13.6`
 - channels(root): `"nixos-23.05-small"`
 - channels(peter): `""`

```

DavHau:
Motivation:
fixes #265953

Changes:
- deprecate `services.smokeping.port` in favor of the niginx native option
- mention in release notes

## Description of changes

cc @nh2 

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [x] Tested, as applicable:

---

### DIFF
--- a/nixos/doc/manual/release-notes/rl-2405.section.md
+++ b/nixos/doc/manual/release-notes/rl-2405.section.md
@@ -399,6 +399,10 @@ The pre-existing [services.ankisyncd](#opt-services.ankisyncd.enable) has been m
 
 - `halloy` package was updated past 2024.5 which introduced a breaking change by switching the config format from YAML to TOML. See https://github.com/squidowl/halloy/releases/tag/2024.5 for details.
 
+- If `services.smokeping.webService` was enabled, smokeping is now served via nginx instead of thttpd. This change brings the following consequences:
+  - The default port for smokeping is now the nginx default port 80 instead of 8081.
+  - The option `services.smokeping.port` has been removed. To customize the port, use `services.nginx.virtualHosts.smokeping.listen.*.port`.
+
 - The `wpaperd` package has a breaking change moving to 1.0.1, previous version 0.3.0 had 2 different configuration files, one for wpaperd and one for the wallpapers. Remove the former and move the latter (`wallpaper.toml`) to `config.toml`.
 
 - Ada packages (libraries and tools) have been moved into the `gnatPackages` scope. `gnatPackages` uses the default GNAT compiler, `gnat12Packages` and `gnat13Packages` use the respective matching compiler version.
--- a/nixos/modules/services/networking/smokeping.nix
+++ b/nixos/modules/services/networking/smokeping.nix
@@ -47,6 +47,13 @@ let
 in
 
 {
+  imports = [
+    (mkRemovedOptionModule [ "services" "smokeping" "port" ] ''
+      The smokeping web service is now served by nginx.
+      In order to change the port, you need to change the nginx configuration under `services.nginx.virtualHosts.smokeping.listen.*.port`.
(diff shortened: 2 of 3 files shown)


Lassulus (approved):

nixos/doc/manual/release-notes/rl-2405.section.md
```diff
@@ -399,6 +399,10 @@ The pre-existing [services.ankisyncd](#opt-services.ankisyncd.enable) has been m
 
 - `halloy` package was updated past 2024.5 which introduced a breaking change by switching the config format from YAML to TOML. See https://github.com/squidowl/halloy/releases/tag/2024.5 for details.
 
+- If `services.smokeping.webService` was enabled, smokeping is now served via nginx instead of thttpd. This change brings the following consequences:
+  - The default port for smokeping is now the nginx default port 80 instead of 8081.
+  - The option `services.smokeping.port` has been removed. To customize the port, use `services.nginx.virtualHosts.smokeping.listen.*.port`.
+
```
bjornfor, line 405: Why couldn't services.smokeping.port be kept (option and value)? It seems the interface was abstract, and thttpd could be swapped for nginx without any user visible changes, but now users have to know the implementation detail of the nginx backend?

flokli: That essentially made it impossible to use the SSL functionalities the nginx vhosts did provide, or at least very unpredictable. We would need to carry a lot of complicated fallback logic, just for the sake of still providing the old port option. The maintenance overhead stood in no relation to the usefulness of that port option.

Moving this to plain nginx vhosts makes things much more predictable, using a well-maintained stack. We don't need another webserver, it's now just nginx vhost config, and if you want to use something else, you still can, configuring something similar to the nginx config in your webserver.

DavHau: To be more specific, nginx offers an option to enable ssl for all your vhosts at once, which would break by having this separate port option.

PR #307770, opened 2024-04-29
testers.hasPkgConfigModules: add optional version check

nicoonoclaste:
## Description of changes

- `testers`: add `pkgConfigModulesVersions`, which tests that pkgconf modules' version metadata match the package's
- `miniz`: use the new tester
- `doc`: mention the new tester in `pkg-config` section

**Not** done: making the new tester run as part of `testMetaPkgConfig`, as the new test fails on some drvs.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
- [x] Tested, as applicable:
  - and/or [package tests](https://github.com/NixOS/nixpkgs/blob/master/pkgs/README.md#package-tests)
- [x] Tested compilation of all packages that depend on this change using `nixpkgs-review`.
- [24.05 Release Notes](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2405.section.md) (or backporting [23.05](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2305.section.md) and [23.11](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2311.section.md) Release notes)
  - [ ] (Package updates) Added a release notes entry if the change is major or breaking
- [x] Fits [CONTRIBUTING.md](https://github.com/NixOS/nixpkgs/blob/master/CONTRIBUTING.md).

---

Add a :+1: [reaction] to [pull requests you find important].

[reaction]: https://github.blog/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/
[pull requests you find important]: https://github.com/NixOS/nixpkgs/pulls?q=is%3Aopen+sort%3Areactions-%2B1-desc

PR #311069, opened 2024-05-12
testers.hasPkgConfigModules: fix bug in `versionCheck` handling

nicoonoclaste:
## Description of changes

- add testcase for the bug reported in https://github.com/NixOS/nixpkgs/pull/307770#issuecomment-2102586718
- fix bug

Alternative to #310535

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
- [x] Tested, as applicable:

---

### DIFF
--- a/pkgs/build-support/testers/hasPkgConfigModules/tester.nix
+++ b/pkgs/build-support/testers/hasPkgConfigModules/tester.nix
@@ -55,7 +55,7 @@ runCommand testName {
       fi
     done
 
-    if [[ $notFound -eq 0 ]] && ([[ $versionMismatch -eq 0 ]] || [[ "$versionCheck" == false ]]); then
+    if [[ $notFound -eq 0 ]] && ([[ $versionMismatch -eq 0 ]] || [[ -z "$versionCheck" ]]); then
       exit 0
     fi
     if [[ $notFound -ne 0 ]]; then
--- a/pkgs/build-support/testers/hasPkgConfigModules/tester.nix
+++ b/pkgs/build-support/testers/hasPkgConfigModules/tester.nix
@@ -45,7 +45,7 @@ runCommand testName {
         if [[ "$moduleVersion" == "$version" ]]; then
           echo "✅ pkg-config module $moduleName exists and has version $moduleVersion"
         else
-          echo "${if versionCheck then "❌" else "⚠"} pkg-config module $moduleName exists and has version $moduleVersion when $version was expected"
+          echo "${if versionCheck then "❌" else "ℹ️"} pkg-config module $moduleName exists at version $moduleVersion != $version (drv version)"
           ((versionMismatch+=1))
         fi
         printf '%s\t%s\n' "$moduleName" "$version" >> "$out"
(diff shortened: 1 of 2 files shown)


roberth:
Good idea to just open a PR. I like your idea, but it needs a bit of work.

I didn't bother with a warning, but if we're going to warn, we should be careful not to cause warning fatigue.
Let's consider the possible combinations:

- `versionCheck == true`:
  - **ok**: report ok
  - **mismatch**: report error
- `versionCheck == false`:
  - Caller has decided that they don't want it.
  - **mismatch**: Don't cause warning fatigue. Don't report anything.
  - **ok**: It may *seem* better to report, but probably the caller has a good reason to disable the check; maybe the version is unreliable instead of always incorrect
- unset:
  - **ok**: Caller should set it to true. Tell them to enable the check?
  - **mismatch**: Caller should provide the expected version or disable the check, and we can tell in a message.

What do you think?

roberth:
> Or did you mean in the tester's log output

That's what I meant, yeah.


> callers won't normally see the version-mismatch messages because they are only part of the tester's build log,

Users of the package won't, but maintainers and contributors to the package will, and they won't easily understand what's going on.




> I had some thoughs how to gently push callers towards using `versionCheck`, but that seems out-of-scope for this bugfix PR... and for the 24.05 release cycle TBH

Interesting. I agree that we should first fix the error behavior, but I don't think we should trade it for a release that causes warning fatigue. Maybe remove the message altogether when the version check is disabled? Then we can merge that and add a helpful message without hurry.

pkgs/build-support/testers/hasPkgConfigModules/tester.nix
```diff
@@ -45,7 +45,7 @@ runCommand testName {
         if [[ "$moduleVersion" == "$version" ]]; then
           echo "✅ pkg-config module $moduleName exists and has version $moduleVersion"
         else
-          echo "❌ pkg-config module $moduleName exists and has version $moduleVersion when $version was expected"
+          echo "${if versionCheck then "❌" else "⚠"} pkg-config module $moduleName exists and has version $moduleVersion when $version was expected"
           ((versionMismatch+=1))
```
roberth, line 49: To make my suggestion more concrete, from https://github.com/NixOS/nixpkgs/pull/311069#issuecomment-2106355589

This is a little silly and untested, but I guess it gets the intended behavior across, fwiw.

```suggestion
          ${if versionCheck
            then ''
              echo "❌ pkg-config module $moduleName exists and has version $moduleVersion when $version was expected"
              ((versionMismatch+=1))
            ''
            # TODO: figure out the UX for enabling the version check; see https://github.com/NixOS/nixpkgs/pull/311069
            else ''
              echo "✅ pkg-config module $moduleName exists and has version $moduleVersion"
            ''}
```

roberth: It's tempting to add more info to the message, but that will just serve to distract whoever reads it, if we don't explain it, which is out of scope for this PR.

roberth:
LGTM if you could rebase away the fixup.

> I find it odd to talk about “warning fatigue” in that context, since it's usually meant for warnings emitted in nix evaluation.

It's a concept that's applicable to all warnings. Eval warnings can be more visible, so it's worse there. Still bad in build logs though.

> In any case, we could easily just not emit the warning at all unless `versionCheck` if that's a blocker

Let's just fix it first. We can get into friendly nudging later.

roberth:
Thanks!

PR #311114, opened 2024-05-12
nixos/vim: use full path in EDITOR environment variable

amozeo:
## Description of changes

use `/run/current-system/sw/bin/vim` in EDITOR environment variable when user set `programs.vim.defaultEditor` to `true`. This will result in a full path to the vim executable. It is not possible to set that environment variable if vim is not enabled. Previously it was relied of `vim` binary being in PATH, which is not always the case, for example using `nix-shell --pure --keep EDITOR`

related PR: https://github.com/NixOS/nixpkgs/pull/310982

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
- For non-Linux: Is sandboxing enabled in `nix.conf`? (See [Nix manual](https://nixos.org/manual/nix/stable/command-ref/conf-file.html))
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:
  - [NixOS test(s)](https://nixos.org/manual/nixos/unstable/index.html#sec-nixos-tests) (look inside [nixos/tests](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests))
  - and/or [package tests](https://github.com/NixOS/nixpkgs/blob/master/pkgs/README.md#package-tests)
  - or, for functions and "core" functionality, tests in [lib/tests](https://github.com/NixOS/nixpkgs/blob/master/lib/tests) or [pkgs/test](https://github.com/NixOS/nixpkgs/blob/master/pkgs/test)
  - made sure NixOS tests are [linked](https://nixos.org/manual/nixpkgs/unstable/#ssec-nixos-tests-linking) to the relevant packages
- [ ] Tested compilation of all packages that depend on this change using `nix-shell -p nixpkgs-review --run "nixpkgs-review rev HEAD"`. Note: all changes have to be committed, also see [nixpkgs-review usage](https://github.com/Mic92/nixpkgs-review#usage)
- [ ] Tested basic functionality of all binary files (usually in `./result/bin/`)
[...]

PR #310982, opened 2024-05-12
nixos/neovim: use full path in EDITOR environment variable

amozeo:
## Description of changes

use `/run/current-system/sw/bin/nvim` in EDITOR environment variable when user set `programs.neovim.defaultEditor` to `true`. This will result in a full path to the nvim executable. It is not possible to set that environment variable if nvim is not enabled. Previously it was relied of `nvim` binary being in PATH, which is not always the case, for example using `nix-shell --pure --keep EDITOR`

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
- For non-Linux: Is sandboxing enabled in `nix.conf`? (See [Nix manual](https://nixos.org/manual/nix/stable/command-ref/conf-file.html))
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:
  - [NixOS test(s)](https://nixos.org/manual/nixos/unstable/index.html#sec-nixos-tests) (look inside [nixos/tests](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests))
  - and/or [package tests](https://github.com/NixOS/nixpkgs/blob/master/pkgs/README.md#package-tests)
  - or, for functions and "core" functionality, tests in [lib/tests](https://github.com/NixOS/nixpkgs/blob/master/lib/tests) or [pkgs/test](https://github.com/NixOS/nixpkgs/blob/master/pkgs/test)
  - made sure NixOS tests are [linked](https://nixos.org/manual/nixpkgs/unstable/#ssec-nixos-tests-linking) to the relevant packages
- [ ] Tested compilation of all packages that depend on this change using `nix-shell -p nixpkgs-review --run "nixpkgs-review rev HEAD"`. Note: all changes have to be committed, also see [nixpkgs-review usage](https://github.com/Mic92/nixpkgs-review#usage)
- [ ] Tested basic functionality of all binary files (usually in `./result/bin/`)
- [24.05 Release Notes](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2405.section.md) (or backporting [23.05](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2305.section.md) and [23.11](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2311.section.md) Release notes)
[...]

PR #311157, opened 2024-05-12
nixos/emacs: use full path in EDITOR environment variable

amozeo:
## Description of changes

use `/run/current-system/sw/bin/emacseditor` in EDITOR environment variable when user set `services.emacs.defaultEditor` to `true`. This will result in a full path to the emacseditor executable. It is not possible to set that environment variable if emacs service is not enabled. Previously it was relied of `emacseditor` binary being in PATH, which is not always the case, for example using `nix-shell --pure --keep EDITOR`

related PRs: https://github.com/NixOS/nixpkgs/pull/310982 https://github.com/NixOS/nixpkgs/pull/311114

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/nixos/modules/services/editors/emacs.nix
+++ b/nixos/modules/services/editors/emacs.nix
@@ -85,7 +85,7 @@ in
 
     environment.systemPackages = [ cfg.package editorScript ];
 
-    environment.variables.EDITOR = mkIf cfg.defaultEditor (mkOverride 900 "emacseditor");
+    environment.variables.EDITOR = mkIf cfg.defaultEditor (mkOverride 900 "/run/current-system/sw/bin/emacseditor");
   };
 
   meta.doc = ./emacs.md;


Qyriad (approved):
Seems reasonable. Plenty of stuff in NixOS assumes `/run/current-system`, though I don't see any existing uses in this kind of environment variable specifically.

jian-lin (dismissed):
LGTM

Atemu (changes requested):
I'm not convinced this is a good idea as this breaks the principle of least surprise. When I run `nix-shell -p emacs -I nixpkgs=...`, I'd expect any command using `$EDITOR` to use the emacs from the Nix shell.

`nix-shell --pure --keep EDITOR` is a contrived example and I'd actually expect it to behave exactly the way it does right now. If your `$PATH` is broken, that's not a problem that the system-wide env var config should need to solve.

Could you point to an actual issue which would be solved by this?

jian-lin:
> When I run nix-shell -p emacs -I nixpkgs=..., I'd expect any command using $EDITOR to use the emacs from the Nix shell

This is a valid argument for other editors such as `vim`.

But it is not valid in this case because `EDITOR=emacseditor` is a wrapper for `emacsclient` which talks to an Emacs daemon.  If you use `emacsclient` in that `nix-shell` environment, talking to **the** Emacs daemon is a reasonable behaviour.

Atemu:
That is a good point and for the Emacs part, I'd agree. If there is only supposed to be a single instance of something, that instance should also be used. For the case of `emacsclient`, I'd absolutely agree. However, the `emacseditor` script also contains `--alternate-editor` which points at an absolute emacs path again.

Much more of an edge-case but the problem is still evident.

amozeo:
I have created an issue https://github.com/NixOS/nixpkgs/issues/311207 which should be a good place for discussion.

> Let's discuss whether this is a good policy first.
@Atemu [here](https://github.com/NixOS/nixpkgs/pull/310982#pullrequestreview-2052706455)

PR #311171, opened 2024-05-12
cvc5: add python package

tsheinen:
## Description of changes

cvc5 has python bindings which aren't built by default. This PR allows for cvc5 to be built with python bindings and a python package which takes advantage of those bindings. 

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/applications/science/logic/cvc5/default.nix
+++ b/pkgs/applications/science/logic/cvc5/default.nix
@@ -1,7 +1,5 @@
 { lib, stdenv, fetchFromGitHub, pkg-config, cmake, flex, cadical, symfpu, gmp, python3, gtest, libantlr3c, antlr3_4, boost, jdk, pythonBindings ? false }:
 
-with lib;
-
 stdenv.mkDerivation rec {
   pname = "cvc5";
   version = "1.2.0";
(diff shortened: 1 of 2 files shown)


pkgs/applications/science/logic/cvc5/default.nix
```diff
@@ -1,4 +1,6 @@
-{ lib, stdenv, fetchFromGitHub, pkg-config, cmake, flex, cadical, symfpu, gmp, python3, gtest, libantlr3c, antlr3_4, boost, jdk }:
+{ lib, stdenv, fetchFromGitHub, pkg-config, cmake, flex, cadical, symfpu, gmp, python3, gtest, libantlr3c, antlr3_4, boost, jdk, pythonBindings ? false }:
+
+with lib;
```
pluiedev, line 3: [Never use global `with lib;`.](https://github.com/NixOS/nixpkgs/issues/208242)

pluiedev (changes requested):
I would call the overridable variable something more like `withPythonBindings` to further show that it's part of the package.

Also, the current formatting is generating a lot of diffs that are frankly a bit unsightly. Perhaps consider reformatting it with `nixfmt-rfc-style` while you're at it?

pkgs/applications/science/logic/cvc5/default.nix
tsheinen: Oh, I see -- that makes sense in retrospect.  I'll adjust it to not use `with`. 

Could you elaborate on the issue with the diffs? I can reformat it while I'm at it but I don't understand which diffs we're talking about.

pluiedev: Having all the dependencies on one line makes the diff appear as if the entire line has deleted and added back again, whereas in reality only `pythonBindings` has been added. 

This is why modern formatting styles put a newline between every dependency, so that additions like these are much easier to read. [Example](https://github.com/pluiedev/nixpkgs/blob/f7c5d7966670a65b45306aecf7eb209bbb1686fb/pkgs/by-name/nz/nzp-fteqw/package.nix#L1-L40)

pluiedev (approved):

tsheinen:
Ah wait this breaks darwin tests in OfBorg looks like? Maybe?  It's unclear to me why "all checks pass" with an explicit error `error: cycle detected in build of '/nix/store/6ixhxkn3hmyi6y4dz2hwfx3i5vy199ns-cvc5-1.1.2.drv' in the references of output 'lib' from output 'out'` -- but it looks like it did not do this before so i will investigate further

pkgs/applications/science/logic/cvc5/default.nix
```diff
@@ -1,39 +1,101 @@
...
+    repo = "cvc5";
+    rev = "cvc5-${version}";
+    hash = "sha256-v+3/2IUslQOySxFDYgTBWJIDnyjbU2RPdpfLcIkEtgQ=";
   };
 
-  nativeBuildInputs = [ pkg-config cmake flex ];
+  pythonic = fetchFromGitHub {
+    owner = "cvc5";
+    repo = "cvc5_pythonic_api";
+    rev = "0bb7f092113e3b933dcd61b0959f3e9be9638109";
```
risicle, line 34: How do you propose to manage the version of this?

I suggest trying to synchronize it from the version specified in https://github.com/cvc5/cvc5/blob/887ecb98a2dd682c19888cde3789871b3dc158fe/cmake/FindCVC5PythonicAPI.cmake#L37. Actually _extracting_ it from there (thus daisy-chaining the two `fetchFromGitHub`s somehow) would be quite painful, but you might be able to add a line that checks the fetched version corresponds with the version expected in `FindCVC5PythonicAPI.cmake` and abort if not, prompting the updater to bump the fetched version.

pkgs/applications/science/logic/cvc5/default.nix
```diff
@@ -1,39 +1,101 @@
...
+    ))
   ];
 
   preConfigure = ''
     patchShebangs ./src/
   '';
 
+  postInstall = lib.optionalString withPythonBindings ''
+    mkdir -p $python/lib
+    mv $out/lib/python3.11/ $python/lib/.
```
risicle, line 72: This needs to be made to work with arbitrary python versions.

PR #311299, opened 2024-05-13
pkgs/formats: add generator for PHP config files

thenhnn:
## Description of changes

Some PHP apps are configurable through evaluation of `config.php` file (file should contain php code that should set options on hashmap), for example [Filesender](https://github.com/filesender/filesender) or [Flarum](https://docs.flarum.org/config/).

This PR adds generation of PHP configs through Nix DSL.

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [x] Tested, as applicable:

---

### DIFF
--- a/nixos/doc/manual/development/settings-options.section.md
+++ b/nixos/doc/manual/development/settings-options.section.md
@@ -146,6 +146,27 @@ have a predefined type and string generator already declared under
     :   Outputs the given attribute set as an Elixir map, instead of the
         default Elixir keyword list
 
+`pkgs.formats.php { finalVariable }` []{#pkgs-formats-php}
+
+:   A function taking an attribute set with values
+
+    `finalVariable`
+
+    :   The variable that will store generated expression (usually `config`). If set to `null`, generated expression will contain `return`.
+
+    It returns a set with PHP-Config-specific attributes `type`, `lib`, and
+    `generate` as specified [below](#pkgs-formats-result).
+
+    The `lib` attribute contains functions to be used in settings, for
+    generating special PHP values:
+
+    `mkRaw phpCode`
+
+    :   Outputs the given string as raw PHP code
+
+    `mkMixedArray list set`
+
+    :   Creates PHP array that contains both indexed and associative values. For example, `lib.mkMixedArray [ "hello" "world" ] { "nix" = "is-great"; }` returns `['hello', 'world', 'nix' => 'is-great']`
 
 []{#pkgs-formats-result}
 These functions all return an attribute set with these values:
--- a/pkgs/pkgs-lib/formats.nix
+++ b/pkgs/pkgs-lib/formats.nix
@@ -43,6 +43,8 @@ rec {
 
   hocon = (import ./formats/hocon/default.nix { inherit lib pkgs; }).format;
 
+  php = (import ./formats/php/default.nix { inherit lib pkgs; }).format;
+
   json = {}: {
 
     type = with lib.types; let
--- a/pkgs/pkgs-lib/formats/php/default.nix
+++ b/pkgs/pkgs-lib/formats/php/default.nix
@@ -1,0 +1,69 @@
+{pkgs, lib}: {
+  # Format for defining configuration of some PHP services, that use "include 'config.php';" approach.
+  format = {finalVariable ? null}: let
+    toPHP = value: {
+        "null" = "null";
+        "bool" = if value then "true" else "false";
+        "int" = toString value;
+        "float" = toString value;
+        "string" = string value;
+        "set" = attrs value;
+        "list" = list value;
+      }
+      .${builtins.typeOf value} or
+        (abort "should never happen: unknown value type ${builtins.typeOf value}");
+
+    # https://www.php.net/manual/en/language.types.string.php#language.types.string.syntax.single
+    escapeSingleQuotedString = lib.escape [ "'" "\\" ];
+    string = value: "'${escapeSingleQuotedString value}'";
+
+    listContent = values: lib.concatStringsSep ", " (map toPHP values);
+    list = values: "[" + (listContent values) + "]";
+
+    attrsContent = values: lib.pipe values [
+      (lib.mapAttrsToList (k: v: "${toPHP k} => ${toPHP v}"))
+      (lib.concatStringsSep ", ")
+    ];
+    attrs = set:
+      if set ? _phpType then specialType set
+      else
+          "[" + attrsContent set + "]";
+
+    mixedArray = {list, set}: if list == [] then attrs set else "[${listContent list}, ${attrsContent set}]";
+
+    specialType = {value, _phpType}: {
+      "mixed_array" = mixedArray value;
+      "raw" = value;
+    }.${_phpType};
+
+    type = with lib.types;
+      nullOr (oneOf [
+        bool
+        int
+        float
+        str
+        (attrsOf type)
+        (listOf type)
+      ])
+      // {
+        description = "PHP value";
+      };
+  in {
+
+    inherit type;
+
+    lib = {
+      mkMixedArray = list: set: {_phpType = "mixed_array"; value = { inherit list set;}; };
+      mkRaw = raw: {_phpType = "raw"; value = raw;};
+    };
+
+    generate = name: value: pkgs.writeTextFile {
+      inherit name;
+      text = let
+        # strict_types enabled here to easily debug problems when calling functions of incorrect type using `mkRaw`.
+        phpHeader = "<?php\ndeclare(strict_types=1);\n";
+      in if finalVariable == null then phpHeader + "return ${toPHP value};\n" else phpHeader + "\$${finalVariable} = ${toPHP value};\n";
+    };
+
+  };
+}
--- a/pkgs/pkgs-lib/tests/formats.nix
+++ b/pkgs/pkgs-lib/tests/formats.nix
@@ -425,4 +425,48 @@ in runBuildTests {
       \u0627\u0644\u062c\u0628\u0631 = \u0623\u0643\u062b\u0631 \u0645\u0646 \u0645\u062c\u0631\u062f \u0623\u0631\u0642\u0627\u0645
     '';
   };
+
+  phpAtoms = shouldPass rec {
+    format = formats.php { finalVariable = "config"; };
+    input = {
+      null = null;
+      false = false;
+      true = true;
+      int = 10;
+      float = 3.141;
+      str = "foo";
+      str_special = "foo\ntesthello'''";
+      attrs.foo = null;
+      list = [ null null ];
+      mixed = format.lib.mkMixedArray [ 10 3.141 ] {
+        str = "foo";
+        attrs.foo = null;
+      };
+      raw = format.lib.mkRaw "random_function()";
+    };
+    expected = ''
+      <?php
+      declare(strict_types=1);
+      $config = ['attrs' => ['foo' => null], 'false' => false, 'float' => 3.141000, 'int' => 10, 'list' => [null, null], 'mixed' => [10, 3.141000, 'attrs' => ['foo' => null], 'str' => 'foo'], 'null' => null, 'raw' => random_function(), 'str' => 'foo', 'str_special' => 'foo
+      testhello\'\'\'${"'"}, 'true' => true];
+    '';
+  };
+
+  phpReturn = shouldPass {
+    format = formats.php { };
+    input = {
+      int = 10;
+      float = 3.141;
+      str = "foo";
+      str_special = "foo\ntesthello'''";
+      attrs.foo = null;
+    };
+    expected = ''
+      <?php
+      declare(strict_types=1);
+      return ['attrs' => ['foo' => null], 'float' => 3.141000, 'int' => 10, 'str' => 'foo', 'str_special' => 'foo
+      testhello\'\'\'${"'"}];
+    '';
+  };
+
 }


pkgs/pkgs-lib/formats.nix
```diff
@@ -274,6 +274,68 @@ rec {
...
+      if value == true then "true" else
+      if value == false then "false" else
+      if isInt value || isFloat value then toString value else
+      if isString value then string value else
+      if isAttrs value then attrs value else
+      if isList value then list value else
+      abort "formats.php: should never happen (value = ${value})";
+
+    escapePHP = escape [ "\\" "\"" ];
+    string = value: "\"${escapePHP value}\"";
```
Ma27, line 290: Why a double quote? This means that `${foo}`, `{$foo}` and `$foo` are substitutable.

thenhnn: Oops, I didn't know that. Thanks, I will fix.

pkgs/pkgs-lib/formats.nix
```diff
@@ -274,6 +274,68 @@ rec {
...
+    inherit type;
+
+    lib = {
+      mkMixedArray = list: assert list != []; set: {_phpType = "mixed_array"; value = { inherit list set;}; };
+      mkRaw = raw: {_phpType = "raw"; value = raw;};
+    };
+
+    generate = name: value: pkgs.writeTextFile {
+      name = "config.php";
+      text = if finalVariable == null then "<?php\nreturn ${toPHP value};\n" else "<?php\n${finalVariable} = ${toPHP value};\n";
```
drupol, line 334: Could we add `declare(strict_types=1);` ?

drupol (commented):
I quickly reviewed and left a comment.

I'm wondering if we could validate the config with `php -l` by default and eventually format it using PSR-12 with PHP-CS-Fixer.

pkgs/pkgs-lib/formats.nix
thenhnn: What's the win from that? 

> Strict typing applies to function calls made from within the file with strict typing enabled, not to the functions declared within that file. If a file without strict typing enabled makes a call to a function that was defined in a file with strict typing, the caller's preference (coercive typing) will be respected, and the value will be coerced.

This only applies for function calls, which is very rare case in configs.

thenhnn:
> I quickly reviewed and left a comment.
> 
> I'm wondering if we could validate the config with `php -l` by default and eventually format it using PSR-12 with PHP-CS-Fixer.

I'm afraid that it will slow down evaluation a lot.

pkgs/pkgs-lib/formats.nix
drupol: That's true too.

Essentially, I use `declare(strict_types=1);` by default and for promoting best practices when one open a PHP file. To me it should be by default.

thenhnn: Is it supported only in PHP 8? I can't find mention of versions on PHP website.

drupol: It's available since ... PHP 4: https://3v4l.org/IePpo#v4.3.11

thenhnn: Done

pkgs/pkgs-lib/tests/formats.nix
```diff
@@ -425,4 +425,48 @@ in runBuildTests {
...
+    format = formats.php { finalVariable = "$config"; };
+    input = {
+      null = null;
+      false = false;
+      true = true;
+      int = 10;
+      float = 3.141;
+      str = "foo";
+      str_special = "foo\ntesthello'''";
+      attrs.foo = null;
```
drupol, line 439: If adding `attrs.bar = "bar";` just under this line, will it convert it correctly into a single `attrs => [ 'bar' => 'bar', 'foo' => null ]` ?

thenhnn: Yeah. From nix perspective `attrs.bar = "bar"; attrs.foo = null;` is equal to `attrs = { bar = "bar"; foo = null; };`.

drupol: That's cool then.

drupol (approved):
LGTM, I'll wait for @jtojnar to give its feedback on this before merging.

pkgs/pkgs-lib/formats.nix
```diff
@@ -274,6 +274,70 @@ rec {
...
+
+    escapePHP = escape [ "'" ];
+    string = value: "'${escapePHP value}'";
+
+    listContent = values: concatStringsSep ", " (map toPHP values);
+    list = values: "[" + (listContent values) + "]";
+
+    attrsContent = values: pipe values [
+      (mapAttrsToList (k: v: "${toPHP k} => ${toPHP v}"))
+      (concatStringsSep ", ")
```
jtojnar, line 297: For easier debugging, it might be nice to format the output so that it  is not squashed into a single unreadable line.

pkgs/pkgs-lib/formats.nix
```diff
@@ -274,6 +274,70 @@ rec {
...
+    lib = {
+      mkMixedArray = list: assert list != []; set: {_phpType = "mixed_array"; value = { inherit list set;}; };
+      mkRaw = raw: {_phpType = "raw"; value = raw;};
+    };
+
+    generate = name: value: pkgs.writeTextFile {
+      name = "config.php";
+      text = let
+        phpHeader = "<?php\ndeclare(strict_types=1);\n";
+      in if isNull finalVariable then phpHeader + "return ${toPHP value};\n" else phpHeader + "${finalVariable} = ${toPHP value};\n";
```
jtojnar, line 336: `isNull` is deprecated. Use `== null` instead.

pkgs/pkgs-lib/formats.nix
```diff
@@ -274,6 +274,70 @@ rec {
...
+  in {
+
+    inherit type;
+
+    lib = {
+      mkMixedArray = list: assert list != []; set: {_phpType = "mixed_array"; value = { inherit list set;}; };
+      mkRaw = raw: {_phpType = "raw"; value = raw;};
+    };
+
+    generate = name: value: pkgs.writeTextFile {
```
jtojnar, line 332: `name` is unused.

jtojnar (changes requested):

nixos/doc/manual/development/settings-options.section.md
```diff
@@ -146,6 +146,27 @@ have a predefined type and string generator already declared under
...
+    `generate` as specified [below](#pkgs-formats-result).
+
+    The `lib` attribute contains functions to be used in settings, for
+    generating special PHP values:
+
+    `mkRaw phpCode`
+
+    :   Outputs the given string as raw PHP code
+
+    `mkMixedArray list set`
```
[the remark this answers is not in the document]
thenhnn: They're used in [SimpleSAMLphp](https://simplesamlphp.org/docs/stable/simplesamlphp-sp.html).
Basically main point of this PR is to supported mixed lists :)

pkgs/pkgs-lib/formats.nix
thenhnn: Which formatter should I use?

pkgs/pkgs-lib/formats.nix
```diff
@@ -274,6 +274,70 @@ rec {
 
   };
 
+  # Format for defining configuration of some PHP services, that use "include 'config.php';" approach.
+  php = {finalVariable ? null}: with lib; let
```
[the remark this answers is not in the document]
thenhnn: Is there any Nix primitive type that is not listed here?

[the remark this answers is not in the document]
thenhnn: I don't know how to make it prettier.

nixos/doc/manual/development/settings-options.section.md
```diff
@@ -146,6 +146,27 @@ have a predefined type and string generator already declared under
     :   Outputs the given attribute set as an Elixir map, instead of the
         default Elixir keyword list
 
+`pkgs.formats.php { finalVariable }`
```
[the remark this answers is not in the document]
jtojnar: Just add the anchor to the end of the line.

jtojnar:
> > I'm wondering if we could validate the config with `php -l` by default and eventually format it using PSR-12  with PHP-CS-Fixer.
> 
> I'm afraid that it will slow down evaluation a lot.

It would be done at build time so there should be negligible effect on evaluation time.

nixos/doc/manual/development/settings-options.section.md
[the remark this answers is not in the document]
jtojnar: Maybe specify the order at least?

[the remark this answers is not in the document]
thenhnn: What do you mean? The order is `array(<list items>, <keys> => <values>)`.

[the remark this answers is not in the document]
jtojnar: Yes, it is not clear from the docs and as demonstrated above, it matters.

pkgs/pkgs-lib/formats.nix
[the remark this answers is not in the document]
jtojnar: Now I wonder if it maybe should be moved into separate file like the java-properties.

pkgs/pkgs-lib/formats.nix
jtojnar: php-cs-fixer should be fine. Or alternately, you could do a minimal formatting for legibility by hand by tracking indentation level of the printed value.

drupol: PHP-CS-Fixer is the way to go indeed.

pkgs/pkgs-lib/formats.nix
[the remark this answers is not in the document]
thenhnn: IMO it's not big enough for external file.

pkgs/pkgs-lib/formats.nix
thenhnn: Looks like I'm dumb, but I can't figure out how to make php-cs-fixer to actually format file.

php-cs-fixer fix doesn't work...

drupol: That works here:

```
❯ php-cs-fixer fix --rules=@PSR12 .
PHP CS Fixer 3.51.0 Insomnia by Fabien Potencier, Dariusz Ruminski and contributors.
PHP runtime: 8.2.19
Loaded config default.
 77/77 [▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓] 100%

   1) tests/unit/TypedIterableAggregateTest.php
   2) tests/unit/SortIterableAggregateTest.php
   3) tests/static-analysis/closure-iterator.php
   4) src/SortIterableAggregate.php
   5) src/RandomIterableAggregate.php
   6) src/ConcatIterableAggregate.php
   7) src/UnpackIterableAggregate.php
   8) src/FilterIterableAggregate.php
   9) src/NormalizeIterableAggregate.php
  10) src/ChunkIterableAggregate.php
  11) src/PackIterableAggregate.php
  12) src/InterruptableIterableIteratorAggregate.php
  13) src/MapIterableAggregate.php
  14) src/StringIteratorAggregate.php
  15) src/IterableIteratorAggregate.php
  16) src/RandomIntegerAggregate.php
  17) src/ReductionIterableAggregate.php
  18) src/MultipleIterableAggregate.php
  19) src/ReduceIterableAggregate.php
  20) src/LimitIterableAggregate.php
  21) stable-sort.php

Fixed 21 of 77 files in 0.387 seconds, 16.000 MB memory used
```

drupol:
Question: Is the Nix indentation mechanism and function still necessary if you pass the resulting PHP code through PHP-CS-Fixer?

pkgs/pkgs-lib/formats.nix
thenhnn: @drupol: Maybe I'm doing something wrong?
```
$ cat default.php
<?php

return ['attrs' => ['foo' => null],
'float' => 3.141000,
'int' => 10, 'str' => 'foo', 'hell' => ['attrs' => ['foo' => null], 'float' => 3.141000, 'int' => 10, 'str' => 'foo']];

$ php-cs-fixer fix --rules=@PSR12 default.php
PHP CS Fixer 3.51.0 Insomnia by Fabien Potencier, Dariusz Ruminski and contributors.
PHP runtime: 8.3.6
Loaded config default.
 1/1 [▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓] 100%


Fixed 0 of 1 files in 0.004 seconds, 14.000 MB memory used

$ cat default.php
<?php

return ['attrs' => ['foo' => null],
'float' => 3.141000,
'int' => 10, 'str' => 'foo', 'hell' => ['attrs' => ['foo' => null], 'float' => 3.141000, 'int' => 10, 'str' => 'foo']];
```

drupol: You need to pass a directory, not a file.

drupol: Passing a file will be soon available, ref: https://twitter.com/_Codito_/status/1789049130867650665

drupol: Indeed, I'm also unable to get it working now. WTH ?!?!

jtojnar:
> Question: Is the Nix indentation mechanism and function still necessary if you pass the resulting PHP code through PHP-CS-Fixer?

If we decide including PHP-CS-Fixer in the closure is fine, then Nix
indentation mechanism is pointless.
If we decide that the former is too heavy, then doing basic formatting
ourselves would be preferred.

thenhnn:
> Question: Is the Nix indentation mechanism and function still necessary if you pass the resulting PHP code through PHP-CS-Fixer?
> If we decide including PHP-CS-Fixer in the closure is fine, then Nix indentation mechanism is pointless. If we decide that the former is too heavy, then doing basic formatting ourselves would be preferred.

I'm a bit confused, why format auto-generated file at all?

jtojnar:
To make it easier for humans to read when debugging.

pkgs/pkgs-lib/formats.nix
thenhnn: I'm afraid to write custom indentation code because I'm not a PHP programmer. Is there any other formatters for PHP?

drupol: no.

thenhnn: ugh... i tried php_codesniffer but it does nothing too.. php is cursed.

jtojnar:
Formatting is just nice to have, IMO, so no need to block on that.

thenhnn:
> Formatting is just nice to have, IMO, so no need to block on that.

Can we just omit it for now?

thenhnn:
@jtojnar I think it's ready to merge.

pkgs/pkgs-lib/formats.nix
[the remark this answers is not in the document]
jtojnar: Well, lot of small changes does add up too. But I do not feel strongly about this.

jtojnar:
> Can we just omit it for now?

Yes, that is what I meant.



> I think it's ready to merge.

No other input from me, other than the minor preference for moving it to a separate file.

----

We should probably give @infinisil as a code owner a bit more time to react if they want.

thenhnn:
> > Can we just omit it for now?
> 
> Yes, that is what I meant.
> 
> > I think it's ready to merge.
> 
> No other input from me, other than the minor preference for moving it to a separate file.
> 
> We should probably give @infinisil as a code owner a bit more time to react if they want.

I moved it to separate file, yeah, it's better :)

Got it, thank you!

pkgs/pkgs-lib/formats/php/default.nix
```diff
@@ -0,0 +1,67 @@
...
+      ])
+      // {
+        description = "PHP value";
+      };
+  in {
+
+    inherit type;
+
+    lib = {
+      mkMixedArray = list: assert list != []; set: {_phpType = "mixed_array"; value = { inherit list set;}; };
```
infinisil, line 55: Why the assert here? If PHP supports mixed arrays with no indexed elements then this shouldn't be done

nixos/doc/manual/development/settings-options.section.md
```diff
@@ -146,6 +146,27 @@ have a predefined type and string generator already declared under
     :   Outputs the given attribute set as an Elixir map, instead of the
         default Elixir keyword list
 
+`pkgs.formats.php { finalVariable }` []{#pkgs-formats-php}
+
+:   A function taking an attribute set with values
+
+    `finalVariable`
+
+    :   The variable that will store generated expression (usually `$config`). If set to `null`, generated expression will contain `return`.
```
infinisil, line 155: It's a bit confusing on whether the `$` should be part of it. Consider either cleaning up the docs, or removing the need to specify the `$`

infinisil (changes requested):
Not bad! Leaving some feedback :)

pkgs/pkgs-lib/formats/php/default.nix
```diff
@@ -0,0 +1,67 @@
...
+    attrsContent = values: lib.pipe values [
+      (lib.mapAttrsToList (k: v: "${toPHP k} => ${toPHP v}"))
+      (lib.concatStringsSep ", ")
+    ];
+    attrs = set:
+      if set ? _phpType then specialType set
+      else
+          "[" + attrsContent set + "]";
+
+    mixedArray = {list, set}: "[${listContent list}, ${attrsContent set}]";
```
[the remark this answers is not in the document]
Ma27: (fwiw PHP supports trailing commas in lists. In fact, I know of coding styles that mandate those ;)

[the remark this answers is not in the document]
thenhnn: Yeah, that's why there's no check

pkgs/pkgs-lib/formats/php/default.nix
thenhnn: It's kinda bad, because if there's no indexed elements - why not use just set?

I replaced implementation to handle this case, not just throwing an error.

thenhnn:
@jtojnar Can you approve changes? Or I missed something?

pkgs/pkgs-lib/formats/php/default.nix
infinisil: This would be problematic if there's some third-party code making use of this function to just propagate some values, which would have to switch between a set and a mixed list depending on whether the list is empty. Not having an assertion avoids this

pkgs/pkgs-lib/tests/formats.nix
```diff
@@ -425,4 +425,48 @@ in runBuildTests {
       \u0627\u0644\u062c\u0628\u0631 = \u0623\u0643\u062b\u0631 \u0645\u0646 \u0645\u062c\u0631\u062f \u0623\u0631\u0642\u0627\u0645
     '';
   };
+
+  phpAtoms = shouldPass rec {
+    format = formats.php { finalVariable = "$config"; };
```
infinisil, line 430:
```suggestion
    format = formats.php { finalVariable = "config"; };
```

Though I don't see any test failure in CI 🤔

infinisil (approved):
Other than the above this looks fine to me!

pkgs/pkgs-lib/tests/formats.nix
thenhnn: oops D:

jtojnar (approved):

thenhnn:
so, it's ready to merge, right?

drupol:
Thank you!

PR #311307, opened 2024-05-13
Remove myself from some packages I no longer use

michaelpj:
## Description of changes

I still want to be on the maintainers list, but I happen to no longer be using these programs.


drupol:
Thank you for what you did already.

PR #311394, opened 2024-05-13
nixos/systemd: allow running shellcheck on generated systemd scripts

r-vdp:
## Description of changes

With this PR we can optionally run shellcheck on generated systemd scripts.
Our nixos closure already depends on GHC (and shellcheck), so this does not affect the runtime closure.

~~I fixed the scripts that I use on my machines, but there is definitely a lot more stuff to fix before this could become the default.~~
I will submit these as separate PRs.

For now, this sits behind an enable-option. It would be great if some people would enable this and submit fixes for existing scripts.

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---


nixos/lib/systemd-lib.nix
```diff
@@ -335,18 +335,27 @@ in rec {
...
     let
       scriptName = replaceStrings [ "\\" "@" ] [ "-" "_" ] (shellEscape name);
-      out = (pkgs.writeShellScriptBin scriptName ''
-        set -e
-        ${text}
-      '').overrideAttrs (_: {
+      out = (
+        if ! enableStrictShellChecks then
+          pkgs.writeShellScriptBin scriptName ''
+            set -e
```
SuperSandro2000, line 344: Naive question: Should set set -eou pipefail?

nixos/lib/systemd-lib.nix
```diff
@@ -335,18 +335,27 @@ in rec {
...
+            name = scriptName;
+            inherit text;
+          }
+      ).overrideAttrs (_: {
         # The derivation name is different from the script file name
         # to keep the script file name short to avoid cluttering logs.
         name = "unit-script-${scriptName}";
       });
-    in "${out}/bin/${scriptName}";
+    in lib.getExe out;
```
SuperSandro2000, line 358: nice 👍🏼

nixos/modules/virtualisation/nixos-containers.nix
```diff
@@ -85,8 +85,13 @@ let
...
     ''
-      mkdir -p -m 0755 "$root/etc" "$root/var/lib"
-      mkdir -p -m 0700 "$root/var/lib/private" "$root/root" /run/nixos-containers
+      # Declare root explicitly to avoid shellcheck warnings, it comes from the env
+      declare root
+
+      mkdir -p "$root/etc" "$root/var/lib"
+      chmod 0755 "$root/etc" "$root/var/lib"
+      mkdir -p "$root/var/lib/private" "$root/root" /run/nixos-containers
+      chmod 0700 "$root/var/lib/private" "$root/root" /run/nixos-containers
```
SuperSandro2000, line 94: I am not sure what the layout is before this. If $root already exists, we could also drop the -p where possible.

Also I don't think your change is fixing the issue shellcheck is noting. The directories along the path still will have the same permissions as before and the ones created by -p still have the default umask permissions.

nixos/modules/virtualisation/nixos-containers.nix
```diff
@@ -151,8 +159,8 @@ let
       # containerInit and stage2 have to handle this as well.
       exec ${config.systemd.package}/bin/systemd-nspawn \
         --keep-unit \
-        -M "$INSTANCE" -D "$root" $extraFlags \
-        $EXTRA_NSPAWN_FLAGS \
+        -M "$INSTANCE" -D "$root" "$extraFlags" \
+        "$EXTRA_NSPAWN_FLAGS" \
```
SuperSandro2000, line 163: Since extra flags is a string here, this will work. Normally, especially in stdenv, extraFlags is an array, which would mean we would mess with quoting here.

nixos/modules/system/activation/activation-script.nix
```diff
@@ -168,7 +168,7 @@ in
 
       apply = set: {
         script = ''
-          unset PATH
+          export PATH=/empty
```
SuperSandro2000, line 171: We would add /empty to the beginning of the path here which I don't think we want to do.

```suggestion
          export PATH=
```

nixos/modules/virtualisation/ec2-data.nix
```diff
@@ -46,19 +47,20 @@ with lib;
...
 
-            mkdir -m 0755 -p /etc/ssh
+            mkdir -p /etc/ssh
+            chown 0755 /etc/ssh
 
             if [ -s "$userData" ]; then
               key="$(sed 's/|/\n/g; s/SSH_HOST_DSA_KEY://; t; d' $userData)"
               key_pub="$(sed 's/SSH_HOST_DSA_KEY_PUB://; t; d' $userData)"
-              if [ -n "$key" -a -n "$key_pub" -a ! -e /etc/ssh/ssh_host_dsa_key ]; then
+              if [ -n "$key" ] && [ -n "$key_pub" ] && [ ! -e /etc/ssh/ssh_host_dsa_key ]; then
```
SuperSandro2000, line 56: I think we are always using bash here, right? Then we might as well use [[

nixos/modules/security/acme/default.nix
```diff
@@ -167,7 +167,7 @@ let
     # ensure all required lock files exist, but none more
     script = ''
       GLOBIGNORE="${concatStringsSep ":" concurrencyLockfiles}"
-      rm -f *
+      rm -f -- *
```
SuperSandro2000, line 170: good catch 👍🏼

nixos/modules/tasks/filesystems/zfs.nix
```diff
@@ -62,6 +62,7 @@ let
   # latter case it makes one last attempt at importing, allowing the system to
   # (eventually) boot even with a degraded pool.
   importLib = {zpoolCmd, awkCmd, cfgZfs}: ''
+    # shellcheck disable=SC2013
```
SuperSandro2000, line 65: We should add a short comment, why this was added

SuperSandro2000 (commented):
Very good idea. I think the only thing that really needs some discussion, is what we are going to do about the shellcheck mkdir -p error. Maybe we ignore it by default for now?

Also please remind me in a week or so, that I want to test this on my deployments.

ElvishJerricco (dismissed):
Seems like this should be something that can be enabled per-unit, with the global setting acting as the default.

nixos/lib/systemd-lib.nix
r-vdp: Yeah but that can be done in another PR then. It can also cause breakage, especially `-u`.

r-vdp:
> Seems like this should be something that can be enabled per-unit, with the global setting acting as the default.

@ElvishJerricco that's already how it's done. There is a global switch (defaults to `false` for now) and a per-unit switch that defaults to the value of the global switch.

So it's possible to enable or disable this on a per-unit basis, independent of the global option.

nixos/lib/systemd-lib.nix
SuperSandro2000: Yeah, that's true

ElvishJerricco:
Ah, I misunderstood what I was reading. Great!

nixos/modules/virtualisation/nixos-containers.nix
r-vdp: So there's nothing to be done here? Would you do this differently?

For an array, I guess you'd want to do `"${extraFlags[@]}"` instead, but then we'd need to test first whether the variable is an array or a string and act accordingly. I'm not sure that that's needed here currently?

ElvishJerricco (changes requested):

nixos/modules/tasks/filesystems/zfs.nix
```diff
@@ -871,12 +872,12 @@ in
...
         };
         script = ''
-          ${cfgZfs.package}/bin/zpool scrub -w ${
+          ${cfgZfs.package}/bin/zpool scrub -w "${
             if cfgScrub.pools != [] then
               (concatStringsSep " " cfgScrub.pools)
             else
               "$(${cfgZfs.package}/bin/zpool list -H -o name)"
-            }
+            }"
```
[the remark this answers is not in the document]
r-vdp: @phaer I think we don't need the quotes around the nix interpolation, do we? It should be sufficient to properly quote the generated bash code.

Could you test if that works? I think you made this change because of your zfs config, right?

nixos/modules/tasks/filesystems/zfs.nix
r-vdp: I rewrote this part such that we no longer need to ignore this warning.

nixos/modules/system/activation/activation-script.nix
r-vdp: Done. The same is already done on line 38 in the same file, not sure why exactly.

nixos/modules/virtualisation/ec2-data.nix
SuperSandro2000: > it may be slightly nicer to stick to posix?

Probably not, using [ wrong is pretty easy and [[ avoids some quoting traps.

nixos/modules/virtualisation/nixos-containers.nix
SuperSandro2000: yeah, unfortunate naming

nixos/modules/tasks/filesystems/zfs.nix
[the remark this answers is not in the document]
phaer: Sorry, I only saw this notification now. I think your assumption is correct; thanks for the catch @ElvishJerricco. Its just that I am in a train right now and won't be able to properly test this on a zfs machine for a good week, when I return home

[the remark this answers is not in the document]
r-vdp: I dropped this commit in order to unblock this PR, since I don't know zfs well enough to test this.

r-vdp:
Any other remarks on this? I will still add a changelog entry in any case.

emilazy (approved):
ACME changes LGTM. Didn’t review anything else.

ElvishJerricco (dismissed):
***Whoa***. Upon review, there are *a ton of changes* that have nothing to do with the logic of this PR. I will not review that. Please remove those changes and open a separate PR for them.

emilazy:
One PR per service or group of services is probably reasonable since they can be reviewed separately by the people responsible for them.

r-vdp:
OK, sure, they are already split in different commits, I'll make separate PRs for them.

r-vdp:
I removed the fixes for different services. I will make separate PRs for these.

You won't be able to build a NixOS system with shellcheck for services enabled until we also merge fixes for some of those.

Mindavi (approved):
I think splitting this up was the right call and on surface this looks good to me and I ACK the idea too.

nixos/modules/virtualisation/nixos-containers.nix
r-vdp: This was pulled out of this PR.

phaer (approved):
Very nice feature! Code looks good to me and feature works as intended. PR quite easy to review IMO.

PR #311459, opened 2024-05-13
nixpkgs-manual: init

philiptaron:
## Motivation

Make it easier and more natural to build the contents of the `doc/` directory.

## Description of changes

- [x] Reshape the derivations from `doc/` to `callPackage` style.
- [x] When needed, extract a derivation into its own file.
- [x] Organize them under the `nixpkgs-manual` name.
- [x] Make `doc/default.nix` and `doc/shell.nix` forward to them.

The diff is pretty chunky, but it's not too bad when you have the old and new side by side... sadly GitHub makes this pretty impossible.

Incidental changes:

- [x] `nixfmt` on new files
- [x] `stdenvNoCC` instead of `stdenv` as appropriate
- [x] Derivations accessible through `passthru`

## Why not `pkgs/by-name`?

Because it has a rule that packages in `pkgs/by-name` cannot refer to files outside their own directory, which this obviously needs to do. I've opened https://github.com/NixOS/nixpkgs-check-by-name/issues/64 to ideate about this use case -- it's not broad, but it would be nice -- but in the meantime, this derivation can't use `pkgs/by-name`.

The CI failure says that merging won't break the main branch.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [x] Tested, as applicable:
  - [x] `nix-build doc/`
  - [x] `nix-build -A nixpkgs-manual`
  - [x] `pushd doc/ && nix-shell`
  - [x] On master, `nix-build doc/`, on branch `nix-build doc/`, and then `diffoscope <original> <new>` (no changes except `nix-support/hydra-build-products`)


philiptaron:
@infinisil, what do you recommend I do about the `pkgs-check-by-name` failures? They're rightly pointing out that this package is reaching outside its directory... but, uh, it's OK in this one case? 😁

philiptaron:
I also need to mock out `lib` [in order to get ofBorg to pass](https://gist.github.com/GrahamcOfBorg/90c32ba673d7f6343202932a570588c2).

doc/default.nix
```diff
@@ -1,176 +1,5 @@
...
-    '') {
-      script = ./tests/manpage-urls.py;
-      urlsFile = ./manpage-urls.json;
-    };
-}
+{
+  pkgs ? (import ./.. { }),
+  nixpkgs ? { },
+}:
+pkgs.nixpkgs-manual.override { inherit nixpkgs; }
```
hsjobeki, line 5: I highly appreciate this change. For no reason the manual was not a pkg itself before. :)

pkgs/by-name/ni/nixpkgs-manual/options-json.nix
```diff
@@ -0,0 +1,39 @@
...
+  transformOptions = opt: opt // { declarations = map transformDeclarations opt.declarations; };
+
+  optionsDoc = nixosOptionsDoc {
+    inherit (modules) options;
+    inherit transformOptions;
+    documentType = "none";
+  };
+in
+
+optionsDoc
```
hsjobeki, line 39: the name of the file might be a bit misleading. It doesn't (strictly) return a json.

hsjobeki (changes requested):
Maybe we should not add the docs with pkgs-by-name, but via an overlay or add an exclude from checks since it will always import files from /doc ?

philiptaron:
@hsjobeki said:

> Maybe we should not add the docs with `pkgs-by-name`, but via an overlay or add an exclude from checks since it will always import files from `doc/`?

This is the direction I went. The current `nixpkgs-check-by-name` failure [says this](https://github.com/NixOS/nixpkgs/commit/4f7ce0474c46fbba2de0382d43a68d5c01aae4a7/checks/27487167585/logs):

```
- Attribute `pkgs.nixpkgs-manual` is a new top-level package using `pkgs.callPackage ./pkgs/data/documentation/nixpkgs-manual { /* ... */ }`.
  Please define it in pkgs/by-name/ni/nixpkgs-manual/package.nix instead.
  See `pkgs/by-name/README.md` for more details.
  Since the second `callPackage` argument is `{ }`, no manual `callPackage` in pkgs/top-level/all-packages.nix is needed anymore.

This PR introduces additional instances of discouraged patterns as listed above.
Merging is discouraged but would not break the base branch.
```

This is now ready for review, I think.

pkgs/by-name/ni/nixpkgs-manual/options-json.nix
philiptaron: Renamed.

philiptaron:
There is one thing I could do that would satisfy `nixpkgs-check-by-name` and the existing code: move everything, web assets and all, into `pkgs/by-name/ni/nixpkgs-manual` and make a symlink to that directory replacing the `doc/` directory.

philiptaron:
I missed that `pkgs/tools/nix/web-devmode.nix` was also used by `nixos/doc/manual/shell.nix`.

fricklerhandwerk:
Re-doing the construction to be `callPackage` style is defintely a good thing, but I fear that hiding the sources in `pkgs` will make it pretty hard to find. The manual is pretty important and keeping it in `/doc` would be vital to make it as likely as possible for people to find it on their own.

pkgs/data/documentation/nixpkgs-manual/lib-function-locations.nix
```diff
@@ -2,80 +2,100 @@
...
+  inherit (lib)
+    attrNames
+    concatMap
+    concatStringsSep
+    filter
+    flatten
+    isAttrs
+    length
+    listToAttrs
+    map
```
SuperSandro2000, line 18: map is just available without builtins or lib, we don't need to take it from lib.

pkgs/data/documentation/nixpkgs-manual/options-doc.nix
```diff
@@ -0,0 +1,32 @@
+# NB: This file describes the Nixpkgs manual, which happens to use module docs infra originally developed for NixOS.
+{ lib, nixosOptionsDoc }:
+
+let
+  inherit (lib.strings) hasPrefix removePrefix;
+
+  modules = lib.evalModules {
+    modules = [ ../../../top-level/config.nix ];
```
hsjobeki, line 8: We could use dependency injection for this module and for the root path as well. Less predictions would make it more portable.

eclairevoyant:
> The manual is pretty important and keeping it in `/doc` would be vital to make it as likely as possible for people to find it on their own.

Can we add a symlink? (doc/nixpkgs-manual -> pkgs/.../nixpkgs-manual)

philiptaron:
> > The manual is pretty important and keeping it in `/doc` would be vital to make it as likely as possible for people to find it on their own.
> 
> Can we add a symlink? (doc/nixpkgs-manual -> pkgs/.../nixpkgs-manual) Even the GH web ui can follow symlinks.

This it what I suggested in https://github.com/NixOS/nixpkgs/pull/311459#issuecomment-2229970474 -- and yeah, it's quite possible, I believe.

pkgs/data/documentation/nixpkgs-manual/options-doc.nix
philiptaron: Agree. I'll do that separately, I think.

infinisil:
You have my blessing to ignore the `pkgs/by-name` check failure! This kind of failure won't break master, and it's not worth trying to work around it when it's such a special case.

The real solution imo is to get away from the messy single `pkgs` namespace, which currently contains both packages, but also a whole lot else. But that's for another day :)

doc/doc-support/options-doc.nix
```diff
@@ -0,0 +1,28 @@
...
+    decl:
+    let
+      declStr = toString decl;
+      subpath = lib.removePrefix "/" (lib.removePrefix root declStr);
+    in
+    assert lib.hasPrefix root declStr;
+    {
+      url = "https://github.com/NixOS/nixpkgs/blob/master/${subpath}";
+      name = subpath;
+    };
```
infinisil, line 22: I noticed you changed this to an `assert` when it was this before: https://github.com/NixOS/nixpkgs/blob/6a037089e1d7f14ac4b1d0f560a45d01b7fd0f8e/doc/default.nix#L89-L93

Notably I'd expect this to break manual rendering when building non-Nixpkgs modules using [`documentation.nixos.includeAllModules`](https://search.nixos.org/options?channel=24.05&show=documentation.nixos.includeAllModules&from=0&size=50&sort=relevance&type=packages&query=AllModules)

infinisil (changes requested):
I reviewed all commits, looking good except one thing!

doc/doc-support/options-doc.nix
philiptaron: Yes! I thought that for the nixpkgs manual, there wouldn't be ... non-nixpkgs modules. Am I wrong there?

infinisil: Ah yes you're right! Mixed things up here :)

infinisil (approved):

philiptaron:
Thank you as always @infinisil!

Artturin:
This shows up in the rebuild count by ofborg in unrelated packages https://gist.github.com/GrahamcOfBorg/e02bf2c3d2c6a52e1bfb155f6462a9a4

philiptaron:
> This shows up in the rebuild count by ofborg in unrelated packages https://gist.github.com/GrahamcOfBorg/e02bf2c3d2c6a52e1bfb155f6462a9a4

I'll be able to take a look tomorrow. Likely something is taking a dependency on the whole Nixpkgs tree somewhere.

philiptaron:
OK, I root-caused this rebuild. Step to repro, in the nixpkgs root:

```
git checkout master
nix-build -A nixpkgs-manual
mv result result-on-master
git commit --allow-empty -m "testing"
nix-build -A nixpkgs-manual
nix-diff ./result-on-master ./result
```

Here's the diff:

```diff
- ./result-on-master:{out}
+ ./result:{out}
• The environments do not match:
    buildPhase=''
        /nix/store/k26j1ix7r2sj83f7rdrjjnsw3gb0zfyi-documentation-highlighter/loader.js

      cp -t out ./style.css ./anchor.min.js ./anchor-use.js

      nixos-render-docs manual html \
        --manpage-urls ./manpage-urls.json \
    -   --revision f4f322d1424aa547eba9cb092f905f5ceb9b639c \
    +   --revision 65abf7019bd37b6ad20ea008843ab4a9d316eaa5 \
        --stylesheet style.css \
        --stylesheet highlightjs/mono-blue.css \
        --script ./highlightjs/highlight.pack.js \
        --script ./highlightjs/loader.js \
''
```

The revision is being incorporated with these lines:

https://github.com/NixOS/nixpkgs/blob/4089e292e56c928b9cc6a2700567f9ecf89f3b80/doc/doc-support/package.nix#L61-L65

`lib.trivial.revisionWithDefault` has this contents:

https://github.com/NixOS/nixpkgs/blob/4089e292e56c928b9cc6a2700567f9ecf89f3b80/lib/trivial.nix#L437-L445

Which makes the whole thing make sense. Every git revision makes this rebuild, because the documentation includes the git revision it was built with.

philiptaron:
My proposed fix is here:
- https://github.com/NixOS/nixpkgs/pull/330915

PR #311507, opened 2024-05-14
bash-kernel: fix and add smoke test

thomasjm:
## Description of changes

The `bash-kernel` in Nixpkgs is currently broken due to https://github.com/takluyver/bash_kernel/issues/142.

The underlying issue is that `bash-kernel` uses invisible characters in its prompt, which are currently not being output.

This PR fixes it by using `bashInteractive` instead of `bash`.  I don't totally understand why this fix works--AFAICT, the main difference between the two is that `bashInteractive` has the `readline` library included. In any case, using `bashInteractive` is probably closer to how most people use `bash-kernel`, as it just runs whatever `bash` is on the PATH.

This PR also adds a smoke test, in the form of a test notebook that gets evaluated by Papermill. If this succeeds, we know basic kernel evaluation works.

## Things done

- Built on platform(s)
  - [X] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

CC @zimbatm @teto

### DIFF
--- a/pkgs/development/python-modules/bash-kernel/default.nix
+++ b/pkgs/development/python-modules/bash-kernel/default.nix
@@ -6,7 +6,7 @@
 , ipykernel
 , python
 , pexpect
-, bash
+, bashInteractive
 , substituteAll
 }:
 
@@ -24,7 +24,7 @@ buildPythonPackage rec {
   patches = [
     (substituteAll {
       src = ./bash-path.patch;
-      bash = lib.getExe bash;
+      bash = lib.getExe bashInteractive;
     })
   ];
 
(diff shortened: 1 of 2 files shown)


teto (approved):
Haven't tested.

pkgs/development/python-modules/bash-kernel/default.nix
```diff
@@ -45,8 +45,16 @@ buildPythonPackage rec {
     ${python.pythonOnBuildForHost.interpreter} -m bash_kernel.install --prefix $out
   '';
 
-  # no tests
-  doCheck = false;
+  checkPhase = ''
+    # Create a JUPYTER_PATH with the kernelspec
```
teto, line 49: maybe add the `runHook preCheck` / `runHook postCheck`. Dunno how it could matter but it helps keep some consistency.

pkgs/development/python-modules/bash-kernel/default.nix
```diff
@@ -45,8 +45,16 @@ buildPythonPackage rec {
...
+    export JUPYTER_PATH=$(mktemp -d)
+    mkdir -p $JUPYTER_PATH/kernels/bash
+    echo '{ "language": "bash", "argv": [ "${python}/bin/python", "-m", "bash_kernel", "-f", "{connection_file}" ] }' > $JUPYTER_PATH/kernels/bash/kernel.json
+
+    # Evaluate a test notebook with papermill
+    cd $(mktemp -d)
+    ${python.withPackages (ps: [ps.papermill])}/bin/papermill --kernel bash ${./test.ipynb} out.ipynb
```
teto, line 56: so this would fail without bashInteractive ? 
I wonder if it should go through passtru.tests or in checkPhase since papermill becomes a new dependency ? without your fixes, the kernel still works but with a funky prompt ? (sry I read a bit fast)

thomasjm: The kernel is totally broken without `bashInteractive`. It crashes right away with an error about being unable to parse the prompt.

I don't know about the current best practices regarding `passthru.tests`. It seems like a good idea to always run a smoke check to prevent regressions like the one we currently have.

If we do the `checkPhase` option, Papermill would only be a build-time dependency right? So users getting it from the Nixpkgs cache wouldn't experience any change.

PR #296581, opened 2024-03-17
Minimal set of changes to get FreeBSD stdenv building again

rhelmot:
## Description of changes

- Subset of https://github.com/NixOS/nixpkgs/pull/254801 actually small enough to be reviewable, hopefully

## Things done

- Built on platform(s)
  - [X] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [X] x86_64-freebsd
- For non-Linux: Is sandboxing enabled in `nix.conf`? (See [Nix manual](https://nixos.org/manual/nix/stable/command-ref/conf-file.html))
  - [ ] `sandbox = relaxed`
  - [X] `sandbox = true`
- [ ] Tested, as applicable:
  - [NixOS test(s)](https://nixos.org/manual/nixos/unstable/index.html#sec-nixos-tests) (look inside [nixos/tests](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests))
  - and/or [package tests](https://nixos.org/manual/nixpkgs/unstable/#sec-package-tests)
  - or, for functions and "core" functionality, tests in [lib/tests](https://github.com/NixOS/nixpkgs/blob/master/lib/tests) or [pkgs/test](https://github.com/NixOS/nixpkgs/blob/master/pkgs/test)
  - made sure NixOS tests are [linked](https://nixos.org/manual/nixpkgs/unstable/#ssec-nixos-tests-linking) to the relevant packages
- [ ] Tested compilation of all packages that depend on this change using `nix-shell -p nixpkgs-review --run "nixpkgs-review rev HEAD"`. Note: all changes have to be committed, also see [nixpkgs-review usage](https://github.com/Mic92/nixpkgs-review#usage)
- [ ] Tested basic functionality of all binary files (usually in `./result/bin/`)
- [24.05 Release Notes](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2405.section.md) (or backporting [23.05](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2305.section.md) and [23.11](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2311.section.md) Release notes)
  - [ ] (Package updates) Added a release notes entry if the change is major or breaking
  - [ ] (Module updates) Added a release notes entry if the change is significant
  - [ ] (Module addition) Added a release notes entry if adding a new NixOS module
[...]

PR #311760, opened 2024-05-14
autoconf-archive: Fix build on FreeBSD

rhelmot:
## Description of changes

texinfo is required for building docs on FreeBSD native.

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [x] x86_64-freebsd
  - [ ] `sandbox = relaxed`
  - [x] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/development/tools/misc/autoconf-archive/default.nix
+++ b/pkgs/development/tools/misc/autoconf-archive/default.nix
@@ -1,4 +1,4 @@
-{ lib, stdenv, fetchurl, xz }:
+{ lib, stdenv, fetchurl, xz, texinfo }:
 
 stdenv.mkDerivation rec {
   pname = "autoconf-archive";
@@ -13,6 +13,10 @@ stdenv.mkDerivation rec {
   enableParallelBuilding = true;
 
   buildInputs = [ xz ];
+  # texinfo being required on FreeBSD precipitates from the autotoolsAbspathHook.
+  # by patching the impure uname path on the FreeBSD code path, this triggers a
+  # file-was-modified check which rebuilds some doc information requiring texinfo.
+  nativeBuildInputs = lib.optionals stdenv.buildPlatform.isFreeBSD [ texinfo ];
 
   meta = with lib; {
     description = "Archive of autoconf m4 macros";


Ericson2314:
1. I think we should target `master` where possible. We can merge `master` into `staging` at any time (last I checked).

2. What exactly is the reason for this? build-time conditional deps, or cross vs native divergence, feels rather peculiar to me.

rhelmot:
```
       last 25 log lines:                                                                                                          
       > cd "$am__cwd"; \                                                                                                          
       > if /nix/store/5fxljnxqia0f76xfykzqrr6p60yrkv1c-bash/bin/bash '/build/autoconf-archive-2023.02.20/build-aux/missing' makeinfo   -I . \
       >  -o autoconf-archive.info autoconf-archive.texi; \                                                                        
       > then \                                                                                                                    
       >   rc=0; \                                                                                                                 
       >   CDPATH="${ZSH_VERSION+.}:" && cd .; \                                                                                   
       > else \                                                                                                                    
       >   rc=$?; \                                                                                                                
       >   CDPATH="${ZSH_VERSION+.}:" && cd . && \                                                                                 
       >   $restore $backupdir/* `echo "./autoconf-archive.info" | sed 's|[^/]*$||'`; \                          
       > fi; \                                                                                                                     
       > rm -rf $backupdir; exit $rc                                                                                               
       > /build/autoconf-archive-2023.02.20/build-aux/missing: line 81: makeinfo: command not found              
       > WARNING: 'makeinfo' is missing on your system.                                                                            
       >          You should only need it if you modified a '.texi' file, or                                  
       >          any other file indirectly affecting the aspect of the manual.                                     
       >          You might want to install the Texinfo package:                                                                                                                                                                                                      
       >          <https://www.gnu.org/software/texinfo/>
       >          The spurious makeinfo call might also be the consequence of
       >          using a buggy 'make' (AIX, DU, IRIX), in which case you might
       >          want to install GNU make:
       >          <https://www.gnu.org/software/make/>
       > make[1]: *** [Makefile:325: autoconf-archive.info] Error 127
       > make[1]: Leaving directory '/build/autoconf-archive-2023.02.20/doc'
       > make: *** [Makefile:420: install-recursive] Error 1
       For full logs, run 'nix-store -l /nix/store/3ypq5y2pdmdwh3m0rqjgmf6jivm8hrxm-autoconf-archive-2023.02.20.drv'.
```

I believe this is in turn because of https://github.com/NixOS/nixpkgs/pull/311795/files#diff-4b5da784f51cdf538452c429273232dce5d0155b75dd461ab67c0d7665c07322. We have to change `uname` unqualified to `/nix/store/.../bin/uname`, which triggers the rebuild.

pkgs/development/tools/misc/autoconf-archive/default.nix
```diff
@@ -13,6 +13,7 @@ stdenv.mkDerivation rec {
   enableParallelBuilding = true;
 
   buildInputs = [ xz ];
+  nativeBuildInputs = lib.optionals stdenv.buildPlatform.isFreeBSD [ texinfo ];
```
alyssais, line 16: Why just on FreeBSD?

rhelmot: See conversation with John on the main thread. The unqualified-uname problem which precipitates this is present for a few operating systems, but I believe FreeBSD is the only one we support building from. We _can_ enable this for all platforms and it will be harmless, but I think this signals the mechanism more clearly. I can also add a comment.

Ericson2314: what does `texinfo` have to do with `uname`?

rhelmot: Patching the unqualified uname trips some modified-file check which causes some additional configure time logic which invokes texinfo. See above for the exact error message which explains this.

rhelmot: done

alyssais (approved):
Looks good with a squash.

pkgs/development/tools/misc/autoconf-archive/default.nix
Ericson2314: Ah, it finally clicked what the "modified-file check" means / how that matters. Got it.

Ericson2314:
Sorry one more question, how come the `/usr/bin/uname` isn't a problem normally? Does it depend on whether we need to regenerate `configure`?

Seems like this should be tied to `autoreconfHook` somehow...

rhelmot:
horrifyingly enough, in the autoconf code which is THIS old, some of the invocations of uname look it up via PATH and some hardcode /usr/bin/uname.

rhelmot:
Okay I investigated the consequences of the uname patch some more.

It looks like `autoreconfHook` DOES solve the same problem as the [uname patch hook in stdenv](https://github.com/NixOS/nixpkgs/pull/311795/files#diff-4b5da784f51cdf538452c429273232dce5d0155b75dd461ab67c0d7665c07322) that necessitates the texinfo package being added here, on account of [this very recent autoconf commit](https://git.savannah.gnu.org/cgit/config.git/commit/?id=e4786449e1c26716e3f9ea182caf472e4dbc96e0). I set up a branch that removes the stdenv hook and instead just adds autoreconfHook to everything that will take it, and just does the patch inline for everything else. Unfortunately, due to the nature of stdenv, this was 36 files changed, of which 17 of them were unable to use autoreconfHook. I think we would just be better just doing it the way I was doing it initially, with the stdenv hook. I can add a comment with the nature of this dependency.

alyssais:
Using a hook sounds okay, but it shouldn't be a default stdenv hook, because that'd be next to impossible to remove later.  We can add it as an input to the 17 packages that need it, and then phase it out as their configure scripts are updated.

Ericson2314:
What about `updateAutotoolsGnuConfigScriptsHook`? It always has good support in the existing native stdenvs too. (I think I was confusing `autoreconfHook` and that.)

rhelmot:
Looks like since the linked PR this is no longer needed :)

PR #311763, opened 2024-05-14
cmake: replace propagated ps input with abspath patch

rhelmot:
## Description of changes

Similar to the Darwin case. Should be self-explanatory.

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [x] x86_64-freebsd
  - [ ] `sandbox = relaxed`
  - [x] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/by-name/cm/cmake/package.nix
+++ b/pkgs/by-name/cm/cmake/package.nix
@@ -101,7 +101,7 @@ stdenv.mkDerivation (finalAttrs: {
   ++ lib.optional stdenv.isDarwin CoreServices
   ++ lib.optional (stdenv.isDarwin && !isMinimalBuild) SystemConfiguration;
 
-  propagatedBuildInputs = lib.optional stdenv.isDarwin ps;
+  propagatedBuildInputs = lib.optional (stdenv.isDarwin || stdenv.isFreeBSD) ps;
 
   preConfigure = ''
     fixCmakeFiles .
--- a/pkgs/by-name/cm/cmake/package.nix
+++ b/pkgs/by-name/cm/cmake/package.nix
@@ -1,6 +1,7 @@
 { lib
 , stdenv
 , fetchurl
+, substituteAll
 , buildPackages
 , bzip2
 , curlMinimal
@@ -66,7 +67,13 @@ stdenv.mkDerivation (finalAttrs: {
   # Derived from https://github.com/curl/curl/commit/31f631a142d855f069242f3e0c643beec25d1b51
   ++ lib.optional (stdenv.isDarwin && isMinimalBuild) ./005-remove-systemconfiguration-dep.diff
   # On Darwin, always set CMAKE_SHARED_LIBRARY_RUNTIME_C_FLAG.
-  ++ lib.optional stdenv.isDarwin ./006-darwin-always-set-runtime-c-flag.diff;
+  ++ lib.optional stdenv.isDarwin ./006-darwin-always-set-runtime-c-flag.diff
+  # On platforms where ps is not part of stdenv, patch the invocation of ps to use an absolute path.
+  ++ lib.optional (stdenv.isDarwin || stdenv.isFreeBSD) (
+    substituteAll {
+      src = ./007-darwin-bsd-ps-abspath.diff;
+      ps = lib.getExe ps;
+    });
 
   outputs = [ "out" ] ++ lib.optionals buildDocs [ "man" "info" ];
   separateDebugInfo = true;
@@ -100,8 +107,6 @@ stdenv.mkDerivation (finalAttrs: {
   ++ lib.optional qt5UI qtbase
   ++ lib.optional stdenv.isDarwin CoreServices
   ++ lib.optional (stdenv.isDarwin && !isMinimalBuild) SystemConfiguration;
-
-  propagatedBuildInputs = lib.optional (stdenv.isDarwin || stdenv.isFreeBSD) ps;
 
   preConfigure = ''
     fixCmakeFiles .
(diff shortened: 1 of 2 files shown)


SuperSandro2000 (approved):

pkgs/by-name/cm/cmake/package.nix
```diff
@@ -101,7 +101,7 @@ stdenv.mkDerivation (finalAttrs: {
   ++ lib.optional stdenv.isDarwin CoreServices
   ++ lib.optional (stdenv.isDarwin && !isMinimalBuild) SystemConfiguration;
 
-  propagatedBuildInputs = lib.optional stdenv.isDarwin ps;
+  propagatedBuildInputs = lib.optional (stdenv.isDarwin || stdenv.isFreeBSD) ps;
```
AndersonTorres, line 104: Why not propagate it over Linux too?

alyssais: Propagation is weird and difficult to debug and should be avoided where possible.

What's this used for?  The library or the executable?  If executable, could we wrap instead?

rhelmot: Looks to be both. However, the place it's used is extremely well isolated and unlikely to change (it's a vendored lib), so we can assuredly just patch it to use an absolute path. This can remove the propagation on Darwin as well.

I suspect ps is part of stdenv on linux so it is not needed there, so I will make the patch only apply to Darwin | FreeBSD in order to minimize rebuilds.

pkgs/by-name/cm/cmake/package.nix
```diff
@@ -66,7 +67,13 @@ stdenv.mkDerivation (finalAttrs: {
...
   ++ lib.optional (stdenv.isDarwin && isMinimalBuild) ./005-remove-systemconfiguration-dep.diff
   # On Darwin, always set CMAKE_SHARED_LIBRARY_RUNTIME_C_FLAG.
-  ++ lib.optional stdenv.isDarwin ./006-darwin-always-set-runtime-c-flag.diff;
+  ++ lib.optional stdenv.isDarwin ./006-darwin-always-set-runtime-c-flag.diff
+  # On platforms where ps is not part of stdenv, patch the invocation of ps to use an absolute path.
+  ++ lib.optional (stdenv.isDarwin || stdenv.isFreeBSD) (
+    substituteAll {
+      src = ./007-darwin-bsd-ps-abspath.diff;
+      ps = "${lib.getBin ps}/bin/ps";
+    });
```
alyssais, line 76: This might be more concisely done with a substituteInPlace or sed.

rhelmot: I was under the impression that this was the kosher way to do this, since direct comparisons are hard to debug if the targeted code changes.

alyssais: Since the patch is conditional, doing it this way means that it's quite likely that Linux users will update the package, without realising that the patch no longer applies, meaning the package ends up not even building on FreeBSD and Darwin.

(OfBorg won't help here, because CMake updates have to go to staging, and OfBorg is generally useless on staging because it can never get past building LLVM without timing out.)

rhelmot: isn't not building preferable to failing in mysterious ways down the line?

rhelmot: I can also make it unconditional on linux, but this will be another mass rebuild.

AndersonTorres: A mass rebuild will happen anyway in other machines.

alyssais: > isn't not building preferable to failing in mysterious ways down the line?

That can happen either way, if they start using it elsewhere in the file, for example.

> I can also make it unconditional on linux, but this will be another mass rebuild.

Probably best not to if it would bring in a dependency that doesn't get used.

alyssais: But I suppose even with sed or substituteInPlace we'd have the same risk of it not being caught due to having to do it conditionally, so I guess this is fine.

rhelmot:
I went ahead and made the patch unconditional on all platforms. ps won't be unused on linux, it'll just override the stdenv ps.

rhelmot:
On never mind. It looks like that triggers infinite recursion on linux.

pkgs/by-name/cm/cmake/package.nix
```diff
@@ -66,7 +67,13 @@ stdenv.mkDerivation (finalAttrs: {
   # Derived from https://github.com/curl/curl/commit/31f631a142d855f069242f3e0c643beec25d1b51
   ++ lib.optional (stdenv.isDarwin && isMinimalBuild) ./005-remove-systemconfiguration-dep.diff
   # On Darwin, always set CMAKE_SHARED_LIBRARY_RUNTIME_C_FLAG.
-  ++ lib.optional stdenv.isDarwin ./006-darwin-always-set-runtime-c-flag.diff;
+  ++ lib.optional stdenv.isDarwin ./006-darwin-always-set-runtime-c-flag.diff
+  # On platforms where ps is not part of stdenv, patch the invocation of ps to use an absolute path.
+  ++ lib.optional (stdenv.isDarwin || stdenv.isFreeBSD) (
+    substituteAll {
+      src = ./007-darwin-bsd-ps-abspath.diff;
+      ps = "${lib.getBin ps}/bin/ps";
```
SuperSandro2000, line 75:
```suggestion
      ps = lib.getExe ps;
```

rhelmot: looks good, committed.

alyssais (approved):
Looks good!  The change to propagation means it's possibly a breaking change, which is currently only allowed in staging, so I'll merge this there.

PR #311764, opened 2024-05-14
coreutils: Fix build on native FreeBSD

rhelmot:
## Description of changes

The singleBinary diff is so old and has been tossed through so many commits that it I no longer remember the nature of the horrors that are unleashed without it, but it is required last I checked. The configure diff is explained inline. The comment moving is a formatting silly and can be reverted if it is not appropriate.

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [x] x86_64-freebsd
  - [ ] `sandbox = relaxed`
  - [x] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/tools/misc/coreutils/default.nix
+++ b/pkgs/tools/misc/coreutils/default.nix
@@ -16,7 +16,7 @@
 , minimal ? true
 , withOpenssl ? !minimal, openssl
 , withPrefix ? false
-, singleBinary ? "symlinks" # you can also pass "shebangs" or false
+, singleBinary ? if stdenv.isFreeBSD then false else "symlinks" # you can also pass "shebangs" or false
 }:
 
 # Note: this package is used for bootstrapping fetchurl, and thus cannot use
--- a/pkgs/tools/misc/coreutils/default.nix
+++ b/pkgs/tools/misc/coreutils/default.nix
@@ -99,13 +99,12 @@ stdenv.mkDerivation rec {
   separateDebugInfo = true;
 
   nativeBuildInputs = [
-    # autoreconfHook is due to patch, normally only needed for cygwin
-    autoreconfHook
     perl
     xz.bin
   ]
   ++ optionals stdenv.hostPlatform.isCygwin [
     # due to patch
+    autoreconfHook
     texinfo
   ];
 
--- a/pkgs/os-specific/bsd/freebsd/lib/install-wrapper.sh
+++ b/pkgs/os-specific/bsd/freebsd/lib/install-wrapper.sh
@@ -14,7 +14,6 @@ while (( $# )); do
   case $1 in
     -C) ;;
     -o | -g) shift ;;
-    -s) ;;
     -m | -l)
       # handle next arg so not counted as path arg
       args+=("$1" "$2")


pkgs/tools/misc/coreutils/default.nix
```diff
@@ -16,7 +16,7 @@
 , minimal ? true
 , withOpenssl ? !minimal, openssl
 , withPrefix ? false
-, singleBinary ? "symlinks" # you can also pass "shebangs" or false
+, singleBinary ? if stdenv.isFreeBSD then false else "symlinks" # you can also pass "shebangs" or false
```
alyssais, line 19: Does shebangs work?

rhelmot: looks like it does, thanks for the idea!

pkgs/tools/misc/coreutils/default.nix
```diff
@@ -138,7 +138,10 @@ stdenv.mkDerivation rec {
     # the boot time is set to the epoch because the system has no RTC. We
     # explicitly enable it for cases where it can't be detected automatically,
     # such as when cross-compiling.
-    ++ optional stdenv.hostPlatform.isLinux "gl_cv_have_proc_uptime=yes";
+    ++ optional stdenv.hostPlatform.isLinux "gl_cv_have_proc_uptime=yes"
+    # this function exists in libc but is not provided by any header, leading
+    # to its detection at configure flag and failure at build time, I think
+    ++ optional stdenv.hostPlatform.isFreeBSD "ac_cv_func_random_r=no";
```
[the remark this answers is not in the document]
rhelmot: No, on non-Nix builds this value is detected automatically. I'm really not sure what could cause this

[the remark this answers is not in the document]
alyssais: Might be worth comparing config.log between Nix and non-Nix builds?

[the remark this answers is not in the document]
rhelmot: Without a doubt, chasing this rabbit hole was the most awful experience I've had debugging freebsd/nix so far. Our libc had a bunch of local symbols in it, which are usually removed by stripping. These are what was being picked up by coreutils configure to believe that random_r existed. The reason they weren't being stripped was that *our install wrapper purposefully discarded the flag that would cause it to invoke strip*. The nixpkgs blame goes back to the original init freebsd commit. I can't for the life of me figure out why this would be considering that the same commit specifies that that flag should be passed! I suppose it has fallen to me to debug this weighty error. Everyone say, "Thank you, Atlas". Fix pushed.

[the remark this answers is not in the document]
alyssais: Thank you so much for the wonderful investigation!

alyssais (commented):
Can you rebase to have one commit fixing the strip wrapper, then one commit fixing coreutils?  Then LGTM.

pkgs/tools/misc/coreutils/default.nix
```diff
@@ -98,8 +98,8 @@ stdenv.mkDerivation rec {
   outputs = [ "out" "info" ];
   separateDebugInfo = true;
 
+  # autoreconfHook is due to patch, normally only needed for cygwin
   nativeBuildInputs = [
-    # autoreconfHook is due to patch, normally only needed for cygwin
     autoreconfHook
```
alyssais, line 103: This hunk can be dropped.

pkgs/tools/misc/coreutils/default.nix
rhelmot: Looks like it does not work. FreeBSD doesn't seem to support using interpreted scripts as interpreters, so anything with coreutils/bin/env as the interpreter will fail. I'm debugging symlinks right now to see what the deal is.

rhelmot: Best I can tell, coreutils is busted for the single-binary build, and freebsd just happens to trigger it. [This](https://github.com/coreutils/coreutils/blob/master/src/local.mk#L485) seems to be the root cause - mixing DEPENDENCIES (i.e. make target prerequisites) and LDADD (i.e. flags for ld) is a really bad idea, since suddenly you have -lintl as a make target.

I don't know why shebangs was working two days ago. Maybe suddenly stripping all our libs causes coreutils configure to detect different LDADD flags?

I'm going to revert to singleBinary = false, submit an upstream issue, link to said upstream issue, and move on with my life.

pkgs/tools/misc/coreutils/default.nix
```diff
@@ -99,13 +99,12 @@ stdenv.mkDerivation rec {
...
 
   nativeBuildInputs = [
-    # autoreconfHook is due to patch, normally only needed for cygwin
-    autoreconfHook
     perl
     xz.bin
   ]
   ++ optionals stdenv.hostPlatform.isCygwin [
     # due to patch
+    autoreconfHook
```
rhelmot, line 107: as per https://github.com/NixOS/nixpkgs/pull/311789/files/977ab5018bd37332386fd36af9227b7fa7d42eab#r1601167922

pkgs/tools/misc/coreutils/default.nix
alyssais, line 19: Link to upstream issue, either here or in commit message?

pkgs/tools/misc/coreutils/default.nix
alyssais: Separate commit?  Doesn't really belong in one called "disable single-binary on native FreeBSD".

pkgs/tools/misc/coreutils/default.nix
rhelmot: Is there a coreutils bug tracker other than the mailing list? I emailed them like 12h ago and it still hasn't shown up in their archives for a link.

alyssais: To bug-coreutils@gnu.org?  Did you send it in plain text?

rhelmot: I did

alyssais: I guess it probably got held for moderation/spam review then?  You could mention the message ID and date of the message, and then if it ever does show up it'll be findable.

rhelmot: Looks like it got approved overnight. It's in the commit message now.

alyssais (approved):

PR #311767, opened 2024-05-14
python3: fix build on native FreeBSD

rhelmot:
## Description of changes

- Deterministic uname is required since cpython will otherwise embed the current host kernel version.
- invocations of uname actually need to be modified a little bit in order to create a cpython which, separately from above, is for x86_64-freebsd and not x86_64-freebsd14.
- libxcrypt is explained inline

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [x] x86_64-freebsd
  - [ ] `sandbox = relaxed`
  - [x] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/development/interpreters/python/cpython/default.nix
+++ b/pkgs/development/interpreters/python/cpython/default.nix
@@ -121,6 +121,9 @@ let
     versionOlder
   ;
 
+  # mixes libc and libxcrypt headers and libs and causes segfaults on importing crypt
+  libxcrypt = if stdenv.hostPlatform.isFreeBSD then null else inputs.libxcrypt;
+
   buildPackages = pkgsBuildHost;
   inherit (passthru) pythonOnBuildForHost;
 
@@ -261,6 +264,7 @@ let
 
     multiarch =
       if isDarwin then "darwin"
+      else if isFreeBSD then ""
       else if isWindows then ""
       else "${multiarchCpu}-${machdep}-${pythonAbiName}";
 
@@ -445,7 +449,11 @@ in with passthru; stdenv.mkDerivation (finalAttrs: {
     "LDFLAGS=-static"
   ];
 
-  preConfigure = optionalString (pythonOlder "3.12") ''
+  preConfigure = ''
+    # Attempt to purify some of the host info collection
+    sed -E -i -e 's/uname -r/echo/g' -e 's/uname -n/echo nixpkgs/g' config.guess
+    sed -E -i -e 's/uname -r/echo/g' -e 's/uname -n/echo nixpkgs/g' configure
+  '' + optionalString (pythonOlder "3.12") ''
     # Improve purity
     for path in /usr /sw /opt /pkg; do
       substituteInPlace ./setup.py --replace-warn $path /no-such-path
@@ -476,10 +484,10 @@ in with passthru; stdenv.mkDerivation (finalAttrs: {
   postInstall = let
     # References *not* to nuke from (sys)config files
     keep-references = concatMapStringsSep " " (val: "-e ${val}") ([
-      (placeholder "out") libxcrypt
-    ] ++ optionals tzdataSupport [
-      tzdata
-    ]);
+      (placeholder "out")
+    ] ++ lib.optional (libxcrypt != null) libxcrypt
+      ++ lib.optional tzdataSupport tzdata
+    );
   in lib.optionalString enableFramework ''
     for dir in include lib share; do
       ln -s $out/Library/Frameworks/Python.framework/Versions/Current/$dir $out/$dir
@@ -661,7 +669,7 @@ in with passthru; stdenv.mkDerivation (finalAttrs: {
     '';
     license = licenses.psfl;
     pkgConfigModules = [ "python3" ];
-    platforms = platforms.linux ++ platforms.darwin ++ platforms.windows;
+    platforms = platforms.linux ++ platforms.darwin ++ platforms.windows ++ platforms.freebsd;
     mainProgram = executable;
   };
 })


pkgs/development/interpreters/python/cpython/default.nix
```diff
@@ -469,15 +477,18 @@ in with passthru; stdenv.mkDerivation (finalAttrs: {
   # https://fedoraproject.org/wiki/Changes/PythonNoSemanticInterpositionSpeedup
   optionalString enableNoSemanticInterposition ''
     export CFLAGS_NODIST="-fno-semantic-interposition"
+  '' + optionalString (stdenv.buildPlatform.isFreeBSD && stdenv.hostPlatform.isFreeBSD) ''
+    sed -E -i -e 's/uname -p/uname -m/g' -e 's/uname -r/echo/g' config.guess
+    sed -E -i -e 's/uname -p/uname -m/g' -e 's/uname -r/echo/g' configure
```
alyssais, line 482: If we're patching these anyway, why not just patch them to a static string and not bother with deterministic uname?

pkgs/development/interpreters/python/cpython/default.nix
```diff
@@ -121,6 +122,9 @@ let
     versionOlder
   ;
 
+  # mixes libcc and libxcrypt headers and libs and causes segfautlts on importing crypt
```
alyssais, line 125:
```suggestion
  # mixes libc and libxcrypt headers and libs and causes segfaults on importing crypt
```

alyssais (changes requested):

pkgs/development/interpreters/python/cpython/default.nix
rhelmot: indeed that would be simpler. I've pushed a patch that does this.

pkgs/development/interpreters/python/cpython/default.nix
```diff
@@ -478,8 +475,8 @@ in with passthru; stdenv.mkDerivation (finalAttrs: {
   optionalString enableNoSemanticInterposition ''
     export CFLAGS_NODIST="-fno-semantic-interposition"
   '' + optionalString (stdenv.buildPlatform.isFreeBSD && stdenv.hostPlatform.isFreeBSD) ''
```
alyssais, line 477: Any reason not to make this unconditional?

rhelmot: I'm worried that other code paths for other operating systems won't take kindly to the empty uname -r substitute.

alyssais: Well, uname isn't going to work on them anyway, right?  And we don't have any better value for them.

rhelmot: I was referring to other unix-likes like NetBSD, OpenBSD, and SunOS. I'm totally amenable to simply doing this and seeing what happens, though!

rhelmot:
It looks like you (@alyssais) pushed a typo fix commit at some point, and I just force-pushed over it by accident (did a squash that did not include that commit). Can you reapply that commit from your local pull and squash it with the rest of the things?

alyssais:
I have never pushed to any of your PRs, as far as I remember.

rhelmot:
Ah, it was that you suggested a diff and I applied it with the github interface, which credited you. I've reapplied the diff myself. I believe everything has been addressed now.

pkgs/development/interpreters/python/cpython/default.nix
```diff
@@ -469,15 +474,17 @@ in with passthru; stdenv.mkDerivation (finalAttrs: {
   # https://fedoraproject.org/wiki/Changes/PythonNoSemanticInterpositionSpeedup
   optionalString enableNoSemanticInterposition ''
     export CFLAGS_NODIST="-fno-semantic-interposition"
+    sed -E -i -e 's/uname -r/echo/g' -e 's/uname -n/echo nixpkgs/g' config.guess
+    sed -E -i -e 's/uname -r/echo/g' -e 's/uname -n/echo nixpkgs/g' configure
```
alyssais, line 478: Guessing this is not supposed to be inside the `enableNoSemanticInterposition` branch?

alyssais (changes requested):

pkgs/development/interpreters/python/cpython/default.nix
rhelmot: eep! yes.

alyssais (approved):

alyssais:
Commit message should be a buildable attribute, like "python3: …".  (This will mean OfBorg knows what to build, among other reasons.)

pkgs/development/interpreters/python/cpython/default.nix
```diff
@@ -445,7 +449,11 @@ in with passthru; stdenv.mkDerivation (finalAttrs: {
     "LDFLAGS=-static"
   ];
 
-  preConfigure = optionalString (pythonOlder "3.12") ''
+  preConfigure = ''
+    # Attempt to purify some of the host info collection
+    sed -E -i -e 's/uname -r/echo/g' -e 's/uname -n/echo nixpkgs/g' config.guess
+    sed -E -i -e 's/uname -r/echo/g' -e 's/uname -n/echo nixpkgs/g' configure
```
Ericson2314, line 455: Can we use `updateAutotoolsGnuConfigScriptsHook` per the other PR? (or does it already do that in there?)

pkgs/development/interpreters/python/cpython/default.nix
```diff
@@ -476,8 +484,8 @@ in with passthru; stdenv.mkDerivation (finalAttrs: {
   postInstall = let
     # References *not* to nuke from (sys)config files
     keep-references = concatMapStringsSep " " (val: "-e ${val}") ([
-      (placeholder "out") libxcrypt
-    ] ++ optionals tzdataSupport [
+      (placeholder "out") ] ++ ((optionals (libxcrypt != null)) [libxcrypt])
+    ++ optionals tzdataSupport [
       tzdata
     ]);
```
Ericson2314, line 490:
```suggestion
      (placeholder "out")
    ] ++ lib.optional (libxcrypt != null) libxcrypt
      ++ lib.optional tzdataSupport tzdata
    );
```

pkgs/development/interpreters/python/cpython/default.nix
Ericson2314: Oh this is different

pkgs/development/interpreters/python/cpython/default.nix
```diff
@@ -121,6 +121,9 @@ let
     versionOlder
   ;
 
+  # mixes libc and libxcrypt headers and libs and causes segfaults on importing crypt
+  libxcrypt = if stdenv.hostPlatform.isFreeBSD then null else inputs.libxcrypt;
```
Ericson2314, line 125: If this is an bona fide optional dependency, it would be more idiomatic to make a `withLibxcrypt ? libxcrypt != null && !stdenv.hostPlatform.isFreeBSD` flag, and change the logic to refer to that.

rhelmot: This is not a bona fide optional dependency - freebsd libc provides xcrypt, and so it's just not needed on this platform. I don't think we can replace libxcrypt with libc at top-level, just because pretty much every (?) other package seems to want or at least be okay with the separate libxcrypt.

Ericson2314: OK, fair enough!

Ericson2314:
I think this is "eval" error is spurious. The error is, after all, nothing in the code, but just the daemon socket going away.

PR #311770, opened 2024-05-14
expect: Fix compilation on FreeBSD

rhelmot:
## Description of changes

- Enable-cross-compilation.patch hardcodes several linux defaults. Vendor the patch and improve so it does not do this.

- Make the darwin compat patch also help out for FreeBSD

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [x] x86_64-freebsd
  - [ ] `sandbox = relaxed`
  - [x] `sandbox = true`
- [ ] Tested, as applicable:

---


alyssais:
> add a new patch, link-unversioned, which fixes an issue with some software expecting libexpect.so when only libexpect.so.1 is present, or possibly vice versa? It has been so long that the precise nature of this beast is lost to me.

This looks really weird, and I'd like to understand why it only affects FreeBSD, so it'd be great to have more information on this.

pkgs/tools/misc/expect/default.nix
```diff
@@ -29,12 +30,17 @@ tcl.mkTclDerivation rec {
...
+  ] ++ lib.optionals (stdenv.isFreeBSD) [
+    ./link-unversioned.patch
   ];
 
   postPatch = ''
     sed -i "s,/bin/stty,$(type -p stty),g" configure.in
   '';
 
+  # software doesn't support clang16 yet...
+  env.NIX_CFLAGS_COMPILE = "-Wno-implicit-function-declaration -Wno-implicit-int";
```
alyssais, line 42: Has this been reported upstream?

pkgs/tools/misc/expect/fix-darwin-clang16.patch
```diff
@@ -14,14 +14,16 @@ diff -ur a/exp_win.c b/exp_win.c
...
  
  */
  
 -#include <pty.h> /* openpty */
 +/* openpty */
-+#ifdef __APPLE__
++#if defined(__APPLE__)
 +#include <util.h>
++#elif defined(__FreeBSD__)
++#include <termios.h>
```
alyssais, line 26: Could this be upstreamed?  If not, the patch file should at least be renamed.

pkgs/tools/misc/expect/default.nix
```diff
@@ -9,11 +9,12 @@ tcl.mkTclDerivation rec {
     hash = "sha256-Safag7C92fRtBKBN7sGcd2e7mjI+QMR4H4nK92C5LDQ=";
   };
 
-  patches = [
+  patches = lib.optionals (!stdenv.isFreeBSD) [
     (fetchpatch {
       url = "https://raw.githubusercontent.com/buildroot/buildroot/c05e6aa361a4049eabd8b21eb64a34899ef83fc7/package/expect/0001-enable-cross-compilation.patch";
       hash = "sha256-yyzE0Jjac5qaj7Svn4VpMiAqSNLYrw7VZbtFqgMVncs=";
     })
+  ] ++ [
```
alyssais, line 17: Generally, conditionally-applied patches are bad news, but looking at the contents of this patch, it also looks like bad news, so I guess it's fine but I wonder if there's a better way we could do this…

If we are to make it conditional, could it at least be moved after the unconditional ones?

pkgs/tools/misc/expect/default.nix
```diff
@@ -29,12 +30,17 @@ tcl.mkTclDerivation rec {
     })
     # Include `sys/ioctl.h` and `util.h` on Darwin, which are required for `ioctl` and `openpty`.
     ./fix-darwin-clang16.patch
```
SuperSandro2000, line 32:
```suggestion
    ./fix-bsd-clang16.patch
```

SuperSandro2000:
@alyssais upstream is really dead for years.
There is another patch in https://sourceforge.net/p/expect/patches/24/ we could apply which maybe fixes the implicit function declaration
Maybe we can also borrow patches from fedora https://src.fedoraproject.org/rpms/expect/tree/rawhide

but I doubt that upstreaming here will be a fruitful way.

alyssais:
Got it.  In that case applying patches from other distros would be preferable than patching ourselves or doing wrapper hacks, but either is fine.

rhelmot:
The patches you linked are actually already being applied - they are 0001-Add-prototype-to-function-definitions.patch and expect-configure-c99.patch. None of the other patches in the fedora repo seem to matter for this issue. I'm fine with replacing things with other distros' patches, but I'll need someone to point me to some archives to look through. I have already looked at the freebsd-ports tree - all of its patches are too broad to apply.

rhelmot:
@alyssais it seems I was misremembering the nature of the bug - without the unversioned patch, I get this on build:

```
error: builder for '/nix/store/979mi789mfdfzd2xrjh8nqhws5q51bs4-expect-5.45.4.drv' failed with exit code 2;
       last 25 log lines:
       > ^
       > exp_event.c:335:1: warning: a function definition without a prototype is deprecated in all versions of C and is not supported in C2x [-Wdeprecated-non-prototype]
       > exp_event_exit_real(interp)
       > ^
       > 15 warnings generated.
       > rm -f libexpect.so.1
       > clang -shared -o libexpect.so.1 exp_command.o expect.o exp_inter.o exp_regexp.o exp_tty.o exp_log.o exp_main_sub.o exp_pty.o exp_trap.o exp_strf.o exp_console.o exp_glob.o exp_win.o exp_clib.o exp_closetcl.o exp_memmove.o exp_tty_comm.o exp_chan.o Dbg.o pty_termios.o exp_select.o exp_event.o -lutil -lm -rpath /nix/store/fln0d0kdb2wnf0prf2fcbi5zy0wcgyp8-expect-5.45.4/lib -L/nix/store/34q2qrvmch594n0w8a5i3rdi7q7z9362-tcl-8.6.13/lib -ltclstub8.6
       > : libexpect.so.1
       > clang -DPACKAGE_NAME=\"expect\" -DPACKAGE_TARNAME=\"expect\" -DPACKAGE_VERSION=\"5.45.4\" -DPACKAGE_STRING=\"expect\ 5.45.4\" -DPACKAGE_BUGREPORT=\"\" -DPACKAGE_URL=\"\" -DBUILD_expect=/\*\*/ -DHAVE_STDIO_H=1 -DHAVE_STDLIB_H=1 -DHAVE_STRING_H=1 -DHAVE_INTTYPES_H=1 -DHAVE_STDINT_H=1 -DHAVE_STRINGS_H=1 -DHAVE_SYS_STAT_H=1 -DHAVE_SYS_TYPES_H=1 -DHAVE_UNISTD_H=1 -DHAVE_SYS_TIME_H=1 -DSTDC_HEADERS=1 -DNO_VALUES_H=1 -DHAVE_LIMITS_H=1 -DHAVE_SYS_PARAM_H=1 -DUSE_THREAD_ALLOC=1 -D_REENTRANT=1 -D_THREAD_SAFE=1 -DTCL_THREADS=1 -DMODULE_SCOPE=extern\ __attribute__\(\(__visibility__\(\"hidden\"\)\)\) -DTCL_WIDE_INT_IS_LONG=1 -DUSE_TCL_STUBS=1 -DUSE_TCL_STUBS=1 -DHAVE_SYS_SELECT_H=1 -DHAVE_SYS_FCNTL_H=1 -DRETSIGTYPE=void -DHAVE_SYS_TIME_H=1 -DTIME_WITH_SYS_TIME=1 -DHAVE_STRUCT_TM_TM_ZONE=1 -DHAVE_TM_ZONE=1 -DHAVE_GMTIME_R=1 -DHAVE_LOCALTIME_R=1 -DHAVE_TM_GMTOFF=1 -DNO_UNION_WAIT=1 -DHAVE_OPENPTY=1 -DSTTY_READS_STDOUT=1 -DHAVE_MEMMOVE=1 -DHAVE_SYSCONF=1 -DHAVE_STRFTIME=1 -DHAVE_STRCHR=1 -DHAVE_TIMEZONE=1 -DHAVE_SIGLONGJMP=1 -DHAVE_MEMCPY=1 -DWNOHANG_BACKUP_VALUE=1 -DSELECT_MASK_TYPE=fd_set -DRETSIGTYPE=void -DHAVE_TCSETATTR=1 -DPOSIX=1 -DHAVE_TERMIOS=1 -DHAVE_TIOCGWINSZ_IN_TERMIOS_H=1 -DHAVE_SV_TIMEZONE=1 -DHAVE_LONG_FILE_NAMES=1 -DTCL_DEBUGGER -DUSE_NON_CONST -DSCRIPTDIR=\"/nix/store/fln0d0kdb2wnf0prf2fcbi5zy0wcgyp8-expect-5.45.4/lib/expect5.45.4\" -DEXECSCRIPTDIR=\"/nix/store/fln0d0kdb2wnf0prf2fcbi5zy0wcgyp8-expect-5.45.4/lib/expect5.45.4\" -DSTTY_BIN=\"/nix/store/23l4dqky4ywywlsbma2ii6c7yl52cl0y-coreutils/bin/stty\" -DDFLT_STTY="\"sane\"" -I. -I"." -I"/nix/store/34q2qrvmch594n0w8a5i3rdi7q7z9362-tcl-8.6.13/include"    -pipe  -O2 -Wall -fPIC  -c `echo exp_main_exp.c` -o exp_main_exp.o
       > exp_main_exp.c:20:1: warning: a function definition without a prototype is deprecated in all versions of C and is not supported in C2x [-Wdeprecated-non-prototype]
       > main(argc, argv)
       > ^
       > 1 warning generated.
       > clang \
       >       -pipe  -O2 -Wall -fPIC  \
       >        \
       >   -o expect exp_main_exp.o \
       >     -L/build/expect5.45.4 -lexpect5454 \
       >   -L/nix/store/34q2qrvmch594n0w8a5i3rdi7q7z9362-tcl-8.6.13/lib -ltcl8.6 \
       >        -ldl  -lm \
       >    "-Wl,-rpath,/nix/store/fln0d0kdb2wnf0prf2fcbi5zy0wcgyp8-expect-5.45.4/lib" \
       >   -Wl,-rpath,/nix/store/fln0d0kdb2wnf0prf2fcbi5zy0wcgyp8-expect-5.45.4/lib/expect5.45.4:/nix/store/34q2qrvmch594n0w8a5i3rdi7q7z9362-tcl-8.6.13/lib
       > /nix/store/9jwffv1bqg56m7wksfv1658xlv5agq0c-binutils/bin/ld: cannot find -lexpect5454: No such file or directory
       > clang-16: error: linker command failed with exit code 1 (use -v to see invocation)
       > make: *** [Makefile:391: expect] Error 1
```

I spent a solid hour trying to root cause this issue and as far as I can tell some combination of the original code being broken and the patches we've added clashing with the freebsd setup is producing this. I can't find any sort of way to patch this more generically and I can't find any sort of patch from the freebsd ports tree to help out.

pkgs/tools/misc/expect/default.nix
```diff
@@ -29,8 +24,15 @@ tcl.mkTclDerivation rec {
...
-    ./fix-darwin-clang16.patch
+    # Include `termios.h` on FreeBSD for `openpty`
+    ./fix-darwin-bsd-clang16.patch
+  ] ++ lib.optionals (!stdenv.isFreeBSD) [
+    (fetchpatch {
+      url = "https://raw.githubusercontent.com/buildroot/buildroot/c05e6aa361a4049eabd8b21eb64a34899ef83fc7/package/expect/0001-enable-cross-compilation.patch";
+      hash = "sha256-yyzE0Jjac5qaj7Svn4VpMiAqSNLYrw7VZbtFqgMVncs=";
+    })
   ] ++ lib.optionals (stdenv.isFreeBSD) [
+    # a nightmare, truly. context: https://github.com/NixOS/nixpkgs/pull/311770#issuecomment-2113440745
```
alyssais, line 35: Links rot.  There should be at least some explanation inline.

pkgs/tools/misc/expect/default.nix
```diff
@@ -28,13 +24,25 @@ tcl.mkTclDerivation rec {
...
     })
     # Include `sys/ioctl.h` and `util.h` on Darwin, which are required for `ioctl` and `openpty`.
-    ./fix-darwin-clang16.patch
+    # Include `termios.h` on FreeBSD for `openpty`
+    ./fix-darwin-bsd-clang16.patch
+  ] ++ lib.optionals (!stdenv.isFreeBSD) [
+    (fetchpatch {
+      url = "https://raw.githubusercontent.com/buildroot/buildroot/c05e6aa361a4049eabd8b21eb64a34899ef83fc7/package/expect/0001-enable-cross-compilation.patch";
+      hash = "sha256-yyzE0Jjac5qaj7Svn4VpMiAqSNLYrw7VZbtFqgMVncs=";
+    })
```
alyssais, line 33: So on FreeBSD it cross compiles fine without this patch?  Does that mean we don't need it (and can use the other workarounds) for other platforms too?

alyssais (changes requested):

pkgs/tools/misc/expect/default.nix
rhelmot: I doubt this is tenable - the patch hardcodes some linux defaults. It can probably be replaced by moving those defaults into the nix code, but we would need separate defaults for each target system we're cross compiling for.

alyssais: Got it.  I wonder if it should be stdenv.isLinux, then.  (I don't think Darwin->Darwin cross works anyway, but it sounds like this would not be helpful in that situation either…)

rhelmot: Can you or someone else with access to a darwin machine test the native build with this diff before I commit it?

alyssais: I don't have easy access.  (I run Linux on a MacBook but Nix does not agree with my macOS installation.)  @Ericson2314 do you?

Ericson2314: It seems to me like we are making FreeBSD cross compilation of this package more broken than it was before with this one change?

> It can probably be replaced by moving those defaults into the nix code, but we would need separate defaults for each target system we're cross compiling for.

That sounds like a good strategy to me.

rhelmot: I think I've fixed it. it looked like there was only one piece which was really bad for FreeBSD, and it required another patch to configure.in because the previous patch bypassed the normal autoconf mechanisms.

Ericson2314: Is moving the order of the patches intentional / part of the fix?

Ericson2314:
I am currently fixing up the patches

Ericson2314:
OK I think this is good now. I cross compiled this (copy the package to a new directory on `master` to avoid mass rebuild) to FreeBSD and build for Linux, and it built fine in both. I had no need for the library name hack thing, so I just removed it.

@alyssais or @SuperSandro2000, do you want to do a final review then?

rhelmot:
The library name hack is for native freebsd, and I posted at length about not being able to figure out why it was necessary. I'm going to try again one more time to see if I can hack around it in stdenv-0 and have it magically be fixed in the final stdenv, but I'm not optimistic.

Ericson2314:
Yes per

> I spent a solid hour trying to root cause this issue and as far as I can tell some combination of the original code being broken *and the patches we've added clashing with the freebsd setup* is producing this.

I am crossing my fingers that it is the latter, and the problem is magically resolved.

rhelmot:
Well... I can't say I understand it even a little bit, but everything builds fine with this version of the PR. Review away!

rhelmot:
Actually, it is _not_ fixed. I must have done a merge with an out of date branch when I was testing previously, but it's really not working now:

```
/nix/store/q4avkpjcp6ka9vqbm3n282zzxyknpyy8-binutils/bin/ld: cannot find -lexpect5454: No such file or directory
```

It's simply linking a file that does not exist and is not scheduled to be built. I really believe that the makefiles are simply busted for freebsd, since from my intensive reverse engineering it just seemed like everything was working as intended.

Ericson2314 (approved):
OK I think we should do this as is and then do the other as a temp workaround, either as an additional commit in this PR or a separate PR or part of another PR. That will make undoing it once we have a proper fix easier.

Also that patch broke cross compilation for me.

rhelmot:
I think I found a patch which can be applied unconditionally. There's some code which is missing in the Linux and Darwin configure paths but is present in, uh, most other operating systems which seems to produce this exact effect for some reason. I'm going to test it native on FreeBSD and Linux and cross on FreeBSD.

rhelmot:
It works :)

Ericson2314 (approved):
Hooray!!

PR #311773, opened 2024-05-14
gnome.gnome-keyring: support cross compilation

uninsane:
## Description of changes

built (and deployed) from an x86_64-linux machine:
- `nix build '.#pkgsCross.aarch64-multiplatform.gnome.gnome-keyring'`
- `nix build '.#gnome.gnome-keyring'`

the `LIBGCRYPT_CONFIG` flag is lifted from the [libgnome-keyring](https://github.com/NixOS/nixpkgs/blob/d1414f543e40172dfd20e5a42ccd8c9108219693/pkgs/development/libraries/libgnome-keyring/default.nix#L25) nix package.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/desktops/gnome/core/gnome-keyring/default.nix
+++ b/pkgs/desktops/gnome/core/gnome-keyring/default.nix
@@ -10,7 +10,6 @@
 , libxslt
 , gettext
 , gcr
-, autoreconfHook
 , libcap_ng
 , libselinux
 , p11-kit
@@ -37,8 +36,6 @@ stdenv.mkDerivation rec {
     pkg-config
     gettext
     libxslt
-    # Upstream uses ancient autotools to pre-generate the scripts.
-    autoreconfHook
     docbook-xsl-nons
     docbook_xml_dtd_43
     wrapGAppsHook3
@@ -63,6 +60,10 @@ stdenv.mkDerivation rec {
     # gnome-keyring doesn't build with ssh-agent by default anymore, we need to
     # switch to using gcr https://github.com/NixOS/nixpkgs/issues/140824
     "--enable-ssh-agent"
+    # cross compilation requires these paths to be explicitly declared:
+    "LIBGCRYPT_CONFIG=${lib.getExe' (lib.getDev libgcrypt) "libgcrypt-config"}"
+    "SSH_ADD=${lib.getExe' openssh "ssh-add"}"
+    "SSH_AGENT=${lib.getExe' openssh "ssh-agent"}"
   ];
 
   # Tends to fail non-deterministically.


pkgs/desktops/gnome/core/gnome-keyring/default.nix
```diff
@@ -72,6 +75,7 @@ stdenv.mkDerivation rec {
 
   postPatch = ''
     patchShebangs build
+    addToSearchPath ACLOCAL_PATH ${lib.getDev libgcrypt}/share/aclocal
```
jtojnar, line 78: Why is this not handled by the automake setup hook?

pkgs/desktops/gnome/core/gnome-keyring/default.nix
```diff
@@ -63,6 +63,9 @@ stdenv.mkDerivation rec {
     # gnome-keyring doesn't build with ssh-agent by default anymore, we need to
     # switch to using gcr https://github.com/NixOS/nixpkgs/issues/140824
     "--enable-ssh-agent"
+    "LIBGCRYPT_CONFIG=${lib.getExe' (lib.getDev libgcrypt) "libgcrypt-config"}"
```
jtojnar, line 66: Please mention these are for cross in the comment or they will be likely removed in the future.

pkgs/desktops/gnome/core/gnome-keyring/default.nix
uninsane: if i understand it right, `libgcrypt` includes m4 macros in its `dev` output which are used during build time. the `automake` hook populates `ACLOCAL_PATH` for this kind of thing, but [only for packages in `nativeBuildInputs`](https://github.com/NixOS/nixpkgs/blob/bfb7a882678e518398ce9a31a881538679f6f092/pkgs/development/tools/misc/automake/setup-hook.sh#L5) (slight simplification).

other ways to fix this is:
1. place `libgcrypt` in both `buildInputs` and `nativeBuildInputs`.
2. tweak that automake setup-hook to populate ACLOCAL_PATH also from the `$targetOffset` (?) in addition to `$hostOffset`, and keep libgcrypt in just `buildInputs`.
3. build gnome-keyring in such a way that it doesn't need libgcrypt's m4 file.

i suspect most users are doing option 1. it's pretty wasteful in terms of compute/builder resources, but i suppose it's more maintainable than manually setting ACLOCAL_PATH here.

uninsane: oh, or we can just avoid this altogether by not invoking autoconf and using their vendored `configure` instead 🧠

uninsane:
i was able to clean up this patch a bit by removing `autoreconfHook`. rebuilt for native x86_64-linux and cross aarch64-multiplatform, seems to be working just as well.

JohnRTitor:
@ofborg build gnome.gnome-keyring libgnome-keyring

JohnRTitor (approved):
Builds for me.

jtojnar (approved):
I compared the outputs of normal build with and without `autoreconfHook` with diffoscope and diffed the build logs but there do not appear to be any significant differences.

```diff
---gk-sans-arh.log
+++gk-avec-arh.log
@@ -118,7 +272,8 @@
 checking for pid_t... yes
 checking return type of signal handlers... void
 checking for size_t... yes
-checking for uid_t in sys/types.h... yes
+checking for uid_t... yes
+checking for gid_t... yes
 checking for glib-2.0 >= 2.44.0... yes
 checking for gmodule-no-export-2.0... yes
 checking for glib-2.0 gobject-2.0... yes
@@ -131,6 +286,7 @@
 checking for ld used by gcc... ld -m elf_x86_64
 checking if the linker (ld -m elf_x86_64) is GNU ld... yes
 checking for shared library run path origin... done
+checking for egrep -e... (cached) /nix/store/28gpmx3z6ss3znd7fhmrzmvk3x5lnfbk-gnugrep-3.11/bin/grep -E
 checking for CFPreferencesCopyAppValue... no
 checking for CFLocaleCopyCurrent... no
 checking for GNU gettext in libc... yes
```

With the xz debacle we might be a bit wary of using pre-generated files but `autoreconfHook` does not really make us safe anyway.

PR #211230, opened 2023-01-17
llvmPackages: use libcxxrt on FreeBSD

alyssais:
###### Description of changes

FreeBSD doesn't use LLVM's cxxabi implementation, for backwards compatibility reasons.  Software expects the libcxxrt API when building on FreeBSD.  This fixes the build of `pkgsCross.x86_64-freebsd.boost`.

###### Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
- [ ] For non-Linux: Is `sandbox = true` set in `nix.conf`? (See [Nix manual](https://nixos.org/manual/nix/stable/command-ref/conf-file.html))
- [ ] Tested, as applicable:
  - [NixOS test(s)](https://nixos.org/manual/nixos/unstable/index.html#sec-nixos-tests) (look inside [nixos/tests](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests))
  - and/or [package tests](https://nixos.org/manual/nixpkgs/unstable/#sec-package-tests)
  - or, for functions and "core" functionality, tests in [lib/tests](https://github.com/NixOS/nixpkgs/blob/master/lib/tests) or [pkgs/test](https://github.com/NixOS/nixpkgs/blob/master/pkgs/test)
  - made sure NixOS tests are [linked](https://nixos.org/manual/nixpkgs/unstable/#ssec-nixos-tests-linking) to the relevant packages
- [ ] Tested compilation of all packages that depend on this change using `nix-shell -p nixpkgs-review --run "nixpkgs-review rev HEAD"`. Note: all changes have to be committed, also see [nixpkgs-review usage](https://github.com/Mic92/nixpkgs-review#usage)
- [ ] Tested basic functionality of all binary files (usually in `./result/bin/`)
- [23.05 Release Notes (or backporting 22.11 Release notes)](https://github.com/NixOS/nixpkgs/blob/master/CONTRIBUTING.md#generating-2305-release-notes)
  - [ ] (Package updates) Added a release notes entry if the change is major or breaking
  - [ ] (Module updates) Added a release notes entry if the change is significant
  - [ ] (Module addition) Added a release notes entry if adding a new NixOS module
  - [ ] (Release notes changes) Ran `nixos/doc/manual/md-to-db.sh` to update generated release notes
- [x] Fits [CONTRIBUTING.md](https://github.com/NixOS/nixpkgs/blob/master/CONTRIBUTING.md).

PR #311777, opened 2024-05-14
llvmPackages: add support for native FreeBSD

rhelmot:
## Description of changes

Tests were done with llvmPackages_16

* asan doesn't work quite right, disable it
* support for 32-bit FreeBSD uses by default logic tuned for a very old FreeBSD; fix that
* Some llvm tests fail for the same reason as MacOS; disable them
* libcxx no longer needs a FreeBSD special case
* add a flag for the inclusion of terminfo, which helps compile static bootstrap files

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [x] x86_64-freebsd
  - [ ] `sandbox = relaxed`
  - [x] `sandbox = true`
- [ ] Tested, as applicable:
- [x] Fits [CONTRIBUTING.md](https://github.com/NixOS/nixpkgs/blob/master/CONTRIBUTING.md).]

---

### DIFF
--- a/pkgs/development/compilers/llvm/16/compiler-rt/freebsd-i386.patch
+++ b/pkgs/development/compilers/llvm/16/compiler-rt/freebsd-i386.patch
@@ -1,0 +1,21 @@
+--- a/lib/builtins/fp_lib.h	1969-12-31 16:00:01.000000000 -0800
++++ b/lib/builtins/fp_lib.h	2023-12-21 23:39:36.066927293 -0800
+@@ -26,18 +26,6 @@
+ #include <stdbool.h>
+ #include <stdint.h>
+ 
+-// x86_64 FreeBSD prior v9.3 define fixed-width types incorrectly in
+-// 32-bit mode.
+-#if defined(__FreeBSD__) && defined(__i386__)
+-#include <sys/param.h>
+-#if __FreeBSD_version < 903000 // v9.3
+-#define uint64_t unsigned long long
+-#define int64_t long long
+-#undef UINT64_C
+-#define UINT64_C(c) (c##ULL)
+-#endif
+-#endif
+-
+ #if defined SINGLE_PRECISION
+ 
+ typedef uint16_t half_rep_t;
--- a/pkgs/development/compilers/llvm/common/compiler-rt/default.nix
+++ b/pkgs/development/compilers/llvm/common/compiler-rt/default.nix
@@ -128,11 +128,17 @@ stdenv.mkDerivation ({
   '') + ''
     substituteInPlace lib/builtins/int_util.c \
       --replace "#include <stdlib.h>" ""
+  '' + (if stdenv.hostPlatform.isFreeBSD then
+    # As per above, but in FreeBSD assert is a macro and simply allowing it to be implicitly declared causes Issues!!!!!
+    ''
+    substituteInPlace lib/builtins/clear_cache.c lib/builtins/cpu_model.c \
+      --replace "#include <assert.h>" "#define assert(e) ((e)?(void)0:__assert(__FUNCTION__,__FILE__,__LINE__,#e))"
+    '' else ''
     substituteInPlace lib/builtins/clear_cache.c \
       --replace "#include <assert.h>" ""
     substituteInPlace lib/builtins/cpu_model${lib.optionalString (lib.versionAtLeast version "18") "/x86"}.c \
       --replace "#include <assert.h>" ""
-  '');
+  ''));
 
   # Hack around weird upsream RPATH bug
   postInstall = lib.optionalString (stdenv.hostPlatform.isDarwin) ''
--- a/pkgs/development/compilers/llvm/common/libcxx/default.nix
+++ b/pkgs/development/compilers/llvm/common/libcxx/default.nix
@@ -13,11 +13,15 @@
 , python3
 , fixDarwinDylibNames
 , version
-, cxxabi ? if stdenv.hostPlatform.isFreeBSD then libcxxrt else null
+, cxxabi ? null
 , libcxxrt
 , libunwind
 , enableShared ? !stdenv.hostPlatform.isStatic
 }:
+
+# note: our setup using libcxxabi instead of libcxxrt on FreeBSD diverges from
+# normal FreeBSD. This may cause issues with binary patching down the line.
+# If this becomes an issue, try adding as symlink libcxxrt.so -> libc++abi.so
 
 # external cxxabi is not supported on Darwin as the build will not link libcxx
 # properly and not re-export the cxxabi symbols into libcxx
--- a/pkgs/development/compilers/llvm/common/llvm/default.nix
+++ b/pkgs/development/compilers/llvm/common/llvm/default.nix
@@ -125,22 +125,22 @@ stdenv.mkDerivation (rec {
     substituteInPlace cmake/modules/AddLLVM.cmake \
       --replace 'set(_install_name_dir INSTALL_NAME_DIR "@rpath")' "set(_install_name_dir)" \
       --replace 'set(_install_rpath "@loader_path/../''${CMAKE_INSTALL_LIBDIR}''${LLVM_LIBDIR_SUFFIX}" ''${extra_libdir})' ""
-  '' + (optionalString (lib.versionAtLeast release_version "15") (''
-
+  '' +
     # As of LLVM 15, marked as XFAIL on arm64 macOS but lit doesn't seem to pick
     # this up: https://github.com/llvm/llvm-project/blob/c344d97a125b18f8fed0a64aace73c49a870e079/llvm/test/MC/ELF/cfi-version.ll#L7
+    (optionalString (lib.versionAtLeast release_version "15") (''
     rm test/MC/ELF/cfi-version.ll
 
+  '' +
     # This test tries to call `sw_vers` by absolute path (`/usr/bin/sw_vers`)
     # and thus fails under the sandbox:
-  '' + (if lib.versionAtLeast release_version "16" then ''
+    (if lib.versionAtLeast release_version "16" then ''
     substituteInPlace unittests/TargetParser/Host.cpp \
       --replace '/usr/bin/sw_vers' "${(builtins.toString darwin.DarwinTools) + "/bin/sw_vers" }"
   '' else ''
     substituteInPlace unittests/Support/Host.cpp \
       --replace '/usr/bin/sw_vers' "${(builtins.toString darwin.DarwinTools) + "/bin/sw_vers" }"
-  '') + optionalString (lib.versionAtLeast release_version "16") ''
-
+  '') +
     # This test tries to call the intrinsics `@llvm.roundeven.f32` and
     # `@llvm.roundeven.f64` which seem to (incorrectly?) lower to `roundevenf`
     # and `roundeven` on macOS and FreeBSD.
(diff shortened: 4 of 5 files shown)


pkgs/development/compilers/llvm/common/libcxx/default.nix
```diff
@@ -13,7 +13,7 @@
 , python3
 , fixDarwinDylibNames
 , version
-, cxxabi ? if stdenv.hostPlatform.isFreeBSD then libcxxrt else null
+, cxxabi ? null
```
annaleeleaves, line 16: not sure if it matters, but to get `libcxxrt` working correctly i had to add `-DLIBCXX_ENABLE_NEW_DELETE_DEFINITIONS=ON` to the cxxCMakeFlags  https://libcxx.llvm.org/BuildingLibcxx.html#using-libcxxrt-on-linux to get freebsd cross and linux pkgsLLVM to work for various packages that used c++.

rhelmot: I can't speak for cross, but I was just able to compile gmp-with-cxx on a freebsd machine using the stdenv derived from these changes.

pkgs/development/compilers/llvm/common/llvm/default.nix
```diff
@@ -381,6 +393,8 @@ stdenv.mkDerivation (rec {
         nativeInstallFlags
       ])
     )
+  ] ++ optionals (!enableTerminfo) [
+    "-DLLVM_ENABLE_TERMINFO=OFF"
```
rrbutani, line 397: Not opposed to this, just curious: what required this?

pkgs/development/compilers/llvm/16/compiler-rt/freebsd-i386.patch
```diff
@@ -0,0 +1,21 @@
+--- a/lib/builtins/fp_lib.h	1969-12-31 16:00:01.000000000 -0800
```
rrbutani, line 1: Apologies if I'm missing something obvious: do these two patch files need to be specified in `llvm/16/default.nix`?

rhelmot: Yes... they do need to be specified. It seems they haven't been, which means they're not strictly required for the native FreeBSD stdenv, but will be required for other packages at some point. I'll fix this...

pkgs/development/compilers/llvm/common/libcxx/default.nix
annaleeleaves: i was under the impression that libcxxrt was required for freebsd https://github.com/NixOS/nixpkgs/pull/211230 tho if it isn't it would be nice to get rid of it at some point -- it's been a minor pain point when working on libcxx (i tried to remove it a few months ago).

pkgs/development/compilers/llvm/common/llvm/default.nix
rhelmot: By plunging into the icy depths of the commit history from several branches ago, I was able to determine this was added while I was trying to get a bunch of things to compile cross-static for the bootstrap files. Sounds like I couldn't get a staticlib out of terminfo.

pkgs/development/compilers/llvm/common/libcxx/default.nix
rhelmot: yeah, it seems like for whatever reason it's not required native (I double checked by compiling boost just now). Someone should take a look at cross though...

annaleeleaves: cross builds work fine using `libcxxabi` for all available boost versions
```bash
for x in {168..185} '' 16x 17x 18x; do nix-build -A pkgsCross.x86_64-freebsd.boost$x; done
```
only failures were ones aliased to throw or missing versions.

rhelmot: I'm not sure what you're referring to - is "with libcxxabi" the consequence of my diff?

annaleeleaves: > I'm not sure what you're referring to - is "with libcxxabi" the consequence of my diff?

yes. if `cxxabi == null` you'll get `libcxxabi`.

alyssais: Doesn't normal FreeBSD use libcxxrt?  Are we sure we'd want to be different?

annaleeleaves: > Doesn't normal FreeBSD use libcxxrt? Are we sure we'd want to be different?

one reason to keep it is to support binaries compiled on vanilla freebsd that expect libcxxrt -- we wouldn't have to add any additional logic when binary patching.

i can't imagine there is code trying to dlopen libcxxrt, but that doesn't mean it doesn't exist.

removing libcxxrt (and external cxxabi) from the libcxx derivation reduces complexity and maintenance burden so i'd be happy to see it removed.

if it stays i needs to add `-DLIBCXX_ENABLE_NEW_DELETE_DEFINITIONS=ON`. freebsd has used the libcxx `new.cpp` since llvm12, i think. `libcxxrt` doesn't provide `new [] / delete []` (or some variation of) which breaks linux llvm17+ when building against it, and c++ programs don't build (duckdb) on freebsd cross without it (tho i thought duckdb worked via cross a few months ago).

pkgs/development/compilers/llvm/common/llvm/default.nix
alyssais: Can you mention this in the commit message?

rhelmot: Done. I was also able to clarify some of the other parts of the commit message.

pkgs/development/compilers/llvm/common/libcxx/default.nix
alyssais: It seems like boost started working with libcxxabi when cross compiling to FreeBSD with the LLVM 12 -> 16 update, so maybe it was just a bug in libcxxabi all along that got fixed?  I'm fine with unconditionally using libcxxabi, as long as we bear in mind that we might find reasons in the future that we need to match what normal FreeBSD does and switch back to libcxxrt.

rhelmot:
I integrated the above discussion into a comment in common/llvm/default.nix. I think things have been addressed at this point, so I'm open to new reviews.

pkgs/development/compilers/llvm/common/compiler-rt/default.nix
```diff
@@ -128,10 +128,15 @@ stdenv.mkDerivation ({
   '') + ''
     substituteInPlace lib/builtins/int_util.c \
       --replace "#include <stdlib.h>" ""
+  '' + lib.optionalString (useLLVM && !haveLibc && !stdenv.hostPlatform.isFreeBSD) ''
```
Ericson2314, line 131: This doesn't feel right. If we don't have a libc, then how will we have an `<assert.h>`?

pkgs/development/compilers/llvm/common/compiler-rt/default.nix
```diff
@@ -128,10 +128,15 @@ stdenv.mkDerivation ({
...
     substituteInPlace lib/builtins/clear_cache.c \
       --replace "#include <assert.h>" ""
     substituteInPlace lib/builtins/cpu_model${lib.optionalString (lib.versionAtLeast version "18") "/x86"}.c \
       --replace "#include <assert.h>" ""
+  '' + lib.optionalString (useLLVM && !haveLibc && stdenv.hostPlatform.isFreeBSD) ''
+    # As per above, but in FreeBSD assert is a macro and simply allowing it to be implicitly declared causes Issues!!!!!
+    substituteInPlace lib/builtins/clear_cache.c lib/builtins/cpu_model.c \
+      --replace "#include <assert.h>" "#define assert(e) ((e)?(void)0:__assert(__FUNCTION__,__FILE__,__LINE__,#e))"
   '');
```
Ericson2314, line 140:
```suggestion
  '' + (if stdenv.hostPlatform.isFreeBSD then 
    # As per above, but in FreeBSD assert is a macro and simply allowing it to be implicitly declared causes Issues!!!!!
    ''
    substituteInPlace lib/builtins/clear_cache.c lib/builtins/cpu_model.c \
      --replace "#include <assert.h>" "#define assert(e) ((e)?(void)0:__assert(__FUNCTION__,__FILE__,__LINE__,#e))"
    '' else ''
    substituteInPlace lib/builtins/clear_cache.c \
      --replace "#include <assert.h>" ""
    substituteInPlace lib/builtins/cpu_model${lib.optionalString (lib.versionAtLeast version "18") "/x86"}.c \
      --replace "#include <assert.h>" ""
  ''));
```
Since this is within a `lib.optionalString (useLLVM && !haveLibc)`, I think this is clearer. Someday `nixfmt` will make that more obvious :).

pkgs/development/compilers/llvm/common/libcxx/default.nix
annaleeleaves: so if freebsd is going to use libcxxabi then all the libcxxrt code can be removed as it is now dead. here is the diff to remove it against staging-next. tested `nix-build -A tests.cc-wrapper.llvmTests` and `nix-build -A pkgsCross.x86_64-freebsd.llvmPackages_17.libcxx`.

<details><summary>drop-libcxxrt.diff</summary>

```diff
diff --git a/pkgs/development/compilers/llvm/common/libcxx/default.nix b/pkgs/development/compilers/llvm/common/libcxx/default.nix
index 5b4c2ca56ae1..346a6f6e3fcb 100644
--- a/pkgs/development/compilers/llvm/common/libcxx/default.nix
+++ b/pkgs/development/compilers/llvm/common/libcxx/default.nix
@@ -8,27 +8,17 @@
 , runCommand
 , substitute
 , cmake
-, lndir
 , ninja
 , python3
 , fixDarwinDylibNames
 , version
-, cxxabi ? if stdenv.hostPlatform.isFreeBSD then libcxxrt else null
-, libcxxrt
 , libunwind
 , enableShared ? !stdenv.hostPlatform.isStatic
 }:
 
-# external cxxabi is not supported on Darwin as the build will not link libcxx
-# properly and not re-export the cxxabi symbols into libcxx
-# https://github.com/NixOS/nixpkgs/issues/166205
-# https://github.com/NixOS/nixpkgs/issues/269548
-assert cxxabi == null || !stdenv.hostPlatform.isDarwin;
 let
   basename = "libcxx";
   pname = basename;
-  cxxabiName = "lib${if cxxabi == null then "cxxabi" else cxxabi.libName}";
-  runtimes = [ "libcxx" ] ++ lib.optional (cxxabi == null) "libcxxabi";
 
   # Note: useLLVM is likely false for Darwin but true under pkgsLLVM
   useLLVM = stdenv.hostPlatform.useLLVM or false;
@@ -46,9 +36,8 @@ let
       cp -r ${monorepoSrc}/third-party "$out"
     '') + ''
       cp -r ${monorepoSrc}/runtimes "$out"
-    '' + (lib.optionalString (cxxabi == null) ''
       cp -r ${monorepoSrc}/libcxxabi "$out"
-    '')) else src;
+    '') else src;
 
   cxxabiCMakeFlags = lib.optionals (lib.versionAtLeast release_version "18") [
     "-DLIBCXXABI_USE_LLVM_UNWINDER=OFF"
@@ -66,15 +55,13 @@ let
   ];
 
   cxxCMakeFlags = [
-    "-DLIBCXX_CXX_ABI=${cxxabiName}"
-  ] ++ lib.optionals (cxxabi == null && lib.versionAtLeast release_version "16") [
+    "-DLIBCXX_CXX_ABI=libcxxabi"
+  ] ++ lib.optionals (lib.versionAtLeast release_version "16") [
     # Note: llvm < 16 doesn't support this flag (or it's broken); handled in postInstall instead.
     # Include libc++abi symbols within libc++.a for static linking libc++;
     # dynamic linking includes them through libc++.so being a linker script
     # which includes both shared objects.
     "-DLIBCXX_STATICALLY_LINK_ABI_IN_STATIC_LIBRARY=ON"
-  ] ++ lib.optionals (cxxabi != null) [
-    "-DLIBCXX_CXX_ABI_INCLUDE_PATHS=${lib.getDev cxxabi}/include"
   ] ++ lib.optionals (stdenv.hostPlatform.isMusl || stdenv.hostPlatform.isWasi) [
     "-DLIBCXX_HAS_MUSL_LIBC=1"
   ] ++ lib.optionals (lib.versionAtLeast release_version "18" && !useLLVM && stdenv.hostPlatform.libc == "glibc" && !stdenv.hostPlatform.isStatic) [
@@ -92,7 +79,7 @@ let
   ];
 
   cmakeFlags = [
-    "-DLLVM_ENABLE_RUNTIMES=${lib.concatStringsSep ";" runtimes}"
+    "-DLLVM_ENABLE_RUNTIMES=libcxx;libcxxabi"
   ] ++ lib.optionals (useLLVM && !stdenv.hostPlatform.isWasm) [
     # libcxxabi's CMake looks as though it treats -nostdlib++ as implying -nostdlib,
     # but that does not appear to be the case for example when building
@@ -104,7 +91,7 @@ let
     "-DCMAKE_CXX_COMPILER_WORKS=ON"
     "-DUNIX=ON" # Required otherwise libc++ fails to detect the correct linker
   ] ++ cxxCMakeFlags
-    ++ lib.optionals (cxxabi == null) cxxabiCMakeFlags;
+    ++ cxxabiCMakeFlags;
 
 in
 
@@ -120,19 +107,15 @@ stdenv.mkDerivation (rec {
   '';
 
   nativeBuildInputs = [ cmake ninja python3 ]
-    ++ lib.optional stdenv.isDarwin fixDarwinDylibNames
-    ++ lib.optional (cxxabi != null) lndir;
-
-  buildInputs = [ cxxabi ]
-    ++ lib.optionals (useLLVM && !stdenv.hostPlatform.isWasm) [ libunwind ];
-
-  # libc++.so is a linker script which expands to multiple libraries,
-  # libc++.so.1 and libc++abi.so or the external cxxabi. ld-wrapper doesn't
-  # support linker scripts so the external cxxabi needs to be symlinked in
-  postInstall = lib.optionalString (cxxabi != null) ''
-    lndir ${lib.getDev cxxabi}/include $dev/include/c++/v1
-    lndir ${lib.getLib cxxabi}/lib $out/lib
-    libcxxabi=$out/lib/lib${cxxabi.libName}.a
+    ++ lib.optional stdenv.isDarwin fixDarwinDylibNames;
+
+  buildInputs = lib.optionals (useLLVM && !stdenv.hostPlatform.isWasm) [ libunwind ];
+
+  # vanilla FreeBSD uses libcxxrt to provide the cxxabi functionality, where
+  # nixpkgs is using the libc++abi bundled with LLVM.
+  postInstall = lib.optionalString (stdenv.hostPlatform.isFreeBSD) ''
+    ${lib.optionalString enableShared ''ln -s libc++abi.so $out/lib/libcxxrt.so''}
+    ln -s libc++abi.a $out/lib/libcxxrt.a
   ''
   # LIBCXX_STATICALLY_LINK_ABI_IN_STATIC_LIBRARY=ON doesn't work for LLVM < 16 or
   # external cxxabi libraries so merge libc++abi.a into libc++.a ourselves.
@@ -140,13 +123,12 @@ stdenv.mkDerivation (rec {
   # GNU binutils emits objects in LIFO order in MRI scripts so after the merge
   # the objects are in reversed order so a second MRI script is required so the
   # objects in the archive are listed in proper order (libc++.a, libc++abi.a)
-  + lib.optionalString (cxxabi != null || lib.versionOlder release_version "16") ''
-    libcxxabi=''${libcxxabi-$out/lib/libc++abi.a}
-    if [[ -f $out/lib/libc++.a && -e $libcxxabi ]]; then
+  + lib.optionalString (lib.versionOlder release_version "16") ''
+    if [[ -f $out/lib/libc++.a && -f $out/lib/libc++abi.a ]]; then
       $AR -M <<MRI
         create $out/lib/libc++.a
         addlib $out/lib/libc++.a
-        addlib $libcxxabi
+        addlib $out/lib/libc++abi.a
         save
         end
     MRI
```
</details>

rhelmot: Can this be a followup PR?

pkgs/development/compilers/llvm/common/llvm/default.nix
rhelmot: ...I'm very curious what your perception of the need to go to staging is, if not "rebuild-darwin-stdenv". But yes.

Ericson2314: Oh I didn't see that, haha. But why is it the case? The code looks pretty innocent.

rhelmot: https://github.com/NixOS/nixpkgs/pull/311777/files#diff-ce640f96a649fa4a9dd17247b89883c49921ef7e013fb8ca31bafc65e03d298fL145 adds to a comment in a darwin-gated shell script

Ericson2314: Oh. Let's just add a comment outside the script in Nix code then. And next `staging` we should move the rest of the comment outside too so we don't get this problem again.

rhelmot:
well... with the new cmakeBool we are now mass-rebuilding linux too. Might as well just get the comment movement done now.

pkgs/development/compilers/llvm/common/llvm/default.nix
alyssais: Why would it even matter if it went through staging?  There's no binary cache for FreeBSD anyway.

Ericson2314: @alyssais Good point. But I guess it's still nice to quickly test that cross still works etc. too so that can help.

Issue #309892, opened 2024-05-07
Defining variables with default values in package arguments can lead to problems


Aleksanaa:
# Problem

To make overriding more feasible, people often write something like

```nix
{
  sth,
  doCheck? true,
  patches? [],
  foo? callPackage ./foo.nix,
}:

{ ... }
```

But this is actually dangerous. We don't know when a top-level package named "doCheck", "patches" or "foo" will appear. While the first two can easily be found on ofBorg failures, the third might be quietly passed, causing more problems (and it even looks unrelated!).

# Reproduce

```nix
# default.nix
{ aaa? "1" }:

aaa
```

This should be "1", right?

```
=> nix eval --impure --expr 'with import <nixpkgs> {}; callPackage ./default.nix {}' 
«derivation /nix/store/l36n95rwc21rn1f8sld428j3sarv3ssd-aaa-1.1.1.drv»
```

Well it's not.

# Solution

My idea is that we set some "reserved attribute prefixes", such as names starting with "var-", and make sure that they cannot be used for top-level package names. We may also need to take the same approach for other sets that define `callPackage`.

nim65s:
Prefixing those variables with pname could also be an option. For example, in package zblarg, one could probably use zblargDoCheck, zblargPatches,  zblargWithPython and zblargAaa with a minimal risk of a new unrelated package being named that. 

It would also be more explicit, as zblargWithPython and fooWithPython would be different, while var-withPython and var-withPython in packages zblarg and foo wouldn't be.

And I think this is a common enough practice in various language and build systems already, so this wouldn't be too hard to adopt here.

TomaSajt:
An example for when I've encountered this unexpected default populating logic was when taking in an `src` value (it turns out it's also the name of a VCS package)

reckenrode:
Does https://github.com/NixOS/rfcs/pull/169 at least partially address this issue?

jtojnar:
I would say  that if you need to pass `doCheck`, `src`, `patches` or other `mkDerivation` arguments, you are not calling a package but a package builder/factory, so you should not use `callPackage`. And if your goal is overridable, those can be overridden nicely with `overrideAttrs` (assuming `finalAttrs` pattern).

And if you are using `callPackage` inside `callPackage` trying to follow DRY or something, that will just break `override` anyway https://github.com/NixOS/nixpkgs/pull/311780#discussion_r1601106653

[...]

PR #311780, opened 2024-05-14
cdemu: updates, fixes and build simplification

### LINKED ISSUE #311690: cdemu: not mounting images
### Describe the bug

cdemu is not mounting images

### Steps To Reproduce
1. Install cdemu with `programs.cdemu.enable`, rebuild
2. Attempt to mount an image either manually or cdemu-daemon/cdemu GUI like gCDEmu
3. It fails to mount
4. Manually run `cdemu` and receive error:
```
  File "/nix/store/33zk628s8rpb19inv5y12ai0dvn196py-cdemu-client-3.2.5/bin/.cdemu-wrapped", line 33, in <module>
    from gi.repository import GLib
  File "/nix/store/9njwgmbs7jlpsrnnr731k2vv5xqghqr4-python3.11-pygobject-3.48.2/lib/python3.11/site-packages/gi/importer.py", line 133, in create_module
    raise ImportError('cannot import name %s, '
ImportError: cannot import name GLib, introspection typelib not found
```
### Expected behavior
cdemu mounts the image

### Additional context
vhba module is loaded

### Notify maintainers

@bendlas 

### Metadata

```console
 - system: `"x86_64-linux"`
 - host os: `Linux 6.8.9, NixOS, 24.05 (Uakari), 24.05.20240512.2057814`
 - multi-user?: `yes`
 - sandbox: `yes`
 - channels(root): `"nixos"`
 - nixpkgs: `/nix/store/vsbi6gfwf2bdamw0d0d9xndb1rrxfnbk-source`
```

---

bendlas:
## Description of changes

- rely on wrapGApps*Hook for gobject-introspection access
- use callPackage as intended

fix https://github.com/NixOS/nixpkgs/issues/311690

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin

---

### DIFF
--- a/pkgs/applications/emulators/cdemu/analyzer.nix
+++ b/pkgs/applications/emulators/cdemu/analyzer.nix
@@ -1,20 +1,14 @@
-{ callPackage, makeWrapper, gobject-introspection, cmake
+{ callPackage, makeWrapper, gobject-introspection, cmake, wrapGAppsHook3
 , python3Packages, gtk3, glib, libxml2, gnuplot, gnome, gdk-pixbuf, librsvg, intltool, libmirage }:
-let pkg = import ./base.nix {
-  version = "3.2.5";
+callPackage ./base.nix {
+  version = "3.2.6";
   pname = "image-analyzer";
-  pkgSha256 = "00906lky0z1m0bdqnjmzxgcb19dzvljhddhh42lixyr53sjp94cc";
-};
-in callPackage pkg {
+  sha256 = "sha256-7I8RUgd+k3cEzskJGbziv1f0/eo5QQXn62wGh/Y5ozc=";
   buildInputs = [ glib gtk3 libxml2 gnuplot libmirage gnome.adwaita-icon-theme gdk-pixbuf librsvg
-                  python3Packages.python python3Packages.pygobject3 python3Packages.matplotlib ];
-  drvParams = {
-    nativeBuildInputs = [ gobject-introspection cmake makeWrapper intltool ];
-    postFixup = ''
-      wrapProgram $out/bin/image-analyzer \
-        --set PYTHONPATH "$PYTHONPATH" \
-        --set GI_TYPELIB_PATH "$GI_TYPELIB_PATH" \
-        --prefix XDG_DATA_DIRS : "$out/share:$XDG_ICON_DIRS:$GSETTINGS_SCHEMAS_PATH"
-    '';
-  };
+                  python3Packages.pygobject3 python3Packages.matplotlib ];
+  nativeBuildInputs = [ gobject-introspection cmake wrapGAppsHook3 intltool ];
+  postFixup = ''
+    wrapProgram $out/bin/image-analyzer \
+      --set PYTHONPATH "$PYTHONPATH"
+  '';
 }
--- a/pkgs/applications/emulators/cdemu/analyzer.nix
+++ b/pkgs/applications/emulators/cdemu/analyzer.nix
@@ -1,14 +1,21 @@
-{ callPackage, makeWrapper, gobject-introspection, cmake, wrapGAppsHook3
+{ stdenv, cmake, pkg-config, callPackage, makeWrapper, gobject-introspection, wrapGAppsHook3
 , python3Packages, gtk3, glib, libxml2, gnuplot, gnome, gdk-pixbuf, librsvg, intltool, libmirage }:
-callPackage ./base.nix {
-  version = "3.2.6";
-  pname = "image-analyzer";
-  hash = "sha256-7I8RUgd+k3cEzskJGbziv1f0/eo5QQXn62wGh/Y5ozc=";
-  buildInputs = [ glib gtk3 libxml2 gnuplot libmirage gnome.adwaita-icon-theme gdk-pixbuf librsvg
+stdenv.mkDerivation {
+
+  inherit (callPackage ./common-drv-attrs.nix {
+    version = "3.2.6";
+    pname = "image-analyzer";
+    hash = "sha256-7I8RUgd+k3cEzskJGbziv1f0/eo5QQXn62wGh/Y5ozc=";
+  }) pname version src meta;
+
+  buildInputs = [ libxml2 gnuplot libmirage gnome.adwaita-icon-theme gdk-pixbuf librsvg
                   python3Packages.pygobject3 python3Packages.matplotlib ];
-  nativeBuildInputs = [ gobject-introspection cmake wrapGAppsHook3 intltool ];
+  nativeBuildInputs = [ cmake pkg-config wrapGAppsHook3 intltool ];
+
+  dontWrapGApps = true;
   postFixup = ''
     wrapProgram $out/bin/image-analyzer \
+      ''${gappsWrapperArgs[@]} \
       --set PYTHONPATH "$PYTHONPATH"
   '';
 }
--- a/pkgs/applications/emulators/cdemu/analyzer.nix
+++ b/pkgs/applications/emulators/cdemu/analyzer.nix
@@ -3,7 +3,7 @@
 callPackage ./base.nix {
   version = "3.2.6";
   pname = "image-analyzer";
-  sha256 = "sha256-7I8RUgd+k3cEzskJGbziv1f0/eo5QQXn62wGh/Y5ozc=";
+  hash = "sha256-7I8RUgd+k3cEzskJGbziv1f0/eo5QQXn62wGh/Y5ozc=";
   buildInputs = [ glib gtk3 libxml2 gnuplot libmirage gnome.adwaita-icon-theme gdk-pixbuf librsvg
                   python3Packages.pygobject3 python3Packages.matplotlib ];
   nativeBuildInputs = [ gobject-introspection cmake wrapGAppsHook3 intltool ];
--- a/pkgs/applications/emulators/cdemu/base.nix
+++ b/pkgs/applications/emulators/cdemu/base.nix
@@ -1,12 +1,16 @@
-{ pname, version, pkgSha256 }:
-attrs@{ lib, stdenv, fetchurl, cmake, pkg-config, buildInputs, drvParams ? {}, ... }:
-stdenv.mkDerivation ( rec {
-  inherit pname version buildInputs;
+{ lib, stdenv, fetchurl, cmake, pkg-config
+, pname, version, sha256, buildInputs
+, nativeBuildInputs ? [ ]
+, postFixup ? ""
+, extraDrvParams ? { }
+}:
+stdenv.mkDerivation ( {
+  inherit pname version buildInputs postFixup;
   src = fetchurl {
     url = "mirror://sourceforge/cdemu/${pname}-${version}.tar.xz";
-    sha256 = pkgSha256;
+    inherit sha256;
   };
-  nativeBuildInputs = (attrs.nativeBuildInputs or [ ]) ++ [ pkg-config cmake ];
+  nativeBuildInputs = nativeBuildInputs ++ [ pkg-config cmake ];
   setSourceRoot = ''
     mkdir build
     cd build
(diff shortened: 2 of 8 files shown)


keenanweaver:
Result of `nixpkgs-review pr 311780` run on x86_64-linux [1](https://github.com/Mic92/nixpkgs-review)
<details>
  <summary>5 packages built:</summary>
  <ul>
    <li>cdemu-client</li>
    <li>cdemu-daemon</li>
    <li>gcdemu</li>
    <li>image-analyzer</li>
    <li>libmirage</li>
  </ul>
</details>

keenanweaver (approved):
Fixes the issue! Thank you!

pkgs/applications/emulators/cdemu/base.nix
```diff
@@ -1,12 +1,16 @@
-{ pname, version, pkgSha256 }:
```
jtojnar, line 1: I do not think this is how `callPackage` is intended to be used. IMO it is intended to auto-wire (in dependency injection parlance) packages from `pkgs` set (or other scope) and to allow `override`ing the auto-wired arguments.

Mixing it with parameters and attributes configuring the derivation is messy, and, in case of arguments with default value, is prone to future breakage. See also https://github.com/NixOS/nixpkgs/issues/309892

Additionally, calling `callPackage` inside `callPackage` will make it impossible to `override` the dependencies.

pkgs/applications/emulators/cdemu/base.nix
```diff
@@ -1,12 +1,16 @@
...
-attrs@{ lib, stdenv, fetchurl, cmake, pkg-config, buildInputs, drvParams ? {}, ... }:
-stdenv.mkDerivation ( rec {
-  inherit pname version buildInputs;
+{ lib, stdenv, fetchurl, cmake, pkg-config
+, pname, version, sha256, buildInputs
+, nativeBuildInputs ? [ ]
+, postFixup ? ""
+, extraDrvParams ? { }
+}:
+stdenv.mkDerivation ( {
```
jtojnar, line 7: I would just copy the contents of `base.nix` to each individual  package expression and remove `base.nix`.

It's not like there is much overlap – it really feels like DRY at any cost. But at the end the added piping adds as much duplication as it saves, while being much harder to read and modify (e.g. unable to use `buildPythonApplication` without adding yet another parameter).

pkgs/applications/emulators/cdemu/analyzer.nix
```diff
@@ -1,20 +1,14 @@
...
-      wrapProgram $out/bin/image-analyzer \
-        --set PYTHONPATH "$PYTHONPATH" \
-        --set GI_TYPELIB_PATH "$GI_TYPELIB_PATH" \
-        --prefix XDG_DATA_DIRS : "$out/share:$XDG_ICON_DIRS:$GSETTINGS_SCHEMAS_PATH"
-    '';
-  };
+                  python3Packages.pygobject3 python3Packages.matplotlib ];
+  nativeBuildInputs = [ gobject-introspection cmake wrapGAppsHook3 intltool ];
+  postFixup = ''
+    wrapProgram $out/bin/image-analyzer \
```
jtojnar, line 11: Take care to avoid [double wrapping](https://nixos.org/manual/nixpkgs/unstable/#ssec-gnome-common-issues-double-wrapped).

pkgs/applications/emulators/cdemu/analyzer.nix
```diff
@@ -1,20 +1,14 @@
-{ callPackage, makeWrapper, gobject-introspection, cmake
+{ callPackage, makeWrapper, gobject-introspection, cmake, wrapGAppsHook3
 , python3Packages, gtk3, glib, libxml2, gnuplot, gnome, gdk-pixbuf, librsvg, intltool, libmirage }:
-let pkg = import ./base.nix {
-  version = "3.2.5";
+callPackage ./base.nix {
+  version = "3.2.6";
```
jtojnar, line 4: Probably best to do cleanup first and then update in a separate commit so that it is easier to review.

pkgs/applications/emulators/cdemu/base.nix
```diff
@@ -1,12 +1,16 @@
...
+  inherit pname version buildInputs postFixup;
   src = fetchurl {
     url = "mirror://sourceforge/cdemu/${pname}-${version}.tar.xz";
-    sha256 = pkgSha256;
+    inherit sha256;
   };
-  nativeBuildInputs = (attrs.nativeBuildInputs or [ ]) ++ [ pkg-config cmake ];
+  nativeBuildInputs = nativeBuildInputs ++ [ pkg-config cmake ];
   setSourceRoot = ''
```
jtojnar, line 14: Why is `setSourceRoot` even needed? Also why not use `cmakeDir` and `cmakeFlags` and rely on the default `cmakeConfigurePhase`?

pkgs/applications/emulators/cdemu/analyzer.nix
bendlas: I'm very aware, and that was also why I asked in matrix about it. I'll try your suggestion of using `buildPythonApplication` with `pyproject = false`

pkgs/applications/emulators/cdemu/base.nix
bendlas: I think cdemu might even predate the cmake build hook. In any case, it was my very first PR in this repo, so please forgive the clumsy style. I'll try your suggestion.

pkgs/applications/emulators/cdemu/base.nix
bendlas: Judging by the state of the issue you linked, the proliferation of non - auto-wired default-parameters throughout nixpkgs, and the lack of established alternatives, I'd still judge this to be the intended use of callPackage, despite being dangeroous and even if just intended by toleration.

Curious though, how in this light, the previous solution with two lambdas was actually "more correct" ...

pkgs/applications/emulators/cdemu/base.nix
bendlas: yeah, I had already considered this. especially if we can use the cmake build hook, I'll go for that.

pkgs/applications/emulators/cdemu/analyzer.nix
bendlas: Turns out, that `buildPythonApplication` with `pyproject = false` has the same behavior as `mkDerivation` with `wrapPython`.

- we've still got double-wrapping from `wrapGAppsHook`
- the resulting binary still doesn't find its path:

I think I'll merge this as is for now. We can improve in a separate PR. Maybe that's also a bug with wrapPython? Is that supposed to pick up buildInputs?

```nix
{ fetchurl, python3Packages, cmake, pkg-config, intltool, wrapGAppsNoGuiHook }:
python3Packages.buildPythonApplication rec {
  version = "3.2.5";
  pname = "cdemu-client";
  src = fetchurl {
    url = "mirror://sourceforge/cdemu/${pname}-${version}.tar.xz";
    hash = "sha256-py2F61v8vO0BCM18GCflAiD48deZjbMM6wqoCDZsOd8=";
  };
  pyproject = false;
  nativeBuildInputs = [ cmake pkg-config intltool wrapGAppsNoGuiHook ];
  buildInputs = with python3Packages; [ dbus-python pygobject3 ];
}
```

```console
± ./result/bin/cdemu status
  File "/nix/store/r0rkqb7zv750k4jckpj4cys3nl07jqql-cdemu-client-3.2.5/bin/..cdemu-wrapped-wrapped", line 33, in <module>
    import gi
ModuleNotFoundError: No module named 'gi'
```

pkgs/applications/emulators/cdemu/base.nix
bendlas: yep, turns out that nowadays, the cmake hook does the right thing out of the box

pkgs/applications/emulators/cdemu/base.nix
bendlas: I avoided the nested callPackage in latest commit

pkgs/applications/emulators/cdemu/base.nix
bendlas: I ended up just renaming `base.nix` to `common-drv-attrs.nix` and inherit into the attribute map from it. That avoids the nested `callPackage`, while still deduplicating source info ...

`buildPythonApplication` can also be used easily with this, but it turns out that this doesn't work right now anyway (or I might be using it wrong, see other comment)

pkgs/applications/emulators/cdemu/analyzer.nix
bendlas: I fixed the double wrapping, but still explicitly set PYTHONPATH. I think I'll leave this in for now. Most pressing feedback is fixed and we can always revisit

bendlas: Partially my bad: I managed to fix this with the info you provided @jtojnar 

```sh
makeWrapperArgs+=("''${gappsWrapperArgs[@]}")
```

However: When going that route, we suddenly only have `dconf` and `gobject-introspection` on the `GI_TYPELIB_PATH`, which makes the gtk3 import fail:

```console
ValueError: Namespace Gtk not available
```

bendlas:
still not able to use `buildPythonApplication` because of https://github.com/NixOS/nixpkgs/pull/311780#discussion_r1601451210, but this is good enough for now.

pkgs/applications/emulators/cdemu/analyzer.nix
jtojnar: > Is that supposed to pick up buildInputs?

No. Python modules are only picked from `propagatedBuildInputs` and `pythonPath`.

jtojnar: See [the manual](https://nixos.org/manual/nixpkgs/unstable/#ssec-gnome-common-issues-double-wrapped) for how to resolve the double wrapping.

bendlas: I did both. It already works. But I change it to `buildPythonApplication` together with `makeWrapperArgs+=("''${gappsWrapperArgs[@]}")`, then it suddenly doesn't any more.

bendlas: @jtojnar am I doing something wrong here? https://github.com/bendlas/nixpkgs/commit/e4ab3dfcbbc298993c9e59fca6d3c57947237d4b

```console
± $(nix-build . -A image-analyzer --no-out-link)/bin/image-analyzer
  File "/nix/store/lgnv7ay3hi94rzwxvxj1c7k0xi7swy3x-image-analyzer-3.2.6/bin/.image-analyzer-wrapped", line 26, in <module>
    gi.require_version('Gtk', '3.0')
  File "/nix/store/9njwgmbs7jlpsrnnr731k2vv5xqghqr4-python3.11-pygobject-3.48.2/lib/python3.11/site-packages/gi/__init__.py", line 122, in require_version
    raise ValueError('Namespace %s not available' % namespace)
ValueError: Namespace Gtk not available
```

As I said, `GI_TYPELIB_PATH` is missing a lot of stuff, compared to the `stdenv.mkDerivation` variant, that's currently in master

```console
± less $(nix-build . -A image-analyzer --no-out-link)/bin/image-analyzer
...
# The C-code for this binary wrapper has been generated using the following command:
makeCWrapper '/nix/store/lgnv7ay3hi94rzwxvxj1c7k0xi7swy3x-image-analyzer-3.2.6/bin/.image-analyzer-wrapped' \
   ...
    --prefix 'GI_TYPELIB_PATH' ':' '/nix/store/bcms6vbac6jyffwyj8y630brpy1il4ks-glib-2.80.0/lib/girepository-1.0:/nix/store/ys2ajfa1bbx0a3axb722na18ymw8h9jy-gobject-introspection-1.80.1/lib/girepository-1.0'
...
```

jtojnar: You are missing `gobject-introspection.setupHook` from `nativeBuildInputs`.

bendlas: that works. let's continue here https://github.com/NixOS/nixpkgs/pull/311959

PR #311762, opened 2024-05-14
bintools: fix wrapper build on FreeBSD native

rhelmot:
Also port a similar fix to cc wrapper

part of https://github.com/NixOS/nixpkgs/pull/296581

This one is... complicated. The dynamic loader path should be self explanatory, but the output filtering in meta was the only thing I could figure out how to do to assuage a recursion error in stdenv/freebsd (linked). If someone else knows what exactly I'm messing with here, please feel free to issue a better fix and I'll try it out.

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [x] x86_64-freebsd
- For non-Linux: Is sandboxing enabled in `nix.conf`? (See [Nix manual](https://nixos.org/manual/nix/stable/command-ref/conf-file.html))
  - [ ] `sandbox = relaxed`
  - [x] `sandbox = true`
- [ ] Tested, as applicable:
  - [NixOS test(s)](https://nixos.org/manual/nixos/unstable/index.html#sec-nixos-tests) (look inside [nixos/tests](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests))
  - and/or [package tests](https://github.com/NixOS/nixpkgs/blob/master/pkgs/README.md#package-tests)
  - or, for functions and "core" functionality, tests in [lib/tests](https://github.com/NixOS/nixpkgs/blob/master/lib/tests) or [pkgs/test](https://github.com/NixOS/nixpkgs/blob/master/pkgs/test)
  - made sure NixOS tests are [linked](https://nixos.org/manual/nixpkgs/unstable/#ssec-nixos-tests-linking) to the relevant packages
- [ ] Tested compilation of all packages that depend on this change using `nix-shell -p nixpkgs-review --run "nixpkgs-review rev HEAD"`. Note: all changes have to be committed, also see [nixpkgs-review usage](https://github.com/Mic92/nixpkgs-review#usage)
- [x] Tested basic functionality of all binary files (usually in `./result/bin/`)
[...]

PR #311789, opened 2024-05-14
top-level/all-packages: misc changes to support native FreeBSD

rhelmot:
## Description of changes

* autoPatchelfHook works on FreeBSD as well
* bootstrapping perl needs bootstrapped zlib
* fix some invocation parameters for bootstrapping coreutils
* fix some parameters for invoking bintools and cc wrappers. related to https://github.com/NixOS/nixpkgs/pull/311762
* cyrus_sasl no longer needs its special case for FreeBSD
* add native-libc support to libiconv-libc. This is probably not needed anymore but it doesn't hurt to have!
* iconv-bin should use the binary from the freebsd tree
* libuuid is provided by e2fsprogs on freebsd

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [x] x86_64-freebsd
  - [ ] `sandbox = relaxed`
  - [x] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/top-level/all-packages.nix
+++ b/pkgs/top-level/all-packages.nix
@@ -20443,9 +20443,7 @@ with pkgs;
 
   cypress = callPackage ../development/web/cypress { };
 
-  cyrus_sasl = callPackage ../development/libraries/cyrus-sasl {
-    libkrb5 = if stdenv.isFreeBSD then heimdal else libkrb5;
-  };
+  cyrus_sasl = callPackage ../development/libraries/cyrus-sasl { };
 
   cyrus-sasl-xoauth2 = callPackage ../development/libraries/cyrus-sasl-xoauth2 { };
 
--- a/pkgs/top-level/all-packages.nix
+++ b/pkgs/top-level/all-packages.nix
@@ -22484,6 +22484,8 @@ with pkgs;
       lib.getBin stdenv.cc.libc
     else if stdenv.hostPlatform.isDarwin then
       lib.getBin libiconv
+    else if stdenv.hostPlatform.isFreeBSD then
+      lib.getBin freebsd.iconv
     else
       lib.getBin libiconvReal;
 
--- a/pkgs/top-level/all-packages.nix
+++ b/pkgs/top-level/all-packages.nix
@@ -1146,7 +1146,7 @@ with pkgs;
             fetchurl = stdenv.fetchurlBoot;
           };
         });
-        perl = buildPackages.perl.override { fetchurl = stdenv.fetchurlBoot; };
+        perl = buildPackages.perl.override { inherit zlib; fetchurl = stdenv.fetchurlBoot; };
         openssl = buildPackages.openssl.override {
           fetchurl = stdenv.fetchurlBoot;
           buildPackages = {
@@ -1154,7 +1154,7 @@ with pkgs;
               fetchurl = stdenv.fetchurlBoot;
               inherit perl;
               xz = buildPackages.xz.override { fetchurl = stdenv.fetchurlBoot; };
-              gmp = null;
+              gmpSupport = false;
               aclSupport = false;
               attrSupport = false;
             };


alyssais:
> libuuid is provided by e2fsprogs on freebsd

Can you explain this some more?  Can we not use the same libuuid implementation as on Linux?  (util-linux certainly builds a subset of itself for Darwin.)

pkgs/top-level/all-packages.nix
```diff
@@ -1146,17 +1147,18 @@ with pkgs;
...
           buildPackages = {
             coreutils = buildPackages.coreutils.override {
               fetchurl = stdenv.fetchurlBoot;
               inherit perl;
               xz = buildPackages.xz.override { fetchurl = stdenv.fetchurlBoot; };
-              gmp = null;
+              gmpSupport = false;
               aclSupport = false;
               attrSupport = false;
+              autoreconfHook = null;
```
alyssais, line 1161: Setting inputs to null like this can be very confusing, and is something we're trying to phase out.  There's a comment that says it's only needed here on Cygwin, so maybe we could only conditionally add it to nativeBuildInputs in that case.

pkgs/top-level/all-packages.nix
```diff
@@ -193,6 +193,7 @@ with pkgs;
       pythonInterpreter = "${python3.withPackages (ps: [ ps.pyelftools ])}/bin/python";
       autoPatchelfScript = ../build-support/setup-hooks/auto-patchelf.py;
     };
+    meta.platforms = lib.platforms.linux ++ lib.platforms.freebsd;
```
alyssais, line 196: Is this necessary?  I already see FreeBSD in there:

```
nix-repl> autoPatchelfHook.meta.platforms
[
  "aarch64-linux"
  "armv5tel-linux"
  "armv6l-linux"
  "armv7a-linux"
  "armv7l-linux"
  "i686-linux"
  "loongarch64-linux"
  "m68k-linux"
  "microblaze-linux"
  "microblazeel-linux"
  "mips-linux"
  "mips64-linux"
  "mips64el-linux"
  "mipsel-linux"
  "powerpc64-linux"
  "powerpc64le-linux"
  "riscv32-linux"
  "riscv64-linux"
  "s390-linux"
  "s390x-linux"
  "x86_64-linux"
  "i686-freebsd"
  "x86_64-freebsd"
]
```

pkgs/top-level/all-packages.nix
```diff
@@ -17006,6 +17008,7 @@ with pkgs;
     nativeLibc = stdenv.targetPlatform == stdenv.hostPlatform && stdenv.cc.nativeLibc or false;
     nativePrefix = stdenv.cc.nativePrefix or "";
     noLibc = !self.nativeLibc && (self.libc == null);
+    propagateDoc = if self.nativeTools then false else ((stdenv.cc ? man) && (cc ? man));
```
alyssais, line 17011: Why set this here instead of modifying the default?

pkgs/top-level/all-packages.nix
```diff
@@ -17026,8 +17029,8 @@ with pkgs;
     nativeTools = stdenv.targetPlatform == stdenv.hostPlatform && stdenv.cc.nativeTools or false;
     nativeLibc = stdenv.targetPlatform == stdenv.hostPlatform && stdenv.cc.nativeLibc or false;
     nativePrefix = stdenv.cc.nativePrefix or "";
-
-    noLibc = (self.libc == null);
+    propagateDoc = if self.nativeTools then false else ((stdenv.cc ? man) && (bintools ? man));
+    noLibc = !self.nativeLibc && (self.libc == null);
```
alyssais, line 17033: @Ericson2314 what does nativeLibc mean?  Could we have a comment or something somewhere explaining it?

pkgs/top-level/all-packages.nix
```diff
@@ -22495,7 +22496,13 @@ with pkgs;
...
       then libcIconv (if stdenv.hostPlatform != stdenv.buildPlatform
-        then libcCross
+          then libcCross
+        else if stdenv.cc.nativeLibc
+          then {
+            pname = "libc-${stdenv.hostPlatform.system}";
+            version = "native";
+            dev = stdenv.cc.nativePrefix;
+          }
         else stdenv.cc.libc)
```
alyssais, line 22506: Do we not have a stdenv.cc.libc on FreeBSD?

alyssais (commented):
This is small enough to review in a single PR, but it would still be nice if all the little independent changes in here could be separated into their own commits.

pkgs/top-level/all-packages.nix
rhelmot: This is another nativeLibc case. nativeLibc is where you're using nix on a non-nixos system and you're using the libc from that system. I originally added these lines for the original, impure bootstrap, which used that for the first phase. I can remove everything related to this if you want.

pkgs/top-level/all-packages.nix
rhelmot: Using the psychic onset from standing on a mountaintop in the crisp spring air (in front of my computer, at home) I determined that you really don't want this diff. It's gone now.

pkgs/top-level/all-packages.nix
rhelmot: Indeed. It is gone.

rhelmot:
I removed the libuuid diff because I determined it was not actually needed for the stdenv. If I rediscover the thing that prompted it, I'll investigate whether util-linux can satisfy it.

alyssais (approved):

SuperSandro2000 (approved):

SuperSandro2000 (approved):

PR #311794, opened 2024-05-14
make-binary-wrapper: add cc to propagatedBuildInputs

rhelmot:
This is required for stdenv on native FreeBSD.

## Description of changes

This is an @artemist change. While I don't understand it myself, I recall her telling me that this was something that should have been included for linux as well but was just working anyway by chance. It is required for the native FreeBSD stdenv.

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [x] x86_64-freebsd
  - [ ] `sandbox = relaxed`
  - [x] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/build-support/setup-hooks/make-binary-wrapper/default.nix
+++ b/pkgs/build-support/setup-hooks/make-binary-wrapper/default.nix
@@ -10,7 +10,7 @@
 
 makeSetupHook {
   name = "make-binary-wrapper-hook";
-  propagatedBuildInputs = [ dieHook ];
+  propagatedBuildInputs = [ dieHook cc ];
 
   substitutions = {
     cc = "${cc}/bin/${cc.targetPrefix}cc ${lib.escapeShellArgs (map (s: "-fsanitize=${s}") sanitizers)}";


artemist:
Adding `cc` to `propagatedBuildInputs` makes derivations with `makeBinaryWrapper` in `nativeBuildInputs` run the cc-wrapper setup hook.
This isn't an issue for derivations using `stdenv`, as the cc setup hook is already run by default. However, derivations that are made with `stdenvNoCC`, e.g. because they're made with `runCommand`, will not run the cc-wrapper setup hook without this change.

For some reason, I don't recall why, the FreeBSD native compiler will not work correctly without the setup hook.

alyssais:
Sounds sensible to me.  Can the commit message be updated with @artemist's explanation?

alyssais:
This seems to be causing some weird side effects.  pkgsMusl.systemd no longer builds:

```
Command line: `clang /run/user/1008/tmp.WuVNjgdfHT/source/build/meson-private/tmpdd3qc57v/testfile.c -o /run/user/1008/tmp.WuVNjgdfHT/source/build/meson-private/tmpdd3qc57v/output.obj -c -D_FILE_OFFSET_BITS=64 -O0 -Werror=implicit-function-declaration -Werror=unknown-warning-option -Werror=unused-command-line-argument -Werror=ignored-optimization-argument -std=gnu11` -> 1
stderr:
clang: error: argument unused during compilation: '-pie' [-Werror,-Wunused-command-line-argument]
```

Given that this is a stdenv rebuild, and it seems unlikely to me that it's something specific about pkgsMusl.systemd that's broken, I think we need to revert this before it makes its way into the next staging cycle and is no longer changeable until the next one.

rhelmot:
I finally ran into the error which necessitated this change: while cross-building a perl-env for FreeBSD:

```
clang: error: unable to execute command: Executable "ld" doesn't exist!
```

This seems to be because perl's buildEnv explicitly includes makeBinaryWrapper but buildEnv uses runCommand (not runCommandCC). What should be the correct fix in this case?

Ericson2314:
I think the issue with this is that the hardening disable flags are not done per platform like the other cc-/bintools-wrapper env vars. (I ran into this with OpenBSD.)

PR #311811, opened 2024-05-15
binlore: migrate override lore to package passthru

abathur:
Move lore overrides from binlore's source repo to the passthru of the target packages.

> Note: to keep this from sprawling, see issue/previous PR below for more background/explanation:
> - #295029
> - #304956

## Description of changes

- Change how lore is merged/gathered:
	- Don't apply overrides from binlore's source. 
	- Merge lore from `$out/nix-support/<loretype>` if it exists. Prepares binlore to collect lore generated by `makeWrapper` and friends. 

	    (Saving that for a follow-up, since it's a huge rebuild that makes this harder to test.)

	- Merge lore from `<package>.binlore` if it exists.

	    (Happy to debate this name. Avoiding `overrideLore` since `override` is roughly a contract in nixpkgs.)
	    
	    > *Note*: this has been changed from `<package>.lore` to `<package>.binlore` during review in order to mitigate confusion about what this is for. We might need to revisit this later if there are more lore providers (or we come up with better terms all around).

- Add a utility function (`binlore.synthesize`) for ~synthesizing lore overrides. It rolls in a small Shell DSL for efficiently expressing overrides.

    (Happy to debate this name. Avoiding `override` since it's roughly a contract in nixpkgs.) 

- Reimplement most overrides using `passthru.binlore = binlore.synthesize ...;`. (I'm skipping some to pare them back to the set needed for nixpkgs and other public repos found via search.)

- Expand resholve's passthru tests to better exercise the machinery.

I also have a [branch with these changes](https://github.com/abathur/nixpkgs/tree/move_lore_overrides_to_passthru_unstable) based on a recent nixpkgs-unstable commit for easier testing. I generally test with `nix-build -A resholve.tests`.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [x] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [x] Tested, as applicable:

---

### DIFF
--- a/pkgs/by-name/pd/pdf2odt/package.nix
+++ b/pkgs/by-name/pd/pdf2odt/package.nix
@@ -49,6 +49,10 @@ resholve.mkDerivation rec {
       imagemagick
       zip
     ];
+    execer = [
+      # zip can exec; confirmed 2 invocations in pdf2odt don't
+      "cannot:${zip}/bin/zip"
+    ];
   };
 
   meta = with lib; {
--- a/pkgs/development/libraries/libarchive/default.nix
+++ b/pkgs/development/libraries/libarchive/default.nix
@@ -23,6 +23,9 @@
 , cmake
 , nix
 , samba
+
+# for passthru.lore
+, binlore
 }:
 
 assert xarSupport -> libxml2 != null;
@@ -125,4 +128,11 @@ stdenv.mkDerivation (finalAttrs: {
   passthru.tests = {
     inherit cmake nix samba;
   };
+
+  # bsdtar is detected as "cannot" because its exec is internal to
+  # calls it makes into libarchive itself. If binlore gains support
+  # for detecting another layer down into libraries, this can be cut.
+  passthru.binlore.out = binlore.synthesize finalAttrs.finalPackage ''
+    execer can bin/bsdtar
+  '';
 })
--- a/pkgs/development/tools/analysis/binlore/default.nix
+++ b/pkgs/development/tools/analysis/binlore/default.nix
@@ -56,58 +56,169 @@ let
     #   in here, but I'm erring on the side of flexibility
     #   since this form will make it easier to pilot other
     #   uses of binlore.
-    callback = lore: drv: overrides: ''
+    callback = lore: drv: ''
       if [[ -d "${drv}/bin" ]] || [[ -d "${drv}/lib" ]] || [[ -d "${drv}/libexec" ]]; then
         echo generating binlore for $drv by running:
         echo "${yara}/bin/yara --scan-list --recursive ${lore.rules} <(printf '%s\n' ${drv}/{bin,lib,libexec}) | ${yallback}/bin/yallback ${lore.yallback}"
       else
         echo "failed to generate binlore for $drv (none of ${drv}/{bin,lib,libexec} exist)"
       fi
-    '' +
-    /*
-    Override lore for some packages. Unsure, but for now:
-    1. start with the ~name (pname-version)
-    2. remove characters from the end until we find a match
-       in overrides/
-    3. execute the override script with the list of expected
-       lore types
-    */
-    ''
-      i=''${#identifier}
-      filter=
-      while [[ $i > 0 ]] && [[ -z "$filter" ]]; do
-        if [[ -f "${overrides}/''${identifier:0:$i}" ]]; then
-          filter="${overrides}/''${identifier:0:$i}"
-          echo using "${overrides}/''${identifier:0:$i}" to generate overriden binlore for $drv
-          break
-        fi
-        ((i--)) || true # don't break build
-      done # || true # don't break build
+
       if [[ -d "${drv}/bin" ]] || [[ -d "${drv}/lib" ]] || [[ -d "${drv}/libexec" ]]; then
-        ${yara}/bin/yara --scan-list --recursive ${lore.rules} <(printf '%s\n' ${drv}/{bin,lib,libexec}) | ${yallback}/bin/yallback ${lore.yallback} "$filter"
+        ${yara}/bin/yara --scan-list --recursive ${lore.rules} <(printf '%s\n' ${drv}/{bin,lib,libexec}) | ${yallback}/bin/yallback ${lore.yallback}
       fi
     '';
   };
-  overrides = (src + "/overrides");
 
 in rec {
+  /*
+    Output a directory containing lore for multiple drvs.
+
+    This will `make` lore for drv in drvs and then combine lore
+    of the same type across all packages into a single file.
+
+    When drvs are also specified in the strip argument, corresponding
+    lore is made relative by stripping the path of each drv from
+    matching entries. (This is mainly useful in a build process that
+    uses a chain of two or more derivations where the output of one
+    is the source for the next. See resholve for an example.)
+  */
   collect = { lore ? loreDef, drvs, strip ? [ ] }: (runCommand "more-binlore" { } ''
     mkdir $out
     for lorefile in ${toString lore.types}; do
       cat ${lib.concatMapStrings (x: x + "/$lorefile ") (map (make lore) (map lib.getBin (builtins.filter lib.isDerivation drvs)))} > $out/$lorefile
-      substituteInPlace $out/$lorefile ${lib.concatMapStrings (x: "--replace '${x}/' '' ") strip}
+      substituteInPlace $out/$lorefile ${lib.concatMapStrings (x: "--replace-quiet '${x}/' '' ") strip}
     done
   '');
-  # TODO: echo for debug, can be removed at some point
+
+  /*
+    Output a directory containing lore for a single drv.
+
+    This produces lore for the derivation (via lore.callback) and
+    appends any lore that the derivation itself wrote to nix-support
+    or which was overridden in drv.binlore.<outputName> (passthru).
+
+    > *Note*: Since the passthru is attached to all outputs, binlore
+    > is an attrset namespaced by outputName to support packages with
+    > executables in more than one output.
+
+    Since the last entry wins, the effective priority is:
+    drv.binlore.<outputName> > $drv/nix-support > lore generated here by callback
+  */
   make = lore: drv: runCommand "${drv.name}-binlore" {
-      identifier = drv.name;
       drv = drv;
     } (''
     mkdir $out
     touch $out/{${builtins.concatStringsSep "," lore.types}}
 
-    ${lore.callback lore drv overrides}
+    ${lore.callback lore drv}
+    '' +
+    # append lore from package's $out and drv.binlore.${drv.outputName} (last entry wins)
+    ''
+    for lore_type in ${builtins.toString lore.types}; do
+      if [[ -f "${drv}/nix-support/$lore_type" ]]; then
+        cat "${drv}/nix-support/$lore_type" >> "$out/$lore_type"
+      fi
+      '' + lib.optionalString (builtins.hasAttr "binlore" drv && builtins.hasAttr drv.outputName drv.binlore) ''
+      if [[ -f "${drv.binlore."${drv.outputName}"}/$lore_type" ]]; then
+        cat "${drv.binlore."${drv.outputName}"}/$lore_type" >> "$out/$lore_type"
+      fi
+      '' + ''
+    done
 
     echo binlore for $drv written to $out
   '');
+
+  /*
+    Utility function for creating override lore for drv.
+
+    We normally attach this lore to `drv.passthru.binlore.<outputName>`.
+
+    > *Notes*:
+    > - Since the passthru is attached to all outputs, binlore is an
+    >   attrset namespaced by outputName to support packages with
+    >   executables in more than one output. You'll generally just use
+    >   `out` or `bin`.
+    > - We can reconsider the passthru attr name if someone adds
+    >   a new lore provider. We settled on `.binlore` for now to make it
+    >   easier for people to figure out what this is for.
+
+    The lore argument should be a Shell script (string) that generates
+    the necessary lore. You can use arbitrary Shell, but this function
+    includes a shell DSL you can use to declare/generate lore in most
+    cases. It has the following functions:
+
+    - `execer <verdict> [<path>...]`
+    - `wrapper <wrapper_path> <original_path>`
+
+    Writing every override explicitly in a Nix list would be tedious
+    for large packages, but this small shell DSL enables us to express
+    many overrides efficiently via pathname expansion/globbing.
+
+    Here's a very general example of both functions:
+
+    passthru.binlore.out = binlore.synthesize finalAttrs.finalPackage ''
+      execer can bin/hello bin/{a,b,c}
+      wrapper bin/hello bin/.hello-wrapped
+    '';
+
+    And here's a specific example of how pathname expansion enables us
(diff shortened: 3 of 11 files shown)


abathur:
@ofborg build resholve.tests

abathur:
@ofborg build resholve.tests

roberth (approved):
Raised minor concerns about documentation out-of-band; which @abathur will handle with infinisil / docs team, and async from this because staging PRs aren't particularly suited for iterating on docs and such.

pkgs/development/tools/analysis/binlore/default.nix
```diff
@@ -56,58 +56,160 @@ let
...
-      substituteInPlace $out/$lorefile ${lib.concatMapStrings (x: "--replace '${x}/' '' ") strip}
+      substituteInPlace $out/$lorefile ${lib.concatMapStrings (x: "--replace-quiet '${x}/' '' ") strip}
     done
   '');
-  # TODO: echo for debug, can be removed at some point
+
+  /*
+    Output a directory containing lore for a single drv.
+
+    This produces lore for the derivation (via lore.callback) and
```
jtojnar, line 97: It is bit confusing that `lore` argument is used for multiple different things. Maybe use something like `loreGen.callback` here.

pkgs/development/tools/analysis/binlore/default.nix
```diff
@@ -56,58 +56,160 @@ let
...
     touch $out/{${builtins.concatStringsSep "," lore.types}}
 
-    ${lore.callback lore drv overrides}
+    ${lore.callback lore drv}
+    '' +
+    # append lore from package's $out and drv.binlore (last entry wins)
+    ''
+    for lore_type in ${builtins.toString lore.types}; do
+      if [[ -f "${drv}/nix-support/$lore_type" ]]; then
+        cat "${drv}/nix-support/$lore_type" >> "$out/$lore_type"
```
jtojnar, line 116: Having two different directory paths is weird. Would it be feasible to standardize on e.g. `lib/binlore/$lore_type`?

pkgs/development/tools/analysis/binlore/default.nix
```diff
@@ -56,58 +56,160 @@ let
...
+    - `execer <verdict> [<path>...]`
+    - `wrapper <wrapper_path> <original_path>`
+
+    Writing every override explicitly in a Nix list would be tedious
+    for large packages, but this small shell DSL enables us to express
+    many overrides efficiently via pathname expansion/globbing.
+
+    Here's a very general example of both functions:
+
+    passthru.binlore = (binlore.synthesize finalAttrs.finalPackage ''
```
jtojnar, line 151: Do you plan to support multiple outputs? For example, `glib` has programs in `out`, `bin` and `dev`.

pkgs/development/tools/analysis/binlore/default.nix
```diff
@@ -56,58 +56,160 @@ let
...
+  synthesize = drv: lore: runCommand "${drv.name}-lore-override" {
+    drv = drv;
+  } (''
+    execer(){
+      local verdict="$1"
+
+      shift
+
+      for path in "$@"; do
+        if [[ -e "$PWD/$path" ]]; then
```
jtojnar, line 182: Are directories supported intentionally?

pkgs/by-name/pd/pdf2odt/package.nix
```diff
@@ -49,6 +49,10 @@ resholve.mkDerivation rec {
       imagemagick
       zip
     ];
+    execer = [
+      # zip can exec; confirmed 2 invocations in pdf2odt don't
+      "cannot:${zip}/bin/zip"
```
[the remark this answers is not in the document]
abathur: The global override was a mistake (but I'm glad you're scrubbing hard enough to notice).

When going back over it, I noticed user-controlled exec in the `-TT cmd/--unzip-command cmd` options:
> Use command cmd instead of 'unzip -tqq' to test an archive when the -T option is used."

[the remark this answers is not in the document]
abathur: In general, there's an open problem here about how to ~operationalize this kind of manual research by documenting what we looked at, what we found, and pinning ~assertions about the command(s) to the underlying source in a way that we can re-evaluate after updates that merit a fresh look. I imagined an advantage of keeping this stuff in the binlore repo was going to be having an obvious place to try to collect that knowledge, but I wasn't looking forward to figure out the right mix of automation/coupling to keep up with nixpkgs from there.

If the nixpkgs immune system doesn't reject this current override effort I'm thinking we might be able to figure out how to add passthru tests that assert knowledge about the underlying source for a CLI. I haven't put too much thought into that yet since it would be moot if this one's a nonstarter.

pkgs/development/tools/analysis/binlore/default.nix
abathur: I've waffled a little about where to put this. 

Open to changing it (whether here or later) so I'll just think out loud for now: 
- I think if we moved it into a `lib/` dir it'd be at risk of getting split off into a lib output when it's present (so I guess this is at least somewhat related to your other question about multiple outputs)?
- When I was first thinking about doing this I didn't feel like it was my place to take up part of the nix-support namespace and imagined this actually belongs in some new namespace that doesn't imply it's just for nix/nixpkgs internal purposes. I opened an issue in the architecture repo last year but it didn't go anywhere and has been since transferred into this repo (#295029).
- In #314937 I'm also tweaking the wrapper generators to output this data, and `nix-support/` does feel _more_ semantically appropriate from that perspective. Since this path is ~anticipating the other change, maybe it feels less out-of-place with that in frame?

pkgs/development/tools/analysis/binlore/default.nix
abathur: Yes and no? :)

I couldn't remember when I was writing it whether the "regular file" bit of `-f` would exclude anything we would actually want to match here. I've had trouble in the past finding a lucid definition of what counts as a regular file and how symlink resolution fits in to the specific context of file property testing (POSIX's own definition strikes me as obtuse). Since this check is mostly just here to make sure we catch typos and removed paths, I decided overmatching was the smaller crime.

I'm happy to change it if you're confident `-f` won't leave out any edge cases we need?

pkgs/tools/misc/coreutils/default.nix
```diff
@@ -176,6 +178,18 @@ stdenv.mkDerivation rec {
     rm -r "$out/share"
   '';
 
+  passthru = {} // optionalAttrs (singleBinary != false) {
+    # everything in the single binary gets the same verdict, so we
+    # override _that case_ with verdicts from separate binaries.
+    #
+    # binlore only spots exec in runcon on some platforms (i.e., not
```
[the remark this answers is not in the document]
abathur: Am I reading you right that the question is why I didn't override the runcon verdict for split binaries?

Fair. I was kinda punting on it here. The command help makes it clear that it is for running other commands, so I'm assuming the macOS version of it is just ~neutered and therefore safe-ish to go ahead and resolve cleanly on macOS.

That said, you've made me think a little harder about it and there's also a case that it's unnecessarily annoying to let people PR something that resholve doesn't object to on the split binary on macOS that then breaks in ofBorg from linux.

I'll chew on it, but I'm leaning towards hard-setting runcon to "can" for both platforms in my next push. If it actually disrupts some legitimate use of the command on macOS, someone will likely show up to question it at that point.

pkgs/tools/misc/coreutils/default.nix
```diff
@@ -176,6 +178,18 @@ stdenv.mkDerivation rec {
...
+    # darwin; I have a note that the behavior may need selinux?)
+    binlore = (binlore.synthesize coreutils ''
+      execer can bin/{chroot,env,install,nice,nohup,runcon,sort,split,stdbuf,timeout}
```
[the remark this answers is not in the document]
abathur: I don't think it would work quite like this, but the underlying point is interesting:
- synthesize isn't collecting the underlying lore for the package passed, so this would produce an empty override
- lore is emitted with absolute paths, so we'd need something to relativize and remap it

pkgs/tools/text/esh/default.nix
```diff
@@ -30,6 +30,14 @@ stdenv.mkDerivation rec {
   doCheck = true;
   checkTarget = "test";
 
+  # working around a bug in file. Was fixed in
```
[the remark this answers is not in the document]
abathur: Yep!

pkgs/top-level/unixtools.nix
```diff
@@ -30,7 +30,9 @@ let
         priority = 10;
         platforms = platforms.${stdenv.hostPlatform.parsed.kernel.name} or platforms.all;
       };
-      passthru = { inherit provider; };
+      passthru = { inherit provider; } // lib.optionalAttrs (builtins.hasAttr "binlore" providers) {
```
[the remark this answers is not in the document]
abathur: If by this you mean push them out into the packages where these all come from, I think I mostly agree.

I tried that first and ended up settling with implementing them here for a few expedience reasons, the foremost of which was not really being able to figure out how to splice it in to some of the macOS commands which are already fairly abstractly sliced out of source releases in a way that makes it hard to tell where and how to intervene and also be able to pair up lore overrides for the commands we care about.

Looking around at the implementation of `diskdev_cmds` in https://github.com/NixOS/nixpkgs/tree/master/pkgs/os-specific/darwin/apple-source-releases may clarify why I backed off of that (but I'm happy to do it if you can tell how we could accomplish it in a way that's as or more straightforward than what I have now?)

pkgs/development/libraries/libarchive/default.nix
```diff
@@ -125,4 +128,11 @@ stdenv.mkDerivation (finalAttrs: {
   passthru.tests = {
     inherit cmake nix samba;
   };
+
+  # bsdtar is detected as "cannot" because its exec is internal to
```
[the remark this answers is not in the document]
abathur: Is this on the right one? I'm not sure about moving it into the multiline string. I do like moving it closer, but I think I prefer the clearer legibility of it being a comment at the Nix level. 

If we need to leave breadcrumbs for more executables, the comments can just be converted into a bulleted list?

pkgs/os-specific/linux/procps-ng/default.nix
```diff
@@ -61,6 +64,12 @@ stdenv.mkDerivation rec {
     install -m 0644 -D watch.1 $out/share/man/man1/watch.1
   '';
 
+  # no obvious exec in documented arguments; haven't trawled source
+  # to figure out what exec binlore hits on
+  passthru.binlore = (binlore.synthesize procps ''
```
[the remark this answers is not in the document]
abathur: Fair. Habit for clarity, but I'll drop them on next push.

pkgs/development/tools/analysis/binlore/default.nix
abathur: I didn't really like this, either, but I'm trying to keep the change from sprawling out. 

In the next push I'll optimistically rename the _new_ argument to `loreSynthesizingScript` since that's a bit more honest anyways. Do you think that satisfies the immediate concern?

(If not and you feel like it's a blocker, I'd prefer to rename them in a standalone clarity-focused changeset that does nothing but fiddle with the names?)

pkgs/development/tools/analysis/binlore/default.nix
abathur: Good point/question, though I may need some advice.

I may be misunderstanding since I haven't intentionally crossed this bridge, but some naive play from CLI suggests passthru is reachable from each output, so I guess all we'd _have_ to do is make sure it's possible to generate one lore attr covering all of the outputs.

I guess I can imagine several approaches, but the best are probably:
1. Don't try to do the current convenience bit where the API cds into the output directory to expand/test the paths--make the call-sites explicitly inline the relevant output and do the expansion (and path testing) from there. Script format gets a little less efficient, but for most uses it'd be roughly balanced by dropping the 1st arg.
2. Accept a list instead of a single drv, try each path on every output in the list and only error if there's no match at all. Keeps the script format efficient but the implementation will get more complex in order to try expanding and testing each path on each drv and failing if a particular path didn't exist in any of them.
3. Basically the above, but auto-enumerate all of the outputs and try them all. This would keep the API simpler, but I guess on the fringe it'll force Nix to realize some outputs it wouldn't otherwise need if the lore override isn't cached.

I was imagining the second at first, but thinking it out a few steps has me leaning towards the first. WDYT? :)

pkgs/tools/misc/coreutils/default.nix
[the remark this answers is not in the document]
abathur: I've made this change in my local copy and it'll go up on next push.

pkgs/development/tools/analysis/binlore/default.nix
roberth: `passthru` is indeed intended to be added to all outputs, and it does not support customization per individual output.
Adding that is somewhat non-trivial, and `mkDerivation` needs to ditch `extendDerivation` for something more capable and integrated, which is arguably a good thing to do anyway. (Extending must never be done externally like that, but be handled by the abstraction that manages the fixpoint, which in this case is `mkDerivation` itself.)

pkgs/development/tools/analysis/binlore/default.nix
abathur: Marking resolved since I've ~optimistically changed this in the latest force-push. Can always revisit if it goes awry in some case.

pkgs/development/tools/analysis/binlore/default.nix
abathur: I explored the first approach, but I didn't like how verbose it was in `finalAttrs.finalPackage`  cases (and that it'd pull in extra dependencies unnecessarily in some cases).

Latest push adopts a 4th approach here that required less all-around change. `passthru.binlore` is now an attrset namespaced by output name. 

While there are no cases that need this behavior in the current set, I did test this out with a smaller change on top of nixpkgs-unstable by adding lore overrides for both glib.bin and glib.dev. I can push that up on a separate branch if anyone needs ~proof :)

Tentatively marking resolved, but I'm open to re-litigating the approach.

pkgs/development/tools/analysis/binlore/default.nix
abathur: Tentatively resolving since this rename is pushed now.

abathur:
@ofborg build resholve.tests

abathur:
@ofborg build resholve.tests

thiagokokada:
Looks good to me, but please fix the commit messages.

thiagokokada:
Eval error.

abathur:
Sounds like the problem this one's addressing:
- https://github.com/NixOS/nixpkgs/pull/324580

thiagokokada:
@ofborg eval

thiagokokada:
@ofborg eval

thiagokokada:
Eval is still failing. Maybe you need to rebase?

abathur:
It looks like the periodic merges into staging have been failing, so I think it isn't in staging yet: https://github.com/NixOS/nixpkgs/actions/workflows/periodic-merge-6h.yml

abathur:
@ofborg eval

PR #311836, opened 2024-05-15
openbsd: init at 7.5

ali-abrar:
## Description of changes

Add OpenBSD support similar to FreeBSD/NetBSD. We use some NetBSD tools to help bootstrap. Currently these are known o build:

```
nix-build -A pkgsCross.x86_64-openbsd.hello
nix-build -A pkgsCross.x86_64-openbsd.zlib
```

Actually *running* these OpenBSD executables on OpenBSD will take more work, but @ericson2314 thinks that is fine for now, because it is easier to fix that after there is a building foundation.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---


pkgs/development/compilers/llvm/17/default.nix
```diff
@@ -150,6 +150,9 @@ in let
           src = ../common/clang/clang-at-least-16-LLVMgold-path.patch;
           libllvmLibdir = "${tools.libllvm.lib}/lib";
         })
+      ] ++ lib.optionals stdenv.targetPlatform.isOpenBSD [
+        # TODO make unconditional
+        ../common/clang/openbsd-no-assume-path-exists.patch
```
Ericson2314, line 155: https://github.com/llvm/llvm-project/pull/92183

pkgs/development/compilers/llvm/17/default.nix
```diff
@@ -405,6 +408,11 @@ in let
           url = "https://github.com/macports/macports-ports/raw/acd8acb171f1658596ed1cf25da48d5b932e2d19/lang/llvm-17/files/0042-mbstate_t-not-defined.patch";
           hash = "sha256-jo+DYA6zuSv9OH3A0bYwY5TlkWprup4OKQ7rfK1WHBI=";
         })
+      ] ++ lib.optionals stdenv.hostPlatform.isOpenBSD [
+        # See https://lists.boost.org/boost-bugs/2015/07/41690.php and
+        # https://github.com/boostorg/log/commit/c98b1f459add14d5ce3e9e63e2469064601d7f71
+        # for a description of an analogou problem an fix for Boost.
+        ../common/libcxx/openbsd.patch
```
Ericson2314, line 415: https://github.com/llvm/llvm-project/pull/92186

pkgs/os-specific/bsd/openbsd/pkgs/make-rules/ar.patch
Ericson2314, line 1: Submitted an email to `tech@openbsd.org`. Hopefully it will appear shortly.

pkgs/os-specific/bsd/openbsd/pkgs/libc/package.nix
```diff
@@ -0,0 +1,117 @@
...
+    make -C $BSDSRCDIR/lib/librthread $makeFlags install
+    symlink_so pthread
+
+    make -C $BSDSRCDIR/lib/librpcsvc $makeFlags
+    make -C $BSDSRCDIR/lib/librpcsvc $makeFlags install
+    symlink_so rpcsv
+
+    make -C $BSDSRCDIR/lib/libutil $makeFlags
+    make -C $BSDSRCDIR/lib/libutil $makeFlags install
+    symlink_so util
```
rhelmot, line 113: What packages have you tried compiling with this libc? I found that there were actually a bunch more libraries on freebsd which were required pretty much constantly, once you got into more interesting applications, like rust packages and stuff built with meson.

pkgs/os-specific/bsd/netbsd/pkgs/make-rules.nix
```diff
@@ -0,0 +1,75 @@
...
+  fetchNetBSD,
+  stdenv,
+  bsdSetupHook,
+  netbsdSetupHook,
+  rsync,
+}:
+
+mkDerivation {
+  path = "share/mk";
+  sha256 = "0w9x77cfnm6zwy40slradzi0ip9gz80x6lk7pvnlxzsr2m5ra5sy";
```
rhelmot, line 13: does this hash need to be here?

pkgs/os-specific/bsd/openbsd/pkgs/mkDerivation.nix
```diff
@@ -0,0 +1,91 @@
...
+      # Since STRIP below is the flag
+      STRIPBIN = "${stdenv.cc.bintools.targetPrefix}strip";
+
+      makeFlags = [
+        "STRIP=-s" # flag to install, not command
+        "-B"
+      ];
+
+      # amd64 not x86_64 for this on unlike NetBSD
+      MACHINE_ARCH = "amd64";
```
rhelmot, line 55: You may, but are not required to, make this more generic sooner rather than later. I've heard openbsd has the best hardware support for other architectures out of all the bsds!

rhelmot (commented):
no showstoppers, just some questions and advice from the freebsd mines

pkgs/development/compilers/llvm/17/default.nix
Ericson2314: @RossComputerGuy I suppose maybe when they are merged to master, we can backport to at least 18 and switch to using that for this?

RossComputerGuy: @Ericson2314 Sounds good to me.

pkgs/os-specific/bsd/openbsd/pkgs/libc/package.nix
Ericson2314: Just the basics, we will also need to add more stuff

pkgs/os-specific/bsd/openbsd/pkgs/include/package.nix
```diff
@@ -0,0 +1,46 @@
...
+    rpcgen
+  ];
+
+  patches = [
+    # ./skip-rdirs.patch
+    ./no-perms.patch
+  ];
+
+  makeFlags = [
+    "RPCGEN_CPP=${buildPackages.stdenv.cc.cc}/bin/cpp"
```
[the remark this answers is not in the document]
Ericson2314: This build-time CPP (not that it really matters) is not on the path

pkgs/os-specific/bsd/openbsd/pkgs/make-rules/ar.patch
Ericson2314: https://marc.info/?l=openbsd-tech&m=171575284906018&w=2

pkgs/os-specific/bsd/openbsd/pkgs/include/package.nix
```diff
@@ -0,0 +1,46 @@
...
+  mtree,
+  pax,
+  buildPackages,
+}:
+mkDerivation {
+  path = "include";
+
+  extraPaths = [
+    "lib"
+    "sys"
```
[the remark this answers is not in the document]
Ericson2314: It is just some dirs now

pkgs/os-specific/bsd/netbsd/pkgs/make-rules.nix
Ericson2314: It does because that's NetBSD, which has many separate fetches (probably not good).

pkgs/development/compilers/llvm/common/libcxx/openbsd.patch
```diff
@@ -0,0 +1,127 @@
...
+
+    [libc++] Fix build on OpenBSD
+    
+    - No indirect syscalls on OpenBSD
+    
+    - Clock:
+    
+      See https://lists.boost.org/boost-bugs/2015/07/41690.php and
+      https://github.com/boostorg/log/commit/c98b1f459add14d5ce3e9e63e2469064601d7f71
+      for a description of an analogou problem an fix for Boost.
```
alyssais, line 13:
```suggestion
      for a description of an analogous problem an fix for Boost.
```

pkgs/os-specific/bsd/netbsd/pkgs/make-rules.nix
```diff
@@ -0,0 +1,75 @@
...
+  version = "9.2";
+  noCC = true;
+
+  buildInputs = [ ];
+  nativeBuildInputs = [
+    bsdSetupHook
+    netbsdSetupHook
+    rsync
```
alyssais, line 21: What's rsync used for?

pkgs/os-specific/bsd/netbsd/pkgs/make.nix
```diff
@@ -11,53 +12,12 @@ mkDerivation {
...
-         --replace /bin/rm rm
-    ''
-    + lib.optionalString stdenv.isDarwin ''
-      substituteInPlace $BSDSRCDIR/share/mk/bsd.sys.mk \
-        --replace '-Wl,--fatal-warnings' "" \
-        --replace '-Wl,--warn-shared-textrel' ""
+    make-rules.postPatch
+    + ''
+      # make needs this to pick up our sys make files
+      export NIX_CFLAGS_COMPILE+=" -D_PATH_DEFSYSPATH=\"$out/share/mk\""
```
alyssais, line 18: appendToVar?

pkgs/development/compilers/llvm/17/default.nix
```diff
@@ -150,6 +150,9 @@ in let
           src = ../common/clang/clang-at-least-16-LLVMgold-path.patch;
           libllvmLibdir = "${tools.libllvm.lib}/lib";
         })
+      ] ++ lib.optionals stdenv.targetPlatform.isOpenBSD [
+        # TODO make unconditional
```
alyssais, line 154: Is this a promise to submit a PR to staging making it unconditional as soon as this is merged to staging?

pkgs/development/compilers/llvm/17/default.nix
```diff
@@ -408,6 +408,11 @@ in let
           url = "https://github.com/macports/macports-ports/raw/acd8acb171f1658596ed1cf25da48d5b932e2d19/lang/llvm-17/files/0042-mbstate_t-not-defined.patch";
           hash = "sha256-jo+DYA6zuSv9OH3A0bYwY5TlkWprup4OKQ7rfK1WHBI=";
         })
+      ] ++ lib.optionals stdenv.hostPlatform.isOpenBSD [
```
alyssais, line 411: Why conditional?

pkgs/development/compilers/llvm/common/libcxx/default.nix
```diff
@@ -87,12 +87,6 @@ let
 
   cmakeFlags = [
     "-DLLVM_ENABLE_RUNTIMES=${lib.concatStringsSep ";" runtimes}"
-  ] ++ lib.optionals (useLLVM && !stdenv.hostPlatform.isWasm) [
-    # libcxxabi's CMake looks as though it treats -nostdlib++ as implying -nostdlib,
-    # but that does not appear to be the case for example when building
-    # pkgsLLVM.libcxxabi (which uses clangNoCompilerRtWithLibc).
-    "-DCMAKE_EXE_LINKER_FLAGS=-nostdlib"
-    "-DCMAKE_SHARED_LINKER_FLAGS=-nostdlib"
```
alyssais, line 95: Can this please be explained in the commit message rather than just a GitHub link that may disappear?

pkgs/development/compilers/llvm/common/compiler-rt/default.nix
```diff
@@ -149,7 +150,9 @@ stdenv.mkDerivation ({
     ln -s $out/lib/*/clang_rt.crtbegin_shared-*.o $out/lib/crtbeginS.o
     ln -s $out/lib/*/clang_rt.crtend_shared-*.o $out/lib/crtendS.o
   '' + lib.optionalString doFakeLibgcc ''
-     ln -s $out/lib/freebsd/libclang_rt.builtins-*.a $out/lib/libgcc.a
+     ln -s $out/lib/*/libclang_rt.builtins-*.a $out/lib/libgcc.a
+  '' + lib.optionalString forceLinkCompilerRt ''
+     ln -s $out/lib/*/libclang_rt.builtins-*.a $out/lib/libcompiler_rt.a
```
alyssais, line 155: Any idea why we need this?

pkgs/os-specific/bsd/openbsd/pkgs/include/no-perms.patch
```diff
@@ -0,0 +1,12 @@
+diff --git a/include/Makefile b/include/Makefile
+index a2d539f1c3b..c256c2e2b43 100644
+--- a/include/Makefile
++++ b/include/Makefile
+@@ -109,7 +109,6 @@ includes:
+ 		echo installing in ${.CURDIR}/$$i; \
+ 		cd ${.CURDIR}/$$i && ${RUN_MAKE}; \
+ 	done
+-	chown -RP ${BINOWN}:${BINGRP} ${DESTDIR}/usr/include
```
alyssais, line 9: Could we avoid the patch by setting BINOWN and BINGRP to the current user and group in a preBuild?

pkgs/os-specific/bsd/openbsd/pkgs/libc/disable-librebuild.patch
```diff
@@ -0,0 +1,12 @@
+diff --git a/lib/libc/Makefile b/lib/libc/Makefile
+index 4bb4b67fcbb..1c8a8e08e60 100644
+--- a/lib/libc/Makefile
++++ b/lib/libc/Makefile
+@@ -6,7 +6,6 @@
+ .include <bsd.own.mk>
+
+ LIB=c
+-LIBREBUILD=y
```
alyssais, line 9: We can't override this from the make command line?

pkgs/os-specific/bsd/openbsd/pkgs/libc/libc-no-tags.patch
```diff
@@ -0,0 +1,24 @@
...
+index 4bb4b67fcbb..daa9275d318 100644
+--- a/lib/libc/Makefile
++++ b/lib/libc/Makefile
+@@ -26,6 +26,10 @@ SYMBOL_LISTS=	${LIBCSRCDIR}/Symbols.list \
+ LIBCSRCDIR=${.CURDIR}
+ .include "${LIBCSRCDIR}/Makefile.inc"
+ 
++all: ${SRCS}
++
++LIBC_TAGS ?= yes
```
alyssais, line 11: Why `?=`?

pkgs/os-specific/bsd/openbsd/pkgs/libc/libpthread-no-werror.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/lib/librthread/Makefile b/lib/librthread/Makefile
+index 67a2678574d..6bc8d407d76 100644
+--- a/lib/librthread/Makefile
++++ b/lib/librthread/Makefile
+@@ -3,7 +3,7 @@
+ LIB=pthread
+ LIBCSRCDIR=    ${.CURDIR}/../libc
+
+-CFLAGS+=-Wall -g -Werror -Wshadow
++CFLAGS+=-Wall -g -Wshadow
```
alyssais, line 10: Could we use NIX_CFLAGS_COMPILE rather than a patch?

pkgs/os-specific/bsd/openbsd/pkgs/libc/package.nix
```diff
@@ -0,0 +1,117 @@
...
+    "lib/libm"
+    "lib/libpthread"
+    "lib/librpcsvc"
+    "lib/librpcsvc"
+    "lib/librthread"
+    "lib/libutil"
+  ];
+
+  patches = [
+    # NetBSD's make usess `${variable:tl}` not `${variable:L}`.
```
alyssais, line 37: Could we move this into the patch file (everything at the start is interpreted as a comment) so the explanation lives closer to the code change?

pkgs/os-specific/bsd/openbsd/pkgs/libc/package.nix
```diff
@@ -0,0 +1,117 @@
...
+    lorder
+    tsort
+  ];
+
+  buildInputs = [
+    include
+    csu
+  ];
+
+  env.NIX_CFLAGS_COMPILE = builtins.toString [ "-B${csu}/lib" ];
```
alyssais, line 63: Why make a single element list and then convert it back to a string?

pkgs/os-specific/bsd/openbsd/pkgs/make-rules/ar.patch
alyssais: Great! Let's fetchurl this from https://marc.info/?l=openbsd-tech&m=171575284906018&q=raw.

pkgs/os-specific/bsd/openbsd/pkgs/make-rules/package.nix
```diff
@@ -0,0 +1,28 @@
+{ mkDerivation, stdenv }:
+
+mkDerivation {
+  path = "share/mk";
+  noCC = true;
+
+  buildInputs = [ ];
+  nativeBuildInputs = [ ];
```
alyssais, line 8: Are these lines necessary?

alyssais (changes requested):
First, thank you so much for splitting this into small, easily reviewable commits.

How come we're using NetBSD make and working around its differences instead of just building OpenBSD make?

pkgs/os-specific/bsd/openbsd/pkgs/make-rules/package.nix
Ericson2314: Yes because there are default deps to be cancelled.

pkgs/os-specific/bsd/openbsd/pkgs/libc/package.nix
Ericson2314: There used to be more things in the list :sweat_smile:

pkgs/development/compilers/llvm/common/compiler-rt/default.nix
Ericson2314: I think the build system simply doesn't bother to make the "normally named" library when its only statically linking? I am a bit confused why too. But note how the previous hack also had to work from the more obscure `libclang_rt.builtins`.

pkgs/development/compilers/llvm/17/default.nix
Ericson2314: If we switch to LLVM 18 perhaps it doesn't even need to go to staging!

Ericson2314:
@alyssais 

> How come we're using NetBSD make and working around its differences instead of just building OpenBSD make?

OpenBSD doesn't have any `tools/compat` or similar way to build its programs with host platform != OpenBSD. So rather than fight hard to build the `nativeBuildInputs` we'd need, we just cut the Gordian knot and used NetBSD ones.

I am still chuckling about this :)

pkgs/os-specific/bsd/openbsd/pkgs/mkDerivation.nix
```diff
@@ -0,0 +1,99 @@
...
+
+      MACHINE_CPUARCH = MACHINE_ARCH;
+
+      COMPONENT_PATH = attrs.path or null;
+
+      strictDeps = true;
+
+      meta = with lib; {
+        maintainers = with maintainers; [ ericson2314 ];
+        platforms = platforms.unix;
```
alyssais, line 77: > OpenBSD doesn't have any tools/compat or similar way to build its programs with host platform != OpenBSD.

In that case, why are we aren't we defaulting this to OpenBSD?

Ericson2314: Good catch! :)

pkgs/os-specific/bsd/openbsd/pkgs/libc/disable-librebuild.patch
Ericson2314: it unfortunately just checks that the variable is defined, not that the variable value is `y`.

pkgs/os-specific/bsd/openbsd/pkgs/libc/libc-no-tags.patch
Ericson2314: To hint that it's just a default. I vaguely recall `?=` isn't as useful for BSD Make as it is for GNU Make, but I forget the details.

pkgs/os-specific/bsd/openbsd/pkgs/libc/libpthread-no-werror.patch
Ericson2314: With `-Wno-error`?

pkgs/os-specific/bsd/openbsd/pkgs/libc/libc-no-tags.patch
alyssais: Every make variable that doesn't use := is just a default.  The only different the ? makes is that it will prefer a value set in the environment (rather than as a make flag) to the default, which IMO is quite fragile unless it's a conventional environment variable like CC.

pkgs/os-specific/bsd/openbsd/pkgs/libc/libpthread-no-werror.patch
alyssais: Yeah.

pkgs/os-specific/bsd/netbsd/pkgs/make-rules.nix
Ericson2314: NetBSD's combining the extra paths, which is not as good as what we did for FreeBSD and OpenBSD. We should really switch it over to the same way of doing things as the others at some point.

pkgs/development/compilers/llvm/17/default.nix
Ericson2314: Just to avoid rebuilds. Like before I hope we can avoid that problem by switching to 18.

pkgs/os-specific/bsd/openbsd/pkgs/libc/libc-no-tags.patch
Ericson2314: Ah right, thanks!

pkgs/development/compilers/llvm/17/default.nix
Ericson2314: https://github.com/llvm/llvm-project/pull/92293 is a backport PR (that unlike the original I cannot merge). We can cherry-pick instead for extra precision for LLVM 18.

pkgs/os-specific/bsd/netbsd/pkgs/make-rules.nix
ali-abrar: I'd like to do that in a separate PR.

pkgs/development/compilers/llvm/common/libcxx/openbsd.patch
Ericson2314: (We'll be replacing this with `fetchpatch` of https://github.com/llvm/llvm-project/pull/92186)

pkgs/os-specific/bsd/netbsd/pkgs/make-rules.nix
Ericson2314: Oh yes, I meant that. It is definitely out of scope for this one,

pkgs/development/compilers/llvm/common/libcxx/openbsd.patch
ali-abrar: The typo also exists in pkgs/development/compilers/llvm/17/default.nix, and I've fixed it there.

pkgs/development/compilers/llvm/17/default.nix
ali-abrar: Switched to LLVM 18 with these patches.

pkgs/development/compilers/llvm/17/default.nix
ali-abrar: This is no longer present in 17, moved to 18 and fetched from the linked PR instead of using an in-tree patch.

pkgs/development/compilers/llvm/common/libcxx/openbsd.patch
ali-abrar: We're now using fetchpatch here.

pkgs/development/compilers/llvm/common/compiler-rt/default.nix
Ericson2314: I suppose we can add a comment explaining this

pkgs/os-specific/bsd/openbsd/pkgs/libc/libpthread-no-werror.patch
ali-abrar: done

pkgs/os-specific/bsd/openbsd/pkgs/include/no-perms.patch
ali-abrar: fixed

pkgs/development/compilers/llvm/common/libcxx/default.nix
Ericson2314: Done

pkgs/development/compilers/llvm/common/compiler-rt/default.nix
Ericson2314: Added those comments

Ericson2314:
OK all review items are addressed!

pkgs/os-specific/bsd/netbsd/pkgs/make-rules.nix
```diff
@@ -0,0 +1,75 @@
...
+        --replace '-g ''${RUMPBINGRP}' "" \
+        --replace '-o ''${DEBUGOWN}' "" \
+        --replace '-g ''${DEBUGGRP}' ""
+
+       substituteInPlace $BSDSRCDIR/share/mk/bsd.lib.mk \
+         --replace '_INSTRANLIB=''${empty(PRESERVE):?-a "''${RANLIB} -t":}' '_INSTRANLIB='
+       substituteInPlace $BSDSRCDIR/share/mk/bsd.kinc.mk \
+         --replace /bin/rm rm
+    ''
+    + lib.optionalString stdenv.targetPlatform.isDarwin ''
```
rrbutani, line 66: Was changing this from `stdenv.isDarwin` to `stdenv.targetPlatform.isDarwin` intentional?

pkgs/os-specific/bsd/netbsd/pkgs/makeMinimal.nix
```diff
@@ -22,7 +23,9 @@ mkDerivation {
 
   postPatch = ''
     patchShebangs $COMPONENT_PATH/configure
-    ${make.postPatch}
+
+    # make needs this to pick up our sys make files
+    export NIX_CFLAGS_COMPILE+=" -D_PATH_DEFSYSPATH=\"$out/share/mk\""
```
rrbutani, line 28: Can we use `appendToVar` here as well, for consistency?

```suggestion
    appendToVar NIX_CFLAGS_COMPILE -D_PATH_DEFSYSPATH="$out/share/mk"
```

pkgs/os-specific/bsd/openbsd/pkgs/libc/package.nix
```diff
@@ -0,0 +1,120 @@
...
+  ];
+
+  env.NIX_CFLAGS_COMPILE = builtins.toString [
+    "-B${csu}/lib"
+    "-Wno-error"
+  ];
+
+  # Suppress lld >= 16 undefined version errors
+  # https://github.com/freebsd/freebsd-src/commit/2ba84b4bcdd6012e8cfbf8a0d060a4438623a638
+  env.NIX_LDFLAGS = lib.optionalString (stdenv.targetPlatform.linker == "lld") "--undefined-version";
```
rrbutani, line 72: Apologies if I'm missing something; should this be `hostPlatform`?

rrbutani (commented):
The LLVM changes all look good to me now. I'm not really qualified to review the rest but I had a couple of questions:

pkgs/os-specific/bsd/netbsd/pkgs/make-rules.nix
Ericson2314: The make rules are always a build-time input to something, so I don't really get the sense of getting in the host platform of the rules themselves (i.e. the build platform of the thing to be built).

Ericson2314: I don't know the purpose of this bit though, which doesn't help either.

pkgs/os-specific/bsd/netbsd/pkgs/make.nix
```diff
@@ -11,53 +12,12 @@ mkDerivation {
...
-         --replace /bin/rm rm
-    ''
-    + lib.optionalString stdenv.isDarwin ''
-      substituteInPlace $BSDSRCDIR/share/mk/bsd.sys.mk \
-        --replace '-Wl,--fatal-warnings' "" \
-        --replace '-Wl,--warn-shared-textrel' ""
+    make-rules.postPatch
+    + ''
+      # make needs this to pick up our sys make files
+      appendToVar NIX_CFLAGS_COMPILE -D_PATH_DEFSYSPATH="$out/share/mk"
```
Ericson2314, line 18:
```suggestion
      appendToVar NIX_CFLAGS_COMPILE -D_PATH_DEFSYSPATH="$out/share/mk"
```
```suggestion
      appendToVar NIX_CFLAGS_COMPILE "-D_PATH_DEFSYSPATH=$out/share/mk"
```

pkgs/os-specific/bsd/openbsd/pkgs/libc/package.nix
Ericson2314: Yes it should

pkgs/development/compilers/llvm/18/default.nix
```diff
@@ -146,6 +146,11 @@ in let
         # https://reviews.llvm.org/D51899
         ./clang/gnu-install-dirs.patch
         ../common/clang/add-nostdlibinc-flag.patch
+        (fetchpatch {
+          url = "https://github.com/llvm/llvm-project/commit/5b09e6bf5aaeb21256e86455f8d6c42d01f71b95.patch";
```
alyssais, line 150: Shouldn't this be https://github.com/llvm/llvm-project/commit/48c1364200b5649dda2f9ccbe382b0bd908b99de, the commit that actually exists in that repo?

Would be nice for the patch to have a name too, so it's obvious what it's for.

pkgs/os-specific/bsd/openbsd/pkgs/source.nix
```diff
@@ -0,0 +1,8 @@
+{ fetchFromGitHub }:
+
+fetchFromGitHub {
+  owner = "openbsd";
+  repo = "src";
+  rev = "19e99d061340840dc52e10ae9c02d78a0a455689";
+  sha256 = "1061vb5m7k7q502bqsfrnnszjv9v66m0ksg5jka5jsdn6qa0fd9n";
+}
```
alyssais, line 8: I think we should fetch from the canonical CVS repo, not the GitHub mirror.

pkgs/os-specific/bsd/openbsd/pkgs/mkDerivation.nix
```diff
@@ -0,0 +1,91 @@
...
+
+      meta = with lib; {
+        maintainers = with maintainers; [ ericson2314 ];
+        platforms = platforms.openbsd;
+        license = licenses.bsd2;
+      };
+    }
+    // lib.optionalAttrs stdenv'.hasCC {
+      # TODO should CC wrapper set this?
+      CPP = "${stdenv'.cc.targetPrefix}cpp";
```
alyssais, line 83: No real need for this to be conditional, right?

pkgs/development/compilers/llvm/18/default.nix
```diff
@@ -396,7 +396,12 @@ in let
     # so: we use the clang from this LLVM package set instead of the regular
     # stdenv's compiler.
     libcxx = callPackage ../common/libcxx {
-      patches = lib.optionals (stdenv.isDarwin && lib.versionOlder stdenv.hostPlatform.darwinMinVersion "10.13") [
+      patches = [
+        (fetchpatch {
+          url = "https://github.com/llvm/llvm-project/commit/24ef89b6a3c2513221dc2e154642ce379cd701f6.patch";
```
alyssais, line 401: Similarly I think this should be https://github.com/llvm/llvm-project/commit/1118c2e05e67a36ed8ca250524525cdb66a55256.

pkgs/development/compilers/llvm/18/default.nix
ali-abrar: this is done

pkgs/development/compilers/llvm/18/default.nix
ali-abrar: done

pkgs/os-specific/bsd/openbsd/pkgs/mkDerivation.nix
ali-abrar: fixed

Ericson2314: Actually yes, because if we don't have a CC then we don't have a `stdenv'.cc.targetPrefix`. And trying to be clever about this can introduce infinite recursion.

pkgs/os-specific/bsd/openbsd/pkgs/source.nix
Ericson2314: We're having trouble fetching this, e.g. 

```bash
cvs -d :pserver:anoncvs@anoncvs4.usa.openbsd.org:/cvs checkout -rOPENBSD_7_5 -P src
```
(with a shell with `openssh` `cvs` `inetutils`) is not working.

Ericson2314 (approved):
I have a feeling moving from Git is not worth it.

pkgs/os-specific/bsd/openbsd/pkgs/source.nix
alyssais: This looks like it's working for me:

```nix
fetchcvs {
  cvsRoot = "anoncvs@anoncvs.fr.openbsd.org/cvs";
  module = "src";
  tag = "OPENBSD_7_5";
  sha256 = lib.fakeSha256;
}
```

In a shell you'd have to set CVS_RSH=ssh to match the behaviour of fetchcvs.

But actually, why can't we just use the source tarball they publish?

https://cdn.openbsd.org/pub/OpenBSD/7.5/src.tar.gz

Ericson2314: > In a shell you'd have to set CVS_RSH=ssh to match the behaviour of fetchcvs.

oh!

> But actually, why can't we just use the source tarball they publish?

Oh great. I like the look of "CDN". Yes let's do that.

alyssais: Probably wouldn't be too bad using `srcs`, but this is also fine.

pkgs/os-specific/bsd/openbsd/pkgs/source.nix
```diff
@@ -0,0 +1,8 @@
+{ lib, fetchcvs }:
```
alyssais, line 1: Presumably forgot to remove this after using `lib.fakeSha256`? :)

alyssais (changes requested):

alyssais (approved):

afh:
Glad to see this PR has been merged, thank you for your work, @ali-abrar, can't wait to try it out! 💯

foolnotion:
Sorry but this seems to break `clang-18.1.6` on linux:

```
❯ nix log /nix/store/synw5z7kjmp7gmg4knh4ikaffprbzr3m-clang-18.1.6.drv^\*
@nix { "action": "setPhase", "phase": "unpackPhase" }
Running phase: unpackPhase
unpacking source archive /nix/store/pkdznhza3cbgdixj2czjw6qi3v8wfpir-clang-src-18.1.6
source root is clang-src-18.1.6/clang
@nix { "action": "setPhase", "phase": "patchPhase" }
Running phase: patchPhase
applying patch /nix/store/4qh1l55wx654v0ls51dg1wdi5xy93w9b-purity.patch
patching file lib/Driver/ToolChains/Gnu.cpp
applying patch /nix/store/fk8qk07cl9m7wkjb0xc0hphbh3vrxfqh-gnu-install-dirs.patch
patching file cmake/modules/AddClang.cmake
patching file lib/Headers/CMakeLists.txt
Hunk #1 succeeded at 498 (offset 25 lines).
patching file tools/libclang/CMakeLists.txt
Hunk #1 succeeded at 246 (offset 12 lines).
patching file tools/scan-build-py/CMakeLists.txt
applying patch /nix/store/gxqzh9x4wmiv4jpbvla3z7i26zkkacqd-add-nostdlibinc-flag.patch
patching file lib/Driver/Driver.cpp
Hunk #1 succeeded at 499 (offset 8 lines).
applying patch /nix/store/ycw68pzfjx00fm99rqlvfnn7hssnzakc-48c1364200b5649dda2f9ccbe382b0bd908b99de.patch
patching file lib/Driver/ToolChains/OpenBSD.cpp
Reversed (or previously applied) patch detected!  Assume -R? [n] 
Apply anyway? [n] 
Skipping patch.
1 out of 1 hunk ignored -- saving rejects to file lib/Driver/ToolChains/OpenBSD.cpp.rej
```

alyssais:
Ah I wonder if that's because we updated LLVM and now the patch is already there and doesn't need backporting any more.

paparodeo:
patches reverted in: https://github.com/NixOS/nixpkgs/pull/315790
patches were the last 2 changes applied before 18.1.6 was tagged.
https://github.com/llvm/llvm-project/commits/llvmorg-18.1.6/

Ericson2314:
Sorry about that!

PR #311872, opened 2024-05-15
Revert "nixos/garage: drop replication_mode setting"

yu-re-ka:
Reverts NixOS/nixpkgs#309643

Broke the NixOS test

```
node4 # [   16.586440] garage[858]: Error: TOML decode error: TOML parse error at line 3, column 20
node4 # [   16.588235] garage[858]:   |
node4 # [   16.588473] garage[858]: 3 | replication_mode = 3
node4 # [   16.588802] garage[858]:   |                    ^
node4 # [   16.589127] garage[858]: invalid type: integer `3`, expected a string


flokli:
I think we need some time porting the test harness to support Garage 1.x, yes.

PR #294601, opened 2024-03-10
electron: add ozoneFlags to passthru

getchoo:
## Description of changes

There are two goals for this PR:

- Create a standard attribute future packagers can reference to avoid mistakes when implementing `NIXOS_OZONE_WL` in electron apps
  - This is similar to `addOpenGLRunpath.driverLink` and how it's sometimes used
- Allow for easier treewide maintenance when it comes to introducing and removing flags
- Allow for the arguments passed when `NIXOS_OZONE_WL` is set to be easily overriden at a global level
  - A good use case for this currently is passing `--enable-wayland-ime` to all apps -- though this PR makes that a default
  - Thanks to @pluiedev for this idea :)

This *should* be a treewide change, though i'd imagine some apps that bundle their electron or use an older version could have issues here. The initial treewide changes are also mainly the result of `sed` and some manual `grep`ing, so this will be a draft until I can get all changed packages to at least evaluate

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
- For non-Linux: Is sandboxing enabled in `nix.conf`? (See [Nix manual](https://nixos.org/manual/nix/stable/command-ref/conf-file.html))
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:
  - [NixOS test(s)](https://nixos.org/manual/nixos/unstable/index.html#sec-nixos-tests) (look inside [nixos/tests](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests))
  - and/or [package tests](https://nixos.org/manual/nixpkgs/unstable/#sec-package-tests)
  - or, for functions and "core" functionality, tests in [lib/tests](https://github.com/NixOS/nixpkgs/blob/master/lib/tests) or [pkgs/test](https://github.com/NixOS/nixpkgs/blob/master/pkgs/test)
  - made sure NixOS tests are [linked](https://nixos.org/manual/nixpkgs/unstable/#ssec-nixos-tests-linking) to the relevant packages
- [ ] Tested compilation of all packages that depend on this change using `nix-shell -p nixpkgs-review --run "nixpkgs-review rev HEAD"`. Note: all changes have to be committed, also see [nixpkgs-review usage](https://github.com/Mic92/nixpkgs-review#usage)
- [ ] Tested basic functionality of all binary files (usually in `./result/bin/`)
[...]

PR #311937, opened 2024-05-15
webcord: Implement ${commandLineArgs}

Kreyren:
## Description of changes

Taking inspiration from the packaging of `brave`:

https://github.com/NixOS/nixpkgs/blob/6308cb8b066539d2fbc7637d2de02349e985e177/pkgs/applications/networking/browsers/brave/make-brave.nix#L49

To enable configuration alike:

https://github.com/Kreyren/nixos-config/tree/e16b1776ffca750cd7ed872efa0becf279780ab3/src/nixos/users/kreyren/home/machines/pelagus/home-configuration.nix#L54-L57

To manage a scenario where the system is using a system-proxy which is causing issues for the user while avoiding doing impure declaration through using `environment.localBinInPath` and making script:

```sh
#!/usr/bin/env sh

# Instruct webcord's electron to not use system proxy
/home/USER/.nix-profile/bin/webcord --no-proxy-server $*
```

## Things done

- Built on platform(s)
  - [X] x86_64-linux
  - [-] aarch64-linux
  - [-] x86_64-darwin
  - [-] aarch64-darwin
  - [-] `sandbox = relaxed`
  - [-] `sandbox = true`
- [-] Tested, as applicable:
- [-] Tested compilation of all packages that depend on this change using `nix-shell -p nixpkgs-review --run "nixpkgs-review rev HEAD"`. Note: all changes have to be committed, also see [nixpkgs-review usage](https://github.com/Mic92/nixpkgs-review#usage)
- [-] Tested basic functionality of all binary files (usually in `./result/bin/`)
  - [X] (Package updates) Added a release notes entry if the change is major or breaking
  - [-] (Module updates) Added a release notes entry if the change is significant
  - [-] (Module addition) Added a release notes entry if adding a new NixOS module

---

### DIFF
--- a/pkgs/by-name/we/webcord/package.nix
+++ b/pkgs/by-name/we/webcord/package.nix
@@ -1,12 +1,15 @@
 {
-  lib,
-  buildNpmPackage,
-  fetchFromGitHub,
-  copyDesktopItems,
-  python3,
-  xdg-utils,
-  electron,
-  makeDesktopItem,
+  lib
+  , buildNpmPackage
+  , fetchFromGitHub
+  , copyDesktopItems
+  , python3
+  , xdg-utils
+  , electron
+  , makeDesktopItem
+
+  # command line arguments which are always set e.g "--no-proxy-server"
+  , commandLineArgs ? ""
 }:
 
 buildNpmPackage rec {
@@ -56,7 +59,8 @@ buildNpmPackage rec {
       makeWrapper '${lib.getExe electron}' $out/bin/webcord \
         --suffix PATH : "${binPath}" \
         --add-flags "\''${NIXOS_OZONE_WL:+\''${WAYLAND_DISPLAY:+--ozone-platform-hint=auto --enable-features=WaylandWindowDecorations}}" \
-        --add-flags $out/lib/node_modules/webcord/
+        --add-flags $out/lib/node_modules/webcord/ \
+        --add-flags ${lib.escapeShellArg commandLineArgs}
 
       runHook postInstall
     '';


eclairevoyant:
Not sure I see the point, sure we could add this kind of arg to every GUI app, but I think it makes more sense for the user to just use e.g. `wrapProgram` on their own rather than overriding (thereby rebuilding) the app itself.

BTW, the file is formatted with nixfmt-rfc-style, please don't reformat it.

huantianad:
There's also #294601 as a solution to consider, which I kinda like more, but I'm not sure what the status of is.

Kreyren:
> Not sure I see the point, sure we could add this kind of arg to every GUI app, but I think it makes more sense for the user to just use  e.g. `wrapProgram` on their own rather than overriding (thereby rebuilding) the app itself. -- @eclairevoyant (https://github.com/NixOS/nixpkgs/pull/311937#issuecomment-2112756777)

Can you elaborate? The problem is to configure the package in a declarative way to not use system proxy or use non-system proxy

Tbh i kept it as a draft, bcs i wanted to brainstorm this in nixpkgs contributors as i also don't think this is a good idea in terms of management for that same reason

> BTW, the file is formatted with nixfmt-rfc-style, please don't reformat it with some other formatter. -- @eclairevoyant (https://github.com/NixOS/nixpkgs/pull/311937#issuecomment-2112756777)

My editor should automatically do that, thanks for highlighting that i will investigate where is the issue with it.

Kreyren:
> There's also #294601 as a solution to consider, which I kinda like more, but I'm not sure what the status of is. -- @huantianad (https://github.com/NixOS/nixpkgs/pull/311937#issuecomment-2113301095)

if i understand the PR correctly this seems to be already implemented? But i don't see an easy way to utilize this in a nix configs.

https://github.com/NixOS/nixpkgs/pull/311937#issuecomment-2113301095

https://github.com/NixOS/nixpkgs/blob/eb2449cce5170f2fec6b35456bd5057afffb9cdb/pkgs/by-name/we/webcord/package.nix#L61

eclairevoyant:
> Can you elaborate?

https://nixos.org/manual/nixpkgs/stable/#fun-wrapProgram
Use this in a new package that depends on webcord (or whatever program you want to wrap)

Kreyren:
Agree that wrapping that is better implementation, implemented as https://github.com/NiXium-org/NiXium/blob/ec5a7e8b4f113de4cbd295221d6265e52454d629/src/nixos/users/kira/home/machines/tupac/home-configuration.nix#L34-L51

PR #312253, opened 2024-05-16
nixos/nvidia-container-toolkit: add option for driver path

loicreynier:
## Description of changes

Add an option `nvidia-driver-path` (in `hardware.nvidia-container-toolkit`) for systems where NVIDIA drivers are not set up by NixOS. This is useful for example in WSL systems where the Windows drivers can be used and are located in `/usr/lib/wsl/lib`.

Tested with Podman

```bash
podman run --rm --device=nvidia.com/gpu=all \
    nvcr.io/nvidia/nvhpc:24.3-devel-cuda_multi-ubuntu22.04 \
    nvidia-smi -L
```

on a WSL NixOS system with 

```nix
hardware.nvidia-container-toolkit = {
  enable = true;
  nvidia-driver-path = "/usr/lib/wsl";
  mount-nvidia-executables = false;
  mount-nvidia-docker-1-directories = false;
};
```

Also tested the compilation and run of [simple CUDA programs](https://github.com/loicreynier/sandbox/tree/main/cuda-demo) on that container.

Not sure yet if this option should always disables the `mount-nvidia-executables` and `mount-nvidia-docker-1-directories
` options.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/nixos/modules/services/hardware/nvidia-container-toolkit/cdi-generate.nix
+++ b/nixos/modules/services/hardware/nvidia-container-toolkit/cdi-generate.nix
@@ -17,10 +17,13 @@
     (mount:
       ["${lib.getExe jq} '${jqAddMountExpression} ${builtins.toJSON (mkMount mount)}'"])
     mounts;
+  ldLibraryPathExport = if builtins.isString nvidia-driver then "export LD_LIBRARY_PATH=${nvidia-driver}/lib" else "";
 in
 writeScriptBin "nvidia-cdi-generator"
 ''
 #! ${runtimeShell}
+
+${ldLibraryPathExport}
 
 function cdiGenerate {
   ${lib.getExe' nvidia-container-toolkit "nvidia-ctk"} cdi generate \
--- a/nixos/modules/services/hardware/nvidia-container-toolkit/default.nix
+++ b/nixos/modules/services/hardware/nvidia-container-toolkit/default.nix
@@ -37,6 +37,16 @@
         '';
       };
 
+      nvidia-driver-path = lib.mkOption {
+        type = with lib.types; nullOr path;
+        default = null;
+        description = ''
+          Custom NVidia driver path for systems not using NixOS to setup NVidia drivers.
+
+          On WSL for example, it should be set to /usr/lib/wsl.
+        '';
+      };
+
       mounts = lib.mkOption {
         type = lib.types.listOf (lib.types.submodule mountType);
         default = [];
@@ -103,7 +113,11 @@
            containerPath = "/usr/local/nvidia/lib64"; }])
     ]);
 
-    systemd.services.nvidia-container-toolkit-cdi-generator = lib.mkIf config.hardware.nvidia-container-toolkit.enable {
+    systemd.services.nvidia-container-toolkit-cdi-generator = let
+      nvidia-driver = if config.hardware.nvidia-container-toolkit.nvidia-driver-path == null
+        then config.hardware.nvidia.package
+        else config.hardware.nvidia-container-toolkit.nvidia-driver-path;
+    in lib.mkIf config.hardware.nvidia-container-toolkit.enable {
       description = "Container Device Interface (CDI) for Nvidia generator";
       wantedBy = [ "multi-user.target" ];
       after = [ "systemd-udev-settle.service" ];
@@ -114,7 +128,7 @@
           let
             script = pkgs.callPackage ./cdi-generate.nix {
               inherit (config.hardware.nvidia-container-toolkit) mounts;
-              nvidia-driver = config.hardware.nvidia.package;
+              inherit nvidia-driver;
             };
           in
           lib.getExe script;


nixos/modules/services/hardware/nvidia-container-toolkit/default.nix
```diff
@@ -37,6 +37,16 @@
         '';
       };
 
+      nvidia-driver-path = lib.mkOption {
+        type = with lib.types; nullOr path;
+        default = null;
+        description = ''
+          Custom NVidia driver path for systems not using NixOS to setup NVidia drivers.
```
ereslibre, line 44:
```suggestion
          Custom Nvidia driver path for systems not using NixOS to setup Nvidia drivers.
```

ereslibre: Just to be consistent with current casing.

nixos/modules/services/hardware/nvidia-container-toolkit/cdi-generate.nix
```diff
@@ -17,11 +17,14 @@
     (mount:
       ["${lib.getExe jq} '${jqAddMountExpression} ${builtins.toJSON (mkMount mount)}'"])
     mounts;
+  ldLibraryPathExport = if builtins.isString nvidia-driver then "export LD_LIBRARY_PATH=${nvidia-driver}/lib" else "";
```
ereslibre, line 20: Why this behavior only when the provided path is a string? Please explain a little further how this new attribute might be used.

I would like to avoid surprises when `nvidia-driver` is something other than `null` and a string, we will silently fallback to not use the provided parameter. Is there any chance for the user to have provided a nix path?

Also, can we use a flag of `nvidia-ctk cdi generate` rather than using an envvar? I see `--driver-root`; others might be relevant too. I think this is more explicit and lets less room for unexpected behaviors.

ereslibre (commented):
Thank you @loicreynier, trying to understand a bit here, since I'm not familiar with WSL myself.

We had a related case at https://github.com/nix-community/NixOS-WSL/issues/433; how is it different from this, does enabling `wsl.useWindowsDriver` work for you? Or is this PR needed even in that case?

loicreynier:
Thanks for pointing out https://github.com/nix-community/NixOS-WSL/issues/433. I wasn't aware that NixOS WSL had this `wsl.useWindowsDriver` which solves the issue. 

Linking the Window driver files (as done by this option) is probably a better approach than what I proposed here. I don't think this PR is needed anymore.

For reference, I couldn't make it work (without the above mentioned option) using the `--driver-root` and `--library-search-path` flags from `nvidia-ctk cdi generate`: 

```shell
nvidia-ctk --debug cdi generate \
  --discovery-mode wsl \
  --driver-root /usr/lib/wsl/lib/ \
  --library-search-path /usr/lib/wsl/lib/
```

<details><summary>Output</summary>

```
DEBU[0000] Locating NVIDIA Container Toolkit CLI as nvidia-ctk
DEBU[0000] Locating "nvidia-ctk" in ...
DEBU[0000] Checking candidate '/run/current-system/sw/bin/nvidia-ctk'
DEBU[0000] Found 1 candidates; ignoring further candidates
DEBU[0000] Found nvidia-ctk candidates: [/run/current-system/sw/bin/nvidia-ctk]
DEBU[0000] Using NVIDIA Container Toolkit CLI path nvidia-ctk
DEBU[0000] Locating /dev/dxg
DEBU[0000] Locating "/dev/dxg" in [/usr/lib/wsl/lib /usr/lib/wsl/lib/dev]
WARN[0000] Could not locate /dev/dxg: pattern /dev/dxg not found
ERRO[0000] failed to generate CDI spec: failed to create edits common for entities: failed to create discoverer for WSL driver: failed to initialize dxcore: failed to initialize dxcore context
```

</details>

ereslibre:
> Thanks for pointing out https://github.com/nix-community/NixOS-WSL/issues/433. I wasn't aware that NixOS WSL had this wsl.useWindowsDriver which solves the issue.

I cannot help but think that there must be something we can do (at the NixOS-WSL level) to improve this user experience. If:

- `config.wsl.enable` is true, and
- `config.hardware.nvidia-container-toolkit.enable` is true,

Then, we can set `wsl.useWindowsDriver = true`, would that be reasonable from the NixOS-WSL standpoint?

> For reference, I couldn't make it work (without the above mentioned option) using the --driver-root and --library-search-path flags from nvidia-ctk cdi generate:

Thanks for the input.

loicreynier:
> Then, we can set `wsl.useWindowsDriver = true`, would that be reasonable from the NixOS-WSL standpoint?

I think it would be reasonable since, as far as I know, the easiest way to make it work is to implement something similar to what this option does.

Additionally, I think it is worth noting that since WSL aims to simplify the interaction between Linux and the Windows host as much as possible, we should avoid duplicates and rely solely on the native drivers.

ereslibre:
> I think it would be reasonable since, as far as I know, the easiest way to make it work is to implement something similar to what this option does.

Thank you, I have created https://github.com/nix-community/NixOS-WSL/pull/478. Let me know if that looks fine or if you think it can be improved somehow.

> Additionally, I think it is worth noting that since WSL aims to simplify the interaction between Linux and the Windows host as much as possible, we should avoid duplicates and rely solely on the native drivers.

I might be missing here a bit, sorry. With this comment you mean that you encourage to use the `useWindowsDriver=true` option? This is what you mean with the native drivers, right? Or am I misunderstanding?

loicreynier:
> I might be missing here a bit, sorry. With this comment you mean that you encourage to use the `useWindowsDriver=true` option? This is what you mean with the native drivers, right? Or am I misunderstanding?

Yes exactly this is what I meant, sorry for the unclear comment.

Issue #162562, opened 2022-03-03
Packaging Request: GameScope


Krutonium:
**Project description**
This is the Wayland based compositor that Valve is using, that when paired with a running application, can fix minimizing and other such issues, as well as allowing AMD FSR on any application.

**Metadata**

* homepage URL: https://github.com/Plagman/gamescope
* source URL: https://github.com/Plagman/gamescope
* license: BSD 2 Clause
* platforms: Linux

SuperSamus:
For those interested, it's available in the NUR: https://github.com/nix-community/nur-combined/blob/master/repos/dukzcry/pkgs/gamescope.nix.

AtilaSaraiva:
I can take care of porting the NUR derivation to nixpkgs and becoming the official maintainer. Just need a few weeks to understand the codebase a bit before committing to it.

IvarWithoutBones:
Are you still interested in doing that? Would love to see this in nixpkgs :)

SuperSamus:
I read on NixOS's discord that https://github.com/Plagman/gamescope/issues/380 is a blocker.

AtilaSaraiva:
Yeah, that issue kinda upset me. I've been accompanying its progress but not being able to use controller for gamescope games suck. I do intend to free up some time to work on this. The NUR package was very oudated and not so well written at the time I checked it out, so I'd most likely would have to refactor it, which would take some effort.

Atemu:
@samueldr managed to package it for their project [here](https://github.com/Jovian-Experiments/Jovian-NixOS/blob/d27526e7917671ad3a64f823f4b95b761f96807d/pkgs/gamescope/default.nix). They opted to re-build the submodules using the Nix packages' srcs which should be remedied for inclusion into Nixpkgs.

AtilaSaraiva:
Oh nice, that helps a lot.

samueldr:
I did it this way as a shortcut.

It would probably be best to have an "updater" script next to the derivation that builds a `srcs.nix` attrset with the same exact source trees the git submodules point to. Though I haven't seen any weirdness, so I don't think that it would matter currently.

Atemu:
Of course, yes.

I think it'd be sufficient to make gamescope dependent on the output paths instead of the subproject build result; just like any other dependency.  
Should gamescope require special versions of those dependencies in the future, we should be able to easily provide those.

[...]

PR #312268, opened 2024-05-16
nixos/steam: add option `extraPackages` `fontPackages`, fix CJK fonts on client

### LINKED ISSUE #355: Characters are displayed as boxes in some languages in Steam.
In some languages, like Japanese and Chinese, characters will be displayed as boxes in either Game Mode or Desktop Mode in Steam. As mentioned in the [ArchWiki](https://wiki.archlinux.org/title/Steam), Steam requires the `wqy_zenhei` package to display CJK characters. Additionally, I found that Steam is unable to find fonts correctly on NixOS. I followed the [NixOS Wiki](https://nixos.wiki/wiki/Fonts) to configure fonts for Flatpak in an attempt to solve this problem and discovered that Steam searches `$HOME/.local/share/fonts` to find fonts. I think it would be better to provide a systemd service to create a symbolic link to the current profile's fonts folder so that Steam can find fonts correctly.

902a3f407a849f4f1a759d3d353a1cf

```Nix
  i18n.fonts.fontDir.enable = true;
  systemd.user.services.link-fonts = {
    description = "Create symbolic link for fonts";
    wantedBy = [ "default.target" "graphical-session.target" ];
    before = [ "graphical-session.target" ];
    serviceConfig = {
      Type = "oneshot";
      RemainAfterExit = true;
      ExecStartPre = ''
        ${pkgs.bash}/bin/bash -c "${pkgs.coreutils}/bin/test ! -e %h/.local/share && ${pkgs.coreutils}/bin/mkdir -p %h/.local/share || ${pkgs.coreutils}/bin/true"
      '';
      ExecStart = ''
        ${pkgs.bash}/bin/bash -c "${pkgs.coreutils}/bin/test ! -e %h/.local/share/fonts && ${pkgs.coreutils}/bin/ln -s /run/current-system/sw/share/X11/fonts %h/.local/share/fonts || ${pkgs.coreutils}/bin/true"
      '';
    };
  };
```

### LINKED ISSUE #178121: Steam client Asian fonts issue
Asian characters are all rendered as squares in client(not in games). Looks like steam client is not picking up system font configs?

### Steps To Reproduce
```
 ;; with latest unstable branch
  programs.steam.enable = true;
```
### Screenshots
image
image

### Metadata

```console
 - system: `"x86_64-linux"`
 - host os: `Linux 5.17.14, NixOS, 22.11 (Raccoon), 22.11.20220614.6616de3`
 - multi-user?: `yes`
 - sandbox: `yes`
 - version: `nix-env (Nix) 2.9.1`

```

### LINKED ISSUE #214275: steam: gamescope fails to launch when used from within steam
### Describe the bug
gamescope crashes with this error when run from within steam's fhsenv:
 ```
 wlserver: Running compositor on wayland display 'gamescope-0'
wlserver: [backend/headless/backend.c:18] Starting headless backend
wlserver: [xwayland/sockets.c:99] /tmp/.X11-unix not owned by root or us
wlserver: [xwayland/sockets.c:99] /tmp/.X11-unix not owned by root or us
[...]
wlserver: [xwayland/sockets.c:99] /tmp/.X11-unix not owned by root or us
wlserver: [xwayland/sockets.c:216] No display available in the first 33
```
This check in wlroots causes that error
https://gitlab.freedesktop.org/wlroots/wlroots/-/blob/0.15.1/xwayland/sockets.c#L98
because inside steam's fhsenv the  /tmp/.X11-unix directory is owned by nobody:nogroup.
I have tried simply removing the check from wlroots but in that case gamescope simply hangs.
(Note that it's gamescope that depends on wlroots so this bug happens under x11 window managers too.)

### Steps To Reproduce
1. start steam
2. add `gamescope -- %command%` to the launch options of any game
3. run the game

### Notify maintainers
@jagajaga @jonringer 

### Metadata
```console
 - system: `"x86_64-linux"`
 - host os: `Linux 6.1.7, NixOS, 23.05 (Stoat), 23.05pre446447.1b1f50645af`
 - multi-user?: `yes`
 - sandbox: `yes`
 - version: `nix-env (Nix) 2.12.0`
 - channels(root): `"nixos"`
 - channels(hadis): `"home-manager, nixos"`
```

Cryolitia:
## Description of changes

1. Add a new option `programs.steam.extraPackages` to add to the Steam environment.
1. Add `config.fonts.packages` to `extraPackages` to fix all Chinese and Japanese text in the Steam interface display as boxes ("tofu"). As we use the existing fonts on the system, this avoids closure impact. It was first came up by #302591 that reverted by #302748
1. ~~Add dependencies required by gamescope, to fix undefined symbols in X11 session. Came up by @BillHuang2001 in https://github.com/NixOS/nixpkgs/issues/162562#issuecomment-1229444338 .~~
    ~~- fix #214275~~
1. Format `nixos/modules/programs/steam.nix` by `nixfmt-rfc-style`
1. Add a new option `programs.steam.fontPackages`, just let user to choose whatever font they like, to end argue in #302862 
   - close #302862
   - fix #178121
   - fix https://github.com/Jovian-Experiments/Jovian-NixOS/issues/355

This PR has tested on my machine.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

CC:

- Steam maintainer team: @Atemu @eclairevoyant @jonringer @K900 @mkg20001 
- #302591 author: @eclairevoyant 
- @BillHuang2001

---

### DIFF
--- a/nixos/modules/programs/steam.nix
+++ b/nixos/modules/programs/steam.nix
@@ -102,6 +102,18 @@ in {
       '';
     };
 
+    fontPackages = lib.mkOption {
+      type = lib.types.listOf lib.types.package;
+      default = config.fonts.packages;
+      defaultText = lib.literalExpression "fonts.packages";
+      example = lib.literalExpression "with pkgs; [ source-han-sans ]";
+      description = ''
+        Font packages to use in Steam.
+
+        Defaults to system fonts, but could be overridden to use other fonts — useful for users who would like to customize CJK fonts used in Steam. According to the [upstream issue](https://github.com/ValveSoftware/steam-for-linux/issues/10422#issuecomment-1944396010), Steam only follows the per-user fontconfig configuration.
+      '';
+    };
+
     remotePlay.openFirewall = lib.mkOption {
       type = lib.types.bool;
       default = false;
@@ -178,6 +190,8 @@ in {
         setuid = true;
       };
     };
+
+    programs.steam.extraPackages = cfg.fontPackages;
 
     programs.gamescope.enable = lib.mkDefault cfg.gamescopeSession.enable;
     services.displayManager.sessionPackages = lib.mkIf cfg.gamescopeSession.enable [ gamescopeSessionFile ];
--- a/nixos/modules/programs/steam.nix
+++ b/nixos/modules/programs/steam.nix
@@ -55,6 +55,7 @@ in {
             then [ package ] ++ extraPackages
             else [ package32 ] ++ extraPackages32;
         in prevLibs ++ additionalLibs;
+        extraPkgs = p: (cfg.extraPackages ++ lib.optionals (prev ? extraPkgs) (prev.extraPkgs p));
       } // lib.optionalAttrs (cfg.gamescopeSession.enable && gamescopeCfg.capSysNice)
       {
         buildFHSEnv = pkgs.buildFHSEnv.override {
@@ -68,6 +69,19 @@ in {
 
         Use this option to customise the Steam package rather than adding your
         custom Steam to {option}`environment.systemPackages` yourself.
+      '';
+    };
+
+    extraPackages = lib.mkOption {
+      type = lib.types.listOf lib.types.package;
+      default = [ ];
+      example = lib.literalExpression ''
+        with pkgs; [
+          gamescope
+        ]
+      '';
+      description = ''
+        Additional packages to add to the Steam environment.
       '';
     };
 


nixos/modules/programs/steam.nix
```diff
@@ -110,6 +110,16 @@ in
       '';
     };
 
+    fontPackages = lib.mkOption {
+      type = lib.types.listOf lib.types.package;
+      default = config.fonts.packages;
+      defaultText = lib.literalExpression "fonts.packages";
+      example = lib.literalExpression "with pkgs; [ source-han-sans ]";
+      description = ''
+        Font packages to use in Steam, it is helpful for Asian users to add CJK fonts.
```
pluiedev, line 119:
```suggestion
        Font packages to use in Steam.
        
        Defaults to system fonts, but could be overridden to use other fonts — useful for users who would like to customize CJK fonts used in Steam.
```

pluiedev (changes requested):
Slight wording change

nixos/modules/programs/steam.nix
Cryolitia: thx

pluiedev (approved):

nixos/modules/programs/steam.nix
```diff
@@ -69,6 +75,19 @@ in {
       '';
     };
 
+    extraPackages = lib.mkOption {
```
K900, line 100: This should probably be a function as well, no?

nixos/modules/programs/steam.nix
```diff
@@ -177,7 +177,21 @@ in {
       };
     };
 
-    programs.steam.extraPackages = config.fonts.packages;
+    programs.steam.extraPackages = config.fonts.packages ++ (
```
K900: This seems extremely jank. Are you sure we need all of those? And why?

nixos/modules/programs/steam.nix
Cryolitia: No. For package's argument  `extraPkgs` , we may not have the same `pkgs` as which called the pcakcge itself while overriding. So providing a function to pass `pkgs` is necessary.

For NixOS option 'extraPackages', users should always get the same `pkgs` in their NixOS configuration globally, as the flake input or `<nixpkgs>` .

I have checked the already existing `extraPackages` in NixOS Search - options, most of their types is list of packages, the remain, whose type are function that evaluates to a(n) list of package, has all argument is `pythonPackages` or `haskellPakcages`.

nixos/modules/programs/steam.nix
Cryolitia: We don't magically know which part of packages in `fonts.packages`  in user's own NixOS config could fix CJK problem. The best solution to solve it is to provide all users' fonts packages to steam The vast majority of users should be satisfied to have this issue fixed, and the others who have their own ideas about CJK donts in Steam could override the default value.

Packages in `fonts.packages` should be always on the disk, so it won't waste any disk space.

K900: I mean the gamescope bit. I understand the fonts.

nixos/modules/programs/steam.nix
K900: Does this not get used for multilib as well?

nixos/modules/programs/steam.nix
Cryolitia: > I mean the gamescope bit. I understand the fonts.

This part was come up with in @BillHuang2001 https://github.com/NixOS/nixpkgs/issues/162562#issuecomment-1229444338 , and do solve our problems. I don't know why and whether it is the minium, orz.

Cryolitia: > That comment is nearly two years old. Are you sure this is still necessary? I really don't want to have code cargo-culted around without fully understanding why we're doing things.

Yes, I'm sure. I met the same problem for gamescope this year, and fixed it as this. The issue #214275 is still open.

K900: That is an entirely different issue, and it should have been fixed by the buildFHSEnv changes a long time ago. Can you find or file a new issue for whatever gamescope behavior you're seeing?

nixos/modules/programs/steam.nix
Cryolitia: > Does this not get used for multilib as well?

Sorry, I don't understand your question. Could you please describe it in detail.

K900: Those functions should be called twice, once for 64-bit packages (with the 64-bit package set), and once for 32-bit packages. Or am I wrong here?

nixos/modules/programs/steam.nix
urioTV: > That comment is nearly two years old. Are you sure this is still necessary? I really don't want to have code cargo-culted around without fully understanding why we're doing things.

That comment may be old but the solution is still accurate. Gamescope is not working without those tweaks.

K900: Even if it is, gating this on gamescopeSession is incorrect, because that starts Steam inside Gamescope, not Gamescope inside Steam. I'd also much prefer to see a solution that doesn't involve adding random things into the fhsenv.

K900: At the very least, the condition to activate it is wrong.

nixos/modules/programs/steam.nix
K900: Oh, I see, I confused it with `extraLibraries`, sorry. This is fine then.

nixos/modules/programs/steam.nix
Cryolitia: > At the very least, the condition to activate it is wrong.

Ok I understand. It's my fault.

So which condition would you suggest. Would you think `programs.gamescope.enable` acceptable. Or we can make a new option `programs.steam.gamescopeEnhancement` or sth.

Cryolitia: > I'm pretty sure I have a better fix for both of your issues.

So what's it.

K900: I'm testing it now and will submit a PR if it works.

eclairevoyant: `libkrb5` and `keyutils` are the only packages needed here, btw.

K900:
Please just add those two packages unconditionally into the standard Steam fhsenv, and please check if `fonts.fontconfig.cache32Bit` without any of the changes here fixes your Steam fonts.

Cryolitia:
image
On my machine, `fonts.fontconfig.cache32Bit` not works for me.

fontconfig:
```
    fontconfig = {
      defaultFonts = {
        emoji = [
          "Source Han Serif SC"
          "JetBrainsMono Nerd Font"
          "Noto Color Emoji"
        ];
        monospace = [
          "Source Han Serif SC"
          "Sarasa Mono SC"
          "JetBrainsMono Nerd Font Mono"
        ];
        sansSerif = [ "Source Han Serif SC" ];
        serif = [ "Source Han Serif SC" ];
      };
      cache32Bit = true;
    };
```

K900:
Is there an easy way to test this locally? I don't see anything wrong with that screenshot.

Cryolitia:
> Is there an easy way to test this locally? I don't see anything wrong with that screenshot.

I set `Source Han Serif SC`. All fonts should be serif whatever the application actually asked. For example, all CJK in kitty is serif. But steam is still sans. To proof that steam is not following font-config.

K900:
And adding the font to the fhsenv actually makes it work? Can you run `fc-match` in the fhsenv?

eclairevoyant:
> For fonts, it's very likely we just need to enable `fonts.fontconfig.cache32Bit` and Steam will work automagically with system fonts.

This doesn't work for me.

image

Feel free to check https://store.steampowered.com/app/1647250/KUUKIYOMI_3_Consider_It_More_and_More__Father_to_Son/ as an example.

Cryolitia:
> In case of providing font packages to steam and setting as fontconfig:

And in case of not providing, steam is still Tofu block.

Cryolitia:
> And adding the font to the fhsenv actually makes it work? Can you run `fc-match` in the fhsenv?

Adding the fonts to the fhsenv just make steam to pick up a lucky font , still not following font-config. It was discussed in https://github.com/NixOS/nixpkgs/pull/302862#issuecomment-2046308403 . @CoelacanthusHex reports that steam follows font-config on her Arch, that's wired on NixOS......

Cryolitia:
[log.txt](https://github.com/NixOS/nixpkgs/files/15404455/log.txt)
The log of the case of not providing font packages to the fhsenv, and font-config like:
<details>

```nix
    fontconfig = {
      defaultFonts = {
        emoji = [
          "Source Han Serif SC"
          "JetBrainsMono Nerd Font"
          "Noto Color Emoji"
        ];
        monospace = [
          "Source Han Serif SC"
          "Sarasa Mono SC"
          "JetBrainsMono Nerd Font Mono"
        ];
        sansSerif = [ "Source Han Serif SC" ];
        serif = [ "Source Han Serif SC" ];
      };
      cache32Bit = true;
    };
```

</details>

K900:
That is still wrong. Please add the packages directly to the fhsenv.

K900:
I've submitted the correct version of the gamescope change here: https://github.com/NixOS/nixpkgs/pull/313697

Please test that and we can reduce the scope of this PR to fixing the font setup.

Cryolitia:
> Please test that and we can reduce the scope of this PR to fixing the font setup.

Though, the option `programs.steam.extraPackages` is needed insetad of override package, in case of user would like to do other things to steam.

Cryolitia:
Latest progress:

Case: not providing font packages to steam FHS

NixOS config:
<details>

```nix
  fonts = {
    packages = with pkgs; [
      sarasa-gothic
      (nerdfonts.override { fonts = [ "JetBrainsMono" ]; })
      source-han-serif
      source-han-sans
      noto-fonts-emoji
      shanggu-fonts
    ];

    fontconfig = {
      defaultFonts = {
        emoji = [
          "JetBrainsMono Nerd Font"
          "Noto Color Emoji"
        ];
        monospace = [
          "Sarasa Mono SC"
          "JetBrainsMono Nerd Font Mono"
        ];
        sansSerif = [ "Sarasa Gothic SC" ];
        serif = [ "Source Han Serif SC" ];
      };
      cache32Bit = true;
    };
  };
```

</details>

Extra fontconfig:
```
  fonts.fontconfig.localConf = ''
<?xml version="1.0"?>
<!DOCTYPE fontconfig SYSTEM "urn:fontconfig:fonts.dtd">
<fontconfig>
  <description>Load local customization file</description>
	<!-- Load local system customization file -->
	<match>
	    <test name="prgname" compare="contains" ignore-blanks="true">
		    <string>steam</string>
    	</test>
		<test name="family" compare="eq">
			<string>Arial</string>
		</test>
		<edit binding="same" mode="prepend" name="family">
			<string>sans-serif</string>
		</edit>
 	</match>
</fontconfig>
```

`strace -f steam` :

```log
[pid 11930] access("/etc/fonts/local.conf", R_OK) = -1 ENOENT (No such file or directory)
```

fully strace has a so big size 686.0 MB, `grep font` upload here:

[grep.txt](https://github.com/NixOS/nixpkgs/files/15406194/grep.txt)

---

Why couldn't steam open the existing file `/etc/fonts/local.conf` ?

K900:
Does that file actually exist outside the sandbox? Can you reach it from steam-run?

Cryolitia:
Add font packages to steam FHSenv is just boken all font-config, force it to use only god knows which lucky font randomly. It's not the right dirction.

Cryolitia:
> Does that file actually exist outside the sandbox?

Yes.

I also managed to launch a kitty inside steam, file is existing, font is correct:

image

> Can you reach it from steam-run?

Yes

Cryolitia:
Steam do access the true font file

`cat strace.txt |grep sarasa`

<details>

```log
[pid 11553] statx(AT_FDCWD, "/nix/store/xw74k7lqbbn993qpsx7xxd7hwhigcg53-sarasa-gothic-1.0.12", AT_STATX_SYNC_AS_STAT|AT_NO_AUTOMOUNT, STATX_BASIC_STATS, {stx_mask=STATX_ALL|STATX_MNT_ID, stx_attributes=0, stx_mode=S_IFDIR|0555, stx_size=10, ...}) = 0
[pid 11553] openat(AT_FDCWD, "/nix/store/xw74k7lqbbn993qpsx7xxd7hwhigcg53-sarasa-gothic-1.0.12", O_RDONLY|O_LARGEFILE|O_CLOEXEC) = 27
[pid 11553] openat(AT_FDCWD, "/nix/store/xw74k7lqbbn993qpsx7xxd7hwhigcg53-sarasa-gothic-1.0.12/.uuid", O_RDONLY|O_LARGEFILE|O_CLOEXEC) = -1 ENOENT (No such file or directory)
[pid 11553] openat(AT_FDCWD, "/nix/store/xw74k7lqbbn993qpsx7xxd7hwhigcg53-sarasa-gothic-1.0.12/.uuid", O_RDONLY|O_LARGEFILE|O_CLOEXEC) = -1 ENOENT (No such file or directory)
[pid 11553] statx(AT_FDCWD, "/nix/store/xw74k7lqbbn993qpsx7xxd7hwhigcg53-sarasa-gothic-1.0.12/share", AT_STATX_SYNC_AS_STAT|AT_NO_AUTOMOUNT, STATX_BASIC_STATS,  <unfinished ...>
[pid 11553] openat(AT_FDCWD, "/nix/store/xw74k7lqbbn993qpsx7xxd7hwhigcg53-sarasa-gothic-1.0.12/share", O_RDONLY|O_LARGEFILE|O_CLOEXEC <unfinished ...>
[pid 11553] openat(AT_FDCWD, "/nix/store/xw74k7lqbbn993qpsx7xxd7hwhigcg53-sarasa-gothic-1.0.12/share/.uuid", O_RDONLY|O_LARGEFILE|O_CLOEXEC <unfinished ...>
[pid 11553] openat(AT_FDCWD, "/nix/store/xw74k7lqbbn993qpsx7xxd7hwhigcg53-sarasa-gothic-1.0.12/share/.uuid", O_RDONLY|O_LARGEFILE|O_CLOEXEC <unfinished ...>
[pid 11553] statx(AT_FDCWD, "/nix/store/xw74k7lqbbn993qpsx7xxd7hwhigcg53-sarasa-gothic-1.0.12/share/fonts", AT_STATX_SYNC_AS_STAT|AT_NO_AUTOMOUNT, STATX_BASIC_STATS, {stx_mask=STATX_ALL|STATX_MNT_ID, stx_attributes=0, stx_mode=S_IFDIR|0555, stx_size=16, ...}) = 0
[pid 11553] openat(AT_FDCWD, "/nix/store/xw74k7lqbbn993qpsx7xxd7hwhigcg53-sarasa-gothic-1.0.12/share/fonts", O_RDONLY|O_LARGEFILE|O_CLOEXEC <unfinished ...>
[pid 11553] openat(AT_FDCWD, "/nix/store/xw74k7lqbbn993qpsx7xxd7hwhigcg53-sarasa-gothic-1.0.12/share/fonts/.uuid", O_RDONLY|O_LARGEFILE|O_CLOEXEC <unfinished ...>
[pid 11553] openat(AT_FDCWD, "/nix/store/xw74k7lqbbn993qpsx7xxd7hwhigcg53-sarasa-gothic-1.0.12/share/fonts/.uuid", O_RDONLY|O_LARGEFILE|O_CLOEXEC <unfinished ...>
[pid 11553] statx(AT_FDCWD, "/nix/store/xw74k7lqbbn993qpsx7xxd7hwhigcg53-sarasa-gothic-1.0.12/share/fonts/truetype", AT_STATX_SYNC_AS_STAT|AT_NO_AUTOMOUNT, STATX_BASIC_STATS, strace: Process 11665 attached
[pid 11553] openat(AT_FDCWD, "/nix/store/xw74k7lqbbn993qpsx7xxd7hwhigcg53-sarasa-gothic-1.0.12/share/fonts/truetype", O_RDONLY|O_LARGEFILE|O_CLOEXEC <unfinished ...>
[pid 11553] openat(AT_FDCWD, "/nix/store/xw74k7lqbbn993qpsx7xxd7hwhigcg53-sarasa-gothic-1.0.12/share/fonts/truetype/.uuid", O_RDONLY|O_LARGEFILE|O_CLOEXEC) = -1 ENOENT (No such file or directory)
[pid 11553] openat(AT_FDCWD, "/nix/store/xw74k7lqbbn993qpsx7xxd7hwhigcg53-sarasa-gothic-1.0.12/share/fonts/truetype/.uuid", O_RDONLY|O_LARGEFILE|O_CLOEXEC <unfinished ...>
[pid 11553] access("/nix/store/xw74k7lqbbn993qpsx7xxd7hwhigcg53-sarasa-gothic-1.0.12/share/fonts/truetype/Sarasa-Regular.ttc", F_OK <unfinished ...>
[pid 11553] openat(AT_FDCWD, "/nix/store/xw74k7lqbbn993qpsx7xxd7hwhigcg53-sarasa-gothic-1.0.12/share/fonts/truetype/Sarasa-Regular.ttc", O_RDONLY|O_LARGEFILE <unfinished ...>
[pid 11553] access("/nix/store/xw74k7lqbbn993qpsx7xxd7hwhigcg53-sarasa-gothic-1.0.12/share/fonts/truetype/Sarasa-Bold.ttc", F_OK <unfinished ...>
[pid 11553] openat(AT_FDCWD, "/nix/store/xw74k7lqbbn993qpsx7xxd7hwhigcg53-sarasa-gothic-1.0.12/share/fonts/truetype/Sarasa-Bold.ttc", O_RDONLY|O_LARGEFILE <unfinished ...>
[pid 11553] access("/nix/store/xw74k7lqbbn993qpsx7xxd7hwhigcg53-sarasa-gothic-1.0.12/share/fonts/truetype/Sarasa-SemiBold.ttc", F_OK <unfinished ...>
[pid 11553] openat(AT_FDCWD, "/nix/store/xw74k7lqbbn993qpsx7xxd7hwhigcg53-sarasa-gothic-1.0.12/share/fonts/truetype/Sarasa-SemiBold.ttc", O_RDONLY|O_LARGEFILE <unfinished ...>
```

</details>

Cryolitia:
As a temporary mitigation measure, we can require users to pass in only one font package, forcing Steam to use and only use this font package. Passing in more than two packets will inevitably cause greater confusion.

K900:
The fonts issue is unrelated to this change. Does this resolve your gamescope issue?

Cryolitia:
> The fonts issue is unrelated

The fonts issue is related with this PR.

K900:
Sorry, wrong PR. Can you maybe get this down to a minimal version?

Cryolitia:
All clear:

## Reason

Accroding to https://github.com/ValveSoftware/steam-for-linux/issues/10422

> The steamwebhelper (Steam UI) in the beta client is using a vaguely Flatpak-like container (the same Steam Linux Runtime containers used by Proton, Dota 2, CS2) to get a consistent library stack and runtime environment across all Linux distros

> It isn't currently possible to override files in /etc in the Steam Linux Runtime container

That's why `fonts.fontconfig` not work —— we all use nixosModules.

> The container that is now used to run the steamwebhelper is meant to have behaviour similar to Flatpak: it will find fonts in /usr/share/fonts and ~/.local/share/fonts automatically, but those fonts will only be used if an application asks for them by name.

That's why providing font packages to FHSenv works, it links all font files to the fake path `/usr/share/fonts`

## Solution

1. Providing only one font package to FHSenv, that would work, but may not satisfied all users, especially East Asian users who would like to use multi fonts for cmn, jp, even emoji.
2. Linking all `fonts.packages` into FHSenv, by default to at least barely works out of the box. Document it somewhere to teach user how to write the fontconfig in their own home directory, may could linkage with home-manager. But it's hard to warn CJK users precisely, without bothering the latin world.
3. Pass font-config by environment variable, but causes steam boot loop, issue open upstream https://github.com/ValveSoftware/steam-for-linux/issues/10915

CC @eclairevoyant What's your point.

K900:
Can you drop the reformat change? There's likely going to be a treewide format at some point, and it's going to cause merge conflicts on other changes.

nixos/modules/programs/steam.nix
```diff
@@ -110,6 +110,18 @@ in
...
 
+    fontPackages = lib.mkOption {
+      type = lib.types.listOf lib.types.package;
+      default = config.fonts.packages;
+      defaultText = lib.literalExpression "fonts.packages";
+      example = lib.literalExpression "with pkgs; [ source-han-sans ]";
+      description = ''
+        Font packages to use in Steam.
+
+        Defaults to system fonts, but could be overridden to use other fonts — useful for users who would like to customize CJK fonts used in Steam. According to the [upstream issue](https://github.com/ValveSoftware/steam-for-linux/issues/10422#issuecomment-1944396010), steam only follows the fontconfig in user's home.
```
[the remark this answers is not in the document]
Cryolitia: Sorry, my English is poor. But I think we do need to notice users some where.

The Steam's behavior is irregular. Users may not notice that.  They may not use `home-manager` or these things to config user-side fontconfig, and just believe that steam don't follows fontconfig at all. Then they may use varity kinds of hack to fix it in a worse way.

To make matters worse, introducing the whole `fonts.packages` to Steam without user-side fontconfig seems to broken the Japanese if they install a 「草书」font, or something even weirder happens in other languages.

related discussion: https://github.com/NixOS/nixpkgs/pull/302862#issuecomment-2046308403

CJK fonts and display are indeed a very difficult problem.

K900:
All right, let's merge it and we can iterate later. Thanks!

K900:
Actually the current version doesn't pass EditorConfig. Please fix that.

Cryolitia:
Re-tested on my computer, all looks good

Cryolitia:
> Actually the current version doesn't pass EditorConfig. Please fix that.

Apologize, fixed

gador:
This caused a build failure for me:
```
A definition for option `programs.steam.fontPackages."[definition 1-entry 24]"' is not of type `package'. Definition values:
       - In `/nix/store/mw4dx9syd793x0s5z4rq88mrh2i65vxm-source/nixos/modules/programs/steam.nix': "/nix/store/r4qnpfh6s2s6d0kw938syll5pp3gy76c-ghostscript-with-X-10.02.1/share/ghostscript/fonts
 ```

probably caused by my config:
```
fonts = {
    enableGhostscriptFonts = true;
}
```

Which adds 
```
config = mkIf config.fonts.enableGhostscriptFonts {
    fonts.packages = [ "${pkgs.ghostscript}/share/ghostscript/fonts" ];
  };
``` 
[source](https://github.com/NixOS/nixpkgs/blob/nixos-unstable/nixos/modules/config/fonts/ghostscript.nix)

Cryolitia:
@gador Could you please open a new issue and mention me by @Cryolitia , this thread has too many participants.

PR #312389, opened 2024-05-17
wildcard: init at 0.3.3

ASHGOLDOFFICIAL:
## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/by-name/wi/wildcard/package.nix
+++ b/pkgs/by-name/wi/wildcard/package.nix
@@ -57,7 +57,7 @@ stdenv.mkDerivation (finalAttrs: {
     homepage = "https://gitlab.gnome.org/World/Wildcard";
     downloadPage = "https://gitlab.gnome.org/World/Wildcard/-/releases/v${finalAttrs.version}";
     license = lib.licenses.gpl3Plus;
-    maintainers = with lib.maintainers; [];
+    maintainers = with lib.maintainers; [ aleksana ];
     mainProgram = "wildcard";
     platforms = lib.platforms.linux;
   };
--- a/pkgs/by-name/wi/wildcard/package.nix
+++ b/pkgs/by-name/wi/wildcard/package.nix
@@ -29,7 +29,7 @@ stdenv.mkDerivation (finalAttrs: {
 
   cargoDeps = rustPlatform.fetchCargoTarball {
     inherit (finalAttrs) src;
-    hash = "sha256-RWEUS7WA/mKCjHLUvkdzsHGg9xQrEVGnvqncBTy1U8c=";
+    hash = "sha256-8jNNCcZRoLyOHdaWmYTOGD7Nf7NkmJ1MIxBXLJGrm5Y=";
     name = "wildcard-${finalAttrs.version}";
   };
 
--- a/pkgs/by-name/wi/wildcard/package.nix
+++ b/pkgs/by-name/wi/wildcard/package.nix
@@ -25,22 +25,28 @@ stdenv.mkDerivation (finalAttrs: {
     hash = "sha256-jOv0l1vnfDePWF7SAbsBFipPAONliPdc47xj79BJ+rc=";
   };
 
+  strictDeps = true;
+
   cargoDeps = rustPlatform.fetchCargoTarball {
     inherit (finalAttrs) src;
     hash = "sha256-RWEUS7WA/mKCjHLUvkdzsHGg9xQrEVGnvqncBTy1U8c=";
+    name = "wildcard-${finalAttrs.version}";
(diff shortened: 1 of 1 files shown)


pkgs/by-name/wi/wildcard/package.nix
```diff
@@ -0,0 +1,58 @@
...
+
+  meta = {
+    description = "Test your regular expressions";
+    longDescription = ''
+      Wildcard gives you a nice and simple to use interface to test/practice regular expressions.
+    '';
+    homepage = "https://gitlab.gnome.org/World/Wildcard";
+    downloadPage = "https://gitlab.gnome.org/World/Wildcard/-/releases/v${finalAttrs.version}";
+    license = lib.licenses.gpl3Plus;
+    maintainers = with lib.maintainers; [];
```
Aleksanaa, line 54: You don't want to add yourself to maintainers? (See `maintainers/maintainer-list.nix`) first

pkgs/by-name/wi/wildcard/package.nix
```diff
@@ -0,0 +1,58 @@
...
+    rev = "v${finalAttrs.version}";
+    hash = "sha256-jOv0l1vnfDePWF7SAbsBFipPAONliPdc47xj79BJ+rc=";
+  };
+
+  cargoDeps = rustPlatform.fetchCargoTarball {
+    inherit (finalAttrs) src;
+    hash = "sha256-RWEUS7WA/mKCjHLUvkdzsHGg9xQrEVGnvqncBTy1U8c=";
+  };
+
+  nativeBuildInputs = [
```
Aleksanaa, line 33: `nativeBuildInputs` means the architecture of the package is the same of the architecture of the build environment, which should be packages that are only available in build time. `buildInputs` are packages (that should be) available at runtime, but they are not simply listed as dependencies, and whether it can become the final runtime dependencies is determined by whether the result has a relationship (linking, including path, etc) with it.

See https://nixos.org/manual/nixpkgs/unstable/#ssec-stdenv-dependencies

Aleksanaa: In this case, libadwaita should go to `buildInputs`, and others are still `nativeBuildInputs`. It's worth noting that enabling `strictDeps = true` can sometimes help you point out these mistakes.

pkgs/by-name/wi/wildcard/package.nix
```diff
@@ -0,0 +1,58 @@
...
+  src = fetchFromGitLab {
+    domain = "gitlab.gnome.org";
+    owner = "World";
+    repo = "Wildcard";
+    rev = "v${finalAttrs.version}";
+    hash = "sha256-jOv0l1vnfDePWF7SAbsBFipPAONliPdc47xj79BJ+rc=";
+  };
+
+  cargoDeps = rustPlatform.fetchCargoTarball {
+    inherit (finalAttrs) src;
```
Aleksanaa, line 29:
```suggestion
    inherit (finalAttrs) src;
    name = "wildcard-${finalAttrs.version}";
```

pkgs/by-name/wi/wildcard/package.nix
ASHGOLDOFFICIAL: I've only began learning nix and packaging, so I think I don't have enough experience to actually maintain a package. So, no, I don't want to add myself to maintainers (not now at least).

pkgs/by-name/wi/wildcard/package.nix
ASHGOLDOFFICIAL: Added it.

pkgs/by-name/wi/wildcard/package.nix
ASHGOLDOFFICIAL: Thank you for explanation.

pkgs/by-name/wi/wildcard/package.nix
Aleksanaa: But you are still submitting it, how do you think you cannot maintain it?

ASHGOLDOFFICIAL: I think maintaining a package is a big responsibility that requires experience and involvement. I don't think I am yet ready for this it: I don't have enough experience with packaging in nix and I don't yet know if I am interested in this kind of stuff. So I want to leave maintaining at more capable hands for now.

PR #296979, opened 2024-03-18
lib: add option parameter for default value priority

Atemu:
It's sometimes the case that your option's default value should be additive such that setting it results in the default value being merged with whatever was set.

Previously, you had to add the default value in the implementation which might be many lines away from the option declaration and is generally an odd pattern.

This is a lot clearer, simpler and doesn't add much complexity.

If this becomes used more widely, it should be rendered in the generated options' manual and nixos-search.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
- For non-Linux: Is sandboxing enabled in `nix.conf`? (See [Nix manual](https://nixos.org/manual/nix/stable/command-ref/conf-file.html))
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:
  - [NixOS test(s)](https://nixos.org/manual/nixos/unstable/index.html#sec-nixos-tests) (look inside [nixos/tests](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests))
  - and/or [package tests](https://nixos.org/manual/nixpkgs/unstable/#sec-package-tests)
  - or, for functions and "core" functionality, tests in [lib/tests](https://github.com/NixOS/nixpkgs/blob/master/lib/tests) or [pkgs/test](https://github.com/NixOS/nixpkgs/blob/master/pkgs/test)
  - made sure NixOS tests are [linked](https://nixos.org/manual/nixpkgs/unstable/#ssec-nixos-tests-linking) to the relevant packages
- [ ] Tested compilation of all packages that depend on this change using `nix-shell -p nixpkgs-review --run "nixpkgs-review rev HEAD"`. Note: all changes have to be committed, also see [nixpkgs-review usage](https://github.com/Mic92/nixpkgs-review#usage)
- [ ] Tested basic functionality of all binary files (usually in `./result/bin/`)
- [24.05 Release Notes](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2405.section.md) (or backporting [23.05](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2305.section.md) and [23.11](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2311.section.md) Release notes)
  - [ ] (Package updates) Added a release notes entry if the change is major or breaking
[...]

Issue #295030, opened 2023-08-08
Performance benchmark for drv-parts/dream2nix compared to bare derivations


infinisil:
Ping @DavHau, let's have this in an issue instead of the Matrix channel

https://github.com/nix-community/dream2nix/tree/main/misc/benchmarks

infinisil:
Here's the message @DavHau shared [on Matrix](https://matrix.to/#/!djTaTBQyWEPRQxrPTb:nixos.org/$muE2XE-sj2n-9LsM88zq6L3Mttk3z0opDb5-zYyShsE?via=nixos.org&via=matrix.org&via=nixos.dev) that I think is valuable enough to be more persistent:

> DISCLAIMER: This is just an experiment
> 
> I made [this fork](https://github.com/DavHau/nixpkgs/tree/derivation-modules) of nixpkgs, which uses the module system for all mkDerivation based packages. All changes are contained in the latest commit.
> 
> Done:
> 
> - in make-derivation.nix replace call to `builtins.derivation` with `lib.evalModules`
> - declare interface using the module system
> - its based on Robert Hensing (roberth) 's [WIP record type](https://github.com/NixOS/nixpkgs/pull/257511) for performance reasons (submodule+freeform was quite slow)
> - the defined options are incomplete -> unknown fields default to `types.raw`
> - added `modules` argument to mkDerivation allowing user to pass extra modules (cool, but still requires calling overrideAttrs which is weird)
> - pass through `.mix` function for each derivation, that allows passing a single module to extend the derivation.
> 
> Example usage:
> 
> ## Overriding
> 
> overriding before:
> 
> ```
> firefox.overrideAttrs (old: {buildInputs = old.buildInputs or [] ++ [hello];})
> ```
> 
> overriding after:
> 
> ```
> firefox.mix {derivation.buildInputs = [hello];}
> ```
> 
> - no need to pass function
> - no need to handle previous value
> - no need to handle default
> - no need to implement merge
> 
> ## Type Checking
> 
> ```
> nix-repl> firefox.mix {derivation.buildInputs = {};}
> error:
> [...]
>        error: A definition for option `derivation.buildInputs' is not of type `list of (null or package or path)'.
>        [...]
> ```
> 
> ## Performance
> 
> (None of these findings are thoroughly tested or statistically sound)
> 
> - Overhead on packages with small closure is little (~10% for pkgs.hello)
> - Overhead on packages with large closure is high: (~100% for pkgs.firefox)
[...]

PR #312432, opened 2024-05-17
package overrides using a mini module system

Atemu:
## Description of changes

Putting a package's configuration options into function arguments in order to expose them via `override` is a well established pattern but has major issues:

- No discoverablilty/documentation of available options
- Namespace collisions (calling an option "xorg" would result in callPackage inserting the xorg package set)
- Odd syntax and scoping limitations when declaring options (`?` and no ability to use let bindings)
- No type checking
- No good error messages
- No sane way to transparently communicate breaking changes

But we have a well-established solution to all those problems: The module system.

This is a little thought that came to my mind while thinking about [RFC 169](https://github.com/NixOS/rfcs/pull/169):  
Why couldn't we just use the module system for configuring packages?

The extent to which this is done here is merely the configuration options that were explicitly exposed by the package before which is similar to how modules are used NixOS; explicit options provided by the package maintainer to customise some specific aspect of the package.

The only limitation I currently see is that it is a bit awkward to compose multiple overrides:

```Nix
# This is fine
(mlterm.override {
  configuration = {
    gui = {
      wayland = false;
      sdl2 = false;
    };
  };
}).override
  # This requires advanced knowledge of the moule system and overrides
  (prev: {
    configuration = {
      imports = [ prev.configuration ];
      gui.sdl2 = true;
    };
  })
```

This could however be remedied using a dedicated `overrideFeatures` function.

As for performance:

Edit: Out of date. See below.

Previous:

```
[atemu@HEPHAISTOS nixpkgs]$ hyperfine --warmup 1 -r 30 'nix-instantiate -A ffmpeg'
Benchmark 1: nix-instantiate -A ffmpeg
  Time (mean ± σ):     716.5 ms ±  13.9 ms    [User: 430.6 ms, System: 143.5 ms]
  Range (min … max):   658.7 ms … 738.2 ms    30 runs
  
[atemu@HEPHAISTOS nixpkgs]$ hyperfine --warmup 10 -r 30 'nix-instantiate -A mlterm'
Benchmark 1: nix-instantiate -A mlterm
  Time (mean ± σ):     739.5 ms ±  33.0 ms    [User: 445.2 ms, System: 151.8 ms]
  Range (min … max):   688.9 ms … 792.6 ms    30 runs
```
  
With modules:

```
[atemu@HEPHAISTOS nixpkgs]$ hyperfine --warmup 1 -r 30 'nix-instantiate -A ffmpeg'
Benchmark 1: nix-instantiate -A ffmpeg
  Time (mean ± σ):     712.8 ms ±  21.4 ms    [User: 428.8 ms, System: 143.0 ms]
  Range (min … max):   657.3 ms … 748.0 ms    30 runs
  
[atemu@HEPHAISTOS nixpkgs]$ hyperfine --warmup 10 -r 30 'nix-instantiate -A mlterm'
Benchmark 1: nix-instantiate -A mlterm
  Time (mean ± σ):     740.9 ms ±  35.9 ms    [User: 446.8 ms, System: 152.2 ms]
  Range (min … max):   687.8 ms … 795.2 ms    30 runs
```

There might be a slight difference here but none that I could measure using runtime.

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---


roberth:
My thoughts on this draft:

Unfortunately we can't measure the performance impact by adding only a handful of examples.
The performance impact of the module system is bounded by how much you ask it to do.
That means that we don't have to expect the impact of, say, loading NixOS's massive module list, but something very small for small configs like these.
However, that means that for a realistic picture, we need to apply this solution to all packages in ffmpeg's closure, or a known portion for which we get a statistically significant difference.

With the Nixpkgs architecture team, we've looked into applying the module system more broadly. See e.g.
- https://github.com/NixOS/nixpkgs/issues/295030

Reducing the scope of module system usage may alleviate the performance issue somewhat, but then the question is whether this is the most effective allocation of our evaluation performance "budget". (I don't have an answer)
By applying the solution to only part of the problem, I feel like we're _almost_ making the situation worse because of the added complexity of having different solutions for what are actually instances of the same problem - package configuration is just parameters, just like dependencies are, or "environment" things like the platforms. Shouldn't they all be discoverable, mergeable, etc?

doronbehar (commented):
I like this! Not because I'm hopeful about the performance improvement (in fact I'm skeptic about this detail), but mostly  because finally options have a description written with Nix and not in a comment.

AndersonTorres:
Obviouslt my question is: how hard will be to teach this to newcomers?

Atemu:
Teach what to newcomers? The module system they likely already use in NixOS?

AndersonTorres:
Yes, given that Nixpkgs is not tied to NixOS.

Atemu:
What part do you expect to be hard to teach to newcomers? From the perspective of a user, they provide an attrset of configuration either way. I don't see how *usage* of this API would be any harder to understand than the status quo; using a module system requires basically no knowledge of its internal mechanisms.

I'd argue it'd be slightly simpler to use actually because error messages would be a lot easier to understand: "anonymous function at ... called with unexpected argument withFoo" vs. "the option foo does not exist". And that's ignoring the errors you could catch via typecheck instead of incredibly confusing "cannot coerce ... into a string".

AndersonTorres:
> What part do you expect to be hard to teach to newcomers? 

Any part not well documented, to begin with.

https://nixos.org/manual/nixpkgs/stable/#module-system

> Note
> 
> This chapter **is new and not complete yet**. For a gentle introduction to the module system, in the context of NixOS, see [Writing NixOS Modules](https://nixos.org/manual/nixos/unstable/#sec-writing-modules) in the NixOS manual.

(bold mine)

Atemu:
I do not see how the developer documentation for setting up a module system would be relevant to users merely using a module system that someone else already set up.

As per my previous comment, no newcomer needs to know about the module system's internal mechanisms in order to configure something with it.

Atemu:
I just noticed that there has been prominent precedent for this in Nixpkgs for many years now: The structured Linux kernel config drv.

It uses much the same pattern as what I "came up" with. (I didn't actively think of the kernel config but it may have subconsciously influenced me.)

Introduced all the way back in 2018 by @teto: https://github.com/NixOS/nixpkgs/pull/42838

Atemu:
Over the past few weeks, I've been chugging away at an expansion on this idea: making full use of the module system and its type system to define all the core aspects of the derivation (configureFlags and buildInputs).

This new design is quite elegant IMHO; utilising the basic building blocks of the module system to massively cut down the complex and repetitive derivation implementation code into simple data that is even conveniently malleable at "runtime" by end-users.

The line count is roughly double but that is mostly data declaration syntax overhead (`{` and `};`s) and the rest is, as mentioned, mostly data rather than code, so I don't think that's bad at all.  
I much prefer this over the mess that is the function arguments. 

What I especially like is that the packages required for a given feature are now declared in the same place as the feature flag itself. This just feels right to me and also allows for further introspection as you can now programatically figure out which feature pulls in which dependency.

I have not done proper performance testing yet. A quick and dirty test on my laptop showed a significant regression but that was well within the margin of error for a power-limited laptop CPU, so that's hardly conclusive.

This is based on a much older state of ffmpeg, so it's missing a few options that have been added or changed since. I will rebase this in due time but I only care about a working POC for now to fine-tune the design.

pkgs/development/libraries/ffmpeg/options.nix
```diff
@@ -0,0 +1,753 @@
...
+        '';
+      };
+
+      gate = lib.mkOption {
+        default = true;
+        description = ''
+          Gates the default value for {option}`enable` behind a boolean. Use
+          this to disable certain features on certain platforms by default.
+        '';
+      };
```
Atemu, line 77: Perhaps this should be readOnly too.

pkgs/development/libraries/ffmpeg/options.nix
```diff
@@ -0,0 +1,752 @@
...
+
+let
+  # Checks whether the given variant of the flag is contained in the currently
+  # active variant. full ⊃ small ⊃ headless
+  isInVariant = flagVariant:
+    let
+      checks = {
+        full = variant == variants.full;
+        small = variant == variants.small || checks.full;
+        headless = variant == variants.headless || checks.small;
```
bew, line 21: Shouldn't the checks be:
- headless: check only headless
- small: check both headless & small
- full: check all
?

Atemu: The function works the other way around: checking whether the feature should be enabled in a certain variant. Any feature that is enabled in headless should also be enabled in full but not vice versa.

This is analogous to how `with*Deps` worked before.

Atemu:
Before:
```
$ hyperfine --warmup 10 -r 30 'nix-instantiate -A ffmpeg-headless'
Benchmark 1: nix-instantiate -A ffmpeg-headless
  Time (mean ± σ):     464.4 ms ±  15.9 ms    [User: 289.1 ms, System: 100.4 ms]
  Range (min … max):   429.5 ms … 486.9 ms    30 runs

$ hyperfine --warmup 10 -r 30 'nix-instantiate -A ffmpeg'
Benchmark 1: nix-instantiate -A ffmpeg
  Time (mean ± σ):     709.7 ms ±  33.3 ms    [User: 450.4 ms, System: 134.3 ms]
  Range (min … max):   649.3 ms … 750.7 ms    30 runs

$ hyperfine --warmup 10 -r 30 'nix-instantiate -A ffmpeg-full'
Benchmark 1: nix-instantiate -A ffmpeg-full
  Time (mean ± σ):     778.4 ms ±  36.8 ms    [User: 472.2 ms, System: 166.5 ms]
  Range (min … max):   710.5 ms … 828.2 ms    30 runs
```
After:
```
$ hyperfine --warmup 10 -r 30 'nix-instantiate -A ffmpeg-headless'
Benchmark 1: nix-instantiate -A ffmpeg-headless
  Time (mean ± σ):     484.4 ms ±  19.0 ms    [User: 316.3 ms, System: 92.1 ms]
  Range (min … max):   441.3 ms … 510.0 ms    30 runs

$ hyperfine --warmup 10 -r 30 'nix-instantiate -A ffmpeg'
Benchmark 1: nix-instantiate -A ffmpeg
  Time (mean ± σ):     727.0 ms ±  30.0 ms    [User: 459.6 ms, System: 148.2 ms]
  Range (min … max):   666.8 ms … 774.0 ms    30 runs

$ hyperfine --warmup 10 -r 30 'nix-instantiate -A ffmpeg-full'
Benchmark 1: nix-instantiate -A ffmpeg-full
  Time (mean ± σ):     815.5 ms ±  38.0 ms    [User: 523.3 ms, System: 157.6 ms]
  Range (min … max):   762.0 ms … 884.0 ms    30 runs
```

There appears to be an effect of about 5% but that only corresponds to ~1σ which isn't really significant, so it's hard to say whether there's an actual negative effect through runtime testing.

Here's the diff of `NIX_SHOW_STATS=1` of `ffmpeg-full` which I expect to show the greatest difference since it contains 3 separate evals of ffmpeg:

```
│  1 │warning: you did not specify '--add-ro↵ │  1 │warning: you did not specify '--add-ro↵
│    │ot'; the result might be removed by th↵ │    │ot'; the result might be removed by th↵
│    │e garbage collector                     │    │e garbage collector
│  2 │/nix/store/3ykq8wg61ibq7xaxcz8h7ygf08i↵ │  2 │/nix/store/4pwzjp1bb4axh7r8ap3jdp777x0
│    │2l8zi-ffmpeg-full-6.1.1.drv!bin         │    │8a4k1-ffmpeg-full-6.1.1.drv!bin
│  3 │{                                       │  3 │{
│  4 │  "cpuTime": 0.46145400404930115,       │  4 │  "cpuTime": 0.5295500159263611,
│  5 │  "envs": {                             │  5 │  "envs": {
│  6 │    "bytes": 15570136,                  │  6 │    "bytes": 27651016,
│  7 │    "elements": 1173009,                │  7 │    "elements": 2054566,
│  8 │    "number": 773258                    │  8 │    "number": 1401811
│  9 │  },                                    │  9 │  },
│ 10 │  "gc": {                               │ 10 │  "gc": {
│ 11 │    "heapSize": 402915328,              │ 11 │    "heapSize": 402915328,
│ 12 │    "totalBytes": 195629904             │ 12 │    "totalBytes": 253786144
│ 13 │  },                                    │ 13 │  },
│ 14 │  "list": {                             │ 14 │  "list": {
│ 15 │    "bytes": 1935720,                   │ 15 │    "bytes": 3137688,
│ 16 │    "concats": 53881,                   │ 16 │    "concats": 78153,
│ 17 │    "elements": 241965                  │ 17 │    "elements": 392211
│ 18 │  },                                    │ 18 │  },
│ 19 │  "nrAvoided": 931140,                  │ 19 │  "nrAvoided": 1577220,
│ 20 │  "nrFunctionCalls": 693286,            │ 20 │  "nrFunctionCalls": 1228708,
│ 21 │  "nrLookups": 349445,                  │ 21 │  "nrLookups": 755979,
│ 22 │  "nrOpUpdateValuesCopied": 4800101,    │ 22 │  "nrOpUpdateValuesCopied": 5003831,
│ 23 │  "nrOpUpdates": 101294,                │ 23 │  "nrOpUpdates": 137413,
│ 24 │  "nrPrimOpCalls": 392853,              │ 24 │  "nrPrimOpCalls": 615520,
│ 25 │  "nrThunks": 1323307,                  │ 25 │  "nrThunks": 2033500,
│ 26 │  "sets": {                             │ 26 │  "sets": {
│ 27 │    "bytes": 96291744,                  │ 27 │    "bytes": 109203312,
│ 28 │    "elements": 5856951,                │ 28 │    "elements": 6502876,
│ 29 │    "number": 161283                    │ 29 │    "number": 322331
│ 30 │  },                                    │ 30 │  },
│ 31 │  "sizes": {                            │ 31 │  "sizes": {
│ 32 │    "Attr": 16,                         │ 32 │    "Attr": 16,

───────────────────────────────────────────────────────────────────────────────────────┐
• 35: warning: you did not specify '--add-root'; the result might be removed by the ga │
───────────────────────────────────────────────────────────────────────────────────────┘
│ 35 │    "Value": 24                         │ 35 │    "Value": 24
│ 36 │  },                                    │ 36 │  },
│ 37 │  "symbols": {                          │ 37 │  "symbols": {
│ 38 │    "bytes": 432883,                    │ 38 │    "bytes": 432598,
│ 39 │    "number": 41660                     │ 39 │    "number": 41663
│ 40 │  },                                    │ 40 │  },
│ 41 │  "values": {                           │ 41 │  "values": {
│ 42 │    "bytes": 47772936,                  │ 42 │    "bytes": 70875288,
│ 43 │    "number": 1990539                   │ 43 │    "number": 2953137
│ 44 │  }                                     │ 44 │  }
│ 45 │}                                       │ 45 │}
```

The only thing I take from this is that there are twice as many function calls now.

doronbehar:
I don't think we have that many packages that would want to use this module system, hence I don't consider this price as too high for gradual insertion of this feature into Nixpkgs. I'd like to see this getting moved forward by rebasing etc, and see ofborg's checks will be green. I'd also like to practice afterwards in converting Gnuradio's and `mpd`'s different features to use this module system.

I'd also like to discuss how the docs would look like.

Atemu:
I'm confident ofBorgs checks will pass because this evals to the exact same result (modulo order).

Atemu:
I've done a few more cleanups and refactors and I think I'm at a point where I'm quite happy with how the ffmpeg derivation turned out and can't immediately think of anything I'd change about the design anymore.

pkgs/development/libraries/ffmpeg/generic.nix
```diff
@@ -456,214 +1047,28 @@ stdenv.mkDerivation (finalAttrs: {
     "--target_os=${if stdenv.hostPlatform.isMinGW then "mingw64" else stdenv.hostPlatform.parsed.kernel.name}"
     "--arch=${stdenv.hostPlatform.parsed.cpu.name}"
     "--pkg-config=${buildPackages.pkg-config.targetPrefix}pkg-config"
```
Atemu, line 1049: I've decided to keep these flags be managed manually; without implying them from the modules eval. Rationale being that it's just a few and they're not really something you'd usually (if ever) want to touch.

Atemu:
I'm pretty much on the same page w.r.t. eval time increase. For now, this only really makes sense to do in complex packages such as ffmpeg or Linux. A suspected eval time increase by ~1 standard deviation (10^-2 s) isn't a great cost to pay considering a typical NixOS eval takes 2-3 orders of magnitude longer.

While I'd like better documentation, the status quo is that nothing about package customisation is documented outside of code and its comments which is retained. I'd rather wait until we have a proper place and system coordinated for that at which point it should be trivial to plug ffmpeg's options into that.

What I'd like to discuss more is how to handle the override situation as it's all but intuitive or easy and I can't think of an easy solution. The best I can currently come up with is a dedicated override function (i.e. overrideFeatures).

pkgs/development/libraries/ffmpeg/generic.nix
```diff
@@ -337,49 +142,835 @@
...
+          };
+          x264 = {
+            gate = enable.gpl;
+            packages = { inherit x264; };
+            flagPrefix = "lib";
+          };
+          x265 = {
+            gate = enable.gpl;
+            packages = { inherit x265; };
+            flagPrefix = "lib";
```
tobim, line 843: Somehow `flagPrefix` doesn't seem to work. With this shell.nix
```
{ pkgs ? import ../nixpkgs {} }:
pkgs.mkShell {
  packages = [
    (
      pkgs.ffmpeg.override {
        configuration = {
          features.gpl.enable = true;
          features.x265.enable = true;
        };
      }
    )
  ];
}
```
I get
```
       > configure flags: --disable-static --prefix=/nix/store/la3im0f739w6faba4k2h35mlambdciii-ffmpeg-6.1.1 --target_os=linux --arch=x86_64 --pkg-config=pkg-config --datadir=/nix/store/r6z9z0qrjs7blsphfc2z1licwmip7sac-ffmpeg-6.1.1-data/share/ffmpeg --bindir=/nix/store/kwxif6j3p3138g12dvvfyjl8mnvxpxms-ffmpeg-6.1.1-bin/bin --libdir=/nix/store/dx9j1hi5pxqag8l7ns4lwpywfb7h9wxa-ffmpeg-6.1.1-lib/lib --incdir=/nix/store/zmn11vw7kzh205yl3m09wsnjb72dpq95-ffmpeg-6.1.1-dev/include --mandir=/nix/store/v9hyndmvzpirv0ayv77ak9gg8m88ld56-ffmpeg-6.1.1-man/share/man --docdir=/nix/store/mivsjfqgnj1difymzwnyvzzivsz20qs6-ffmpeg-6.1.1-doc/share/doc/ffmpeg --enable-alsa --disable-libaom --disable-appkit --disable-libaribcaption --enable-libass --disable-audiotoolbox --enable-avcodec --enable-avdevice --enable-avfilter --enable-avformat --disable-avfoundation --disable-avisynth --enable-avutil --disable-libbluray --disable-libbs2b --enable-bzlib --disable-libcaca --disable-libcelt --disable-chromaprint --disable-libcodec2 --disable-coreimage --disable-cuda --disable-cuda-llvm --enable-cuvid --enable-libdav1d --disable-libdc1394 --disable-debug --enable-doc --enable-libdrm --disable-extra-warnings --disable-libfdk-aac --enable-ffmpeg --enable-ffplay --enable-ffprobe --disable-libflite --enable-fontconfig --enable-libfontconfig --enable-libfreetype --disable-frei0r --disable-libfribidi --disable-libgme --enable-gnutls --enable-gpl --disable-gray --disable-libgsm --enable-hardcoded-tables --enable-libharfbuzz --enable-htmlpages --enable-iconv --disable-libjack --disable-libjxl --disable-ladspa --enable-lzma --enable-manpages --disable-libmfx --disable-libmodplug --enable-libmp3lame --disable-libmysofa --enable-network --enable-ffnvcodec --enable-nvdec --enable-nvenc --disable-openal --disable-opencl --disable-libopencore-amrnb --disable-libopencore-amrwb --disable-opengl --disable-libopenh264 --disable-libopenjpeg --disable-libopenmpt --enable-optimizations --enable-libopus --disable-os2threads --enable-pic --enable-pixelutils --disable-libplacebo --enable-podpages --enable-postproc --enable-pthreads --enable-libpulse --disable-librav1e --disable-librtmp --enable-runtime-cpudetect --enable-safe-bitstream-reader --disable-libsmbclient --enable-sdl2 --disable-libshaderc --enable-shared --disable-small --enable-libsoxr --enable-libspeex --enable-libsrt --enable-libssh --disable-static --disable-stripping --disable-librsvg --enable-libsvtav1 --enable-swresample --enable-swscale --enable-swscale-alpha --disable-libtensorflow --enable-libtheora --disable-thumb --enable-txtpages --disable-nonfree --enable-libv4l2 --enable-v4l2-m2m --enable-vaapi --enable-vdpau --enable-version3 --disable-videotoolbox --disable-libvidstab --disable-libvmaf --disable-libvo-amrwbenc --enable-libvorbis --disable-libvpl --enable-libvpx --enable-vulkan --disable-w32threads --disable-libwebp --enable-libx264 --enable-x265 --disable-libxavs --disable-libxcb --disable-libxcb-shape --disable-libxcb-shm --disable-libxcb-xfixes --disable-xlib --disable-libxml2 --enable-libxvid --enable-libzimg --enable-zlib --disable-libzmq
       > Unknown option "--enable-x265".
       > See ./configure --help for available options.
```

Atemu: Overriding one value of an ffmpegFlag appears to override the entire ffmpegFlag. Back to the drawing board...

Perhaps we do need https://github.com/NixOS/nixpkgs/pull/296979 for this.

Atemu: The fix wasn't as bad as I imagined.

doronbehar:
> What I'd like to discuss more is how to handle the override situation as it's all but intuitive or easy and I can't think of an easy solution. The best I can currently come up with is a dedicated override function (i.e. overrideFeatures).

Which would just be a thin wrapper around `.override { configuration = ...; }`?

> While I'd like better documentation, the status quo is that nothing about package customisation is documented outside of code and its comments which is retained. I'd rather wait until we have a proper place and system coordinated for that at which point it should be trivial to plug ffmpeg's options into that.

It doesn't feel like a big deal to me - we have a place already to document how to use Nix expressions of packages, it's just not automatically generated. I agree we don't have to settle now on how exactly to build the workflow of adding these docs automatically, but I'd like to see that it is at least possible to generate some documentation files automatically from these expressions.

Atemu:
> Which would just be a thin wrapper around `.override { configuration = ...; }`?

It'd work as described in the OP.

Though implementing it could be tricky because an override functions' scope is not local to the package.

> we have a place already to document how to use Nix expressions of packages

Where? That's news to me.

> I'd like to see that it is at least possible to generate some documentation files automatically from these expressions.

This already works for NixOS, I don't see why it wouldn't work here aswell.

In fact, you can already introspect the documentation using a repl/nix-instantiate:

```
$ nix-instantiate --eval -A ffmpeg.eval.options.features.ass.description
"Control ass support in ffmpeg: (Advanced) SubStation Alpha subtitle rendering."
```

Atemu:
> This already works for NixOS, I don't see why it wouldn't work here aswell.

I tried it and it's trivial: `pkgs.nixosOptionsDoc { inherit (ffmpeg.eval) options; }`

What I noticed is that it puts a literal `‹name›` where the submodule's name should be in sentences like `The dependencies required to enable ‹name› support.`. Not sure what's going on there.

doronbehar:
> > Which would just be a thin wrapper around `.override { configuration = ...; }`?
> 
> It'd work as described in the OP.
> 
> Though implementing it could be tricky because an override functions' scope is not local to the package.

Hmm OK I see. Now I read in the OP that it is only an issue with multiple overrides? Why would someone need to do that?

Atemu:
It's only an issue with composing overrides indeed but you never know what people might need to do with your derivation.

aanderse:
multiple overrides are definitely a thing in nixos

consider a module which provides a `package` option watch requires a specific option - and allows the user to further customize options

i don't recall the concrete example to link at the moment but i if you look you will find it

Atemu:
Steam does this for instance:

https://github.com/NixOS/nixpkgs/blob/8c9630d265fa4665bd2ad86e39eadcb930470544/nixos/modules/programs/steam.nix#L45-L65

As you can see, it also needs to take great care of overriding in a composable manner but that's only necessary when composing values of arguments; composing top-level arguments themselves without "deeper" values is trivial.

PR #312472, opened 2024-05-17
nixos/networkd: get rid of *Config attributes in lists

Ma27:
## Description of changes
This patch is about removing `wireguardPeerConfig`, `dhcpServerStaticLeaseConfig` - a.k.a. the
AbstractSingletonProxyFactoryBean of nixpkgs - and friends.

As a former colleague said

> worst abstraction ever

I second that. I've written enough networkd config for NixOS systems so far to have a strong dislike. In fact, these don't even make sense: `netdevs.wireguardPeers._.wireguardPeerConfig` will be rendered into the key `[WireGuardPeer]` and every key from `wireguardPeerConfig` is in there. Since it's INI, there's no place where sections on the same level as wireguardPeerConfig fit into. Hence, get rid of it all.

For the transition, using the old way is still allowed, but gives a warning. I think we could drop this after one release.

The tests of rosenpass and systemd-networkd-dhcpserver-static-leases were broken on the rev before, hence they were updated, but are still not building.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [x] Tested, as applicable:

---

### DIFF
--- a/nixos/modules/services/networking/rosenpass.nix
+++ b/nixos/modules/services/networking/rosenpass.nix
@@ -130,8 +130,8 @@ in
             relevant = config.systemd.network.enable;
             root = config.systemd.network.netdevs;
             peer = (x: x.wireguardPeers);
-            key = (x: if x.wireguardPeerConfig ? PublicKey then x.wireguardPeerConfig.PublicKey else null);
-            description = "${options.systemd.network.netdevs}.\"<name>\".wireguardPeers.*.wireguardPeerConfig.PublicKey";
+            key = x: x.PublicKey or null;
+            description = "${options.systemd.network.netdevs}.\"<name>\".wireguardPeers.*.PublicKey";
           }
           {
             relevant = config.networking.wireguard.enable;
--- a/nixos/modules/system/boot/networkd.nix
+++ b/nixos/modules/system/boot/networkd.nix
@@ -386,7 +386,7 @@ let
         (assertValueOneOf "UDP6ZeroChecksumRx" boolValues)
       ];
 
-      sectionL2TPSession = checkUnitConfig "L2TPSession" [
+      sectionL2TPSession = checkUnitConfigWithLegacyKey "l2tpSessionConfig" "L2TPSession" [
         (assertOnlyFields [
           "Name"
           "SessionId"
@@ -421,7 +421,7 @@ let
       # NOTE The PresharedKey directive is missing on purpose here, please
       # do not add it to this list. The nix store is world-readable,let's
       # refrain ourselves from providing a footgun.
-      sectionWireGuardPeer = checkUnitConfig "WireGuardPeer" [
+      sectionWireGuardPeer = checkUnitConfigWithLegacyKey "wireguardPeerConfig" "WireGuardPeer" [
         (assertOnlyFields [
           "PublicKey"
           "PresharedKeyFile"
@@ -712,7 +712,7 @@ let
         (assertValueOneOf "KeepConfiguration" (boolValues ++ ["static" "dhcp-on-stop" "dhcp"]))
       ];
 
-      sectionAddress = checkUnitConfig "Address" [
+      sectionAddress = checkUnitConfigWithLegacyKey "addressConfig" "Address" [
         (assertOnlyFields [
           "Address"
           "Peer"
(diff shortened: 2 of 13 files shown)


WilliButz (approved):
Very nice! Also great that you managed to clean up the abstraction in a non-breaking way.

>  I've written enough networkd config for NixOS systems so far to have a strong dislike. In fact, these don't even make sense

+1 on this :)

nixos/modules/services/networking/rosenpass.nix
```diff
@@ -130,8 +130,8 @@ in
             relevant = config.systemd.network.enable;
             root = config.systemd.network.netdevs;
             peer = (x: x.wireguardPeers);
-            key = (x: if x.wireguardPeerConfig ? PublicKey then x.wireguardPeerConfig.PublicKey else null);
-            description = "${options.systemd.network.netdevs}.\"<name>\".wireguardPeers.*.wireguardPeerConfig.PublicKey";
+            key = (x: if x ? PublicKey then x.PublicKey else null);
```
lheckemann, line 133:
```suggestion
            key = (x: x.PublicKey or null);
```

nixos/modules/system/boot/networkd.nix
```diff
@@ -1627,34 +1627,21 @@ let
...
-      wireguardPeerConfig = mkOption {
-        default = {};
-        type = types.addCheck (types.attrsOf unitOption) check.netdev.sectionWireGuardPeer;
-        description = ''
-          Each attribute in this set specifies an option in the
-          `[WireGuardPeer]` section of the unit.  See
-          {manpage}`systemd.netdev(5)` for details.
-        '';
-      };
+  mkSubsectionType = oldKey: checkF:
```
lheckemann, line 1630: (bikeshedding potential, feel free to dismiss) should this be called mkSectionType? systemd's config format has sections but no "subsections", as far as I understand.

nixos/modules/tasks/network-interfaces-systemd.nix
```diff
@@ -95,65 +95,63 @@ let
...
-              Table = route.options.table;
-            } //
-            optionalAttrs (route.options ? advmss) {
-              TCPAdvertisedMaximumSegmentSize = route.options.advmss;
-            } //
-            optionalAttrs (route.options ? ttl-propagate) {
-              TTLPropagate = route.options.ttl-propagate == "enabled";
-            };
-        });
+          optionalAttrs (route.address != null && route.prefixLength != null) {
```
lheckemann, line 101: Could we be using `mkIf` rather than chaining `optionalAttrs` here?

lheckemann (approved):
:heart:

Consider all my review comments as nonblocking nits.

nixos/modules/system/boot/networkd.nix
Ma27: I called it subsection because it's one level deeper than matchConfig/networkConfig/... in the Nix abstraction.
That said, I wouldn't worry too much since I plan to rip this out with my bare hands right after the 24.11 branch-off.

nixos/modules/services/networking/rosenpass.nix
Ma27: Done.

nixos/modules/tasks/network-interfaces-systemd.nix
Ma27: Yes indeed.
Used an mkMerge/mkIf chain. Tests don't seem to have changed, so it was indeed a no-op here, but still the correct thing to do I think.

fpletz (approved):
:heart:

fpletz:
We have to wait with the merge until after the branch-off since this is a breaking change and thus unfortunately too late for 24.05.

ElvishJerricco:
Isn't the idea of these `*Config` options that we could have abstractions in sibling options that get rendered into the `*Config` option? Granted, I guess that hasn't actually happened yet, or else this PR wouldn't be possible.

nixos/modules/tasks/network-interfaces-systemd.nix
lheckemann: I meant as in https://github.com/lheckemann/nixpkgs/commit/cf50bd0aa1a0d9271fd0c547f435b5c9a2a0f2c0 -- without mkMerge.

Ma27:
@ElvishJerricco for systemd.network, the structure has pretty much always been a direct mapping to how the .ini files look like, no? (Sure, with a few modifications such as using lists for sections that can appear multiple times).

I don't expect much to happen there given what the options are for. Also, I don't know anyone who likes the way this is currently structured and the usually recommended "workaround" is using "extraConfig" which is IMHO worse since it doesn't have any eval-time validation, so I'd strongly prefer my solution over the current state.

fpletz:
As 24.05 got branched off let's merge now and fix possible fallout of this change in the next weeks.

PR #312811, opened 2024-05-19
linuxPackages.nvidia_x11.libXNVCtrl: make the shared library available

aidalgol:
## Description of changes
The libXNVCtrl library is used by several programs in nixpkgs to get Nvidia GPU stats, and they all load the library dynamically, but the libXNVCtrl package only builds a static library.  There is no build flag to enable this, so we have to patch the Makefile to produce a shared library.  We then add the package to `hardware.opengl.extraPackages` in the NixOS module `hardware/video/nvidia.nix` so that it is available on Nvidia-enabled systems.

Based on a patch by @negativo17 for Fedora: https://github.com/negativo17/nvidia-settings/blob/master/nvidia-settings-libXNVCtrl.patch

## Things done
- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/nixos/modules/hardware/video/nvidia.nix
+++ b/nixos/modules/hardware/video/nvidia.nix
@@ -285,7 +285,7 @@ in
             KERNEL=="nvidia_uvm", RUN+="${pkgs.runtimeShell} -c 'mknod -m 666 /dev/nvidia-uvm-tools c $$(grep nvidia-uvm /proc/devices | cut -d \  -f 1) 1'"
           '';
           hardware.opengl = {
-            extraPackages = [ nvidia_x11.out ];
+            extraPackages = [ nvidia_x11.out nvidia_x11.settings.libXNVCtrl ];
             extraPackages32 = [ nvidia_x11.lib32 ];
           };
           environment.systemPackages = [ nvidia_x11.bin ];
--- a/pkgs/os-specific/linux/nvidia-x11/settings.nix
+++ b/pkgs/os-specific/linux/nvidia-x11/settings.nix
@@ -147,6 +147,6 @@ stdenv.mkDerivation {
     license = licenses.unfreeRedistributable;
     platforms = nvidia_x11.meta.platforms;
     mainProgram = "nvidia-settings";
-    maintainers = with maintainers; [ abbradar ];
+    maintainers = with maintainers; [ abbradar aidalgol ];
   };
 }
--- a/pkgs/os-specific/linux/nvidia-x11/settings.nix
+++ b/pkgs/os-specific/linux/nvidia-x11/settings.nix
@@ -43,6 +43,14 @@ let
 
     makeFlags = [
       "OUTPUTDIR=." # src/libXNVCtrl
+      "libXNVCtrl.a"
+      "libXNVCtrl.so"
+    ];
+
+    patches = [
+      # Patch the Makefile to also produce a shared library.
+      (if lib.versionOlder nvidia_x11.settingsVersion "400" then ./libxnvctrl-build-shared-3xx.patch
+      else ./libxnvctrl-build-shared.patch)
     ];
 
     installPhase = ''
@@ -52,6 +60,7 @@ let
       cp libXNVCtrl.a $out/lib
       cp NVCtrl.h     $out/include/NVCtrl
       cp NVCtrlLib.h  $out/include/NVCtrl
+      cp -P libXNVCtrl.so* $out/lib
     '';
   };
 
--- a/pkgs/os-specific/linux/nvidia-x11/libxnvctrl-build-shared-3xx.patch
+++ b/pkgs/os-specific/linux/nvidia-x11/libxnvctrl-build-shared-3xx.patch
@@ -1,0 +1,24 @@
+--- a/src/libXNVCtrl/Makefile
++++ b/src/libXNVCtrl/Makefile
+@@ -33,6 +33,8 @@
+ 
+ LIBXNVCTRL = libXNVCtrl.a
+ 
++LIBXNVCTRL_SHARED = $(OUTPUTDIR)/libXNVCtrl.so
++
+ LIBXNVCTRL_PROGRAM_NAME = "libXNVCtrl"
+ 
+ LIBXNVCTRL_VERSION := $(NVIDIA_VERSION)
+@@ -62,6 +64,12 @@
+ $(LIBXNVCTRL) : $(OBJS)
+ 	$(AR) ru $@ $(OBJS)
+ 
++$(LIBXNVCTRL_SHARED): $(LIBXNVCTRL_OBJ)
++	$(RM) $@ $@.*
++	$(CC) -shared -Wl,-soname=$(@F).0 -o $@.0.0.0 $(LDFLAGS) $^ -lXext -lX11
++	ln -s $(@F).0.0.0 $@.0
++	ln -s $(@F).0 $@
++
+ # define the rule to build each object file
+ $(foreach src,$(SRC),$(eval $(call DEFINE_OBJECT_RULE,TARGET,$(src))))
+ 
--- a/pkgs/os-specific/linux/nvidia-x11/libxnvctrl-build-shared.patch
+++ b/pkgs/os-specific/linux/nvidia-x11/libxnvctrl-build-shared.patch
@@ -1,0 +1,21 @@
+--- a/src/libXNVCtrl/xnvctrl.mk
++++ b/src/libXNVCtrl/xnvctrl.mk
+@@ -39,6 +39,8 @@
+ 
+ LIBXNVCTRL = $(OUTPUTDIR)/libXNVCtrl.a
+ 
++LIBXNVCTRL_SHARED = $(OUTPUTDIR)/libXNVCtrl.so
++
+ LIBXNVCTRL_SRC = $(XNVCTRL_DIR)/NVCtrl.c
+ 
+ LIBXNVCTRL_OBJ = $(call BUILD_OBJECT_LIST,$(LIBXNVCTRL_SRC))
+@@ -47,3 +49,9 @@
+ 
+ $(LIBXNVCTRL) : $(LIBXNVCTRL_OBJ)
+ 	$(call quiet_cmd,AR) ru $@ $(LIBXNVCTRL_OBJ)
++
++$(LIBXNVCTRL_SHARED): $(LIBXNVCTRL_OBJ)
++	$(RM) $@ $@.*
++	$(CC) -shared -Wl,-soname=$(@F).0 -o $@.0.0.0 $(LDFLAGS) $^ -lXext -lX11
++	ln -s $(@F).0.0.0 $@.0
++	ln -s $(@F).0 $@


pkgs/os-specific/linux/nvidia-x11/settings.nix
```diff
@@ -45,13 +45,22 @@ let
       "OUTPUTDIR=." # src/libXNVCtrl
     ];
 
+    patches = [
+      # Patch the Makefile to also produce a shared library.
+      (if (lib.versions.major nvidia_x11.settingsVersion) == "470" then ./libxnvctrl-build-shared-470.patch
+      else if (lib.versions.major nvidia_x11.settingsVersion) == "390" then ./libxnvctrl-build-shared-390.patch
+      else if (lib.versions.major nvidia_x11.settingsVersion) == "340" then ./libxnvctrl-build-shared-340.patch
```
ConnorBaker, line 52: Function application has a higher precedence than equality so there's no need for parenthesis around the calls to `lib.versions.major`.

ConnorBaker (commented):
One minor nit; I can try the builds later to verify they build.

Do you have an easy way to verify this is working as intended?

Also, would you mind adding yourself to the list of maintainers for pkgs/os-specific/linux/nvidia-x11/settings.nix?

pkgs/os-specific/linux/nvidia-x11/libxnvctrl-build-shared-340.patch
SomeoneSerge, line 1: Looking at https://github.com/NVIDIA/nvidia-settings/blob/4e4bb3fecbbdd902d6f4bc6bbfb9902043c5c7f6/src/libXNVCtrl/Makefile#L39-L53, it seems that we could build the `.so` just by invoking `make` with appropriate arguments. I think that'd be more maintainable than patching `Makefile` per-release.

If we wish to build both the static and the shared object in a single derivation we'd still have to define a custom `buildPhase`, but I think that's OK.

aidalgol:
> Do you have an easy way to verify this is working as intended?

The program I was trying to get to work with my nvidia GPU was MangoHUD, so running `vkcube` with that was how I was testing my fix.

Before changes:
```console
❯ mangohud vkcube
Selected GPU 0: NVIDIA GeForce RTX 3080, type: DiscreteGpu
[2024-05-20 06:44:07.254] [MANGOHUD] [error] [loader_nvctrl.cpp:39] Failed to open 64bit libXNVCtrl.so.0: libXNVCtrl.so.0: cannot open shared object file: No such file or directory
[2024-05-20 06:44:07.254] [MANGOHUD] [error] [nvctrl.cpp:48] XNVCtrl loader failed to load
```

After changes:
```console
❯ mangohud vkcube
Selected GPU 0: NVIDIA GeForce RTX 3080, type: DiscreteGpu
```

(With `gpu_stats` in my `~/.config/MangoHud/MangoHud.conf`.)

pkgs/os-specific/linux/nvidia-x11/libxnvctrl-build-shared-340.patch
aidalgol: There's no make rule for a `.so` file in `xnvctrl.mk` (or anywhere that I can see), so I don't think we can do this via `make` arguments.  We could possibly just bypass `make` entirely and invoke the commands directly in a custom `buildPhase`, but I'm not sure whether that would be more maintainable.

pkgs/os-specific/linux/nvidia-x11/settings.nix
aidalgol: Fixed.

pkgs/os-specific/linux/nvidia-x11/libxnvctrl-build-shared-340.patch
SomeoneSerge: Oh I notice the `$(AR)` part, I was naive to think they had some kind of a generic rule for linkage...

SomeoneSerge: I'd still like to get rid of the conditional patches. A part of the diff doesn't seem to be essential to the derivation, e.g. the `clean` and `all` rules. Maybe omitting those we could have a single patch that successfully applies to all releases: adding nothing but the `.so` rule.

aidalgol: I was able to simplify down to two patches, because `xnvctrl.mk` does not exist in the `3xx` versions.

SomeoneSerge (approved):
I've some general comments left, in the line of "are we solving the right problem":

1. It seems that patching `Makefile`s takes approximately the same amount of code as shipping our own ad hoc `CMakeLists.txt` would have...
2. It seems that upstream does not distribute shared libraries, so it feels kind of arbitrary that MangoHUD should use it as one.
3. Further on MangoHUD, these probably should be in `buildInputs`: https://github.com/NixOS/nixpkgs/blob/dbeb624729a7c6308d6a945b0a1fbafd001c2813/pkgs/tools/graphics/mangohud/default.nix#L190-L194

That said, if you @aidalgol are convinced this is the way to address your use-case, I'll merge

aidalgol:
> 1. It seems that patching `Makefile`s takes approximately the same amount of code as shipping our own ad hoc `CMakeLists.txt` would have...

Do you mean for all of `nvidia_x11`, or just `libXNVCtrl`?

> 2. It seems that upstream does not distribute shared libraries, so it feels kind of arbitrary that MangoHUD should use it as one.

MangoHUD is not the only package in nixpkgs that references `libXNVCtrl`.

- `psensor`
- `nvfancontrol`
- `monitor` (yes, it's really called just "monitor")
- `vibrantlinux`
- `conky`

But I don't know whether any of those try to link dynamically.  It sounds as if Conky does: [Installing libxnvctrl (525.60.11-2) breaks dependency ‘libXNVCtrl.so=0-64’ required by conky](https://forum.manjaro.org/t/installing-libxnvctrl-525-60-11-2-breaks-dependency-libxnvctrl-so-0-64-required-by-conky/127981/2).

Looking at other Linux distributions, at least Arch patches `libXNVCtrl` similarly.

- https://bugs.archlinux.org/task/76702
- https://archlinux.org/packages/extra/x86_64/libxnvctrl/

> 3. Further on MangoHUD, these probably should be in `buildInputs`: https://github.com/NixOS/nixpkgs/blob/dbeb624729a7c6308d6a945b0a1fbafd001c2813/pkgs/tools/graphics/mangohud/default.nix#L190-L194

That I'm not too sure about.  If someone is using `mangohud` installed via "standalone" home-manager, and they update that more frequently than their system configuration, wouldn't `mangohud` have a newer version than the system's driver?

SomeoneSerge:
> Do you mean for all of nvidia_x11, or just libXNVCtrl?

libXNVCtrl

> Looking at other Linux distributions, at least Arch patches libXNVCtrl similarly.

Seems like the patch belongs in NVIDA/nvidia-settings then, let's aim to eventually move it there.

SomeoneSerge:
> You mean upstream it?

Yes

> In that case, I would not be opposed to instead replacing the upstream Makefile, if you think that would be simpler or easier to maintain.

I'd be in favour of that. But also if we're upstreaming a patch, we'll want to switch to `fetchpatch`, and it's safer to avoid trying to impose cmake (or anything) on people...

If you need time to choose but you're committed to following up on this, I can merge the current PR as a temporary fix

aidalgol:
I have submitted a PR upstream and have changed the `libXNVCtrl` derivation to use the diff from the PR, except for the `3xx` legacy version.

pkgs/os-specific/linux/nvidia-x11/settings.nix
```diff
@@ -45,13 +45,23 @@ let
       "OUTPUTDIR=." # src/libXNVCtrl
     ];
 
+    patches = [
+      # Patch the Makefile to also produce a shared library.
+      (if lib.versionOlder nvidia_x11.settingsVersion "400" then ./libxnvctrl-build-shared-3xx.patch
+      else (fetchpatch {
+        url = "https://github.com/NVIDIA/nvidia-settings/pull/105/commits/0ddfd1d4f871846ff79de6630ddf9f09e9acaf39.patch";
```
SomeoneSerge, line 52: - Iirc the `pull/xxxx/commits/yyyy` URL may get invalidated after force-pushing or even rebase-merging (I may be wrong)
- The `fetchpatch` now has an updated `all` rule, but the 3xx patch doesn't

aidalgol: I think you're right.  I thought that would make it robust against PR changes, but I think GitHub does invalidate commit URLs on a force push, at least after a while.  I think I should back out this commit until the PR is merged.

pkgs/os-specific/linux/nvidia-x11/settings.nix
```diff
@@ -43,6 +43,14 @@ let
 
     makeFlags = [
       "OUTPUTDIR=." # src/libXNVCtrl
+      "libXNVCtrl.a"
+      "libXNVCtrl.so"
```
SomeoneSerge, line 47: Since there's a merge conflict anyway, another nit: `makeTargets` is more explicit

aidalgol: I tried `makeTargets` first, and it did not work.  I don't see any mention of this attribute in the [nixpkgs manual](https://nixos.org/manual/nixpkgs/unstable/#var-stdenv-makeFlags).

SomeoneSerge: Hm, I was sure I saw the attribute being used somewhere, but I don't see any mentions in `pkgs/stdenv/generic/setup.sh` so I must be wrong

nixos/modules/hardware/video/nvidia.nix
SomeoneSerge, line 1: Let's solve conflicts, merge, and backport to the last two releases (counting 24.05)

aidalgol: Branch rebased on recent `master`.

nixos/modules/hardware/video/nvidia.nix
```diff
@@ -285,7 +285,7 @@ in
             KERNEL=="nvidia_uvm", RUN+="${pkgs.runtimeShell} -c 'mknod -m 666 /dev/nvidia-uvm-tools c $$(grep nvidia-uvm /proc/devices | cut -d \  -f 1) 1'"
           '';
           hardware.opengl = {
-            extraPackages = [ nvidia_x11.out ];
+            extraPackages = [ nvidia_x11.out nvidia_x11.settings.libXNVCtrl ];
```
SomeoneSerge, line 288: I had another look at the derivation and it actually doesn't seem to depend on `nvidia_x11`, except through `nvidia_x11.settingsVersion`. This makes me question if we actually have to deploy this impurely rather than link directly

megheaiulian, line 288: :exclamation:  This will break things. Depending on config ( `useSettings` or `nvidiaSettings` ) `nvidia_x11.settings` might be a empty set.  See https://github.com/NixOS/nixpkgs/blob/nixos-unstable/pkgs/os-specific/linux/nvidia-x11/generic.nix#L215
This should handle that case or be reverted.

aidalgol: Submitted https://github.com/NixOS/nixpkgs/pull/316516 to address this.

SomeoneSerge: @megheaiulian thanks for the heads up!

PR #119444, opened 2021-04-14
gradle: Decouple from JDK 8 and support Java Toolchains

lorenzleutgeb:
This sets a default Java installation per Gradle version, and allows to specify [Java Toolchains](https://docs.gradle.org/current/userguide/toolchains.html).

Gradle Version | Attribute | Default Java Version | `defaultJava`
---------------|-----------|----------------------|--------------
7.3 | `gradle_7` | 17 | `jdk8`
6.9.1 | `gradle_6` | 11 | `jdk11`
5.6.4 | `gradle_5` | 11 | `jdk11`
4.10.3 | `gradle_4` | 8 | `jdk8`

(see also Gradle's [Compatibility Matrix](https://docs.gradle.org/current/userguide/compatibility.html))

E.g. with `gradle_7.override { javaToolchains = [ jdk8 jdk11 ]; }` I get

```
$ gradle -q javaToolchains

 + Options
     | Auto-detection:     Enabled
     | Auto-download:      Enabled

 + N/A JDK 17.0.1+12-nixos
     | Location:           /nix/store/1vi5p9gliab13jc3v56kz7z78x5bq0da-openjdk-17.0.1+12/lib/openjdk
     | Language Version:   17
     | Vendor:             N/A
     | Is JDK:             true
     | Detected by:        Current JVM

 + OpenJDK 1.8.0_272-b10
     | Location:           /nix/store/4f0d8c9lmj174hz3qvn1qhc70n22bai4-openjdk-8u272-b10
     | Language Version:   8
     | Vendor:             Oracle
     | Is JDK:             true
     | Detected by:        environment variable 'JAVA_TOOLCHAIN_NIX_0'

 + OpenJDK 11.0.12+0-adhoc..source
     | Location:           /nix/store/7r5b2r4kjsbflgxs8sy7i86c1wy773is-openjdk-11.0.12+7
     | Language Version:   11
     | Vendor:             Oracle
     | Is JDK:             true
     | Detected by:        environment variable 'JAVA_TOOLCHAIN_NIX_1'
```

PR #312887, opened 2024-05-19
java: Populate `$out/lib/jvm`

lorenzleutgeb:
## Description of changes

This change should make it easier to detect available JDKs by providing symlinks at the canonical location `$out/lib/jvm/$name`.

In the Java ecosystem, this is a common pattern. For example, the [Gradle](https://gradle.org/) build tool supports "crawling" many of those locations:

 - [`/usr/lib{,64}/jvm`, `{/opt,/usr/{,local/}}java`](https://github.com/gradle/gradle/blob/766a6288794e4f5b78ad44e9ba8b0a2204d3c439/platforms/jvm/jvm-services/src/main/java/org/gradle/jvm/toolchain/internal/LinuxInstallationSupplier.java#L36)
 - [`$HOME/.jdks`](https://github.com/gradle/gradle/blob/766a6288794e4f5b78ad44e9ba8b0a2204d3c439/platforms/jvm/jvm-services/src/main/java/org/gradle/jvm/toolchain/internal/IntellijInstallationSupplier.java)
 - [`$HOME/.sdkman/candidates`](https://github.com/gradle/gradle/blob/766a6288794e4f5b78ad44e9ba8b0a2204d3c439/platforms/jvm/jvm-services/src/main/java/org/gradle/jvm/toolchain/internal/SdkmanInstallationSupplier.java)

The Scala tool [sbt](https://www.scala-sbt.org/) follows does [something very similar](https://github.com/sbt/sbt/blob/4e46a331d9b94cf80dfba15fa8466061bd354474/main/src/main/scala/sbt/internal/CrossJava.scala#L486-L492) and an [attempt to integrate with NixOS](https://github.com/sbt/sbt/issues/4260) has stalled.

There's been some discussion about how to make integration of these tools with NixOS easier, see https://discourse.nixos.org/t/4677 and an earlier attempt at https://github.com/NixOS/nixpkgs/pull/76699

Since the attempts by @raboof, I have made some progress with Gradle in https://github.com/NixOS/nixpkgs/pull/119444. The approach gives more control but is not ergonomic since it requires overriding Gradle to pass in JDKs. Note that this proposal is not meant to replace passing toolchains to Gradle but proposed in addition.

Recently @kravemir has reignited the discussion and has suggested to implement detection based on environments/profiles in https://github.com/gradle/gradle/pull/29214#discussion_r1605954040. This PR is an attempt to standardize a detection location per-derivation, namely `$out/lib/jvm/$name` and to enable linking these paths whenever someone sets `programs.java.enable`.

A hook is provided so that all JDKs can use the same code for creating the symlink.

Once this has landed, it'd be possible to have tools like Gradle and sbt crawl `/run/current-system/sw/lib/jvm`, `~/.nix-profile/lib/jvm`, or other locations via `nix-shell`.

For now I have only changed four JDKs, but that should be sufficient to see what's going on and test the approach. Of course I am happy to add the hook to other JDKs too.

## Alternative Considered

I also considered generating a `pkgs.linkFarm` in the Java module, but this solution is not as portable. By creating the link in each derivation, downstream package consumers can also make use of the mechanism, e.g. home-manager could also link them somewhere without having to reimplement linking.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/nixos/modules/programs/java.nix
+++ b/nixos/modules/programs/java.nix
@@ -61,6 +61,8 @@ in
       };
     };
 
+    environment.pathsToLink = [ "/lib/jvm" ];
+
     environment.systemPackages = [ cfg.package ];
 
     environment.shellInit = ''
--- a/pkgs/by-name/ho/hookLibJvm/hook.sh
+++ b/pkgs/by-name/ho/hookLibJvm/hook.sh
@@ -1,0 +1,13 @@
+# shellcheck shell=bash
+
+postInstallHooks+=(libJvm)
+
+libJvm() {
+    if [ "${dontLibJvm-}" = 1 ]; then return; fi
+
+    if [ ! -d "${!outputBin:?}" ]; then return; fi
+
+    base="/lib/jvm"
+    mkdir -p "${!outputBin:?}${base}"
+    ln -fs "${!outputBin:?}" "${!outputBin:?}${base}/${name}"
+}
--- a/pkgs/by-name/ho/hookLibJvm/package.nix
+++ b/pkgs/by-name/ho/hookLibJvm/package.nix
@@ -1,0 +1,1 @@
+{ makeSetupHook }: makeSetupHook { name = "hook-lib-jvm"; } ./hook.sh
--- a/pkgs/development/compilers/adoptopenjdk-bin/jdk-darwin-base.nix
+++ b/pkgs/development/compilers/adoptopenjdk-bin/jdk-darwin-base.nix
@@ -4,6 +4,7 @@
 , lib, stdenv
 , fetchurl
 , setJavaClassPath
+, hookLibJvm
 }:
 
 assert (stdenv.isDarwin && stdenv.isx86_64);
@@ -19,10 +20,14 @@ let cpuName = stdenv.hostPlatform.parsed.cpu.name;
     inherit (sourcePerArch.${cpuName}) url sha256;
   };
 
+  nativeBuildInputs = [ hookLibJvm ];
+
   # See: https://github.com/NixOS/patchelf/issues/10
   dontStrip = 1;
 
   installPhase = ''
+    runHook preInstall
+
     cd ..
 
     mv $sourceRoot $out
@@ -47,6 +52,8 @@ let cpuName = stdenv.hostPlatform.parsed.cpu.name;
     cat <<EOF >> $out/nix-support/setup-hook
     if [ -z "\''${JAVA_HOME-}" ]; then export JAVA_HOME=$out; fi
     EOF
+
+    runHook postInstall
   '';
 
   # FIXME: use multiple outputs or return actual JRE package
--- a/pkgs/development/compilers/adoptopenjdk-bin/jdk-linux-base.nix
+++ b/pkgs/development/compilers/adoptopenjdk-bin/jdk-linux-base.nix
@@ -20,6 +20,7 @@
 , cairo
 , glib
 , gtk3
+, hookLibJvm
 }:
 
 let
@@ -56,12 +57,14 @@ let result = stdenv.mkDerivation rec {
     zlib
   ] ++ lib.optional stdenv.isAarch32 libffi;
 
-  nativeBuildInputs = [ autoPatchelfHook makeWrapper ];
+  nativeBuildInputs = [ autoPatchelfHook makeWrapper hookLibJvm ];
 
   # See: https://github.com/NixOS/patchelf/issues/10
   dontStrip = 1;
 
   installPhase = ''
+    runHook preInstall
+
     cd ..
 
     mv $sourceRoot $out
@@ -98,6 +101,8 @@ let result = stdenv.mkDerivation rec {
         wrapProgram "$bin" --prefix LD_LIBRARY_PATH : "${runtimeLibraryPath}"
       fi
     done
+
+    runHook postInstall
   '';
 
   preFixup = ''
--- a/pkgs/development/compilers/openjdk/21.nix
+++ b/pkgs/development/compilers/openjdk/21.nix
@@ -4,6 +4,7 @@
 , libXi, libXinerama, libXcursor, libXrandr, fontconfig, openjdk21-bootstrap
 , ensureNewerSourcesForZipFilesHook
 , setJavaClassPath
+, hookLibJvm
 # TODO(@sternenseemann): gtk3 fails to evaluate in pkgsCross.ghcjs.buildPackages
 # which should be fixable, this is a no-rebuild workaround for GHC.
 , headless ? stdenv.targetPlatform.isGhcjs
@@ -32,7 +33,7 @@ let
       hash = "sha256-fA8nRWBuTL87S8mwapmNfCPPQoI2aKHjbHJ6PDN3khs=";
     };
 
-    nativeBuildInputs = [ pkg-config autoconf unzip ensureNewerSourcesForZipFilesHook ];
+    nativeBuildInputs = [ pkg-config autoconf unzip ensureNewerSourcesForZipFilesHook hookLibJvm ];
     buildInputs = [
       cpio file which zip perl zlib cups freetype alsa-lib libjpeg giflib
       libpng zlib lcms2 libX11 libICE libXrender libXext libXtst libXt libXtst
@@ -114,6 +115,8 @@ let
     buildFlags = [ "images" ];
 
     installPhase = ''
+      runHook preInstall
+
       mkdir -p $out/lib
 
       mv build/*/images/jdk $out/lib/openjdk
@@ -139,6 +142,8 @@ let
       ''}
 
       ln -s $out/lib/openjdk/bin $out/bin
+
+      runHook postInstall
     '';
 
     preFixup = ''
--- a/pkgs/development/compilers/openjdk/22.nix
+++ b/pkgs/development/compilers/openjdk/22.nix
@@ -34,6 +34,7 @@
 , openjdk22-bootstrap
 , ensureNewerSourcesForZipFilesHook
 , setJavaClassPath
+, hookLibJvm
   # TODO(@sternenseemann): gtk3 fails to evaluate in pkgsCross.ghcjs.buildPackages
   # which should be fixable, this is a no-rebuild workaround for GHC.
 , headless ? stdenv.targetPlatform.isGhcjs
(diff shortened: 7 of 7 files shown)


pkgs/by-name/sh/share-jvm/hook.sh
```diff
@@ -0,0 +1,12 @@
...
+
+postInstallHooks+=(shareJvm)
+
+shareJvm() {
+    if [ "${dontShareJvm-}" = 1 ]; then return; fi
+
+    if [ ! -d "${!outputBin:?}" ]; then return; fi
+
+    mkdir -p "${!outputBin:?}/share/jvm"
+    ln -fs "${!outputBin:?}" "${!outputBin:?}/share/jvm/${name}"
```
kravemir, line 11: I would rather opt for `lib`, which is usually meant for platform specific things, and is already more of a convention in Gradle:

> [/usr/lib{,64}/jvm, {/opt,/usr/{,local/}}java](https://github.com/gradle/gradle/blob/766a6288794e4f5b78ad44e9ba8b0a2204d3c439/platforms/jvm/jvm-services/src/main/java/org/gradle/jvm/toolchain/internal/LinuxInstallationSupplier.java#L36)

```suggestion
    mkdir -p "${!outputBin:?}/lib/jvm"
    ln -fs "${!outputBin:?}" "${!outputBin:?}/share/lib/${name}"
```


The `share` is described in [Wiki about FHS](https://en.wikipedia.org/wiki/Filesystem_Hierarchy_Standard) as:

```
Architecture-independent (shared) data.
```

And, JVM is architecture dependent.

---

Though, IMO, `lib` isn't totally correct either. As JVM is collection of binaries and libraries. But, it's an already adopted pattern.

kravemir (commented):
I'm only Nix beginner, so I can't review Nix files reliably.

One comment (concern) about the new location pattern to be founded here (`share` instead of `lib`).

Otherwise, the concept LGTM.

raboof:
I'm somewhat reluctant about this change: 
* for 'end-user' programs, the JDK should just be specified in the derivation, so for those you shouldn't need this module
* for 'developer' environments, the 'appropriate' JDKs might differ per project, so installing them 'globally' as a module doesn't make too much sense.

Having let this simmer for a while, I think I would prefer the approach of having each JDK contribute itself to a `JAVA_HOMES` environment variable (similar to how it sets the `JAVA_HOME`, but allowing multiple values).

That said, as this topic has been stagnant for a while, I'd be supportive of this change as well, just to get things unstuck and gain some more experience on how this works out in practice ;)

lorenzleutgeb:
> * for 'end-user' programs, the JDK should just be specified in the derivation, so for those you shouldn't need this module

I agree. This is not the intended use-case here. I mention how it is possible to pass JDKs to Gradle, for example, which fits this use-case.

> * for 'developer' environments, the 'appropriate' JDKs might differ per project, so installing them 'globally' as a module doesn't make too much sense.

This is why it is opt-in (currently tied to `programs.java.enable`). Note however, that the proposal also works with regular profiles, without NixOS. You don't have to use the "global" profile via `/run/current-system/sw/share/jvm`, you can point your detection mechanism at an arbitrary profile/environment. For example:

```
$ nix profile install .#adoptopenjdk-hotspot-bin-8 --priority 7
$ ls ~/.nix-profile/share/jvm
adoptopenjdk-hotspot-bin-8.0.372
$ nix profile install .#adoptopenjdk-hotspot-bin-11 --priority 7
$ ls ~/.nix-profile/share/jvm
adoptopenjdk-hotspot-bin-8.0.372  adoptopenjdk-hotspot-bin-11.0.19
$ nix profile list
Index:              0
Store paths:        [...]

Index:              1
Store paths:        /nix/store/...-home-manager-path

Index:              2
Flake attribute:    legacyPackages.x86_64-linux.adoptopenjdk-hotspot-bin-8
Original flake URL: git+file:///home/lorenz/src/github.com/NixOS/nixpkgs
Locked flake URL:   git+file:///home/lorenz/src/github.com/NixOS/nixpkgs?ref=refs/heads/jvm&rev=3a9db7691302d62a6d9020c927a791f0ddbfba86
Store paths:        /nix/store/4q4d79yp3sj9310kcxajh6y7gyjg5j4q-adoptopenjdk-hotspot-bin-8.0.372

Index:              3
Flake attribute:    legacyPackages.x86_64-linux.adoptopenjdk-hotspot-bin-11
Original flake URL: git+file:///home/lorenz/src/github.com/NixOS/nixpkgs
Locked flake URL:   git+file:///home/lorenz/src/github.com/NixOS/nixpkgs?ref=refs/heads/jvm&rev=3a9db7691302d62a6d9020c927a791f0ddbfba86
Store paths:        /nix/store/yd3x5x25k66byrwrl6qq7z3caqqqips0-adoptopenjdk-hotspot-bin-11.0.19
% nix profile remove /nix/store/4q4d79yp3sj9310kcxajh6y7gyjg5j4q-adoptopenjdk-hotspot-bin-8.0.372
removing 'git+file:///home/lorenz/src/github.com/NixOS/nixpkgs#legacyPackages.x86_64-linux.adoptopenjdk-hotspot-bin-8'
$ ls ~/.nix-profile/share/jvm
adoptopenjdk-hotspot-bin-11.0.19
```

I believe that this is also useful with `nix-shell`. Consider the following `shell.nix` placed at the root of nixpkgs:

```nix
let
  pkgs = import ./default.nix {};
  env = pkgs.buildEnv {
    name = "development";
    paths = with pkgs; [
      (adoptopenjdk-hotspot-bin-8  // { meta.priority = 6; })
      (adoptopenjdk-hotspot-bin-11 // { meta.priority = 5; })
    ];
  };
in
pkgs.mkShellNoCC {
  packages = [ env ];
  shellHook = ''
    echo "Your JVMs are in '${env}/share/jvm' check it out:"
    ls -o ${env}/share/jvm
  '';
}
```

This gives you:
```
$ nix-shell
Your JVMs are in '/nix/store/wn8c7mvnrlw83gsm6laq24haq94xp1b1-development/share/jvm' check it out:
total 8
lrwxrwxrwx 3 root 119 Jan  1  1970 adoptopenjdk-hotspot-bin-11.0.19 -> /nix/store/yd3x5x25k66byrwrl6qq7z3caqqqips0-adoptopenjdk-hotspot-bin-11.0.19/share/jvm/adoptopenjdk-hotspot-bin-11.0.19
lrwxrwxrwx 3 root 119 Jan  1  1970 adoptopenjdk-hotspot-bin-8.0.372 -> /nix/store/4q4d79yp3sj9310kcxajh6y7gyjg5j4q-adoptopenjdk-hotspot-bin-8.0.372/share/jvm/adoptopenjdk-hotspot-bin-8.0.372
```

> Having let this simmer for a while, I think I would prefer the approach of having each JDK contribute itself to a `JAVA_HOMES` environment variable (similar to how it sets the `JAVA_HOME`, but allowing multiple values).

I know that this is your preferred approach, I have linked https://github.com/NixOS/nixpkgs/pull/76699 above, but I think that what I am proposing here both fits the "Nix way" (convention on derivation outputs instead of environment variables) and also the "Java ecosystem way" (crawling magic directories).

pkgs/by-name/sh/share-jvm/hook.sh
lorenzleutgeb: I have skimmed [FHS 2.3](https://refspecs.linuxfoundation.org/FHS_2.3/fhs-2.3.pdf) before choosing `$out/share/jvm`. It reads like from another planet. I think that `/lib` is wrong, `/share` is also wrong, and the FHS won't really help you because it is just not written to spec what we are doing here.

`/share` is (ab)used by many others: https://github.com/NixOS/nixpkgs/blob/e7cc61784ddf51c81487637b3031a6dd2d6673a2/nixos/modules/config/system-path.nix#L154-L165

rhendric:
Forgive the bikeshedding, but why `share`? A JVM is mostly object files, so I'd have expected `$out/lib/jvm` (and this is a closer match to the `/usr/lib/jvm` path expected by Gradle and SBT).

pkgs/by-name/sh/share-jvm/hook.sh
rhendric: `lib` canonically doesn't have to consist of *only* binaries. `pkgconfig` files go in there, for example. The rule of thumb is architecture-dependent or not, and I think all the things I recognize in that system-path.nix excerpt are architecture-independent. A JVM isn't.

lorenzleutgeb: Hmm, the biggest reason for me to avoid `/lib` was the corresponding comment suggesting that it could be removed:

https://github.com/NixOS/nixpkgs/blob/e7cc61784ddf51c81487637b3031a6dd2d6673a2/nixos/modules/config/system-path.nix#L152

However this FIXME has been sitting there for about ~10 years, see https://github.com/NixOS/nixpkgs/commit/e636e0a5329f546ebcc603d56bb99a0d016370f8#diff-5c7b357730b03feb709650ecc8c17df225af3b31ad01b5142de81b33c4f72639R106

So maybe `$out/lib` would also be okay. If this somehow blocks merging, I'm happy to go for `$out/lib`. But I'd be very curious to understand why `/lib` should be removed.

lorenzleutgeb: Alright, I am changing to `$out/lib/jvm`. @rhendric can you enlighten me regarding the FIXME?

rhendric: > @rhendric can you enlighten me regarding the FIXME?

:sweat_smile: I have no idea what, if anything, is wrong with linking `/lib`; looks like an @edolstra -ism.

A theory is that, since the [file referenced](https://github.com/NixOS/nixpkgs/blob/40c21be908c26dc2b1bc749cde92f46cac92a3a2/nixos/modules/config/debug-info.nix) refers to wanting to link a specific subpath under `/lib`, if the `/lib` entry were to be removed for some reason it would still be correct to add `/lib/jvm` back in.

lorenzleutgeb:
Trying once more to get some attention. Pinging @NixOS/java after @tomodachi94's [comment](https://github.com/NixOS/nixpkgs/pull/294121#pullrequestreview-2471235726). Of course I do realize that there are merge conflicts currently, and I would of course resolve them, but I'd love to get some feedback before I invest more time.

tomodachi94 (commented):
A few preliminary comments. I think it would be awesome if this behavior was specified in a [JEP](https://openjdk.org/jeps/0) (specifically, what the subfolder inside `lib/jvm` should be named and what files it should contain), but that's out of the scope of this PR.

It would also be awesome if our Gradle/Maven/etc was patched or had a setup hook to support this out-of-the-box, but again, this is out-of-scope.

Additionally: Should we add this to other Java distributions?

tomodachi94:
That said, I'm generally supportive of the general direction this is going in.

PR #312945, opened 2024-05-19
texlive.withPackages: use mkDerivation (finalAttrs: { ... }) instead of buildEnv

xworld21:
## Description of changes
Refactor `texlive.withPackages` to use `mkDerivation (finalAttrs: { ... })` instead of `buildEnv`.

The main advantage is having `texliveSmall.overrideAttrs { withDocs = true; }`, which will allow us to deprecate `texlive.combine`, at last. Further benefit is that the build follows traditional phases, so one can add custom behaviours by setting the relevant pre/post-hooks.

Cons: I had to vendor a copy of the `buildEnv` script.

## Things done
- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [X] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [X] Tested, as applicable:

---


veprbl:
Any plan to de-duplicate with the buildEnv script?

xworld21:
> Any plan to de-duplicate with the buildEnv script?

Am I allowed to use the one in `build-support/build-env`? At some point I got the notion that folders should be self contained, but maybe it's just a `by-name` rule. I wouldn't mind using the same script.

xworld21:
The last update (which prevents the overrides from affecting the formats) seems to fix the performance regression:
image
The previous run reported some extra 60s – it might have been a CI fluke, but it got me worried.

veprbl:
> > Any plan to de-duplicate with the buildEnv script?
> 
> Am I allowed to use the one in `build-support/build-env`? At some point I got the notion that folders should be self contained, but maybe it's just a `by-name` rule. I wouldn't mind using the same script.

I think, you are allowed. If we can get the script no by path but via an attribute, that would be better.

veprbl:
Or, perhaps, converting buildEnv to mkDerivation is an option?

xworld21:
> I think, you are allowed. If we can get the script no by path but via an attribute, that would be better.

I have done so (hacky, because `buildEnv` is a function, but it's not outrageous).

> Or, perhaps, converting buildEnv to mkDerivation is an option?

It may be too distruptive. `buildEnv` is written on the assumption that there won't be any phases, at best a kind of build phase (just so you can run `postBuild`). The new `buildTeXEnv` is split into unpack, configure, build, install and it runs fixup. I'd say they are too different at the moment.

A better solution, at the next round of improvements maybe, is to use `buildEnv` to create an 'unwrapped' output, and let `buildTeXEnv` deal only with actual building (formats, `bin/` wrappers, ConTeXt cache and the likes).

xworld21:
@ofborg eval

(Checking if the 30s eval increase is repeatable.)

PR #264902, opened 2023-11-01
protobuf: 24.4 -> 25.3

GaetanLepage:
## Description of changes

- ~~Add `protobuf` v25.2.~~
- Set it as the default version (previously v24.4).

**Note:** I had to manually set the abseil version to a more recent one (202308) instead of the default one (202301).
Maybe it should be consider to bump the default `abseil-cpp` to a newer version.

Changelog: https://github.com/protocolbuffers/protobuf/releases/tag/v25.2

cc @tobim @jonringer 

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [x] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
- For non-Linux: Is sandboxing enabled in `nix.conf`? (See [Nix manual](https://nixos.org/manual/nix/stable/command-ref/conf-file.html))
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:
  - [NixOS test(s)](https://nixos.org/manual/nixos/unstable/index.html#sec-nixos-tests) (look inside [nixos/tests](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests))
  - and/or [package tests](https://nixos.org/manual/nixpkgs/unstable/#sec-package-tests)
  - or, for functions and "core" functionality, tests in [lib/tests](https://github.com/NixOS/nixpkgs/blob/master/lib/tests) or [pkgs/test](https://github.com/NixOS/nixpkgs/blob/master/pkgs/test)
  - made sure NixOS tests are [linked](https://nixos.org/manual/nixpkgs/unstable/#ssec-nixos-tests-linking) to the relevant packages
- [ ] Tested compilation of all packages that depend on this change using `nix-shell -p nixpkgs-review --run "nixpkgs-review rev HEAD"`. Note: all changes have to be committed, also see [nixpkgs-review usage](https://github.com/Mic92/nixpkgs-review#usage)
- [ ] Tested basic functionality of all binary files (usually in `./result/bin/`)
- [23.11 Release Notes](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2311.section.md) (or backporting [23.05 Release notes](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2305.section.md))
  - [ ] (Package updates) Added a release notes entry if the change is major or breaking
  - [ ] (Module updates) Added a release notes entry if the change is significant
[...]

PR #312957, opened 2024-05-19
protobuf_25.tests.pythonProtobuf: disable faulty test

knedlsepp:
## Description of changes

In order to enable bumping the default protobuf version from protobuf_24 to protobuf_25, we address the build failure of pythonPackages.protobuf against that version.
Protobuf's python package is moving away from cpp backend in favor of a µpb backend.
(See: https://github.com/protocolbuffers/protobuf/tree/main/upb) The work on that seems to have lead to the introduction of a broken test in "minimal_test.py":
https://github.com/protocolbuffers/protobuf/commit/501ececd39fed69f0c50c2c87f04ce75932dde72 I suspect that this is not an issue on the nixpkgs packaging end but rather that this file is uncovered code upstream. I don't know enough about Bazel to be sure, but it looks like that file is not part of their protobuf/python/BUILD.bazel file.
(I wanted to prove that in
https://github.com/protocolbuffers/protobuf/pull/16888 but couldn't trigger upstream's CI)
So for now let's just skip that file.
Note that protobuf_26.tests.pythonProtobuf is still broken. This is due to the fact that upstream removed support for building the library directly from the GitHub repo. (See:
https://github.com/protocolbuffers/protobuf/pull/15708) Conda packaging is also currently struggling with this: https://github.com/conda-forge/protobuf-feedstock/pull/215

Related: #264902

cc @GaetanLepage, we may be able to rebase #264902 on this.

### Next steps

I also looked into how to get the same working for the newer protobuf 26 and the upcoming 27 releases. I gave up on the option to generate the source package via `bazel`. Instead we will probably want to use the PyPI tarballs generated from the GitHub repo via bazel. One downside is however that these tarballs don't ship any tests, which makes verifying the correctness of the build harder.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/development/python-modules/protobuf/default.nix
+++ b/pkgs/development/python-modules/protobuf/default.nix
@@ -104,6 +104,8 @@ buildPythonPackage {
     #   https://github.com/protocolbuffers/protobuf/commit/5abab0f47e81ac085f0b2d17ec3b3a3b252a11f1
     #
     "google/protobuf/internal/generator_test.py"
+  ] ++ lib.optionals (lib.versionAtLeast protobuf.version "25") [
+    "minimal_test.py" # ModuleNotFoundError: No module named 'google3'
   ];
 
   pythonImportsCheck = [
@@ -122,6 +124,6 @@ buildPythonPackage {
     maintainers = with maintainers; [ knedlsepp ];
     # Tests are currently failing because backend is unavailable and causes tests to fail
     # Progress tracked in https://github.com/NixOS/nixpkgs/pull/264902
-    broken = lib.versionAtLeast protobuf.version "25";
+    broken = lib.versionAtLeast protobuf.version "26";
   };
 }


GaetanLepage (approved):
Thanks @knedlsepp for this temporary fix.
I had this on my todo list for a while but got discouraged by the new (from 26) situation.
I agree that dealing with bazel doesn't look appealing at all.

Let's have this merged and then work on v26.

SomeoneSerge:
> Instead we will probably want to use the PyPI tarballs generated from the GitHub repo via bazel

I think what we rather want is to reproduce the python source in a simple derivation without Bazel. I think @GaetanLepage and I had looked into that at some point, only we went with trying to decipher the bazel scripts. These proved less than comprehensible. The PR you referenced includes a table ([rendered](https://md.someonex.net/s/LqtQb1mVr#)), maybe we can use that as a starting point.

I believe we should aim for a better transparency standard than pypi tarballs or bazel...

pkgs/development/python-modules/protobuf/default.nix
```diff
@@ -104,6 +104,8 @@ buildPythonPackage {
     #   https://github.com/protocolbuffers/protobuf/commit/5abab0f47e81ac085f0b2d17ec3b3a3b252a11f1
     #
     "google/protobuf/internal/generator_test.py"
+  ] ++ lib.optionals (lib.versionAtLeast protobuf.version "25") [
+    "minimal_test.py" # ModuleNotFoundError: No module named 'google3'
```
SomeoneSerge, line 108: Is it the interpretation that `google3` is a package that comes with upb?

knedlsepp: I don't think so. Grepping through the repo just makes it look like an oversight during moving code around.

SomeoneSerge: Let's update the comment to reflect that, "Refers to a non-existent module 'google3'. Presumably an oversight". An issue upstream?

knedlsepp: I mentioned it in the commit message. My rationale was that I wanted to keep the comment short as people can always check if they get the same error message.

PR #309580, opened 2024-05-06
cargo,clippy,rustc,rustfmt: 1.77.2 -> 1.78.0 

alyssais:
## Description of changes

I guess we want to hold this until breaking changes are allowed again, since we have packages using nightly features that have been renamed.

Built:

- [x] fd on aarch64-linux
- [x] fd on x86_64-linux
- [x] fd on x86_64-darwin
- [x] firefox on aarch64-linux
- [x] firefox on x86_64-linux
- [x] firefox-esr on aarch64-linux
- [x] firefox-esr on x86_64-linux
- [x] thunderbird on aarch64-linux
- [x] thunderbird on x86_64-linux
- [x] betterbird on aarch64-linux
- [x] librewolf on aarch64-linux
- [x] floorp on aarch64-linux
- [x] nixosTests.kernel-rust on aarch64-linux
- [x] nixosTests.kernel-rust on x86_64-linux

I haven't fixed thunderbirdPackages.thunderbird-102, but it's been marked insecure since September, and I think we should just remove it: #308820.

**Could somebody please check fd on x86_64-darwin?**

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [x] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
- For non-Linux: Is sandboxing enabled in `nix.conf`? (See [Nix manual](https://nixos.org/manual/nix/stable/command-ref/conf-file.html))
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:
  - [NixOS test(s)](https://nixos.org/manual/nixos/unstable/index.html#sec-nixos-tests) (look inside [nixos/tests](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests))
  - and/or [package tests](https://github.com/NixOS/nixpkgs/blob/master/pkgs/README.md#package-tests)
  - or, for functions and "core" functionality, tests in [lib/tests](https://github.com/NixOS/nixpkgs/blob/master/lib/tests) or [pkgs/test](https://github.com/NixOS/nixpkgs/blob/master/pkgs/test)
  - made sure NixOS tests are [linked](https://nixos.org/manual/nixpkgs/unstable/#ssec-nixos-tests-linking) to the relevant packages
[...]

PR #285333, opened 2024-01-31
llvmPackages: 16.0.6 -> 17.0.6 on Linux

alyssais:
## Description of changes

rustc is already using LLVM 17, so as long as we don't find that anything widely used part of a NixOS system has a hard requirement of LLVM 16 over LLVM 17, it makes sense to upgrade the default to minimise the number of LLVMs in closures.

All my systems build fine with LLVM 17 as the default.

[Hydra jobset](https://hydra.nixos.org/jobset/nixpkgs/pr-285333-llvmPackages_17-linux)
[Comparison](https://hydra.nixos.org/eval/1804042?compare=1804006&full=0)

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [x] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
- For non-Linux: Is sandboxing enabled in `nix.conf`? (See [Nix manual](https://nixos.org/manual/nix/stable/command-ref/conf-file.html))
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:
  - [NixOS test(s)](https://nixos.org/manual/nixos/unstable/index.html#sec-nixos-tests) (look inside [nixos/tests](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests))
  - and/or [package tests](https://nixos.org/manual/nixpkgs/unstable/#sec-package-tests)
  - or, for functions and "core" functionality, tests in [lib/tests](https://github.com/NixOS/nixpkgs/blob/master/lib/tests) or [pkgs/test](https://github.com/NixOS/nixpkgs/blob/master/pkgs/test)
  - made sure NixOS tests are [linked](https://nixos.org/manual/nixpkgs/unstable/#ssec-nixos-tests-linking) to the relevant packages
- [ ] Tested compilation of all packages that depend on this change using `nix-shell -p nixpkgs-review --run "nixpkgs-review rev HEAD"`. Note: all changes have to be committed, also see [nixpkgs-review usage](https://github.com/Mic92/nixpkgs-review#usage)
- [ ] Tested basic functionality of all binary files (usually in `./result/bin/`)
[...]

PR #312617, opened 2024-05-18
llvmPackages: 17.0.6 -> 18.1.5 on other platforms

alyssais:
## Description of changes

This is supposed to be kept at the latest version.

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
- For non-Linux: Is sandboxing enabled in `nix.conf`? (See [Nix manual](https://nixos.org/manual/nix/stable/command-ref/conf-file.html))
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:
  - [NixOS test(s)](https://nixos.org/manual/nixos/unstable/index.html#sec-nixos-tests) (look inside [nixos/tests](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests))
  - and/or [package tests](https://github.com/NixOS/nixpkgs/blob/master/pkgs/README.md#package-tests)
  - or, for functions and "core" functionality, tests in [lib/tests](https://github.com/NixOS/nixpkgs/blob/master/lib/tests) or [pkgs/test](https://github.com/NixOS/nixpkgs/blob/master/pkgs/test)
  - made sure NixOS tests are [linked](https://nixos.org/manual/nixpkgs/unstable/#ssec-nixos-tests-linking) to the relevant packages
- [ ] Tested compilation of all packages that depend on this change using `nix-shell -p nixpkgs-review --run "nixpkgs-review rev HEAD"`. Note: all changes have to be committed, also see [nixpkgs-review usage](https://github.com/Mic92/nixpkgs-review#usage)
- [ ] Tested basic functionality of all binary files (usually in `./result/bin/`)
- [24.05 Release Notes](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2405.section.md) (or backporting [23.05](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2305.section.md) and [23.11](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2311.section.md) Release notes)
  - [ ] (Package updates) Added a release notes entry if the change is major or breaking
  - [ ] (Module updates) Added a release notes entry if the change is significant
  - [ ] (Module addition) Added a release notes entry if adding a new NixOS module
[...]

PR #312981, opened 2024-05-19
llvmPackages: 17.0.6 -> 18.1.5 on Linux

alyssais:
## Description of changes

The next staging cycle will likely have Rust using LLVM 18 (https://github.com/NixOS/nixpkgs/pull/309580), so to avoid ending up with more LLVMs in closures than necessary, it's time to start working on upgrading the default in general.

I've built all my systems with LLVM 18 with no build regressions (which is better than [16 -> 17](https://github.com/NixOS/nixpkgs/pull/285333), where there was 1), so hopefully it'll just be a few leaf packages that need to have their LLVM versions pinned.  Or maybe we'll be lucky, and the pinning from last round will mean there aren't any?

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [x] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/top-level/all-packages.nix
+++ b/pkgs/top-level/all-packages.nix
@@ -16269,7 +16269,7 @@ with pkgs;
       /**/ if platform.isDarwin then 16
       else if platform.isFreeBSD then 16
       else if platform.isAndroid then 12
-      else if platform.isLinux then 17
+      else if platform.isLinux then 18
       else if platform.isWasm then 16
       # For unknown systems, assume the latest version is required.
       else 17;


pkgs/top-level/all-packages.nix
```diff
@@ -16269,7 +16269,7 @@ with pkgs;
       /**/ if platform.isDarwin then 16
       else if platform.isFreeBSD then 16
       else if platform.isAndroid then 12
-      else if platform.isLinux then 17
+      else if platform.isLinux then 18
       else if platform.isWasm then 16
       # For unknown systems, assume the latest version is required.
```
Smaug123, line 16274: (This comment is now more obviously out of sync than it was - maybe worth a drive-by relaxing the comment to "assume a recent version", or else raising an issue to track an upgrade? I'm not familiar with how we manage LLVM)

RossComputerGuy: Could you please elaborate? I'm not sure I understand what your review means.

Smaug123: The comment says "assume the latest version", and then uses version 17 which is not the latest version. This PR upgrades one of the cases to 18, which (being greater than 17) makes the mismatch between stated intent and actual fact much more obvious.

RossComputerGuy (commented):
I see we use `else 17`, is there a reason why we don't change it to 18 or are we only sticking to updating Linux to 18?

pkgs/top-level/all-packages.nix
RossComputerGuy: Yeah well, in the PR title, it states that this is only a Linux upgrade. I'm not sure why we don't specify `"latest"` instead. However the likely reason is because LLVM is a core package and one bad change that is untested could break a lot. I'm sure with how long LLVM 18 has been out, it is okay to change.

rrbutani: @Smaug123 #312617 bumps the LLVM version for other platforms to `18`.

@RossComputerGuy I believe we can't use `llvmPackages_latest` here because it's an alias:
https://github.com/NixOS/nixpkgs/blob/454d208c022ea7b3fe9007dafaf3a617f74fcd44/pkgs/top-level/aliases.nix#L1418

RossComputerGuy (approved):

RossComputerGuy:
Tried a `nixpkgs-review` (big mistake with all the `nix log` errors) and a lot of packages are borked on `aarch64-linux`.

```
900 packages marked as broken and skipped
23748 packages failed to build
```

RossComputerGuy:
I tried on the aarch64 builder and it appears that works better. Must be Apple Silicon related weirdness.

alyssais:
Builds fine for me on Apple Silicon, fwiw.

alyssais:
If you only tried building with nixpkgs-review, it was very probably the CPU or memory contention issues I mentioned before.

PR #312997, opened 2024-05-19
gmic-qt: build from Git source to avoid mutable tarball

amarshall:
The previous tarball src sometimes gets mutated (see e.g. [this comment][1]). This was changed from the Git src in fd3e2b49f4a3138a1c3e2dd854578f64b9c772cd (see also [upstream discussion][2]). However the delta seems simple; it had error:

```
In file included from /build/source/src/GmicProcessor.cpp:48:
/nix/store/jk1dp7v01pisw0flybqwyjg63in6r9fp-gmic-3.3.5-dev/include/gmic.h:191:21: fatal error: gmic.cpp: No such file or directory
  191 | #define cimg_plugin "gmic.cpp"
```

workaround this by linking `gmic.cpp` into the expected location as it appears in the tarball.

Remove the now-unneeded updateScript, as it is trivial GitHub tags now.

Unclear to me why cimg was not needed in buildInputs before, but it is now.

This may appear to be downgrading, but this is only because the previous drv used the gmic version, not the gmic-qt version.

[1]: https://github.com/NixOS/nixpkgs/pull/311734#issuecomment-2118663822
[2]: https://github.com/c-koi/gmic-qt/pull/175

## Things done

`nix-build -A gmic-qt.passthru.tests`; tested in GIMP as plugin.

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [x] Tested, as applicable:

---

### DIFF
--- a/pkgs/by-name/gm/gmic-qt/package.nix
+++ b/pkgs/by-name/gm/gmic-qt/package.nix
@@ -1,26 +1,22 @@
 { lib
 , cimg
 , cmake
-, coreutils
 , curl
-, fetchzip
+, fetchFromGitHub
 , fftw
 , gimp
 , gimpPlugins
 , gmic
-, gnugrep
-, gnused
 , graphicsmagick
 , libjpeg
 , libpng
 , libsForQt5
 , libtiff
 , ninja
-, nix-update
+, nix-update-script
 , openexr
 , pkg-config
 , stdenv
-, writeShellScript
 , zlib
 , variant ? "standalone"
 }:
@@ -53,14 +49,14 @@ assert lib.assertMsg
 
 stdenv.mkDerivation (finalAttrs: {
   pname = "gmic-qt${lib.optionalString (variant != "standalone") "-${variant}"}";
-  version = "3.3.6";
+  version = "3.3.5";
 
-  src = fetchzip {
-    url = "https://gmic.eu/files/source/gmic_${finalAttrs.version}.tar.gz";
-    hash = "sha256-LZwAMLvQ+X6xkvnL+7LA/UmwDBGgapUtFHNJuV04F+Y=";
+  src = fetchFromGitHub {
+    owner = "c-koi";
+    repo = "gmic-qt";
+    rev = "v.${finalAttrs.version}";
+    hash = "sha256-WApuIWqVgVJAM2WdfOiqoQ2U+9kIuq8fy6wvJ55KoIc=";
   };
-
-  sourceRoot = "${finalAttrs.src.name}/gmic-qt";
 
   nativeBuildInputs = [
     cmake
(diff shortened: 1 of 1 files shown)


pkgs/by-name/gm/gmic-qt/package.nix
```diff
@@ -107,24 +111,6 @@ stdenv.mkDerivation (finalAttrs: {
       gimp-plugin = gimpPlugins.gmic;
       inherit cimg gmic;
     };
-
-    updateScript = writeShellScript "gmic-qt-update-script" ''
```
lilyinstarlight, line 111: We should still have an update script because as-is nix-update and nixpkgs-update/r-ryantm will have a hard time finding updates both due to the lack of GitHub upstream releases (rather than just tags) and the fact that the tags have weird version strings. Maybe something like this?

```nix
    updateScript = nix-update-script {
      extraArgs = [ "--version-regex" "^v\\.(.*)" ];
    };
```

Also please remove the now-unused arguments from the top of the file

lilyinstarlight (commented):
> Unclear to me why cimg was not needed in buildInputs before, but it is now.

The tarball also includes `CImg.h` and now it no longer does, hence the need for the build input

I do still hate that the build system requires more than just the header and shared object for libgmic (although it _does_ also use both and still links to it...), but this looks like a reasonable enough hack (I hate having to use the bespoke tarballs anyway... only trouble for `gmic` package itself is the stdlib has a bootstrap problem so we still have to download that prebuilt header which could change too)

I left one comment but otherwise this looks good, thank you!

pkgs/by-name/gm/gmic-qt/package.nix
jopejoe1: Using gitUpdater would be nicer for this as it is specificly designd for git tags which this is using for versions.

```nix
    updateScript = gitUpdater { 
      rev-prefix = "v.";
    };
```

lilyinstarlight: Both that and nix-update-script are specifically designed for it I guess (nix-update just usually deals with edge cases better), but I'm largely indifferent as long as there is just *something* there :)

lilyinstarlight: Actually this is one of the edge cases that `gitUpdater` does not handle. If I add that `updateScript` then it tries to "update" it to version 2.2.0 rather than keeping the version at 3.3.5

So please use `nix-update-script` unless there is an argument for `gitUpdater` to fix that

lilyinstarlight (approved):
Looks good to me now!

Will merge when ofborg is green if there are no objections

PR #313079, opened 2024-05-20
python311Packages.freud: init at 3.0.0

doronbehar:
###### Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
- [ ] Tested, as applicable:

### DIFF
--- a/pkgs/development/python-modules/freud/default.nix
+++ b/pkgs/development/python-modules/freud/default.nix
@@ -1,0 +1,90 @@
+{ lib
+, stdenv
+, buildPythonPackage
+, fetchFromGitHub
+, cmake
+, cython
+, oldest-supported-numpy
+, scikit-build
+, setuptools
+, tbb
+, numpy
+, rowan
+, scipy
+, pytest
+, gsd
+, matplotlib
+, sympy
+}:
+
+buildPythonPackage rec {
+  pname = "freud";
+  version = "3.0.0";
+  pyproject = true;
+
+  src = fetchFromGitHub {
+    owner = "glotzerlab";
+    repo = "freud";
+    rev = "v${version}";
+    hash = "sha256-aKh2Gub1vU/wzvWkCl8yzlIswp8CtR975USiCr6ijUI=";
+    fetchSubmodules = true;
+  };
+  # Because we prefer to not `leaveDotGit`, we need to fool upstream into
+  # thinking we left the .git files in the submodules, so cmake won't think we
+  # didn't initialize them. Upstream doesn't support using the system wide
+  # installed version of these libraries, and it's probably aint's worth the
+  # hassle, because upstream also doesn't distribute all of these dependencies'
+  # libraries, and probably it uses only what it needs.
+  preConfigure = ''
+    touch extern/{voro++,fsph,Eigen}/.git
+  '';
+
+  nativeBuildInputs = [
+    cmake
+    cython
+    oldest-supported-numpy
+    scikit-build
+    setuptools
+  ];
+  dontUseCmakeConfigure = true;
+  buildInputs = [
+    tbb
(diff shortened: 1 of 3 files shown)


doronbehar:
@ofborg build python312Packages.freud

pkgs/development/python-modules/rowan/default.nix
```diff
@@ -0,0 +1,45 @@
...
+  src = fetchFromGitHub {
+    owner = "glotzerlab";
+    repo = "rowan";
+    rev = "v${version}";
+    hash = "sha256-klIqyX04w1xYmYtAbLF5jwpcJ83oKOaENboxyCL70EY=";
+  };
+
+  nativeBuildInputs = [
+    setuptools
+    wheel
```
SuperSandro2000, line 25: Why do we need wheel here?

pkgs/development/python-modules/freud/default.nix
```diff
@@ -0,0 +1,63 @@
...
+  # thinking we left the .git files in the submodules, so cmake won't think we
+  # didn't initialize them.
+  preConfigure = builtins.map (submodule: ''
+    touch extern/${submodule}/.git
+  '') [ "fsph" "Eigen" "voro++" ];
+
+  nativeBuildInputs = [
+    cmake
+    cython
+    oldest-supported-numpy
```
SuperSandro2000, line 38: We should probably patch that tp just numpy

pkgs/development/python-modules/freud/default.nix
```diff
@@ -0,0 +1,63 @@
...
+    scikit-build
+    setuptools
+    wheel
```
SuperSandro2000, line 41: Same question here

pkgs/development/python-modules/freud/default.nix
```diff
@@ -0,0 +1,63 @@
...
+    rev = "v${version}";
+    hash = "sha256-aKh2Gub1vU/wzvWkCl8yzlIswp8CtR975USiCr6ijUI=";
+    fetchSubmodules = true;
+  };
+  # Because we prefer to not `leaveDotGit`, we need to fool upstream into
+  # thinking we left the .git files in the submodules, so cmake won't think we
+  # didn't initialize them.
+  preConfigure = builtins.map (submodule: ''
+    touch extern/${submodule}/.git
+  '') [ "fsph" "Eigen" "voro++" ];
```
SuperSandro2000, line 33: Why not use our Eigen?

pkgs/development/python-modules/freud/default.nix
doronbehar: > We should probably patch that tp just numpy

There is a difference between the two, and many packages require `oldest-supported-numpy`. TBH I don't understand the idea behind this meta package. Note that even in Nixpkgs it is defined to simply propagate numpy so I think it's OK.

pkgs/development/python-modules/freud/default.nix
doronbehar: > Why not use our Eigen?

Indeed it would have been nice, I haven't thought about it. However, upstream doesn't seem to support that out of the box via a cmake option or a like. Also I'm not sure it will save us space, because upstream doesn't distribute eventually all of eigen's libraries. I added a sentence to the comment explaining the reasoning.

doronbehar:
I will wait for CI on all platforms.

pkgs/development/python-modules/freud/default.nix
SuperSandro2000: It is just a meta package which requires the last numpy upstream support IIRC

PR #313144, opened 2024-05-20
lib.strings: deprecate lib.strings.{head,tail}

FireyFly:
These operate on lists and error if you try to call them on strings... and they're already exposed under `lib` and `lib.lists`

## Description of changes

* `lib.strings.head` is no longer an alias for `builtins.head`, ditto for `tail`

## Things done

I checked (somewhat sloppily) for references with `rg lib.strings.head`, `rg '\(lib.strings\) .*head`, `rg lib.strings.tail`, `rg \(lib.strings\) .*tail` and didn't notice any. I checked that `<nixpkgs>` and `<nixpkgs/nixos>` evaluate (on my system/with my config) as well

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/lib/strings.nix
+++ b/lib/strings.nix
@@ -21,7 +21,6 @@ rec {
     filter
     fromJSON
     genList
-    head
     isInt
     isList
     isAttrs
@@ -35,11 +34,21 @@ rec {
     storeDir
     stringLength
     substring
-    tail
     toJSON
     typeOf
     unsafeDiscardStringContext
     ;
+
+  # TODO: warnings added 2025-04-10; remove after NixOS 2025.11
+  head = lib.warn ''
+    lib.strings.head only supports lists; use lib.lists.head instead.
+    The lib.strings.head alias is deprecated and will be removed after the NixOS 2025.11 release.
+  '' lib.head;
+
+  tail = lib.warn ''
+    lib.strings.tail only supports lists; use lib.lists.tail instead.
+    The lib.strings.tail alias is deprecated and will be removed after the NixOS 2025.11 release.
+  '' lib.tail;
 
   /**
     Concatenate a list of strings.
@@ -157,7 +166,7 @@ rec {
     if list == [ ] || length list == 1 then
       list
     else
-      tail (
+      lib.tail (
         lib.concatMap (x: [
           separator
           x
@@ -465,7 +474,7 @@ rec {
       # then the regex may not match and `res` will be `null`.
       res = match regex s;
     in
-    optionalString (res != null) (head res);
+    optionalString (res != null) (lib.head res);
 
   /**
     Construct a Unix-style, colon-separated search path consisting of
@@ -1863,7 +1872,7 @@ rec {
     let
       components = splitString "/" url;
       filename = lib.last components;
-      name = head (splitString sep filename);
+      name = lib.head (splitString sep filename);
     in
     assert name != filename;
     name;
@@ -2532,10 +2541,10 @@ rec {
       strippedInput = matchStripInput str;
 
       # RegEx: Match a leading '0' then one or more digits.
-      isLeadingZero = matchLeadingZero (head strippedInput) == [ ];
+      isLeadingZero = matchLeadingZero (lib.head strippedInput) == [ ];
 
       # Attempt to parse input
-      parsedInput = fromJSON (head strippedInput);
+      parsedInput = fromJSON (lib.head strippedInput);
 
       generalError = "toInt: Could not convert ${escapeNixString str} to int.";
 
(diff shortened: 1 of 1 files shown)


eclairevoyant:
We should probably `throw` an actual error message to explain why it was removed

FireyFly:
Yep, that's very fair, you're probably right... I think after a period it'd be nice to fully remove them though since they still show up in e.g. tabcompletion in the REPL otherwise..

I figured in this case it might make sense to be a bit more bold than usual for deprecation since well... the existing functions don't even work on strings (so chances are pretty slim anyone depends on importing these from `lib.strings` and using them on lists, honestly)

eclairevoyant:
> chances are pretty slim anyone depends on importing these from `lib.strings` and using them on lists, honestly

You'd be surprised what people do, especially accidentally, since nix lacks robust static analysis

lib/strings.nix
```diff
@@ -1338,4 +1336,15 @@ rec {
...
+  let
+    head_tail_warn_msg = name:
+      ''
+        lib.strings.${name} only supports lists; use lib.lists.${name} instead.
+        The lib.strings.${name} alias is deprecated and will be removed after the NixOS 2024.11 release.
+      '';
+  in {
+    # warnings added 2024-05-21
+    head = x: lib.warn (head_tail_warn_msg "head") (builtins.head x);
+    tail = x: lib.warn (head_tail_warn_msg "tail") (builtins.tail x);
```
infinisil, line 1349:
```suggestion
    head = lib.warn (head_tail_warn_msg "head") builtins.head;
    tail = lib.warn (head_tail_warn_msg "tail") builtins.tail;
```

This way the error gets thrown only once (generally how deprecations are done in `lib`)

lib/strings.nix
```diff
@@ -1338,4 +1336,15 @@ rec {
...
         else levenshtein ainfix binfix <= k;
       in f;
-}
+} // (
+  let
+    head_tail_warn_msg = name:
+      ''
+        lib.strings.${name} only supports lists; use lib.lists.${name} instead.
+        The lib.strings.${name} alias is deprecated and will be removed after the NixOS 2024.11 release.
+      '';
```
infinisil, line 1345: Imo it's a bit overkill to abstract such an error message, and as implemented this actually slows down `lib` evaluation a little bit, because of the extra `//` used. I think it would be fine to just inline the error twice

infinisil (commented):
Just some relatively minor comments, but otherwise this looks good!

lib/strings.nix
FireyFly: Ah, I pulled it out because the main attrset is `rec` and already uses `head`/`tail` a few times but we can change the couple uses to explicit `builtins.{head,tail}` in the module

FireyFly: I moved it into the main `rec {}` attrset with exports and changed occurrences of `head`/`tail` in the file to use `builtins.{head,tail}` explicitly

lib/strings.nix
```diff
@@ -85,7 +94,7 @@ rec {
     list:
     if list == [] || length list == 1
     then list
-    else tail (lib.concatMap (x: [separator x]) list);
+    else builtins.tail (lib.concatMap (x: [separator x]) list);
```
infinisil, line 97: Instead of doing this, you can add
```nix
  inherit (lib) head tail;
```

to the top `let in`

(the preference should be to get symbols from `lib`, because it's more flexible)

infinisil (approved):
Just a minor comment, looking good after that!

lib/strings.nix
FireyFly: The problem with that is the top-level `let-in` is above the `rec { }` attrset where we define the deprecated `head` and `tail` with warnings... so any unqualified references will refer to those and throw warnings.

(That's why at first I had pulled those references out into a separate attrset that was merged in with `//` before...)

I can access them from `lib` instead of `builtins` though if that's preferred

FireyFly: Moved to using `lib.{head,tail}` instead of `builtins.{head,tail}` as requested. As mentioned I don't see how a `let-inherit` would help here since the `head`/`tail` with warnings in the returned recursive attrset would shadow them. Please let me know if I'm missing something, or if there's a different approach you would prefer...

FireyFly: It'd still be nice to get rid of these confusing aliases since they don't work as expected (for their name/location). Could you let me know if there's anything else left to do here, @infinisil?

PR #313173, opened 2024-05-20
erlang_27: init 27.0; erlang_27-rc3: remove

chiroptical:
## Description of changes

Erlang 27 was officially released https://erlangforums.com/t/erlang-otp-27-0-released/3636

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [x] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [x] `sandbox = true`
- [ ] Tested, as applicable:

```
> nix-build -A erlang_27
/nix/store/70y7d02jfqh6nfpaydd6dkwbnpcm9mfx-erlang-27.0
> ./result/bin/erl
Erlang/OTP 27 [erts-15.0] [source] [64-bit] [smp:8:8] [ds:8:8:10] [async-threads:1] [jit]

Eshell V15.0 (press Ctrl+G to abort, type help(). for help)
1> io:format("hello world~n").
hello world
ok
2> q().
ok
```

```
[nix-shell:~/.cache/nixpkgs-review/rev-a251d7e27849119564bb72e3ab157128f4200add]$ erl
Erlang/OTP 27 [erts-15.0] [source] [64-bit] [smp:8:8] [ds:8:8:10] [async-threads:1] [jit]

Eshell V15.0 (press Ctrl+G to abort, type help(). for help)
1> io:format(~"hello world~n").
hello world
ok
```

The deprecation,

```
> nix-build -A erlang_27-rc3
error:

         at /Users/chiroptical/programming/nix/nixpkgs/pkgs/top-level/aliases.nix:33:5:

           32|   removeDistribute = alias: with lib;
           33|     if isDerivation alias then
             |     ^
           34|       dontDistribute alias

         at /Users/chiroptical/programming/nix/nixpkgs/pkgs/top-level/aliases.nix:26:5:

           25|   removeRecurseForDerivations = alias: with lib;
           26|     if alias.recurseForDerivations or false
             |     ^
           27|     then removeAttrs alias [ "recurseForDerivations" ]

       error: erlang_27-rc3 has been removed in favor of erlang_27
```

---

### DIFF
--- a/pkgs/top-level/all-packages.nix
+++ b/pkgs/top-level/all-packages.nix
@@ -17180,7 +17180,7 @@ with pkgs;
   };
 
   inherit (beam.interpreters)
-    erlang erlang_27-rc3 erlang_26 erlang_25 erlang_24
+    erlang erlang_27-rc3 erlang_27 erlang_26 erlang_25 erlang_24
     erlang_odbc erlang_javac erlang_odbc_javac
     elixir elixir_1_16 elixir_1_15 elixir_1_14 elixir_1_13 elixir_1_12 elixir_1_11 elixir_1_10
     elixir-ls;
--- a/pkgs/top-level/all-packages.nix
+++ b/pkgs/top-level/all-packages.nix
@@ -17180,7 +17180,7 @@ with pkgs;
   };
 
   inherit (beam.interpreters)
-    erlang erlang_27-rc3 erlang_27 erlang_26 erlang_25 erlang_24
+    erlang erlang_27 erlang_26 erlang_25 erlang_24
     erlang_odbc erlang_javac erlang_odbc_javac
     elixir elixir_1_16 elixir_1_15 elixir_1_14 elixir_1_13 elixir_1_12 elixir_1_11 elixir_1_10
     elixir-ls;
--- a/pkgs/top-level/beam-packages.nix
+++ b/pkgs/top-level/beam-packages.nix
@@ -41,6 +41,15 @@ in
 
     # Standard Erlang versions, using the generic builder.
 
+    erlang_27 = self.beamLib.callErlang ../development/interpreters/erlang/27.nix {
+      wxGTK = wxGTK32;
+      parallelBuild = true;
+      autoconf = buildPackages.autoconf269;
+      exdocSupport = true;
+      exdoc = self.packages.erlang_26.ex_doc;
+      inherit wxSupport systemdSupport;
+    };
+
     erlang_27-rc3 = self.beamLib.callErlang ../development/interpreters/erlang/27-rc3.nix {
       wxGTK = wxGTK32;
       parallelBuild = true;
--- a/pkgs/top-level/beam-packages.nix
+++ b/pkgs/top-level/beam-packages.nix
@@ -50,15 +50,6 @@ in
       inherit wxSupport systemdSupport;
     };
 
-    erlang_27-rc3 = self.beamLib.callErlang ../development/interpreters/erlang/27-rc3.nix {
-      wxGTK = wxGTK32;
-      parallelBuild = true;
-      autoconf = buildPackages.autoconf269;
-      exdocSupport = true;
-      exdoc = self.packages.erlang_26.ex_doc;
-      inherit wxSupport systemdSupport;
-    };
-
     erlang_26 = self.beamLib.callErlang ../development/interpreters/erlang/26.nix {
       wxGTK = wxGTK32;
       parallelBuild = true;
--- a/pkgs/development/interpreters/erlang/27-rc3.nix
+++ b/pkgs/development/interpreters/erlang/27-rc3.nix
@@ -1,6 +1,0 @@
-{ mkDerivation }:
-
-mkDerivation {
-  version = "27.0-rc3";
-  sha256 = "sha256-c2DTIqBd7UxpSv84F1cCB9K+MkJb5OwnWSghtewnw/4=";
-}
--- a/pkgs/development/interpreters/erlang/27.nix
+++ b/pkgs/development/interpreters/erlang/27.nix
@@ -1,0 +1,6 @@
+{ mkDerivation }:
+
+mkDerivation {
+  version = "27.0";
+  sha256 = "sha256-YZWBLcpkm8B4sjoQO7I9ywXcmxXL+Dvq/JYsLsr7TO0=";
+}
--- a/pkgs/top-level/aliases.nix
+++ b/pkgs/top-level/aliases.nix
@@ -341,6 +341,7 @@ mapAliases ({
   erlangR22 = erlang_22;
   erlang_23 = throw "erlangR23 has been removed in favor of newer versions."; # added 2023-09-11
   erlangR23 = erlang_23;
+  erlang_27-rc3 = throw "erlang_27-rc3 has been removed in favor of erlang_27"; # added 2024-05-20
   etcd_3_3 = throw "etcd_3_3 has been removed because upstream no longer maintains it"; # Added 2023-09-29
   etcher = throw "'etcher' has been removed because it depended on an insecure version of Electron"; # Added 2024-03-14
   eterm = throw "eterm was removed because it is still insecure: https://github.com/mej/Eterm/issues/7"; # Added 2023-09-10


pkgs/top-level/all-packages.nix
```diff
@@ -17180,7 +17180,7 @@ with pkgs;
   };
 
   inherit (beam.interpreters)
-    erlang erlang_27-rc3 erlang_26 erlang_25 erlang_24
+    erlang erlang_27 erlang_26 erlang_25 erlang_24
```
superherointj, line 17183: If you remove versions like that you'll be breaking `erlang_27-rc3` without notice for whoever is using this version.

Split the different objectives:
1) Add new erlang version (`erlang_27: init 27.x.x`)
2) Upgrade main erlang (`erlang: erlang_27-rc3 -> erlang_27`)
3) Remove erlang_27-rc3 (`erlang_27-rc3: remove`) => It should be an alias, to be turned into a warning on release and then to be removed after release. Minimally it should be a warning, such that the user gets some warning of what is going on.

chiroptical: I have never turned these things into warnings. Do you have an example?

FWIW, the `rc{N}` versions were only added for testing purposes. I agree a warning would be ideal though.

superherointj: https://github.com/NixOS/nixpkgs/blob/master/pkgs/top-level/aliases.nix#L338

superherointj: In retrospect, likely `erlang_27-rc3` should have been tagged to `erlang_latest` or `erlang_edge` (for pre-releases).

chiroptical: Yeah, that is a good idea. Do you think it would be sufficient to just tag the warning for `erlang_27-rc3` and initialize `erlang_27` proper here? (again, this was my fault for not setting this up for success)

chiroptical: Also, I'll fix the commit message tonight

superherointj: To remove the package, add an alias:

> erlang_27-rc3 = throw "erlang_27-rc3 has been removed in favor of newer versions."; # added 2024-05-20

To: https://github.com/NixOS/nixpkgs/blob/master/pkgs/top-level/aliases.nix#L343

You need 3 commits: One for:
* adding the new erlang version.
* bumping the erlang tag
* removing the old version. (Which then creates the alias warning that such package is now unavailable.) I think this package should be removed because it is a release candidate, and it is not for an edge tag.

chiroptical: Ahh, I thought you wanted me to split these into 3 PRs which felt a little tedious. Thanks for the clarification.

I'll get this cleaned up tonight.

chiroptical: To be honest, the commit messages for the latter two I am unsure of the convention

- `erlang: init at 27.0`: seems correct
- `erlang: erlang_27-rc3 -> erlang_27`: seems okay, deprecate one in favor of another
- `erlang: erlang_25 -> erlang_27 (latest version)`: totally unsure, `erlang: latest version to 27.0`?

superherointj: > The following commit message doesn't follow nixpkgs guidelines on commit messages:
> 
> > erlang_27-rc3 -> erlang_27
> 
> When removing a package tag (erlang_27-rc3), add a warning (at aliases) for notifying users of such deprecation.
> 
> Split the different objectives:
> 
>     1. Add new erlang version (`erlang_27: init 27.x.x`)
> 
>     2. Upgrade main erlang (`erlang: erlang_27-rc3 -> erlang_27`)
> 
>     3. Remove erlang_27-rc3 (`erlang_27-rc3: remove`) => It should be an alias, to be turned into a warning on release and then to be removed after release. Minimally it should be a warning, such that the user gets some warning instead of a not found error.
> 
> 
> *Notice: The text inside parameters is the expected commit messages title.

Please, re-read this. ^

chiroptical: I could also these as being acceptable?

```
erlang_27: init at 27.0
erlang: erlang_27-rc3 -> erlang_27
erlang: latestVersion -> erlang_27
```

superherointj: Please, make the commit removing version `erlang_27-rc3` a discrete commit.

It should be titled as:

> `erlang_27-rc3: remove`

Then, certify only changes related to the removal of package exist in this commit. Such as package remove and adding alias deprecation warning.

chiroptical: Okay, I see my confusion now. In my brain, `erlang: erlang_27-rc3 -> erlang_27` is saying "replace erlang_27-rc3 with erlang_27" which implies I am adding the alias for `erlang_27-rc3` in the same commit (because it was replaced). You want it in a separate one (which is why I was thinking you meant `latestVersion`). Sorry, this is genuinely confusing to me.

chiroptical: I guess we should also update the "latest version" though. How is that supposed to look?

superherointj: > Okay, I see my confusion now. (...) Sorry, this is genuinely confusing to me.

If you have patience, we'll figure out this. Communication is hard. The good side is, once you learn the pattern, reviews will be easier.

cole-h:
@ofborg eval

chiroptical:
@superherointj I think I got it now. Thanks for sticking with me. I learned a lot about tag management. I also learned it would have been a lot simpler if I did `erlang_27: init 27-rc3` instead of what I did.

pkgs/top-level/beam-packages.nix
```diff
@@ -50,14 +50,7 @@ in
...
 
-    erlang_27-rc3 = self.beamLib.callErlang ../development/interpreters/erlang/27-rc3.nix {
-      wxGTK = wxGTK32;
-      parallelBuild = true;
-      autoconf = buildPackages.autoconf269;
-      exdocSupport = true;
-      exdoc = self.packages.erlang_26.ex_doc;
-      inherit wxSupport systemdSupport;
-    };
+    erlang_27-rc3 = self.interpreters.erlang_27;
```
superherointj: Commit: `erlang: erlang_27-rc3 -> erlang_27 `

This change does not make sense.

This commit should be bumping erlang tag. And not redefining what `erlang_27-rc3` package is.

chiroptical: I legit have no idea what you want here. I don't want to touch the `erlang` tag. At the end of this, I only want `erlang_27` to be initialized. This is why I am mega confused. I want to **only**

- add `erlang_27`
- remove `erlang_27-rc3`

Updating the erlang tag requires me to add all of these

```
      erlang = self.interpreters.${self.latestVersion};
      erlang_odbc = self.interpreters."${self.latestVersion}_odbc";
      erlang_javac = self.interpreters."${self.latestVersion}_javac";
      erlang_odbc_javac = self.interpreters."${self.latestVersion}_odbc_javac";
```

which I also **don't** want to do because I don't understand their existence.

chiroptical: I understand _what they are_ but I don't understand _why they are here_

superherointj: Then, remove commit:

> erlang: erlang_27-rc3 -> erlang_27

superherointj: To bump erlang tag, you would have to only update `latestVersion` version. But yes, you don't have to do that.
But again, I only assumed that because you actually said without wanting to say that.
PR name has to be changed now, then.

chiroptical: Done, if you are in the `#nix` channel of the Erlanger slack would love to discuss ways we could simplify the beam tags. Looks like you did the replacement of `erlangR` -> `erlang_` but we never deprecated the old tags. Also looks like there are some dangling aliases for old versions too.

chiroptical: I also started a 🧵 there on updating the latest version to gather feedback.

chiroptical: Sorry for being a pain in your butt 😆

superherointj: > if you are in the `#nix` channel of the Erlanger slack 

Sorry. I don't use slack and I don't know this channel. I stopped with Erlang/Elixir several years ago. I still appreciate the technology. But I'm no longer active in it.

> would love to discuss ways we could simplify the erlang tags.

What do you mean?

> Looks like you did the replacement of `erlangR` -> `erlang_` but we never deprecated the old tags. 

Feel free to do it. It's been a while.

> Also looks like there are some dangling aliases for old versions too.

Erlang & Elixir are lacking upkeep. There are few people doing upkeep. The work you are doing here is important. Because you took an interest in it. I remember finding several problems in how things are structure for Elixir. Like, overrides not working, or some other issue that I don't recall correctly. It is possible to use a better Nix structure, but I suggest you to start slowly with basic things before diving deeper.

I remember having thought about a way to speed up builds with cache for dependencies. But it never materialized. Nix and Elixir is a good match. I wish I had found Nix earlier when I was still using Erlang/Elixir.

Communication today was messy. But I think, you came out of it with a better understanding.

chiroptical: > Sorry. I don't use slack and I don't know this channel. I stopped with Erlang/Elixir several years ago. I still appreciate the technology. But I'm no longer active in it.

No worries.

> would love to discuss ways we could simplify the erlang tags.
> > What do you mean?

For example, are the `_odbc`, `_javac`, and `_odbc_javac` widely used? They aren't exposed in `all-packages.nix` so are they cached? AFAICT, they aren't very discoverable and _could_ be nice to trim.

> Looks like you did the replacement of erlangR -> erlang_ but we never deprecated the old tags.
Feel free to do it. It's been a while.

Awesome, I'll add it to my list.

> I remember finding several problems in how things are structure for Elixir. Like, overrides not working, or some other issue that I don't recall correctly. It is possible to use a better Nix structure, but I suggest you to start slowly with basic things before diving deeper.

Thanks for the extra context!

> Communication today was messy. But I think, you came out of it with a better understanding.

Yeah, partially my fault. I should have explained better in the description **what** I wanted to accomplish. I most certainly did learn a bunch.

I am also glad we **didn't** end up doing the `_odbc_javac` versions because we can see who is relying on them.

superherointj:
Result of `nixpkgs-review pr 313173` run on x86_64-linux [1](https://github.com/Mic92/nixpkgs-review)
<details>
  <ul>
    <li>erlang_27</li>
  </ul>
</details>

superherointj (approved):

camelpunch:
Also looks like we didn't make an `interpreters.erlangRxx` alias this time?

superherointj:
> Also looks like we didn't make an `interpreters.erlangRxx` alias this time?
 
That is deprecated. And to be removed.

chiroptical:
> It looks like erlang 27 depends on erlang 26 inside beam_packages.nix:
> 
> 
> 
> https://github.com/NixOS/nixpkgs/blob/86e62b8133807f3d12836348c6e29e8e4169bcbd/pkgs/top-level/beam-packages.nix#L49
> 
> 
> 
> Is this intentional?

It is actually. Ex_doc is built using Erlang 26 (because we don't yet have Erlang 27) and then used as an input to Erlang 27 to build the documentation. 

Is it ideal? Maybe not, but it was an intentional decision. We could instead just get the ex_doc binary instead of building it. (FWIW, this could also be a misunderstanding on how Elixir packages are built). Happy to fix if you have suggestions!

chiroptical:
As for the other comments, I am planning to put together a post on Erlang forums to gather some nix related feedback. But I think removing the Rx versions is less controversial than removing the `javac ` versions. Unsure though.

superherointj:
> As for the other comments, I am planning to put together a post on Erlang forums to gather some nix related feedback. But I think removing the Rx versions is less controversial than removing the `javac ` versions. Unsure though.

No version was removed. Using `_` for separating package name and version is standard in nixpkgs. The use of `R` was non-standard.

adamcstephens:
> We could instead just get the ex_doc binary instead of building it. (FWIW, this could also be a misunderstanding on how Elixir packages are built).

I would vote for fetching their binary script. Using a different erlang version to get ex_doc means using erlang 27 requires fetching and storing two different erlangs. It's even more painful when using `beam_minimal` with a custom erlang version, as you have to build both 26 and 27.

The otp docs say you can use thse: <https://github.com/elixir-lang/ex_doc/releases/tag/v0.34.1>

If we don't want to do that, I'd advocate for disabling doc generation by default or for the smaller beam package sets.

camelpunch:
Exactly. And if you're using beam_minimal it's not like you want those docs anyway. I've previously had lots of fun trying to exclude them from builds.

superherointj:
Docs should be a different package output.

adamcstephens:
> Docs should be a different package output.

All outputs are built together, so this won’t save any build time.

superherointj:
> All outputs are built together, so this won’t save any build time.

Then, what is missing is the means of globally disabling the build of manuals?
I understand the need particularly for containers.

adamcstephens:
I've opened https://github.com/NixOS/nixpkgs/pull/313173

chiroptical:
> If we don't want to do that, I'd advocate for disabling doc generation by default or for the smaller beam package sets.

Seems fine to me. We can easily add an attribute to override at minimum.

As for ex_doc, the current nix package comes with the run time. This makes sense. If we just used the binary people would also have to pull in a run time to use it. However, I am not opposed to only bringing in the executable in the generic erlang builder because we are building the run time before needing ex_doc. However, then you are maintaining two ex_docs. Just my two cents.

https://github.com/NixOS/nixpkgs/pull/321367 <- this one gives us the new ex_doc for OTP 27.1 when it releases. Just let me know what you want to do with it.

chiroptical:
> I've opened #313173

This link leads back to this PR. Might need to update.

PR #313248, opened 2024-05-20
add force-build-new-FODs.sh maintainer script

risicle:
## Description of changes
See the later discussions in https://github.com/NixOS/nix/issues/969, https://github.com/NixOS/ofborg/issues/68, https://github.com/NixOS/rfcs/pull/171 around concerns that the nixos binary cache could be "seeded" with a malicious source package in a quiet under-reviewed corner of nixpkgs which then gets re-used in a more security-critical place through a hash confusion attack.

The following script can be used as a mitigation to this, enabling reviewers to check whether new or changed Fixed-Output-Derivations actually produce the purported output hash when not pulled from a cache. It actually works by producing a "tainted" copy of any new FODs with a random hash and checking what "`got:`" hash nix lists when the hashes fail to agree.

There are obvious synergies here with what ofborg does - particularly the most computationally intensive part, performing the full nixpkgs evaluation. However it appears that the ofborg team don't have the capacity to act on this now, so it's probably better to get a standalone copy of this script in more peoples' hands.

The script in fact uses a modified version of ofborg's `outpaths.nix`, along with a few other embedded sibling scripts. There are of course downsides with using embedded scripts, but I chose to go this way because it keeps the script very self-contained and means I don't have to solve problems like locating these other scripts at runtime.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/maintainers/scripts/force-build-new-FODs.sh
+++ b/maintainers/scripts/force-build-new-FODs.sh
@@ -1,0 +1,285 @@
+#!/usr/bin/env nix-shell
+#! nix-shell -i bash -p coreutils gnugrep gnused nix curl jq
+#
+# Given directories <before-nixpkgs> and <after-nixpkgs>, will find
+# new Fixed-Output-Derivations ("FODs") introduced by <after-nixpkgs>,
+# and attempt to build/download them locally, disregarding any caches
+# that claim to have a substitutable binary with the given hash, and
+# compares the actual hash with the purported output hash.
+#
+# On completion, outputs to stdout a series of json objects, each
+# representing a FOD that has been checked, containing at least
+# an `origDrv` and `status` key.
+#
+# Exit code is nonzero if at least one FOD failed the check.
+#
+# Also recognizes $REALIZE_PROCESSES and $REALIZE_TIMEOUT environment
+# vars
+#
+
+set -e -o pipefail
+
+if [ $# != '2' ] ; then
+  echo "usage: $0 <before-nixpkgs> <after-nixpkgs>" >&2
+  exit 2
+fi
+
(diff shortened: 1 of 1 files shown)


tobiasBora:
Thanks, I will try to give it a try when I've some time. Would be great to integrate it in nixpkgs review if ofborg is not going to run it automatically. But just wanted to drop a line saying that this is a nice to have and "easy" (compared to fixing nix directly) to implement mitigation, but a true fix would be to do it directly at nix's level as discussed in https://github.com/NixOS/nix/issues/969#issuecomment-2083382876 

For instance, I guess that this script would be unable to detect this kind of attack targeting a different architecture than the one of the current user?

jonringer:
How long does it take to run this? usually doing a nixpkgs-wide eval is pretty ram and time intensive for me.

risicle:
:grin: 

It takes just about as long as you suspect, which is why I was interested in piggy-backing on ofborg - it already does the required evals.

There's actually no limit to how long it can take when e.g. a common fetcher's hash changes - it can technically want to build quite a few non-FOD derivations in order to be able to build the FOD derivations. But ofborg itself isn't immune to this sort of thing and usually people ignore crazy ofborg results for staging PRs and the like.

It occurred to me the other day that this script would still have some value if it were simply run periodically against master. Might not catch things pre-merge, but it's better than the current situation.

qweered (approved):

PR #313345, opened 2024-05-21
check-by-name: remove concurrency group

katexochen:
Due to the trigger on PR edit, every change to the PR body will trigger a re-run of the check-by-name workflow. This is needed to catch of the base branch. However, an action like ticking the checkbox of a PR will also trigger a re-run. Furthermore, due to the concurrency group, any already running check-by-name workflow will be canceled, leading to a notification sent to the PR author.

To reduce the noise, remove the concurrency group. As a consequence, multiple runs will run to completion on the same PR/commit. While concurrent runs are often canceled after a few seconds, the workflow will take about a minute to complete.

image

## Things done

---

### DIFF
--- a/.github/workflows/check-by-name.yml
+++ b/.github/workflows/check-by-name.yml
@@ -14,16 +14,16 @@ on:
     # While `edited` is also triggered when the PR title/body is changed,
     # this PR action is fairly quick, and PR's don't get edited that often,
     # so it shouldn't be a problem
+    # There is a feature request for adding a `base_changed` event:
+    # https://github.com/orgs/community/discussions/35058
     types: [opened, synchronize, reopened, edited]
 
 permissions: {}
 
-# Create a check-by-name concurrency group based on the pull request number. if
-# an event triggers a run on the same PR while a previous run is still in
-# progress, the previous run will be canceled and the new one will start.
-concurrency:
-  group: check-by-name-${{ github.event.pull_request.number }}
-  cancel-in-progress: true
+# We don't use a concurrency group here, because the action is triggered quite often (due to the PR edit
+# trigger), and contributers would get notified on any canceled run.
+# There is a feature request for supressing notifications on concurrency-canceled runs:
+# https://github.com/orgs/community/discussions/13015
 
 jobs:
   check:


willbush:
I didn't think about notifications from canceled runs when adding the concurrency group.

I'm wondering if we can do some api inspection on `edited` events and early return if base branch is not changed.

gpt-4o came up with this:

```yaml
jobs:
  check:
    runs-on: ubuntu-latest
    timeout-minutes: 10

    steps:
      - name: Checkout code
        uses: actions/checkout@v2

      - name: Check event name and base branch change
        id: check_event
        run: |
          if [[ "${{ github.event_name }}" == "edited" ]]; then
            # Fetch the past event data to compare the base branch
            pr_number=${{ github.event.pull_request.number }}
            repo=${{ github.repository }}

            # Use GitHub CLI to fetch pull request events
            pr_events=$(gh api -H "Accept: application/vnd.github+json" /repos/${repo}/pulls/${pr_number}/events)
            last_event=$(echo "$pr_events" | jq -r '. | map(select(.event == "edited")) | last')

            # Compare the base branch in the current and previous event
            base_before=$(echo "$last_event" | jq -r '.base.ref')
            base_after=${{ github.event.pull_request.base.ref }}

            if [[ "$base_before" == "$base_after" ]]; then
              echo "The base branch did not change. Exiting."
              exit 0
            else
              echo "The base branch changed from $base_before to $base_after. Continuing."
            fi
          fi
```

I'm ok with removing the concurrency group for now. I may look into the above more later.

willbush (approved):

infinisil:
See also:
- Feature request to not get notified for concurrency-canceled runs: https://github.com/orgs/community/discussions/13015
- Feature request for a `base_changed` event: https://github.com/orgs/community/discussions/35058

.github/workflows/check-by-name.yml
```diff
@@ -18,13 +18,6 @@ on:
 
 permissions: {}
 
-# Create a check-by-name concurrency group based on the pull request number. if
-# an event triggers a run on the same PR while a previous run is still in
-# progress, the previous run will be canceled and the new one will start.
```
infinisil, line 23: Can you replace this with a comment saying why we don't want a concurrency group? Otherwise somebody might just want to add this again in the future :)

Ideally also link to the github feature requests

katexochen:
> I'm wondering if we can do some api inspection on `edited` events and early return if base branch is not changed.

@willbush While this is possible for sure, API requests are a pretty limited resource. Not sure what nixpkgs' quota is, but in other projects I run into rate-limiting quite often.

willbush:
>API requests are a pretty limited resource

That's a good point. I've hit them too using `gh api`.

infinisil (approved):

PR #313358, opened 2024-05-21
python3Packages.libnbd: init at 1.18.2

physics-enthusiast:
## Description of changes

Enable compilation of python bindings in the main libnbd derivation, and expose them in python3Packages.
## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [x] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/development/libraries/libnbd/default.nix
+++ b/pkgs/development/libraries/libnbd/default.nix
@@ -4,6 +4,7 @@
 , bash-completion
 , pkg-config
 , perl
+, buildPythonBindings ? false, python3
 , libxml2
 , fuse
 , fuse3
@@ -23,7 +24,8 @@ stdenv.mkDerivation rec {
     bash-completion
     pkg-config
     perl
-  ];
+  ]
+    ++ lib.optionals buildPythonBindings [ python3 ];
 
   buildInputs = [
     fuse
@@ -32,7 +34,15 @@ stdenv.mkDerivation rec {
     libxml2
   ];
 
+  configureFlags = lib.optionals buildPythonBindings [ "--with-python-installdir=${placeholder "out"}/${python3.sitePackages}" ];
+
   installFlags = [ "bashcompdir=$(out)/share/bash-completion/completions" ];
+
+  postInstall = lib.optionalString buildPythonBindings ''
+    LIBNBD_PYTHON_METADATA='${placeholder "out"}/${python3.sitePackages}/nbd-${version}.dist-info/METADATA'
+    install -Dm644 -T ${./libnbd-metadata} $LIBNBD_PYTHON_METADATA
+    substituteAllInPlace $LIBNBD_PYTHON_METADATA
+  '';
 
   meta = with lib; {
     homepage = "https://gitlab.com/nbdkit/libnbd";
@@ -59,5 +69,4 @@ stdenv.mkDerivation rec {
 }
 # TODO: package the 1.6-stable version too
 # TODO: git version needs ocaml
-# TODO: bindings for go, ocaml and python
-
+# TODO: bindings for go and ocaml
--- a/pkgs/development/libraries/libnbd/libnbd-metadata
+++ b/pkgs/development/libraries/libnbd/libnbd-metadata
@@ -1,0 +1,3 @@
+Metadata-Version: 2.1
+Name: nbd
+Version: @version@
--- a/pkgs/top-level/python-packages.nix
+++ b/pkgs/top-level/python-packages.nix
@@ -6617,6 +6617,11 @@ self: super: with self; {
     inherit (pkgs) libsodium;
   };
 
+  libnbd = toPythonModule (pkgs.libnbd.override {
+    buildPythonBindings = true;
+    python3 = python;
+  });
+
   libpcap = callPackage ../development/python-modules/libpcap {
     pkgsLibpcap = pkgs.libpcap; # Needs the C library
   };


pkgs/development/libraries/libnbd/default.nix
```diff
@@ -4,6 +4,7 @@
 , bash-completion
 , pkg-config
 , perl
+, withPython ? true, python3
```
AndersonTorres, line 7:
```suggestion
, withPython ? true
, python3
```

pkgs/development/libraries/libnbd/default.nix
```diff
@@ -32,8 +34,24 @@ stdenv.mkDerivation rec {
...
+      Metadata-Version: 2.1
+      Name: nbd
+      Version: ${version}
+    EOT
+  '';
+
+  outputs = [
+    "out"
+  ];
+
```
AndersonTorres, line 54:
```suggestion
```

Or is this somehow needed?

pkgs/development/libraries/libnbd/default.nix
```diff
@@ -32,8 +34,24 @@ stdenv.mkDerivation rec {
     libxml2
   ];
 
+  configureFlags = []
+    ++ lib.optionals withPython [ "--with-python-installdir=${placeholder "out"}/${python3.sitePackages}" ];
+
   installFlags = [ "bashcompdir=$(out)/share/bash-completion/completions" ];
 
+  postInstall = lib.optionalString withPython ''
+    mkdir -p ${placeholder "out"}/${python3.sitePackages}/nbd-${version}.dist-info
```
AndersonTorres, line 43: Instead of a heredoc, use a substituteAll template and something like

```suggestion
    install -Dm644 -t ${placeholder "out"}/${python3.sitePackages}/nbd-${version}.dist-info/METADATA ${./libndb-metadata}
```

pkgs/development/libraries/libnbd/default.nix
```diff
@@ -59,5 +77,4 @@ stdenv.mkDerivation rec {
 }
 # TODO: package the 1.6-stable version too
 # TODO: git version needs ocaml
-# TODO: bindings for go, ocaml and python
-
+# TODO: bindings for go and ocaml
```
AndersonTorres, line 80: Many thanks!

pkgs/development/libraries/libnbd/default.nix
AndersonTorres: Can this be unconditional?

pkgs/development/libraries/libnbd/default.nix
physics-enthusiast: Oops, that was from and earlier draft with multiple outputs ("out" and "python"). It can be safely removed.

pkgs/development/libraries/libnbd/default.nix
physics-enthusiast: Sorry, I don't quite understand the question? Python is only pulled in as a `buildInput` if `withPython` is set to `true` (to allow a user to override it if necessary for some reason). Is there other desired behaviour?

pkgs/development/libraries/libnbd/default.nix
physics-enthusiast: That does seem more elegant.

pkgs/development/libraries/libnbd/default.nix
AndersonTorres: In a certain sense, Python is employed here to create the bindings.
But the bindings would be created anyway - Hydra cache will contain both nbd and the python-nbd.
(IIRC the toPythonModule works mostly like a thin wrapper.)

So I don't see why not to build this bindings unconditionally.

physics-enthusiast: Hmm, I was thinking more along the lines of a user on a nonstandard `buildPlatform` compiling it themselves. For the purposes of Hydra it does indeed make no difference since we default the setting to `true` and don't override it to `false` anywhere. I feel like it's just a general good practice to provide the option to not pull in an entire language ecosystem as a build dependency since upstream made the bindings optional, although I recognise that this is less applicable with python being so ubiquitous that it's probably getting pulled in somewhere else anyway. If you still want it, I can go ahead and make it unconditional.

AndersonTorres: That being said, a better name for this variable would be `buildPythonModule` instead of `withPython`.
```suggestion
, buildPythonModule ? true
, python3
```

physics-enthusiast: How about just `pythonBindings`?

pkgs/development/libraries/libnbd/default.nix
```diff
@@ -32,8 +34,17 @@ stdenv.mkDerivation rec {
     libxml2
   ];
 
+  configureFlags = []
+    ++ lib.optionals withPython [ "--with-python-installdir=${placeholder "out"}/${python3.sitePackages}" ];
```
SuperSandro2000, line 38: We don't need the [] here

pkgs/development/libraries/libnbd/default.nix
SuperSandro2000, line 7: This should by default be false because we don't need the binding in the normal package, do we?

physics-enthusiast: I'm honestly on the fence with this one. I've searched for similar options in other packages and it seems to be fairly 50-50 whether the default is off or on. I guess the counterpoint is that defaulting to `true` means you have one fewer copy of the derivation floating around in the cache?

AndersonTorres: This is a small hint from an [old book](https://www.amazon.com/Practice-Programming-Addison-Wesley-Professional-Computing/dp/020161586X) about naming ~~variables~~ attributes:
Boolean variables should have Boolean-ish names.

In Scheme this is easy: just add a question mark at the end.
But this is not Guix, so we need another convention.

The purpose of this attribute is to control if the Python module will be built. Then a good name would be `buildPythonModule`, or maybe `buildPythonBindings` (to not clash with the already used function name `buildPythonModule`). 

Think on this like asking a question, "do you want to build the Python bindings?" Yes = true, No = false.

On the other hand, the other two are a bit ambiguous:

- `withPython` looks like it will bring all Python ecosystem bundled
- `pythonBindings` looks like a list of Python bindings provided after the build.

AndersonTorres: I think Nixpkgs will be smart enough to not to do useless duplication.

physics-enthusiast:
Changes have been made as requested

AndersonTorres (changes requested):
Squash all commits.

Otherwise, code is fine.

pkgs/development/libraries/libnbd/default.nix
```diff
@@ -32,8 +34,16 @@ stdenv.mkDerivation rec {
     libxml2
   ];
 
+  configureFlags = lib.optionals buildPythonBindings [ "--with-python-installdir=${placeholder "out"}/${python3.sitePackages}" ];
+
   installFlags = [ "bashcompdir=$(out)/share/bash-completion/completions" ];
 
+  postInstall = lib.optionalString buildPythonBindings ''
+    LIBNBD_PYTHON_METADATA='${placeholder "out"}/${python3.sitePackages}/nbd-${version}.dist-info/METADATA'
```
AndersonTorres, line 42: Being overzealous:

```suggestion
    local LIBNBD_PYTHON_METADATA='${placeholder "out"}/${python3.sitePackages}/nbd-${version}.dist-info/METADATA'
```

AndersonTorres (approved):
LGTM

AndersonTorres:
Is this PR rebased over Master? Here my libnbd is 1.20.1.

SuperSandro2000:
@ofborg build python3Packages.libnbd

PR #250627, opened 2023-08-21
ride: init at 4.5.4097

TomaSajt:
## Description of changes

This PR adds 1 package: `ride`

This name is quite simple so another option could be `dyalog-ride`.

Homepage: https://github.com/dyalog/ride

RIDE is a **Remote** IDE for Dyalog, so Dyalog is not a hard dependency of RIDE.
The package `dyalog` can be found in Nixpkgs, so you can use that for testing out the program.

For testing: In the startup window click on the arrow in the top-right corner and put "dyalog" as the executable path (the names of the binaries in `dyalog`.

----

Outside of Nixpkgs, RIDE would detect the installed versions of Dyalog (`/opt/mdyalog`).
We could patch this so that it lists the interpreters specified in an attribute. The issue with that is that when the interpreter's store path changes, the old interpreters nix path will remain in the interpreter selection box, and you would have to select the updated interpreter from the dropdown. 
We could patch this also, so that if its a `/nix/store/` path, it gets cleared if it isn't specified in the attribute.

Anyways, that's outside of the scope of this PR

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
- [ ] For non-Linux: Is `sandbox = true` set in `nix.conf`? (See [Nix manual](https://nixos.org/manual/nix/stable/command-ref/conf-file.html))
- [ ] Tested, as applicable:
  - [NixOS test(s)](https://nixos.org/manual/nixos/unstable/index.html#sec-nixos-tests) (look inside [nixos/tests](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests))
  - and/or [package tests](https://nixos.org/manual/nixpkgs/unstable/#sec-package-tests)
  - or, for functions and "core" functionality, tests in [lib/tests](https://github.com/NixOS/nixpkgs/blob/master/lib/tests) or [pkgs/test](https://github.com/NixOS/nixpkgs/blob/master/pkgs/test)
  - made sure NixOS tests are [linked](https://nixos.org/manual/nixpkgs/unstable/#ssec-nixos-tests-linking) to the relevant packages
- [ ] Tested compilation of all packages that depend on this change using `nix-shell -p nixpkgs-review --run "nixpkgs-review rev HEAD"`. Note: all changes have to be committed, also see [nixpkgs-review usage](https://github.com/Mic92/nixpkgs-review#usage)
[...]

PR #313397, opened 2024-05-21
ride: add darwin support

TomaSajt:
## Description of changes

This PR adds darwin support to the `ride` package.

I was working on supporting darwin in https://github.com/NixOS/nixpkgs/pull/250627, but I gave up since I didn't have a darwin machine to do debugging.

I have finally set up a `macos-ventura` VM so I can test on Darwin now. 

Turns out I was very very close, the only thing missing was that I accidentally removed some of the execute permissions from the electron directory, which made it not be able to launch.

Now, instead of doing `cp -r --no-preserve=all` I do `cp -r` with `chmod -R u+w`, this way the directory becomes writable without messing up the other permissions.

---

Note: since `dyalog` is not packaged for Darwin, you'll have to connect to a `dyalog` instance remotely if you want to test it out. (Or just run the non-nix package)

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [x] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/by-name/ri/ride/package.nix
+++ b/pkgs/by-name/ri/ride/package.nix
@@ -12,6 +12,7 @@
   copyDesktopItems,
   makeDesktopItem,
   electron,
+  darwin,
 }:
 
 let
@@ -31,6 +32,8 @@ let
   };
 
   platformInfo = platformInfos.${stdenv.system};
+
+  electronDist = electron + (if stdenv.isDarwin then "/Applications" else "/libexec/electron");
 in
 buildNpmPackage rec {
   pname = "ride";
@@ -83,13 +86,20 @@ buildNpmPackage rec {
     popd
   '';
 
-  nativeBuildInputs = [
-    zip
-    makeWrapper
-    copyDesktopItems
-  ];
+  nativeBuildInputs =
+    [
+      zip
+      makeWrapper
+    ]
+    ++ lib.optionals (!stdenv.isDarwin) [ copyDesktopItems ]
+    ++ lib.optionals stdenv.isDarwin [ darwin.cctools ];
 
   env.ELECTRON_SKIP_BINARY_DOWNLOAD = "1";
+
+  # Fix error: no member named 'aligned_alloc' in the global namespace
+  env.NIX_CFLAGS_COMPILE = lib.optionalString (
+    stdenv.isDarwin && lib.versionOlder stdenv.hostPlatform.darwinSdkVersion "11.0"
+  ) "-D_LIBCPP_HAS_NO_LIBRARY_ALIGNED_ALLOCATION=1";
 
   npmBuildFlags = platformInfo.buildCmd;
 
@@ -97,8 +107,12 @@ buildNpmPackage rec {
   # Here, we create a local cache of electron zip-files, so electron-packager can copy from it
   postConfigure = ''
     mkdir local-cache
-    cp -r --no-preserve=all ${electron}/libexec/electron electron
-    pushd electron
+
+    # electron files need to be writable on Darwin
+    cp -r ${electronDist} electron-dist
+    chmod -R u+w electron-dist
+
+    pushd electron-dist
     zip -qr ../local-cache/electron-v${electron.version}-${platformInfo.zipSuffix}.zip *
     popd
   '';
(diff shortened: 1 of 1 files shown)


pkgs/by-name/ri/ride/package.nix
```diff
@@ -109,12 +124,20 @@ buildNpmPackage rec {
 
     install -Dm644 ThirdPartyNotices.txt -t $out/share/doc/ride
 
-    mkdir -p $out/share/ride
-    cp -r locales resources{,.pak} $out/share/ride
-    makeWrapper ${lib.getExe electron} $out/bin/ride \
-      --add-flags $out/share/ride/resources/app.asar \
-      --add-flags "\''${NIXOS_OZONE_WL:+\''${WAYLAND_DISPLAY:+--ozone-platform-hint=auto --enable-features=WaylandWindowDecorations}}" \
-      --inherit-argv0
+    ${lib.optionalString (!stdenv.isDarwin) ''
```
SuperSandro2000, line 127:
```suggestion
  '' + lib.optionalString (!stdenv.isDarwin) ''
```
that's easier than string interpolation inside strings

pkgs/by-name/ri/ride/package.nix
```diff
@@ -109,12 +124,20 @@ buildNpmPackage rec {
...
+      mkdir -p $out/share/ride
+      cp -r locales resources{,.pak} $out/share/ride
+      makeWrapper ${lib.getExe electron} $out/bin/ride \
```
SuperSandro2000, line 130: Nice usage of getExe 👍🏼

pkgs/by-name/ri/ride/package.nix
```diff
@@ -109,12 +124,20 @@ buildNpmPackage rec {
...
+        --add-flags $out/share/ride/resources/app.asar \
+        --add-flags "\''${NIXOS_OZONE_WL:+\''${WAYLAND_DISPLAY:+--ozone-platform-hint=auto --enable-features=WaylandWindowDecorations}}" \
+        --inherit-argv0
+    ''}
+
+    ${lib.optionalString (stdenv.isDarwin) ''
```
SuperSandro2000, line 136: We could probably work with if then else here, too.

pkgs/by-name/ri/ride/package.nix
```diff
@@ -79,11 +88,13 @@ buildNpmPackage rec {
...
-    makeWrapper
-    copyDesktopItems
-  ];
+  nativeBuildInputs =
+    [
+      zip
+      makeWrapper
+    ]
+    ++ lib.optionals (!stdenv.isDarwin) [ copyDesktopItems ]
+    ++ lib.optionals stdenv.isDarwin [ darwin.cctools ];
```
SuperSandro2000, line 97:
```suggestion
  nativeBuildInputs = [
    zip
    makeWrapper
  ] ++ lib.optionals stdenv.isLinux [ copyDesktopItems ]
    ++ lib.optionals stdenv.isDarwin [ darwin.cctools ];
```

pkgs/by-name/ri/ride/package.nix
```diff
@@ -33,8 +41,9 @@ let
 
   # Electron 27 is the latest version that works as of RIDE version 4.5.4097
   electron = electron_27;
+  electronDist = "${electron}/${if stdenv.isDarwin then "Applications" else "libexec/electron"}";
```
SuperSandro2000, line 44:
```suggestion
  electronDist = electron + if stdenv.isDarwin then "Applications" else "libexec/electron";
```
is that valid nix?

SuperSandro2000 (commented):
already looking good

pkgs/by-name/ri/ride/package.nix
TomaSajt: Should I replace `!stdenv.isDarwin` with `stdenv.isLinux` everywhere?

Also, the formatting is done by `nixfmt-rfc-style`, so I won't change that.

pkgs/by-name/ri/ride/package.nix
TomaSajt: I think you meant something like this:
```nix
electronDist = electron + (if stdenv.isDarwin then "/Applications" else "/libexec/electron");
```
Not sure which one is better, though this new one doesn't use strings inside strings.

Ideally, this should be a property provided by the electron packages as `electron.dist`, but I haven't yet talked about this with the maintainers.

pkgs/by-name/ri/ride/package.nix
```diff
@@ -79,42 +81,61 @@ buildNpmPackage rec {
...
+    ++ lib.optionals (!stdenv.isDarwin) [ copyDesktopItems ]
+    ++ lib.optionals stdenv.isDarwin [ darwin.cctools ];
 
   env.ELECTRON_SKIP_BINARY_DOWNLOAD = "1";
 
+  # Fix error: no member named 'aligned_alloc' in the global namespace
+  env.NIX_CFLAGS_COMPILE = lib.optionalString (
+    stdenv.isDarwin && lib.versionOlder stdenv.hostPlatform.darwinSdkVersion "11.0"
+  ) "-D_LIBCPP_HAS_NO_LIBRARY_ALIGNED_ALLOCATION=1";
+
```
TomaSajt, line 98: I added this instead of doing `overrideSDK` since this requires less overriding.

pkgs/by-name/ri/ride/package.nix
TomaSajt: I don't really agree with this. This is very similar to how bash's `if` statements look. I'm not using `if then else` because that would add extra levels of indentation and it doesn't really look much more readable.

pkgs/by-name/ri/ride/package.nix
TomaSajt: https://github.com/NixOS/nixpkgs/pull/313397#discussion_r1636176238

PR #313497, opened 2024-05-21
nixos-option: rewrite as a nix script

### LINKED ISSUE #97855: nixos-option does not support flakes
**Describe the bug**
Running `nixos-option nix.nixPath` fails because I don't have a `nixpkgs` entry in my NIX_PATH. This requires me to use my compatibility wrapper (https://github.com/cole-h/nixos-config/tree/c02cdf4702db3e383073426396dea31c40ae6e71/compat) with `-I nixpkgs=compat`; otherwise, the program aborts.

**To Reproduce**
1. `nixos-option nix.nixPath` without `nixpkgs` in NIX_PATH
2. Observe abort due to lack of `<nixpkgs/nixos>`

**Expected behavior**
I would expect a `--flake` flag, just like `nixos-rebuild` has, to be able to e.g. `nixos-option nix.nixPath --flake .#hostname`.

**Notify maintainers**
cc @Ma27 @chkno 

**Metadata**
 - system: `"x86_64-linux"`
 - host os: `Linux 5.8.7-zen1, NixOS, 20.09.20200912.ac8a276-cosmere (Nightingale)`
 - multi-user?: `yes`
 - sandbox: `yes`
 - version: `nix-env (Nix) 3.0pre20200903_8a945d6`
 - channels(vin): `""`
 - channels(root): `""`
 - nixpkgs: `/home/vin/.config/nixpkgs/compat`

Maintainer information:
```yaml
# a list of nixpkgs attributes affected by the problem
attribute:
# a list of nixos modules affected by the problem
module: installer.tools.nixos-option
```

### LINKED ISSUE #293543: `nixos-option` needs rewriting
## Issue description

I had the displeasure of touching this code in https://github.com/NixOS/nixpkgs/pull/293542, and it is obvious that this tool has utterly bitrotted. I am not convinced it should be using the C++ API to begin with.

It really should be rewritten to use the CLI or C API. All the info is there and can absolutely be acquired via the CLI with a clever bit of Nix scripting.

## Technical details

FireyFly:
This ports the functionality of the C++ `nixos-option` to a nix script (with a tiny shellscript for argument processing and invoking the nix script). The main impetus for this is that the current C++ version seems a bit verbose, unnecessarily tightly coupled to cppnix internals, fragile (with the occasional segfaults if you hold it slightly wrong).

While I'm rewriting it anyway, I opted to improve the output format slightly to make it easier on the eyes (there's no point in the type/description strings and declared/defined by lists as nix values; instead just make sure each output is indented with two spaces).

There's some changes in behaviour especially when it comes to errors whilst evaluating options. In recursive mode, errors that can be caught with `builtins.tryEval` render as just `«error»` whereas fatal errors bubble to the top and stop the recursion (the old C++ version prints the full stacktrace inline in place of the value for each option where evaluating the value errors).

The new version supports descending into `attrsOf`/`listOf` submodule options (as long as they're defined in the user's config)--the previous version doesn't seem to understand them (and segfaults in the recursive case, e.g. `nixos-option -r users.users.youruser`).

The new version (well, the shell wrapper part of it) supports `--show-trace` for full error traces.

Performance is basically unchanged vs the existing version (I think most of the time is spent evaluating nixos anyway).

Fixes #293543
Fixes  #97855

---

Before this is merged, it'd be nice if it was a bit more widely tested (so far I've just tested it a bit on random options/paths locally).

The manual page should probably also be updated, and release notes added, but I figured it makes sense to get some feedback on the derivation changes first.

## Description of changes

* C++ implementation of `nixos-option` removed 
* nixlang implementation added, with a small shellscript wrapper to invoke it
  - (this uses `nix-instantiate --eval --json | jq -r` as a hack to output raw strings; suggestions for improvements welcome)
* new `default.nix`; I took the liberty to change the license since it's a fresh implementation anyway and doesn't link against nix (and if anything I took some notes from the `nixos-rebuild` derivation, which is also MIT licensed)

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---


K900 (commented):
Have not looked closely yet, but yes please.

pkgs/tools/nix/nixos-option/nixos-option.nix
```diff
@@ -0,0 +1,134 @@
...
+      renderShort = n: v: "${lib.showOption (path' ++ n)} = ${safeToPretty v};";
+      mapAttrsRecursive' = safeMapAttrsRecursiveCond (x: !lib.isDerivation x);
+    in
+      if lib.isAttrs config
+      then lib.concatStringsSep "\n" (lib.collect lib.isString (mapAttrsRecursive' renderShort config))
+      else renderShort [] config;
+
+in
+  if !lib.hasAttrByPath path' nixos.config
+  then throw "Couldn't resolve config path '${lib.showOption path'}'"
```
stuebinm, line 126: perhaps it would be worth it not to `throw` here, and instead return some structured error which the wrapper script can display nicely?

The output for nonexistent config options is the only thing which feels worse than in the old version to me.

stuebinm (commented):
I really like this rewrite, and would love to see it in nixpkgs.

There are some slight regressions compared to the existing implementation:
 - the old implementation supported a couple more flags (notably `--help`), though these were apparently entirely undocumented
 - the old version had a marginally nicer error message for nonexisting option paths; the new version currently just throws a (Nix) error (see below)

Additionally, I think it'd be nice to update the man page as well — it contains some example output which has changed slightly, and the `--show-trace` option is not mentioned there.

pkgs/tools/nix/nixos-option/nixos-option.nix
FireyFly: hmm, at first I just returned strings in case of errors, but then figured we probably want nonzero exit codes and so opted to `throw` instead

Maybe it'd be an idea to still `throw` but format the errors from `nixos-option` itself with a specific prefix that we can look for in stderr in the wrapper script, and if present cut off the stacktrace and only show the error message (plus perhaps also improve the message a bit)

I like sticking to throwing in the .nix script since it makes it less strictly coupled to being used with the wrapper script

Good point re `--help`--we could handle that and show the manpage in the wrapper script easily enough. The manpage & release notes should be updated but I wanted feedback on the functionality first ^^

pkgs/tools/nix/nixos-option/nixos-option.sh
```diff
@@ -0,0 +1,47 @@
...
+while [[ $# -gt 0 ]]; do
+  case "$1" in
+    -r|--recursive)
+      recursive=true
+      shift
+      ;;
+
+    -I)
+      nix_args+=("$1" "$2")
+      shift; shift
```
SuperSandro2000, line 17:
```suggestion
      shift 2
```

r-vdp:
I guess we can also drop the version pin for nix in `all-packges.nix` then, and in `nixos/modules/installer/tools/tools.nix` we can then override the nix input of nixos-option with `config.nix.package`.

eclairevoyant:
Let's keep that as a separate PR; I'd personally say this PR should be for establishing parity with the original tool (as much as possible), which is already a fair bit of work. Keeping new functionality as a separate feature would be easier for contributors and reviewers IMO.

pkgs/tools/nix/nixos-option/nixos-option.nix
```diff
@@ -0,0 +1,134 @@
...
+    let
+      renderShort = n: v: "${lib.showOption (path' ++ n)} = ${safeToPretty v};";
+      mapAttrsRecursive' = safeMapAttrsRecursiveCond (x: !lib.isDerivation x);
+    in
+      if lib.isAttrs config
+      then lib.concatStringsSep "\n" (lib.collect lib.isString (mapAttrsRecursive' renderShort config))
+      else renderShort [] config;
+
+in
+  if !lib.hasAttrByPath path' nixos.config
```
AndersonTorres, line 125: Stylistic issue: 

instead of `if !P then A else B`
use `if P then B else A`.

pkgs/tools/nix/nixos-option/nixos-option.sh
```diff
@@ -0,0 +1,51 @@
+#! @runtimeShell@
```
AndersonTorres, line 1: `shellcheck` is a nice tool to enable.

pkgs/tools/nix/nixos-option/nixos-option.nix
FireyFly: Hm, that was intentional since I found it easier to read (kind of as an "early return" of sorts, taking care of the error case first) but I'm happy to rephrase it if it's easier to read without the negation

FireyFly:
@azuwis:
> I've added flake support in this commit https://github.com/azuwis/nixpkgs/commit/1bd0fb37624cc63976e8421a7d2102807f15b366, feel free to merge it into this PR.

I appreciate covering flake support and think it's something that would be nice to have, but I think I agree with @eclairevoyant to first focus this PR on essentially parity with the existing C++ version.. that said, after that it'd be nice to add such support.

---

Apart from that, I got a bit distracted by other things around me, but i should be able to get back to this PR now. I appreciate all the feedback and I'll try to incorporate them into this PR soon.

I didn't do thorough tests of running my tool on every possible option subtree, but @philiptaron poked me on Matrix as well with some feedback taking that kind of route, hence also the linked PRs opened referencing this one. Appreciate this as well!

pkgs/tools/nix/nixos-option/nixos-option.sh
FireyFly: Good point, i haven't really used it since I haven't written many "real" (not just for my own use) shellscripts. I didn't see any specific hooks for it but I guess I can just add `# shellcheck shell=bash` and explicitly invoke `shellcheck` on the script in the checkPhase

philiptaron (commented):
I'm actively testing this out. There are definitely some sharp edges, some of them issues in this tool and some of them issues in the codebase.

- https://github.com/NixOS/nixpkgs/pull/320850
- https://github.com/NixOS/nixpkgs/pull/320846 (but this will be likely rejected)
- https://github.com/NixOS/nixpkgs/pull/320836
- https://github.com/NixOS/nixpkgs/pull/320830

There are likely many others. This was sufficient to get most options printed.

As I do use flakes to configure my NixOS system, I am using a mildly modified version as testing fodder, where I pass `--arg nixos (builtins.getFlake "github:philiptaron/flock.nix/eeeeeee").nixosConfigurations.zebul` to interrogate that flake's configuration.

pkgs/tools/nix/nixos-option/nixos-option.nix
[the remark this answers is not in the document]
FireyFly: Hm, I find to me it's a mix of improving or worsening readability... e.g. not sure about what it does with `flatMapSlices` or the call to it in `readOption`, and some of rewrapping `if then else` to one line is a bit harder to read than having it split up in my opinion...

I could see if I can rephrase those bits in a way that the `nixfmt-rfc-style`-reformatted ends up more sensical too, if you think that would be valuable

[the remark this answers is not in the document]
eclairevoyant: per the current CI checks, this should be formatted with that formatter now; please rebase on master, `nix-shell`, then `nixfmt <file>`

Aleksanaa:
> Compared to the cpp version of nixos-option, the nix script version is 3 times faster when showing options that display derivations like `programs.sway.package`:

Thanks for testing, but this may have some eval caching involved?

Aleksanaa:
@philiptaron we may want to skip the parsing check on `nixos-option.nix` file, which is not used here but will be installed into the package. Is this currently possible?

(I don't want to rename it to `nixos-option.nix.in`, as I still want some CI checks on this, like `nixf-tidy`, and my editor won't highlight it by default)

pkgs/tools/nix/nixos-option/default.nix
```diff
@@ -1,38 +1,38 @@
...
-, pkg-config
+, runtimeShell
+, substitute
 , installShellFiles
 , nix
+, jq
 }:
 
-stdenv.mkDerivation {
+substitute {
```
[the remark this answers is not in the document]
Aleksanaa: We cannot install that manpage currently with `writeShellApplication`.

Aleksanaa:
So I finally decided to use `resholve.mkDerivation`.

`substitute`'s params are a little bit ugly as they can't be gathered into a string, and look even worse after formatting, and it cannot resolve dependencies, so it will make the script a bit stiff to write.

`writeShellApplication` looks good at first, as we can simply replace `@...@` with `lib.replaceStrings` after `lib.readFile`, but we cannot easily add the manpage together with it, unless we add another layer of `symlinkJoin`, which seems a bit dull. So I wanted to add a `extraInstallCommands` like appimage builder does, but then I thought there's a reason behind calling it "write". It's not for more complex derivations.

`resholve.mkDerivation` is kinda good, as it automatically completes executables to absolute paths and throws errors when dependencies are missing, which is nice to avoid impurity. It's just that we need to call `shellcheck` explicitly, but this can be written as a hook in the future, so there is nothing to complain about.

More opinions are welcomed!

pkgs/tools/nix/nixos-option/nixos-option.8
Aleksanaa, line 1: The manpage still has some room for improvements, and I haven't figured out how's specifying a non-flake directory working. The `NIXOS_CONFIG` can still be useful but seems not implemented? Maybe just add some simple processing in the script

nixos/doc/manual/release-notes/rl-2411.section.md
```diff
@@ -16,6 +16,8 @@
 - `hardware.display` is a new module implementing workarounds for misbehaving monitors
   through setting up custom EDID files and forcing kernel/framebuffer modes.
 
+- `nixos-option` has been rewritten to a Nix expression called by a simple bash script. This lowers our maintenance threshold, makes eval errors less verbose, and adds support for descending into `attrsOf` and `listOf` submodule options, and `--show-trace` support.
```
Aleksanaa, line 19: Forgot to mention flakes

lf-:
> Also is there really no better option than relying on resholve? It uses _python 2_ of all things...
oh this is oil shell's fault. i thought they had forked Python 2 or something for their purposes. not quite certain but whatever they're up to it's strange.

someone should probably rewrite the whole thing in haskell using shellcheck's parser instead or so, if that's not already happened.

eclairevoyant:
It's completely unnecessary to use `resholve` here.

Rather than using some tooling that uses some hidden magic, isn't properly documented, and only one person knows how to use correctly, let's stick with standard builders. Since we're calling `shellcheck` manually anyway, the benefit of using `writeShellApplication` is gone, so we can fall back to the well-known `stdenvNoCC.mkDerivation`.

Pushed a fix for that, as well as the manpage issue below:
```
$ result/bin/nixos-option --help
/nix/store/k46l5ki5ppfsrbz0wpzxs707zxk9s669-man-db-2.12.1/bin/man: /nix/store/qg0qhvikfg846fq0l9iqrqbfflm223iv-nixos-option-unresholved-0/share/man/man8/nixos-option.8: No such file or directory
```

eclairevoyant:
Something still seems to be broken, though, it still fail(ed/s) to eval on my (flake-based) config. Whereas the `nix eval $(dirname $(realpath /etc/nixos/flake.nix))#nixosConfigurations.$(hostname).options.<optionName>.definitionsWithLocations` oneliner works fine.

I'll try to get an MWE together, once I get some more time.

Aleksanaa:
> it still fail(ed/s) to eval on my (flake-based) config.

Works for me:

<details>
<summary>Output</summary>

```
=> ./result/bin/nixos-option boot.loader.efi.efiSysMountPoint
Value:
  "/boot"

Default:
  "/boot"

Type:
  string

Description:
  Where the EFI System Partition is mounted.

Declared by:
  /nix/store/l6spwdi1xs20q24fqfi3xwxx0d2x9q2m-source/nixos/modules/system/boot/loader/efi.nix

Defined by:
  /nix/store/l6spwdi1xs20q24fqfi3xwxx0d2x9q2m-source/nixos/modules/system/boot/loader/efi.nix

=> ./result/bin/nixos-option boot.loader.efi.efiSysMountPoint --flake /etc/nixos
Value:
  "/boot"

Default:
  "/boot"

Type:
  string

Description:
  Where the EFI System Partition is mounted.

Declared by:
  /nix/store/l6spwdi1xs20q24fqfi3xwxx0d2x9q2m-source/nixos/modules/system/boot/loader/efi.nix

Defined by:
  /nix/store/l6spwdi1xs20q24fqfi3xwxx0d2x9q2m-source/nixos/modules/system/boot/loader/efi.nix

=> ./result/bin/nixos-option boot.loader.efi.efiSysMountPoint --flake /etc/nixos#Aleksana-Laptop
Value:
  "/boot"

Default:
  "/boot"

Type:
  string

Description:
  Where the EFI System Partition is mounted.

Declared by:
  /nix/store/l6spwdi1xs20q24fqfi3xwxx0d2x9q2m-source/nixos/modules/system/boot/loader/efi.nix

Defined by:
  /nix/store/l6spwdi1xs20q24fqfi3xwxx0d2x9q2m-source/nixos/modules/system/boot/loader/efi.nix
```

</details>

Aleksanaa:
> One thing I've noticed is the cpp version of nixos-option can handle options with a trailing dot

Fixed now

Aleksanaa:
Weird, seems I cannot let GitHub parse @eclairevoyant 's anonymous email placeholder, and when I add it azuwis's entry is also gone

pkgs/tools/nix/nixos-option/nixos-option.sh
```diff
@@ -0,0 +1,94 @@
...
+      hostname=default
+    fi
+    flakeAttr="nixosConfigurations.\"$hostname\""
+  else
+    flakeAttr="nixosConfigurations.\"$flakeAttr\""
+  fi
+  nix_args+=(--arg nixos "(builtins.getFlake \"$flake\").$flakeAttr")
+fi
+
+if [[ -z $flake ]]; then
```
Aleksanaa, line 74: Still wrong, fix later

eclairevoyant:
Turns out to be a [nix ux bug](https://github.com/NixOS/nix/issues/5836), lol

So anyway if you're using flakes and git for the config, and there are untracked files, it will evaluate differently with this `nixos-option` script unless you explicitly specify `git+file://` prefix

lf-:
lol lmao. we could detect it in the shell script though, since we could look for .git in parent directories or even just abuse `git rev-parse --show-toplevel` to figure out if it's a git repo i guess.

Aleksanaa:
I'm a bit confused on that, my flake is in `/etc/nixos`, is a git repo, has unstaged changes, but it works just as expected?

eclairevoyant:
I said unstaged _files_ not unstaged changes.

Flake-based nix in a git repo (i.e. `git+<whatever>://` urls) will only see paths that are part of the git index. However, if you use `path://` urls, nix will see whatever the filesystem sees - in fact it will even copy the `.git` folder (and of course `.gitignored` files) to the store.

When you don't explicitly specify a protocol, but provide a path, nix on the CLI and when reading flake inputs will try to intelligently guess whether it's a git-type flake or not. On the other hand, `getFlake` (as I just rediscovered) will not, and just assumes it's `path://`; to override that behaviour requires using `git+file://` explicitly.

That discrepancy in behaviour will lead to different evaluation if you were to use, say, `nixos-rebuild` vs this `nixos-option`. Unless we avoid using `getFlake` or use the workaround above of trying to "intelligently" determine the protocol ourselves.

lf-:
I can do you one worse, I believe there is behaviour inconsistency that 9999years found while fiddling with CppNix 2.20 a while ago (which may well exist in lix also) where `git+file://something` behaves differently than `git+file:something`.

The exact details elude me since it was several months ago, but it's truly Nix levels of broken.

nixos/doc/manual/release-notes/rl-2411.section.md
```diff
@@ -16,6 +16,8 @@
 - `hardware.display` is a new module implementing workarounds for misbehaving monitors
   through setting up custom EDID files and forcing kernel/framebuffer modes.
 
+- `nixos-option` has been rewritten to a Nix expression called by a simple bash script. This lowers our maintenance threshold, makes eval errors less verbose, adds support for flake-based configurations, desending into `attrsOf` and `listOf` submodule options, and `--show-trace`.
```
FireyFly, line 19:
```suggestion
- `nixos-option` has been rewritten to a Nix expression called by a simple bash script. This lowers our maintenance threshold, makes eval errors less verbose, adds support for flake-based configurations, descending into `attrsOf` and `listOf` submodule options, and `--show-trace`.
```

FireyFly (commented):
Sorry about dropping this and going kinda AWOL.. thanks for picking it up @Aleksanaa, and adding flakes support/addressing the other feedback here.

Aleksanaa:
Also resolved flake inside git repo. `-e` is used instead of `-d` in case someone is using flake in a git submodule, although unlikely to happen.

eclairevoyant:
(it was in the wrong `if` block and missing a `/`)

lf-:
That git detection is broken. You cannot assume the flake dir itself is a git repo because the flake could well be in a subdirectory. I suggested `git rev-parse --show-toplevel` because it avoids this bug.

eclairevoyant:
I think we should just drop flake support.

lf-:
I also don't think we've at all exhausted the ways to not use getFlake, like --apply. So yeah I think we should merge this without flake support and figure that out in a separate change.

Mindavi:
I agree with this conceptually, haven't really used the tool before so I find it hard to give in-depth comments.

nyabinary:
> I think we should just drop flake support. It wasn't in the original script anyway and it needs a full design in a separate PR.

I mean, flakes are very widely used, so I don't know if that would be the wisest decision? Idk.

lf-:
> > I think we should just drop flake support. It wasn't in the original script anyway and it needs a full design in a separate PR.
> 
> I mean, flakes are very widely used, so I don't know if that would be the wisest decision? Idk.

The reason we should do this is that it has taken three months to merge this, which is very silly given that it was in a state to be an incremental improvement (minus new flake functionality) a month or two ago. Thus the flake functionality should be split out and worked on separately.

nyabinary:
What further work is needed to get this in?

lf-:
> What further work is needed to get this in?

The flakes pieces need to be removed and then it can be merged with no other changes.

Mic92:
How are you planning to support flakes? With the C++ version we wouldn't have to rely on `builtins.getFlake`

lf-:
> How are you planning to support flakes? With the C++ version we wouldn't have to rely on `builtins.getFlake`

There are a few ways to do this and get around the fact that running nix code in the context of a flake is way too hard for no reason. One way to do this is using `--apply`on `nix eval`. Being involved in maintaining the Nix implementations, this is our problem, and if there is no viable option, we will have to fix the CLI to be able to do this kind of obvious thing that has been broken for years.

The flake support in `nixos-option` is already nonexistent, and I think it is a *super* bad example to be setting to have this kind of Nix language inspection tooling be impossible to write except in C++.

My immediate answer is "don't worry about it", and my longer-term answer is "this *must* be possible [or made to be possible] because it would be fundamentally unserious of Nix/Lix to not be able to build this".

aviallon:
I used your PR to debug my NixOS config using Flakes, and it works marvelously. Thank you.

thiagokokada:
I am also interested in seeing this merged. IMO right now I already have a wrapper for `nixos-option` to make it work with flakes:

```nix
    nixos-option =
      let
        prefix = ''
          (import ${inputs.flake-compat} {
            src = ${flake};
          }).defaultNix.nixosConfigurations.\$(hostname)
        '';
      in
      prev.runCommand "nixos-option"
        {
          buildInputs = with prev; [
            makeWrapper
            installShellFiles
          ];
        }
        ''
          makeWrapper ${prev.nixos-option}/bin/nixos-option $out/bin/nixos-option \
          --add-flags --config_expr \
          --add-flags "\"${prefix}.config\"" \
          --add-flags --options_expr \
          --add-flags "\"${prefix}.options\""

          installManPage ${prev.nixos-option}/share/man/**/*
        '';
```

I assume this will continue working with zero or minimal changes. So I concur that not supporting Flakes right now is fine, we can have another PR adding support for Flakes later.

Mic92:
@thiagokokada maybe just make a new pull request.

thiagokokada:
> @thiagokokada maybe just make a new pull request.

I have little to no context about this change, but maybe one of the maintainers can create another PR?

benley:
> > @thiagokokada maybe just make a new pull request.
> 
> I have little to no context about this change, but maybe one of the maintainers can create another PR?

According to github it should be possible for anyone with commit privileges in nixpkgs to amend the existing PR, fwiw:
image

thiagokokada:
> > > @thiagokokada maybe just make a new pull request.
> > 
> > I have little to no context about this change, but maybe one of the maintainers can create another PR?
> 
> According to github it should be possible for anyone with commit privileges in nixpkgs to amend the existing PR, fwiw:
> image
> 
> (though it may be more polite to open a new one?  @FireyFly, what is your preference?)
> 

I know it is possible, but it would be better done by the people that understand what is happening to fix possible issues found in the review.

I may open another PR anyway if there is no one else interested, but would prefer to leave it to the people that already invested in the source code.

thiagokokada:
Opened a new PR: https://github.com/NixOS/nixpkgs/pull/369151.

Issue #166401, opened 2022-03-30
apple-m1 compiler optimisations for aarch64-darwin?


Atemu:
## Issue description

https://github.com/NixOS/nixpkgs/blob/dd697d414853a99d0d5f05ee23430923c6da1176/lib/systems/platforms.nix#L393-L398

adds inappropriate `arch` args for apple-m1. According to Apple clang, `arch = "armv8.5a+fp-armv8+neon+crc+crypto+dotprod+fp16fml+ras+lse+rdm+rcpc+zcm+zcz+fullfp16+sm4+sha3+sha2+aes";`.

(The string isn't even correct for `apple-13` AFAICT, it's `armv8.4a`.)

Is there any reason against setting `arch` to the string above?

Secondly, is there any reason against setting `cpu = "apple-m1";` once llvm12 is default? llvm11 doesn't support that cpu type; what would happen if a package's llvm was overridden to be llvm11?

cc @thefloweringash @NixOS/darwin-maintainers

thefloweringash:
The current flags come from https://github.com/NixOS/nixpkgs/pull/105026#issuecomment-774757005 by @bobrik. 

> Is there any reason against setting arch to the string above?

Not that I know of.

> Secondly, is there any reason against setting `cpu = "apple-m1";` once llvm12 is default?

That does also sound like an improvement, and might even resolve the TODO items added in 4aa95e33121c (https://github.com/NixOS/nixpkgs/pull/115235#discussion_r626226534). We could probably handle llvm11 compatibility in the cc-wrapper as a special case, if it's still needed. I don't expect there to be a large difference between llvm11 and llvm12.

Though we made a concession to llvm11 to revert the new default setting of `-fno-common` (c99904e11307cbade607ad855a29f07694762f37). We'd either need to make the same change in llvm12, or progress https://github.com/NixOS/nixpkgs/pull/110571.

Atemu:
@thefloweringash fno-common migration is merged. Does that mean we can upgrade llvm to 12 (or higher?) for aarch64-darwin?

toonn:
Main blocker rn would be we can't upgrade the LLVM in the bootstraptools currently, relevant information linked from [this PR].

thefloweringash:
> @thefloweringash fno-common migration is merged. Does that mean we can upgrade llvm to 12 (or higher?) for aarch64-darwin?

I don't know of any significant blockers.

[...]

PR #313539, opened 2024-05-22
Add CPU model support to target platform

### LINKED ISSUE #313530: Add CPU model to target platform
## Purpose

To specify CPU level optimizations for the package instance.

## Why?

Some users may want to compile packages for their specific CPU model instead of for the generic model on their platform. This could have the added benefit of extra performance by using CPU instructions for their specific CPU instead of generic instructions.

## How?

Inside nixpkgs, we could update `targetPlatform` to include `cpuModel` or add to `parsed.cpu` a `model` attribute which defaults to `"generic"`.  Furthermore, we can change `crossSystem.config` to allow for a CPU model. An example of an acceptable value could be `aarch64-unknown-linux-gnu-apple-m1`. This could be implemented into CC wrapper so we don't have to always specify the CPU model to the compiler. I know clang has this support built in but I am not sure about GCC.

RossComputerGuy:
## Description of changes

Adds a `cpuModel` attribute to `targetPlatform` and defaults the CPU model to `"generic"`. Might need a lot of work but I think this is a great start. We might want to check the CPU model based on the CPU arch coming in and have a fallback of only `"generic"` when we cannot match a recognized CPU arch.

Fixes #313530

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---


Ericson2314:
This seems close to https://github.com/NixOS/nixpkgs/blob/master/lib/systems/architectures.nix ? My thought at the time with that is that I don't really know what the difference between an instruction set architecture and a "sub instruction set architecture", except for rather arbitrary conventions about what goes in the target triple or not, which I view as sort of historical accident we shouldn't take too seriously.

RossComputerGuy:
> This seems close to https://github.com/NixOS/nixpkgs/blob/master/lib/systems/architectures.nix ? My thought at the time with that is that I don't really know what the difference between an instruction set architecture and a "sub instruction set architecture", except for rather arbitrary conventions about what goes in the target triple or not, which I view as sort of historical accident we shouldn't take too seriously.

Yeah, though this mechanism is simpler. I'm not sure how that method works in Nixpkgs but this way you just add your CPU model to the end of config. It also allows for easy checking of common CPU models so marking certain packages as broken can be simpler. Adding this into the config stuff also allows for easy cross compiling via `systems.examples`.

RossComputerGuy:
Idk what is up with `by-name` failing. I cannot reproduce that error with the information given.

RossComputerGuy:
Just cleaned up the commits, fixed some of the issues found with CI. Made it so this doesn't rebuild everything if the CPU model is generic. I think this is at a state that I am happy.

RossComputerGuy:
@ofborg build pkgsCross.asahi-m1.bash

alyssais:
> > This seems close to https://github.com/NixOS/nixpkgs/blob/master/lib/systems/architectures.nix ? My thought at the time with that is that I don't really know what the difference between an instruction set architecture and a "sub instruction set architecture", except for rather arbitrary conventions about what goes in the target triple or not, which I view as sort of historical accident we shouldn't take too seriously.
> 
> Yeah, though this mechanism is simpler.

Does this allow us to get rid of architectures.nix?  It's really annoying to have to extend that list with boilerplate when adding new architectures.

Ericson2314:
> Does this allow us to get rid of `architectures.nix`? It's really annoying to have to extend that list with boilerplate when adding new architectures.

Yeah bottom line is I want a unified way of doing things. If we are doing this, then I think it's a hard requirement that is replaces `architecture.nix`. Having multiple ways of specifying things was exactly the sort of thing that the big `lib/systems` reorg I did was supposed to solve --- I don't want it creeping back.

RossComputerGuy:
> > > This seems close to https://github.com/NixOS/nixpkgs/blob/master/lib/systems/architectures.nix ? My thought at the time with that is that I don't really know what the difference between an instruction set architecture and a "sub instruction set architecture", except for rather arbitrary conventions about what goes in the target triple or not, which I view as sort of historical accident we shouldn't take too seriously.
> > 
> > 
> > Yeah, though this mechanism is simpler.
> 
> Does this allow us to get rid of architectures.nix? It's really annoying to have to extend that list with boilerplate when adding new architectures.

Yes. This is a simpler alternative so we probably could deprecate it and remove it. This method only requires a couple lines to be changed and models which aren't added to the list could just have all the info specified as an attribute set instead of a string.

pkgs/build-support/cc-wrapper/default.nix
```diff
@@ -634,6 +634,12 @@ stdenvNoCC.mkDerivation {
     + optionalString ((targetPlatform ? gcc.cpu) && (isClang || !(targetPlatform.isDarwin && targetPlatform.isAarch64))) ''
       echo "-mcpu=${targetPlatform.gcc.cpu}" >> $out/nix-support/cc-cflags-before
```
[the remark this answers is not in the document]
Ericson2314: Yes it should replace it too, ideally.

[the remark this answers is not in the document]
RossComputerGuy: Yes, I just didn't strip out the way things are done with `gcc.tune` because I didn't want to break too much. My initial goal was to just get this mechanism to work.

pkgs/development/compilers/zig/generic.nix
```diff
@@ -64,6 +65,9 @@ stdenv.mkDerivation (finalAttrs: {
   };
 
   meta = {
+    # Zig does not support Apple Silicon on Linux yet.
+    # Ref: https://github.com/ziglang/zig/issues/16331
+    broken = stdenv.isLinux && targetPlatform.isAppleSilicon;
```
alyssais, line 70: This is a conflation I don't think we should be making — whether Zig is broken doesn't depend on whether we're specifically building for Apple Silicon.  The package is equally broken on Apple Silicon when built with the generic CPU model.

RossComputerGuy: That is true but I don't see how we could determine the package is broken and the model is generic but being built on Apple Silicon. I mostly did it as a placeholder for writing a patch that makes Zig working on Apple Silicon and to test the `isAppleSilicon` attribute.

RossComputerGuy:
Tried getting ofborg to build bash for Apple Silicon M1 Linux but it seems that bash didn't like the compiler:
```
checking for aarch64-unknown-linux-gnu-gcc... aarch64-unknown-linux-gnu-gcc
checking whether the C compiler works... no
configure: error: in `/build/bash-5.2':
configure: error: C compiler cannot create executables
See `config.log' for more details
```

pkgs/development/compilers/zig/generic.nix
alyssais: But such a patch needs to be applied for all aarch64-linux — we can't force users to specify their CPU and rebuild the world just to have programs that work on their CPU.

alyssais: Can't we just always build Zig on aarch64 for a target with 64K pages, then?

RossComputerGuy: Maybe but who knows what other issues that could bring

RossComputerGuy: Yeah though this feels like it kinda gets off topic on this particular PR. Could raise an issue with the maintainers and get their thoughts.

alyssais: I agree, but I'm bringing it up because I don't think that isAppleSilicon has really been justified.  Simplifying specifying CPU-specific compiler flags seems good, but I don't think it's ever really going to be the right thing to have conditionals based on them.

pkgs/build-support/cc-wrapper/default.nix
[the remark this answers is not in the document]
Ericson2314: Unlike getting rid of `architecture.nix`, I am fine saving that for later :).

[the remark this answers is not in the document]
RossComputerGuy: Yeah, that was my plan. Maybe add a couple warnings that the file is up for removal so that we can give people a buffer

pkgs/development/compilers/zig/generic.nix
RossComputerGuy: So should we remove it from this PR or just keep it as is until a follow up PR is made to actually fix Zig?

RossComputerGuy:
@ofborg build pkgsCross.asahi-m1.pkgsLLVM.bash

RossComputerGuy:
Interesting quick look at the optimizations this can produce. In this example, I used bash.

Standard - 1.1MB
LLVM - 971KB
LLVM w/ CPU model optimizations - 970KB

We save about 1KB of memory on Apple Silicon for bash.

pkgs/development/compilers/zig/generic.nix
alyssais: We should remove it from this PR.  There should be no predicates for CPU type.

(Another reason I thought of overnight: Apple Silicon doesn't even necessarily imply 16K pages! You can run a 4K kernel, although lots of stuff currently won't work, it's not some fundamental limitation of the CPU.)

RossComputerGuy: Done

alyssais: No it isn't?  The CPU type predicates are still there.

RossComputerGuy: Where? I don't see it in the Zig files.

alyssais: Ah, when I say that "there should be no predicates for CPU type", I meant that they shouldn't exist, not just that they shouldn't be used in the Zig package.  As I said, I would be surprised if having conditionals based on target CPU (as opposed to passing it to a compiler, etc.) is ever the right thing to do.

RossComputerGuy: Alright, done.

lib/systems/parse.nix
```diff
@@ -421,6 +421,61 @@ rec {
...
+      };
+      apple_m2 = mkCpuModel {
+        llvm = "apple-m2";
+        zig = "apple_m2";
+      };
+      cortex_a53 = mkCpuModel {
+        llvm = "cortex-a53";
+        gnu = "cortex-a53";
+        zig = "cortex_a53";
+      };
```
alyssais, line 464: So far all of these follow the same pattern, of being the same apart from underscores vs hyphens.  Are there counterexamples?

RossComputerGuy: That's true though in the case of Apple Silicon, the model isn't supported by GCC yet. Should I make a simple wrapper which generates the names from a single name?

RossComputerGuy: Might be able to get rid of the `zig` one and just replace `-` with `_` at least.

RossComputerGuy: I made some adjustments so now the function to make the CPU model definition simply can be used for most systems. I decided to keep `zig` in because of `baseline`.

lib/systems/default.nix
```diff
@@ -82,6 +86,13 @@ let
       isCompatible = _: throw "2022-05-23: isCompatible has been removed in favor of canExecute, refer to the 22.11 changelog for details";
       # Derived meta-data
       useLLVM = final.isFreeBSD;
+      cpuModel = let
+        name = if args ? cpuModel then args.cpuModel else "generic";
```
yunfachi, line 90:
```suggestion
        name = args.cpuModel or "generic";
```
can be simplified with `or`

RossComputerGuy: Done

RossComputerGuy:
Rebased due to conflicts.

RossComputerGuy:
Going to rebase this tonight to work with Zig 0.13 since that PR was merged yesterday. Gentle ping to see what progress we can get.

alyssais:
> Gentle ping to see what progress we can get.

AFAICT this still hasn't been addressed:

> If we are doing this, then I think it's a hard requirement that is replaces `architecture.nix`.

RossComputerGuy:
@alyssais My understanding with that was to deprecate it here and removed it in a follow up PR.

alyssais:
I don't want to see us with _three_ different ways to configure this — two is already too many, and if we put this in without getting rid of (or at least properly deprecating, with a plan to remove — e.g. by implementing the old interface on top of the new one), there's no guarantee we won't end up stuck with three.

(Personally I feel it'd be much easier to get rid of the `gcc` system options than architecture.nix, because they could easily be reimplemented on top of this, but I don't really care which one goes, as long as we don't end up with three.)

RossComputerGuy:
Alright, I can try working on stripping the `architectures.nix` out tonight or whenever I have time next.

Ericson2314:
Yes, removing `architectures.nix` is a requirement for me too

RossComputerGuy:
@Ericson2314 Should that be included with this PR or a separate one?

RossComputerGuy:
I've replaced what I can find using `gcc.arch` and changed a few things around. `lib.systems.architecture` throws an error and I think we can remove it after the next NixOS release. The error message is in place to let people know about the change.

lib/systems/platforms.nix
```diff
@@ -385,16 +403,15 @@ rec {
...
+      zig = "armv8-a";
     };
   };
 
   apple-m1 = {
-    gcc = {
-      arch = "armv8.3-a+crypto+sha2+aes+crc+fp16+lse+simd+ras+rdm+rcpc";
-      cpu = "apple-a13";
-    };
+    cpuModel = "apple_m1";
```
reckenrode, line 414: This should fix https://github.com/NixOS/nixpkgs/issues/166401? The reason for the lengthy arch string was due to LLVM 11, which isn’t even in nixpkgs anymore.

RossComputerGuy: Likely, needs testing though.

RossComputerGuy:
Just added the ability to detect which compiler versions are being used and adjust some stuff accordingly.

RossComputerGuy:
Will fix conflicts tonight

lib/systems/default.nix
```diff
@@ -25,7 +25,24 @@ let
   inspect = import ./inspect.nix { inherit lib; };
   platforms = import ./platforms.nix { inherit lib; };
   examples = import ./examples.nix { inherit lib; };
-  architectures = import ./architectures.nix { inherit lib; };
+
+  architectures = throw "lib.systems.architectures was deprecated and replaced via the new cpuModel attribute";
```
AndersonTorres, line 29: Deprecated has a specific meaning of "still works but we will remove in the future".
```suggestion
  architectures = throw "architectures was replaced by `cpuModel` attribute";
```

RossComputerGuy: Alright, changed.

RossComputerGuy:
Closing in favor of #376197

PR #313729, opened 2024-05-22
gobang: unbreak, modernize

pluiedev:
## Description of changes

Bumped the sqlx version used in order to use num-bigint 0.3.3+, which fixes an incompatibility with Rust >1.77 — very similar to the patch used in #309507.

Also changed src to use an SRI hash, reformatted with `nixfmt-rfc-style` and moved to `pkgs/by-name` because why not

Part of ZHF #309482

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---


risicle:
Could you better describe the steps you took to bump sqlx? The problem is this patch file is effectively unreviewable at face value.

pluiedev:
> Could you better describe the steps you took to bump sqlx? The problem is this patch file is effectively unreviewable at face value.

Well, I just updated Cargo.toml to use sqlx 0.5.13 instead of 0.5.6, and ran `cargo check` to rebuild the lock file

pkgs/by-name/go/gobang/package.nix
```diff
@@ -0,0 +1,38 @@
...
+  darwin,
+}:
+
+rustPlatform.buildRustPackage rec {
+  pname = "gobang";
+  version = "0.1.0-alpha.5";
+
+  src = fetchFromGitHub {
+    owner = "tako8ki";
+    repo = pname;
```
isabelroses, line 15: `repo = pname` is an anti pattern, see here https://github.com/NixOS/nixpkgs/issues/277994

pkgs/by-name/go/gobang/package.nix
```diff
@@ -0,0 +1,38 @@
+{
+  lib,
+  rustPlatform,
+  fetchFromGitHub,
+  stdenv,
+  darwin,
+}:
+
+rustPlatform.buildRustPackage rec {
```
isabelroses, line 9: Consider removing the `rec` for the `let in` syntax see https://nix.dev/guides/best-practices.html

pluiedev:
> > Could you better describe the steps you took to bump sqlx? The problem is this patch file is effectively unreviewable at face value.
> 
> Well, I just updated Cargo.toml to use sqlx 0.5.13 instead of 0.5.6, and ran `cargo check` to rebuild the lock file. I apologize that the diff is this big but apparently the sqlx people change a lot of their deps on minor version bumps 😅 I'll try to find a version that produces a smaller diff

I tried to look for a version of sqlx that produces a smaller diff, but apparently they bumped almost everything in version 0.5.8... Not much we can do then since Cargo doesn't support just manually overriding the version of a transitive dependency like that (the `[patch]` section does not work with crates fulled from crates.io)

risicle (approved):
Can confirm the `sqlx` bump produces the same result for me.

`nixpkgs-review` happy, macos 12 x86_64 & nixos x86_64

PR #313760, opened 2024-05-22
bun: add baseline release

### LINKED ISSUE #298612: Packaging request: Bun on non-AVX/AVX2 architectures
### Describe the bug
The current build for bun is optimized for AVX/AVX2, and if it's run on an architecture which doesn't support these instructions, bun crashes.

### Steps To Reproduce
1. Install bun on an old system
2. Run it

### Expected behavior
Bun should be able to run also on non-AVX/AVX2 systems

### Additional context
See also #215517 and #202331

One could either
- compile the `bun` package like the `bun-baseline`
- add a new package to nixpkgs with the name `bun-baseline`

### Notify maintainers

@DAlperin @jk @thilobillerbeck @cdmistman @coffeeispower

---

x3lfyn:
## Description of changes

regular bun release works only on CPUs that support AVX instructions. bun has separate baseline releases that don't require AVX instructions

added `bunBaseline` package which uses baseline release(fixes https://github.com/NixOS/nixpkgs/issues/298612)

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---


cdmistman (approved):
nice, though tbh i'm not a fan of polluting the packages namespace like this. mind if we make this `passthru.baseline` instead?

x3lfyn:
> nice, though tbh i'm not a fan of polluting the packages namespace like this. mind if we make this `passthru.baseline` instead?

what do you mean? combine two sources into one file and make some flag to switch between them?

sorry, i am pretty new to nix, i don't know all best practices)

pkgs/development/web/bun/bun.nix
```diff
@@ -0,0 +1,87 @@
...
+, unzip
+, installShellFiles
+, openssl
+, writeShellScript
+, curl
+, jq
+, common-updater-scripts
+}:
+
+{ sources, version }:
```
cdmistman, line 14:
```suggestion
{ src, version, baseline ? null }:
```

pkgs/development/web/bun/bun.nix
```diff
@@ -0,0 +1,87 @@
...
+    SHELL="fish" $out/bin/bun completions $completions_dir
+
+    installShellCompletion --name bun \
+      --bash $completions_dir/bun.completion.bash \
+      --zsh $completions_dir/_bun \
+      --fish $completions_dir/bun.fish
+  '';
+
+  passthru = {
+    inherit sources;
```
cdmistman, line 53:
```suggestion
    inherit baseline;
```

cdmistman (commented):
no worries at all! these 2 changes and any other changes needed to get it working :)

ideally we select `src` before calling `buildBun`, and depending on system also add the result of a `buildBun <baseline source>`

pkgs/top-level/all-packages.nix
```diff
@@ -15203,7 +15203,14 @@ with pkgs;
...
 
-  bun = callPackage ../development/web/bun { };
+  inherit
+    ({
+      bun = callPackage ../development/web/bun { };
+      bunBaseline = callPackage ../development/web/bun/baseline.nix { };
+    })
+    bun
+    bunBaseline
+    ;
```
cdmistman, line 15213: oh and in addition to ^ revert this as well

cdmistman: for future reference, the diff you wrote here is also equivalent to:
```suggestion
  bun = callPackage ../development/web/bun { };
  bunBaseline = callPackage ../development/web/bun/baseline.nix { };
```

06kellyjac (commented):
Generally looks good with 1 change.

I probably would lean towards separate attributes in all-pacakges rather than the passthru option as it's more visible in the search.

pkgs/top-level/all-packages.nix
x3lfyn: i've done it to bypass [this](https://github.com/NixOS/nixpkgs/blob/master/.github/workflows/check-by-name.yml) check.  without this, workflow just [failes](https://github.com/NixOS/nixpkgs/actions/runs/9196795760/job/25295724717). or this error can be ignored?

cdmistman: ah yeah, in that case, you should either:
1. move `bun/baseline.nix` to `by-name/bu/bunBaseline/package.nix` as suggested by the error message
  - you can leave `bun/default.nix` where it's at if you'd like
2. OR implement the `passthru` option as i suggested above
  - i'm realizing this would also probably mean setting `bunBaseline = pkgs.bun.passthru.baseline` in `aliases.nix` instead of modifying `all-packages.nix`

@06kellyjac seems to prefer option 1, i personally prefer option 2, i don't think either of us will care which gets implemented so the decision is up to you :)

cdmistman:
also if you rebase off of master, you'll see the hashes are for 1.1.10

x3lfyn:
so i moved `bun` package to `pkgs/by-name/bu/bun/package.nix`. baseline is now passtru of `bun` with `bunBaseline = pkgs.bun.passthru.baseline` in `all-packages.nix`. all that satifsfies check-by-name. if just move `bunBaseline` to `pkgs/by-name/bu/bunBaseline/package.nix`, check-by-name fails with message: 
```
pkgs/by-name/bu/bunBaseline: File package.nix at line 4 contains the path expression "../../../development/web/bun/bun.nix" which may point outside the directory of that package.
This PR introduces the problems listed above. Please fix them before merging, otherwise the base branch would break.
```

cdmistman (approved):
looks great, thank you!

cdmistman (approved):
reapproving

x3lfyn:
sorry to bother you, i understand correctly that we are waiting for the second contributor to complete review?

diogomdp (approved):

nagy:
Before merging though, the two youngest commits should probably be removed. If you rebase this PR, those commits should no longer be necessary.

diogomdp:
Result of `nixpkgs-review pr 313760` run on x86_64-linux [1](https://github.com/Mic92/nixpkgs-review)
<details>
  <ul>
    <li>bunBaseline</li>
  </ul>
</details>

x3lfyn:
hi. regarding [this commit](https://github.com/NixOS/nixpkgs/commit/c5435f8253de8a1cc347b653500a0faecdfde76c) - what decision would be more reasonable?
1) use default version for `x86_64-darwin` system in `bun` package and baseline version in `bunBaseline` package (current state of this PR)
2) use baseline version in `bun` package and exclude `x86_64-darwin` system from `bunBaseline` package

pkgs/top-level/all-packages.nix
Eveeifyeve: I was just about to say to fix this issue do a passthru.

Eveeifyeve:
> hi. regarding [this commit](https://github.com/NixOS/nixpkgs/commit/c5435f8253de8a1cc347b653500a0faecdfde76c) - what decision would be more reasonable?
> 
> 1. use default version for `x86_64-darwin` system in `bun` package and baseline version in `bunBaseline` package (current state of this PR)
> 2. use baseline version in `bun` package and exclude `x86_64-darwin` system from `bunBaseline` package

I would say first one as support as a pr will most likely be made for this version some time in the future. 
but I would say for the baseline use `passthru` as to creating another package will be a lot of maintaining plus it makes it harder to browse in the Nixpkgs repo.

x3lfyn:
> but I would say for the baseline use `passthru` as to creating another package will be a lot of maintaining plus it makes it harder to browse in the Nixpkgs repo.

there are 2 problems:
- new packages can't be created in `top-level/all-packages.nix`, only in `by-name` folders
- expressions that point outside of package's folder are prohibited in `by-name` folders

so i can't just create separate baseline package that shares logic with default package due to this limitations

correct me if i'm wrong and there is some better way to share logic between two packages

Eveeifyeve:
Ops wrong reference.

forsureitsme:
Until this is merged, is there another way to install `bun` for non AVX machines?

x3lfyn:
> Until this is merged, is there another way to install `bun` for non AVX machines?

you can override source archive
```nix
bunBaseline = pkgs.bun.overrideAttrs rec {
  passthru.sources."x86_64-linux" = pkgs.fetchurl {
    url = "https://github.com/oven-sh/bun/releases/download/bun-v1.1.27/bun-linux-x64-baseline.zip";
    hash = "sha256-FwkVP5lb2V9E8YGPkTAqVMsZmaZXMq8x5AR+99cuIX0=";
  };
  src = passthru.sources."x86_64-linux";
};
```

DADA30000:
helloo?

Eveeifyeve (commented):
This is good to me, but however please fix the conflicts I can't approve until the conflicts are fixed.

Baitinq:
I've revived this in #441599

Issue #278348, opened 2024-01-02
CoreFoundation.tbd hook breaks darwin -> non-darwin cross-compilation


r-burns:
The CoreFoundation.tbd setup hook in darwin's CoreFoundation does not take build/host offsets into account, so it unconditionally applies CFLAGS/LDFLAGS which are invalid for non-darwin hosts. This can occur when a package pulls in CoreFoundation via a buildPackage's propagatedBuildInputs. Since CF is propagated by a number of packages, this occurs quite often.

This seems to have technically been an issue for as long as this setup hook has existed, but has only started causing problems now that we are explicitly linking the .tbd file. Previously it was an extraneous directory that could be ignored, but now the explicit .tbd file is seen as an invalid linker script by the target cross linker.

For example, see: [nixpkgs](https://hydra.nixos.org/project/nixpkgs):[cross-trunk](https://hydra.nixos.org/jobset/nixpkgs/cross-trunk):[rpi-musl.nix.x86_64-darwin](https://hydra.nixos.org/job/nixpkgs/cross-trunk/rpi-musl.nix.x86_64-darwin) (https://hydra.nixos.org/build/241448869/nixlog/1):

```
CMake Error at /nix/store/nsk9a663g0khs9li1hwwqlsxsvqd0x9d-cmake-3.27.7/share/cmake-3.27/Modules/CMakeTestCXXCompiler.cmake:60 (message):
  The C++ compiler

    "/nix/store/7hl3605laf3qi72bgzkxkvpg4ziqzlch-armv6l-unknown-linux-musleabihf-gcc-wrapper-12.3.0/bin/armv6l-unknown-linux-musleabihf-g++"

  is not able to compile a simple test program.

  It fails with the following output:

    Change Dir: '/tmp/nix-build-aws-sdk-cpp-armv6l-unknown-linux-musleabihf-1.11.118.drv-0/source/build/CMakeFiles/CMakeScratch/TryCompile-cfogov'
    
    Run Build Command(s): /nix/store/nsk9a663g0khs9li1hwwqlsxsvqd0x9d-cmake-3.27.7/bin/cmake -E env VERBOSE=1 /nix/store/x5dy4sqg1nxsl7nwibanfrbx95wdb228-gnumake-4.4.1/bin/make -f Makefile cmTC_fd3de/fast
[...]

PR #284706, opened 2024-01-29
CoreFoundation: apply hook to correct platform

szlend:
## Description of changes

Cross-compiling from darwin to linux is broken if the derivation includes CoreFoundation somewhere in its (native) input deps. This change makes the cf-setup-hook aware of its platform role, replacing NIX_LDFLAGS with NIX_LDFLAGS_FOR_BUILD when necessary.

Fixes https://github.com/NixOS/nixpkgs/issues/278348 in `23.11` without getting rid of the hook in case anyone was relying on it.

In `unstable` we probably want to get rid of the hook entirely. See #284629.

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [x] x86_64-darwin
  - [x] aarch64-darwin
- For non-Linux: Is sandboxing enabled in `nix.conf`? (See [Nix manual](https://nixos.org/manual/nix/stable/command-ref/conf-file.html))
  - [ ] `sandbox = relaxed`
  - [x] `sandbox = true`
- [ ] Tested, as applicable:
  - [NixOS test(s)](https://nixos.org/manual/nixos/unstable/index.html#sec-nixos-tests) (look inside [nixos/tests](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests))
  - and/or [package tests](https://nixos.org/manual/nixpkgs/unstable/#sec-package-tests)
  - or, for functions and "core" functionality, tests in [lib/tests](https://github.com/NixOS/nixpkgs/blob/master/lib/tests) or [pkgs/test](https://github.com/NixOS/nixpkgs/blob/master/pkgs/test)
  - made sure NixOS tests are [linked](https://nixos.org/manual/nixpkgs/unstable/#ssec-nixos-tests-linking) to the relevant packages
- [ ] Tested compilation of all packages that depend on this change using `nix-shell -p nixpkgs-review --run "nixpkgs-review rev HEAD"`. Note: all changes have to be committed, also see [nixpkgs-review usage](https://github.com/Mic92/nixpkgs-review#usage)
- [ ] Tested basic functionality of all binary files (usually in `./result/bin/`)
[...]

PR #284629, opened 2024-01-28
darwin.apple_sdk.frameworks.CoreFoundation: drop hook

reckenrode:
## Description of changes

As of https://github.com/NixOS/nixpkgs/pull/265102, x86_64-darwin no longer uses the open-source CF release. Since there is no longer a need to switch between implementations, and the hook causes problems with cross-compilation (see https://github.com/NixOS/nixpkgs/issues/278348), drop the hook and make both darwin.CF an alias for darwin.apple_sdk.frameworks.CoreFoundation.

Thank you to @szlend for his work on #284288 and @r-burns for identifying the issue in #278348.

Testing was done by confirming the following packages built on aarch64- and x86_64-darwin:

* bacula (requires patches in commit)
* pkgsCross.muslpi.aws-sdk-cpp
* wine64Packages.unstable

I am targeting this for backporting to fix the cross-compilation issue in 23.11. Since the hook is already not present by default, this change will only affect packages that explicitly use the CoreFoundation framework as a build input.

Fixes #278348
Closes #284288

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [x] x86_64-darwin
  - [x] aarch64-darwin
- For non-Linux: Is sandboxing enabled in `nix.conf`? (See [Nix manual](https://nixos.org/manual/nix/stable/command-ref/conf-file.html))
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:
  - [NixOS test(s)](https://nixos.org/manual/nixos/unstable/index.html#sec-nixos-tests) (look inside [nixos/tests](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests))
  - and/or [package tests](https://nixos.org/manual/nixpkgs/unstable/#sec-package-tests)
  - or, for functions and "core" functionality, tests in [lib/tests](https://github.com/NixOS/nixpkgs/blob/master/lib/tests) or [pkgs/test](https://github.com/NixOS/nixpkgs/blob/master/pkgs/test)
  - made sure NixOS tests are [linked](https://nixos.org/manual/nixpkgs/unstable/#ssec-nixos-tests-linking) to the relevant packages
- [ ] Tested compilation of all packages that depend on this change using `nix-shell -p nixpkgs-review --run "nixpkgs-review rev HEAD"`. Note: all changes have to be committed, also see [nixpkgs-review usage](https://github.com/Mic92/nixpkgs-review#usage)
[...]

PR #313773, opened 2024-05-22
CoreFoundation: apply hook to correct platform

szlend:
## Description of changes

> Cross-compiling from darwin to linux is broken if the derivation includes CoreFoundation somewhere in its (native) input deps. This change makes the cf-setup-hook aware of its platform role, replacing NIX_LDFLAGS with NIX_LDFLAGS_FOR_BUILD when necessary.
>
> Fixes https://github.com/NixOS/nixpkgs/issues/278348 in `23.11` without getting rid of the hook in case anyone was relying on it.
>
> In `unstable` we probably want to get rid of the hook entirely. See #284629.

This was already merged into `23.11` a few months back as #284706, with the plan to remove the CoreFoundation hook entirely by `24.05` in #284629.

Unfortunately #284629 didn't make it into 24.05, so this is now a regression. This PR is cherry-picked from `23.11`.

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/os-specific/darwin/apple-sdk-11.0/cf-setup-hook.sh
+++ b/pkgs/os-specific/darwin/apple-sdk-11.0/cf-setup-hook.sh
@@ -1,6 +1,14 @@
 forceLinkCoreFoundationFramework() {
-  NIX_CFLAGS_COMPILE="-F@out@/Library/Frameworks${NIX_CFLAGS_COMPILE:+ }${NIX_CFLAGS_COMPILE-}"
-  NIX_LDFLAGS+=" @out@/Library/Frameworks/CoreFoundation.framework/CoreFoundation.tbd"
+  local role_post
+  getHostRole
+
+  local cflags_compile_role_var="NIX_CFLAGS_COMPILE${role_post}"
+  local cflags_compile_role="${!cflags_compile_role_var-}"
+  export NIX_CFLAGS_COMPILE${role_post}="-F@out@/Library/Frameworks${cflags_compile_role:+ $cflags_compile_role}"
+
+  local ldflags_role_var="NIX_LDFLAGS${role_post}"
+  local ldflags_role="${!ldflags_role_var-}"
+  export NIX_LDFLAGS${role_post}+="${ldflags_role:+ }@out@/Library/Frameworks/CoreFoundation.framework/CoreFoundation.tbd"
 }
 
-preConfigureHooks+=(forceLinkCoreFoundationFramework)
+forceLinkCoreFoundationFramework
--- a/pkgs/os-specific/darwin/apple-sdk/cf-setup-hook.sh
+++ b/pkgs/os-specific/darwin/apple-sdk/cf-setup-hook.sh
@@ -1,9 +1,17 @@
 linkSystemCoreFoundationFramework() {
-  NIX_CFLAGS_COMPILE="-F@out@/Library/Frameworks${NIX_CFLAGS_COMPILE:+ }${NIX_CFLAGS_COMPILE-}"
+  local role_post
+  getHostRole
+
+  local cflags_compile_role_var="NIX_CFLAGS_COMPILE${role_post}"
+  local cflags_compile_role="${!cflags_compile_role_var-}"
+  export NIX_CFLAGS_COMPILE${role_post}="-F@out@/Library/Frameworks${cflags_compile_role:+ $cflags_compile_role}"
+
   # gross! many symbols (such as _OBJC_CLASS_$_NSArray) are defined in system CF, but not
   # in the opensource release
   # if the package needs private headers, we assume they also want to link with system CF
-  NIX_LDFLAGS+=" @out@/Library/Frameworks/CoreFoundation.framework/CoreFoundation.tbd"
+  local ldflags_role_var="NIX_LDFLAGS${role_post}"
+  local ldflags_role="${!ldflags_role_var-}"
+  export NIX_LDFLAGS${role_post}+="${ldflags_role:+ }@out@/Library/Frameworks/CoreFoundation.framework/CoreFoundation.tbd"
 }
 
-preConfigureHooks+=(linkSystemCoreFoundationFramework)
+linkSystemCoreFoundationFramework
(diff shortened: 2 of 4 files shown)


pkgs/os-specific/darwin/apple-sdk-11.0/cf-setup-hook.sh
```diff
@@ -1,6 +1,14 @@
...
+  local role_post
+  getHostRole
+
+  local cflags_compile_role_var="NIX_CFLAGS_COMPILE${role_post}"
+  local cflags_compile_role="${!cflags_compile_role_var-}"
+  export NIX_CFLAGS_COMPILE${role_post}="-F@out@/Library/Frameworks${cflags_compile_role:+ $cflags_compile_role}"
+
+  local ldflags_role_var="NIX_LDFLAGS${role_post}"
+  local ldflags_role="${!ldflags_role_var-}"
+  export NIX_LDFLAGS${role_post}+="${ldflags_role:+ }@out@/Library/Frameworks/CoreFoundation.framework/CoreFoundation.tbd"
```
toonn, line 11: Aren't we appending here already due to the `+=` assignment?
I think this'll end up duplicating the contents of `NIX_LDFLAGS${role_post}`.

pkgs/os-specific/darwin/apple-sdk-11.0/cf-setup-hook.sh
```diff
@@ -1,6 +1,14 @@
...
 }
 
-preConfigureHooks+=(forceLinkCoreFoundationFramework)
```
toonn, line 6: You mentioned somewhere this change was necessary to keep `role_post` in scope?

When do hooks run without such assignment and might we want to just pass the value in to keep it in scope? `preConfigureHooks+=(forceLinkCoreFoundationFramework getHostRole)` or something?

pkgs/os-specific/darwin/apple-sdk/cf-setup-hook.sh
```diff
@@ -1,9 +1,17 @@
...
+  local cflags_compile_role="${!cflags_compile_role_var-}"
+  export NIX_CFLAGS_COMPILE${role_post}="-F@out@/Library/Frameworks${cflags_compile_role:+ $cflags_compile_role}"
+
   # gross! many symbols (such as _OBJC_CLASS_$_NSArray) are defined in system CF, but not
   # in the opensource release
   # if the package needs private headers, we assume they also want to link with system CF
-  NIX_LDFLAGS+=" @out@/Library/Frameworks/CoreFoundation.framework/CoreFoundation.tbd"
+  local ldflags_role_var="NIX_LDFLAGS${role_post}"
+  local ldflags_role="${!ldflags_role_var-}"
+  export NIX_LDFLAGS${role_post}+="${ldflags_role:+ }@out@/Library/Frameworks/CoreFoundation.framework/CoreFoundation.tbd"
```
toonn, line 14: Same comment re append assignment.

pkgs/os-specific/darwin/apple-sdk/cf-setup-hook.sh
```diff
@@ -1,9 +1,17 @@
...
+  local role_post
+  getHostRole
+
+  local cflags_compile_role_var="NIX_CFLAGS_COMPILE${role_post}"
+  local cflags_compile_role="${!cflags_compile_role_var-}"
+  export NIX_CFLAGS_COMPILE${role_post}="-F@out@/Library/Frameworks${cflags_compile_role:+ $cflags_compile_role}"
+
   # gross! many symbols (such as _OBJC_CLASS_$_NSArray) are defined in system CF, but not
   # in the opensource release
   # if the package needs private headers, we assume they also want to link with system CF
```
toonn, line 11: Is the CF aliasing already a thing? Then we might be able to drop this comment.

toonn (commented):
Any particularly suitable targets to test build?

szlend:
Hey @toonn, thanks for the detailed review. Happy to answer your questions, but just in case you missed my comment, this was already reviewed in #284706. I'm just porting it over from `23.11` to `24.05` because the CF hook removal that was planned for `24.05` didn't make it in (#284629), so this is kind of a regression.

szlend:
> Any particularly suitable targets to test build?

This failed when I originally fixed it for `23.11`:
```
$ nix build -f . pkgsCross.aarch64-multiplatform.aws-sdk-cpp --sandbox --print-out-paths
```

pkgs/os-specific/darwin/apple-sdk-11.0/cf-setup-hook.sh
szlend: `${ldflags_role:+ }` just adds a space. I guess it would have been much clearer to always prepend a space.

pkgs/os-specific/darwin/apple-sdk/cf-setup-hook.sh
szlend: I'm not actually sure, but it will be dropped in #284629. Imo out of scope for this PR.

pkgs/os-specific/darwin/apple-sdk-11.0/cf-setup-hook.sh
szlend: I'm not sure if that's actually possible. In the snippet above "forceLinkCoreFoundationFramework" and "getHostRole" are literal strings appended to an array.

Though I can't think of any reason why `forceLinkCoreFoundationFramework` would need to run specifically in the `configure` phase. All it does is append compiler/linker flags. These flags can be set at any point before the build phase. They could even be set in the `.drv` itself.

pkgs/os-specific/darwin/apple-sdk-11.0/cf-setup-hook.sh
toonn: Ah, my bad, did not squint hard enough at the parameter expansion. This is fine.

toonn (approved):
LGTM and solves the cross build for aws-sdk-cpp as promised.

PR #229439, opened 2023-05-02
musl: 1.2.3 -> 1.2.5

alyssais:
From a packaging perspective, this is the most important thing about this upgrade:

> Users who build applications from source against musl should be aware: 1.2.4 deprecates the "LFS64" interfaces, which were provided via macros (causing lots of frustration by breaking things; thus the motivation to get rid of them).
>
> Builds which break can be fixed for now by adding -D_LARGEFILE64_SOURCE to CFLAGS, but this too will be dropped in a later version.

[Source](https://fosstodon.org/@musl/110297136732196345)

I think we should see which packages are affected, and give it some reasonable amount of time for them to be fixed, because I don't think it's going to always be as simple as modifying CFLAGS — what do we do for rustc, for example?

We also have some work ahead of us to even identify broken packages.  I've done some test builds of the Spectrum live image, which heavily uses musl, to identify a few regressions, but since we don't have Hydra for musl, we're going to need people who use musl in Nixpkgs to do their own test builds to check whether any of the packages they care about are affected.  If you do so, add any more broken packages to the list in this post (or comment if you lack permissions).

- [x] LLVM #229442
- [x] libselinux https://github.com/NixOS/nixpkgs/pull/287471
- [x] acl #229440
- [x] xorg.libpciaccess #229441
- [x] rustc (and cross-compiled programs that use non-musl build rustc) #259989
- [x] gperftools https://github.com/NixOS/nixpkgs/pull/253556
- [x] gdb #229673
- [x] doxygen https://github.com/NixOS/nixpkgs/pull/292630
- [x] libredirect https://github.com/NixOS/nixpkgs/pull/292616
- [x] numactl https://github.com/NixOS/nixpkgs/pull/281724
- [x] autogen https://github.com/NixOS/nixpkgs/pull/292617
- [x] libeatmydata https://github.com/stewartsmith/libeatmydata/pull/36 https://github.com/NixOS/nixpkgs/pull/324772
- [x] bpftools
- [x] libbpf
- [x] gst_all_1.gstreamer
- [x] meson https://github.com/NixOS/nixpkgs/pull/298047
- [x] elfutils https://github.com/NixOS/nixpkgs/pull/293517
- [x] yara https://github.com/NixOS/nixpkgs/pull/306531
- [x] directfb https://github.com/NixOS/nixpkgs/pull/309583
- [x] systemd #311995
[...]

PR #313836, opened 2024-05-23
pkgsMusl.nginx: fix build

superherointj:
pkgsMusl.nginx: fix build

Fixes:

> error: #warning usage of non-standard #include <sys/cdefs.h> is deprecated [-Werror=cpp]

CC Musl: @yu-re-ka @alyssais @ulrikstrid
CC Nginx: @fpletz @raitobezarius @helsinki-systems

### DIFF
--- a/pkgs/servers/http/nginx/generic.nix
+++ b/pkgs/servers/http/nginx/generic.nix
@@ -130,6 +130,9 @@ stdenv.mkDerivation {
     "-Wno-error=deprecated-declarations"
     "-Wno-error=gnu-folding-constant"
     "-Wno-error=unused-but-set-variable"
+  ] ++ lib.optionals stdenv.hostPlatform.isMusl [
+    # fix sys/cdefs.h is deprecated
+    "-Wno-error=cpp"
   ]);
 
   configurePlatforms = [];


alyssais (approved):
We should strongly consider just ditching these compatibility headers and fixing stuff properly upstream, because I have a feeling they cause more build failures (like this one) than they prevent.

But let's worry about that after #229439.

alyssais:
@OfBorg build nginx pkgsMusl.nginx

fpletz (approved):

SuperSandro2000:
I think a very similar fix was already proposed half a year back in https://github.com/NixOS/nixpkgs/pull/263147 which was not accepted. Glad to see that we choose a churn friendly solution 😄

superherointj:
> Glad to see that we choose a churn friendly solution 😄

Hi Sandro,
As always, nice to see you here.
I'm not sure what you mean by that. I didn't have a better solution at the time.
I proposed what I found to be working. I understand that it might not be a perfect solution.
But I don't have perfection information. I don't really know C/C++ well [haven't programmed in that]. So I'm not comfortable in fixing at upstream level.
Also, time is limited, and there's priorities too. It's just too much to be done.
There is also an information imbalance between different people. Sometimes people say something like that is obvious, while that isn't for others.
Anyway, if someone has a better solution, I'm all for it.
Thanks for your understanding.

Mic92:
> > Glad to see that we choose a churn friendly solution 😄
> 
> Hi Sandro, As always, nice to see you here. Thanks for the reference. I'm not sure what you mean by that. I didn't have a better solution at the time. I proposed what I found to be working. I understand that it might not be a perfect solution. But I don't have perfection information. I don't really know C/C++ well [haven't programmed in that]. So I'm not comfortable in fixing at upstream level. Also, time is limited, and there's priorities too. It's just too much to be done. There is also an information imbalance between different people. Sometimes people say something like that is obvious, while that isn't for others. Anyway, if someone has a better solution, I'm all for it. Thanks for your understanding.

I don't think he meant to criticize your approach, but instead was a bit disgruntled that his pull request was rejected. I would have probably fixed it in the same way.

SuperSandro2000:
> But I don't have perfection information. I don't really know C/C++ well [haven't programmed in that]. So I'm not comfortable in fixing at upstream level.
Also, time is limited, and there's priorities too. It's just too much to be done.

Same, same for me. I totally didn't want to critique you and just say that I had the basically the same mindset a while ago and agree with the sentiment in this thread and that I am glad that I am not alone with this opinion.

Ma27:
@SuperSandro2000 can you elaborate please? Seems like the PR was closed by the original author and it was your suggestion in review that got rejected.

SuperSandro2000:
I think the original author and me had the same idea, to make musl maintenance easy and reduce churn with every update and compiler change but it got declined by the maintainer for reasons I do not follow. But everything alright, I don't think we need to change anything here codewise.

PR #313893, opened 2024-05-23
tzdata: add libintl dependency on FreeBSD

rhelmot:
## Description of changes

tzdata doesn't have a configure step. It instead has instructions to edit the make flags until it works. The makefile actually contains the explicit example of FreeBSD as a platform you have to explicitly add -lintl on!

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [x] x86_64-freebsd
  - [ ] `sandbox = relaxed`
  - [x] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/data/misc/tzdata/default.nix
+++ b/pkgs/data/misc/tzdata/default.nix
@@ -1,4 +1,4 @@
-{ lib, stdenv, fetchurl, buildPackages }:
+{ lib, stdenv, fetchurl, buildPackages, libintl }:
 
 stdenv.mkDerivation (finalAttrs: {
   pname = "tzdata";
@@ -23,6 +23,7 @@ stdenv.mkDerivation (finalAttrs: {
 
   outputs = [ "out" "bin" "man" "dev" ];
   propagatedBuildOutputs = [ ];
+  buildInputs = lib.optionals stdenv.hostPlatform.isFreeBSD [ libintl ];
 
   makeFlags = [
     "TOPDIR=${placeholder "out"}"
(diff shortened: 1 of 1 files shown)


wolfgangwalther (approved):
> The makefile actually contains the explicit example of FreeBSD as a platform you have to explicitly add -lintl on!

Not 100%.

The `Makefile` for tzdata contains:

```
# Non-default libraries needed to link.
# On some hosts, this should have -lintl unless CFLAGS has -DHAVE_GETTEXT=0.
LDLIBS=

# Add the following to an uncommented "CFLAGS=" line as needed
# to override defaults specified in the source code or by the system.
[...]
#  -DHAVE_GETTEXT if gettext works (e.g., GNU/Linux, FreeBSD, Solaris),
#	where LDLIBS also needs to contain -lintl on some hosts;
#	-DHAVE_GETTEXT=0 to avoid using gettext
```

I can't judge the "why", i.e. whether FreeBSD is one of "those hosts", but the "how" seems good. But given that it fixed things for you, the "why" seems pretty close.

Ericson2314 (commented):
I think it might be good to have a `libIntlSeparateFromLibc ? stdenv.hostPlatform.libc` and use `libIntlSeparateFromLibc` in the two conditions?

Also, please add the description to the commit message

rhelmot:
libintl is already defined as

```
# On non-GNU systems we need GNU Gettext for libintl.
libintl = if stdenv.hostPlatform.libc != "glibc" then gettext else null;
```

So I guess we can just make the dependency unconditional? Might want to change the default from null to stdenv.cc.libc, if I'm understanding the directions we would like to take for nixpkgs? :)

Ericson2314 (changes requested):
Hmm, I just cross compiled this without any changes, and it built fine. Are we sure we need this?

rhelmot:
EXCELLENT catch! I did some forensics and found that this is only because of the way the FreeBSD bootstrap tools work, and if I override it for just that one stdenv it works fine. 

It turns out that tzdata is doing quite a bit of self-configuration at build time through use of the `__has_include` macro, which I didn't know existed until today!

PR #313982, opened 2024-05-23
nodejs: run JS test suite as part of the checks

aduh95:
## Description of changes

I noticed the JS test suite of Node.js wasn't run as part of the check phase, so I spent some time making it work. I've opened https://github.com/nodejs/node/pull/53105 and https://github.com/nodejs/node/pull/53104 to land the fixes upstream.

I've only the changes with `nix-build -A nodejs-slim_20`. I think there are some improvements that could be made (e.g. currently the list of skip tests is shared for all versions, it would probably make more sense to have per-version specific lists instead).

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [x] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/development/web/nodejs/v18.nix
+++ b/pkgs/development/web/nodejs/v18.nix
@@ -1,4 +1,4 @@
-{ callPackage, lib, overrideCC, pkgs, buildPackages, openssl, python3, enableNpm ? true }:
+{ callPackage, lib, overrideCC, pkgs, buildPackages, openssl, python3, fetchpatch2, enableNpm ? true }:
 
 let
   # Clang 16+ cannot build Node v18 due to -Wenum-constexpr-conversion errors.
@@ -27,5 +27,10 @@ buildNodejs {
     ./revert-arm64-pointer-auth.patch
     ./node-npm-build-npm-package-logic.patch
     ./trap-handler-backport.patch
+    ./use-correct-env-in-tests.patch
+    (fetchpatch2 {
+      url = "https://github.com/nodejs/node/commit/534c122de166cb6464b489f3e6a9a544ceb1c913.patch";
+      hash = "sha256-4q4LFsq4yU1xRwNsM1sJoNVphJCnxaVe2IyL6AeHJ/I=";
+    })
   ];
 }
--- a/pkgs/development/web/nodejs/nodejs.nix
+++ b/pkgs/development/web/nodejs/nodejs.nix
@@ -137,10 +137,16 @@ let
       "build-node-api-tests"
       "tooltest"
       "cctest"
+      "test-ci-js"
     ];
 
-    # Do not create __pycache__ when running tests.
-    checkFlags = [ "PYTHONDONTWRITEBYTECODE=1" ];
+    checkFlags = [
+      # Do not create __pycache__ when running tests.
+      "PYTHONDONTWRITEBYTECODE=1"
+      "FLAKY_TESTS=skip"
+      # Skip some tests that are not passing in this context
+      "CI_SKIP_TESTS=test-setproctitle,test-tls-cli-max-version-1.3,test-tls-client-auth,test-child-process-exec-env,test-fs-write-stream-eagain,test-tls-sni-option,test-https-foafssl,test-child-process-uid-gid,test-process-euid-egid,test-process-initgroups,test-process-uid-gid,test-process-setgroups"
+    ];
 
     postInstall = ''
       HOST_PATH=$out/bin patchShebangs --host $out
--- a/pkgs/development/web/nodejs/use-correct-env-in-tests.patch
+++ b/pkgs/development/web/nodejs/use-correct-env-in-tests.patch
@@ -1,0 +1,24 @@
+`/usr/bin/env` is not available.
+
+--- old/test/common/assertSnapshot.js
++++ new/test/common/assertSnapshot.js
+@@ -81,2 +81,2 @@ async function spawnAndAssert(filename, transform = (x) => x, { tty = false, ...
+-  const executable = tty ? 'tools/pseudo-tty.py' : process.execPath;
+-  const args = tty ? [process.execPath, ...flags, filename] : [...flags, filename];
++  const executable = tty ? 'python3' : process.execPath;
++  const args = tty ? ['tools/pseudo-tty.py', process.execPath, ...flags, filename] : [...flags, filename];
+--- old/test/parallel/test-child-process-default-options.js
++++ new/test/parallel/test-child-process-default-options.js
+@@ -35 +35 @@ if (isWindows) {
+-  child = spawn('/usr/bin/env', [], {});
++  child = spawn('env', [], {});
+--- old/test/parallel/test-child-process-env.js
++++ new/test/parallel/test-child-process-env.js
+@@ -51 +51 @@ if (isWindows) {
+-  child = spawn('/usr/bin/env', [], { env });
++  child = spawn('env', [], { env });
+--- old/test/parallel/test-child-process-exec-env.js
++++ new/test/parallel/test-child-process-exec-env.js
+@@ -47 +47 @@ if (!isWindows) {
+-  child = exec('/usr/bin/env', { env: { 'HELLO': 'WORLD' } }, after);
++  child = exec('env', { env: { 'HELLO': 'WORLD' } }, after);
--- a/pkgs/development/web/nodejs/v20.nix
+++ b/pkgs/development/web/nodejs/v20.nix
@@ -1,4 +1,4 @@
-{ callPackage, openssl, python3, enableNpm ? true }:
+{ callPackage, fetchpatch2, openssl, python3, enableNpm ? true }:
 
 let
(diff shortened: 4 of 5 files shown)


lheckemann:
Diff looks good, but this should go to `staging` rather than straight to master since it causes a lot of rebuilds. See the [contributing docs section](https://github.com/NixOS/nixpkgs/blob/ac5cb0b650c56e0ca4e7427d5db221c11bd7190c/CONTRIBUTING.md#rebasing-between-branches-ie-from-master-to-staging) for how (just changing the target branch will notify a bunch of unrelated people due to github weirdness).

The commit metadata also seems wrong, check your Git `user.name` and `user.email` and use `git commit --amend --reset-author` to fix that.

lheckemann:
@ofborg build nodejs_18 nodejs_20 nodejs_22

lheckemann:
Looks like you rebased on staging (as opposed to the merge base of staging) before changing the target branch -- the rebase on staging shouldn't happen until _after_ the target branch change -- but thankfully this didn't result in a huge amount of notifications so I think we can keep this PR rather than opening a new one (sorry to the unrelated requested reviewers!).

lheckemann (approved):
Looks good, will merge if the builds pass

lheckemann:
The build seems to be failing on ofborg: https://logs.ofborg.org/?key=nixos/nixpkgs.313982&attempt_id=fdb75382-9baf-43dc-9dfa-990f53fc001f maybe this is an issue with the non-slim build?

aduh95:
There are some tests that try to spawn `/usr/bin/env`; is there a better solution that just skipping those tests?



> Looks like you rebased on staging (as opposed to the merge base of staging) before changing the target branch -- the rebase on staging shouldn't happen until _after_ the target branch change -- but thankfully this didn't result in a huge amount of notifications so I think we can keep this PR rather than opening a new one (sorry to the unrelated requested reviewers!).

Sorry for the unwelcome pings, I should have read the docs more carefully.

lheckemann:
No worries, this github bug is basically a rite of passage and you got off easy with only a few pings :D

The only fix I can think of for the /usr/bin/env reliance is patching the tests to refer to `${coreutils}/bin/env`, since we can't really control the contents of the sandbox purely.

Darwin build is timing out due to an LLVM build, I'll throw it at my mac.

aduh95:
> The only fix I can think of for the /usr/bin/env reliance is patching the tests to refer to `${coreutils}/bin/env`, since we can't really control the contents of the sandbox purely.

Is `coreutils` available on the environment? Or can I make it available as an env variable?

lheckemann:
It's part of stdenv, so you should be able to use it from PATH (i.e. patch in the output of `type -p env`).

pkgs/development/web/nodejs/v18.nix
```diff
@@ -27,5 +27,14 @@ buildNodejs {
...
     ./trap-handler-backport.patch
+    ./use-correct-env-in-tests.patch
+    (fetchpatch2 {
+      url = "https://github.com/nodejs/node/commit/e818eee5afa2b6c902954f2f6e198858c5f3e79c.patch";
+      hash = "sha256-XA4CDWla8Uq/mq9BCXjAW0v9fw5T9YPAZUlZiYK2Idc=";
+    })
+    (fetchpatch2 {
+      url = "https://github.com/nodejs/node/pull/53105.patch";
+      hash = "sha256-K0EttzKpcPf3pKm3HVqn9xpPlKcNd3rkQSSgotDcSjQ=";
+    })
```
aduh95, line 38: We should wait for the PR to land upstream

lheckemann:
Ah, I assumed it was used in a shebang -- using the implicit PATH lookup rather than a fixed path to bin/env is much nicer of course!

aduh95:
I'm not I think this is ready to land. AFAICT the macOS failures are unrelated.

lheckemann:
Yes, the macOS failures are real but weren't introduced here. Not sure how to parse the first part of your comment though -- ready or not? :)

aduh95:
Yes sorry, I meant that IMO it's ready to land.

alyssais:
This has broken nodejs_18 and nodejs_22.

aduh95:
@alyssais can you give more info regarding what's broken?

alyssais:
On Mon, Jun 17, 2024 at 05:35:40AM -0700, Antoine du Hamel wrote:
> @alyssais can you give more info regarding what's broken?

They no longer build on x86_64-linux.

(On current staging, nodejs_20 doesn't build either, but that wasn't introduced by this PR, rather by the OpenSSL upgrade, and nodejs has an upstream patch for it.)

aduh95:
Do you have a link or could you copy the output please?

alyssais:
[nodejs_22.log](https://github.com/user-attachments/files/15888471/nodejs_22.log)

alyssais:
nodejs_18 on b26563aae2a1fce011ac52e28ff23a460866e82f is building for me now, so for that one it either fails intermittently or I made a mistake.

aduh95:
Hum unfortunately some tests are flaky (it's also a problem for Node.js CI, but unfortunately quite tricky to fix). I'll try to make a patch to automatically re-run failing tests to avoid failing the build just because of a flake.

alyssais:
If test failures are likely to be due to flakiness, is it worth it for us to run the tests in the first place?

alyssais:
@lheckemann wdyt?

lheckemann:
I'm guessing it's only individual tests that are flaky (and known to be flaky)? I think it would be worth skipping those specifically, or having the tests retry (if they can't be fixed in order not to be flaky). I imagine that would be a welcome improvement upstream as well?

aduh95:
There's already such system upstream, and this PR is already taking advantage of it and skip the tests that have been marked as flaky upstream: https://github.com/NixOS/nixpkgs/blob/667f4bf8524fa8a282e2c80ae8dbfa5b92bca5e6/pkgs/development/web/nodejs/nodejs.nix#L147

However, because the Node.js project has a relatively high bar for marking tests as flaky (the idea being a test marked as flaky has less chance of getting fixed, because flaky tests do not block PRs from merging); however, for NixOS, those flakes can be very annoying, and IMO it's not realistic to try to manage our own list of FLAKY tests, so I think the best course of action would be to allow to retry failing tests – if a test fails once, and succeeds e.g. the 9 following times, it's safe to assume it was just a flake, no?

lheckemann:
Right, but we don't want to have to restart drv builds every time the tests fail (especially since that will involve building the code again too). And I'd expect tests that aren't marked as flaky... not to be flaky :upside_down_face:

lheckemann:
Yep, nodejs/node@14863e80584e579fd48c55f6373878c821c7ff7e seems to be the fix for this, so it must have been an unfortunate timing coincidence with the merging of this PR.

lheckemann:
Asked about backporting the nodejs-side fix: https://github.com/nodejs/node/pull/53373#issuecomment-2181240666

aduh95:
> Asked about backporting the nodejs-side fix: [nodejs/node#53373 (comment)](https://github.com/nodejs/node/pull/53373#issuecomment-2181240666)

I cherry-picked that commit to the 20.x update PR

> Update, I don't think these are actually flaky, but failing maybe for OpenSSL-related reasons?

It's different from the failure reported in https://github.com/NixOS/nixpkgs/pull/313982#issuecomment-2176315033:

         not ok 5 - appends positional arguments
          ---
          duration_ms: 16.503608
          location: '/build/node-v22.2.0/test/parallel/test-node-run.js:57:3'
          failureType: 'testCodeFailure'
          error: |-
            The input did not match the regular expression /Arguments: '--help "hello world test" A B C'/. Input:
            
            ''
            
          code: 'ERR_ASSERTION'
          name: 'AssertionError'

I tried running that test 999 times on my machine for a stress test on the latest `main` branch of nodejs/node, and got no failures, so either it's a very rare flake, either it was already fixed.



> I'd expect tests that aren't marked as flaky... not to be flaky 🙃

It's certainly not an unreasonable expectation :) however it's inevitable there are some flaky tests that are going to slip through, and I guess the question is "are we fine with those flakes affecting us (Nix)?"

> we don't want to have to restart drv builds every time the tests fail

That's precisely why I'm suggesting we retry the failing tests to not fail the build if it hits a flaky test (that is, a test flaky in practice even though not marked as flaky for whatever reason).

alyssais:
> Update, I don't think these are actually flaky, but failing maybe for OpenSSL-related reasons?

That's not the test failures we're talking about here. (I mentioned it already in https://github.com/NixOS/nixpkgs/pull/313982#issuecomment-2175584046)

lheckemann:
Argh, sorry. I'll throw it at a builder a couple of times to see if I can get an idea of how often it happens.

vcunat:
Basically reverting this on darwin:
https://github.com/NixOS/nixpkgs/pull/325844

PR #313995, opened 2024-05-23
python312Packages.soundcard init at 0.4.3

matthiasdotsh:
## Description of changes

https://github.com/bastibe/SoundCard

A python package to interact with your soundcard.

## Things done

- Built on platform(s)
  - [X] x86_64-linux
  - [X] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [X] Tested, as applicable:

---

### DIFF
--- a/pkgs/development/python-modules/soundcard/default.nix
+++ b/pkgs/development/python-modules/soundcard/default.nix
@@ -1,0 +1,50 @@
+{
+  buildPythonPackage,
+  cffi,
+  fetchPypi,
+  lib,
+  libpulseaudio,
+  numpy,
+  setuptools,
+  testers,
+}:
+let
+  version = "0.4.3";
+in
+buildPythonPackage {
+  inherit version;
+  pname = "soundcard";
+  pyproject = true;
+
+  src = fetchPypi {
+    inherit version;
+    pname = "SoundCard";
+    hash = "sha256-QQg1UUuhCAmAPLmIfUJw85K1nq82WRW7lFFq8/ix0Dc=";
+  };
+
+  patchPhase = ''
+    substituteInPlace soundcard/pulseaudio.py \
+      --replace "'pulse'" "'${libpulseaudio}/lib/libpulse.so'"
+  '';
+
+  build-system = [ setuptools ];
+
+  dependencies = [
+    cffi
+    numpy
+  ];
+
+  # doesn't work because there are not many soundcards in the
+  # sandbox. See VM-test
+  # pythonImportsCheck = [ "soundcard" ];
+
+  passthru.tests.vm-with-soundcard = testers.runNixOSTest ./test.nix;
+
+  meta = {
+    description = "Pure-Python Real-Time Audio Library";
+    homepage = "https://github.com/bastibe/SoundCard";
+    changelog = "https://github.com/bastibe/SoundCard/blob/${version}/README.rst#changelog";
+    license = lib.licenses.bsd3;
+    maintainers = with lib.maintainers; [ matthiasdotsh ];
+  };
+}
--- a/pkgs/development/python-modules/soundcard/test.nix
+++ b/pkgs/development/python-modules/soundcard/test.nix
@@ -1,0 +1,47 @@
+{
+  name = "Python Soundcard Library Test";
+
+  nodes.machine =
+    { pkgs, ... }:
+    {
+      environment.systemPackages = [ (pkgs.python3.withPackages (ps: [ ps.soundcard ])) ];
+
+      hardware.pulseaudio.enable = true;
+      hardware.pulseaudio.systemWide = true;
+
+      virtualisation.qemu.options = [
+        "-device virtio-sound-pci,audiodev=my_audiodev"
+        "-audiodev wav,id=my_audiodev"
+      ];
+    };
+
+  interactive.nodes.machine =
+    { ... }:
+    {
+      services.openssh.enable = true;
+      services.openssh.settings.PermitRootLogin = "yes";
+      users.extraUsers.root.initialPassword = "";
+      users.extraUsers.root.hashedPassword = null;
+      users.extraUsers.root.hashedPasswordFile = null;
+      users.extraUsers.root.initialHashedPassword = null;
+      virtualisation.forwardPorts = [
+        {
+          from = "host";
+          host.port = 2222;
+          guest.port = 22;
+        }
+      ];
+    };
+
+  testScript =
+    let
+      script = builtins.toFile "import.py" ''
+        import soundcard
+      '';
+    in
+    ''
+      start_all()
+      machine.wait_for_unit("multi-user.target")
+      machine.succeed("python ${script}")
+    '';
+}
--- a/maintainers/maintainer-list.nix
+++ b/maintainers/maintainer-list.nix
@@ -12686,6 +12686,12 @@
     githubId = 427866;
     name = "Matthias Beyer";
   };
+  matthiasdotsh = {
+    email = "git@matthias.sh";
+    name = "Matthias";
+    github = "matthiasdotsh";
+    githubId = 142118899;
+  };
   matthiasq = {
     email = "matthias.queitsch@mailbox.org";
     github = "matthias-Q";
--- a/pkgs/top-level/python-packages.nix
+++ b/pkgs/top-level/python-packages.nix
@@ -14562,6 +14562,8 @@ self: super: with self; {
 
   sortedcontainers = callPackage ../development/python-modules/sortedcontainers { };
 
+  soundcard = callPackage ../development/python-modules/soundcard { };
+
   soundcloud-v2 = callPackage ../development/python-modules/soundcloud-v2 { };
 
   sounddevice = callPackage ../development/python-modules/sounddevice { };


pkgs/development/python-modules/soundcard/default.nix
```diff
@@ -0,0 +1,47 @@
...
+
+buildPythonPackage rec {
+  pname = "SoundCard";
+  version = "0.4.3";
+  pyproject = true;
+
+  src = fetchPypi {
+    inherit pname version;
+    hash = "sha256-QQg1UUuhCAmAPLmIfUJw85K1nq82WRW7lFFq8/ix0Dc=";
+  };
```
TomaSajt, line 20: IIRC `pname` is no longer allowed to contain uppercase characters. This means you can no longer do `inherit pname` in `fetchPypi`.

```suggestion
  pname = "soundcard";
  version = "0.4.3";
  pyproject = true;

  src = fetchPypi {
    pname = "SoundCard";
    inherit version;
    hash = "sha256-QQg1UUuhCAmAPLmIfUJw85K1nq82WRW7lFFq8/ix0Dc=";
  };
```

TomaSajt: Though it might also be better to use `fetchFromGitHub` instead of `fetchPypi` since `fetchFromGitHub` is a more easily inspectable source.

matthiasdotsh: Thanks for the comment. I was not aware about the `pname` thing. I just pushed a version with all lowercase `pname` and `fetchFromGitHub`.

Anything else I should change before this can be merged?

matthiasdotsh:
fore-push: update `pname` to all lowercase + use `fetchFromGitHub`

pkgs/development/python-modules/soundcard/default.nix
```diff
@@ -0,0 +1,49 @@
...
+    owner = "bastibe";
+    repo = pname;
+    rev = version;
+    hash = "sha256-d91fhgkjnRFamMIn7hmXuHZt6kGBysAhQq8IsZ/WqHs=";
+  };
+
+  patchPhase = ''
+    substituteInPlace soundcard/pulseaudio.py \
+      --replace "'pulse'" "'${libpulseaudio}/lib/libpulse.so'"
+  '';
```
TomaSajt, line 27: The newer `--replace-fail` flags should be used instead of the legacy `--replace` flag, since it won't silently fail.

Also, I think it's safer to also include `dlopen` in the replacement to not accidentally replace other instances of the `'pulse'` literal.

And finally, `libpulse.so` is only valid on linux systems. On darwin/Mac it's not `.so` but `.dylib`. `stdenv.hostPlatform.extensions.sharedLibrary` can be used instead, which evaluates to `.so` or `.dylib` depending on which system you're building for.


```suggestion
  patchPhase =
    let
      libpulse = "${libpulseaudio}/lib/libpulse${stdenv.hostPlatform.extensions.sharedLibrary}";
    in
    ''
      substituteInPlace soundcard/pulseaudio.py \
          --replace-fail "_ffi.dlopen('pulse')" "_ffi.dlopen('${libpulse}')"
    '';
```

pkgs/development/python-modules/soundcard/default.nix
```diff
@@ -0,0 +1,49 @@
...
+}:
+
+buildPythonPackage rec {
+  pname = "soundcard";
+  version = "0.4.3";
+  pyproject = true;
+
+  src = fetchFromGitHub {
+    owner = "bastibe";
+    repo = pname;
```
TomaSajt, line 19: This is a bit of a nit, but nowadays it's better practice not to use `pname` for substituting occurrences of the package name.

```suggestion
    repo = "soundcard";
```

TomaSajt (changes requested):
The commit message (and PR title) should be the following:
`python312Packages.soundcard: init at 0.4.3`

This is needed because the automatic build system CI won't use aliases. (`pyton3Packages` is an alias of `python311Packages`)

pkgs/development/python-modules/soundcard/default.nix
matthiasdotsh: Thank you very much for your advice here.
I don't have a MacOS to test right now, so I'm super grateful for feedback.

The [Upstream project](https://github.com/bastibe/SoundCard) says.

> SoundCard is a library for playing and recording audio without resorting to a CPython extension. Instead, it is implemented using the wonderful [CFFI](http://cffi.readthedocs.io/en/latest/) and the native audio libraries of Linux, Windows and macOS.

> SoundCard is cross-platform, and supports Linux/pulseaudio, Mac/coreaudio, and Windows/WASAPI. 


I'm not sure whether `pulseaudio` or `coraudio` should be used on a macOS with nix as package manager.

TomaSajt: I'll try to look into it and try it out

TomaSajt:
The `python312Packages` name is case-sensitive, so please update it to use lowercase starting `p`

pkgs/development/python-modules/soundcard/default.nix
TomaSajt: Okay, I asked around on the Matrix chat about `dlopen` on darwin, still waiting for a response.

The current patching should be changed, then, to only happen on linux.
This can be done with
```nix
postPatch = lib.optionalString stdenv.isLinux ''
  substituteInPlace ....
'';
```

pkgs/development/python-modules/soundcard/default.nix
```diff
@@ -0,0 +1,49 @@
...
+  pyproject = true;
+
+  src = fetchFromGitHub {
+    owner = "bastibe";
+    repo = "soundcard";
+    rev = version;
+    hash = "sha256-d91fhgkjnRFamMIn7hmXuHZt6kGBysAhQq8IsZ/WqHs=";
+  };
+
+  patchPhase = ''
```
TomaSajt, line 24: This might seem a bit weird, but `postPatch` should be used instead of `patchPhase`, since if you override the default logic of `patchPhase`, the `patches = [ ... ];` list will not work, if you ever want to use `.patch` files for this package in the future.

```suggestion
  postPatch = ''
```

pkgs/development/python-modules/soundcard/default.nix
matthiasdotsh: I just pushed a variant with `replace-fail`, where the patch is only applied (in `postPatch`) when we are on a linux system.
In this case I assume it's fine to use `.so` as extension.

TomaSajt (approved):
The package itself LGTM

I don't have experience writing/reading NixOS tests, so I am not able to provide much feedback for that.

---

A final nit would be the ordering of the top attrset:
I'd really recommend putting the python dependencies next to each other and putting `lib` above everything.

---

_In general, this is my personal preference, which you don't have to follow:
`lib`, `stdenv`, the builder-function, fetchers, hooks, build-time deps, normal deps, test deps, other._
_I use this because usually this is the order they appear in the derivation._

matthiasdotsh:
> A final nit would be the ordering of the top attrset:
> I'd really recommend putting the python dependencies next to each other and putting lib above everything.

I have just sorted the dependencies.
The way you do it sounds good to me at first. I'm still new to the nix environment and would be grateful for any feedback.

I've learned a lot from your review, thank you very much!


> The package itself LGTM

What happens now with the MR? Can you merge it directly? Is there anything else I need to do?

TomaSajt:
Well, I don't have any commit access to the repo, so you'll probably have to wait for someone who does to comes along.

The waiting time can be improved a bit by submitting this PR into some other place, such as
the "PRs already reviewed" thread on the NixOS discourse (https://discourse.nixos.org/t/prs-already-reviewed/2617/1599)
or maybe the Matrix chat for Review Requests (https://matrix.to/#/#review-requests:nixos.org)

I'll probably post it to "PRs already reviewed" after I do some tests on a Mac emulator later today.

pkgs/development/python-modules/soundcard/test.nix
```diff
@@ -0,0 +1,41 @@
...
+    services.openssh.enable = true;
+    services.openssh.settings.PermitRootLogin = "yes";
+    users.extraUsers.root.initialPassword = "";
+    users.extraUsers.root.hashedPassword = null;
+    users.extraUsers.root.hashedPasswordFile = null;
+    users.extraUsers.root.initialHashedPassword = null;
+    virtualisation.forwardPorts = [
+      { from = "host"; host.port = 2222; guest.port = 22; }
+    ];
+  };
```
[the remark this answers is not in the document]
TomaSajt: Hmm, interesting...
Could you share the resources about how to run this?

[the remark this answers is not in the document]
matthiasdotsh: Sure: https://nixcademy.com/2023/12/01/nixos-integration-tests-part-2/

matthiasdotsh:
Force push: rebase to latest master and fix merge conflict in `maintainers/maintainer-list.nix`

pkgs/development/python-modules/soundcard/default.nix
```diff
@@ -0,0 +1,47 @@
...
+  fetchPypi,
+  lib,
+  libpulseaudio,
+  numpy,
+  setuptools,
+  testers,
+}:
+
+buildPythonPackage rec {
+  pname = "SoundCard";
```
TomaSajt, line 13:
```suggestion
  pname = "soundcard";
```

pkgs/development/python-modules/soundcard/default.nix
```diff
@@ -0,0 +1,47 @@
...
+  version = "0.4.3";
+  pyproject = true;
+
+  src = fetchPypi {
+    inherit pname version;
```
TomaSajt, line 18:
```suggestion
    pname = "SoundCard";
    inherit version;
```

since `pname` is no longer the same as the pypi name, we have to manually provide the name

pkgs/development/python-modules/soundcard/default.nix
TomaSajt: still, the the `pname` that is provided for `buildPythonPackage` should be lowercase.

only the `fetchPypi`'s `pname` should have any upper case letters.

You can keep `version` in the `let in` block if you  don't want to use `rec`, but I think `pname` should just stay inside `buildPythonPackage`, since it's not needed anywhere else.

Sigmanificient (changes requested):

pkgs/development/python-modules/soundcard/default.nix
```diff
@@ -0,0 +1,49 @@
...
+  ];
+
+  # doesn't work because there are not many soundcards in the
+  # sandbox. See VM-test
+  # pythonImportsCheck = [ "soundcard" ];
+
+  passthru.tests.vm-with-soundcard = testers.runNixOSTest ./test.nix;
+
+  meta = with lib; {
+    description = "A Pure-Python Real-Time Audio Library";
```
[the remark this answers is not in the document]
matthiasdotsh: it is copied from the [upstream project](https://github.com/bastibe/SoundCard) ...

pkgs/development/python-modules/soundcard/default.nix
```diff
@@ -0,0 +1,49 @@
...
+    homepage = "https://github.com/bastibe/SoundCard";
+    changelog = "https://soundcard.readthedocs.io/en/latest/#changelog";
```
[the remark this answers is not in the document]
TomaSajt: It might be better to change it to
`"https://github.com/bastibe/SoundCard/blob/${version}/README.rst#changelog"`
since we can use the current version this way.

pkgs/development/python-modules/soundcard/test.nix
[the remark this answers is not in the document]
matthiasdotsh: You can run the test with `nix-build -A pkgs.python3Packages.soundcard.passthru.tests.vm-with-soundcard`

pkgs/development/python-modules/soundcard/default.nix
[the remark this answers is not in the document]
matthiasdotsh: do you have a suggestion? it looks to me like there are no tags available?

pkgs/development/python-modules/soundcard/default.nix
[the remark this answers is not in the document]
Sigmanificient: See https://github.com/NixOS/nixpkgs/tree/master/pkgs#meta-attributes

`meta.description` must:

- Be short, just one sentence.
- Be capitalized.
- Not start with the definite or an indefinite article.
- Not start with the package name.
  -  More generally, it should not refer to the package name.
- Not end with a period (or any punctuation for that matter).
- Provide factual information.
  - Avoid subjective language.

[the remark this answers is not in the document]
matthiasdotsh: Thank you, I wasn't aware of that.

PR #314058, opened 2024-05-23
lib.modules: add mkForAllItems

rhendric:
## Description of changes

For motivation, see [Discourse](https://discourse.nixos.org/t/how-to-change-the-default-nixos-option-in-a-list-of-attrsets-described-as/45561).

For explanation, see additions to nixos/doc/manual/development/option-def.section.md

**tl;dr:** if you wish you could write (pseudocode)
```nix
swapDevices.*.options = mkDefault [ /* ... */ ];
```

you can now do that without the dreaded infinite recursion by writing
```nix
swapDevices = mkForAllItems {
  options = mkDefault [ /* ... */ ];
};
```

Is this stupid? Nixpkgs has gotten by this far without it, and there are some alternative patterns available:
* Defining a template function somewhere and using that on every element (disadvantage: undesirable coupling?)
* Redefining the option itself and injecting defaults into its type (disadvantage: overly verbose, not intuitive?)

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [x] Tested, as applicable:

---

### DIFF
--- a/lib/modules.nix
+++ b/lib/modules.nix
@@ -1036,6 +1036,17 @@ let
   defaultOrderPriority = 1000;
   mkAfter = mkOrder 1500;
 
+  /*
+    For use with options of type `attrsOf *`, `lazyAttrsOf *`, or `listOf *`.
+    Accepts a value that is merged with every element defined elsewhere. May
+    also accept a function that takes the attribute name or index (1-based),
+    the result of which is used as above.
+  */
+  mkForAllItems = content:
+    { _type = "forAllItems";
+      content = if isFunction content then content else _: content;
+    };
+
   # Convenient property used to transfer all definitions and their
   # properties from one option to another. This property is useful for
   # renaming options, and also for including properties from another module
@@ -1421,6 +1432,7 @@ private //
     mkDefault
     mkDerivedConfig
     mkFixStrictness
+    mkForAllItems
     mkForce
     mkIf
     mkImageMediaOverride
--- a/lib/types.nix
+++ b/lib/types.nix
@@ -30,6 +30,7 @@ let
     imap1
     last
     length
+    partition
     tail
     ;
   inherit (lib.attrsets)
@@ -548,17 +549,23 @@ rec {
       name = "listOf";
       description = "list of ${optionDescriptionPhrase (class: class == "noun" || class == "composite") elemType}";
       descriptionClass = "composite";
-      check = isList;
+      check = x: isList x || x._type or "" == "forAllItems";
       merge = loc: defs:
+        let
+          # partitionedDefs.right: all forAllItems definitions
+          # partitionedDefs.wrong: all regular definitions
+          partitionedDefs = partition (d: d.value._type or "" == "forAllItems") defs;
+          extraDefs = idx: map (def: { inherit (def) file; value = def.value.content idx; }) partitionedDefs.right;
+        in
         map (x: x.value) (filter (x: x ? value) (concatLists (imap1 (n: def:
           imap1 (m: def':
             (mergeDefinitions
               (loc ++ ["[definition ${toString n}-entry ${toString m}]"])
               elemType
-              [{ inherit (def) file; value = def'; }]
+              ([{ inherit (def) file; value = def'; }] ++ extraDefs m)
             ).optionalValue
           ) def.value
-        ) defs)));
+        ) partitionedDefs.wrong)));
       emptyValue = { value = []; };
       getSubOptions = prefix: elemType.getSubOptions (prefix ++ ["*"]);
       getSubModules = elemType.getSubModules;
@@ -581,11 +588,17 @@ rec {
       descriptionClass = "composite";
       check = isAttrs;
       merge = loc: defs:
+        let
+          # partitionedDefs.right: all forAllItems definitions
+          # partitionedDefs.wrong: all regular definitions
+          partitionedDefs = partition (d: d.value._type or "" == "forAllItems") defs;
+          extraDefs = name: map (def: { inherit (def) file; value = def.value.content name; }) partitionedDefs.right;
+        in
         mapAttrs (n: v: v.value) (filterAttrs (n: v: v ? value) (zipAttrsWith (name: defs:
-            (mergeDefinitions (loc ++ [name]) elemType defs).optionalValue
+            (mergeDefinitions (loc ++ [name]) elemType (defs ++ extraDefs name)).optionalValue
           )
           # Push down position info.
-          (map (def: mapAttrs (n: v: { inherit (def) file; value = v; }) def.value) defs)));
+          (map (def: mapAttrs (n: v: { inherit (def) file; value = v; }) def.value) partitionedDefs.wrong)));
       emptyValue = { value = {}; };
       getSubOptions = prefix: elemType.getSubOptions (prefix ++ ["<name>"]);
       getSubModules = elemType.getSubModules;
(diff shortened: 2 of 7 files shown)


lib/modules.nix
```diff
@@ -1036,6 +1036,17 @@ let
   defaultOrderPriority = 1000;
   mkAfter = mkOrder 1500;
 
+  /*
+    For use with options of type `attrsOf *`, `lazyAttrsOf *`, or `listOf *`.
+    Accepts a value that is merged with every element defined elsewhere. May
+    also accept a function that takes the attribute name or index (1-based),
+    the result of which is used as above.
+  */
+  mkForAllValues = content:
```
rhendric, line 1045: This thread preallocated for bikeshedding the name of this function.

`forEach` is used elsewhere in the libraries, but there it means ‘flipped `map`’, and this function is not a map at all (if it is given a function, the argument provided to that function is key, not value). So I went with `ForAll` to draw the distinction.

`Values` splits the difference between `Elems` for lists and `Attrs` for attrsets, but I'm not wedded to it.

lib/modules.nix
```diff
@@ -1036,6 +1036,17 @@ let
   defaultOrderPriority = 1000;
   mkAfter = mkOrder 1500;
 
+  /*
+    For use with options of type `attrsOf *`, `lazyAttrsOf *`, or `listOf *`.
+    Accepts a value that is merged with every element defined elsewhere. May
+    also accept a function that takes the attribute name or index (1-based),
```
rhendric, line 1042: This thread preallocated for bikeshedding the choice of 1-based indexing.

I made an arbitrary choice here but AFAIK (not far) Nixpkgs is somewhat incoherent in which of these it uses by default. `imap` defaults to 1-indexed (but has explicit `imap0` and `imap1` options), `elemAt` defaults to 0-indexed (and has no `elemAt1` alternative). Values displayed to users seem to be 1-indexed.

lib/modules.nix
roberth: `Items` might be better. Everything except an attribute is already value in the expression language.

`types.nix` currently has `elemType` in `nestedTypes`, which I think was a mistake because `item` is a word and `elem` is not. (Also `elem` is a _function_ in Haskell, fwiw)

roberth (commented):
I feel ambivalent towards this addition.
While it's good to make this functionality more accessible, it introduces more complexity and a performance overhead.
It'd help to quantify that impact.
It might be possible to optimize the common case of no `mkForAllValues`.

rhendric:
> It'd help to quantify that impact.

Not sure how representative this'll be, but I ran this command before and after the patch:

```
NIX_SHOW_STATS=1 NIX_PATH=nixpkgs=$PWD:nixos-config=/etc/nixos/configuration.nix nix-instantiate '<nixpkgs/nixos>' -A system --eval
```

Selected results:

| | Before | After | Δ |
|--|--:|--:|--|
|`cpuTime`| 5.008959770202637 | 5.019791126251221 | +0.22%
|`envs.bytes`| 191776744 | 192550400 | +0.4%
|`envs.elements`| 9965909 | 10008938 | +0.43%
|`envs.number`| 7003092 | 7029931 |  +0.38%
|`gc.totalBytes`| 1315467280 | 1318265568 | +0.21%
|`list.bytes`| 19593080 | 19662352 | +0.35%
|`list.concats`| 375631 | 382919 | +1.9%
|`list.elements`| 2449135 | 2457794 | +0.35%
|`nrFunctionCalls`| 6312840 | 6334283 | +0.34%
|`nrThunks`| 10340995 | 10380775 | +0.38%
|`sets.bytes`| 480846832 | 481106752 | +0.054%
|`sets.elements`| 28250011 | 28260856 | +0.038%
|`sets.number`| 1802916 | 1808316 | +0.3%
|`values.bytes`| 362088648 | 363302568 |  +0.34%
|`values.number`| 15087027 | 15137607 | +0.34%

roberth:
> > It'd help to quantify that impact.
> 
> Not sure how representative this'll be,

The CPU time and even the GC stats need to be taken with a grain of salt, but the other stats are good proxies for performance. (except concats, which is too specific)
IMO this improvement is worth the cost. :+1: 

I wonder what's the cost of the implicit `{ _module.args.name = last loc; }` that we have in `submodule`, which is technically unnecessary after this, but that might be hard to test, and I don't think we should act on it anyway.

lib/types.nix
```diff
@@ -548,17 +549,23 @@ rec {
       name = "listOf";
       description = "list of ${optionDescriptionPhrase (class: class == "noun" || class == "composite") elemType}";
       descriptionClass = "composite";
-      check = isList;
+      check = x: isList x || x._type or "" == "forAllValues";
       merge = loc: defs:
+        let
+          partitionedDefs = partition (d: d.value._type or "" == "forAllValues") defs;
+          forAllValueses = partitionedDefs.right;
+          defs' = partitionedDefs.wrong;
```
roberth, line 557: By inlining single-use bindings like these, we could get rid of a few `Env` entries.

To avoid losing the documentation-like role of the names, those could be documented in a comment on `partitionedDefs` instead.

```suggestion
          # partitionedDefs.right: all forAllValues definitions
          # partitionedDefs.wrong: all regular definitions
          partitionedDefs = partition (d: d.value._type or "" == "forAllValues") defs;
```

roberth: Same in the other types.

roberth (commented):
Found a micro-optimization. I couldn't think of something more significant, except maybe it's worth optimizing the cases where we have no definitions? Only if you're interested in that kind of thing.

nixos/doc/manual/development/option-def.section.md
```diff
@@ -123,3 +123,29 @@ they were declared in separate modules. This can be done using
...
+overrides on all of the elements or attributes of that option independently of
+the population of the list or set. For example, one module can define a list of
+`swapDevices`, while another module can define that for every element of the list
+of `swapDevices`, `swapDevices.*.options` list should have a particular value. This
+can be done using `mkForAllValues`:
+
+```nix
+{
+  swapDevices = mkForAllValues {
+    options = mkDefault [ /* ... */ ];
```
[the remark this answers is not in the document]
rhendric: I'd be worried about the home directory example because, as implemented, it uses the value of one attribute in a `mkIf` used to compute another. This is doable with `mkForAllValues` as well but there is some subtlety to it and I don't want that to distract.

Is the issue with `swapDevices.*.options` just the word `options`? Can we take `swapDevices.*.discardPolicy` instead, or something? Or how about `services.wordpress.sites.<name>.uploadsDir`, if we want to showcase the parameter? (I'm leaning toward both.)

nixos/doc/manual/development/option-def.section.md
```diff
@@ -123,3 +123,29 @@ they were declared in separate modules. This can be done using
...
+```
+
+The argument to `mkForAllValues` will be merged with every element of the list,
+or every attribute of the attrset. As usual, use `mkDefault`, `mkForce`,
+`mkBefore`, `mkAfter`, etc. to control how multiple definitions of an option
+are combined.
+
+`mkForAllValues` can also accept a function which will receive the element
+index (starting at 1) or attribute name for each element or attribute on which
+it is used. The result of this function is used as described above.
```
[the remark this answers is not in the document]
rhendric: Sorry, I don't understand why this is connected to `mkForAllValues`. It's always the case that if you want to define a new option you need to declare it. Why is this something that should be explained in this section?

lib/modules.nix
rhendric: Changed to `Items`.

PR #314175, opened 2024-05-24
setup-hooks/strip: uniqify files by inode number before stripping

tpwrules:
## Description of changes

Mesa, among other packages, has binaries that are linked together and
can end up corrupted when the same binary is stripped through two
different names.

To resolve this, print out the device and inode number before each file
name, sort/uniq based on that, then cut it back out before stripping.

The symlink resolution logic is removed as the same file accessed
through two different links in `$paths` will necessarily have the same
numbers. File/directory within the paths listed in `$paths` are
correctly not (and were never) processed due to the `-type f` predicate
and (implied) `-P` option to `find`.

## Things done

- Built on platform(s)
  - [X] x86_64-linux
  - [x] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/build-support/setup-hooks/strip.sh
+++ b/pkgs/build-support/setup-hooks/strip.sh
@@ -74,13 +74,17 @@ stripDirs() {
         echo "stripping (with command $cmd and flags $stripFlags) in $paths"
         local striperr
         striperr="$(mktemp --tmpdir="$TMPDIR" 'striperr.XXXXXX')"
-        # Do not strip lib/debug. This is a directory used by setup-hooks/separate-debug-info.sh.
-        find $paths -type f "${excludeFlags[@]}" -a '!' -path "$prefix/lib/debug/*" -print0 |
-            # Make sure we process files under symlinks only once. Otherwise
-            # 'strip` can corrupt files when writes to them in parallel:
-            #   https://github.com/NixOS/nixpkgs/issues/246147#issuecomment-1657072039
-            xargs -r -0 -n1 -- realpath -z | sort -u -z |
+        # Make sure we process files only once. `strip`ping the same file through different
+        # links in parallel can corrupt it:
+        #   https://github.com/NixOS/nixpkgs/issues/246147#issuecomment-1657072039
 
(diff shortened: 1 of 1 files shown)


alyssais:
Why not just move the symlinkification to before the stripping, like @puckipedia [suggested](https://github.com/tpwrules/nixos-apple-silicon/issues/199#issuecomment-2111036503)?

tpwrules:
Mesa changes are going to need to go through staging anyway, and this does fix the root cause as also suggested.

However, fixing Mesa might be easier to review. I'll investigate, we should quite possibly do both anyway.

pkgs/build-support/setup-hooks/strip.sh
```diff
@@ -74,13 +74,17 @@ stripDirs() {
...
+        # links in parallel can corrupt it:
+        #   https://github.com/NixOS/nixpkgs/issues/246147#issuecomment-1657072039
 
+        # Do not strip lib/debug. This is a directory used by setup-hooks/separate-debug-info.sh.
+        # Print out each file's device and inode (which will be the same if two files are hardlinked
+        # or are the same file found through different symlinks), followed by its path...
+        find $paths -type f "${excludeFlags[@]}" -a '!' -path "$prefix/lib/debug/*" -printf '%D-%i,%p\0' |
+            # ... sort/uniq by device/inode, then cut them out and keep the path, ...
+            sort -t, -k1,1 -u -z | cut -d, -f2- -z |
+            # and finally strip each unique path in parallel.
```
Artturin, line 87:
```suggestion
            # and finally strip each unique path in parallel.
```

philiptaron (approved):
Yeah, this works.

trofi (approved):

lheckemann (approved):
I haven't built anything, but the diff looks good.

winterqt:
I wonder if it's worth sending this to 23.11 too, since we have at least a few more staging cycles to go in it.

tpwrules:
That would fix the issue in all stable releases which seems like a good thing. I would be happy to at least build such a patch on my machine.

PR #314268, opened 2024-05-24
rustup: addressing a series of issues encountered while using Rust's self-contained `ld.lld`

### LINKED ISSUE #312661: Latest rustup update version breaks compilation
### Describe the bug

Running a fresh `rustup install nightly && cargo build` fails, the core message being:

```
note: Could not start dynamically linked executable: /home/[xxx]/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/x86_64-unknown-linux-gnu/bin/gcc-ld/ld.lld
          NixOS cannot run dynamically linked executables intended for generic
          linux environments out of the box. For more information, see:
          https://nix.dev/permalink/stub-ld
          collect2: error: ld returned 127 exit status
```

### Steps To Reproduce
1. have a rust project
2. `rustup install nightly` (or `rustup update` if you already have one)
3. `cargo build`

### Expected behavior
Either the rustup gcc/ld to be patched for NixOS use, or rust to use the system gcc/ld.

### Metadata

```console
 - system: `"x86_64-linux"`
 - host os: `Linux 6.1.90, NixOS, 24.05 (Uakari), 24.05pre625274.205781405197`
 - multi-user?: `yes`
 - sandbox: `yes`
 - channels(root): `"nixos-23.11, nixpkgs-23.11"`
```

dianqk:
## Description of changes

Fixes #312661.

This PR primarily makes the following changes:

- Patching `ld.lld` and other files by determining whether they are ELF files.
- Adding `ld-wrapped.sh` to `ld.lld` (inspired entirely by @oxalica's work in https://github.com/oxalica/rust-overlay/pull/174).
- Adding `rpath` to `rust-lld`.

Determining whether to patch based on the pathname is inaccurate, which results in the loss of lld under `bin/gcc-ld` and attempts to patch `*.rlib` files.

I determine the patching method by looking up `.interp`, which should be simpler than the code from `readelf`.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [x] Tested, as applicable:

---

### DIFF
--- a/pkgs/development/tools/rust/rustup/default.nix
+++ b/pkgs/development/tools/rust/rustup/default.nix
@@ -48,7 +48,12 @@ rustPlatform.buildRustPackage rec {
   checkFeatures = [ ];
 
   patches = lib.optionals stdenv.isLinux [
-    (runCommand "0001-dynamically-patchelf-binaries.patch" { CC = stdenv.cc; patchelf = patchelf; libPath = "$ORIGIN/../lib:${libPath}"; } ''
+    (runCommand "0001-dynamically-patchelf-binaries.patch"
+      {
+        CC = stdenv.cc;
+        patchelf = patchelf;
+        libPath = "${libPath}";
+      } ''
       export dynamicLinker=$(cat $CC/nix-support/dynamic-linker)
       substitute ${./0001-dynamically-patchelf-binaries.patch} $out \
         --subst-var patchelf \
@@ -96,7 +101,22 @@ rustPlatform.buildRustPackage rec {
     # Note: fish completion script is not supported.
     $out/bin/rustup completions bash cargo > "$out/share/bash-completion/completions/cargo"
     $out/bin/rustup completions zsh cargo >  "$out/share/zsh/site-functions/_cargo"
+
+    # add a wrapper script for ld.lld
+    mkdir -p $out/nix-support
+    substituteAll ${../../../../../pkgs/build-support/wrapper-common/utils.bash} $out/nix-support/utils.bash
+    substituteAll ${../../../../../pkgs/build-support/bintools-wrapper/add-flags.sh} $out/nix-support/add-flags.sh
+    substituteAll ${../../../../../pkgs/build-support/bintools-wrapper/add-hardening.sh} $out/nix-support/add-hardening.sh
+    export prog='$PROG'
+    export use_response_file_by_default=0
+    substituteAll ${../../../../../pkgs/build-support/bintools-wrapper/ld-wrapper.sh} $out/nix-support/ld-wrapper.sh
+    chmod +x $out/nix-support/ld-wrapper.sh
   '';
+
+  env = lib.optionalAttrs (pname == "rustup") {
+    inherit (stdenv.cc.bintools) expandResponseParams shell suffixSalt wrapperName coreutils_bin;
+    hardening_unsupported_flags = "";
+  };
 
   meta = with lib; {
     description = "The Rust toolchain installer";
--- a/pkgs/development/tools/rust/rustup/0001-dynamically-patchelf-binaries.patch
+++ b/pkgs/development/tools/rust/rustup/0001-dynamically-patchelf-binaries.patch
@@ -1,39 +1,185 @@
 diff --git a/src/dist/component/package.rs b/src/dist/component/package.rs
-index 73a533b5..408ab815 100644
+index dfccc661..85233f3b 100644
 --- a/src/dist/component/package.rs
 +++ b/src/dist/component/package.rs
-@@ -113,6 +113,7 @@ fn install<'a>(
+@@ -113,6 +113,7 @@ impl Package for DirectoryPackage {
                      } else {
                          builder.move_file(path.clone(), &src_path)?
                      }
-+                    nix_patchelf_if_needed(&target.prefix().path().join(path.clone()), &src_path)
++                    nix_patchelf_if_needed(&target.prefix().path().join(path.clone()))
                  }
                  "dir" => {
                      if self.copy {
-@@ -135,6 +136,29 @@ fn components(&self) -> Vec<String> {
+@@ -135,6 +136,175 @@ impl Package for DirectoryPackage {
      }
  }
  
-+fn nix_patchelf_if_needed(dest_path: &Path, src_path: &Path) {
-+    let (is_bin, is_lib) = if let Some(p) = src_path.parent() {
-+        (p.ends_with("bin") || p.ends_with("libexec"), p.ends_with("lib"))
-+    } else {
-+        (false, false)
-+    };
++fn nix_wrap_lld(dest_lld_path: &Path) -> Result<()> {
++    use std::fs;
++    use std::io::Write;
++    use std::os::unix::fs::PermissionsExt;
 +
-+    if is_bin {
-+        let _ = ::std::process::Command::new("@patchelf@/bin/patchelf")
-+            .arg("--set-interpreter")
-+            .arg("@dynamicLinker@")
-+            .arg(dest_path)
-+            .output();
++    let path = dest_lld_path.parent().unwrap();
++    let mut unwrapped_name = path.file_name().unwrap().to_string_lossy().to_string();
++    unwrapped_name.push_str("-unwrapped");
++    let unwrapped_dir = path.with_file_name(unwrapped_name);
++    fs::create_dir(&unwrapped_dir).context("failed to create unwrapped directory")?;
++    let mut unwrapped_lld = unwrapped_dir;
++    unwrapped_lld.push(dest_lld_path.file_name().unwrap());
++    fs::rename(dest_lld_path, &unwrapped_lld).context("failed to move file")?;
++    let mut ld_wrapper_path = std::env::current_exe()?
++        .parent()
++        .ok_or(anyhow!("failed to get parent directory"))?
++        .with_file_name("nix-support");
++    let mut file = std::fs::File::create(dest_lld_path)?;
++    ld_wrapper_path.push("ld-wrapper.sh");
++
++    let wrapped_script = format!(
++        "#!/usr/bin/env bash
++set -eu -o pipefail +o posix
++shopt -s nullglob
++export PROG=\"{}\"
++\"{}\" $@",
++        unwrapped_lld.to_string_lossy().to_string(),
++        ld_wrapper_path.to_string_lossy().to_string(),
++    );
++    file.write_all(wrapped_script.as_bytes())?;
++    let mut permissions = file.metadata()?.permissions();
++    permissions.set_mode(0o755);
++    file.set_permissions(permissions)?;
++    Ok(())
++}
++
++fn nix_patchelf_if_needed(dest_path: &Path) {
++    use std::fs::File;
++    use std::os::unix::fs::FileExt;
++
++    struct ELFReader<'a> {
++        file: &'a mut File,
++        is_32bit: bool,
++        is_little_end: bool,
 +    }
-+    else if is_lib {
-+        let _ = ::std::process::Command::new("@patchelf@/bin/patchelf")
-+            .arg("--set-rpath")
-+            .arg("@libPath@")
-+            .arg(dest_path)
-+            .output();
++
++    impl<'a> ELFReader<'a> {
++        const MAGIC_NUMBER: &'static [u8] = &[0x7F, 0x45, 0x4c, 0x46];
++        const ET_EXEC: u16 = 0x2;
++        const ET_DYN: u16 = 0x3;
++        const PT_INTERP: u32 = 0x3;
++
++        fn new(file: &'a mut File) -> Option<Self> {
++            let mut magic_number = [0; 4];
++            file.read_exact(&mut magic_number).ok()?;
++            if Self::MAGIC_NUMBER != magic_number {
++                return None;
++            }
++            let mut ei_class = [0; 1];
++            file.read_exact_at(&mut ei_class, 0x4).ok()?;
++            let is_32bit = ei_class[0] == 1;
++            let mut ei_data = [0; 1];
++            file.read_exact_at(&mut ei_data, 0x5).ok()?;
++            let is_little_end = ei_data[0] == 1;
++            Some(Self {
++                file,
++                is_32bit,
++                is_little_end,
++            })
++        }
++
++        fn is_exec_or_dyn(&self) -> bool {
++            let e_type = self.read_u16_at(0x10);
++            e_type == Self::ET_EXEC || e_type == Self::ET_DYN
++        }
++
++        fn e_phoff(&self) -> u64 {
++            if self.is_32bit {
++                self.read_u32_at(0x1C) as u64
++            } else {
++                self.read_u64_at(0x20)
++            }
++        }
++
++        fn e_phentsize(&self) -> u64 {
++            let offset = if self.is_32bit { 0x2A } else { 0x36 };
++            self.read_u16_at(offset) as u64
++        }
++
++        fn e_phnum(&self) -> u64 {
++            let offset = if self.is_32bit { 0x2C } else { 0x38 };
++            self.read_u16_at(offset) as u64
++        }
++
++        fn has_interp(&self) -> bool {
++            let e_phoff = self.e_phoff();
++            let e_phentsize = self.e_phentsize();
++            let e_phnum = self.e_phnum();
++            for i in 0..e_phnum {
++                let p_type = self.read_u32_at(e_phoff + i * e_phentsize);
++                if p_type == Self::PT_INTERP {
++                    return true;
++                }
++            }
++            false
++        }
++
++        fn read_u16_at(&self, offset: u64) -> u16 {
++            let mut data = [0; 2];
++            self.file.read_exact_at(&mut data, offset).unwrap();
++            if self.is_little_end {
++                u16::from_le_bytes(data)
++            } else {
++                u16::from_be_bytes(data)
++            }
++        }
++
++        fn read_u32_at(&self, offset: u64) -> u32 {
++            let mut data = [0; 4];
++            self.file.read_exact_at(&mut data, offset).unwrap();
++            if self.is_little_end {
++                u32::from_le_bytes(data)
++            } else {
++                u32::from_be_bytes(data)
++            }
++        }
++
++        fn read_u64_at(&self, offset: u64) -> u64 {
++            let mut data = [0; 8];
++            self.file.read_exact_at(&mut data, offset).unwrap();
++            if self.is_little_end {
++                u64::from_le_bytes(data)
++            } else {
++                u64::from_be_bytes(data)
++            }
++        }
++    }
++
++    let Some(mut dest_file) = File::open(dest_path).ok() else {
++        return;
++    };
++    let Some(elf) = ELFReader::new(&mut dest_file) else {
++        return;
++    };
++    if !elf.is_exec_or_dyn() {
++        return;
++    }
++    let mut patch_command = std::process::Command::new("@patchelf@/bin/patchelf");
++    if elf.has_interp() {
++        patch_command
++            .arg("--set-interpreter")
++            .arg("@dynamicLinker@");
++    }
++    if Some(std::ffi::OsStr::new("rust-lld")) == dest_path.file_name() || !elf.has_interp() {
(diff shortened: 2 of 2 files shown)


DianQK:
ping? @mic92

samueltardieu:
Result of `nixpkgs-review pr 314268` run on x86_64-linux [1](https://github.com/Mic92/nixpkgs-review)
<details>
  <summary>3 packages built:</summary>
  <ul>
    <li>cargo-dist</li>
    <li>cargo-msrv</li>
    <li>rustup</li>
  </ul>
</details>

samueltardieu (approved):

pkgs/development/tools/rust/rustup/default.nix
```diff
@@ -96,8 +102,23 @@ rustPlatform.buildRustPackage rec {
     # Note: fish completion script is not supported.
     $out/bin/rustup completions bash cargo > "$out/share/bash-completion/completions/cargo"
     $out/bin/rustup completions zsh cargo >  "$out/share/zsh/site-functions/_cargo"
+
+    # add a wrapper script for ld.lld
+    mkdir -p $out/nix-support
+    substituteAll ${path + "/pkgs/build-support/wrapper-common/utils.bash"} $out/nix-support/utils.bash
```
Mic92, line 108: We shouldn't reference `path` in nixpkgs as this will cause to create a copy nixpkgs when evaluating this, which is quite expensive. Please use a relative path here.

pkgs/development/tools/rust/rustup/default.nix
```diff
@@ -96,8 +102,23 @@ rustPlatform.buildRustPackage rec {
...
+    substituteAll ${path + "/pkgs/build-support/bintools-wrapper/add-flags.sh"} $out/nix-support/add-flags.sh
+    substituteAll ${path + "/pkgs/build-support/bintools-wrapper/add-hardening.sh"} $out/nix-support/add-hardening.sh
+    export prog='$PROG'
+    export use_response_file_by_default=0
+    substituteAll  ${path + "/pkgs/build-support/bintools-wrapper/ld-wrapper.sh"} $out/nix-support/ld-wrapper.sh
```
Mic92, line 113:
```suggestion
    substituteAll ${path + "/pkgs/build-support/bintools-wrapper/add-flags.sh"} $out/nix-support/add-flags.sh
    substituteAll ${path + "/pkgs/build-support/bintools-wrapper/add-hardening.sh"} $out/nix-support/add-hardening.sh
    export prog='$PROG'
    export use_response_file_by_default=0
    substituteAll  ${path + "/pkgs/build-support/bintools-wrapper/ld-wrapper.sh"} $out/nix-support/ld-wrapper.sh
```

Same here

Mic92:
What command can I use for testing, which did not work before?

Mic92:
Looks like `nightly` is affected according to https://github.com/NixOS/nixpkgs/issues/312661

DianQK:
> What command can I use for testing, which did not work before? i.e. in which rust version does this first break?

The latest nightly rustc has switched the linker to the self-contained lld. This is an example for testing:

`Cargo.toml`:
```toml
[package]
name = "testlld"
version = "0.1.0"
edition = "2021"

[build-dependencies]
cc = "1.0"
```

`src/main.rs`:
```rust
extern "C" {
    fn add(a: i32, b: i32) -> i32;
}

fn main() {
    unsafe { add(1, 2) };
}
```

`build.rs`:
```rust
fn main() {
    cc::Build::new()
        .cpp(true)
        .file("foo.cpp")
        .cpp_link_stdlib("stdc++")
        .compile("foo");
}
```

`foo.cpp`:
```cpp
#include <iostream>
#include <vector>

extern "C" {
    int add(int a, int b) {
          std::vector<int> numbers = {1, 2, 3, 4, 5};
          numbers.push_back(6);
          for (int number : numbers) {
            std::cout << number << " ";
          }
          std::cout << std::endl;
          return a + b;
    }
}
```

Then run the script:
```sh
rustup uninstall nightly-2024-05-20 
rustup install nightly-2024-05-20
cargo clean
unset LD_LIBRARY_PATH
cargo +nightly-2024-05-20 run
```

We will get the following error:
```
= note: Could not start dynamically linked executable: /home/dianqk/.rustup/toolchains/nightly-2024-05-20-x86_64-unknown-linux-gnu/lib/rustlib/x86_64-unknown-linux-gnu/bin/gcc-ld/ld.lld
          NixOS cannot run dynamically linked executables intended for generic
          linux environments out of the box. For more information, see:
          https://nix.dev/permalink/stub-ld
          clang: error: unable to execute command: No such file or directory
          clang: error: linker command failed due to signal (use -v to see invocation)
```

After using `patchelf` to fix it, `cargo run` results in the following error:
```sh
$ patchelf --set-interpreter $(patchelf --print-interpreter $(which patchelf)) ~/.rustup/toolchains/nightly-2024-05-20-x86_64-unknown-linux-gnu/lib/rustlib/x86_64-unknown-linux-gnu/bin/gcc-ld/ld.lld
$ unset LD_LIBRARY_PATH
$ cargo +nightly-2024-05-20 run
target/debug/testlld: error while loading shared libraries: libstdc++.so.6: cannot open shared object file: No such file or directory
```

DianQK:
Reading the ELF file format on 32-bit systems is different. I don't have the environment to test it. :'(

Mic92:
Testing 32-bit is hopefully as easy as building `nix-build -A pkgsi686Linux.rustup` with your example above. I will have a look.

Mic92:
@ofborg build pkgsi686Linux.rustup

Mic92:
Interestingly on x86_64-linux I get the following build error:

```
failures:

---- suite::cli_self_upd::as_rustup_setup stdout ----
inprocess: true
status: Some(0)
duration: 0.002s
stdout:
====

====

stderr:
====

====

inprocess: true
status: Some(0)
duration: 0.001s
stdout:
====

====

stderr:
====
info: auto-self-update mode set to 'disable'

====

thread 'suite::cli_self_upd::as_rustup_setup' panicked at tests/mock/clitools.rs:750:25:
Unable to run test command: Os { code: 26, kind: ExecutableFileBusy, message: "Text file busy" }

failures:
    suite::cli_self_upd::as_rustup_setup

test result: FAILED. 440 passed; 1 failed; 8 ignored; 0 measured; 4 filtered out; finished in 39.94s

```

However this might be a different issue, i.e. maybe too many build cores on the system I am building from.

DianQK:
> However this might be a different issue, i.e. maybe too many build cores on the system I am building from.
> 
> **UPDATE** Yes. When building with less cores, it becomes better.

I can't reproduce it on my 16-core, 32-thread system. Could you test the upstream repository? I'm curious whether the issue lies with `cargo test` or the test case itself.

You can test it with the following script:
```bash
git clone https://github.com/rust-lang/rustup.git --branch 1.26.0 --depth 1
cd rustup
rustup install 1.77.2
cargo +1.77.2 test
```

Mic92:
> > However this might be a different issue, i.e. maybe too many build cores on the system I am building from.
> > **UPDATE** Yes. When building with less cores, it becomes better.
> 
> I can't reproduce it on my 16-core, 32-thread system. Could you test the upstream repository? I'm curious whether the issue lies with `cargo test` or the test case itself.
> 
> You can test it with the following script:
> 
> ```shell
> git clone https://github.com/rust-lang/rustup.git --branch 1.26.0 --depth 1
> cd rustup
> rustup install 1.77.2
> cargo +1.77.2 test
> ```

I don't think we need to fix this in this PR. It's a AMD EPYC 7713P 64-Core, I often run into the problem that build systems and test suite are not exercised with this core count.

DianQK:
> Testing 32-bit is hopefully as easy as building `nix-build -A pkgsi686Linux.rustup` with your example above. I will have a look.

I mainly want to test whether `rustup` correctly fixes the files for i686 during the installation of the toolchain.
Ah, I can use the script to test it:
```bash
$ env RUSTUP_DEBUG=true rustup toolchain install nightly-i686-unknown-linux-gnu --force-non-host

debug: patching "/home/dianqk/.rustup/toolchains/nightly-i686-unknown-linux-gnu/lib/rustlib/i686-unknown-linux-gnu/lib/libstd-4ebf06893886ecff.so" using patchelf
...
debug: patching "/home/dianqk/.rustup/toolchains/nightly-i686-unknown-linux-gnu/bin/rustc" using patchelf
debug: patching "/home/dianqk/.rustup/toolchains/nightly-i686-unknown-linux-gnu/bin/rustdoc" using patchelf
debug: patching "/home/dianqk/.rustup/toolchains/nightly-i686-unknown-linux-gnu/lib/librustc_driver-370689965a44b509.so" using patchelf
debug: patching "/home/dianqk/.rustup/toolchains/nightly-i686-unknown-linux-gnu/lib/libstd-4ebf06893886ecff.so" using patchelf
debug: patching "/home/dianqk/.rustup/toolchains/nightly-i686-unknown-linux-gnu/lib/rustlib/i686-unknown-linux-gnu/bin/gcc-ld/ld.lld" using patchelf
debug: patching "/home/dianqk/.rustup/toolchains/nightly-i686-unknown-linux-gnu/lib/rustlib/i686-unknown-linux-gnu/bin/gcc-ld/ld64.lld" using patchelf
debug: patching "/home/dianqk/.rustup/toolchains/nightly-i686-unknown-linux-gnu/lib/rustlib/i686-unknown-linux-gnu/bin/gcc-ld/lld-link" using patchelf
debug: patching "/home/dianqk/.rustup/toolchains/nightly-i686-unknown-linux-gnu/lib/rustlib/i686-unknown-linux-gnu/bin/gcc-ld/wasm-ld" using patchelf
debug: patching "/home/dianqk/.rustup/toolchains/nightly-i686-unknown-linux-gnu/lib/rustlib/i686-unknown-linux-gnu/bin/rust-lld" using patchelf
debug: patching "/home/dianqk/.rustup/toolchains/nightly-i686-unknown-linux-gnu/libexec/rust-analyzer-proc-macro-srv" using patchelf
```

DianQK:
I pushed a commit for this patch file: https://github.com/DianQK/rustup/commit/f14f5f2683c67543cc18166f3564f66e0bda3b26, which should help in reviewing the changes.

Mic92:
Looks like the 32bit rustup build is broken anyway...

Mic92:
So far the 32-bit headers still seem to be corrupted:

```
% file ./rustc
./rustc: ELF 32-bit LSB pie executable, Intel 80386, version 1 (SYSV), dynamically linked, interpreter /nix/store/k7zgvzp2r31zkg9xqgjim7mbknryv6bs-glibc-2.39-52/lib/ld-linux-x86-64.so.2, for GNU/Linux 2.6.32, with debug_info, not stripped
% ./rustc
zsh: accessing a corrupted shared library: ./rustc

``

**Comment by Mic92** [comment]
What seems to work is `rustup target add i686-unknown-linux-gnu` as opposed to a 32-bit rustc.

File: pkgs/development/tools/rust/rustup/default.nix [hunk 3]
```diff
@@ -96,8 +101,23 @@ rustPlatform.buildRustPackage rec {
...
+    export prog='$PROG'
+    export use_response_file_by_default=0
+    substituteAll ${../../../../../pkgs/build-support/bintools-wrapper/ld-wrapper.sh} $out/nix-support/ld-wrapper.sh
+    chmod +x $out/nix-support/ld-wrapper.sh
   '';
 
+  env = lib.optionalAttrs (pname == "rustup") {
+    inherit (stdenv.cc.bintools) expandResponseParams shell suffixSalt wrapperName coreutils_bin;
+    hardening_unsupported_flags = "";
+  };
```
Mic92, line 119: I wonder if we could make a better regression test here, given the growing complexity of machinery that gets added here?
If we cannot make it pure, even a self-contained script would be nice i.e. based on the example project that you already provided.

pkgs/development/tools/rust/rustup/default.nix
```diff
@@ -48,7 +48,12 @@ rustPlatform.buildRustPackage rec {
...
   patches = lib.optionals stdenv.isLinux [
-    (runCommand "0001-dynamically-patchelf-binaries.patch" { CC = stdenv.cc; patchelf = patchelf; libPath = "$ORIGIN/../lib:${libPath}"; } ''
+    (runCommand "0001-dynamically-patchelf-binaries.patch"
+      {
+        CC = stdenv.cc;
+        patchelf = patchelf;
+        libPath = "${libPath}";
+      } ''
       export dynamicLinker=$(cat $CC/nix-support/dynamic-linker)
       substitute ${./0001-dynamically-patchelf-binaries.patch} $out \
```
Mic92, line 58: I think 32-bit support never properly worked here, I would leave it to you, if you want to fix this in this pull request or later.

DianQK:
> So far the 32-bit headers still seem to be corrupted:
> 
> ```
> % file ./rustc
> ./rustc: ELF 32-bit LSB pie executable, Intel 80386, version 1 (SYSV), dynamically linked, interpreter /nix/store/k7zgvzp2r31zkg9xqgjim7mbknryv6bs-glibc-2.39-52/lib/ld-linux-x86-64.so.2, for GNU/Linux 2.6.32, with debug_info, not stripped
> % ./rustc
> zsh: accessing a corrupted shared library: ./rustc
> ```
> 
> I now actually doubt, this ever worked because we are trying to use the 64-bit link-loader for a 32-bit executable.

Is `rustup` used to install `rustc` on the i686 platform? It is not common practice to use `rustup` on the x86-64 platform to install the i686 toolchain. I think we only need to ensure that `rustup` on i686 can successfully install and run rustc.

DianQK:
> What seems to work is `rustup target add i686-unknown-linux-gnu` as opposed to a 32-bit rustc.
> 
> ```
> with import <nixpkgs> {};
> pkgsi686Linux.stdenv.mkDerivation {
>   name = "hello";
> }
> ```
> 
> ```
> % file ./target/i686-unknown-linux-gnu/debug/testlld
> ./target/i686-unknown-linux-gnu/debug/testlld: ELF 32-bit LSB pie executable, Intel 80386, version 1 (SYSV), dynamically linked, interpreter /nix/store/756q2acnjgi6ym709gw1f3jp0ciqwib1-glibc-2.39-52/lib/ld-linux.so.2, for GNU/Linux 3.10.0, with debug_info, not stripped
> 
> %  ./target/i686-unknown-linux-gnu/debug/testlld
> 1 2 3 4 5 6
> ```

I'm not quite sure what you're expressing here. Are you referring to results entirely on i686?

pkgs/development/tools/rust/rustup/default.nix
DianQK: Sure, we do need a test case. I plan to add an internal test case for the rustup project in a subsequent PR.

pkgs/development/tools/rust/rustup/default.nix
DianQK: I don't think this introduces any new regressions. I removed `$ORIGIN/../lib`, but I replaced it with `--add-rpath`.
Furthermore, I changed this because that `rust-lld` requires `$ORIGIN/../../lib`.

DianQK:
> So far the 32-bit headers still seem to be corrupted:
> 
> ```
> % file ./rustc
> ./rustc: ELF 32-bit LSB pie executable, Intel 80386, version 1 (SYSV), dynamically linked, interpreter /nix/store/k7zgvzp2r31zkg9xqgjim7mbknryv6bs-glibc-2.39-52/lib/ld-linux-x86-64.so.2, for GNU/Linux 2.6.32, with debug_info, not stripped
> % ./rustc
> zsh: accessing a corrupted shared library: ./rustc
> ```
> 
> I now actually doubt, this ever worked because we are trying to use the 64-bit link-loader for a 32-bit executable.

I can use `rustup`  of the master branch or the PR to build `pkgsi686Linux.rustup` on x86-64, and then use `rustup` to install `rustc` for `x86-64` to reproduce a similar issue.

pkgs/development/tools/rust/rustup/default.nix
Mic92: No, it's not a new regression.

Mic92:
> > I now actually doubt, this ever worked because we are trying to use the 64-bit link-loader for a 32-bit executable.
> 
> I can use `rustup` of the master branch or the PR to build `pkgsi686Linux.rustup` on x86-64, and then use `rustup` to install `rustc` for `x86-64` to reproduce a similar issue.

If you can fix it, it's nice, but it would be also not a new regression because it's broken just now.

Mic92:
> > What seems to work is `rustup target add i686-unknown-linux-gnu` as opposed to a 32-bit rustc.
> > ```
> > with import <nixpkgs> {};
> > pkgsi686Linux.stdenv.mkDerivation {
> >   name = "hello";
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > ```
> > % file ./target/i686-unknown-linux-gnu/debug/testlld
> > ./target/i686-unknown-linux-gnu/debug/testlld: ELF 32-bit LSB pie executable, Intel 80386, version 1 (SYSV), dynamically linked, interpreter /nix/store/756q2acnjgi6ym709gw1f3jp0ciqwib1-glibc-2.39-52/lib/ld-linux.so.2, for GNU/Linux 3.10.0, with debug_info, not stripped
> > 
> > %  ./target/i686-unknown-linux-gnu/debug/testlld
> > 1 2 3 4 5 6
> > ```
> 
> I'm not quite sure what you're expressing here. Are you referring to results entirely on i686?

I thought, this would also require patchelf, but actually it doesn't because everything is compiled afterwards.

DianQK:
I found that `doCheck` fails on CI. Would this affect the i686 build after merging? Perhaps the host for this CI is i686?

Mic92:
Only pkgsi686Linux.rustup failed but we also have `rustup, rustup.passthru.tests on x86_64-linux Success`. However I checked and `pkgsi686Linux.rustup` also doesn't work on master. We are slowly phasing out 32-bit to be honest. We do not even have a NixOS installer for it anymore: https://nixos.org/download/

DianQK:
> Only pkgsi686Linux.rustup failed but we also have `rustup, rustup.passthru.tests on x86_64-linux Success`. However I checked and `pkgsi686Linux.rustup` also doesn't work on master. We are slowly phasing out 32-bit to be honest. We do not even have a NixOS installer for it anymore: https://nixos.org/download/

This seems reasonable as well. I'm not currently considering dealing with this particular cross-compilation, but I'll try to add a test case.

BTW, *we should backport this to 24.05*. The backport note could be: fixed rustc's self-contained lld not functioning.

oxalica:
Should we use nix-ld instead of patchelf in rustup for this? I've seen some people (from nixos_zhcn) complaining about rustup's toolchains break constantly on each system updates, possibly because it's not added to GC roots? I'm not sure if this PR is a good place to discuss it, but this PR does make patchelf in rustup more and more complicated.

DianQK:
> Should we use nix-ld instead of patchelf in rustup for this? 

I think we should remain the behavior of rustc itself. `rust-lld` might perform some proprietary operations, and silently replacing it could cause issues. Additionally, `rustc` offers the option to use an external `ld`.

> I've seen some people (from nixos_zhcn) complaining about rustup's toolchains break constantly on each system updates, possibly because it's not added to GC roots?

I think we could add a sub-command like `rustup repatch`.

> I'm not sure if this PR is a good place to discuss it, but this PR does make patchelf in rustup more and more complicated.

I don't have any good ideas here.

DianQK:
Writing such test cases on rustup isn't easy. I read through some test code, and it seems like rustup's internal tests only focus on rustup's own behavior. I can only add this test in nixpkgs, but I'm not familiar with nix. Perhaps we should create an additional package or add a test script directly in the `postInstall`.

Mic92:
> Writing such test cases on rustup isn't easy. I read through some test code, and it seems like rustup's internal tests only focus on rustup's own behavior. I can only add this test in nixpkgs, but I'm not familiar with nix. Perhaps we should create an additional package or add a test script directly in the `postInstall`.

Maybe write a test that just uses nix for dependencies but has internet access. Than we can at least test manually when we change things.

DianQK:
@Mic92 @NickCao 
Could you help me add a backport label? `rustc` will switch the default linker to lld in the next stable version.
We can continue to improve the garbage collection issue in `rustup`.

PR #314270, opened 2024-05-24
qemu: vendor subprojects locally

Mic92:
i.e. when building on arm64, we need dtc, which is not available in the tarball.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [x] aarch64-linux (pkgsCross.aarch64-multiplatform)
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/applications/virtualization/qemu/default.nix
+++ b/pkgs/applications/virtualization/qemu/default.nix
@@ -1,4 +1,5 @@
-{ lib, stdenv, fetchurl, fetchpatch, python3Packages, zlib, pkg-config, glib, buildPackages
+{ lib, stdenv, fetchzip, fetchpatch, python3Packages, zlib, pkg-config, glib, buildPackages
+, git, cacert
 , pixman, vde2, alsa-lib, texinfo, flex
 , bison, lzo, snappy, libaio, libtasn1, gnutls, nettle, curl, dtc, ninja, meson
 , sigtool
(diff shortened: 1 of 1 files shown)


Mic92:
I also got reports that aarch64 can no longer build qemu... So we should probably also make this part of the next NixOS release?

brianmcgillion:
verified to build again in 3 scenarios
x-compile x86_64 -> aarch64 for nvidia jetson agx target
native compile : x86_64 and aarch64

pkgs/applications/virtualization/qemu/default.nix
```diff
@@ -57,9 +58,22 @@ stdenv.mkDerivation (finalAttrs: {
...
+  # So instead we just download all subprojects.
+  src = fetchzip {
     url = "https://download.qemu.org/qemu-${finalAttrs.version}.tar.xz";
-    hash = "sha256-7PVTf+q5JkG5nXSC9VHyGV06W9NKzvnVK/v/NTpgc5c=";
+    hash = "sha256-PY/I4qCQj6Q0n8iFzQSHMGimMk0T6o8QAzUGQjF70YI=";
+    nativeBuildInputs = [ cacert git meson ];
+    postFetch = ''
+      (
+        cd "$out"
+        for prj in subprojects/*.wrap; do
```
nagy, line 71: According to the error message in the upstream `configure` script ( https://github.com/qemu/qemu/blob/v8.2.4/configure#L782 ) it should be enough to invoke `meson subprojects download` once in the top-level folder. This has worked well for me and then we would not need this for-loop. Does that work for your case as well?

alyssais: In FODs it's important to ensure we depend on the behaviour of fetcher programs as little as possible, because an update to e.g. Meson that changes how it lays things out would break hashes.  So it's best to do as much ourselves as possible.

Mic92: meson could be moved out, but we still need to remove `.git` as it is not binary reproducible.

alyssais:
We don't need to have QEMU build its own dtc — we just need to provide it as a build input: #314806.  Presumably that's why upstream doesn't include this in its tarball.  I think it's mostly there as a development aid.

PR #314377, opened 2024-05-24
temurin-bin: init at 22, update sources

Infinidoge:
## Description of changes

Split from #286267

CCing @thiagokokada @philiptaron as previous reviewers

Adds temurin bin 22, and regenerates the sources file

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/top-level/java-packages.nix
+++ b/pkgs/top-level/java-packages.nix
@@ -227,7 +227,7 @@ in {
       ../development/compilers/openjdk/22.nix
       ../development/compilers/zulu/22.nix
       {
-        openjdk22-bootstrap = temurin-bin.jdk-21;
+        openjdk22-bootstrap = temurin-bin.jdk-22;
         openjfx = openjfx22;
       };
 
--- a/pkgs/development/compilers/temurin-bin/generate-sources.py
+++ b/pkgs/development/compilers/temurin-bin/generate-sources.py
@@ -6,7 +6,7 @@ import re
 import requests
 import sys
 
-feature_versions = (8, 11, 16, 17, 18, 19, 20, 21)
+feature_versions = (8, 11, 16, 17, 18, 19, 20, 21, 22)
 oses = ("mac", "linux", "alpine-linux")
 types = ("jre", "jdk")
 impls = ("hotspot",)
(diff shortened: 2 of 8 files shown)


Infinidoge:
Retargeted to staging due to the rebuild count

philiptaron (approved):

pkgs/top-level/java-packages.nix
```diff
@@ -227,7 +227,7 @@ in {
       ../development/compilers/openjdk/22.nix
       ../development/compilers/zulu/22.nix
       {
-        openjdk22-bootstrap = temurin-bin.jdk-21;
+        openjdk22-bootstrap = temurin-bin.jdk-22;
```
jlesquembre, line 230: Not a big deal, but I think we should keep the version at 21 here, to align with the recommended version in the [OpenJDK build documentation](https://openjdk.org/groups/build/doc/building.html#boot-jdk-requirements):
> ... the boot JDK for building JDK major version N should be a JDK of major version N-1

Infinidoge: All recent versions of Java build with major version N, so this keeps it consistent.

raboof: Do we know why? Seems to make more sense to bootstrap with `N-1` tbh...

raboof (approved):
bit weird with the bootstrap, but LGTM nonetheless

pkgs/top-level/java-packages.nix
jlesquembre: If the concern is to keep it consistent, then we should use `N-1` everywhere. Using the same version introduces a cyclic dependency. Also, it makes sense to follow the recommendation from OpenJDK build docs.

Infinidoge: I'm not sure how it introduces a cyclic dependency, because the binary package has no dependency on Java.

IMO LTS releases should definitely be built with the `N` version, because otherwise it introduces a dependency on a package that will be marked insecure. Non-LTS, either way works.

jlesquembre: I hadn't considered the dependency on an EOL JDK. That's a good point.

 By cyclic dependency, I mean that Temurin depends on the same OpenJDK source code that the derivation does.

jlesquembre (approved):

PR #314450, opened 2024-05-25
koboldcpp: init at 1.66

### LINKED ISSUE #295900: Package request: koboldcpp
**Project description**

A simple one-file way to run various GGML and GGUF models with KoboldAI's UI.

**Metadata**

* homepage URL: https://github.com/LostRuins/koboldcpp
* source URL: https://github.com/LostRuins/koboldcpp
* license: GNU Affero General Public License v3.0
* platforms: darwin, linux

---

maxstrid:
## Description of changes

Closes #295900

Adds koboldcpp, a llama.cpp fork which makes running models with various backends a bit easier. Currently untested on CUDA and Apple silicon.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [x] `sandbox = relaxed`
  - [x] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/by-name/ko/koboldcpp/package.nix
+++ b/pkgs/by-name/ko/koboldcpp/package.nix
@@ -1,0 +1,154 @@
+{
+  lib,
+  fetchFromGitHub,
+  stdenv,
+  makeWrapper,
+  gitUpdater,
+  python311Packages,
+  tk,
+
+  darwin,
+
+  koboldLiteSupport ? true,
+
+  config,
+  cudaPackages ? { },
+
+  openblasSupport ? !stdenv.isDarwin,
+  openblas,
+
+  cublasSupport ? config.cudaSupport,
+
+  clblastSupport ? stdenv.isLinux,
+  clblast,
+  ocl-icd,
+
+  vulkanSupport ? true,
+  vulkan-loader,
+
+  metalSupport ? stdenv.isDarwin && stdenv.isAarch64,
+
+  # You can find list of x86_64 options here: https://gcc.gnu.org/onlinedocs/gcc/x86-Options.html
+  # For ARM here: https://gcc.gnu.org/onlinedocs/gcc/ARM-Options.html
+  # If you set "march" to "native", specify "mtune" as well; otherwise, it will be set to "generic". (credit to: https://lemire.me/blog/2018/07/25/it-is-more-complicated-than-i-thought-mtune-march-in-gcc/)
+  # For example, if you have an AMD Ryzen CPU, you will set "march" to "x86-64" and "mtune" to "znver2"
+  march ? "",
+  mtune ? "",
+}:
+
+let
+  makeBool = option: bool: (if bool then "${option}=1" else "");
+
+  effectiveStdenv = if cublasSupport then cudaPackages.backendStdenv else stdenv;
+in
+effectiveStdenv.mkDerivation (finalAttrs: {
+  pname = "koboldcpp";
+  version = "1.66";
+
+  src = fetchFromGitHub {
+    owner = "LostRuins";
+    repo = "koboldcpp";
+    rev = "refs/tags/v${finalAttrs.version}";
+    sha256 = "sha256-3Gsbfxh3K3xa6Ofs9M9cfKWg0sxrolL8Ivx40iqvlA8=";
+  };
+
+  enableParallelBuilding = true;
+
+  nativeBuildInputs = [
+    makeWrapper
+    python311Packages.wrapPython
+  ];
+
+  pythonInputs = builtins.attrValues { inherit (python311Packages) tkinter customtkinter packaging; };
+
+  buildInputs =
+    [ tk ]
+    ++ finalAttrs.pythonInputs
+    ++ lib.optionals effectiveStdenv.isDarwin [
+      darwin.apple_sdk_11_0.frameworks.Accelerate
+      darwin.apple_sdk_11_0.frameworks.CoreVideo
+      darwin.apple_sdk_11_0.frameworks.CoreGraphics
+      darwin.apple_sdk_11_0.frameworks.CoreServices
+    ]
+    ++ lib.optionals metalSupport [
+      darwin.apple_sdk_11_0.frameworks.MetalKit
+      darwin.apple_sdk_11_0.frameworks.Foundation
+      darwin.apple_sdk_11_0.frameworks.MetalPerformanceShaders
+    ]
+    ++ lib.optionals openblasSupport [ openblas ]
+    ++ lib.optionals cublasSupport [
+      cudaPackages.libcublas
+      cudaPackages.cuda_nvcc
+      cudaPackages.cuda_cudart
+      cudaPackages.cuda_cccl
+    ]
+    ++ lib.optionals clblastSupport [
+      clblast
+      ocl-icd
+    ]
+    ++ lib.optionals vulkanSupport [ vulkan-loader ];
+
+  pythonPath = finalAttrs.pythonInputs;
+
+  darwinLdFlags = lib.optionals stdenv.isDarwin [
+    "-F${darwin.apple_sdk_11_0.frameworks.CoreServices}/Library/Frameworks"
+    "-F${darwin.apple_sdk_11_0.frameworks.Accelerate}/Library/Frameworks"
+    "-framework CoreServices"
+    "-framework Accelerate"
+  ];
+  metalLdFlags = lib.optionals metalSupport [
+    "-F${darwin.apple_sdk_11_0.frameworks.Foundation}/Library/Frameworks"
+    "-F${darwin.apple_sdk_11_0.frameworks.Metal}/Library/Frameworks"
+    "-framework Foundation"
+    "-framework Metal"
+  ];
+
+  env.NIX_LDFLAGS = lib.concatStringsSep " " (finalAttrs.darwinLdFlags ++ finalAttrs.metalLdFlags);
+
+  env.NIX_CFLAGS_COMPILE =
+    lib.optionalString (march != "") "-march=${march}" + lib.optionalString (mtune != "") "-mtune=${mtune}";
+
+  makeFlags = [
+    (makeBool "LLAMA_OPENBLAS" openblasSupport)
+    (makeBool "LLAMA_CUBLAS" cublasSupport)
+    (makeBool "LLAMA_CLBLAST" clblastSupport)
+    (makeBool "LLAMA_VULKAN" vulkanSupport)
+    (makeBool "LLAMA_METAL" metalSupport)
+  ];
+
+  installPhase = ''
+    runHook preInstall
+
+    mkdir -p "$out/bin"
+
+    install -Dm755 koboldcpp.py "$out/bin/koboldcpp.unwrapped"
+    cp *.so "$out/bin"
+    cp *.embd "$out/bin"
+
+    ${lib.optionalString (!koboldLiteSupport) ''
+      rm "$out/bin/kcpp_docs.embd"
+      rm "$out/bin/klite.embd"
+    ''}
+
+    runHook postInstall
+  '';
+
+  postFixup = ''
+    wrapPythonProgramsIn "$out/bin" "$pythonPath"
+    makeWrapper "$out/bin/koboldcpp.unwrapped" "$out/bin/koboldcpp" \
+    --prefix PATH ${lib.makeBinPath [ tk ]}
+  '';
+
+  passthru.updateScript = gitUpdater { rev-prefix = "v"; };
+
+  meta = {
+    description = "A way to run various GGML and GGUF models";
+    license = lib.licenses.agpl3Only;
+    maintainers = with lib.maintainers; [
+      maxstrid
+      donteatoreo
+    ];
+    mainProgram = "koboldcpp";
+    platforms = lib.platforms.unix;
+  };
+})
--- a/maintainers/maintainer-list.nix
+++ b/maintainers/maintainer-list.nix
@@ -12827,6 +12827,12 @@
       fingerprint = "1DE4 424D BF77 1192 5DC4  CF5E 9AED 8814 81D8 444E";
     }];
   };
+  maxstrid = {
+    email = "mxwhenderson@gmail.com";
+    github = "maxstrid";
+    githubId = 115441224;
+    name = "Maxwell Henderson";
+  };
   maxux = {
     email = "root@maxux.net";
     github = "maxux";


DontEatOreo:
I own a `aarch64-darwin` device so I would be able to test this package.

I would like to note that for `aarch64-darwin` builds, you should disable OpenBLAS as it's somewhat useless since it's very slow, and all `aarch64-darwin` devices come with Metal Support.

For ROCm, we should use <https://github.com/YellowRoseCx/koboldcpp-rocm> (as mentioned in <https://github.com/LostRuins/koboldcpp/tree/v1.66?tab=readme-ov-file>).

pkgs/by-name/ko/koboldcpp/package.nix
```diff
@@ -0,0 +1,98 @@
...
+    ]}
+  '';
+
+  meta = with lib; {
+    description = "A way to run various GGML and GGUF models";
+    license = licenses.agpl3Only;
+    mainProgram = "koboldcpp";
+    platforms = platforms.unix;
+    maintainers = with maintainers; [ maxstrid ];
+  };
```
DontEatOreo, line 96:
```suggestion
  meta = {
    description = "A way to run various GGML and GGUF models";
    license = lib.licenses.agpl3Only;
    maintainers = with lib.maintainers; [ maxstrid ];
    mainProgram = "koboldcpp";
    platforms = lib.platforms.unix;
  };
```

pkgs/by-name/ko/koboldcpp/package.nix
```diff
@@ -0,0 +1,98 @@
...
+, cublasSupport ? config.cudaSupport
+
+, clblastSupport ? true
+, clblast
+, ocl-icd
+
+, vulkanSupport ? true
+, vulkan-loader
+
+, metalSupport ? stdenv.isDarwin && stdenv.isAarch64 && !openblasSupport
```
DontEatOreo, line 24:
```suggestion
, openblasSupport ? !stdenv.isDarwin
, openblas

, cublasSupport ? config.cudaSupport

, clblastSupport ? true
, clblast
, ocl-icd

, vulkanSupport ? true
, vulkan-loader

, metalSupport ? stdenv.isDarwin && stdenv.isAarch64
```

DontEatOreo:
I would also suggest adding an optional option for `march` and `mtune`

```nix
# You can find list of x86_64 options here: https://gcc.gnu.org/onlinedocs/gcc/x86-Options.html
# For ARM here: https://gcc.gnu.org/onlinedocs/gcc/ARM-Options.html
# If you set "march" to "native", specify "mtune" as well; otherwise, it will be set to "generic". (credit to: https://lemire.me/blog/2018/07/25/it-is-more-complicated-than-i-thought-mtune-march-in-gcc/)
# For example, if you have an AMD Ryzen CPU, you will set "march" to "x86-64" and "mtune" to "znver2"
march ? "",
mtune ? "",
```

pkgs/by-name/ko/koboldcpp/package.nix
```diff
@@ -0,0 +1,98 @@
...
+  ++ optionals vulkanSupport [ "LLAMA_VULKAN=1" ]
+  ++ optionals metalSupport [ "LLAMA_METAL=1" ];
+
+  installPhase = ''
+    runHook preInstall
+
+    mkdir -p $out/run-env
+
+    install -Dm755 koboldcpp.py $out/run-env/koboldcpp-unwrapped
+    cp *.so $out/run-env
```
DontEatOreo, line 77:
```suggestion
    mkdir -p "$out/run-env"

    install -Dm755 koboldcpp.py "$out/run-env/koboldcpp-unwrapped"
    cp *.so "$out/run-env"
```

You should move all `*.embd` and `quantize_*` files as well

pkgs/by-name/ko/koboldcpp/package.nix
DontEatOreo, line 1: You could add an `koboldLiteSupport` option so the Web UI is optional. 

To "disable" the Web UI, you just need to **not** move over the `klite.embd` and `kcpp_docs.embd` files into `$out/run-env`

pkgs/by-name/ko/koboldcpp/package.nix
```diff
@@ -0,0 +1,98 @@
...
+
+, config
+, cudaPackages ? { }
+
+, openblasSupport ? true
+, openblas
+
+, cublasSupport ? config.cudaSupport
+
+, clblastSupport ? true
```
DontEatOreo, line 17:
```suggestion
, clblastSupport ? stdenv.isLinux
```

`clblast` is only available on `platforms.linux` https://github.com/NixOS/nixpkgs/blob/5710852ba686cc1fd0d3b8e22b3117d43ba374c2/pkgs/development/libraries/science/math/clblast/default.nix#L42

DontEatOreo:
Pleas rebase and squash e3d0e64bba9194b0ce40d3abb19c9df96699ac35 into e9539623e612e3aaf898cf0cdc4be0789211afae

pkgs/by-name/ko/koboldcpp/package.nix
maxstrid: The reason I did `run-env` is that koboldcpp needs the koboldcpp_default.so file to be in the same directory.

I wasn't sure if it was proper to leave the shared object file in `bin` so I put them both into a separate directory and put the wrapper in `bin`. But I don't think this is supposed to be the proper solution here.

DontEatOreo: You should use `$out/bin` instead of `$out/run-env` as there is no mention or usage of `run-env` in the manual or anywhere on nixpkgs

All `embd` and `so` files should be in same directory as the `koboldcpp.py`

pkgs/by-name/ko/koboldcpp/package.nix
maxstrid: Would there be any reason to make the Web UI optional? The [koboldcpp docs](https://github.com/LostRuins/koboldcpp/wiki#what-is-kobold-lite-how-do-i-use-it) treat it as something thats packaged by default into koboldcpp.

DontEatOreo: I would say it's nice to have option. It should be enabled by default, but I think it's a good idea to give the choice to not use the Web UI if someone just wants to use `koboldcpp` as an API instance

pkgs/by-name/ko/koboldcpp/package.nix
```diff
@@ -0,0 +1,98 @@
...
+      tk
+    ]}
+  '';
+
+  meta = with lib; {
+    description = "A way to run various GGML and GGUF models";
+    license = licenses.agpl3Only;
+    mainProgram = "koboldcpp";
+    platforms = platforms.unix;
+    maintainers = with maintainers; [ maxstrid ];
```
DontEatOreo, line 95:
```suggestion
    maintainers = with maintainers; [ maxstrid donteatoreo ];
```

I would like to be a maintainer for `koboldcpp` as well

maxstrid:
For rocm would I just change the package to be koboldcpp-rocm when rocmSupport is enabled?

pkgs/by-name/ko/koboldcpp/package.nix
```diff
@@ -0,0 +1,115 @@
...
+, clblast
+, ocl-icd
+
+, vulkanSupport ? true
+, vulkan-loader
+
+, metalSupport ? stdenv.isDarwin && stdenv.isAarch64
+
+, march ? ""
+, mtune ? ""
```
DontEatOreo, line 29: Shouldn't we leave a comment here like this one:

```
# You can find list of x86_64 options here: https://gcc.gnu.org/onlinedocs/gcc/x86-Options.html
# For ARM here: https://gcc.gnu.org/onlinedocs/gcc/ARM-Options.html
# If you set "march" to "native", specify "mtune" as well; otherwise, it will be set to "generic". (credit to: https://lemire.me/blog/2018/07/25/it-is-more-complicated-than-i-thought-mtune-march-in-gcc/)
# For example, if you have an AMD Ryzen CPU, you will set "march" to "x86-64" and "mtune" to "znver2"
```

Showing what options are possible?

pkgs/by-name/ko/koboldcpp/package.nix
```diff
@@ -0,0 +1,115 @@
...
+
+    install -Dm755 koboldcpp.py "$out/bin/koboldcpp.unwrapped"
+    cp *.so "$out/bin"
+    cp *.embd "$out/bin"
+
+    if [ ${lib.boolToString koboldLiteSupport} == "false" ]; then
+      echo "koboldLiteSupport disabled, removing appropriate filew."
+      rm "$out/bin/kcpp_docs.embd"
+      rm "$out/bin/klite.embd"
+    fi
```
DontEatOreo, line 94:
```suggestion
    ${optionalString (!koboldLiteSupport) ''
      rm "$out/bin/kcpp_docs.embd"
      rm "$out/bin/klite.embd"
    ''}
```
I think this is a better approach. Also, since the option is enabled by default, I don't think it's worth leaving a comment with `echo`

pkgs/by-name/ko/koboldcpp/package.nix
```diff
@@ -0,0 +1,115 @@
...
+  pythonInputs = with python311Packages; [
+    tkinter
+    customtkinter
+    packaging
+    tk
+  ];
+
+  cudaBuildInputs = with cudaPackages; [
+      libcublas
+  ];
```
DontEatOreo, line 44:
```suggestion
  pythonInputs = builtins.attrValues { inherit (python311Packages) tkinter customtkinter packaging; };

  cudaBuildInputs = builtins.attrValues { inherit (cudaPackages) cudaPackages libcublas; };
```

https://nix.dev/guides/best-practices#with-scopes

There is no `python311Packages.tk`?

DontEatOreo:
> For rocm would I just change the package to be koboldcpp-rocm when rocmSupport is enabled?

`src = fetchFromGitHub` would just be conditional

Maybe something like this?

```nix
src = if cublasSupport then srcDefault else srcROCm;

srcDefault = fetchFromGitHub {
  owner = "LostRuins";
  repo = "koboldcpp";
  sha256 = "sha256-LAAINN49gjJPye3rfbA8oh29/qcVVk99SSRZD8WQNKQ=";
};

srcROCm = fetchFromGitHub {
  owner = "YellowRoseCx";
  repo = "koboldcpp-rocm";
  rev = "refs/tags/v${version}.yr1-ROCm";
  sha256 = "sha256-AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA=";
};
```

4evy (changes requested):

DontEatOreo:
For macOS Support you will need to add `CoreServices` and `Accelerate`

For Metal support you will need to add `Foundation` and `Metal`

To the `NIX_CFLAGS_COMPILE`

https://github.com/NixOS/nixpkgs/blob/e10e8912a8b8681c43857dcd9948e4499d14f80d/pkgs/by-name/ko/koboldcpp/package.nix#L99-L110

maxstrid:
I've tested the package compiles with CUDA, I'm having a hard time getting `koboldcpp-rocm` to build with `hipblas`. Is there a good place to reach you to get your insight on this? I want to avoid flooding the comments with debugging info.

DontEatOreo:
If you want you can join the [NixOS Unofficial Discord](https://discord.com/invite/RbvHtGa) with same name as your GitHub and I will reach out to you

maxstrid:
Result of `nixpkgs-review pr 314450` run on x86_64-linux [1](https://github.com/Mic92/nixpkgs-review)
<details>
  <ul>
    <li>koboldcpp</li>
  </ul>
</details>

DontEatOreo:
Result of `nixpkgs-review pr 314450` run on aarch64-darwin [1](https://github.com/Mic92/nixpkgs-review)
<details>
  <ul>
    <li>koboldcpp</li>
  </ul>
</details>

4evy (approved):
LGTM!

4evy (approved):

nicoonoclaste (approved):

pkgs/by-name/ko/koboldcpp/package.nix
```diff
@@ -0,0 +1,154 @@
...
+    mkdir -p "$out/bin"
+
+    install -Dm755 koboldcpp.py "$out/bin/koboldcpp.unwrapped"
+    cp *.so "$out/bin"
+    cp *.embd "$out/bin"
+
+    ${lib.optionalString (!koboldLiteSupport) ''
+      rm "$out/bin/kcpp_docs.embd"
+      rm "$out/bin/klite.embd"
+    ''}
```
SuperSandro2000, line 131:
```suggestion
  '' + lib.optionalString (!koboldLiteSupport) ''
    rm "$out/bin/kcpp_docs.embd"
    rm "$out/bin/klite.embd"
  '' + ''
```
FYI this is the clean way to write this

pkgs/by-name/ko/koboldcpp/package.nix
```diff
@@ -0,0 +1,154 @@
...
+  };
+
+  enableParallelBuilding = true;
+
+  nativeBuildInputs = [
+    makeWrapper
+    python311Packages.wrapPython
+  ];
+
+  pythonInputs = builtins.attrValues { inherit (python311Packages) tkinter customtkinter packaging; };
```
SuperSandro2000, line 62: Does this exactly need to be python311? If not python3 should be just used

pkgs/by-name/ko/koboldcpp/package.nix
DontEatOreo: In my version of the package, I actually had this initially but thought about it, in the opposite way. Why isn't string interpolation the better way?

SuperSandro2000: Not really, Doing it like my suggestion uses less code and string interpolation has usually some edge cases and is generally more complex in multiline strings.

KiruyaMomochi:
The CUDA support is not work as expected - as we need to add `autoAddDriverRunpath` hook to `nativeBuildInputs`.

However, only adding the hook is not enough, because by default it uses `-arch=native`.
We may need to set `LLAMA_PORTABLE`, or set `CUDA_DOCKER_ARCH` according to `config.cudaCapabilities`. I'm not sure if the latter one is possible or not without patching there Makefile. 

https://github.com/LostRuins/koboldcpp/blob/1487a4bc812d66654a6c0049bd72e227ba6e4627/Makefile#L166-L178

```makefile
ifdef CUDA_DOCKER_ARCH
	NVCCFLAGS += -Wno-deprecated-gpu-targets -arch=$(CUDA_DOCKER_ARCH)
else
ifdef LLAMA_PORTABLE
ifdef LLAMA_COLAB #colab does not need all targets, all-major doesnt work correctly with pascal
	NVCCFLAGS += -Wno-deprecated-gpu-targets -arch=all-major
else
	NVCCFLAGS += -Wno-deprecated-gpu-targets -arch=all
endif #LLAMA_COLAB
else
	NVCCFLAGS += -arch=native
endif #LLAMA_PORTABLE
endif # CUDA_DOCKER_ARCH
```

pkgs/by-name/ko/koboldcpp/package.nix
DontEatOreo: Yea, it should be `python3`, I suppose

DontEatOreo:
@KiruyaMomochi Do you still face this issue? I tested out `koboldcpp` yesterday *(version 1.68)* with CUDA and everything worked as expected

mirrexagon:
@DontEatOreo I'm not sure if it is the same issue as @KiruyaMomochi is reporting, but for me, attempting to run koboldcpp with `--usecublas` gives me a message like this whenever it attempts to use CUDA:

```
ggml_cuda_init: failed to initialize CUDA: CUDA driver is a stub library
```

I notice that this issue appears to be similar: https://github.com/NixOS/nixpkgs/issues/320145
And it was fixed in the linked PR involving `addDriverRunpath`: https://github.com/NixOS/nixpkgs/pull/324387

I don't yet know enough about how that works to fix it here.

KiruyaMomochi:
@DontEatOreo Apologize for my tardy reply. I'm still facing this issue. I need to add `addDriverRunpath` and define `CUDA_DOCKER_ARCH`. I have already set `hardware.graphics.enable`.

DontEatOreo:
@KiruyaMomochi @mirrexagon I will later today (or tomorrow) take a closer look and add `addDriverRunpath` and `CUDA_DOCKER_ARCH` and make a PR. Thanks for the feedback!

PR #314670, opened 2024-05-25
gettext, texinfo: Remove spurious xz 

rhelmot:
## Description of changes

If xz is overridden with a derivation that does not have an explicit bin output, these will fail. These changed instances are the ones impacted by the FreeBSD stdenv providing a bootstrap-files xz for early boot.

diffoscope indicates that the builds with and without xz are identical, so just remove.

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [x] x86_64-freebsd
  - [ ] `sandbox = relaxed`
  - [x] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/development/libraries/gettext/default.nix
+++ b/pkgs/development/libraries/gettext/default.nix
@@ -1,4 +1,4 @@
-{ stdenv, lib, fetchurl, libiconv, xz, bash
+{ stdenv, lib, fetchurl, libiconv, bash
 }:
 
 # Note: this package is used for bootstrapping fetchurl, and thus
@@ -28,7 +28,7 @@ stdenv.mkDerivation rec {
   LDFLAGS = lib.optionalString stdenv.isSunOS "-lm -lmd -lmp -luutil -lnvpair -lnsl -lidmap -lavl -lsec";
 
   configureFlags = [
-     "--disable-csharp" "--with-xz"
+     "--disable-csharp"
   ] ++ lib.optionals (stdenv.hostPlatform != stdenv.buildPlatform) [
     # On cross building, gettext supposes that the wchar.h from libc
     # does not fulfill gettext needs, so it tries to work with its
@@ -50,10 +50,6 @@ stdenv.mkDerivation rec {
   '';
 
   strictDeps = true;
-  nativeBuildInputs = [
-    xz
-    xz.bin
-  ];
   buildInputs = lib.optionals (!stdenv.hostPlatform.isMinGW) [
     bash
   ]
--- a/pkgs/development/tools/misc/texinfo/common.nix
+++ b/pkgs/development/tools/misc/texinfo/common.nix
@@ -1,6 +1,6 @@
 { version, sha256, patches ? [] }:
 
-{ lib, stdenv, buildPackages, fetchurl, perl, xz, libintl, bash
+{ lib, stdenv, buildPackages, fetchurl, perl, libintl, bash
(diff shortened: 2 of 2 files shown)


wolfgangwalther (approved):

pkgs/development/libraries/gettext/default.nix
```diff
@@ -52,7 +52,7 @@ stdenv.mkDerivation rec {
   strictDeps = true;
   nativeBuildInputs = [
     xz
-    xz.bin
+    (lib.getBin xz)
```
SuperSandro2000, line 55: Isn't this automatically chosen here?

pkgs/development/tools/misc/texinfo/common.nix
```diff
@@ -49,7 +49,7 @@ stdenv.mkDerivation {
   # A native compiler is needed to build tools needed at build time
   depsBuildBuild = [ buildPackages.stdenv.cc perl ];
 
-  buildInputs = [ xz.bin bash libintl ]
+  buildInputs = [ (lib.getBin xz) bash libintl ]
```
SuperSandro2000, line 52: This seems wrong to me especially for cross compilation

Ericson2314: I think @SuperSandro2000 means this should be a `nativeBuildInput`

Ericson2314 (changes requested):
Please put the description in the commit messages (goes for many of the other PRs too).

I think it's OK to just leave the `texinfo` one as-is with the suspicious use of `buildInputs` as that's an orthogonal problem, but if you are feeling motivated you could add a `strictDeps = true` and see if `nativeBuildInputs` is better (simulating cross).

pkgs/development/tools/misc/texinfo/common.nix
SuperSandro2000: > Is your comment about `xz.bin` in the first place

yes

SuperSandro2000: We don't need that here anymore. At the time the tarball was in the xz format but now we have gz.

pkgs/development/libraries/gettext/default.nix
SuperSandro2000: We don't need that here anymore. At the time the tarball was in the xz format but now we have gz and strictDeps is already set.

```suggestion
  buildInputs = [ bash libintl ]
```

The --with-xz seems to be not properly checked and for that to work we would need to move xz to buildInputs

pkgs/development/tools/misc/texinfo/common.nix
SuperSandro2000: We don't need that here anymore. At the time the tarball was in the xz format but now we have gz and strictDeps is already set.

pkgs/development/libraries/gettext/default.nix
SuperSandro2000, line 55:
```suggestion
```
I've run diffoscope on a gettext build with and without xz and booth are equal. I think we can just drop xz entirely from here.

and we can also drop the ``"--with-xz"`` arg in line 31.

rhelmot:
I have changed the commits as per feedback, removing both uses!

Ericson2314 (approved):

PR #312934, opened 2024-05-19
haskellPackages: update stackage and hackage

ncfavier:
### This Merge

This PR is the regular merge of the `haskell-updates` branch into `master`.

This branch is being continually built and tested by hydra at https://hydra.nixos.org/jobset/nixpkgs/haskell-updates. You may be able to find an up-to-date Hydra build report at [cdepillabout/nix-haskell-updates-status](https://github.com/cdepillabout/nix-haskell-updates-status).

We roughly aim to merge these `haskell-updates` PRs at least once every two weeks. See the @NixOS/haskell [team calendar](https://cloud.maralorn.de/apps/calendar/p/H6migHmKX7xHoTFa) for who is currently in charge of this branch.

### haskellPackages Workflow Summary

Our workflow is currently described in [`pkgs/development/haskell-modules/HACKING.md`](https://github.com/NixOS/nixpkgs/blob/haskell-updates/pkgs/development/haskell-modules/HACKING.md).

The short version is this:
* We regularly update the Stackage and Hackage pins on `haskell-updates` (normally at the beginning of a merge window).
* The community fixes builds of Haskell packages on that branch.
* We aim at at least one merge of `haskell-updates` into `master` every two weeks.
* We only do the merge if the [`mergeable`](https://hydra.nixos.org/job/nixpkgs/haskell-updates/mergeable) job is succeeding on hydra.
* If a [`maintained`](https://hydra.nixos.org/job/nixpkgs/haskell-updates/maintained) package is still broken at the time of merge, we will only merge if the maintainer has been pinged 7 days in advance. (If you care about a Haskell package, become a maintainer!)

More information about Haskell packages in nixpkgs can be found [in the nixpkgs manual](https://nixos.org/manual/nixpkgs/unstable/#haskell).

---

This is the follow-up to #311940. Come to [#haskell:nixos.org](https://matrix.to/#/#haskell:nixos.org) if you have any questions.

PR #314679, opened 2024-05-25
[24.05] Haskell backports (cabal2nix, GHC expression refactor)

sternenseemann:
## Description of changes

Backport of #312934. We mainly want

- The GHC expression refactors (which only change patch order / whitespace), so we can backport future GHC changes.
- A new (minor) version of distribution-nixpkgs, so cabal2nix deals with `javascript-unknown-ghcjs` as it should. Sadly, this requires a Hackage bump and some unrelated changes to keep the package set working properly.


maralorn (approved):
I agree this is a good idea.

PR #314716, opened 2024-05-25
gnumake: do not use MAKE_CXX

tie:
## Description of changes

Removes unnecessary C++ compiler reference when CXX environment variable is set to an absolute path.

See also https://savannah.gnu.org/bugs/?63668

**Parent PR:**
- #314920

## Motivation

I’m trying to fix cross-compilation for platforms where C/C++ compiler’s `targetPrefix` is identical. It’s common to set
```
depsBuildBuild = [ buildPackages.stdenv.cc ];
```
when cross-compiling, however, cc/bintools wrapper setup hook currently sets {CC,CXX}{,_FOR_BUILD} et cetera to program name instead of absolute path. That obviously doesn’t work because only the first program from `PATH` is used (for example, this is part of the reason why static glibc doesn’t work in Nixpkgs and we have pkgsStatic use musl that has different target prefix). There are a lot of other things that start falling apart in stdenv when we set these variables to absolute paths, this is just one of them: gnumake is part of the stdenv and shouldn’t refer to the bootstrap tools.

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [x] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/development/tools/build-managers/gnumake/default.nix
+++ b/pkgs/development/tools/build-managers/gnumake/default.nix
@@ -22,17 +22,18 @@ stdenv.mkDerivation rec {
     sha256 = "sha256-3Rb7HWe/q3mnL16DkHNcSePo5wtJRaFasfgd23hlj7M=";
   };
 
-  # to update apply these patches with `git am *.patch` to https://git.savannah.gnu.org/git/make.git
-  patches = [
-    # See patch message.
-    ./make-cxx.patch
-    # Replaces /bin/sh with sh, see patch file for reasoning
-    ./0001-No-impure-bin-sh.patch
-    # Purity: don't look for library dependencies (of the form `-lfoo') in /lib
-    # and /usr/lib. It's a stupid feature anyway. Likewise, when searching for
-    # included Makefiles, don't look in /usr/include and friends.
-    ./0002-remove-impure-dirs.patch
-  ];
+  # To update patches:
+  #  $ version=4.4.1
+  #  $ git clone https://git.savannah.gnu.org/git/make.git
+  #  $ cd make && git checkout -b nixpkgs $version
+  #  $ git am --directory=../patches
+  #  $ # make changes, resolve conflicts, etc.
+  #  $ git format-patch --output-directory ../patches --diff-algorithm=histogram $version
+  #
+  # TODO: stdenv’s setup.sh should be aware of patch directories. It’s very
+  # convenient to keep them in a separate directory but we can defer listing the
+  # directory until derivation realization to avoid unnecessary Nix evaluations.
+  patches = lib.filesystem.listFilesRecursive ./patches;
 
   nativeBuildInputs = [ autoreconfHook pkg-config ];
   buildInputs = lib.optionals guileEnabled [ guile ];
--- a/pkgs/development/tools/build-managers/gnumake/default.nix
+++ b/pkgs/development/tools/build-managers/gnumake/default.nix
@@ -1,7 +1,7 @@
 { lib
 , stdenv
 , fetchurl
-, updateAutotoolsGnuConfigScriptsHook
+, autoreconfHook
 , guileSupport ? false, guile
 # avoid guile depend on bootstrap to prevent dependency cycles
 , inBootstrap ? false
@@ -24,6 +24,8 @@ stdenv.mkDerivation rec {
 
   # to update apply these patches with `git am *.patch` to https://git.savannah.gnu.org/git/make.git
   patches = [
+    # See patch message.
+    ./make-cxx.patch
     # Replaces /bin/sh with sh, see patch file for reasoning
     ./0001-No-impure-bin-sh.patch
     # Purity: don't look for library dependencies (of the form `-lfoo') in /lib
@@ -32,7 +34,7 @@ stdenv.mkDerivation rec {
     ./0002-remove-impure-dirs.patch
   ];
 
-  nativeBuildInputs = [ updateAutotoolsGnuConfigScriptsHook ] ++ lib.optionals guileEnabled [ pkg-config ];
+  nativeBuildInputs = [ autoreconfHook pkg-config ];
   buildInputs = lib.optionals guileEnabled [ guile ];
 
   configureFlags = lib.optional guileEnabled "--with-guile"
--- a/pkgs/development/tools/build-managers/gnumake/patches/0002-Remove-impure-dirs.patch
+++ b/pkgs/development/tools/build-managers/gnumake/patches/0002-Remove-impure-dirs.patch
@@ -1,20 +1,24 @@
-From 795d63d3c8b5c0dbb7e544954f75507b371b7228 Mon Sep 17 00:00:00 2001
+From 2db52008be2e2d504889f4f19318c2ba5a2a4797 Mon Sep 17 00:00:00 2001
 From: =?UTF-8?q?J=C3=B6rg=20Thalheim?= <joerg@thalheim.io>
 Date: Sat, 24 Apr 2021 10:20:16 +0200
-Subject: [PATCH 2/2] remove impure dirs
+Subject: [PATCH 2/3] Remove impure dirs
 
+Purity: don't look for library dependencies (of the form `-lfoo') in
+/lib and /usr/lib. It's a stupid feature anyway. Likewise, when
+searching for included Makefiles, don't look in /usr/include and
+friends.
 ---
  src/read.c   | 3 ---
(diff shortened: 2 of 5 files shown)


tie:
Yikes, gnumake doesn’t have a maintainer now (#324539) 🫠

@AndersonTorres, would you mind reviewing this PR?

AndersonTorres:
Let's start with this!

https://github.com/NixOS/nixpkgs/issues/327779

pkgs/development/tools/build-managers/gnumake/default.nix
```diff
@@ -32,6 +32,17 @@ stdenv.mkDerivation rec {
     ./0002-remove-impure-dirs.patch
   ];
 
+  # Remove unnecessary reference to C++ compiler. Note that MAKE_CXX is
+  # always set to $CXX when running confugure script. If it is not set,
+  # a reasonable fallback is used.
+  #
+  # See also https://savannah.gnu.org/bugs/?63668 and make commit
+  # ffa28f3914ff402b3915f75e4fed86ac6fb1449d.
```
AndersonTorres, line 40:
```suggestion
  #
  # References:
  # - https://savannah.gnu.org/bugs/?63668
  # - https://git.savannah.gnu.org/cgit/make.git/commit/?id=ffa28f3914ff402b3915f75e4fed86ac6fb1449d
  #
```

AndersonTorres (changes requested):
I have some extra doubts about this specific project.

From a puristic perspective, it should be bootstrapped.
However, how can it be bootstrapped, given that a truckload of things that compose stdenv depend on it?

But let's postpone this. For now we need to test.

AndersonTorres:
P.S.: technically this PR should wait for next staging merge so that the meta info can be updated. But let's hope it does not conflict.

pkgs/development/tools/build-managers/gnumake/default.nix
tie: Done.

Sigmanificient:
> Let's start with this!
> 
> #327779

Do we have a place that list all tracking issues? I always find myself to search them which can something be tedious as i don't remember their name well

pkgs/development/tools/build-managers/gnumake/default.nix
```diff
@@ -32,6 +32,18 @@ stdenv.mkDerivation rec {
     ./0002-remove-impure-dirs.patch
   ];
 
+  # Remove unnecessary reference to C++ compiler. Note that MAKE_CXX is
+  # always set to $CXX when running confugure script. If it is not set,
+  # a reasonable fallback is used.
```
AndersonTorres, line 37:
```suggestion
  # always set to $CXX when running Autotools-based configure script,
  # falling back to a reasonable fallback when CXX is not set.
```

P.S.: the "it" refers to what? CXX, as I have  wrote above?

AndersonTorres:
> Do we have a place that list all tracking issues? I always find myself to search them which can something be tedious as i don't remember their name well

Search for the label "scope tracking":  
https://github.com/NixOS/nixpkgs/labels/5.%20scope%3A%20tracking

pkgs/development/tools/build-managers/gnumake/default.nix
tie: If MAKE_CXX is not set (a preprocessor macro that is set to a build-time CXX value), the embedded default value for run-time CXX uses a reasonable fallback like `g++` (or `gpp` on MSDOS). I’ve updated the comment to elaborate on this behavior on more details.

tie: I’ve updated the PR to use autoreconfHook with a patch to configure.ac.

tie:
I’ve rebased the PR and updated the change to use autoreconfHook with a patch to configure.ac and src/default.c. The only downside is that now we always need pkg-config to regenerate configure script (hence the updated nativeBuildInputs).

rhelmot:
Isn't updateAutotoolsGnuConfigScriptsHook a subset of autoreconfHook, so you only need the latter?

pkgs/development/tools/build-managers/gnumake/default.nix
```diff
@@ -24,6 +25,8 @@ stdenv.mkDerivation rec {
 
   # to update apply these patches with `git am *.patch` to https://git.savannah.gnu.org/git/make.git
   patches = [
+    # See patch message.
+    ./make-cxx.patch
```
AndersonTorres, line 29:
```suggestion
    ./0000-make-cxx.patch
```

Non blocking

AndersonTorres (approved):

philiptaron (approved):
I built and tested the stdenv, Nix, and a few other things (gcc, clang, zig.)

LGTM.

philiptaron:
I intend to merge this tomorrow; I'm holding it open for more commentary.

tie:
> Isn't updateAutotoolsGnuConfigScriptsHook a subset of autoreconfHook, so you only need the latter?

Uh, I’m not that familiar with Autotools and related build infrastructure in Nixpkgs. I guess this is true if autoreconfHook regenerates the files that updateAutotoolsGnuConfigScriptsHook updates 😅

I’ll remove it today when I get home if I don’t fall asleep, otherwise in the morning.

pkgs/development/tools/build-managers/gnumake/default.nix
tie: Updated the patch set as suggested in the comment above `patches`, cleaned up that a bit to use separate patches directory with improved docs.

tie:
As suggested in https://github.com/NixOS/nixpkgs/pull/314716#discussion_r1720778248, recreated MAKE_CXX patch with `git format-patch` and updated the patch set. Also removed `updateAutotoolsGnuConfigScriptsHook`, https://github.com/NixOS/nixpkgs/pull/314716#issuecomment-2294656582.

pkgs/development/tools/build-managers/gnumake/default.nix
```diff
@@ -22,17 +22,20 @@ stdenv.mkDerivation rec {
...
+  #  $ version=4.4.1
+  #  $ git clone https://git.savannah.gnu.org/git/make.git
+  #  $ cd make && git checkout -b nixpkgs $version
+  #  $ git am --directory=../patches
+  #  $ # make changes, resolve conflicts, etc.
+  #  $ git format-patch --output-directory ../patches --diff-algorithm=histogram $version
+  #
+  # TODO: stdenv’s setup.sh should be aware of patch directories. It’s very
+  # convenient to keep them in a separate directory but we can defer listing the
+  # directory until derivation realization to avoid unnecessary Nix evaluations.
```
tie, line 35: Created #335579

pkgs/development/tools/build-managers/gnumake/default.nix
```diff
@@ -22,17 +22,20 @@ stdenv.mkDerivation rec {
...
+  patches = lib.filesystem.listFilesRecursive ./patches;
```
philiptaron, line 36: As I wrote on #335579, I sort of love the old approach where the patches are just listed out. I'm willing to hear the case, but my "do the most simple thing" [grugbrain](https://grugbrain.dev/) triggers when there's magical patch application.

I'm willing to experiment, though.

pkgs/development/tools/build-managers/gnumake/patches/0002-Remove-impure-dirs.patch
```diff
@@ -1,20 +1,24 @@
...
+From 2db52008be2e2d504889f4f19318c2ba5a2a4797 Mon Sep 17 00:00:00 2001
 From: =?UTF-8?q?J=C3=B6rg=20Thalheim?= <joerg@thalheim.io>
 Date: Sat, 24 Apr 2021 10:20:16 +0200
-Subject: [PATCH 2/2] remove impure dirs
+Subject: [PATCH 2/3] Remove impure dirs
 
+Purity: don't look for library dependencies (of the form `-lfoo') in
+/lib and /usr/lib. It's a stupid feature anyway. Likewise, when
+searching for included Makefiles, don't look in /usr/include and
+friends.
```
philiptaron, line 9: This was taken from the comment in the Nix file (and I appreciate it here!)

philiptaron (approved):
I'm planning on merging once a test of the stdenv completes, even though I'm not sold on the patches approach.

PR #314775, opened 2024-05-26
jellyfin, jellyfin-web: 10.9.1 -> 10.9.3, 10.9.2 -> 10.9.3

Nanotwerp:
## Description of changes
Updates `jellyfin` from `10.9.1` to `10.9.3` and `jellyfin-web` from `10.9.2` to `10.9.3`. This also contains a change to `jellyfin-web` that adds an `assert` that only builds `jellyfin-web` if it is on the same version as `jellyfin`. Note that using `nixfmt-rfc-style`  changed the `default.nix` of `jellyfin`, due to it not being formatted by it before. `jellyfin-web`'s `web.nix` was unaffected by the formatting.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/servers/jellyfin/default.nix
+++ b/pkgs/servers/jellyfin/default.nix
@@ -1,30 +1,29 @@
-{ lib
-, fetchFromGitHub
-, nixosTests
-, stdenv
-, dotnetCorePackages
-, buildDotnetModule
-, ffmpeg
-, fontconfig
-, freetype
-, jellyfin-web
-, sqlite
+{
+  lib,
+  fetchFromGitHub,
+  nixosTests,
+  stdenv,
+  dotnetCorePackages,
+  buildDotnetModule,
+  ffmpeg,
+  fontconfig,
+  freetype,
+  jellyfin-web,
+  sqlite,
 }:
 
 buildDotnetModule rec {
   pname = "jellyfin";
-  version = "10.9.1"; # ensure that jellyfin-web has matching version
+  version = "10.9.3"; # ensure that jellyfin-web has matching version
 
   src = fetchFromGitHub {
     owner = "jellyfin";
     repo = "jellyfin";
     rev = "v${version}";
-    sha256 = "sha256-ZvXz4gnpYE9bMvOHbmLhqJLUomPmk1K9ysw+Wlsyhr4=";
+    sha256 = "sha256-gJMz2LfxC0JXqGYNKNz1zRbWZOH1UxbcoGtmdymZ/Oo=";
   };
 
-  propagatedBuildInputs = [
-    sqlite
-  ];
+  propagatedBuildInputs = [ sqlite ];
 
   projectFile = "Jellyfin.Server/Jellyfin.Server.csproj";
   executables = [ "jellyfin" ];
@@ -56,7 +55,12 @@ buildDotnetModule rec {
     homepage = "https://jellyfin.org/";
     # https://github.com/jellyfin/jellyfin/issues/610#issuecomment-537625510
     license = licenses.gpl2Plus;
-    maintainers = with maintainers; [ nyanloutre minijackson purcell jojosch ];
+    maintainers = with maintainers; [
+      nyanloutre
+      minijackson
+      purcell
+      jojosch
+    ];
     mainProgram = "jellyfin";
     platforms = dotnet-runtime.meta.platforms;
   };
--- a/pkgs/servers/jellyfin/web.nix
+++ b/pkgs/servers/jellyfin/web.nix
@@ -29,12 +29,14 @@ buildNpmPackage' rec {
   pname = "jellyfin-web";
   version = "10.9.3";
 
-  src = fetchFromGitHub {
-    owner = "jellyfin";
-    repo = "jellyfin-web";
-    rev = "v${version}";
-    hash = "sha256-duq2tilUDEzj7o3Nq3Ku5qVJm4XDqVmqkQQIK/dlTpE=";
-  };
+  src =
+    assert version == jellyfin.version;
+    fetchFromGitHub {
+      owner = "jellyfin";
+      repo = "jellyfin-web";
+      rev = "v${version}";
+      hash = "sha256-duq2tilUDEzj7o3Nq3Ku5qVJm4XDqVmqkQQIK/dlTpE=";
+    };
 
   npmDepsHash = "sha256-nKA/mR1ug1yq4+jJGhWGtAL9Zsx3KjDPqt5rkCE4LFU=";
 
--- a/pkgs/servers/jellyfin/web.nix
+++ b/pkgs/servers/jellyfin/web.nix
@@ -4,6 +4,7 @@
   overrideSDK,
   fetchFromGitHub,
   buildNpmPackage,
+  jellyfin,
   nix-update-script,
   pkg-config,
   xcbuild,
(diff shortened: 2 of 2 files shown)


pkgs/servers/jellyfin/default.nix
```diff
@@ -13,13 +13,13 @@
 
 buildDotnetModule rec {
   pname = "jellyfin";
-  version = "10.9.1"; # ensure that jellyfin-web has matching version
+  version = "10.9.2"; # ensure that jellyfin-web has matching version
```
doronbehar, line 16: BTW you can add a `lib.warn` that will disturb ofborg's evaluation by comparing the versions.

doronbehar: Or you can use an assertion.

Nanotwerp: Thank you. Resolved by [this change](https://github.com/NixOS/nixpkgs/commit/ffb3ea67083513d5d7ffe1dfd5f6fdb7d0922178)!

Tasqa:
Will this be backported to 24.05?

AkechiShiro:
Jellyfin 10.9.3 is out btw, maybe this PR should update jellyfin and jellyfin-web to 10.9.3

Nanotwerp:
> Jellyfin 10.9.3 is out btw, maybe this PR should update jellyfin and jellyfin-web to 10.9.3 @nyanloutre @minijackson @purcell @jojosch

Done!

pkgs/servers/jellyfin/web.nix
```diff
@@ -1,41 +1,44 @@
...
-  stdenv,
-  overrideSDK,
-  fetchFromGitHub,
-  buildNpmPackage,
-  nix-update-script,
-  pkg-config,
-  xcbuild,
-  pango,
-  giflib,
-  darwin,
```
doronbehar: Did you use `nixfmt`? There is no consensus in Nixpkgs about it's formatting rules, and this change enlarges the diff for no good reason.

pkgs/servers/jellyfin/web.nix
```diff
@@ -1,41 +1,44 @@
...
       stdenv;
   buildNpmPackage' = buildNpmPackage.override { stdenv = stdenv'; };
 in
 buildNpmPackage' rec {
   pname = "jellyfin-web";
-  version = "10.9.2";
+  version = "10.9.3";
 
-  src = fetchFromGitHub {
+  # nixpkgs-update: no auto update
```
doronbehar: Why is that? I'm not sure that if `nixpkgs-update` will try to update it, it will fail. I think though, that it won't update a PR, because it won't understand that it should also update `jellyfin`, so that evaluation will be fixed.

pkgs/servers/jellyfin/web.nix
```diff
@@ -1,41 +1,44 @@
...
     if stdenv.isDarwin then
-      overrideSDK stdenv {
-        darwinMinVersion = "10.15";
-        darwinSdkVersion = "11.0";
-      }
+      overrideSDK stdenv
+        {
+          darwinMinVersion = "10.15";
+          darwinSdkVersion = "11.0";
+        }
```
doronbehar: Enlarges the diff for no good reason.

doronbehar (commented):
I had to play with this PR locally, so I pushed a very similar change overall, but with 3 commits.

pkgs/servers/jellyfin/web.nix
Nanotwerp: It seems to be protocol to not let `nixpkgs-update` auto-update packages that should have the same version as its accompanying package. In this case, `no auto update` should help prevent it from getting out of sync with `jellyfin`. I think this may be the best option, unless there's a way for `nixpkgs-update` to couple two packages that should be updated together in one commit.

pkgs/servers/jellyfin/web.nix
Nanotwerp: > Did you use `nixfmt`? There is no consensus in Nixpkgs about it's formatting rules, and this change enlarges the diff for no good reason.

The outdated code that you're commenting on was formatted by whatever `nixd` uses. Now, I  used `nixfmt-rfc-style` on both files and rebased. The usage of `nixfmt-rfc-style` appears to be encouraged, so I'll keep it like this.

pkgs/servers/jellyfin/web.nix
Nanotwerp: Fixed with `nixfmt-rfc-style`.

doronbehar: > Fixed with `nixfmt-rfc-style`.

I'm not sure what tool are you talking about, (is it https://github.com/NixOS/nixfmt ? ) In either case, it's OK to use it if it is official, but I'd recommend apply it's changes in a dedicated commit. Also, I would recommend using 3 different commits for the updates, like I pushed to this PR. Together with the formatting commit, it will become:

```
jellyfin: 10.9.1 -> 10.9.3
jellyfin-web: use version from jellyfin, to avoid version mismatch
jellyfin-web: 10.9.2 -> 10.9.3
jellyfin: format with NixOS/nixfmt
jellyfin-web: format with NixOS/nixfmt
```

Nanotwerp: I rebased to use three commits; the first two for the updates of `jellyfin` and `jellyfin-web`, respectively, and the third for adding the `assert` to `jellyfin-web`. I don't think it's idiomatic of a `nixpkgs` PR to make formatting a separate commit unless the *only* change to the `.nix` file is that it's being formatted.

doronbehar: > I don't think it's idiomatic of a `nixpkgs` PR to make formatting a separate commit unless the _only_ change to the `.nix` file is that it's being formatted.

OK, not a big deal. It'd still be nice if you'd share (ideally in the commit message text), exactly what tool did you use to format the expression, I'm still not sure which one it is.

Nanotwerp: > OK, not a big deal. It'd still be nice if you'd share (ideally in the commit message text), exactly what tool did you use to format the expression, I'm still not sure which one it is.

In the outdated version of my branch, I was using `nixpkgs-fmt`. I'm now using `nixfmt-rfc-style`, which is the package name for the current, [`RFC 166`](https://github.com/NixOS/nixfmt/issues/153)-compliant, version of `nixfmt`. I edited the PR message to explain that I formatted `jellyfin`'s `default.nix`, and that `jellyfin-web`'s `web.nix` was already formatted in this manner, so it didn't need to be formatted.

Nanotwerp: Just edited the [commit message](https://github.com/NixOS/nixpkgs/pull/314775/commits/a7441feb23f2ba370fb3ae130f51113c5c72b8f2) for `jellyfin`'s update to indicate that it was formatted.

doronbehar: OK I wasn't aware of this package, thanks for cooperating :) Will merge when CI is green.

minijackson:
huh? Maybe the backport action didn't like me adding 2 tags at once.

PR #314791, opened 2024-05-26
nixos/pam: add `security.pam.package` option

uninsane:
## Description of changes

this can be used to swap out which pam package is actually used by the
system for things like `pam_unix`, `pam_limits`, etc, without forcing
a mass rebuild the way an overlay would.

see commit message for more info on why places which previously specified just the .so name for `modulePath` now specify the full path.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/nixos/modules/security/pam.nix
+++ b/nixos/modules/security/pam.nix
@@ -99,6 +99,7 @@ let
     }));
   };
 
+  package = config.security.pam.package;
   parentConfig = config;
 
   pamOpts = { config, name, ... }: let cfg = config; in let config = parentConfig; in {
@@ -731,7 +732,7 @@ let
               { name = "gnupg"; enable = cfg.gnupg.enable; control = "optional"; modulePath = "${pkgs.pam_gnupg}/lib/security/pam_gnupg.so"; settings = {
                 store-only = cfg.gnupg.storeOnly;
               }; }
-              { name = "faildelay"; enable = cfg.failDelay.enable; control = "optional"; modulePath = "${pkgs.pam}/lib/security/pam_faildelay.so"; settings = {
+              { name = "faildelay"; enable = cfg.failDelay.enable; control = "optional"; modulePath = "${package}/lib/security/pam_faildelay.so"; settings = {
                 inherit (cfg.failDelay) delay;
               }; }
               { name = "google_authenticator"; enable = cfg.googleAuthenticator.enable; control = "required"; modulePath = "${pkgs.google-authenticator}/lib/security/pam_google_authenticator.so"; settings = {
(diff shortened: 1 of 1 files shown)


SuperSandro2000 (changes requested):

nixos/modules/security/pam.nix
```diff
@@ -1546,9 +1549,9 @@ in
         (name: "r ${config.environment.etc."pam.d/${name}".source},\n")
         (attrNames config.security.pam.services) +
       ''
-      mr ${getLib pkgs.pam}/lib/security/pam_filter/*,
-      mr ${getLib pkgs.pam}/lib/security/pam_*.so,
-      r ${getLib pkgs.pam}/lib/security/,
+      mr ${getLib package}/lib/security/pam_filter/*,
+      mr ${getLib package}/lib/security/pam_*.so,
+      r ${getLib package}/lib/security/,
```
Majiir, line 1554: I think your change makes these rules unnecessary. The section below adds apparmor rules for any configured modules with absolute paths, which is now all of them. So, I think we could remove these modified rules, and then convert the absolute path filter below into an assertion requiring all `modulePath` to be absolute.

Majiir (approved):
Thanks for doing this!

nixos/modules/security/pam.nix
```diff
@@ -92,6 +92,7 @@ let
     }));
   };
 
+  pamPackage = config.security.pam.package;
```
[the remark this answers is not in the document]
Majiir: Hm, I have the opposite feedback: `pamPackage` is better for readability within the module than `package`. This is just a `let` binding, not an option. Either is okay by me for merging.

nixos/modules/security/pam.nix
uninsane: i've deleted this code and added a **warning** to the below codepath whenever we encounter non-absolute paths.

i can do an assertion right here if you're confident, but since i'm not versed in apparmor setups my inclination is to instead show a warning, and  open a "(NOT FOR MERGE)" PR immediately after this one to upgrade that to an assertion. this allows users at least a couple weeks to fix any broken configs/provide feedback before we fail eval for anyone.

Majiir (approved):

nixos/modules/security/pam.nix
```diff
@@ -958,6 +959,8 @@ in
 
   options = {
 
+    security.pam.package = mkPackageOption pkgs.pam "pam" { };
```
SuperSandro2000, line 962:
```suggestion
    security.pam.package = mkPackageOption pkgs "pam" { };
```

SuperSandro2000 (dismissed):

JohnRTitor (approved):
I think this is good to go, will merge after eval passes

PR #314173, opened 2024-05-24
nixos/healthchecks: add settingsFile option

yomaq:
Single unified env file read by systemd

## Description of changes

Current settings option requires multiple files if you want to set multiple environment secrets, as well as being read by the service program, meaning that things like agenix won't work with the default service user.

Added a settingsFile option that allows a single secret file, and doesn't break any current behavior for existing settings options, and plays well with agenix. Env variables can be set in both `settings` and `settingsFile`.

Tested both the webservice, and the `healtchecks-manage` script created by the module.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
- For non-Linux: Is sandboxing enabled in `nix.conf`? (See [Nix manual](https://nixos.org/manual/nix/stable/command-ref/conf-file.html))
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:
  - [NixOS test(s)](https://nixos.org/manual/nixos/unstable/index.html#sec-nixos-tests) (look inside [nixos/tests](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests))
  - and/or [package tests](https://github.com/NixOS/nixpkgs/blob/master/pkgs/README.md#package-tests)
  - or, for functions and "core" functionality, tests in [lib/tests](https://github.com/NixOS/nixpkgs/blob/master/lib/tests) or [pkgs/test](https://github.com/NixOS/nixpkgs/blob/master/pkgs/test)
  - made sure NixOS tests are [linked](https://nixos.org/manual/nixpkgs/unstable/#ssec-nixos-tests-linking) to the relevant packages
- [ ] Tested compilation of all packages that depend on this change using `nix-shell -p nixpkgs-review --run "nixpkgs-review rev HEAD"`. Note: all changes have to be committed, also see [nixpkgs-review usage](https://github.com/Mic92/nixpkgs-review#usage)
- [ ] Tested basic functionality of all binary files (usually in `./result/bin/`)
[...]

PR #314798, opened 2024-05-26
nixos/healthchecks: add settingsFile option

yomaq:
Single unified env file read by systemd

## Description of changes

Current settings option requires multiple files if you want to set multiple environment secrets, as well as being read by the service program, meaning that things like agenix won't work with the default service user.

Added a settingsFile option that allows a single secret file, and doesn't break any current behavior for existing settings options, and plays well with agenix. Env variables can be set in both settings and settingsFile.

Tested both the webservice, and the healtchecks-manage script created by the module.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/nixos/modules/services/web-apps/healthchecks.nix
+++ b/nixos/modules/services/web-apps/healthchecks.nix
@@ -11,7 +11,7 @@ let
   environment = {
     PYTHONPATH = pkg.pythonPath;
     STATIC_ROOT = cfg.dataDir + "/static";
-  } // cfg.settings;
+  } // lib.filterAttrs (_: v: !builtins.isNull v) cfg.settings;
 
   environmentFile = pkgs.writeText "healthchecks-environment" (lib.generators.toKeyValue { } environment);
 
@@ -21,6 +21,7 @@ let
       sudo='exec /run/wrappers/bin/sudo -u ${cfg.user} --preserve-env --preserve-env=PYTHONPATH'
     fi
     export $(cat ${environmentFile} | xargs)
+    ${lib.optionalString (cfg.settingsFile != null) "export $(cat ${cfg.settingsFile} | xargs)"}
     $sudo ${pkg}/opt/healthchecks/manage.py "$@"
   '';
 in
@@ -89,6 +90,12 @@ in
       '';
     };
 
+    settingsFile = lib.mkOption {
+      type = lib.types.nullOr lib.types.path;
+      default = null;
+      description = opt.settings.description;
+    };
+
     settings = lib.mkOption {
       description = ''
         Environment variables which are read by healthchecks `(local)_settings.py`.
@@ -109,6 +116,8 @@ in
           have support for a `_FILE` variant, run:
           - `nix-instantiate --eval --expr '(import <nixpkgs> {}).healthchecks.secrets'`
           - or `nix eval 'nixpkgs#healthchecks.secrets'` if the flake support has been enabled.
+
+        If the same variable is set in both `settings` and `settingsFile` the value from `settingsFile` has priority.
       '';
       type = types.submodule (settings: {
         freeformType = types.attrsOf types.str;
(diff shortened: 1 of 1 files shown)


nixos/modules/services/web-apps/healthchecks.nix
```diff
@@ -109,9 +116,11 @@ in
           have support for a `_FILE` variant, run:
           - `nix-instantiate --eval --expr '(import <nixpkgs> {}).healthchecks.secrets'`
           - or `nix eval 'nixpkgs#healthchecks.secrets'` if the flake support has been enabled.
+
+        If the same variable is set in both `settings` and `settingsFile` the value from `settingsFile` has priority.
       '';
       type = types.submodule (settings: {
-        freeformType = types.attrsOf types.str;
+        freeformType = types.attrsOf (types.nullOr types.str);
```
SuperSandro2000, line 123: This should also already allow {} which is a bit better than null.

SuperSandro2000 (changes requested):
Why are multiple files required? The environmentFile by systemd can hold multiple envs already.

nixos/modules/services/web-apps/healthchecks.nix
amozeo: This is a change that I proposed in https://github.com/NixOS/nixpkgs/pull/314173#discussion_r1613159749
see the comment before it for reason why

amozeo:
It should be noted that pr author originally created https://github.com/NixOS/nixpkgs/pull/314173, but wanted to rename the branch as stated in his last comment there

yomaq:
Previously the module would have you set an env variable that would point to a file for each secret you wanted to set, and the program would read the secret file rather than systemd.

Only one file is required. The `settingsFile` option I can handle all secret requirements just fine, as it is using the systemd environmentFile directly.
But as to not to make any breaking changes I left the previous env variable options as is.

nixos/modules/services/web-apps/healthchecks.nix
yomaq: Removed the change as suggested.
The script issue from [#314173 (comment)](https://github.com/NixOS/nixpkgs/pull/314173#discussion_r1613159749) is fixed by line 133 where `settings.SECRET_KEY_FILE` can still be set as null.

nixos/modules/services/web-apps/healthchecks.nix
```diff
@@ -186,7 +196,10 @@ in
           WorkingDirectory = cfg.dataDir;
           User = cfg.user;
           Group = cfg.group;
-          EnvironmentFile = [ environmentFile ];
+          EnvironmentFile = [
+            environmentFile
+            cfg.settingsFile
```
eclairevoyant, line 201: missing null check?

yomaq: It was, thank you

nixos/modules/services/web-apps/healthchecks.nix
```diff
@@ -186,7 +196,10 @@ in
           WorkingDirectory = cfg.dataDir;
           User = cfg.user;
           Group = cfg.group;
-          EnvironmentFile = [ environmentFile ];
+          EnvironmentFile = [
+            environmentFile
+            (lib.optional (cfg.settingsFile != null) cfg.settingsFile)
+          ];
```
amozeo, line 202: `lib.optional` returns a list
```suggestion
          EnvironmentFile = [
            environmentFile
          ] ++ lib.optional (cfg.settingsFile != null) cfg.settingsFile;
```

yomaq: Module was working as it was, but it does return I list. I've fixed this now.

SuperSandro2000 (approved):

PR #314720, opened 2024-05-26
gawk: move gawkbug to gawkInteractive

tie:
## Description of changes

This change removes gawkbug program from non-interactive build, similar to the bash{,Interactive} package.

**Parent PR:**
- #314920

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [x] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
- For non-Linux: Is sandboxing enabled in `nix.conf`? (See [Nix manual](https://nixos.org/manual/nix/stable/command-ref/conf-file.html))
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:
  - [NixOS test(s)](https://nixos.org/manual/nixos/unstable/index.html#sec-nixos-tests) (look inside [nixos/tests](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests))
  - and/or [package tests](https://github.com/NixOS/nixpkgs/blob/master/pkgs/README.md#package-tests)
  - or, for functions and "core" functionality, tests in [lib/tests](https://github.com/NixOS/nixpkgs/blob/master/lib/tests) or [pkgs/test](https://github.com/NixOS/nixpkgs/blob/master/pkgs/test)
  - made sure NixOS tests are [linked](https://nixos.org/manual/nixpkgs/unstable/#ssec-nixos-tests-linking) to the relevant packages
- [ ] Tested compilation of all packages that depend on this change using `nix-shell -p nixpkgs-review --run "nixpkgs-review rev HEAD"`. Note: all changes have to be committed, also see [nixpkgs-review usage](https://github.com/Mic92/nixpkgs-review#usage)
- [ ] Tested basic functionality of all binary files (usually in `./result/bin/`)
- [24.11 Release Notes](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2411.section.md) (or backporting [23.11](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2311.section.md) and [24.05](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2405.section.md) Release notes)
  - [ ] (Package updates) Added a release notes entry if the change is major or breaking
  - [ ] (Module updates) Added a release notes entry if the change is significant
  - [ ] (Module addition) Added a release notes entry if adding a new NixOS module
[...]

PR #294504, opened 2024-03-09
postgresql: split -lib and -dev outputs cleanly

wolfgangwalther:
## Description of changes

TLDR: This splits outputs for the postgresql package cleanly, using a minimum of patching and/or nuking of references. Instead, dead references to other store paths are removed at compile and link time. The minimal -lib output is then added as a top-level `libpq` package.

This builds on my previous work in #292993 and #293996 and contains commits from those two branches. Please ignore the first ~20 commits here and provide feedback for those in the two other PRs. This PR starts with `treewide: prepare pg_config moving to dev`. This commit (by @jtojnar) and the 4 commits after (by @SuperSandro2000) were cherry-picked from #179962. Compared to #179962, the approach taken in this PR is simpler and requires far less patching of postgresql source code. In fact, the overall size of patches is reduced compared to before and fewer references needed to be removed afterwards.

While the most work here involved splitting the -dev output, a clean -lib output comes with that naturally. Thus, this PR replaces #179962 and #273175. The cleaned up lib output was discussed in https://github.com/NixOS/nixpkgs/issues/61580#issuecomment-493154280 (@thoughtpolice). With this output available, it's easy to create a top-level attribute `libpq` pointing to the latest version's lib output automatically as suggested by @danbst in https://github.com/NixOS/nixpkgs/issues/61580#issuecomment-498161755, which should resolve #61580 (@nh2). A true separate libpq package as proposed in #234470 (@szlend) would then not be necessary. This PR will be followed-up by another PR to build postgresql v16+ based on meson instead of autotools as mentioned in https://github.com/NixOS/nixpkgs/pull/292993#issue-2165314957. With a bit of patching (of which I have a working PoC already), this will allow us to build postgresql (at least libpq, but possibly even more) in `pkgsStatic` - which is probably the biggest thing that didn't work so far and why a separate libpq package was asked for in the first place.

After the split, the outputs contain the following:
- default `out`:
  - All binaries, except `pg_config` and `ecpg` which are in the dev output. The default output thus includes both server-side and client-side binaries.
  - All contrib/core extension related files: server-libraries, extension control files etc.
- `lib`:
  - libpq
  - libecpg, including libecpg_compat and libpgtypes
  - theoretically the `/share/locale` folders, if there were any built, which is not the case right now
- `doc` and `man` as before
- `dev`:
  - all header files in `/include` (both server and client)
  - the `pgxs` infrastructure to build extensions
[...]

PR #290081, opened 2024-02-20
stdenv: log build hooks as they run

Qyriad:
## Description of changes

Hooks are vital to Nixpkgs, and are often the "magic" to make things Just Work, but they're also extremely opaque. Adding something like `undmg` to a derivation's `nativeBuildInputs` adds a hook to the unpackPhase to automatically extract DMG sources — this is fantastic, but a beginner would have *no* way of knowing that adding something as a build input could have such a major effect (`undmg` gets only a [single passing mention](https://nixos.org/manual/nixpkgs/unstable/#unzip) in the Nixpkgs manual), and even to Nixpkgs veterans it's *far* from obvious what the final array of hooks for a build might be, or what order they even run in (a source of confusion if one causes an error!).

This PR changes stdenv's setup.sh to log hooks to stdout as they run, as it does currently with phases. Setup hook files are logged just before they're sourced, and individual build hooks like `patchShebangsAuto` are also logged just before they're run.

With this PR, now anyone building with `--print-build-logs` will have transparency into one of the most important aspects of Nixpkgs, to aid heavily in both debugging and learning.

<details>
<summary>For example, here what it now looks like when building Curlie:</summary>
<code><pre>
sourcing setup hook '/nix/store/33jvc2jsv20lsx8dp0pwf0rr7xhwwpq9-patchelf-0.15.0/nix-support/setuphook'
sourcing setup hook '/nix/store/qmavq4i2rh012ggwb5fxlvp3pzhxmbkl-update-autotools-gnu-config-scripts-hook/nix-support/setuphook'
sourcing setup hook '/nix/store/h9lc1dpi14z7is86ffhl3ld569138595-audit-tmpdir.sh'
sourcing setup hook '/nix/store/m54bmrhj6fqz8nds5zcj97w9s9bckc9v-compress-man-pages.sh'
sourcing setup hook '/nix/store/wgrbkkaldkrlrni33ccvm3b6vbxzb656-make-symlinks-relative.sh'
sourcing setup hook '/nix/store/5yzw0vhkyszf2d179m0qfkgxmp5wjjx4-move-docs.sh'
sourcing setup hook '/nix/store/fyaryjvghbkpfnsyw97hb3lyb37s1pd6-move-lib64.sh'
sourcing setup hook '/nix/store/kd4xwxjpjxi71jkm6ka0np72if9rm3y0-move-sbin.sh'
[...]

PR #209870, opened 2023-01-09
stdenv: Nix-driven bootstrap of gcc

ghost:
#### :heavy_check_mark: Good to go.

- [X] Deferred #221525
- [x] [bootstrap-files `allowedRequisites` issue](https://github.com/NixOS/nixpkgs/pull/209870#issuecomment-1464546176).
- [X] Make a ["last call"](https://github.com/NixOS/nixpkgs/pull/209870#issuecomment-1473175912) for anybody to raise objections
- [x] [all jobs pass on Hydra](https://hydra.nixos.org/eval/1792339) 

#### Summary

By default, when you type `make`, GCC will compile itself three times.  This PR inhibits that behavior by configuring GCC with `--disable-bootstrap`, and reimplements the triple-rebuild using Nix rather than `make`/`sh`.

#### Immediate Benefits

- Allow `gcc11` and `gcc12` on `aarch64` (without needing new  `bootstrapFiles`)
- Faster stdenv rebuilds: the third compilation of gcc  (i.e. stageCompare) is no longer a `drvInput` of the final stdenv.  This allows Nix to build stageCompare in parallel with the rest of  nixpkgs instead of in series.
- No more copying `libgcc_s` out of the bootstrap-files or other  derivations
- No more Frankenstein compiler: the final gcc and the libraries it links against (mpfr, mpc, isl, glibc) are all built by the same compiler (xgcc)  instead of a mixture of the bootstrapFiles' compiler and xgcc.
- No more [static lib{mpfr,mpc,gmp,isl}.a hack]
- Many other small `stdenv` hacks eliminated
- `gcc` and `clang` share the same codepath for more of `cc-wrapper`.

#### Future Benefits

- This should allow using a [foreign] `bootstrap-files` so long as  `hostPlatform.canExecute bootstrapFiles`.
- This should allow each of the libraries that ship with `gcc`  (lib{backtrace, atomic, cc1, decnumber, ffi, gomp, iberty,  offloadatomic, quadmath, sanitizer, ssp, stdc++-v3, vtv}) to be  built in separate (one-liner) derivations which `inherit src;`  from `gcc`, much like https://github.com/NixOS/nixpkgs/pull/132343

#### Incorporates

- https://github.com/NixOS/nixpkgs/pull/210004
- https://github.com/NixOS/nixpkgs/pull/36948 (unreverts it)
- https://github.com/NixOS/nixpkgs/pull/210325
- https://github.com/NixOS/nixpkgs/pull/210118
[...]

PR #314920, opened 2024-05-26
wrapCC, wrapBintools: use absolute paths

tie:
## Description of changes

This change sets {CC,CXX,LD,etc…}{,_FOR_BUILD,_FOR_TARGET} variables to absolute paths in cc-wrapper and bintools-wrapper setup hooks.

It’s common to set
```nix
{
  depsBuildBuild = [ buildPackages.stdenv.cc ];
}
```
when cross-compiling (see doc/stdenv/cross-compilation.chapter.md), however, cc/bintools wrapper setup hook currently sets these variables to program name instead of absolute path. That obviously doesn’t work e.g. when build and host platform have the same C compiler name because only the first program from PATH is used (for example, this is part of the reason why static glibc doesn’t work in Nixpkgs and we have pkgsStatic use musl that has different target prefix).

**Prerequisites:**
- #314720
- #314716
- #316011
- #316023
- ~~#322025~~
  #294504

Note: I’ll be adding new PRs split from this one to the list above.

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [x] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---


tie:
Rebuild from `nixpkgs-review` is going well so far, most of the failures seems to be unrelated to this change (i.e. broken packages, outdated source hashes).
<details>
image
</details>

alyssais:
Looks like some unrelated commits have found their way in.

tie:
@alyssais, I don’t think so? The main change is in 3b6195a82ee60f528852b34e097c0582734deb1e with other commits mostly fixing occasional $CC references introduced by package build systems. I’ll try to merge these separately (hence the draft status for now), but they are still required for this PR, especially stdenv stuff like gnumake and gawk.

alyssais:
ah right, I missed the prerequisites mention in the PR body

tie:
Uh, I think we have a problem, looks like `pkgs/test/stdenv/gcc-stageCompare.nix` is broken (as in, it doesn’t check anything because compiler override has no effect) and this PR elevates the issue to a build failure.

At b1a4e835b68d14b92525468358887b8c3ea80ada on `aarch64-linux`:
```
nix develop --impure --expr 'with (import ./. { });
let ccWrapper = wrapCCWith { cc = stdenv.cc; };
in (gcc-unwrapped.override {
  stdenv = overrideCC stdenv ccWrapper;
}).overrideAttrs (_: { expectedCC = ccWrapper; })' \
-c bash -c 'echo "Expected: $expectedCC"
echo "Actual: $NIX_CC"
echo "PATH lookup: $CC -> $(type -P "$CC")"
echo "CC_FOR_BUILD: $CC_FOR_BUILD"'
```

Note that we use stripped down version of `pkgs/test/stdenv/gcc-stageCompare.nix` in the `--expr` argument above.

Output:
```
Expected: /nix/store/ahdbqcc10l1m9l543msjjiglyphabqz3-gcc-wrapper-wrapper-13.2.0
Actual: /nix/store/ahdbqcc10l1m9l543msjjiglyphabqz3-gcc-wrapper-wrapper-13.2.0
PATH lookup: gcc -> /nix/store/4jwr53s7jxhcqcxqyhc83jb0ksdjj1r0-bootstrap-stage3-gcc-wrapper-13.2.0/bin/gcc
CC_FOR_BUILD: gcc
```

So, stdenv overrides work fine, we get the correct `$NIX_CC`, *but* `$CC` from `PATH` ends up with bootstrap C compiler. No wonder this test fails with this PR—it didn’t check anything and just rebuilt gcc with the same `$CC` since its introduction in #209870.

What would be the right path to move forward with this PR? I’d really like to avoid diving even deeper down this rabbit hole and focus on a single issue at hand 😬

cc @Ericson2314 @trofi @vcunat @SuperSandro2000

trofi:
> That obviously doesn’t work e.g. when build and host platform have the same C compiler name because only the first program from PATH is used

Yeah, I'm not a big fan of absolute paths as they tend to leak everywhere as you have discovered and sometimes are not even supported by the build systems. But if there is no better solution then it might be a reasonable workaround.

Ideally I would say that `targetPrefix` should differ for different enough targets. Otherwise you have to use other means to convince autoconf that `./configure --build=foo --host=foo` is a cross-compilation. It breaks many assumptions in both upstream packages and in `nixpkgs` downstream. At least in `nixpkgs` `isStatic` looks very different from `!isStatic` as they don't share much of `/nix/store` (even if both targets are `musl`). But changing `targetPrefix` for those is a big (and somewhat controversial) change as well.

tie:
>Ideally I would say that targetPrefix should differ for different enough targets.

Nixpkgs platforms can be configured in so many different ways (microarchitectures, compiler flags, etc) that the only reasonable encoding is a hash of all parameters, which we already have in a form of absolute paths in Nix store 😅
In addition to that, most of the configure scripts out there works fine-ish with absolute paths but I can’t say that for some “non-standard” prefix that is not a target triplet in the basename.

Ericson2314:
I think in the past both this and putting the target prefix on native binaries have both been half-attempted.

tie:
CI checks are green after a few rebases and minor changes, that’s good to see. I’ll try to focus on merging the existing changes that I’ve already split (see __Prerequisites__ in the PR description) and restart `nixpkgs-review` rebuild in the background.

paparodeo:
i'm not sure if is relevant, but we found that cmake is adding the `bin` directories of `buildInputs` etc ahead of PATH, which causes issues that seemingly can be fixed by setting CC/CXX to the full path but working on a fix to just not add the bad stuff to the search path.

 https://github.com/NixOS/nixpkgs/pull/318226 (the diff is not updated yet with the new stuff -- still testing)

AndersonTorres:
I remember some time ago that was cited Exherbo employed something like always prefixing the target on the name of CC.
Would this be useful?

tie:
We do prefix target when host ≠ build, but the PATH lookup is still ambiguous in many cases. For example, I recently had to add a [workaround](https://github.com/NixOS/nixpkgs/blob/f299a8e9664cf246bccf5abec16ee4a6ab435c21/pkgs/development/web/nodejs/nodejs.nix#L138C3-L144) for this issue because a package required `libtool` from `darwin.cctools` and the latter shadowed `$AR = ar` from stdenv. Another example is https://github.com/NixOS/nixpkgs/pull/314920#issuecomment-2138797149.

Aside from occasional reference leaks, I think that using absolute paths is better than prefixing toolchain (i.e. __using Nix store paths instead of target prefixes__). That is because nativeBuildInputs and other flags from mkDerivation arguments take precedence over defaults and we also embed some platform-specific options in the wrapper itself.

Fixing unwanted references is easy in most cases and is a common task for Nixpkgs contributors, but I start questioning my sanity every time the build uses wrong toolchain because of PATH lookup.

FWIW, some packages already set these variables to absolute paths for non-trivial builds, e.g.
- gcc: https://github.com/NixOS/nixpkgs/blob/aa28fa2be9cc65c0c4fc3b7e3f6ba55e37d5f5c4/pkgs/development/libraries/gcc/libgcc/default.nix#L59-L75
- fast-downward: https://github.com/NixOS/nixpkgs/blob/aa28fa2be9cc65c0c4fc3b7e3f6ba55e37d5f5c4/pkgs/applications/science/logic/fast-downward/default.nix#L31-L33

AndersonTorres:
> We do prefix target when target ≠ host

Why not to generalize it? Regardless the need of absolute paths, I think we should begin to treat cross-compilation as first  class and native compilation as a particular case of it.
And, given that GCC is not smart enough to run like  `gcc --host=<> --target=<>`, the solution is to use the corresponding  prefix.
Further, I believe it would eliminate the necessity of always writing code like `CC=${stdenv.cc.targetPrefix}cc`. (In my latest RG count, it returned 370 files.)

(I hope this is not too off-topic. If so, feel free to ignore it.)

tie:
>Why not to generalize it? Regardless the need of absolute paths, I think we should begin to treat cross-compilation as first class and native compilation as a particular case of it.

Yes, I completely agree with this! Unfortunately, GCC and the existing C ecosystem makes it much harder than it should be, and potentially having multiple matching binaries on the search path doesn’t help that…

On that note, we’ve had several PRs recently that added useful information to the stdenv (e.g. #290081), so perhaps we can also warn about binaries under the same name present in multiple PATH entries 🤔

tie:
Just had another weird issue while working in #329721 where autotools configure script (generated after `autoreconfHook`) does not detect `$AR` for `pkgsStatic.xar` unless it is set to an absolute path. Not sure if that is the case for other packages (but I suspect it is)…

```console
$ echo $AR
aarch64-unknown-linux-musl-ar
$ AR=$AR ./configure $confugureFlags | fgrep 'checking for ar'
checking for ar... no
$ AR=$NIX_CC/bin/$AR ./configure $confugureFlags | fgrep 'checking for ar'
checking for ar... /nix/store/zzaivbdjmarqdcsfia9g5zfs8mgdhj9q-aarch64-unknown-linux-musl-gcc-wrapper-13.3.0/bin/aarch64-unknown-linux-musl-ar
```

emilazy:
Relevant to the prefixing discussion: https://github.com/NixOS/nixpkgs/pull/44583.

tie:
Note: the PR is not stale and I still think there aren’t any better solutions compared to using absolute paths. I’ll try to squeeze in a rebase in a week or so.

PR #314937, opened 2024-05-26
make(Binary)Wrapper: record created wrappers

abathur:
## Description of changes

Write a machine-readable list of wrappers (each line is a `$wrapper:$original` pair) to `${output-wrapper-is-being-written-to}/nix-support/wrappers`.
    
A wrapper -> original mapping makes it easier to see through wrappers, especially binary wrappers, to the original executable.

One case where this helps is when you need to perform analysis on the actual executables and associate that analysis with exec wrappers.

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/build-support/setup-hooks/make-binary-wrapper/make-binary-wrapper.sh
+++ b/pkgs/build-support/setup-hooks/make-binary-wrapper/make-binary-wrapper.sh
@@ -53,6 +53,33 @@ makeBinaryWrapper() {
         -Os \
         -x c \
         -o "$wrapper" -
+
+    recordWrapperLore "$wrapper" "$original"
+}
+
+findWrapperOutputRoot(){
+    local wrapper="$1"
+
+    for output in $(getAllOutputNames); do
+        if [[ "$wrapper" == "${!output}"* ]]; then
+            echo "${!output}"
+            return
+        fi
+    done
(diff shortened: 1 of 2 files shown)


abathur:
I'm not certain whether this writes into the "right" output.

Before I started trying to figure out how to support multiple outputs here, the first draft of this wrote these into `${!outputBin}`.

For a bit, I had myself convinced that automatic and manual `moveToOutput` behavior would break some of the linkages here (leaving the record in the wrong output and pointing at a stale path) unless we could either figure out how to write the record into the correct output from the start, or perhaps relocate them during fixup.

Now I'm starting to think it may be okay not to sweat this much, since wrappers should _also_ get broken by moveToOutput (probably leading packages that need both to create wrappers in postFixup or later). To that end, I've force-pushed to update this to try to just discover the store path prefix that the wrapper is being written to, and record the wrapper within that output.

(I think the risk of multi-output handling breaking this linkage is technically still here, but any package with this problem is likely already broken and I suspect that fixing the wrapper will generally fix that problem?)

Issue #314080, opened 2024-05-23
rPackages.quarto patch needs to be updated due to changes upstream


alexvorobiev:
The patch https://github.com/NixOS/nixpkgs/blob/3bea02fa06b3e56f68d98ad3ecb71fe634c45027/pkgs/development/r-modules/default.nix#L1306 needs to be updated. The relevant line upstream is now https://github.com/quarto-dev/quarto-r/blob/main/R/quarto.R#L20

### Notify maintainers
@Kupac @jbedo @b-rodrigues

b-rodrigues:
Thank you, will take this into account for the next bump which should take place soon. We’ll have to remember to use `replace-fail` instead of `replace`!

PR #315015, opened 2024-05-27
rPackages, R: R 4.3.3 -> 4.4.1 and bump rPackages

jbedo:
Update R and rPackages. 

Hydra: https://hydra.nixos.org/jobset/nixpkgs/r-updates

nixpkgs-review:

```
7 packages failed to build:
cantor jasp-desktop labplot python311Packages.rpy2 python311Packages.rpy2.dist sage sageWithDoc

41 packages built:
R R.tex diffoscope diffoscope.dist diffoscope.man jetbrains.dataspell ocamlPackages.ocaml-r postgresql12JitPackages.plr postgresql12Packages.plr postgresql13JitPackages.plr postgresql13Packages.plr postgresql14JitPackages.plr postgresql14Packages.plr postgresql15JitPackages.plr postgresql15Packages.plr postgresql16JitPackages.plr postgresql16Packages.plr python311Packages.cnvkit python311Packages.cnvkit.dist python311Packages.radian python311Packages.radian.dist python311Packages.rchitect python311Packages.rchitect.dist python312Packages.cnvkit python312Packages.cnvkit.dist python312Packages.radian python312Packages.radian.dist python312Packages.rchitect python312Packages.rchitect.dist python312Packages.rpy2 python312Packages.rpy2.dist quarto rWrapper radianWrapper rstudio rstudio-server rstudioServerWrapper rstudioWrapper spark spark_3_4 vscode-extensions.reditorsupport.r
```

`jasp-desktop` has warnings being elevated to errors, e.g.,:
```
jaspObject.cpp:508:25: error: format not a string literal and no format arguments [8;;https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#index-Wformat-security-Werror=format-security8;;]
  508 |                 Rf_error(("nested key \"" + nestedKeyToString(nestedKey, "$") + "\"does not exist in the options!").c_str());
```
There doesn't seem to be a newer jasp to bump to, so we could disable format security for the moment. @TomaSajt 

Sage is blocked by `python3Packages.rpy2`, which is failing using 3.11 during testing due to a segfault. This segfault does not happen with python 3.12. Any suggestions @joelmo? Sage builds fine when disabling tests so this is the only problem to resolve for sage.


TomaSajt:
For `jasp-desktop` I guess we could just do
```
env.NIX_CFLAGS_COMPILE = "-Wno-error=format-security";
```
I think it's fine to include it in this PR.

b-rodrigues:
also, let’s not forget about this: https://github.com/NixOS/nixpkgs/issues/314080

jbedo:
@domenkozar can we enable force pushes to this `r-updates` branch?

domenkozar:
done

jbedo:
The only issue remaining is sage being blocked by rpy2: there's no newer rpy2 to bump to, and sage seems to require python 3.11. Any suggestions? @timokau @collares @7c6f434c

collares:
I spent a while yesterday trying to figure out the crash, but didn't have enough time to reach a conclusion. I will try more later, but if I can't figure out something in the next few days I think we should just update to a Sage 10.4beta9 (soon to be RC) and migrate to Python 3.12.

Something goes wrong when R tries to run rpy2's `_handler_wrap` callback via cffi in `tests/Interface/test_embedded.py:test_parse_error_when_evaluating` (Python tries to access memory at 0x46 as if it were a frame), but I can't yet explain why Python 3.11 + R 4.4.0 crashes and other combinations don't.

collares:
Compiling rpy2 without optimizations, I got the following error:

```
Exception ignored from cffi callback <function _handler_wrap at 0x6594220>:
  File "/nix/store/pwi3r33c8xhic5011a5kcxp7q4h6ic6q-python3-3.11.9-env/lib/python3.11/site-packages/rpy2/rinterface_lib/_rinterface_capi.py", line 690, in _handler_wrap
AttributeError: 'set' object has no attribute 'rlib'
```

I can't yet see how `openrlib` can be a set at [_rinterface_capi:690](https://github.com/rpy2/rpy2/blob/RELEASE_3_5_16/rpy2/rinterface_lib/_rinterface_capi.py#L690), but I will keep thinking.

jbedo:
> I can't yet see how `openrlib` can be a set at [_rinterface_capi:690](https://github.com/rpy2/rpy2/blob/RELEASE_3_5_16/rpy2/rinterface_lib/_rinterface_capi.py#L690), but I will keep thinking.

That doesn't seem possible, perhaps the migration to sage 10.4beta9 you suggested is the way to go. If the RC is due soon we probably don't want to spend too much effort on figuring out this python 3.11 bug.

collares:
I still need to double-check this, but I think the error goes away if both `cffi` and `rpy2` are compiled with 
```
NIX_COMPILE_CFLAGS = "-O0";
NIX_HARDENING_ENABLE = "";
```
which makes me think there's some undefined behavior we were very lucky not to hit until now. If you have evidence that this is a `rpy2` bug, let me know and I will file an issue upstream.

A workaround which doesn't require changing `cffi` compile flags is setting the `RPY2_CFFI_MODE="ABI"` environment flag. I think it might be enough to do it at `rpy2` build time, but some extra work is needed to make the tests find `libR.so`.

Either way, I'll soon investigate upgrading Sage to 10.4beta9, even if just to get a head start at upgrading to rc0.

collares:
The error goes away if both `cffi` and `rpy2` are compiled with 
```
NIX_COMPILE_CFLAGS = "-O0";
NIX_HARDENING_ENABLE = "";
```
which makes me think there's some undefined behavior we were very lucky not to hit until now.
I will try to gather more evidence that this is a `rpy2` bug and file an issue upstream. For Sage purposes, I think the following should therefore be acceptable as a workaround while we wait for Sage 10.4 (assuming Sage tests pass, which I haven't checked):
```diff
diff --git a/pkgs/applications/science/math/sage/default.nix b/pkgs/applications/science/math/sage/default.nix
index 6624dcc36528..d03cec7e3ef1 100644
--- a/pkgs/applications/science/math/sage/default.nix
+++ b/pkgs/applications/science/math/sage/default.nix
@@ -29,6 +29,18 @@ let
       sage-setup = self.callPackage ./python-modules/sage-setup.nix {
         inherit sage-src;
       };
+
+      # there is some miscompilation happening with -O1, and the fortify hardening requires -O1.
+      # TODO: remove when sage migrates to Python 3.12, since affects 3.11 only for some reason.
+      cffi = super.cffi.overrideAttrs (old: {
+        env.NIX_CFLAGS_COMPILE = (old.env.NIX_CFLAGS_COMPILE or "") + " -O0";
+        hardeningDisable = [ "fortify" ];
+      });
+
+      rpy2 = super.rpy2.overrideAttrs (old: {
+        env.NIX_CFLAGS_COMPILE = (old.env.NIX_CFLAGS_COMPILE or "") + " -O0";
+        hardeningDisable = [ "fortify" ];
+      });
     });
   };
```

A workaround which doesn't require changing `cffi` compile flags is setting the `RPY2_CFFI_MODE="ABI"` environment flag, and I would prefer this latter option if you can get it to work. It might be enough to set the flag at `rpy2` build time, but I haven't tested it because some extra work is needed to make the tests find `libR.so`.

Either way, I'll soon investigate upgrading Sage to 10.4beta9, even if just to get a head start at upgrading to rc0.

collares:
@jbedo If I understand correctly, the CI runs at https://github.com/rpy2/rpy2/pull/1051 segfault on the same test (`test_embedded_r.py`) running under Python 3.12, which is consistent with the undefined behavior explanation.

collares:
Filed rpy2/rpy2#1111 just in case.

jbedo:
That's great sluthing, and good to know it isn't resolved with 3.12. I'll spend some time trying to get the ABI mode to work.

collares:
Actually, I am not sure the ABI suggestion definitely helps. If you run the minimized tastcase under `valgrind --trace-children=yes`, both the ABI and the API mode report invalid reads when invoking callbacks. But it might be a good enough workaround while we wait for an upstream fix.

jbedo:
I'm unsure as to the best way forward: this bug really needs to be fixed upstream. Changing ABI/hardening/optimisation really does nothing, we just sidestep the bug during the test case with some luck. Is this any better than disabling tests?  rpy2 is also still broken with the old R 4.3.3 version as you've discovered, so there doesn't seem to be much point to holding the R version bump back waiting for a fix and we may as well pick one of these options and then update rpy2 when an upstream fix is available.

collares:
@ofborg build sageWithDoc

collares (approved):

b-rodrigues:
R 4.4.1 got released last week, what shall we do, merge this one and re-bump, or directly bump to 4.4.1?

jbedo:
> R 4.4.1 got released last week, what shall we do, merge this one and re-bump, or directly bump to 4.4.1?

I'm planning to bump to 4.4.1, just preparing it and checking it. It should be a fairly minor change since we already got 4.4.0 in good shape.

jbedo:
49 packages built:
R R.tex cantor diffoscope diffoscope.dist diffoscope.man jasp-desktop jetbrains.dataspell labplot ocamlPackages.ocaml-r postgresql12JitPackages.plr postgresql12Packages.plr postgresql13JitPackages.plr postgresql13Packages.plr postgresql14JitPackages.plr postgresql14Packages.plr postgresql15JitPackages.plr postgresql15Packages.plr postgresql16JitPackages.plr postgresql16Packages.plr python311Packages.cnvkit python311Packages.cnvkit.dist python311Packages.radian python311Packages.radian.dist python311Packages.rchitect python311Packages.rchitect.dist python311Packages.rpy2 python311Packages.rpy2.dist python312Packages.cnvkit python312Packages.cnvkit.dist python312Packages.radian python312Packages.radian.dist python312Packages.rchitect python312Packages.rchitect.dist python312Packages.rpy2 python312Packages.rpy2.dist quarto rWrapper radianWrapper resorter rstudio rstudio-server rstudioServerWrapper rstudioWrapper sage sageWithDoc
spark spark_3_4 vscode-extensions.reditorsupport.r

b-rodrigues (approved):
I don't see any other major packages that must absolutely be fixed, so for me this can be merged!

PR #315025, opened 2024-05-27
dssat: init at 4.8.2.12

pcboy:
## Description of changes

DSSAT is a famous tool used for crop modeling. The first release was in 1989 but this is still in use today. It's in Fortran.

https://en.wikipedia.org/wiki/DSSAT

This package is only for the CLI version.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [x] x86_64-darwin
  - [x] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [x] `sandbox = true`
- [ ] Tested, as applicable:

## Notes

~~This is for DSSAT47. DSSAT 48 also exists. Instead of naming the package dssat, I decided to call it dssat47 as this is common to have both dssat47 and dssat48 on the same system right now as some model files only work for dssat47.  
I don't have yet a working package for DSSAT48 (yet).  
What is the recommended way to deal with this situation? Should I have 2 folders dssat47 and dssat48 in `pkgs/by-name/dss/` and use `dssat47.override{}` in the dssat48 package?  
Or should I maybe have a `dssat` package pointing to latest (48), and a separate `dssat47` package?~~

Not valid anymore. Now it is only DSSAT48.

### DIFF
--- a/pkgs/by-name/ds/dssat/package.nix
+++ b/pkgs/by-name/ds/dssat/package.nix
@@ -1,0 +1,74 @@
+{
+  stdenv,
+  lib,
+  cmake,
+  glibc,
+  gfortran,
+  makeWrapper,
+  fetchFromGitHub,
+  fetchpatch,
+  dos2unix,
+  dataRepo ? fetchFromGitHub {
+    owner = "DSSAT";
+    repo = "dssat-csm-data";
+    rev = "v4.8.2.8";
+    hash = "sha256-hbSBKEvdSd1lfbemfp4Lk4/JcGMXGVjm1x7P7lmmuA0=";
+  },
+}:
+let
+  # Temporary patch to fix 80 chars limit on paths
+  # https://github.com/DSSAT/dssat-csm-os/pull/417/
+  charLimitPatch = fetchpatch {
+    url = "https://github.com/DSSAT/dssat-csm-os/pull/417/commits/9215012a297c074f392b5e7eb90b8c20495f13f7.patch";
+    hash = "sha256-WwJR5lnWtR3aYWZmk8pBC0/qaRqY0UrWHIaYp2ajImE=";
+  };
+in
+stdenv.mkDerivation (final: {
+  pname = "dssat";
+  version = "4.8.2.12";
+
+  src = fetchFromGitHub {
+    owner = "DSSAT";
+    repo = "dssat-csm-os";
+    rev = "refs/tags/v${final.version}";
+    sha256 = "sha256-8OaTM7IXFZjlelx5O4O+bVNQj4dIhGzIk2iCfpqI8uA=";
+  };
+
+  # maintainers are on windows and have CRLF endings in their files
+  # And github returns a patch file in unix format only.
+  patchPhase = ''
+    runHook prePatch
+    cp ${charLimitPatch} ./limit-path.patch
+    unix2dos ./limit-path.patch
+    patch --binary -p1 < ./limit-path.patch
+    runHook postPatch
+  '';
+
+  nativeBuildInputs = [
+    cmake
+    dos2unix
+    gfortran
+    makeWrapper
+  ];
+
+  buildInputs = lib.optionals stdenv.isLinux [ glibc.static ];
+
+  cmakeFlags = [ "-DCMAKE_INSTALL_PREFIX=${placeholder "out"}/share/dssat/" ];
+
+  postInstall = ''
+    mkdir -p $out/share/dssat/Data
+    cp -r $src/Data/* $out/share/dssat/Data/
+    cp -r ${dataRepo}/* $out/share/dssat/Data/
+    makeWrapper $out/share/dssat/dscsm048 $out/bin/dscsm048
+  '';
+
+  meta = {
+    homepage = "https://github.com/DSSAT/dssat-csm-os";
+    description = "Cropping System Model";
+    mainProgram = "dscsm048";
+    license = lib.licenses.bsd3;
+    maintainers = with lib.maintainers; [ pcboy ];
+    platforms = lib.platforms.unix;
+    broken = stdenv.isAarch64 && stdenv.isLinux;
+  };
+})
--- a/maintainers/maintainer-list.nix
+++ b/maintainers/maintainer-list.nix
@@ -15752,6 +15752,12 @@
     githubId = 817039;
     name = "Paulo Casaretto";
   };
+  pcboy = {
+    email = "david@joynetiks.com";
+    github = "pcboy";
+    githubId = 943430;
+    name = "David Hagege";
+  };
   pedrohlc = {
     email = "root@pedrohlc.com";
     github = "PedroHLC";


paparodeo (commented):
there is no need to override buildPhase and installPhase as you can use the the default cmake ones. suggested changes below. ~also, why use a change from 3 years ago and not the current tag -- 4.8.2.xx?~ [edit] just saw the note


<details><summary>suggested changes.diff</summary>

```diff
diff --git a/pkgs/by-name/ds/dssat47/package.nix b/pkgs/by-name/ds/dssat47/package.nix
index 573fa012d4e9..9ff7c6b0bcc5 100644
--- a/pkgs/by-name/ds/dssat47/package.nix
+++ b/pkgs/by-name/ds/dssat47/package.nix
@@ -1,7 +1,6 @@
 { stdenv
 , lib
 , cmake
-, makeWrapper
 , glibc
 , gfortran
 , fetchFromGitHub
@@ -9,42 +8,31 @@
     owner = "DSSAT";
     repo = "dssat-csm-data";
     rev = "v4.8.2.8";
-    sha256 = "sha256-hbSBKEvdSd1lfbemfp4Lk4/JcGMXGVjm1x7P7lmmuA0=";
+    hash = "sha256-hbSBKEvdSd1lfbemfp4Lk4/JcGMXGVjm1x7P7lmmuA0=";
   }
-, ...
 }:
-stdenv.mkDerivation {
+stdenv.mkDerivation (final: {
   pname = "dssat47";
   version = "4.7.5.42";
 
   src = fetchFromGitHub {
     owner = "DSSAT";
     repo = "dssat-csm-os";
-    rev = "3ccad41098ac7c7a162f1d463bd09c82e8931845";
-    sha256 = "sha256-3/9iaz1CJ9iUOlhGvTq9b/0IiLcNZc08Qh1Ij1Ky/44=";
+    rev = "refs/tags/v${final.version}";
+    hash = "sha256-3/9iaz1CJ9iUOlhGvTq9b/0IiLcNZc08Qh1Ij1Ky/44=";
   };
-  dontUseCmakeConfigure = true;
-  nativeBuildInputs = [ cmake makeWrapper ];
+  nativeBuildInputs = [ cmake gfortran ];
 
-  buildInputs = [ gfortran ] ++ (lib.optionals stdenv.isLinux [ glibc.static ]);
+  buildInputs = lib.optionals stdenv.isLinux [ glibc.static ];
 
-  buildPhase = ''
-    runHook preBuild
-    mkdir -p build $out/bin
-    cd build
-    cmake -DCMAKE_BUILD_TYPE=RELEASE -DCMAKE_INSTALL_PREFIX=$out/bin ..
-    runHook postBuild
-  '';
-
-  installPhase = ''
-    runHook preInstall
-    make install
-    mkdir -p $out/bin/Data
-    cp bin/dscsm047 $out/bin/
+  cmakeFlags = [
+    "-DCMAKE_INSTALL_PREFIX=${placeholder "out"}/bin"
+  ];
 
-    cp -rf $src/Data/* $out/bin/Data/
-    cp -rf ${dataRepo}/* $out/bin/Data/
-    runHook postInstall
+  postInstall = ''
+    mkdir $out/bin/Data
+    cp -r $src/Data/* $out/bin/Data/
+    cp -r ${dataRepo}/* $out/bin/Data/
   '';
 
   meta = {
@@ -54,4 +42,4 @@ stdenv.mkDerivation {
     license = lib.licenses.bsd3;
     maintainers = with lib.maintainers; [ pcboy ];
   };
-}
+})
```
</details>

pcboy:
Thank you @paparodeo for the review.

That's very good points, I applied your changes. :+1: 

Somehow I had issues originally just doing the default cmake configure. Must have been because of something else. It seems to be working fine now.

paparodeo (commented):
~i don't know off hand what the preferred method for having multiple versions is in the `by-name` hierarchy and i am not seeing any examples when i look over `pkgs/top-level/all-packages.nix`.~ [edit] see: https://github.com/NixOS/nixpkgs/tree/master/pkgs/by-name#recommendation-for-new-packages-with-multiple-versions

one thing i noticed is all the files are installed into `bin` which seems undesirable. Without modifying the build system, perhaps can install in `$out/libexec/dssat47/` then either using a symlink or `makeWrapper` link the executable from `$out/bin/dssat47` to  `$out/libexec/dssat47/`. tho that still seems a little hacky, i think that data is usually stored in `$out/shared/pkg-name/`

https://github.com/NixOS/nixpkgs/blob/master/pkgs/README.md#new-packages
so the meta sections needs `platforms` which could be unix but looks like it needs to mark aarch64-linux as broken.

note -- i do not have commit access so can not merge a commit and just wanted to give some basic feedback.

pcboy:
> one thing i noticed is all the files are installed into bin which seems undesirable

Yes I completely agree. Thing is this software is super old and working in strange ways, it's extremely sensitive to where the files are.  
I tried to put the files in the `$out/shared` folder and originally it failed, but after changing `pname = "dssat47"` to `name = "dssat47"` it works... which is really strange.

When I have `pname = "dssat47"`, for instance it gives:

```
 File not found. Please check file name or create file.                          
File: ///nix/store/38zb9hx0mz0n051h26x57cl8   Line:     0   Error key: CO2VAL

Additional information may be available in WARNING.OUT file.
```

The WARNING.OUT file contains:

```
 CO2 input file does not exist or has invalid format.                          
///nix/store/38zb9hx0mz0n051h26x57cl8qkb2ik7a-dssat47-4.7.5.42/bin/Standard/CO
 Line number    0                                                              
 Program will stop.                                                            

 CO2VAL  YEAR DOY = 2022 125
 File not found. Please check file name or create file.                        
 File: ///nix/store/zry2zmdnndm44ci05arvnfcn                                   
    Line:     0                                                                
    Error key: CO2VAL                                                          
 Simulations terminated.           
```

but it works fine when I have `name = "dssat47"` in my package... It seems that there is an issue with the resulting path being too long maybe? :thinking: 
 
I put the MR as draft until I investigate more.

pcboy:
The original program had a char path limit of 80 chars. Which was quickly an issue with the nix store paths of course.

I added a patch for it to work. This seems to be working fine on my machine.  
I'm trying to merge the patch upstream, but it may take months.

Notes:

- This is now DSSAT 4.8.2.12, I gave up on trying to have both 47 and 48 because of the path issue and not wanting to maintain both. This is likely fine. Package is then now called `dssat` only.

- I tried to directly use `patches = []` to apply the patch, but it wasn't working because github returns patch files in unix format, while the code is formatted with CLRF (windows devs it seems). Just trying to apply with `--binary` was only working after the patch file was unix2dos'ed.

pkgs/by-name/ds/dssat/package.nix
```diff
@@ -0,0 +1,71 @@
...
+, fetchFromGitHub
+, fetchpatch
+, dos2unix
+, dataRepo ? fetchFromGitHub {
+    owner = "DSSAT";
+    repo = "dssat-csm-data";
+    rev = "v4.8.2.8";
+    hash = "sha256-hbSBKEvdSd1lfbemfp4Lk4/JcGMXGVjm1x7P7lmmuA0=";
+  }
+, ...
```
djacu, line 16: This isn't a module. Is the `...` necessary?

```suggestion
```

djacu (changes requested):
Since this PR adds a new file, please format your `package.nix` file with the official formatter, [`nixfmt-rfc-style`](https://search.nixos.org/packages?channel=24.05&show=nixfmt-rfc-style&from=0&size=50&sort=relevance&type=packages&query=nixfmt-rfc-style). It provides a binary, `nixfmt`.

pkgs/by-name/ds/dssat/package.nix
pcboy: Thank you! Fixed.

I was using `nixpkgs-fmt` until now. Good to know. :+1:

djacu (approved):

djacu:
Result of `nixpkgs-review pr 315025` run on x86_64-linux [1](https://github.com/Mic92/nixpkgs-review)
<details>
  <ul>
    <li>dssat</li>
  </ul>
</details>

pkgs/by-name/ds/dssat/package.nix
```diff
@@ -0,0 +1,73 @@
...
+let
+  # Temporary patch to fix 80 chars limit on paths
+  # https://github.com/DSSAT/dssat-csm-os/pull/417/
+  charLimitPatch = fetchpatch {
+    url = "https://github.com/DSSAT/dssat-csm-os/pull/417/commits/9215012a297c074f392b5e7eb90b8c20495f13f7.patch";
+    hash = "sha256-WwJR5lnWtR3aYWZmk8pBC0/qaRqY0UrWHIaYp2ajImE=";
+  };
+in
+stdenv.mkDerivation (final: {
+  name = "dssat";
```
OPNA2608, line 27:
```suggestion
  pname = "dssat";
```

`name` is `pname` and `version` combined.

pkgs/by-name/ds/dssat/package.nix
```diff
@@ -0,0 +1,73 @@
...
+    owner = "DSSAT";
+    repo = "dssat-csm-os";
+    rev = "refs/tags/v${final.version}";
+    sha256 = "sha256-8OaTM7IXFZjlelx5O4O+bVNQj4dIhGzIk2iCfpqI8uA=";
+  };
+
+  # maintainers are on windows and have CRLF endings in their files
+  # And github returns a patch file in unix format only.
+  patchPhase = ''
+    runHook prePatchPhase
```
OPNA2608, line 40:
```suggestion
    runHook prePatch
```

pkgs/by-name/ds/dssat/package.nix
```diff
@@ -0,0 +1,73 @@
...
+    cp ${charLimitPatch} ./limit-path.patch
+    ${lib.getExe' dos2unix "unix2dos"} ./limit-path.patch
+    patch --binary -p1 < ./limit-path.patch
+    runHook postPatchPhase
```
OPNA2608, line 44:
```suggestion
    runHook postPatch
```

pkgs/by-name/ds/dssat/package.nix
```diff
@@ -0,0 +1,73 @@
...
+    rev = "refs/tags/v${final.version}";
+    sha256 = "sha256-8OaTM7IXFZjlelx5O4O+bVNQj4dIhGzIk2iCfpqI8uA=";
+  };
+
+  # maintainers are on windows and have CRLF endings in their files
+  # And github returns a patch file in unix format only.
+  patchPhase = ''
+    runHook prePatchPhase
+    cp ${charLimitPatch} ./limit-path.patch
+    ${lib.getExe' dos2unix "unix2dos"} ./limit-path.patch
```
OPNA2608, line 42:
```suggestion
    unix2dos ./limit-path.patch
```

pkgs/by-name/ds/dssat/package.nix
```diff
@@ -0,0 +1,73 @@
...
+    patch --binary -p1 < ./limit-path.patch
+    runHook postPatchPhase
+  '';
+
+  nativeBuildInputs = [
+    cmake
```
OPNA2608, line 48:
```suggestion
    cmake
    dos2unix
```

OPNA2608 (changes requested):

pkgs/by-name/ds/dssat/package.nix
pcboy: Thank you @OPNA2608 , all changed applied! :+1:

pcboy:
Thanks @OPNA2608 for the review.  I've addressed all the suggested changes. Would you be willing to revisit and consider merging this one? :pray:

pkgs/by-name/ds/dssat/package.nix
```diff
@@ -0,0 +1,74 @@
...
+    dos2unix
+    gfortran
+    makeWrapper
+  ];
+
+  buildInputs = lib.optionals stdenv.isLinux [ glibc.static ];
+
+  cmakeFlags = [ "-DCMAKE_INSTALL_PREFIX=${placeholder "out"}/shared/dssat/" ];
```
OPNA2608, line 56: Is `shared` vs `share` intentional? Same for all other uses of this path below.

pcboy: @paparodeo originally mentioned `shared` which is why I put the data there. But likely a typo as indeed I see people using `share` instead, which makes sense.

I made the changes. Thank you.

paparodeo: oops -- sorry about that.

OPNA2608 (approved):

PR #315119, opened 2024-05-27
webp-pixbuf-loader: fix cross-comilation

Mic92:
## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [x] Tested, as applicable:

---

### DIFF
--- a/pkgs/development/libraries/webp-pixbuf-loader/default.nix
+++ b/pkgs/development/libraries/webp-pixbuf-loader/default.nix
@@ -52,6 +52,9 @@ stdenv.mkDerivation rec {
     GDK_PIXBUF_MODULEDIR="$out/${moduleDir}" \
     gdk-pixbuf-query-loaders --update-cache
 
+    # gdk-pixbuf disables the thumbnailer in cross-builds (https://gitlab.gnome.org/GNOME/gdk-pixbuf/-/commit/fc37708313a5fc52083cf10c9326f3509d67701f)
+    # and therefore makeWrapper will fail because 'gdk-pixbuf-thumbnailer' the executable does not exist.
+  '' + lib.optionalString (stdenv.hostPlatform == stdenv.buildPlatform) ''
     # It assumes gdk-pixbuf-thumbnailer can find the webp loader in the loaders.cache referenced by environment variable, breaking containment.
     # So we replace it with a wrapped executable.
     mkdir -p "$out/bin"


pkgs/development/libraries/webp-pixbuf-loader/default.nix
```diff
@@ -52,6 +52,8 @@ stdenv.mkDerivation rec {
     GDK_PIXBUF_MODULEDIR="$out/${moduleDir}" \
     gdk-pixbuf-query-loaders --update-cache
 
+    # gdk-pixbuf disables the thumbnailer in cross-builds.
```
jtojnar, line 55: How did you determine this? I see no such thing in https://github.com/aruiz/webp-pixbuf-loader/blob/0.2.7/meson.build

Mic92: It's disabled in **gdk-pixbuf** not webp-pixbuf-loader. Checkout this: https://gitlab.gnome.org/GNOME/gdk-pixbuf/-/blob/master/meson.build?ref_type=heads#L397

jtojnar: Hmm, [indeed](https://gitlab.gnome.org/GNOME/gdk-pixbuf/-/commit/fc37708313a5fc52083cf10c9326f3509d67701f). That makes little sense to me. The only thing that could be preventing thumbnailer from working is not generating the `loaders.cache` but we are creating that ourselves using emulator:

https://github.com/NixOS/nixpkgs/blob/e6719289cdd68c290fa076a6409a2d89535ade5e/pkgs/development/libraries/gdk-pixbuf/default.nix#L123-L124

But I would assume this will be issue in cross here as well, yet you are keeping the line above.

Also, since we do install the `.thumbnailer` file unconditionally in this package, the thumbnailer will not be fully disabled – I believe it will just fail to be discovered at runtime.

Mic92: So what are your suggested next steps? I tried enabling the thumbnailer executable in their buildsystem but this will need a more complex patch inside their buildsystem since this executable also has dependencies on `loaders.cache`. So disabling the wrapper here seemed more straightforward to me.

jtojnar: If this workaround fixes cross for you, it is good enough for me, the change is thankfully unobtrusive. I am just surprised `gdk-pixbuf-query-loaders` above works in cross.

Would be nice to include more detailed explanation of what exactly is broken in the commit message (e.g. `makeWrapper` expects the wrapped file to exists) and limitations (keeping the thumbnailer file pointing to non-existent `gdk-pixbuf-thumbnailer-webp` but it is okay since thumbnailer machinery will disable `thumnailer` files with non-executable paths in `Exec` key).

Mic92: Done.

uninsane:
probably OK to target master if you want, since it only affects cross?

can confirm that everything works as expected with this patch (epiphany web browser, loupe image viewer, fractal matrix client, ...), x86_64 -> aarch64 cross compilation.

i'd strongly encourage including more explanation in the `postInstall` for why we do this though (maybe a link to an upstream issue or relevant code), because i suspect the long term thing here is to fix upstream.

i believe [this gdk-pixbuf changeset](https://gitlab.gnome.org/GNOME/gdk-pixbuf/-/merge_requests/20) was relevant. i recall patching `if not meson.is_cross_build()` to be `if true` and building a whole system (webp-pixbuf-thumbnailer included) just fine ~18 mo ago, but i can't remember why i moved away from that approach (maybe it required emulation).

Mic92:
> i'd strongly encourage including more explanation in the `postInstall` for why we do this though (maybe a link to an upstream issue or relevant code), because i suspect the long term thing here is to fix upstream.
> 

Upstream 



> probably OK to target master if you want, since it only affects cross?
> 
> can confirm that everything works as expected with this patch (epiphany web browser, loupe image viewer, fractal matrix client, ...), x86_64 -> aarch64 cross compilation.
> 
> i'd strongly encourage including more explanation in the `postInstall` for why we do this though (maybe a link to an upstream issue or relevant code), because i suspect the long term thing here is to fix upstream.

Linked to the upstream issue.
 
> i recall patching `if not meson.is_cross_build()` (from the gdk-pixbuf changeset linked earlier) to be `if true` and building a whole system (webp-pixbuf-thumbnailer included) just fine ~18 mo ago, but i can't remember why i moved away from that approach (maybe it required emulation).

That did not work for me, there some more dependencies that don't cross-build

jtojnar:
> Upstream thinks this is not necessary because the icon cache will be build with host binaries during the package process anyway.

Sure, but I was talking about `gdk-pixbuf-thumbnailer`, not icon cache. I do not see why that could not be cross-compiled.



> Linked to the upstream issue.

I do not see any such link.

Mic92:
> > Upstream thinks this is not necessary because the icon cache will be build with host binaries during the package process anyway.
> 
> Sure, but I was talking about `gdk-pixbuf-thumbnailer`, not icon cache. I do not see why that could not be cross-compiled.

Afaik they have some icon cache file that is required for thumbnailer to build. It didn't see a trivial fix for that. Just removing the is cross conditions didn't solve this.

> 
> > Linked to the upstream issue.
> 
> I do not see any such link.

Here is a link where they disabled it.
https://gitlab.gnome.org/GNOME/gdk-pixbuf/-/commit/fc37708313a5fc52083cf10c9326f3509d67701f

PR #315121, opened 2024-05-27
Kernel updates 2024-05-27 (incl 6.10-rc1)

alyssais:
## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/os-specific/linux/kernel/common-config.nix
+++ b/pkgs/os-specific/linux/kernel/common-config.nix
@@ -441,7 +441,8 @@ let
       # (stable) amdgpu support for bonaire and newer chipsets
       DRM_AMDGPU_CIK = yes;
       # Allow device firmware updates
-      DRM_DP_AUX_CHARDEV = yes;
+      DRM_DP_AUX_CHARDEV = whenOlder "6.10" yes;
+      DRM_DISPLAY_DP_AUX_CHARDEV = whenAtLeast "6.10" yes;
       # amdgpu display core (DC) support
       DRM_AMD_DC_DCN1_0 = whenOlder "5.6" yes;
       DRM_AMD_DC_DCN2_0 = whenBetween "5.3" "5.6" yes;
@@ -470,7 +471,8 @@ let
       MEDIA_CEC_RC = whenAtLeast "5.10" yes;
 
       # Enable CEC over DisplayPort
-      DRM_DP_CEC = yes;
+      DRM_DP_CEC = whenOlder "6.10" yes;
+      DRM_DISPLAY_DP_AUX_CEC = whenAtLeast "6.10" yes;
     } // optionalAttrs (stdenv.hostPlatform.system == "x86_64-linux") {
       # Intel GVT-g graphics virtualization supports 64-bit only
       DRM_I915_GVT = yes;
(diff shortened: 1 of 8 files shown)


K900 (approved):
Diff LGTM

Ma27 (changes requested):
The Rust patch seems already applied for `/nix/store/z4sf9jia1rzqlkpgbx3985glaxk8hy9g-linux-6.10-rc1.drv`:

<details><summary>error log</summary>

```
       last 25 log lines:
       > Perhaps you used the wrong -p or --strip option?
       > The text leading up to this was:
       > --------------------------
       > |diff --git a/rust/alloc/vec/mod.rs b/rust/alloc/vec/mod.rs
       > |index 220fb9d6f45b3f..0be27fff4554a1 100644
       > |--- a/rust/alloc/vec/mod.rs
       > |+++ b/rust/alloc/vec/mod.rs
       > --------------------------
       > File to patch:
       > Skip this patch? [y]
       > Skipping patch.
       > 15 out of 15 hunks ignored
       > patching file rust/kernel/lib.rs
       > Reversed (or previously applied) patch detected!  Assume -R? [n]
       > Apply anyway? [n]
       > Skipping patch.
       > 2 out of 2 hunks ignored -- saving rejects to file rust/kernel/lib.rs.rej
       > patching file scripts/Makefile.build
       > Reversed (or previously applied) patch detected!  Assume -R? [n]
       > Apply anyway? [n]
       > Skipping patch.
       > 1 out of 1 hunk ignored -- saving rejects to file scripts/Makefile.build.rej
       > patching file scripts/min-tool-version.sh
       > Hunk #1 FAILED at 33.
       > 1 out of 1 hunk FAILED -- saving rejects to file scripts/min-tool-version.sh.rej
       For full logs, run 'nix log /nix/store/2rarf83b3j9202d533i8bdzdspk5hd1a-linux-config-6.10-rc1.drv'.
```

</details>

alyssais:
Ah right, of course.

Ma27:
fwiw rest of the changed kernels seem good though.

Nanotwerp (approved):
LGTM! The removal of the now-redundant `Rust` patch and the renaming of the config options fixed my compilation of the kernel.

alyssais:
linux_testing doesn't build for riscv64, I assume because of https://git.kernel.org/linus/e79dfcbfb902a99268cc8022031461da7a8e2bc8.  It will now unconditionally try install Image.zst, which we haven't built because platforms.nix says we want Image on Linux, and that can't just be changed to Image.zst because that doesn't exist in older kernels.

pkgs/os-specific/linux/kernel/common-config.nix
```diff
@@ -927,8 +929,9 @@ let
       # i686 issues: https://github.com/NixOS/nixpkgs/pull/117961#issuecomment-812106375
       useZstd = stdenv.buildPlatform.is64bit && versionAtLeast version "5.9";
     in {
-      KERNEL_XZ            = mkIf (!useZstd) yes;
-      KERNEL_ZSTD          = mkIf useZstd yes;
+      KERNEL_UNCOMPRESSED  = mkIf stdenv.hostPlatform.isRiscV yes;
```
Ma27, line 932: A comment on the rationale for riscv may be useful here.

Ma27 (approved):
May I ask why we assume that Risc-V uses an uncompressed kernel? Not sure I understand that part.

Approving anyways, since this is not my area of expertise and I don't want to block an otherwise valid PR because I probably don't know enough about that.

alyssais:
> May I ask why we assume that Risc-V uses an uncompressed kernel?

From the commit message:

We assume RISC-V uses an uncompressed kernel in lib/systems/platforms.nix.  Since this is the first Linux version to support either xz or zstd compressed kernels on RISC-V, we can't change platforms.nix to match the other platforms without breaking builds of older kernel versions, so let's just preserve the old behaviour.

> We could in future decide to use gzip-compressed kernels on RISC-V, which is supported going back further.

pkgs/os-specific/linux/kernel/common-config.nix
alyssais: Added.

Ma27:
Sorry, understood. Thanks!

pkgs/os-specific/linux/kernel/common-config.nix
alyssais: Done.

alyssais:
@blitz @Ma27 @yu-re-ka I've just noticed that this has sadly regressed Rust on aarch64.  It now uses the aarch64-unknown-none target instead of the default one.  We could just build support for that into our default rustc, except that the build fails because it can't generate documentation for std for that target (because there is no std).  I think the only good way to resolve this is going to be for Rust's build system to be fixed to skip trying to generate std documentation only for no_std targets, or to at least provide that option.  I don't think I'm going to be able to work on that for at least a while.

PR #315139, opened 2024-05-27
finamp: init at 0.9.8-beta

DSeeLP:
## Description of changes
[Finamp](https://github.com/jmshrv/finamp)

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [x] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/maintainers/maintainer-list.nix
+++ b/maintainers/maintainer-list.nix
@@ -5351,6 +5351,11 @@
     githubId = 5596239;
     keys = [ { fingerprint = "62BC E2BD 49DF ECC7 35C7  E153 875F 2BCF 163F 1B29"; } ];
   };
+  dseelp = {
+    name = "dsee";
+    github = "DSeeLP";
+    githubId = 46624152;
+  };
   dsferruzza = {
     email = "david.sferruzza@gmail.com";
     github = "dsferruzza";
--- a/pkgs/by-name/fi/finamp/package.nix
+++ b/pkgs/by-name/fi/finamp/package.nix
@@ -1,0 +1,71 @@
+{ lib
+, flutter322
+, mpv-unwrapped
+, xdg-user-dirs
+, patchelf
+, fetchFromGitHub
+, copyDesktopItems
+, makeDesktopItem
+}:
+let
+  version = "0.9.8-beta";
+in
+flutter322.buildFlutterApplication {
+  inherit version;
+  pname = "finamp";
+  src = fetchFromGitHub {
+    owner = "jmshrv";
+    repo = "finamp";
+    rev = version;
+    hash = "sha256-lvjhA+hdCXgDsrNhNw4Tiq6ZgkYlPuMeHha8OJNF1TI=";
+  };
+  pubspecLock = lib.importJSON ./pubspec.lock.json;
+
+  nativeBuildInputs = [ patchelf copyDesktopItems ];
+  buildInputs = [ mpv-unwrapped ];
+
+  gitHashes = {
+    balanced_text = "sha256-lSDR5dDjZ4garRbBPI+wSxC5iScg8wVSD5kymmLbYbk=";
+    isar_generator = "sha256-lWnHmZmYx7qDG6mzyDqYt+Xude2xVOH1VW+BoDCas60=";
+    media_kit_libs_windows_audio = "sha256-p3hRq79whLFJLNUgL9atXyTGvOIqCbTRKVk1ie0Euqs=";
+    palette_generator = "sha256-mnRJf3asu1mm9HYU8U0di+qRk3SpNFwN3S5QxChpIA0=";
+    split_view = "sha256-unTJQDXUUPVDudlk0ReOPNYrsyEpbd/UMg1tHZsmg+k=";
+  };
+
+  postFixup = ''
+    patchelf $out/app/finamp --add-needed libisar.so --add-needed libmpv.so --add-rpath ${lib.makeLibraryPath [ mpv-unwrapped ]}
+  '';
+
+  postInstall = ''
+    install -Dm644 $src/assets/icon/icon_foreground.svg $out/share/icons/hicolor/scalable/apps/finamp.svg
+  '';
+
+  extraWrapProgramArgs = ''
+    --prefix PATH : ${lib.makeBinPath [ xdg-user-dirs ]}
+  '';
+
+  desktopItems = [(makeDesktopItem {
+    name = "Finamp";
+    desktopName = "Finamp";
+    genericName = "Music Player";
+    exec = "finamp";
+    icon = "finamp";
+    startupWMClass = "finamp";
+    comment = "An open source Jellyfin music player";
+    categories = [
+      "AudioVideo"
+      "Audio"
+      "Player"
+      "Music"
+    ];
+  })];
+
+  meta = {
+    description = "Open source Jellyfin music player";
+    homepage = "https://github.com/jmshrv/finamp";
+    license = lib.licenses.mpl20;
+    maintainers = with lib.maintainers; [ dseelp ];
+    mainProgram = "finamp";
+    platforms = lib.platforms.linux;
+  };
+}
--- a/pkgs/by-name/fi/finamp/pubspec.lock.json
+++ b/pkgs/by-name/fi/finamp/pubspec.lock.json
@@ -1,0 +1,2093 @@
+{
+  "packages": {
+    "_fe_analyzer_shared": {
+      "dependency": "transitive",
+      "description": {
+        "name": "_fe_analyzer_shared",
+        "sha256": "0b2f2bd91ba804e53a61d757b986f89f1f9eaed5b11e4b2f5a2468d86d6c9fc7",
+        "url": "https://pub.dev"
+      },
+      "source": "hosted",
+      "version": "67.0.0"
+    },
+    "analyzer": {
+      "dependency": "transitive",
+      "description": {
+        "name": "analyzer",
+        "sha256": "37577842a27e4338429a1cbc32679d508836510b056f1eedf0c8d20e39c1383d",
+        "url": "https://pub.dev"
+      },
+      "source": "hosted",
+      "version": "6.4.1"
+    },
+    "analyzer_plugin": {
+      "dependency": "transitive",
+      "description": {
+        "name": "analyzer_plugin",
+        "sha256": "9661b30b13a685efaee9f02e5d01ed9f2b423bd889d28a304d02d704aee69161",
+        "url": "https://pub.dev"
+      },
+      "source": "hosted",
+      "version": "0.11.3"
+    },
+    "archive": {
+      "dependency": "transitive",
+      "description": {
+        "name": "archive",
+        "sha256": "6bd38d335f0954f5fad9c79e614604fbf03a0e5b975923dd001b6ea965ef5b4b",
+        "url": "https://pub.dev"
+      },
+      "source": "hosted",
+      "version": "3.6.0"
+    },
+    "args": {
+      "dependency": "transitive",
+      "description": {
+        "name": "args",
+        "sha256": "7cf60b9f0cc88203c5a190b4cd62a99feea42759a7fa695010eb5de1c0b2252a",
+        "url": "https://pub.dev"
+      },
+      "source": "hosted",
+      "version": "2.5.0"
+    },
+    "async": {
+      "dependency": "transitive",
+      "description": {
+        "name": "async",
+        "sha256": "947bfcf187f74dbc5e146c9eb9c0f10c9f8b30743e341481c1e2ed3ecc18c20c",
+        "url": "https://pub.dev"
+      },
+      "source": "hosted",
+      "version": "2.11.0"
+    },
+    "audio_service": {
+      "dependency": "direct main",
+      "description": {
+        "name": "audio_service",
+        "sha256": "4547c312a94f9cb2c48b60823fb190767cbd63454a83c73049384d5d3cba4650",
+        "url": "https://pub.dev"
+      },
+      "source": "hosted",
+      "version": "0.18.13"
+    },
+    "audio_service_mpris": {
+      "dependency": "direct main",
+      "description": {
+        "name": "audio_service_mpris",
+        "sha256": "a8d1583f9143d17b2facc994a99bd1ea257cec43adcb8d7349458555c62b570f",
+        "url": "https://pub.dev"
+      },
+      "source": "hosted",
+      "version": "0.1.3"
+    },
+    "audio_service_platform_interface": {
+      "dependency": "direct main",
+      "description": {
+        "name": "audio_service_platform_interface",
+        "sha256": "8431a455dac9916cc9ee6f7da5620a666436345c906ad2ebb7fa41d18b3c1bf4",
+        "url": "https://pub.dev"
+      },
+      "source": "hosted",
+      "version": "0.1.1"
+    },
+    "audio_service_web": {
+      "dependency": "transitive",
+      "description": {
+        "name": "audio_service_web",
+        "sha256": "9d7d5ae5f98a5727f2580fad73062f2484f400eef6cef42919413268e62a363e",
+        "url": "https://pub.dev"
+      },
+      "source": "hosted",
+      "version": "0.1.2"
+    },
+    "audio_session": {
+      "dependency": "direct main",
+      "description": {
+        "name": "audio_session",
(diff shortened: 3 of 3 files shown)


pkgs/by-name/fi/finamp/package.nix
```diff
@@ -0,0 +1,69 @@
...
+    ];
+  };
+
+  meta = with lib; {
+    description = "A Jellyfin music client for mobile and desktop";
+    homepage = "https://github.com/jmshrv/finamp";
+    license = licenses.mpl20;
+    platforms = platforms.linux;
+    mainProgram = "finamp";
+    maintainers = with lib.maintainers; [ dseelp ];
```
nicolas-goudry, line 67: This line is already scoped with `lib`, either:

* use `with maintainers` (preferred)
* keep it as is but unscope other attributes (ie. `license` and `platforms`)
* unscope everything

pkgs/by-name/fi/finamp/package.nix
```diff
@@ -0,0 +1,69 @@
...
+    cp $src/assets/icon/icon_foreground.svg $d/finamp.svg
+
+    mkdir -p $out/share/applications
+    cp $desktopItem/share/applications/*.desktop $out/share/applications
+    substituteInPlace $out/share/applications/*.desktop --subst-var out
+  '';
+
+  desktopItem = makeDesktopItem {
+    name = "Finamp";
+    exec = "@out@/bin/finamp";
```
nicolas-goudry, line 50:
```suggestion
    exec = "finamp";
```

I think you can just use `"finamp"` here, since upon installation it will be in the PATH.

This would also remove line 45 (`substituteInPlace`).

pkgs/by-name/fi/finamp/package.nix
```diff
@@ -0,0 +1,69 @@
...
+    startupWMClass = "finamp";
+    genericName = "A Jellyfin music client for mobile and desktop";
+    categories = [
+      "Audio"
+      "Player"
+    ];
+  };
+
+  meta = with lib; {
+    description = "A Jellyfin music client for mobile and desktop";
```
nicolas-goudry, line 62: Just a suggestion, but since NixOS isn’t on mobile (afaik), I would keep the description concise and simple with something like « A Jellyfin music client ». Up to you!

If you change it, also change the `genericName` attribute of `makeDesktopItem`.

nicolas-goudry (changes requested):

nicolas-goudry:
Result of `nixpkgs-review pr 315139` run on x86_64-linux [1](https://github.com/Mic92/nixpkgs-review)
<details>
  <summary>3 packages built:</summary>
  <ul>
    <li>finamp</li>
    <li>finamp.debug</li>
    <li>finamp.pubcache</li>
  </ul>
</details>

pkgs/by-name/fi/finamp/package.nix
DSeeLP: I did the 1. thing

pkgs/by-name/fi/finamp/package.nix
DSeeLP: I changed it and removed the call to `substituteInPlace`

pkgs/by-name/fi/finamp/package.nix
DSeeLP: I changed it to "An open source Jellyfin music player" and added it to the `makeDesktopItem` `comment` field and I changed the `genericName` field to `Music Player` like it is done upstream.

nicolas-goudry:
I was able to run the program and connect to my Jellyfin server. I don’t have any music on it so I couldn’t test the actual media playback.

DSeeLP:
Media playback works for me, but there is currently no built in volume control.

nicolas-goudry (approved):

maintainers/maintainer-list.nix
```diff
@@ -5532,6 +5532,11 @@
       fingerprint = "62BC E2BD 49DF ECC7 35C7  E153 875F 2BCF 163F 1B29";
     }];
   };
+  dseelp = {
+    name = "dsee";
+    github = "DSeeLP";
+    githubId = "46624152";
```
nicolas-goudry, line 5538: I didn’t catch that earlier, the githubId must be an integer.

nicolas-goudry (changes requested):

nicolas-goudry:
I was able to add some music to my Jellyfin server and can confirm the playback is working.

This may be related to the software being in beta, but there are a few graphical issues in the app.

maintainers/maintainer-list.nix
DSeeLP: Fixed

DSeeLP:
Result of `nixpkgs-review pr 315139` run on x86_64-linux [1](https://github.com/Mic92/nixpkgs-review)
<details>
  <summary>3 packages built:</summary>
  <ul>
    <li>finamp</li>
    <li>finamp.debug</li>
    <li>finamp.pubcache</li>
  </ul>
</details>

DSeeLP:
Updated to 0.9.8-beta which includes a few bug fixes

pkgs/by-name/fi/finamp/package.nix
```diff
@@ -0,0 +1,69 @@
...
+  desktopItem = makeDesktopItem {
+    name = "Finamp";
+    desktopName = "Finamp";
+    genericName = "Music Player";
+    exec = "finamp";
+    icon = "finamp";
+    startupWMClass = "finamp";
+    comment = "An open source Jellyfin music player";
+    categories = [
+      "Audio"
```
nicolas-goudry, line 55: Just noticed this while building:

```
/nix/store/5ddf158z5769lbw9biq97npk0sxfd061-Finamp.desktop/share/applications/Finamp.desktop: error: (will be fatal in the future): value item "Audio" in key "Categories" in group "Desktop Entry" requires another category to be present among the following categories: AudioVideo
```

Adding `AudioVideo` to the categories should fix it.

nicolas-goudry (changes requested):
Most likely not an issue but there is the following error when building:

```
Failed obtaining PRETTY_NAME for Linux: PathNotFoundException: Cannot open file, path = '/usr/lib/os-release' (OS Error: No such file or directory, errno = 2)
```

Don’t know why the build is using `/usr/lib/os-release`, I always use `/etc/os-release` (which exists on NixOS).

Would be great to patch this upstream or in the package! But not mandatory IMO, since the package builds and works fine.

pyrox0:
This depends on the `xdg-user-dir` tool, which I found out because my system configuration does not include this by default. Please add `xdg-user-dirs`(note the extra s on the end) to `buildInputs` so that the package runs properly if this is not installed seperately!

I've confirmed this fix works by getting a nix shell with this tool, and it works properly.

pkgs/by-name/fi/finamp/package.nix
DSeeLP: I added `AudioVideo` to the categories.
[Flutter first tries `/etc/os-release` and falls back to `/usr/lib/os-release` if it doesn't exist.](https://github.com/flutter/flutter/blob/5187cab7bdd434ca74abb45895d17e9fa553678a/packages/flutter_tools/lib/src/base/os.dart#L321)
I noticed that during build `/etc` only contains `group`, `hosts` and `passwd`.

DSeeLP:
I added `AudioVideo` to the categories and added `xdg-user-dirs` to the `buildInputs`.

pkgs/by-name/fi/finamp/package.nix
nicolas-goudry: Oh yes, that may be because builds are sandboxed.

nicolas-goudry (approved):

pyrox0:
LGTM!

pyrox0 (approved):

Titaniumtown (approved):
LGTM

pkgs/by-name/fi/finamp/package.nix
```diff
@@ -0,0 +1,71 @@
...
+  src = fetchFromGitHub {
+    owner = "jmshrv";
+    repo = "finamp";
+    rev = version;
+    hash = "sha256-lvjhA+hdCXgDsrNhNw4Tiq6ZgkYlPuMeHha8OJNF1TI=";
+  };
+  pubspecLock = lib.importJSON ./pubspec.lock.json;
+
+  nativeBuildInputs = [ patchelf ];
+  buildInputs = [ mpv xdg-user-dirs ];
```
Titaniumtown, line 24: `xdg-user-dirs` also needs to be a run-time dependency.

pkgs/by-name/fi/finamp/package.nix
```diff
@@ -0,0 +1,71 @@
...
+    categories = [
+      "AudioVideo"
+      "Audio"
+      "Player"
+      "Music"
+    ];
+  };
+
+  meta = with lib; {
+    description = "An open source Jellyfin music player";
```
Aleksanaa, line 64:
```suggestion
    description = "Open source Jellyfin music player";
```

pkgs/by-name/fi/finamp/package.nix
```diff
@@ -0,0 +1,71 @@
...
+    patchelf $out/app/finamp --add-needed libisar.so --add-needed libmpv.so --add-rpath ${lib.makeLibraryPath [ mpv ]}
+  '';
+
+  postInstall = ''
+    d=$out/share/icons/hicolor/scalable/apps
+    mkdir -p $d
+    cp $src/assets/icon/icon_foreground.svg $d/finamp.svg
+
+    mkdir -p $out/share/applications
+    cp $desktopItem/share/applications/*.desktop $out/share/applications
```
Aleksanaa, line 44: Use `desktopItems` with `copyDesktopItems` hook instead.

pkgs/by-name/fi/finamp/package.nix
```diff
@@ -0,0 +1,71 @@
...
+    comment = "An open source Jellyfin music player";
+    categories = [
+      "AudioVideo"
+      "Audio"
+      "Player"
+      "Music"
+    ];
+  };
+
+  meta = with lib; {
```
Aleksanaa, line 63: Better to avoid `with lib;`

nicolas-goudry: That’s not the first time I see another reviewer say this. What is the rationale? At such a low level, it doesn’t introduce much hassle to understand the code. Would it be placed at a higher level I would agree with you, but I don’t understand why it would be an issue here.

Aleksanaa: See https://github.com/NixOS/nixpkgs/issues/208242

nicolas-goudry: Thanks for pointing to this issue. As I said earlier, I do understand and do agree that using `with` at a high level is clearly misleading and makes the code much harder to read, even more when multiple `with` are used.

However, here I can only agree with SuperSandro and eclairevoyant:

> The problem is not when with is used in meta where it is very local and only covering a few lines. I am completely against changing those treewide.

> The nix.dev antipatterns guide warns against using with at the top of a Nix file - a reasonable suggestion, but does not mention anything about using with in the very predictable/standardised spot of only in meta.

I believe the `with` usages on `meta` attributes are legit since they save a few keystrokes while keeping the code readable, even for beginners. IMO this should not be changed.

DSeeLP: I will change it to match the manual at https://nixos.org/manual/nixpkgs/unstable/#chap-meta

DSeeLP:
I changed the meta section to match the [nixpkgs manual](https://nixos.org/manual/nixpkgs/unstable/#chap-meta) and adjusted the package description.
I removed `xdg-user-dirs` from `buildInputs` and added it to `PATH` using `extraWrapProgramArgs` and also switched to using the `copyDesktopItems` hook instead of copying the desktop file in `postInstall`.

pkgs/by-name/fi/finamp/package.nix
DSeeLP: Fixed

pkgs/by-name/fi/finamp/package.nix
DSeeLP: Fixed

pkgs/by-name/fi/finamp/package.nix
DSeeLP: I added it to `PATH` using `extraWrapProgramArgs` and removed it from `buildInputs`.

DSeeLP:
3 packages built:
finamp finamp.debug finamp.pubcache

Result of `nixpkgs-review pr 315139` run on x86_64-linux [1](https://github.com/Mic92/nixpkgs-review)
<details>
  <summary>3 packages built:</summary>
  <ul>
    <li>finamp</li>
    <li>finamp.debug</li>
    <li>finamp.pubcache</li>
  </ul>
</details>

pkgs/by-name/fi/finamp/package.nix
Titaniumtown: Ok great. Works perfectly now @DSeeLP 
Good work!

Guanran928 (approved):
Works on `x86_64-linux`.

Saw some flickering, but it disappeared later.

https://github.com/NixOS/nixpkgs/assets/68757440/734fa766-f7c1-4448-b455-7c8e2ac721be

pkgs/by-name/fi/finamp/package.nix
```diff
@@ -0,0 +1,73 @@
...
+, xdg-user-dirs
+, patchelf
+, fetchFromGitHub
+, copyDesktopItems
+, makeDesktopItem
+}:
+let
+  version = "0.9.8-beta";
+in
+flutter.buildFlutterApplication {
```
Aleksanaa, line 13: The flutter version should pin to upstream version. See https://github.com/NixOS/nixpkgs/blob/master/doc/languages-frameworks/dart.section.md#flutter-applications-ssec-dart-flutter.

```suggestion
flutter322.buildFlutterApplication {
```

You can find the version in it's workflow run log (of the `0.9.8-beta` commit): https://github.com/jmshrv/finamp/actions/runs/9504242145/job/26196438043#step:5:38

pkgs/by-name/fi/finamp/package.nix
```diff
@@ -0,0 +1,73 @@
+{ lib
+, flutter
+, mpv
```
Aleksanaa, line 3:
```suggestion
, mpv-unwrapped
```

Not so sure, but usually we don't use wrapped ones if we need the library

pkgs/by-name/fi/finamp/package.nix
```diff
@@ -0,0 +1,73 @@
...
+  };
+
+  postFixup = ''
+    patchelf $out/app/finamp --add-needed libisar.so --add-needed libmpv.so --add-rpath ${lib.makeLibraryPath [ mpv ]}
+  '';
+
+  postInstall = ''
+    d=$out/share/icons/hicolor/scalable/apps
+    mkdir -p $d
+    cp $src/assets/icon/icon_foreground.svg $d/finamp.svg
```
Aleksanaa, line 42:
```suggestion
    install -Dm644 $src/assets/icon/icon_foreground.svg $d/finamp.svg
```

No need to `mkdir`.

DSeeLP:
I applied all suggestions from @Aleksanaa to use `install` instead of `mkdir` and `cp` afterwards and to use the unwrapped of mpv version and to specify the flutter release.
@Guanran928 the flickering also happens to me. It also happens on Ubuntu 24.04 with the release binaries. I reported it https://github.com/jmshrv/finamp/issues/801.

pkgs/by-name/fi/finamp/package.nix
DSeeLP: Fixed

pkgs/by-name/fi/finamp/package.nix
DSeeLP: Fixed

pkgs/by-name/fi/finamp/package.nix
DSeeLP: Fixed

DSeeLP:
3 packages built:
finamp finamp.debug finamp.pubcache

Result of `nixpkgs-review pr 315139` run on x86_64-linux [1](https://github.com/Mic92/nixpkgs-review)
<details>
  <summary>3 packages built:</summary>
  <ul>
    <li>finamp</li>
    <li>finamp.debug</li>
    <li>finamp.pubcache</li>
  </ul>
</details>

Aleksanaa (approved):

PR #244332, opened 2023-07-19
nixos/fonts: rename fonts.fonts option to fonts.packages, other cleanups 

SuperSandro2000:
###### Description of changes

fonts.fonts is not a great name and this also resolves a TODO.

###### Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
- [ ] For non-Linux: Is `sandbox = true` set in `nix.conf`? (See [Nix manual](https://nixos.org/manual/nix/stable/command-ref/conf-file.html))
- [ ] Tested, as applicable:
  - [NixOS test(s)](https://nixos.org/manual/nixos/unstable/index.html#sec-nixos-tests) (look inside [nixos/tests](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests))
  - and/or [package tests](https://nixos.org/manual/nixpkgs/unstable/#sec-package-tests)
  - or, for functions and "core" functionality, tests in [lib/tests](https://github.com/NixOS/nixpkgs/blob/master/lib/tests) or [pkgs/test](https://github.com/NixOS/nixpkgs/blob/master/pkgs/test)
  - made sure NixOS tests are [linked](https://nixos.org/manual/nixpkgs/unstable/#ssec-nixos-tests-linking) to the relevant packages
- [ ] Tested compilation of all packages that depend on this change using `nix-shell -p nixpkgs-review --run "nixpkgs-review rev HEAD"`. Note: all changes have to be committed, also see [nixpkgs-review usage](https://github.com/Mic92/nixpkgs-review#usage)
- [ ] Tested basic functionality of all binary files (usually in `./result/bin/`)
- [23.11 Release Notes](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2311.section.md) (or backporting [23.05 Release notes](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2305.section.md))
  - [ ] (Package updates) Added a release notes entry if the change is major or breaking
  - [ ] (Module updates) Added a release notes entry if the change is significant
  - [ ] (Module addition) Added a release notes entry if adding a new NixOS module
- [ ] Fits [CONTRIBUTING.md](https://github.com/NixOS/nixpkgs/blob/master/CONTRIBUTING.md).

PR #315149, opened 2024-05-27
nixos/steam: fix evaluation failure when `fonts.packages` contains path

Cryolitia:
## Description of changes

Fix https://github.com/NixOS/nixpkgs/pull/312268#issuecomment-2133641945

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/nixos/modules/programs/steam.nix
+++ b/nixos/modules/programs/steam.nix
@@ -104,8 +104,9 @@ in {
 
     fontPackages = lib.mkOption {
       type = lib.types.listOf lib.types.package;
-      default = config.fonts.packages;
-      defaultText = lib.literalExpression "fonts.packages";
+      # `fonts.packages` is a list of paths now, filter out which are not packages
+      default = builtins.filter lib.types.package.check config.fonts.packages;
+      defaultText = lib.literalExpression "builtins.filter lib.types.package.check config.fonts.packages";
       example = lib.literalExpression "with pkgs; [ source-han-sans ]";
       description = ''
         Font packages to use in Steam.


nixos/modules/programs/steam.nix
```diff
@@ -191,7 +191,10 @@ in {
       };
     };
 
-    programs.steam.extraPackages = cfg.fontPackages;
+    programs.steam.extraPackages = [ (pkgs.runCommand "X11-fonts" { preferLocalBuild = true; } ''
+      mkdir -p "$out/share"
+      ln -s "${pkgs.symlinkJoin { name = "steam-fonts"; paths = cfg.fontPackages; }}/share/fonts" "$out/share/fonts"
+    '')];
```
jian-lin, line 197: It is enough to just change the type of `extraPackages` and `fontPackages` to `types.listOf types.path`.  After skimming through codes related to steam, buildFHSEnv and [buildEnv][1], I find that the `package` type is stronger than needed and the `path` type can be handled.

[1]: https://github.com/NixOS/nixpkgs/blob/2cd4abd34d2eddacb62f60a4c157a0a0c8d1817d/pkgs/build-support/buildenv/default.nix#L58-L72

jian-lin (commented):
I propose another fix.

nixos/modules/programs/steam.nix
jian-lin: The disadvantage of this proposal is that if an element of `extraPackages` or `fontPackages` is not a `package` (derivation) it will be linked to a wrong place so it won't be available in the steam environment.

Cryolitia: What I'm more curious about is whether there are other practical use cases where fonts.packages is a path. For `ghostscript` this is irrelevant, since `/usr/share/ghostscript/fonts` is not in steam's search path.

jian-lin:
According to [1][1], `types.listOf types.package` is the semantically correct type for `programs.steam.extraPackages` and `programs.steam.fontPackages`.

Better fixes than my last proposal are to filter out non-package elements of `fonts.packages` or wrap them in packages.  I think the former is good enough to be a default value and keeps the module simple.

Another direction is to change the type of `fonts.packages` to `types.listOf types.package`.  This is a breaking change and needs more effect.

[1]: https://github.com/NixOS/nixpkgs/pull/315149#discussion_r1616315129

nixos/modules/programs/steam.nix
jian-lin: BTW, just note that your original implementation also has this disadvantage.

Cryolitia:
1. https://github.com/NixOS/nixpkgs/pull/244332 changes it from `fonts.fonts` to `fonts.packages` , which broken the semantics that it's not packages at all.
2. It is mainly used as an rvalue in https://github.com/NixOS/nixpkgs/blob/fb460e0ba0054cd73258bddb2af9e6e6d111bee6/nixos/modules/config/fonts/fontconfig.nix#L54 , which do actually accept a path whatever it contains.

> Better fixes than my last proposal are to filter out non-package elements of fonts.packages or wrap them in packages. I think the former is good enough to be a default value and keeps the module simple.

I don't know whether there is user really add a path that contains CJK fonts to `fonts.packages`, and expecting that Steam to work out-of-box.

> Another direction is to change the type of fonts.packages to types.listOf types.package. This is a breaking change and needs more effect.

I'd better to rename it back to sth. like `fonts.directories` with type `listOf path`

Whatever, the most important thing now is to fix the broken evalution, then we could discuss how to solve the semantic inconsistency problem.

I would prefer to change `steam.*Packages` 's type to `listOf path`, or symlinkJoin all pathes in `fonts.packages` into a package.

jian-lin:
> I don't know whether there is user really add a path that contains CJK fonts to fonts.packages and expecting that Steam to work out-of-box.

> I would prefer to ... symlinkJoin all pathes in fonts.packages into a package

IMHO, this use case is not common and  is not worth supporting.  I doubt they cannot specify those fonts as packages.  More like a package issue of those fonts.



> Whatever, the most important thing now is to fix the broken evalution

Affected users can change/set programs.steam.fontPackages as a local workaround.

---


I still think changing the type of fonts.packages to `package` is a good idea.  A bit stronger type leads to much simpler steam module code.

Cryolitia:
> Better fixes than my last proposal are to filter out non-package elements of fonts.packages or wrap them in packages. I think the former is good enough to be a default value and keeps the module simple.

done

nixos/modules/programs/steam.nix
```diff
@@ -104,8 +104,9 @@ in {
 
     fontPackages = lib.mkOption {
       type = lib.types.listOf lib.types.package;
-      default = config.fonts.packages;
-      defaultText = lib.literalExpression "fonts.packages";
+      # `fonts.packages` is a list of paths now, filter out which are not packages
+      default = builtins.filter (p: lib.types.package.check p) config.fonts.packages;
+      defaultText = lib.literalExpression "builtins.filter (p: lib.types.package.check p) fonts.packages";
```
jian-lin, line 109: nit
```suggestion
      defaultText = lib.literalExpression "builtins.filter (p: lib.types.package.check p) config.fonts.packages";
```

jian-lin (commented):
LGTM

commit message can be improved: `fix evalute failed` -> `fix evaluation failure`

jian-lin (approved):
I think it is a good idea to backport this fix and the original PR to 24.05.

nixos/modules/programs/steam.nix
```diff
@@ -104,8 +104,9 @@ in {
 
     fontPackages = lib.mkOption {
       type = lib.types.listOf lib.types.package;
-      default = config.fonts.packages;
-      defaultText = lib.literalExpression "fonts.packages";
+      # `fonts.packages` is a list of paths now, filter out which are not packages
+      default = builtins.filter (p: lib.types.package.check p) config.fonts.packages;
```
eclairevoyant, line 108: nit: eta reduction
```suggestion
      default = builtins.filter lib.types.package.check config.fonts.packages;
```

eclairevoyant (approved):
This solution looks good. Thanks for the fix.

linyinfeng:
I think `fonts.packages` is not a good name, but its type `list path` is reasonable since these paths are mainly used as `<dir>FONT_PATH</dir>` in a configuration file of Fontconfig. Fontconfig recursively searches these paths for font files. So `fonts.packages` should be a list of paths of directories containing font files (no matter in which subdirectories).

The current fix and the original PR <https://github.com/NixOS/nixpkgs/pull/312268#issuecomment-2133641945> both assumed that, the packages in `fonts.packages` always install fonts into `$out/share/fonts`. I think this is not a good assumption. Any directory 

But, since all font packages in nixpkgs do install fonts into `$out/share/fonts`, I think it is OK to merge this as a fix.

A more correct method might be:
1. Change the type of `programs.steam.fontPackages` to `listOf path`, aligning to `fonts.packages`.
2. `let` a new package `steamFonts`, which aggregates all paths in `fontPackages` and installs to `$out/share/fonts` (like `x11Fonts` in <https://github.com/nixos/nixpkgs/blob/master/nixos/modules/config/fonts/fontdir.nix#L9-L21>).
4. `programs.steam.extraPackages = [ steamFonts ];`
5. (optional) Rename `fonts.packages` to some other name like `fonts.directories`, and rename `programs.steam.fontPackages` to `programs.steam.fontDirectories` in a separate PR.

jian-lin:
> > I still think changing the type of fonts.packages to package is a good idea. A bit stronger type leads to much simpler steam module code.
> 
> I think `fonts.packages` is not a good name, but its type `list path` is reasonable since these paths are mainly used as `<dir>FONT_PATH</dir>` in a configuration file of Fontconfig. Fontconfig recursively searches these paths for font files. So `fonts.packages` should be a list of paths of directories containing font files (no matter in which subdirectories). `list path` provides flexibility, so that user can easily add some non-standard font packages to their environment, for example, `ghostscript`'s fonts and TeXLive's font packages.
> 

My assumption is that adding "non-standard" fonts to `fonts.packages` is a niche use case.  If it is not true, then we should not change the type of `fonts.packages`.

> 
> A more correct method might be:
> 
> 1. Change the type of `programs.steam.fontPackages` to `listOf path`, aligning to `fonts.packages`.
> 2. `let` a new package `steamFonts`, which aggregates all paths in `fontPackages` and installs to `$out/share/fonts` (like `x11Fonts` in https://github.com/nixos/nixpkgs/blob/master/nixos/modules/config/fonts/fontdir.nix#L9-L21).
> 3. `programs.steam.extraPackages = [ steamFonts ];`
> 4. (optional) Rename `fonts.packages` to some other name like `fonts.directories`, and rename `programs.steam.fontPackages` to `programs.steam.fontDirectories` in a separate PR.
> 5. (further) find a way to let Steam respect system Fontconfig configurations.

I agree that your proposal is more correct.

However, given that using "non-standard" fonts in steam is probably a niche use case, the current implementation of this PR reaches a better balance between correctness (or use case coverage) and module complexity.  Just my two cents.

linyinfeng:
> A more correct method might be:
> 
> 1. Change the type of `programs.steam.fontPackages` to `listOf path`, aligning to `fonts.packages`.
> 2. `let` a new package `steamFonts`, which aggregates all paths in `fontPackages` and installs to `$out/share/fonts` (like `x11Fonts` in https://github.com/nixos/nixpkgs/blob/master/nixos/modules/config/fonts/fontdir.nix#L9-L21).
> 3. `programs.steam.extraPackages = [ steamFonts ];`
> 4. (optional) Rename `fonts.packages` to some other name like `fonts.directories`, and rename `programs.steam.fontPackages` to `programs.steam.fontDirectories` in a separate PR.
> 5. (further) find a way to let Steam respect system Fontconfig configurations.

I tested 1-3 on my machine with `nixos-unstable` and it works fine.
<https://github.com/linyinfeng/dotfiles/blob/main/nixos/modules/programs/steam/fonts.nix>

It is pretty simple.

```nix
let
  ...
  steamFontsDir = pkgs.symlinkJoin {
    name = "steam-fonts-dir";
    paths = cfg.fontPackages;
  };
  steamFonts = pkgs.runCommand "steam-fonts" { preferLocalBuild = true; } ''
    mkdir -p "$out/share"
    ln -s "${steamFontsDir}" "$out/share/fonts"
  '';
in
{
  options.programs.steam = {
    ...
    fontPackages = lib.mkOption {
      type = lib.types.listOf lib.types.path;
      ...
    };
    ...
  };
  ...
  config = {
    ...
    programs.steam.extraPackages = [ steamFonts ];
    ...
  };
}
```

The only problem is that the resulting "/usr/share/fonts" is somehow messy, but Fontconfig does not care and the fonts work in-game.

```
$ steam-run eza /usr/share/fonts --tree --level 3
/usr/share/fonts
├── lib
│  └── X11
│     └── fonts
└── share
   ├── doc
   │  └── liberation-fonts-2.1.5
   ├── fonts
   │  ├── cantarell
   │  ├── fonts.dir
   │  ├── fonts.scale
   │  ├── noto
   │  ├── opentype
   │  ├── truetype
   │  ├── unifont.otb
   │  ├── unifont.pcf.gz
   │  ├── variable
   │  ├── wqy-microhei.ttc
   │  └── wqy-zenhei.ttc
   └── metainfo
      └── org.gnome.cantarell.metainfo.xml
```

image

(RimWorld in Simplified Chinese)

linyinfeng:
> However, given that using "non-standard" fonts in steam is probably a niche use case, the current implementation of this PR reaches a better balance between correctness (or use case coverage) and module complexity. Just my two cents.

I agree, that using "non-standard" fonts is a niche use case. But to support them is not very complex. As you can see in <https://github.com/NixOS/nixpkgs/pull/315149#issuecomment-2134374297>, it only adds six lines of code to create `steamFonts`:

```nix
  steamFontsDir = pkgs.symlinkJoin {
    name = "steam-fonts-dir";
    paths = cfg.fontPackages;
  };
  steamFonts = pkgs.runCommand "steam-fonts" { preferLocalBuild = true; } ''
    mkdir -p "$out/share"
    ln -s "${steamFontsDir}" "$out/share/fonts"
  '';
```

Cryolitia:
I had written it earlier: https://github.com/NixOS/nixpkgs/pull/315149#discussion_r1616297246

But if we do that, should we change it to `steam.fontDirectories`?

jian-lin:
> I had written it earlier: [#315149 (comment)](https://github.com/NixOS/nixpkgs/pull/315149#discussion_r1616297246)
> 

That implementation is not the same as @linyinfeng's duo to the trailing `/share/fonts` in `paths = cfg.fontPackages; }}/share/fonts`.

linyinfeng:
> I had written it earlier: [#315149 (comment)](https://github.com/NixOS/nixpkgs/pull/315149#discussion_r1616297246)

Sorry, I missed this conversation since it has been resolved.

> But if we do that, should we change it to `steam.fontDirectories`?

I'm not sure. If we use `listOf path`, my original idea is to rename `fonts.packages` and `programs.steam.fontPackages` together, keeping the name of these two options consistent. But I think changing one of these options first is totally OK.

> 5. (optional) Rename `fonts.packages` to some other name like `fonts.directories`, and rename `programs.steam.fontPackages` to `programs.steam.fontDirectories` in a separate PR.

But if we do this. Since the option will reach `nixos-unstable` after <https://hydra.nixos.org/eval/1806566> passes. If we use `listOf path`, change the option name to `fontDirectories`, `mkRenamedOptionModule` might be needed? I'm not sure about that.

Cryolitia:
> `ln -s "${steamFontsDir}" "$out/share/fonts"`

I don't like this.

I think the most basic consensus we can reach is that the directory structure within the software package should be consistent with the root directory structure of other FHS distributions. That is, only the /share/fonts directory in the package should be installed to `/usr/share/fonts` in other FHS distributions, also should we build FHS environment as this.

Then, the problem is that, if their's other packages, for example, a tex font chosen to install it in `/usr/share/tex/fonts`, and Steam only read fonts from the FHS's `/usr/share/fonts`, the tex font will take no effect.

It is the Steam that they choose to only read `/usr/share/fonts` but not others like `/usr/share/tex/fonts`, it's by design but not an issue that we should fix. Although I don't know the reason, but I don't think it's a good idea to force Steam to read fonts from where the Steam don't expect. If their's any user really need to do this, they could re-pack it to make it follows the directory structure of FHS.

An obvious problem with forcing all fonts to be mounted to `/usr/share/fonts` is that since we have passed the system fonts by default, if a user turns on `fonts.enableGhostscriptFonts`, these fonts will also be forced to be mounted. What will happen is unpredictable. We already know that Steam will not respect system-level fontconfig configurations, and many users do not write their own user-level fontconfig configurations. Steam will follow rules in this case that we do not yet know. Selecting fonts also includes fonts that Steam originally did not want to select. If we think Steam should search for fonts everywhere anyway, this should be an upstream issue.

Cryolitia:
A more compromise solution would be to add a `programs.steam.fonts.forceSearchAnywhere` option, but I don't think this is really necessary.

linyinfeng:
> But, since all font packages in nixpkgs do install fonts into `$out/share/fonts`, I think it is OK to merge this as a fix.

> I don't like this.

If you believe the current fix does not need to be changed, I will not block this from merging. We can merge this and ignore my previous comments. Also, there is already another reason to merge this as is.

> adding "non-standard" fonts to `fonts.packages` is a niche use case.
> However, given that using "non-standard" fonts in steam is probably a niche use case, the current implementation of this PR reaches a better balance between correctness (or use case coverage) and module complexity.

By the way,

> Steam will follow rules in this case that we do not yet know.

I extracted the Fontconfig configuration files from "Steam Linux Runtime 3.0 (sniper)". The configuration properly defines aliases for `serif`, `sans`, and `monospace`. So the result of adding some random fonts to its search path should be pretty predictable, nothing will change.

`steam-run ~/.local/share/Steam/steamapps/common/SteamLinuxRuntime_sniper/run-in-sniper -- tar -czvf /tmp/fonts.tar.gz /usr/share/fontconfig /etc/fonts`

[fonts.tar.gz](https://github.com/NixOS/nixpkgs/files/15474121/fonts.tar.gz)

Aleksanaa (approved):
Eh okay, though not great, still a good solution suitable for backport

jian-lin:
Let's merge this.  We can iterate later if the non-standard font use case is not that uncommon.

Issue #312971, opened 2024-05-19
LunarVim: `nvim-treesitter` error


cig0:
### Describe the bug

About a week ago, I started seeing this error message at LV launch:

```
Failed to source `/home/cig0/.local/share/lvim/site/pack/lazy/opt/nvim-treesitter/plugin/nvim-treesitter.lua`
                                                                                                                                                                                                           
vim/_editor.lua:0: BufReadPre Autocommands for "*"..script nvim_exec2() called at BufReadPre Autocommands for "*":0../home/cig0/.local/share/lvim/site/pack/lazy/opt/bigfile.nvim/after/plugin/bigfile.lua.
.nvim_exec2() called at /home/cig0/.local/share/lvim/site/pack/lazy/opt/bigfile.nvim/after/plugin/bigfile.lua:0../home/cig0/.local/share/lvim/site/pack/lazy/opt/nvim-treesitter/plugin/nvim-treesitter.lua
: Vim(source):E5113: Error while calling lua chunk: ...d-0.10.0/share/nvim/runtime/lua/vim/treesitter/query.lua:651: Overriding existing predicate has-ancestor?                                           
stack traceback:                                                                                                                                                                                           
^I[C]: in function 'error'                                                                                                                                                                                 
^I...d-0.10.0/share/nvim/runtime/lua/vim/treesitter/query.lua:651: in function 'add_predicate'                                                                                                             
^I...nvim-treesitter/lua/nvim-treesitter/query_predicates.lua:74: in main chunk                                                                                                                            
^I[C]: in function 'require'                                                                                                                                                                               
^I...te/pack/lazy/opt/nvim-treesitter/lua/nvim-treesitter.lua:8: in main chunk                                                                                                                             
^I[C]: in function 'require'                                                                                                                                                                               
^I...pack/lazy/opt/nvim-treesitter/plugin/nvim-treesitter.lua:9: in main chunk                                                                                                                             
^I[C]: in function 'nvim_exec2'                                                                                                                                                                            
^Ivim/_editor.lua: in function 'cmd'                                                                                                                                                                       
^I...im/site/pack/lazy/opt/lazy.nvim/lua/lazy/core/loader.lua:438: in function <...im/site/pack/lazy/opt/lazy.nvim/lua/lazy/core/loader.lua:437>                                                           
^I[C]: in function 'xpcall'                                                                                                                                                                                
^I...                                                                                                                                                                                                      
[...]

PR #315186, opened 2024-05-27
lunarvim: pinning neovim to 0.9.2 for compatibility

lebensterben:
## Description of changes

This commit fixes the dependency package `neovim` to `0.9.2`.

Though `lunarvim` version `1.4.0` supports `neovim` `v0.9.0+` as said in
its [documentation], it turns out that only `v0.9.0` and `v0.9.2` could
run properly without error.

[documentation]: <https://www.lunarvim.org/docs/installation#prerequisites>

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---


lebensterben:
Here are somethings for consideration before merging this.

# What is the commit trying to do?

As reported in https://github.com/NixOS/nixpkgs/issues/312971, `lunarvim` stable releases (currently at `1.4.0`) won't run correctly with latest `neovim`.

As suggested by the upstream developer from `lunarvim`, there are two solutions:
1. Using develop branch of `lunarvim` which is **preferred** if we use `v0.10.0+` version of `neovim`.
2. Pinning `neovim` version to `v0.9.5` and keep using release branch of `lunarvim`.

The problem of the first approach is that there might not be any guarantee that the latest `lunarvim` works with the latest `neovim`. And it in turn increases the maintenance burden should the in-compatibility occurs.

Therefore I opted for the second approach here, though it's worth noting that after testing `lunarvim` only works without error with `neovim` `v0.9.0` or `v0.9.2`.

# Why I pull in the tarball of a specific version of `nixos/nixpkgs`?

To pin the version of a dependency package, one can either use a overlay or override version and appropriate attributes of the dependency.

The first approach is discouraged by [nixpkgs manual](https://nixos.org/manual/nixpkgs/unstable/#sec-overlays-argument):

> NOTE: DO NOT USE THIS in nixpkgs.

and meanwhile the second approach hits the error

```txt
       error: function 'legacyWrapper' called with unexpected argument 'neovim-unwrapped'
```

when I tried

```nix
neovim-pinned = neovim.override (neovim-previous: {
  neovim-unwrapped = neovim-previous.neovim-unwrapped.override (neovim-unwrapped-previous: {
    version = "0.9.5";
  });
});
```

whereas `legacyWrapper` is defined [here](https://github.com/NixOS/nixpkgs/blob/bfb7a882678e518398ce9a31a881538679f6f092/pkgs/applications/editors/neovim/utils.nix#L127-L141) and it limits the attributes I can use override for the underlying `neovim-unwrapped` package.

Therefore I have to manually pull in a specific version of `nixos/nixpkgs` tarball that contains `neovim` `v0.9.2`.

teto:
The fix here is not possible to merge, it doesn't scale. If every package referenced a previous versions of nixpkgs, you end up with 80k different evaluations. Either mark lunarvim as broken or use an unstable version (justified in this case IMO). 
The error referenced in the issue seemed related to an nvim-treesitter incompatibility. Maybe you just need to change that ?

lebensterben:
I haven't used lunarvim a lot recently. And the prior statement that these two versions are compatible could be false if I had tried with more programming languages. I will close this PR.

As for whether marking it as broken:

The problem is it works with some languages but not others. I'm inclined to mark it as broken but I prefer to get more inputs from other users.

PR #315205, opened 2024-05-27
tracy: fix wayland - remove legacy build, add deps

### LINKED ISSUE #308220: Runtime failure: tracy
### Describe the bug

```
Error 65544: Wayland: Failed to load libwayland-client
terminate called without an active exception
zsh: abort (core dumped)  result/bin/tracy
```

### Steps To Reproduce
1. Run `tracy`

### Expected behavior
No crash

### Screenshots
n/a

### Additional context
n/a

### Notify maintainers
@mpickering @nagisa @paveloom

### Metadata

```console
output here
```

---

widlarizer:
## Description of changes

This fixes #308220 on my end, but consider this a draft PR. I am very sleepy rn and don't really know what I'm doing. Tagging @paveloom @mpickering 

Tasks before merging
- [x] To be checked whether this breaks Tracy for X11 users (I guess it does) 
- [x] Try reverting to capstone from capstone_4 (if it works it's probably left as it was before) 

## Things done

 I remove `LEGACY=1`. I add wayland, libxkbcommon as deps and replace capstone with capstone_4 since the current default.nix explicitly references capstone 4.

- Built on platform(s)
  - [X] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:
- [ X Fits [CONTRIBUTING.md](https://github.com/NixOS/nixpkgs/blob/master/CONTRIBUTING.md).


Aleksanaa (approved):
Looks good but if some x11 users see this pls give some feedback

pkgs/development/tools/tracy/default.nix
```diff
@@ -2,11 +2,13 @@
...
+, capstone_4
 , darwin
 , dbus
 , freetype
 , glfw
+, wayland
+, libxkbcommon
 , hicolor-icon-theme
 , pkg-config
 , tbb
```
paveloom, line 14:
```suggestion
, capstone_4
, darwin
, dbus
, freetype
, glfw
, hicolor-icon-theme
, libxkbcommon
, pkg-config
, tbb
, wayland
```

pkgs/development/tools/tracy/default.nix
```diff
@@ -30,9 +32,11 @@ stdenv.mkDerivation rec {
   nativeBuildInputs = [ pkg-config ];
 
   buildInputs = [
-    capstone
+    capstone_4
     freetype
     glfw
+    wayland
+    libxkbcommon
```
paveloom, line 39:
```suggestion
    capstone_4
    freetype
    glfw
    libxkbcommon
    wayland
```

paveloom (changes requested):

paveloom:
It *will* break the app for X11 users since the support for it has been [dropped](https://github.com/wolfpld/tracy/blob/37aff70dfa50cf6307b3fee6074d627dc2929143/NEWS#L68-L75) since version `0.9.1`.

I'm fine with it now, but, uhm... yeah... Also, see https://github.com/wolfpld/tracy/issues/505.

widlarizer:
If X11 requires different deps, how about we split the package between tracy and tracy-x11? I'm sure there's more than zero happy X11 tracy users out there and I don't want to fight them over the steering wheel on who gets to use this program.

paveloom:
Yep, I also think this is the way to approach this.

I would make it somewhat similar to how it's done in the derivation for [`rio`](https://github.com/NixOS/nixpkgs/blob/42dd9b11a64791f15bc0a9316764bf2911fb906a/pkgs/applications/terminal-emulators/rio/default.nix):

1) A separate block of alphabetically sorted inputs with `, withWayland ? true` at the start.
2) Optionally a separate block of `buildInputs` with `lib.optionals withWayland`.
3) A separate derivation in `all-packages.nix` called `tracy-x11` with `withWayland` set to `false`.

widlarizer:
I have implemented what you described, and tested with xeyes that tracy-x11 is an X program and tracy is a wayland program

widlarizer:
No idea what that test failure is supposed to be

pkgs/by-name/tr/tracy/package.nix
```diff
@@ -60,7 +66,7 @@ stdenv.mkDerivation rec {
     make -j $NIX_BUILD_CORES -C csvexport/build/unix release
     make -j $NIX_BUILD_CORES -C import-chrome/build/unix release
     make -j $NIX_BUILD_CORES -C library/unix release
-    make -j $NIX_BUILD_CORES -C profiler/build/unix release LEGACY=1
+    make -j $NIX_BUILD_CORES -C profiler/build/unix release ${if withWayland then "" else "LEGACY=1"}
```
paveloom, line 69: This should only affect Linux builds:

```suggestion
    make -j $NIX_BUILD_CORES -C profiler/build/unix release \
      ${lib.optionalString (stdenv.isLinux && !withWayland) "LEGACY=1"}
```

pkgs/top-level/all-packages.nix
```diff
@@ -7584,7 +7584,7 @@ with pkgs;
 
   tracker = callPackage ../development/libraries/tracker { };
 
-  tracy = callPackage ../development/tools/tracy { };
+  tracy-x11 = callPackage ../by-name/tr/tracy/package.nix { withWayland = false; };
```
paveloom, line 7587: The `LEGACY` flag is about choosing between GLFW and Wayland backends on Linux systems. Meaning, as of right now, the legacy build works both on X11 and Wayland, so the use of the flag is not really an issue of X11 vs. Wayland. In the mean time, GLFW is still the default on all other platforms (including Darwin, which we support).

I agreed with the creation of the `x11` package before, but now I think that we probably shouldn't add the legacy build to the binary cache. This also leads us to having two identical packages on Darwin systems (since the `LEGACY` flag is ignored there). So, this line should be deleted, and the release note should mention the use of the flag instead.

paveloom (changes requested):

pkgs/top-level/all-packages.nix
flokli: > I agreed with the creation of the x11 package before, but now I think that we probably shouldn't add the legacy build to the binary cache

I think it's valid to have both in there. We have both X11 and Wayland users, and shouldn't make it unnecessarily painful for X11 users. We build so much other less useful stuff, I think it's fine to build both flavours.

flokli (approved):
Tested this on Wayland, and the `tracy-x11` attribute still has the same store path as before - so there's not really any way this could have regressed X11 users (other than having to switch to the other attribute, which is clearly documented). I don't think we need to keep this broken for longer ;-)

@widlarizer can you address the cosmetical changes, or should this be done in a followup PR?

pkgs/top-level/all-packages.nix
paveloom: As I described above, this is more of a GLFW vs. custom Wayland backend issue, rather than an X11 vs. Wayland issue. On Darwin, two identical packages will be cached with the current set of changes, and the `-x11` postfix doesn't really mean anything on that platform.

I wouldn't say it's painful to override one attribute if the legacy build is needed, but... If @widlarizer still wants to create a separate package, then he needs to make conditional changes to `pname` (different name is required to cache a different derivation) and `meta.platforms`.

Theoretically, we would have the `tracy-wayland` package for Linux only and `tracy-glfw` for both Linux and Darwin (which is the way it is packaged now). Then, one of those would conditionally be chosen as `tracy`.

flokli: I'm not saying your assessment is wrong, I'm merely proposing to have that being dealt with in a followup (or as an additional commit to this PR).

The PR fixes a crash at runtime, and already did a lot of rounds. The changes requested again might not be respectful to the PR author, who was merely sending a fix for a crash while being "very sleepy rn and don't really know what [...] doing".

We can always juggle things around a bit in unstable, especially editing release notes or removing "redundant packages".

widlarizer: I like the proposed solution (the one starting with "Theoretically,"). Being an X11 user on Linux is still valid and on NixOS shouldn't require overrides. Only problem is, looking at all-packages.nix for inspiration, I don't see a way of having one or two flavors of a package depending on the system. I'll leave that bit to a potential followup by @paveloom

widlarizer:
Cosmetics addressed, feel free to merge if CI goes green

paveloom (approved):

pca006132:
can we backport this to 24.05?

flokli:
> can we backport this to 24.05?

I don't think so, at least not replacing the `legacy` build. You could introduce a `tracy-wayland` there if you really want to, or wrap it with xwayland for the time being.

PR #315251, opened 2024-05-28
jemalloc: fix linking against libc++ with lld

### LINKED ISSUE #315247: Build failure: pkgsLLVM.jemalloc
### Steps To Reproduce

1. build `pkgsLLVM.jemalloc`

### Build log

```
>>> referenced by lib/libjemalloc.so
make: *** [Makefile:534: test/integration/xallocx] Error 1
make: *** [Makefile:534: test/integration/slab_sizes] Error 1
aarch64-unknown-linux-gnu-ld: error: undefined reference due to --no-allow-shlib-undefined: std::__1::mutex::lock()
>>> referenced by lib/libjemalloc.so

aarch64-unknown-linux-gnu-ld: error: undefined reference due to --no-allow-shlib-undefined: std::set_new_handler(void (*)())
>>> referenced by lib/libjemalloc.so

aarch64-unknown-linux-gnu-ld: error: undefined reference due to --no-allow-shlib-undefined: std::__1::mutex::unlock()
>>> referenced by lib/libjemalloc.so

aarch64-unknown-linux-gnu-ld: error: undefined reference due to --no-allow-shlib-undefined: __cxa_begin_catch
>>> referenced by lib/libjemalloc.so

aarch64-unknown-linux-gnu-ld: error: undefined reference due to --no-allow-shlib-undefined: __cxa_end_catch
>>> referenced by lib/libjemalloc.so

aarch64-unknown-linux-gnu-ld: error: undefined reference due to --no-allow-shlib-undefined: std::__throw_bad_alloc()
>>> referenced by lib/libjemalloc.so

aarch64-unknown-linux-gnu-ld: error: undefined reference due to --no-allow-shlib-undefined: std::terminate()
>>> referenced by lib/libjemalloc.so

aarch64-unknown-linux-gnu-ld: error: undefined reference due to --no-allow-shlib-undefined: typeinfo for std::bad_alloc
>>> referenced by lib/libjemalloc.so

aarch64-unknown-linux-gnu-ld: error: undefined reference due to --no-allow-shlib-undefined: __gxx_personality_v0
>>> referenced by lib/libjemalloc.so
make: *** [Makefile:534: test/integration/sdallocx] Error 1
make: *** [Makefile:534: test/integration/thread_arena] Error 1
make: *** [Makefile:534: test/integration/thread_tcache_enabled] Error 1
make: *** [Makefile:534: test/integration/overflow] Error 1
make: *** [Makefile:534: test/integration/rallocx] Error 1
```

### Notify maintainers

@yu-re-ka 

### Metadata

```console
 - system: `"aarch64-linux"`
 - host os: `Linux 5.15.156, NixOS, 24.05 (Uakari), 24.05.20240418.bc279bb`
 - multi-user?: `yes`
 - sandbox: `yes`
 - nixpkgs: `/run/current-system/nixpkgs`
```

---

RossComputerGuy:
## Description of changes

Fixes #315247

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [x] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/development/libraries/jemalloc/default.nix
+++ b/pkgs/development/libraries/jemalloc/default.nix
@@ -26,6 +26,12 @@ stdenv.mkDerivation rec {
       url = "https://github.com/jemalloc/jemalloc/commit/d00ecee6a8dfa90afcb1bbc0858985c17bef6559.patch";
       hash = "sha256-N5i4IxGJ4SSAgFiq5oGRnrNeegdk2flw9Sh2mP0yl4c=";
     })
+    # fix linking with libc++, can be removed in the next update (after 5.3.0).
+    # https://github.com/jemalloc/jemalloc/pull/2348
+    (fetchpatch {
+      url = "https://github.com/jemalloc/jemalloc/commit/4422f88d17404944a312825a1aec96cd9dc6c165.patch";
+      hash = "sha256-dunkE7XHzltn5bOb/rSHqzpRniAFuGubBStJeCxh0xo=";
+    })
   ];
 
   configureFlags =


pkgs/development/libraries/jemalloc/default.nix
```diff
@@ -47,10 +47,15 @@ stdenv.mkDerivation rec {
     ++ lib.optional (stdenv.isDarwin && stdenv.isx86_64) "--with-lg-vaddr=48"
   ;
 
-  env.NIX_CFLAGS_COMPILE = lib.optionalString stdenv.isDarwin "-Wno-error=array-bounds";
+  env = {
+    NIX_CFLAGS_COMPILE = lib.optionalString stdenv.isDarwin "-Wno-error=array-bounds";
+    # Tries to reference undefined shlibs
+    NIX_LDFLAGS = lib.optionalString (stdenv.cc.bintools.isLLVM && lib.versionAtLeast stdenv.cc.bintools.version "17") "--allow-shlib-undefined";
```
paparodeo, line 53: is this necessary? `pkgsLLVM.jemalloc` builds fine for me without this change and the changes to checkFlags below.

pkgs/development/libraries/jemalloc/default.nix
```diff
@@ -47,10 +47,15 @@ stdenv.mkDerivation rec {
...
+  };
 
-  # Tries to link test binaries binaries dynamically and fails
-  doCheck = !stdenv.hostPlatform.isStatic;
+  # static: Tries to link test binaries binaries dynamically and fails
+  # llvm: Tries linking with an undefined symbol and fails.
+  doCheck = !stdenv.hostPlatform.isStatic && !(stdenv.cc.bintools.isLLVM && lib.versionAtLeast stdenv.cc.bintools.version "17");
```
paparodeo, line 58:
```suggestion
  doCheck = !stdenv.hostPlatform.isStatic;
  checkFlags = lib.optionals condition [ "LDFLAGS=-lc++" ];
```

fixes the tests for me. it seems that this is broken prior to llvm17 too.

paparodeo: tho maybe
```nix
buildFlags = lib.optionals condition [ "LDFLAGS=-lc++" ];
```
is better as that will link jemalloc.so to libc++, which is similar to when compiled with gcc which links to `libstdc++`

pkgs/development/libraries/jemalloc/default.nix
paparodeo: the errors in the bug report involve linking test which you don't do given `doCheck` is false, so still can't see how this flag is doing anything.

RossComputerGuy: Removed

pkgs/development/libraries/jemalloc/default.nix
RossComputerGuy: Yes, that fixed it.

pkgs/development/libraries/jemalloc/default.nix
```diff
@@ -49,9 +49,12 @@ stdenv.mkDerivation rec {
 
   env.NIX_CFLAGS_COMPILE = lib.optionalString stdenv.isDarwin "-Wno-error=array-bounds";
 
-  # Tries to link test binaries binaries dynamically and fails
+  # static: Tries to link test binaries binaries dynamically and fails
+  # llvm: Tries linking with an undefined symbol and fails.
   doCheck = !stdenv.hostPlatform.isStatic;
 
+  checkFlags = lib.optional (stdenv.cc.bintools.isLLVM && lib.versionAtLeast stdenv.cc.bintools.version "17") "LDFLAGS=-lc++";
+
```
paparodeo, line 57: so this is not restricted to LLVM17+ as it is an issue for all LLVMs nor does it seem to  be related to lld. the condition doesn't seem correct as it should be something like `libcxx.isLLVM` i think. but i think the underlying issue is that libjemalloc.so is not linking to `libc++.so`.

when built with gcc toolchain, there is a c++ lib linked in but this is not the case with pkgsLLVM

eg:
```console
$ nix-build -A jemalloc
[...]
$ patchelf --print-needed result/lib/libjemalloc.so
libm.so.6
libstdc++.so.6
libgcc_s.so.1
libpthread.so.0
libc.so.6
$ nix-build --expr 'with import ./. {}; pkgsLLVM.jemalloc.overrideAttrs { doCheck = false; }'
[...]
$ patchelf --print-needed result/lib/libjemalloc.so
libm.so.6
libunwind.so.1
libpthread.so.0
libc.so.6
```

so libjemalloc.so when built with pkgsLLVM is not adding `libc++.so`, but one can look at the undefined symbols, `nm -u result/lib/libjemalloc.so` and see that there are undefined symbols that are defined in `libc++.so`. so while checkFlags works, it seems something is off with the build. as noted before, adding `LDFLAGS=-lc++` conditionally to configureFlags would work it is curious why it is not added in the first place.

paparodeo: on x64 darwin it seems to link to libc++ fine.
```console
$ nix-build -A jemalloc
[...]
$ otool -L result/lib/libjemalloc.dylib
result/lib/libjemalloc.dylib:
        /nix/store/h4fzlgr53fngqx2dk7j5xwzvc5jc3hdm-jemalloc-5.3.0/lib/libjemalloc.2.dylib (compatibility version 0.0.0, current version 0.0.0)
        /nix/store/2yjr5xx0qn1cd72lc3l1mw8k90fg5lxs-libcxx-16.0.6/lib/libc++.1.0.dylib (compatibility version 1.0.0, current version 1.0.0)
        /usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1238.60.2)
```

RossComputerGuy: > so this is not restricted to LLVM17+ as it is an issue for all LLVMs nor does it seem to be related to lld.

Interesting, with `pkgsLLVM` there are a good number of cases where LLVM 17 introduces certain defaults that cause linking with lld to have build issues with various packages. (Notably ncurses)

> so libjemalloc.so when built with pkgsLLVM is not adding `libc++.so`, but one can look at the undefined symbols, `nm -u result/lib/libjemalloc.so` and see that there are undefined symbols that are defined in `libc++.so`. so while checkFlags works, it seems something is off with the build. as noted before, adding `LDFLAGS=-lc++` conditionally to configureFlags would work it is curious why it is not added in the first place.

So add the `LDFLAGS` to both `configureFlags` & `checkFlags`?

RossComputerGuy: > the condition doesn't seem correct as it should be something like `libcxx.isLLVM` i think.

Strange, `libcxx.isLLVM` and `pkgsLLVM.libcxx.isLLVM` are true. I think the check was fine before but without the version check.

paparodeo: so created jemalloc2 which is just HEAD from https://github.com/jemalloc/jemalloc with pkgsLLVM and it doesn't require any changes and is linking to `libc++`
```console
$ nix-build -A pkgsLLVM.jemalloc2
[...]
$ patchelf --print-needed result/lib/libjemalloc.so
libm.so.6
libunwind.so.1
libc++.so.1
libc++abi.so.1
libpthread.so.0
libc.so.6
```

so i guess either find the change that introduces the fix and patch or add `LDFLAGS` to configureFlags conditionally with a note that it can be removed on update.

with `LDFLAGS` added to `configureFlags` it will not be required for `checkFlags`.



> > the condition doesn't seem correct as it should be something like `libcxx.isLLVM` i think.
> 
> Strange, `libcxx.isLLVM` and `pkgsLLVM.libcxx.isLLVM` are true. I think the check was fine before but without the version check.

ah, ok, then perhaps `stdenv.cc.libcxx.isLLVM or false` is the correct flag. as we don't want to set the flag when using clangStdenv as that doesn't use libcxx.

RossComputerGuy: > so i guess either find the change that introduces the fix and patch or add `LDFLAGS` to configureFlags conditionally with a note that it can be removed on update.

This is probably the better and easier solution. I'll add the note to remove it on the next release.

> with `LDFLAGS` added to `configureFlags` it will not be required for `checkFlags`.

Alright, I'll remove the `checkFlags`.

paparodeo: this looks like it could be the PR that does the fix: https://github.com/jemalloc/jemalloc/pull/2348
https://github.com/jemalloc/jemalloc/commit/4422f88d17404944a312825a1aec96cd9dc6c165

RossComputerGuy: Oh then we could just fetch the patch and add it in.

paparodeo: yeah, just confirmed that patch works.

RossComputerGuy: Amazing, I added the patch.

pkgs/development/libraries/jemalloc/default.nix
```diff
@@ -26,7 +26,11 @@ stdenv.mkDerivation rec {
       url = "https://github.com/jemalloc/jemalloc/commit/d00ecee6a8dfa90afcb1bbc0858985c17bef6559.patch";
       hash = "sha256-N5i4IxGJ4SSAgFiq5oGRnrNeegdk2flw9Sh2mP0yl4c=";
     })
-  ];
+    # Fix linking against LLVM's libc++
+  ] ++ lib.optional (stdenv.cc.bintools.isLLVM) (fetchpatch {
```
paparodeo, line 30: i think the patch can be applied unconditionally, and needs a comment to remove on next update as well as a link to `https://github.com/jemalloc/jemalloc/pull/2348`. tho, the condition will allow the change to go to master rather than staging.

and as noted previously, the condition is not correct, it needs to reference `libcxx` somehow, as shown below.

```suggestion
  ] ++ lib.optional (stdenv.cc.libcxx.isLLVM or false) (fetchpatch {
```

as that will work correctly for `libcxxStdenv`

paparodeo (approved):

pkgs/development/libraries/jemalloc/default.nix
RossComputerGuy: Alright, moved to staging and I removed the condition. I also added a note with a link to the PR.

lilyinstarlight:
@ofborg eval

It looks like ofborg eval compared against `master` as base branch instead of `staging` and miscalculated (this also seems to be the reason for code-owner pings)

RossComputerGuy:
Ok, labels still look cursed + pinged people who aren't involved. Going to open a new PR.

Issue #266103, opened 2023-11-07
Don't expose all functions in the top-level `lib`, prefer sub-libraries


infinisil:
`lib` is a bit messy right now, where some functions are available in `lib.*`, but some need to be accessed as `lib.<sublibrary>.*`.

Traditionally all functions were available under `lib.*`, but in recent years we've seen that it doesn't work that well. An example is how `lib.hasPrefix` works on strings, it's aliased to `lib.strings.hasPrefix`. But it also makes sense to have such functionality for lists and paths, we should be able to re-use the same name. This is now done:
- https://github.com/NixOS/nixpkgs/pull/243511
- https://github.com/NixOS/nixpkgs/pull/237610

But `lib.hasPrefix` still needs to point to `lib.strings.hasPrefix` for backwards compat, and this previously caused problems when accidentally using `lib.hasPrefix` for paths:
- https://github.com/NixOS/nixpkgs/pull/221204

Furthermore, `lib` is tending towards the style more emphasised since
- https://github.com/NixOS/nixpkgs/pull/101139

```nix
{ lib }:
let
  inherit (lib.lists)
    length;

  inherit (lib.strings)
    hasPrefix;
in
...
```

which makes sublibraries not much harder to use (no need to always use `lists.length`).

We should establish a standard for sublibraries and make sure it's consistent.

Related issues/comments:
- https://github.com/NixOS/nixpkgs/pull/263969
- https://github.com/NixOS/nixpkgs/pull/261676#discussion_r1378009124

Ping @roberth @h7x4

PR #214021, opened 2023-02-01
lib.lists.ifilter0: init

infinisil:
###### Description of changes

Originally part of https://github.com/NixOS/nixpkgs/pull/119286, credit for the smart O(n) implementation goes to @bb010g! I only rebased, improved readability and added comments and tests.

###### Things done

- [x] Wrote and ran tests successfully (`nix-instantiate --eval --strict lib/tests/misc.nix`)
- [x] Has documentation

PR #315411, opened 2024-05-28
lib.strings: add `trim`

MattSturgeon:
`strings.trim` returns a copy of the string with all leading and trailing whitespace removed.

`strings.trimWith` does the same thing, but calling code can decide whether to trim the start and/or end of the string.

The obvious regex `[[:string:]]*(.*)[[:string:]]*` wouldn't work, because the `(.*)` is greedy and matches trailing whitespace. Since [ERE](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap09.html#tag_09_04) doesn't support lazy/reluctant matches (`.*?`) introduced in perl-style regexes, we need to use a regex that fails to match blank/empty strings. This case is handled using `optionalString`.

## Description of changes

`trim` is a fairly core function included in most standard libraries. I've personally come across a couple scenarios where I would've liked to have been able to trim strings.

I've added a few basic tests.

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

Pinging recent contributors to lib.string: @infinisil @AndersonTorres @FireyFly @eclairevoyant @h7x4

---

### DIFF
--- a/lib/strings.nix
+++ b/lib/strings.nix
@@ -157,6 +157,68 @@ rec {
   */
   replicate = n: s: concatStrings (lib.lists.replicate n s);
 
+  /*
+    Remove leading and trailing whitespace from a string.
+
+    Whitespace is defined as any of the following characters:
+      " ", "\t" "\r" "\n"
+
+    Type: trim :: string -> string
+
+    Example:
+      trim "   hello, world!   "
+      => "hello, world!"
+  */
+  trim = trimWith {
+    start = true;
+    end = true;
+  };
+
+  /*
+    Remove leading and/or trailing whitespace from a string.
+
+    Whitespace is defined as any of the following characters:
+      " ", "\t" "\r" "\n"
+
+    Type: trimWith :: Attrs -> string -> string
+
+    Example:
+      trimWith { start = true; } "   hello, world!   "}
+      => "hello, world!   "
+      trimWith { end = true; } "   hello, world!   "}
+      => "   hello, world!"
+  */
+  trimWith =
+    {
+      # Trim leading whitespace
+      start ? false,
+      # Trim trailing whitespace
+      end ? false,
+    }:
+    s:
+    let
+      # Define our own whitespace character class instead of using
+      # `[:space:]`, which is not well-defined.
+      chars = " \t\r\n";
+
+      # To match up until trailing whitespace, we need to capture a
+      # group that ends with a non-whitespace character.
+      regex =
+        if start && end then
+          "[${chars}]*(.*[^${chars}])[${chars}]*"
+        else if start then
+          "[${chars}]*(.*)"
+        else if end then
+          "(.*[^${chars}])[${chars}]*"
+        else
+          "(.*)";
+
+      # If the string was empty or entirely whitespace,
+      # then the regex may not match and `res` will be `null`.
+      res = match regex s;
+    in
+      optionalString (res != null) (head res);
+
   /* Construct a Unix-style, colon-separated search path consisting of
      the given `subDir` appended to each of the given paths.
 
--- a/lib/tests/misc.nix
+++ b/lib/tests/misc.nix
@@ -369,6 +369,72 @@ runTests {
     expected = "hellohellohellohellohello";
   };
 
+  # Test various strings are trimmed correctly
+  testTrimString = {
+    expr =
+    let
+      testValues = f: mapAttrs (_: f) {
+        empty = "";
+        cr = "\r";
+        lf = "\n";
+        tab = "\t";
+        spaces = "   ";
+        leading = "  Hello, world";
+        trailing = "Hello, world   ";
+        mixed = " Hello, world ";
+        mixed-tabs = " \t\tHello, world \t \t ";
+        multiline = "  Hello,\n  world!  ";
+        multiline-crlf = "  Hello,\r\n  world!  ";
+      };
+    in
+      {
+        leading = testValues (strings.trimWith { start = true; });
+        trailing = testValues (strings.trimWith { end = true; });
+        both = testValues strings.trim;
+      };
+    expected = {
+      leading = {
+        empty = "";
+        cr = "";
+        lf = "";
+        tab = "";
+        spaces = "";
+        leading = "Hello, world";
+        trailing = "Hello, world   ";
+        mixed = "Hello, world ";
+        mixed-tabs = "Hello, world \t \t ";
+        multiline = "Hello,\n  world!  ";
+        multiline-crlf = "Hello,\r\n  world!  ";
+      };
+      trailing = {
+        empty = "";
+        cr = "";
+        lf = "";
+        tab = "";
+        spaces = "";
+        leading = "  Hello, world";
+        trailing = "Hello, world";
+        mixed = " Hello, world";
+        mixed-tabs = " \t\tHello, world";
+        multiline = "  Hello,\n  world!";
+        multiline-crlf = "  Hello,\r\n  world!";
+      };
+      both = {
+        empty = "";
+        cr = "";
+        lf = "";
+        tab = "";
+        spaces = "";
+        leading = "Hello, world";
+        trailing = "Hello, world";
+        mixed = "Hello, world";
+        mixed-tabs = "Hello, world";
+        multiline = "Hello,\n  world!";
+        multiline-crlf = "Hello,\r\n  world!";
+      };
+    };
+  };
+
   testSplitStringsSimple = {
     expr = strings.splitString "." "a.b.c.d";
     expected = [ "a" "b" "c" "d" ];
--- a/lib/default.nix
+++ b/lib/default.nix
@@ -105,7 +105,7 @@ let
       hasInfix hasPrefix hasSuffix stringToCharacters stringAsChars escape
       escapeShellArg escapeShellArgs
       isStorePath isStringLike
-      isValidPosixName toShellVar toShellVars
+      isValidPosixName toShellVar toShellVars trim trimWith
       escapeRegex escapeURL escapeXML replaceChars lowerChars
       upperChars toLower toUpper addContextFrom splitString
       removePrefix removeSuffix versionOlder versionAtLeast


eclairevoyant:
Should this be added to the top-level (`lib/default.nix`) as well? (Example PR: #214021)

MattSturgeon:
> Should this be added to the top-level (`lib/default.nix`) as well? (Example PR: #214021)

Perhaps, but based on #266103 I'd guess it's preferred not to add new functions to the top-level?

infinisil:
I'll try to review this a bit more closely soon, but yeah for now it should still be exported under `lib.*` too, because that's the status-quo :)

lib/tests/misc.nix
```diff
@@ -196,6 +196,25 @@ runTests {
     expected = "hellohellohellohellohello";
   };
 
+  testTrimString = {
+    expr = mapAttrs strings.trim {
```
eclairevoyant, line 200:
```suggestion
    expr = mapAttrs (_: strings.trim) {
```
The function passed to `mapAttrs` takes two args (attribute name and value, respectively). By the way, you can run the tests in a local clone with `nix-build lib/tests/release.nix`.

eclairevoyant (dismissed):

lib/tests/misc.nix
MattSturgeon: > By the way, you can run the tests in a local clone with `nix-build lib/tests/release.nix`.

Thanks!

lib/strings.nix
```diff
@@ -158,6 +158,24 @@ rec {
...
+
+    Example:
+      trim "   hello   "
+      => "hello"
+  */
+  trim = s:
+    let
+      res = match "[[:space:]]*(.*[^[:space:]])[[:space:]]*" s;
+    in
+      optionalString (res != null) (head res);
```
infinisil, line 177: Can you explain how this works and why it's correct? :sweat_smile: Ideally as a code comment :)

lib/strings.nix
```diff
@@ -158,6 +158,24 @@ rec {
   */
   replicate = n: s: concatStrings (lib.lists.replicate n s);
 
+  /*
+    Remove leading and trailing whitespace from a string.
+
+    Whitespace is defined as any characters matching the regex
+    `[[:space:]]`.
```
infinisil, line 165: It would be good to be more specific here. You could mention that it uses [`builtins.match`](https://nix.dev/manual/nix/2.18/language/builtins.html?highlight=builtins.match#builtins-match) underneath, though digging down into what it uses for `[:space:]` makes me believe that it might be dependent on the `LC_CTYPE` env var, which would be bad..

infinisil (dismissed):

lib/strings.nix
MattSturgeon: Are you suggesting we use a more explicit character class, along the lines of `[ \xa0\r\n]`?

infinisil: That would definitely avoid the ambiguity! I'm also wondering if non-Ascii spaces would even get matched by Nix

eclairevoyant: Not sure how thorough we're looking to be here, but there's also [these space characters](https://en.wikipedia.org/wiki/Whitespace_character#Unicode)

AndersonTorres (approved):

lib/strings.nix
MattSturgeon: > there's also [these space characters](https://en.wikipedia.org/wiki/Whitespace_character#Unicode)

That's why I'd prefer we use a well defined class like `[:space:]`, but if it is non-deterministic that's less than ideal.

Either way, I'll add some more tests for whatever chars we deem to be whitespace/non-whitespace.

eclairevoyant: > That's why I'd prefer we use a well defined class like `[:space:]`

It's [not well-defined](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap09.html#tag_09_04
), that's precisely the issue ;)

> A character class expression shall represent the union of two sets:
>
> * The set of single characters that belong to the character class, as defined in the `LC_CTYPE` category in the current locale.
>
> * An unspecified set of multi-character collating elements.

MattSturgeon:
Would nixpkgs be open to also having `trimStart` and `trimEnd` (possible aliases/alternate names `trimHead`/`trimTail`, `trimLeading`.`trimTrailing`)?

If so, would you want them in this PR or separately?

lib/strings.nix
MattSturgeon: > Perhaps `[ \t\r\n]`?

This is done. @infinisil Let me know if you'd prefer a different approach.

AndersonTorres:
What is the prevalence of `trim` to begin with?

eclairevoyant:
> What is the prevalence of `trim` to begin with?
> 
> In the sense of: "all these lines of code in Nixpkgs are mere applications of trim".

https://github.com/NixOS/nixpkgs/blob/4eb44db1f53f2cc57330a929d8e27fcd4d73f383/lib/strings.nix#L1060 and friends all try to implement trim

https://github.com/NixOS/nixpkgs/blob/4eb44db1f53f2cc57330a929d8e27fcd4d73f383/lib/strings.nix#L1174 is more of a trimEnd, but still

Also, I'd want to trim option descriptions in some cases (especially for `mkEnableOption`).

eclairevoyant (approved):

lib/strings.nix
```diff
@@ -157,6 +157,29 @@ rec {
...
+    Whitespace is defined as any of the following characters:
+      " ", "\t" "\r" "\n"
+
+    Type: trim :: string -> string
+
+    Example:
+      trim "   hello, world!   "
+      => "hello, world!"
+  */
+  trim = s:
```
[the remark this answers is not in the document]
eclairevoyant: what's "chomp" vs "trim"?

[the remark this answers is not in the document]
AndersonTorres: Basically,

```nix
chompBegin = s:
  trim s true false;
chompEnd = s:
  trim s false true;
chomp = s:
  trim s true true;
```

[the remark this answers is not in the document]
MattSturgeon: We could have a configurable `trim'`:

```nix
trim = trim' {
  start = true;
  end = true;
};

trim' =
  {
    start ? false,
    end ? false,
    ... # forward compatible
  }:
  s:
  # implementation
```

With this approach idk whether or not having `trimStart` and `trimEnd` has any value, since the configurable version can be used.

Not sure if it's better to have `start` and `end` arguments default to true or false though?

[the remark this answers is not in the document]
AndersonTorres: Nice idea!

This "general function plus typical usage" approach is typical around Nixpkgs. We have `getExe'` plus `getExe`, after all.

Putting `begin` and `end` as false by default is most natural, since they will be overridden anyway: if I want to trim the beginning but not the ending, it would be not so ergonomical to set both explicitly.

MattSturgeon:
As suggested (https://github.com/NixOS/nixpkgs/pull/315411#discussion_r1690161175) I've implemented a configurable `trim'` too.

@AndersonTorres does this achieve what you wanted?

@eclairevoyant would you mind re-reviewing since the implementation has changed?

AndersonTorres (approved):

lib/strings.nix
```diff
@@ -157,6 +157,76 @@ rec {
...
+
+    Type: trim' :: Attrs -> string -> string
+
+    Example:
+      trim' { start = true; } "   hello, world!   "}
+      => "hello, world!   "
+      trim' { end = true; } "   hello, world!   "}
+      => "   hello, world!"
+  */
+  trim' =
```
infinisil, line 191:
```suggestion
  trimWith =
```

This is emerging as a convention :)

infinisil (commented):
Just a minor comment, otherwise I think this is good!

lib/strings.nix
MattSturgeon: Sure thing, done.

lib/strings.nix
```diff
@@ -157,6 +157,70 @@ rec {
...
+      => "   hello, world!"
+  */
+  trimWith =
+    {
+      # Trim leading whitespace
+      start ? false,
+      # Trim trailing whitespace
+      end ? false,
+      # Forward compatibility
+      ...
```
infinisil, line 198:
```suggestion
```

I don't think we should have the ellipsis, because if we support an extra argument in the future, the old version shouldn't blindly accept and ignore such an attribute. Instead users should rightly get an error that the old version doesn't accept that attribute.

MattSturgeon: I've seen the pattern used on other functions, but your argument makes sense.

infinisil: Perhaps this one: https://github.com/NixOS/nixpkgs/blob/7af9a82b766258e9cd53e7d1097fe5c748f78672/lib/fileset/default.nix#L799-L804

In that case at least, the difference is that the attributes are provided by the library (and new ones may be added over time), while the user provides the function to handle them. There's something super naughty there to guarantee that the user doesn't forget the `...` in their function: https://github.com/NixOS/nixpkgs/blob/88733d6c42306ab5d094cccdaa8d7f9c90ea63b5/lib/fileset/internal.nix#L835-L837

:laughing:

MattSturgeon: Interesting example!

The only concrete example I can think of is outside of `lib` anyway, so probably not relevant: https://github.com/NixOS/nixpkgs/blob/72b5f58e5bba538ca62c10cd1ab5a4baef03402a/pkgs/applications/editors/neovim/utils.nix#L41-L42

and

https://github.com/NixOS/nixpkgs/blob/b7aed13df5ea800264d72cb76b9db5031afc322f/pkgs/applications/editors/neovim/wrapper.nix#L46

Either way, I've removed the ellipses, thanks for the feedback!

infinisil (approved):
Looking good to me, nice work!

infinisil:
Small follow-up: https://github.com/NixOS/nixpkgs/pull/330034

PR #291559, opened 2024-02-26
libvpl: patch to search `/run/opengl-drivers/lib/` for runtime implem…

RyanGibb:
## Description of changes

libvpl: patch to search `/run/opengl-drivers/lib/` for runtime implementations.

This means instead of setting:
```
  environment.sessionVariables = {
    ONEVPL_SEARCH_PATH = lib.strings.makeLibraryPath (with pkgs; [oneVPL-intel-gpu]);
  };
```

To find `onevpl-intel-gpu` libraries instead it will pick it up from:
```
  hardware.opengl.extraPackages = [ pkgs.onevpl-intel-gpu ];
```

See https://github.com/NixOS/nixpkgs/pull/264621 for more detail.

Tagging @Atemu and @evanrichter for interest.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
- For non-Linux: Is sandboxing enabled in `nix.conf`? (See [Nix manual](https://nixos.org/manual/nix/stable/command-ref/conf-file.html))
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:
  - [NixOS test(s)](https://nixos.org/manual/nixos/unstable/index.html#sec-nixos-tests) (look inside [nixos/tests](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests))
  - and/or [package tests](https://nixos.org/manual/nixpkgs/unstable/#sec-package-tests)
  - or, for functions and "core" functionality, tests in [lib/tests](https://github.com/NixOS/nixpkgs/blob/master/lib/tests) or [pkgs/test](https://github.com/NixOS/nixpkgs/blob/master/pkgs/test)
  - made sure NixOS tests are [linked](https://nixos.org/manual/nixpkgs/unstable/#ssec-nixos-tests-linking) to the relevant packages
- [ ] Tested compilation of all packages that depend on this change using `nix-shell -p nixpkgs-review --run "nixpkgs-review rev HEAD"`. Note: all changes have to be committed, also see [nixpkgs-review usage](https://github.com/Mic92/nixpkgs-review#usage)
- [ ] Tested basic functionality of all binary files (usually in `./result/bin/`)
[...]

PR #264621, opened 2023-10-31
oneVPL for intel gpus

evanrichter:
## Description of changes

fresh PR continuing from #242359, sorry for the mass-ping over there :(

ping for the people who seem interested: @Atemu @PJungkamp @J-Swift

outstanding question for @midchildan: should intel-media-sdk be marked deprecated? The [upstream repo](https://github.com/Intel-Media-SDK/MediaSDK#media-sdk-support-matrix) is archived and suggests transitioning to onevpl

also todo: Atemu says ofborg build failed for onevpl-intel-gpu on aarch-linux, but I can't find the CI run for it. marking as draft until that's re-run and I can debug it

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
- For non-Linux: Is sandboxing enabled in `nix.conf`? (See [Nix manual](https://nixos.org/manual/nix/stable/command-ref/conf-file.html))
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:
  - [NixOS test(s)](https://nixos.org/manual/nixos/unstable/index.html#sec-nixos-tests) (look inside [nixos/tests](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests))
  - and/or [package tests](https://nixos.org/manual/nixpkgs/unstable/#sec-package-tests)
  - or, for functions and "core" functionality, tests in [lib/tests](https://github.com/NixOS/nixpkgs/blob/master/lib/tests) or [pkgs/test](https://github.com/NixOS/nixpkgs/blob/master/pkgs/test)
  - made sure NixOS tests are [linked](https://nixos.org/manual/nixpkgs/unstable/#ssec-nixos-tests-linking) to the relevant packages
- [ ] Tested compilation of all packages that depend on this change using `nix-shell -p nixpkgs-review --run "nixpkgs-review rev HEAD"`. Note: all changes have to be committed, also see [nixpkgs-review usage](https://github.com/Mic92/nixpkgs-review#usage)
- [ ] Tested basic functionality of all binary files (usually in `./result/bin/`)
- [23.11 Release Notes](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2311.section.md) (or backporting [23.05 Release notes](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2305.section.md))
[...]

PR #315415, opened 2024-05-28
Add "/run/opengl-driver" to jellyfin runtimeDeps

pjungkamp:
## Description of changes

Add `"/run/opengl-driver"` to jellyfin server `runtimeDeps`.

This allows intel-media-sdk in the ffmpeg transcoding process spawned by jellyfin to find and load the
`libmfx-gen.so.1.2` from onevpl-intel-gpu if it's included in `hardware.opengl.extraPackages`.

## Workarounds

Adding an overlay with `jellyfin-ffmpeg.override { withVpl = true; withMfx = false}` should therefore not be required anymore.

If you're currently using an overlay like that you might want to try this overlay instead as builds considerably faster:
```nix
final: prev: {
  jellyfin = prev.jellyfin.overrideAttrs (finalAttrs: prevAttrs: {
    makeWrapperArgs = (prevAttrs.makeWrapperArgs or []) ++ ["--suffix LD_LIBRARY_PATH : /run/opengl-driver/lib"];
  });
}
```

## "Worse" approaches

- Adding `"/run/opengl-driver/lib"` to the ffmpeg `LD_LIBRARY_PATH` as a general hack was discarded: https://github.com/NixOS/nixpkgs/pull/264621#issuecomment-1963569055
- We could add an `LD_LIBRARY_PATH` only to `jellyfin-ffmpeg` but that's also a hack.

## A better approach?

This causes more rebuilds, so I opted to do the simple fix above for now.

We could add a patch to `intel-media-sdk` here: https://github.com/Intel-Media-SDK/MediaSDK/blob/7a72de33a15d6e7cdb842b12b901a003f7154f0a/api/mfx_dispatch/linux/mfxloader.cpp#L192-L216

This is an example patch that seems to work: [nixos-search-onevplrt-in-run-opengl-driver-lib.patch](https://github.com/NixOS/nixpkgs/files/15473862/nixos-search-onevplrt-in-run-opengl-driver-lib.patch.txt)

This would be the `intel-media-sdk` equivalent to https://github.com/NixOS/nixpkgs/pull/291559.

## Tags for interest

@RyanGibb 
@evanrichter
@Atemu 

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/servers/jellyfin/default.nix
+++ b/pkgs/servers/jellyfin/default.nix
@@ -29,6 +29,7 @@ buildDotnetModule rec {
   executables = [ "jellyfin" ];
   nugetDeps = ./nuget-deps.nix;
   runtimeDeps = [
+    "/run/opengl-driver"
     ffmpeg
     fontconfig
     freetype


RyanGibb:
Hi @PJungkamp. Just to be clear,  what problem this is solving? I've been using VPL for QSV with jellyfin successfully for a while now, though I am overriding ffmpeg to use libvpl:

```
    ffmpeg_6 = prev.ffmpeg_6.overrideAttrs (old: rec {
      configureFlags =
        # Remove deprecated Intel Media SDK support
        (builtins.filter (e: e != "--enable-libmfx") old.configureFlags)
        # Add Intel VPL support
        ++ [ "--enable-libvpl" ];
      buildInputs = old.buildInputs ++ [
        # VPL dispatcher
        pkgs.libvpl
      ];
    });
```

Perhaps something worth considering; Intel Media SDK looks to be deprecated https://github.com/Intel-Media-SDK/MediaSDK

PJungkamp:
I'm annoyed that this override was necessary.
Overriding ffmpeg shouldn't be needed to get Jellyfin with QSV running.

Intel Media SDK can act as a proxy that dispatches to onevpl-intel-gpu. This allows the ffmpeg package compiled without `--enable-libvpl` (but with `--enable-libmfx`) to dispatch to onevpl-intel-gpu and thus use QSV on platforms like alder lake and newer.

PJungkamp:
I've posted the initial working overlays for Jellyfin in evanrichter's onevpl PR and have been using Jellyfin on NixOS for months now.

But I'd like to see a more "out of the box" solution for ffmpeg with QSV, especially for Jellyfin.

RyanGibb:
> I'm annoyed that this override was necessary.

Makes sense!

> Intel Media SDK can act as a proxy that dispatches to onevpl-intel-gpu.

The intel-media-sdk mentions security issues that makes me think it would be preferable to set ffmpeg to use oneVPL via a flag?

evanrichter:
I agree the override is annoying. But the purpose is to make sure people who know what they need, can do the override, while being 100% sure we don't break people on older hardware. This is mostly my inexperience with intel gpu drivers and dispatch, so I'm sorry I don't know for sure.

Atemu:
> This causes more rebuilds, so I opted to do the simple fix above for now.

As long as you send your PR to staging, rebuilds aren't really a problem. I'd rather see a "proper" upstream-sanctioned solution to this issue than yet another hack.

PJungkamp:
> As long as you send your PR to staging, rebuilds aren't really a problem. I'd rather see a "proper" upstream-sanctioned solution to this issue than yet another hack.

Ok, I'll close this one then and pursue https://github.com/NixOS/nixpkgs/pull/315425 instead.
I'll need some guidance on how to approach a PR with some more rebuilds then.

RyanGibb:
> The intel-media-sdk mentions security issues that makes me think it would be preferable to set ffmpeg to use oneVPL via a flag?

@PJungkamp @Atemu There are a number of CVE's for intel media sdk https://cve.mitre.org/cgi-bin/cvekey.cgi?keyword=Intel+Media+SDK. Intel recommends upgrading to oneVPL https://www.intel.com/content/www/us/en/security-center/advisory/intel-sa-00935.html. It looks like oneVPL supports legacy hardware https://www.intel.com/content/www/us/en/docs/onevpl/upgrade-from-msdk/2023-1/onevpl-hardware-support-details.html.

PJungkamp:
It seems to be possible to dispatch from `onevpl` to `intel-media-sdk`. As far as I can tell our oneVPL package searches the implementation in `/run/opengl-driver/lib`.

Could someone with an Intel iGPU from 10th gen or older try running `ffmpeg_6-full.override { withVpl = true; withMfx = false; }`? I don't have an older device running NixOS...

If it does not work (as I would expect) try adding `intel-media-sdk` to `hardware.opengl.extraPackages` and see if that fixes it.

This command should suffice as a test.

```console
$ ffmpeg -init_hw_device vaapi=va:,kernel_driver=i915,driver=iHD -init_hw_device qsv=qs@va -filter_hw_device qs
```

Atemu:
@RyanGibb please don't create comments on closed unrelated PRs that should have been a separate issue.

Those "C"VEs don't appear particularly critical to me; not to the degree that I think would warrant any sort of immediate response. There are barely any details on these however, so judging this is near impossible. If there was a vuln that'd allow arbitrary code execution, that would be the point where I'd draw the line to mark it as vulnerable and therefore disable it in ffmpeg.

Of course Intel recommends switching to oneVPL. They only care about selling new CPUs/GPUs and don't have much regard for distros who need to care about their users rather than Intel's profits.

FWIW, they've deprecated intel-media-sdk. I think we should keep it enabled in ffmpeg until supporting it is unattainable.

As for supporting older GPUs via oneVPL: I don't care whether we do intel-media-sdk via oneVPL or the other way around. It appears you always still need intel-media-sdk in the loop though. Patches welcome.

PR #315557, opened 2024-05-29
nixos/redis: Use valkey package by default

### LINKED ISSUE #310061: Valkey default package for redis service
## Issue description

Make [Valkey](https://github.com/NixOS/nixpkgs/pull/302935) the default package for redis service

onny:
## Description of changes

[Valkey](https://github.com/valkey-io/valkey) is a fork of the formerly open-source Redis project. It emerged as a response to Redis’s licensing changes. While Redis transitioned to a “source-available” license, Valkey retained a FOSS-compatible license. The Valkey community actively supports and develops the project, making it a viable alternative to Redis.

Valkey is fully compatible with the Redis API, protocol, persistence data, ports, and configuration files. [If you’re migrating from Redis, you can move your existing configuration files and persistence data directly to Valkey without translation steps](https://fedoramagazine.org/how-to-move-from-redis-to-valkey/)[4](https://fedoramagazine.org/how-to-move-from-redis-to-valkey/).

Fixes https://github.com/NixOS/nixpkgs/issues/310061

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/nixos/modules/services/databases/redis.nix
+++ b/nixos/modules/services/databases/redis.nix
@@ -51,7 +51,13 @@ in {
   options = {
 
     services.redis = {
-      package = lib.mkPackageOption pkgs "redis" { };
+      package = lib.mkOption {
+        type = lib.types.package;
+        default = pkgs.valkey;
+        defaultText = lib.literalExpression "pkgs.valkey";
+        description = "Which package to use for the Redis service.";
+        relatedPackages = [ "valkey" "redis" ];
+      };
 
       vmOverCommit = lib.mkEnableOption ''
         set `vm.overcommit_memory` sysctl to 1
--- a/nixos/doc/manual/release-notes/rl-2505.section.md
+++ b/nixos/doc/manual/release-notes/rl-2505.section.md
@@ -576,6 +576,8 @@
 
 - [`system.stateVersion`](#opt-system.stateVersion) is now validated and must be in the `"YY.MM"` format, ideally corresponding to a prior NixOS release.
 
+- The default package for the `redis` service got changed to the drop-in replacement `valkey`. The `redis` package can still used by configuring [`services.redis.package`](#opt-services.redis.package).
+
 - GOverlay has been updated to 1.2, please check the [upstream changelog](https://github.com/benjamimgois/goverlay/releases) for more details.
 
 - [`services.mongodb`](#opt-services.mongodb.enable) is now compatible with the `mongodb-ce` binary package. To make use of it, set [`services.mongodb.package`](#opt-services.mongodb.package) to `pkgs.mongodb-ce`.


nicoonoclaste (approved):

nixos/modules/services/databases/redis.nix
```diff
@@ -54,7 +54,13 @@ in {
...
 
     services.redis = {
-      package = mkPackageOption pkgs "redis" { };
+      package = mkOption {
+        type = types.package;
+        default = pkgs.valkey;
+        defaultText = literalExpression "pkgs.valkey";
+        description = "Which package to use for the Redis service.";
+        relatedPackages = [ "valkey" "redis" ];
+      };
```
SuperSandro2000, line 63: Did you think about guarding this behind system.stateVersion? What's the reasoning to not do it?

I am fully sure Valkey is fully compatible but there is probably some software out there which might be not that small, which matches the product name in some case and we could break that and I would like to avoid doing that silently.

Rucadi:
I also raised my concerns about the command and naming change in discourse, which may be surprising to some degree by users, Sandro's view is right.

yu-re-ka (changes requested):
Long-term the users should choose on their own, which service they replace redis with, since there is not one single successor project, and in the future redis, redict and valkey will very likely diverge to a point where they are not drop-ins for another it might be necessary to run both on one machine.
Consider something like I initially attempted when I added redict: https://github.com/NixOS/nixpkgs/commit/aed7744d1ad452368c4dae0b704d297674322c67#diff-5c454cebbf2adf0ddb06f9c5504c428d14173ec4b8103172186da942f05fd5c2R307 (did not get any feedback from redis maintainers or anyone really, so I removed that commit from that PR)

K900:
Can you please rebase this on top of latest master? It would be nice to land this for 25.05.

onny:
Rebased :)

nixos/modules/services/databases/redis.nix
onny: Can you give me an example on how to do this :)

K900: Please stop using stateVersion for this.

NyCodeGHG:
Can we still get this into 25.05? Or are we too late?
Not sure if this is considered a breaking change, since valkey should be still compatible?

Rucadi:
Relevant: https://redis.io/blog/agplv3/

NyCodeGHG:
I think Valkey would still be the better choice for most users, even with Redis Inc. giving an AGPL option for redis.

Rucadi:
Why not just services/redis and services/valkey? It's clear and non-confusing, for example, Arch is discontinuing redis but not making you install "valkey" when you mean "redis", you have to just install valkey.

For me, if it's decided "redis bad redis gone", deprecate the redis service and mark it for deletion, while creating a new valkey service. However, I would just leave them coexist.

NyCodeGHG:
@Rucadi that would make it harder for services which need a redis-like thing to offer a choice between redis or valkey.

Rucadi:
> @Rucadi that would make it harder for services which need a redis-like thing to offer a choice between redis or valkey.

If it is not desired to have it in different services, then in my opinion then the "default" should be the one to name the service, so I would prefer to rename redis to valkey and offer a deprecation warning on "redis" to avoid breaking existing configurations.

SuperSandro2000:
> Why not just services/redis and services/valkey?

Why do we need two services if they are 100% compatible right now and we effectively would need to do each change to both of them?

pbsds:
https://www.gomomento.com/blog/valkey-turns-one-how-the-community-fork-left-redis-in-the-dust/

mweinelt:
Closing for the same reason as https://github.com/NixOS/nixpkgs/issues/310061.

PR #315616, opened 2024-05-29
doc/meta: Mention --version as a good usecase for installCheckPhase

doronbehar:
## Description of changes

The first 3 commits of this PR are rather trivial, hopefully will be easily accepted. The last commit is a minor change of policy - recommending using `installCheckPhase` over `passthru.tests.version`.

This is in response to a discussion that wasn't fully resolved in [the original PR that added these lines](https://github.com/NixOS/nixpkgs/pull/119731#discussion_r615367755) and [at discourse here](https://discourse.nixos.org/t/improve-the-jdk-infrastructure-on-nixpkgs/45678/14)

The arguments for using `installCheckPhase` over `passthru.tests.version`, for the simple case of a `--version | grep ${version}` test, should be fully explained in the content added. Review is welcome.

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/doc/stdenv/meta.chapter.md
+++ b/doc/stdenv/meta.chapter.md
@@ -137,7 +137,7 @@ Prefer `passthru.tests` for tests that are introduced in nixpkgs because:
 * we can run `passthru.tests` independently
 * `installCheckPhase` adds overhead to each build
 
-For more on how to write and run package tests, see [](#sec-package-tests).
+For more on how to write and run package tests, see [`pkgs/README.md`](https://github.com/NixOS/nixpkgs/blob/master/pkgs/README.md#package-tests).
 
 #### NixOS tests {#var-meta-tests-nixos}
 
--- a/doc/stdenv/meta.chapter.md
+++ b/doc/stdenv/meta.chapter.md
@@ -134,7 +134,7 @@ Tests that are part of the source package are often executed in the `installChec
 Prefer `passthru.tests` for tests that are introduced in nixpkgs because:
 
 * `passthru.tests` tests the 'real' package, independently from the environment in which it was built
-* we can run `passthru.tests` independently
+* We can run and debug a `passthru.tests` independently, after the package was built (useful if it takes a long time).
 * `installCheckPhase` adds overhead to each build
 
 For more on how to write and run package tests, see [`pkgs/README.md`](https://github.com/NixOS/nixpkgs/blob/master/pkgs/README.md#package-tests).
--- a/doc/stdenv/meta.chapter.md
+++ b/doc/stdenv/meta.chapter.md
@@ -129,9 +129,29 @@ Note that Hydra and [`nixpkgs-review`](https://github.com/Mic92/nixpkgs-review) 
 
 #### Package tests {#var-meta-tests-packages}
 
-Tests that are part of the source package are often executed in the `installCheckPhase`.
+Tests that are part of the source package are often executed in the `installCheckPhase`. This phase is also suitable for performing a `--version` test for packages that support such flag. Here's an example:
 
-Prefer `passthru.tests` for tests that are introduced in nixpkgs because:
+```nix
+# Say the package is git
+stdenv.mkDerivation(finalAttrs: {
+  pname = "git";
+  version = "...";
+  # ...
+
+  doInstallCheck = true;
+  installCheckPhase = ''
+    runHook preInstallCheck
+
+    echo checking if 'git --version' mentions ${finalAttrs.version}
+    $out/bin/git --version | grep ${finalAttrs.version}
+
+    runHook postInstallCheck
+  '';
+  # ...
+})
+```
+
+Most programs distributed by Nixpkgs support such a `--version` flag, and it can help give confidence that the package at least got compiled properly. However, tests that are slightly non trivial will better fit into `passthru.tests`, because:
 
 * `passthru.tests` tests the 'real' package, independently from the environment in which it was built
 * We can run and debug a `passthru.tests` independently, after the package was built (useful if it takes a long time).
--- a/doc/stdenv/meta.chapter.md
+++ b/doc/stdenv/meta.chapter.md
@@ -125,6 +125,8 @@ $ cd path/to/nixpkgs
 $ nix-build -A your-package.tests
 ```
 
+Note that Hydra and [`nixpkgs-review`](https://github.com/Mic92/nixpkgs-review) don't build these derivations by default, and that ([`@ofborg`](https://github.com/NixOS/ofborg)) only builds them when evaluating PRs for that particular package (or when manually instructed).
+
 #### Package tests {#var-meta-tests-packages}
 
 Tests that are part of the source package are often executed in the `installCheckPhase`.
--- a/doc/stdenv/meta.chapter.md
+++ b/doc/stdenv/meta.chapter.md
@@ -157,6 +157,8 @@ Most programs distributed by Nixpkgs support such a `--version` flag, and it can
 * We can run and debug a `passthru.tests` independently, after the package was built (useful if it takes a long time).
 * `installCheckPhase` adds overhead to each build
 
+It is also possible to still use `passthru.tests` to test the version, with [testVersion](#tester-testVersion).
+
 For more on how to write and run package tests, see [`pkgs/README.md`](https://github.com/NixOS/nixpkgs/blob/master/pkgs/README.md#package-tests).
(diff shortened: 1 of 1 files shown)


doc/stdenv/meta.chapter.md
```diff
@@ -125,17 +125,38 @@ $ cd path/to/nixpkgs
 $ nix-build -A your-package.tests
 ```
 
+Note that Hydra and [`nixpkgs-review`](https://github.com/Mic92/nixpkgs-review) don't build these derivations by default, but they are useful pointers for Nixpkgs' CI ([`@ofborg`](https://github.com/NixOS/ofborg)) that may help identify issues when changes such as version bumps are directly made to the package.
```
raboof, line 128:
```suggestion
Note that Hydra and [`nixpkgs-review`](https://github.com/Mic92/nixpkgs-review) don't build these derivations by default, and that ([`@ofborg`](https://github.com/NixOS/ofborg)) only builds them when evaluating PRs for that particular package (or when manually instructed).
```

doc/stdenv/meta.chapter.md
```diff
@@ -125,17 +125,38 @@ $ cd path/to/nixpkgs
...
+
 #### Package tests {#var-meta-tests-packages}
 
-Tests that are part of the source package are often executed in the `installCheckPhase`.
+Tests that are part of the source package are often executed in the `installCheckPhase`. This phase is also suitable for performing a `--version` test for packages that support such flag. Here's an example:
+
+```nix
+# Say the package is git
+stdenv.mkDerivation(finalAttrs: {
```
raboof, line 136: this style seems less common than `mkDerivation rec {`, is there a reason we'd want to show `finalAttrs` here?

doc/stdenv/meta.chapter.md
```diff
@@ -125,17 +125,38 @@ $ cd path/to/nixpkgs
...
+  pname = "git";
+  version = "...";
+  # ...
+
+  installCheckPhase = ''
+    runHook preInstallCheck
+
+    echo checking if git --version spits ${finalAttrs.version}
```
raboof, line 144:
```suggestion
    echo checking if 'git --version' mentions ${finalAttrs.version}
```

doc/stdenv/meta.chapter.md
```diff
@@ -125,17 +125,38 @@ $ cd path/to/nixpkgs
...
+Tests that are part of the source package are often executed in the `installCheckPhase`. This phase is also suitable for performing a `--version` test for packages that support such flag. Here's an example:
+
+```nix
+# Say the package is git
+stdenv.mkDerivation(finalAttrs: {
+  pname = "git";
+  version = "...";
+  # ...
+
+  installCheckPhase = ''
```
raboof, line 141: also add `doInstallCheck = true;` here as it is disabled by default

raboof (approved):
The general sentiment makes sense to me: this test is useful and not resource-intensive, so it seems helpful to execute it as part of the 'main' build.

This begs the question whether we should remove `testVersion`, as that now seems duplicate. I think it still makes sense to keep it, to catch issues where the program incorrectly relies on something that is only available in the build environment.

Added a couple of smaller suggestions.

doc/stdenv/meta.chapter.md
doronbehar: > this style seems less common than `mkDerivation rec {`, is there a reason we'd want to show `finalAttrs` here?

using `finalAttrs` is the modern way to `mkDerivation` nowadays, as it makes overriding the package much easier when the package references it self, and especially if you want to make it easy for the user reading this to maybe add a `passthru.tests.comprehensive` that will use `finalAttrs.finalPackage` which is not possible with the old `mkDerivation rec {` pattern.

doc/stdenv/meta.chapter.md
doronbehar: > also add `doInstallCheck = true;` here as it is disabled by default

Thanks!

doc/stdenv/meta.chapter.md
```diff
@@ -125,17 +125,38 @@ $ cd path/to/nixpkgs
...
+```
 
-Prefer `passthru.tests` for tests that are introduced in nixpkgs because:
+Most programs distributed by Nixpkgs support such a `--version` flag, and it can help give confidence that the package at least got compiled properly. However, tests that are slightly non trivial will better fit into `passthru.tests`, because:
 
 * `passthru.tests` tests the 'real' package, independently from the environment in which it was built
-* we can run `passthru.tests` independently
-* `installCheckPhase` adds overhead to each build
+* We can run and debug a `passthru.tests` independently, after the package was built (useful if it takes a long time).
+* `installCheckPhase` adds an overhead to each build.
```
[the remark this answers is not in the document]
doronbehar: > it looks like we have no general consistency on whether items in bulleted lists end with a '.', but at least let's keep it consistent within the list ;)

I agree :).

doronbehar (commented):
> The general sentiment makes sense to me: this test is useful and not resource-intensive, so it seems helpful to execute it as part of the 'main' build.

Great I'm glad you agree.

> This begs the question whether we should remove `testVersion`, as that now seems duplicate. I think it still makes sense to keep it, to catch issues where the program incorrectly relies on something that is only available in the build environment.

Definitely I don't think the right thing to do now is to iterate all usages of `testVersion` and replace them with `installCheckPhase` - that would create too many rebuilds and it needs to be tested manually. Perhaps though, we can only remove the documentation? Which is here BTW:

https://github.com/NixOS/nixpkgs/blob/340d46ceac0640399b3618b52874ce6f5dbf30cf/doc/build-helpers/testers.chapter.md?plain=1#L119-L166

doc/stdenv/meta.chapter.md
raboof: > using `finalAttrs` is the modern way to `mkDerivation` nowadays

Do you have a reference for that? https://nixos.org/manual/nixpkgs/unstable/#sec-using-stdenv is still showing `rec`

> it makes overriding the package much easier when the package references it self, and especially if you want to make it easy for the user reading this to maybe add a `passthru.tests.comprehensive` that will use `finalAttrs.finalPackage` which is not possible with the old `mkDerivation rec {` pattern

TIL, very cool

doronbehar:
> > Perhaps though, we can only remove the documentation?
> 
> I think this test still has value, so I don't think we should remove it or the documentation.

OK, another option is to slightly modify the introduction to it, such that it will mention `installCheckPhase`, and how that is preferred for most cases?

raboof:
> > > Perhaps though, we can only remove the documentation?
> > 
> > I think this test still has value, so I don't think we should remove it or the documentation.
> 
> OK, another option is to slightly modify the introduction to it, such that it will mention `installCheckPhase`, and how that is preferred for most cases?

I don't think I agree `installCheckPhase` should replace `testVersion`, having both makes sense to me. I appreciate you might feel differently - perhaps let's just keep that part as it is for now?

doronbehar:
> having both makes sense to me.

It makes sense to me too. However I felt we lack some consistency here if this part of the docs doesn't mention the other and vise versa. I added a sentence here that mentions `testVersion`, with a link.

doc/stdenv/meta.chapter.md
raboof: Some more ongoing discussion on this in #315337

raboof (approved):

doc/stdenv/meta.chapter.md
```diff
@@ -125,17 +125,41 @@ $ cd path/to/nixpkgs
...
+  version = "...";
+  # ...
+
+  doInstallCheck = true;
+  installCheckPhase = ''
+    runHook preInstallCheck
 
-Prefer `passthru.tests` for tests that are introduced in nixpkgs because:
+    echo checking if 'git --version' mentions ${finalAttrs.version}
+    $out/bin/git --version | grep ${finalAttrs.version}
```
Atemu, line 146: It'd be great if this could simply be a hook which would be configured using a similar interface as the tester.

Artturin, line 146: This doesn't show the whole output on failure, nor is it separated from the build environment, therefore it is inferior to `testVersion`

PR #315643, opened 2024-05-29
vscode-extensions: deprecate aliases

superherointj:
vscode-extensions: deprecate aliases


superherointj:
> why did you decide that aliases are no longer needed?

These aliases are (very) old and deprecated. This should have been done much sooner. Ideally each package should have only a single way of referencing it and it should be standardized.

superherointj:
ping @Garmelon

PR #315120, opened 2024-05-27
[24.05] Revert "nixos/steam: add option fontPackages"

eclairevoyant:
## Description of changes

This reverts commit 7f86c8842a5d4686b86bded1c29e2ef22304eab3, which was a breaking change (see https://github.com/NixOS/nixpkgs/pull/312268#issuecomment-2133641945) and therefore not eligible for backporting.

Original PR #312268
Original backport #314970

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
- For non-Linux: Is sandboxing enabled in `nix.conf`? (See [Nix manual](https://nixos.org/manual/nix/stable/command-ref/conf-file.html))
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:
  - [NixOS test(s)](https://nixos.org/manual/nixos/unstable/index.html#sec-nixos-tests) (look inside [nixos/tests](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests))
  - and/or [package tests](https://github.com/NixOS/nixpkgs/blob/master/pkgs/README.md#package-tests)
  - or, for functions and "core" functionality, tests in [lib/tests](https://github.com/NixOS/nixpkgs/blob/master/lib/tests) or [pkgs/test](https://github.com/NixOS/nixpkgs/blob/master/pkgs/test)
  - made sure NixOS tests are [linked](https://nixos.org/manual/nixpkgs/unstable/#ssec-nixos-tests-linking) to the relevant packages
- [ ] Tested compilation of all packages that depend on this change using `nix-shell -p nixpkgs-review --run "nixpkgs-review rev HEAD"`. Note: all changes have to be committed, also see [nixpkgs-review usage](https://github.com/Mic92/nixpkgs-review#usage)
- [ ] Tested basic functionality of all binary files (usually in `./result/bin/`)
- [24.11 Release Notes](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2411.section.md) (or backporting [23.11](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2311.section.md) and [24.05](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2405.section.md) Release notes)
  - [ ] (Package updates) Added a release notes entry if the change is major or breaking
[...]

PR #315674, opened 2024-05-29
[24.05] Revert "[24.05] Revert "nixos/steam: add option fontPackages"", backport #315149

Cryolitia:
Reverts NixOS/nixpkgs#315120
Backport #315149

In summary, add `programs.steam.fontPackages` option for 24.05 and fix the evaluation error in the previous backport. I would like to fix the CJK problem on 24.05.
The type should be stable this time.

### DIFF
--- a/nixos/modules/programs/steam.nix
+++ b/nixos/modules/programs/steam.nix
@@ -100,6 +100,18 @@ in {
       '';
     };
 
+    fontPackages = lib.mkOption {
+      type = lib.types.listOf lib.types.package;
+      default = config.fonts.packages;
+      defaultText = lib.literalExpression "fonts.packages";
+      example = lib.literalExpression "with pkgs; [ source-han-sans ]";
+      description = ''
+        Font packages to use in Steam.
+
+        Defaults to system fonts, but could be overridden to use other fonts — useful for users who would like to customize CJK fonts used in Steam. According to the [upstream issue](https://github.com/ValveSoftware/steam-for-linux/issues/10422#issuecomment-1944396010), Steam only follows the per-user fontconfig configuration.
+      '';
+    };
+
     remotePlay.openFirewall = lib.mkOption {
       type = lib.types.bool;
       default = false;
@@ -171,6 +183,8 @@ in {
         setuid = true;
       };
     };
+
+    programs.steam.extraPackages = cfg.fontPackages;
 
     programs.gamescope.enable = lib.mkDefault cfg.gamescopeSession.enable;
     services.displayManager.sessionPackages = lib.mkIf cfg.gamescopeSession.enable [ gamescopeSessionFile ];
--- a/nixos/modules/programs/steam.nix
+++ b/nixos/modules/programs/steam.nix
@@ -102,8 +102,9 @@ in {
 
     fontPackages = lib.mkOption {
       type = lib.types.listOf lib.types.package;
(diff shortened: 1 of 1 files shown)


nixos/modules/programs/steam.nix
```diff
@@ -172,6 +185,8 @@ in {
       };
     };
 
+    programs.steam.extraPackages = cfg.fontPackages;
```
wegank, line 188: Isn't this a breaking change?

jian-lin (changes requested):
For the second commit, please use `git cherry-pick -x` to reference the backported commit .  Here is the [doc][1].

[1]: https://github.com/NixOS/nixpkgs/blob/master/CONTRIBUTING.md#manually-backporting-changes

nixos/modules/programs/steam.nix
jian-lin: Why do you think so?

Cryolitia: It should fix the CJK font display on Steam and introduce no other side effects. Whether to treat it as a breaking or fix commit depends on you. I just open it to discuss.

Cryolitia:
> For the second commit, please use `git cherry-pick -x` to reference the backported commit .  Here is the [doc][1].
> 
> [1]: https://github.com/NixOS/nixpkgs/blob/master/CONTRIBUTING.md#manually-backporting-changes

done

nixos/modules/programs/steam.nix
wegank: > Why do you think so?

Previously, `extraPackages` defaulted to `[ ]`.

https://github.com/NixOS/nixpkgs/blob/daa52e1d9e89a4df5b844d26225db10fc9475739/nixos/modules/programs/steam.nix#L75-L77

And now you have `programs.steam.extraPackages = cfg.fontPackages;` in addition. It's not breaking if `fontPackages` also defaults to `[ ]`, but it effectively has the value of `config.fonts.packages` instead. So, 

> It should fix the CJK font display on Steam and introduce no other side effects.

This will not fix the display if no CJK font is installed system-wide. But in addition, there is the side effect of having all fonts passed via `extraPkgs` to `steamPackages.steam-fhsenv` on every installation with Steam enabled.

jian-lin: I still cannot see why the default non-empty `cfg.fontPackages` is a breaking change.  Since the type of `programs.steam.extraPackages` is a list, the value of `programs.steam.extraPackages` is the merge (concat) result of newly added `cfg.fontPackages` and users' config.

jian-lin (approved):
LGTM

nit:

Your first commit message is a bit weird.  It contains unneeded `[24.05]` and is not what `git revert` generates.

I think [this][1] is a good example of what the "revert-revert" commit message should look like.  It references the reverted commit and adds a reason for the revert.

[1]: https://github.com/NixOS/nixpkgs/commit/1284b4f7fa36a3e4e7f05a979f56136277ac8983

nixos/modules/programs/steam.nix
jian-lin: Ah, I see.  This may change the font steam uses.  If the previous steam font is fine for users, this can be seen as a breaking change.

Cryolitia: Feel free to close it if I'm wrong

PR #34636, opened 2018-02-05
Switch to Nix 2.0

edolstra:
Hopefully we can switch to Nix 2.0 for NixOS 18.03. Nix 2.0 should be released in the next few weeks. This PR makes 2.0 the default (by temporarily pointing `nix` to `nixUnstable`) and updates `nixos-install` to work with it. It also adds a command `nixos-enter` that factors out the functionality of `nixos-install --chroot`.

PR #315857, opened 2024-05-30
nixos/tests/misc: support old and new path-info output structure

mweinelt:
## Description of changes

https://github.com/NixOS/nixpkgs/pull/315262#issuecomment-2139134458

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/nixos/tests/misc.nix
+++ b/nixos/tests/misc.nix
@@ -73,16 +73,19 @@ let
 
 
         with subtest("nix-db"):
-            info = get_path_info("${foo}")
+            out = "${foo}"
+            info = get_path_info(out)
             print(info)
 
+            pathinfo = info[0] if isinstance(info, list) else info[out]
+
             if (
-                info[0]["narHash"]
+                pathinfo["narHash"]
                 != "sha256-BdMdnb/0eWy3EddjE83rdgzWWpQjfWPAj3zDIFMD3Ck="
             ):
                 raise Exception("narHash not set")
 
-            if info[0]["narSize"] != 128:
+            if pathinfo["narSize"] != 128:
                 raise Exception("narSize not set")
 
         with subtest("nixos-version"):


fricklerhandwerk:
@nixos/nix-team Why in the world is Nixpkgs even testing against Nix experimental commands? Looking at [commit history](https://github.com/NixOS/nixpkgs/blame/a047bae9f3f9a0f3a00964f1b6bbfef61485880d/nixos/tests/misc.nix), this has been added before experimental features were a thing. I suggest removing that test and only adding it back when the command is stabilized.

K900:
Going to untag as channel-blocker as the Nix bump was reverted.

Mic92:
> @NixOS/nix-team Why in the world is Nixpkgs even testing against Nix experimental commands? That's exactly how you make them unchangeable or annoy everyone with breakages.
> 
> Looking at [commit history](https://github.com/NixOS/nixpkgs/blame/a047bae9f3f9a0f3a00964f1b6bbfef61485880d/nixos/tests/misc.nix), this has been added before experimental features were a thing. I suggest removing that test and only adding it back when the command is stabilized.

I think we are beyond a point where we can reasonable treat these commands as experimental. Even if it doesn't break in nixpkgs, it will still break other users, since these commands have been cooking for too long in an "experimental" state.

mweinelt:
If it broke, then yes, `passthru.tests` seems appropriate.  Naming the test `misc` :shrug:

Mic92:
Yeah. Terrible name. I will come up with something else.

lilyinstarlight:
> I think we are beyond a point where we can no longer reasonable treat these commands as experimental. Even if it doesn't break in nixpkgs, it will still break other users, since these commands have been cooking for too long in an "experimental" state.

This command has been present since Nix 2.0 and was added to the NixOS misc tests during the Nix 2.0 bump PR (<https://github.com/NixOS/nixpkgs/pull/34636/files#diff-5eb2258233433ba8710976a266f30f16a70d92c633d36706072a095f91dc50a9R38>)

I'm just providing context, though, so I won't make any comment either way about whether it's appropriate the command was later made "experimental" when that ability was added (like the `nix repl` command) or whether it should be removed from the test until the command is "stabilized"

Mic92:
Tested that this, fixes the nix upgrade.

Ericson2314:
> I currently don't see a good replacement for this command in the old cli interface.

There is a mishmash of `nix-store` commands to query individual fields of the path info JSON. Agreed that is much less nice to use! 

> If we make sure that we run this NixOS test on every nix version bump, than we can also fix it.

That's true. If we the tests make clear they are not locking down features users should rely on, but just sanity checking Nixpkgs' own internal usage of experimental stuff, it is less bad.

PR #277626, opened 2023-12-29
nixos/pam: Use secure default for `sshAgentAuth.authorizedKeysFiles`

nicoonoclaste:
Follow-up on #277620, blocked until it is merged.

## Description of changes

Change the default value of `security.pam.sshAgentAuth.authorizedKeysFiles` to `[ "/etc/ssh/authorized_keys.d/%u" ]`.
This breaks backwards-compatibility to fix a security vulnerability: #31611

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
- [ ] Tested, as applicable:
  - `nixosTests.ssh-agent-auth` once #266332 is merged.
- [24.05 Release Notes](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2405.section.md)
  - [x] (Module updates) Added a release notes entry if the change is significant
- [x] Fits [CONTRIBUTING.md](https://github.com/NixOS/nixpkgs/blob/master/CONTRIBUTING.md).

---

Add a :+1: [reaction] to [pull requests you find important].

[reaction]: https://github.blog/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/
[pull requests you find important]: https://github.com/NixOS/nixpkgs/pulls?q=is%3Aopen+sort%3Areactions-%2B1-desc

PR #315874, opened 2024-05-30
nixos/gnome-keyring: rewrite module and fix unlocking on GDM session login

JohnRTitor:
## Description of changes

This PR rewrites the `nixos/gnome/gnome-keyring` module.

GDM uses gdm-password as the PAM service name for both logins and unlocks.
So unlock gnome-keyring as part of `gdm-password`.
Without this, keyrings may not be unlocked properly for GDM 45+.

## Things done

- [ ] Tested, as applicable:

---

@NixOS/gnome

---

### DIFF
--- a/nixos/modules/services/desktops/gnome/gnome-keyring.nix
+++ b/nixos/modules/services/desktops/gnome/gnome-keyring.nix
@@ -43,6 +43,9 @@ in
         gdm-password.enableGnomeKeyring = true;
         gdm-autologin.enableGnomeKeyring = true;
       })
+      (lib.mkIf (config.services.xserver.displayManager.gdm.enable && config.services.fprintd.enable) {
+        gdm-fingerprint.enableGnomeKeyring = true;
+      })
     ];
 
     security.wrappers.gnome-keyring-daemon = {
--- a/nixos/modules/services/desktops/gnome/gnome-keyring.nix
+++ b/nixos/modules/services/desktops/gnome/gnome-keyring.nix
@@ -1,41 +1,37 @@
 # GNOME Keyring daemon.
 
-{ config, pkgs, lib, ... }:
-
+{
+  config,
+  pkgs,
+  lib,
+  ...
+}:
+let
+  cfg = config.services.gnome.gnome-keyring;
+in
 {
 
   meta = {
     maintainers = lib.teams.gnome.members;
   };
 
-  ###### interface
-
   options = {
-
     services.gnome.gnome-keyring = {
-
-      enable = lib.mkOption {
-        type = lib.types.bool;
-        default = false;
-        description = ''
-          Whether to enable GNOME Keyring daemon, a service designed to
-          take care of the user's security credentials,
-          such as user names and passwords.
-        '';
-      };
-
+      enable = lib.mkEnableOption ''
+        GNOME Keyring daemon, a service designed to
+        take care of the user's security credentials,
+        such as user names and passwords
+      '';
     };
-
   };
 
-
-  ###### implementation
-
-  config = lib.mkIf config.services.gnome.gnome-keyring.enable {
-
+  config = lib.mkIf cfg.enable {
     environment.systemPackages = [ pkgs.gnome.gnome-keyring ];
 
-    services.dbus.packages = [ pkgs.gnome.gnome-keyring pkgs.gcr ];
+    services.dbus.packages = [
+      pkgs.gnome.gnome-keyring
+      pkgs.gcr
+    ];
 
     xdg.portal.extraPortals = [ pkgs.gnome.gnome-keyring ];
 
@@ -47,7 +43,5 @@
       capabilities = "cap_ipc_lock=ep";
       source = "${pkgs.gnome.gnome-keyring}/bin/gnome-keyring-daemon";
     };
-
   };
-
 }
--- a/nixos/modules/services/desktops/gnome/gnome-keyring.nix
+++ b/nixos/modules/services/desktops/gnome/gnome-keyring.nix
@@ -35,7 +35,15 @@ in
 
     xdg.portal.extraPortals = [ pkgs.gnome.gnome-keyring ];
 
-    security.pam.services.login.enableGnomeKeyring = true;
+    security.pam.services = lib.mkMerge [
+      {
+        login.enableGnomeKeyring = true;
+      }
+      (lib.mkIf config.services.xserver.displayManager.gdm.enable {
+        gdm-password.enableGnomeKeyring = true;
+        gdm-autologin.enableGnomeKeyring = true;
+      })
+    ];
 
     security.wrappers.gnome-keyring-daemon = {
       owner = "root";


JohnRTitor:
Building NixOS manual check [fails](https://github.com/NixOS/nixpkgs/actions/runs/9302524845/job/25602901205?pr=315874), any idea why?

jtojnar:
Docs do not allow `pkgs` references since they would prevent the docs from being cacheable. You need to include `defaultText` for the manual when you use a restricted value in `default`.

nixos/modules/services/desktops/gnome/gnome-keyring.nix
```diff
@@ -1,53 +1,53 @@
...
-      enable = lib.mkOption {
-        type = lib.types.bool;
-        default = false;
-        description = ''
-          Whether to enable GNOME Keyring daemon, a service designed to
-          take care of the user's security credentials,
-          such as user names and passwords.
-        '';
-      };
+  ## Interface ##
```
jtojnar, line 16: We are removing these comments in GNOME modules when refactoring, since they are redundant with `options` and `config` attributes. We still use `options = {}` and `meta` unpacked to make the sections more visible.

nixos/modules/services/desktops/gnome/gnome-keyring.nix
```diff
@@ -1,53 +1,52 @@
...
 
+  options.services.gnome.gnome-keyring = {
+    enable = lib.mkEnableOption null // {
+      default = false;
```
jtojnar, line 20: This is the default `default` for `mkEnableOption`.

Why not just use

```nix
lib.mkEnableOption ''
        GNOME Keyring daemon, a service designed to
        take care of the user's security credentials,
        such as user names and passwords
''

File: nixos/modules/services/desktops/gnome/gnome-keyring.nix [hunk 3]
```diff
@@ -1,53 +1,52 @@
...
+      description = ''
+        Whether to enable GNOME Keyring daemon, a service designed to
+        take care of the user's security credentials,
+        such as user names and passwords.
+      '';
     };
-
+    package = lib.mkPackageOption pkgs "gnome-keyring" { default = pkgs.gnome.gnome-keyring; };
```
jtojnar, line 27: Is there as specific reason for adding the package option? Are users expected to change it? If it is just for debugging, maybe overlays would be sufficient?

jtojnar:
> GDM uses gdm-password as the PAM service name for both logins and unlocks.
> So unlock gnome-keyring as part of `gdm-password`.
> Without this, keyrings may not be unlocked properly for GDM 45+.

Do you have some tips for reproducing this? Keyring unlocks fine for me on 46.

---

The whole GDM PAM setup would deserve a re-do but PAM feels like black magic, and the man page probably contains a curse that erases my memory every time I look at it.

nixos/modules/services/desktops/gnome/gnome-keyring.nix
JohnRTitor: Overlays are pretty complicated for new users. I added the option in case anyone wants to override the gnome-keyring package like use an older version, or patch it themselves.

This package option helps sort that out without resorting to overlays, which can actually rebuild the whole system sometimes (may be not for this package). Overlays are known to cause problems.

As for the `gcrPackage` option, I thought I would use `gcr_4` with this. I assume `gcr` ver 3 is still the default due to compatibility issues but for GNOME 46, shouldn't `gcr_4` be the default?

JohnRTitor:
> Do you have some tips for reproducing this? Keyring unlocks fine for me on 46.

I am not actually using GNOME 46. My setup is Hyprland window manager with a set of GNOME packages.
Keyrings are not automatically unlocked for me and I have to do that manually unless I use this option.

Other users on NixOS discourse have noticed this issue as well. See https://discourse.nixos.org/t/login-keyring-did-not-get-unlocked-hyprland/40869/11

nixos/modules/services/desktops/gnome/gnome-keyring.nix
jtojnar: Well, new users should not need to change packages under modules in the first place. If there is a bug, we should fix it in Nixpgks. And the mass rebuilds are not an issue with leaf packages like GNOME Keyring.

`gcr` contains two D-Bus services that were removed in [GCR 4](https://gitlab.gnome.org/GNOME/gcr/-/commit/f316c8919bfcbe2534271e9729084342db3cc9c6). AFAICT, it does not even make sense to install that on GNOME Shell, since it provides its own implementation.

Looks like `gcr` installation was introduced in 2d280840f844cad6c9478bc9107b02cff82b1c34 to provide fallback prompter for GNOME Keyring (well, it uses the prompter through `gcr` library). But that should probably go into its own module, as GNOME Keyring is [not the only software that expects this](https://codesearch.debian.net/search?q=org.gnome.keyring.SystemPrompter&literal=1).

JohnRTitor: Done. Removed `package` and `gcrPackage` option.

nixos/modules/services/desktops/gnome/gnome-keyring.nix
```diff
@@ -1,53 +1,48 @@
...
-          Whether to enable GNOME Keyring daemon, a service designed to
-          take care of the user's security credentials,
-          such as user names and passwords.
-        '';
-      };
-
+      enable = lib.mkEnableOption ''
+        GNOME Keyring daemon, a service designed to
+        take care of the user's security credentials,
+        such as user names and passwords.
```
jtojnar, line 23: mkEnableOption already includes period

nixos/modules/services/desktops/gnome/gnome-keyring.nix
```diff
@@ -1,53 +1,48 @@
...
 
+  config = lib.mkIf cfg.enable {
+    environment.systemPackages = with pkgs; [ gnome.gnome-keyring ];
 
-  ###### implementation
-
-  config = lib.mkIf config.services.gnome.gnome-keyring.enable {
-
-    environment.systemPackages = [ pkgs.gnome.gnome-keyring ];
+    services.dbus.packages = with pkgs; [
```
jtojnar, line 31: `with` makes reference checking lazy so it can hide errors. I would avoid it here since it does not add much benefit here and actually makes the code slightly longer.

nixos/modules/services/desktops/gnome/gnome-keyring.nix
```diff
@@ -36,6 +36,7 @@ in
     xdg.portal.extraPortals = with pkgs; [ gnome.gnome-keyring ];
 
     security.pam.services.login.enableGnomeKeyring = true;
+    security.pam.services.gdm-password.enableGnomeKeyring = config.services.xserver.displayManager.gdm.enable;
```
jtojnar, line 39: Arch adds this as well so it is probably fine https://gitlab.gnome.org/GNOME/gdm/-/blob/main/data/pam-arch/gdm-password.pam#L4

JohnRTitor: Nice. But I believe this should only be enabled if GDM is used?

Some users (without GNOME) may prefer another login manager such as SDDM, in which case `login` PAM should work.

JohnRTitor: I enabled gnome-keyring for `autologin`, `fingerprint` and `smartcard` authentication mode too.

jtojnar: Right, I just meant that it should be okay to include keyring in `gdm-password`. Will need to double check the other auth modes, in case their presence signals specific behavour to GDM (we previously did not have those files).

JohnRTitor: I can't test fingerprint and smartcard, but autologin seems fine.

JohnRTitor:
@jtojnar Would love to see this patch in 24.05 too since it's also broken there. 

Also, have you checked other authentication modes such as fingerprint and smartcard? 

I don't think these should cause any problems. Arch seems to run this just fine.

https://gitlab.gnome.org/GNOME/gdm/-/blob/main/data/pam-arch/gdm-fingerprint.pam#L10
https://gitlab.gnome.org/GNOME/gdm/-/blob/main/data/pam-arch/gdm-smartcard.pam#L10

jtojnar:
> Also, have you checked other authentication modes such as fingerprint and smartcard?
> 
> I don't think these should cause any problems. Arch seems to run this just fine.

I meant that we do not have those files currently. And adding `security.pam.services.gdm-fingerprint.enableGnomeKeyring = true;` creates the file with only some minimal rules.

I am not confident enough with PAM to be able to predict the effects it will have.

Wanted to look deeper into PAM this weekend but stuff got in the way. Finding someone who understands PAM and/or GDM for review would make this faster.

> Would love to see this patch in 24.05 too since it's also broken there.

Probably would want to run it in unstable for a bit to catch possible issues.

JohnRTitor:
> Wanted to look deeper into PAM this weekend but stuff got in the way. Finding someone who understands PAM and/or GDM for review would make this faster.

Sure but I am not sure who to tag, but probably @nbraud and @LeSuisse who were involved with the last PR to `nixos/pam` (#277626),

> Probably would want to run it in unstable for a bit to catch possible issues.

Sure.

nixos/modules/services/desktops/gnome/gnome-keyring.nix
```diff
@@ -1,53 +1,54 @@
...
+      pkgs.gnome.gnome-keyring
+      pkgs.gcr
+    ];
 
     xdg.portal.extraPortals = [ pkgs.gnome.gnome-keyring ];
 
-    security.pam.services.login.enableGnomeKeyring = true;
+    security.pam.services = {
+      login.enableGnomeKeyring = true;
+    } // lib.optionalAttrs config.services.xserver.displayManager.gdm.enable {
```
SuperSandro2000, line 40: Shouldn't we use mkMerge here to leverage the module system?

JohnRTitor: Is there any benefits in that? I did this way as this looks cleaner.

JohnRTitor: @SuperSandro2000 Done. Please check.

JohnRTitor:
I feel like getting this to be reviewed and tested will take a lot of time. @jtojnar so can we just drop all changes to PAM except for gdm-password, and get this merged?

nixos/modules/services/desktops/gnome/gnome-keyring.nix
```diff
@@ -1,53 +1,57 @@
...
-    security.pam.services.login.enableGnomeKeyring = true;
+    security.pam.services = lib.mkMerge [
+      {
+        login.enableGnomeKeyring = true;
+      }
+      (lib.mkIf config.services.xserver.displayManager.gdm.enable {
+        gdm-password.enableGnomeKeyring = true;
+        gdm-autologin.enableGnomeKeyring = true;
+        gdm-fingerprint.enableGnomeKeyring = true;
+        gd-smartcard.enableGnomeKeyring = true;
```
LeSuisse, line 46:
```suggestion
        gdm-smartcard.enableGnomeKeyring = true;
```

Also, it might reasonable to not set it right now, there is some discussions about `gdm-smartcard` and unwanted interaction with Yubikeys (seems to concern at least some Arch and Debian users).


https://gitlab.gnome.org/GNOME/gdm/-/issues/877

nixos/modules/services/desktops/gnome/gnome-keyring.nix
```diff
@@ -1,53 +1,57 @@
...
 
-    security.pam.services.login.enableGnomeKeyring = true;
+    security.pam.services = lib.mkMerge [
+      {
+        login.enableGnomeKeyring = true;
+      }
+      (lib.mkIf config.services.xserver.displayManager.gdm.enable {
+        gdm-password.enableGnomeKeyring = true;
+        gdm-autologin.enableGnomeKeyring = true;
+        gdm-fingerprint.enableGnomeKeyring = true;
```
LeSuisse, line 45: It might be reasonable to only enable this automatically when `config.services.fprintd.enable` is enabled. It is unlikely to be useful in other scenario (or in the worst case the users can still enable it themselves like today).

nixos/modules/services/desktops/gnome/gnome-keyring.nix
JohnRTitor: Sure. I have removed the smartcard bit. Can be enabled later if proven stable.

nixos/modules/services/desktops/gnome/gnome-keyring.nix
JohnRTitor: Done, anything else I should check for?

nixos/modules/services/desktops/gnome/gnome-keyring.nix
```diff
@@ -1,53 +1,56 @@
...
-    security.pam.services.login.enableGnomeKeyring = true;
+    security.pam.services = lib.mkMerge [
+      {
+        login.enableGnomeKeyring = true;
+      }
+      (lib.mkIf config.services.xserver.displayManager.gdm.enable {
+        gdm-password.enableGnomeKeyring = true;
+        gdm-autologin.enableGnomeKeyring = true;
+        gdm-fingerprint.enableGnomeKeyring = config.services.fprintd.enable;
+      })
```
LeSuisse, line 46:
```suggestion
      (lib.mkIf config.services.xserver.displayManager.gdm.enable {
        gdm-password.enableGnomeKeyring = true;
        gdm-autologin.enableGnomeKeyring = true;
      })
      (lib.mkIf (config.services.xserver.displayManager.gdm.enable && config.services.fprintd.enable) {
        gdm-fingerprint.enableGnomeKeyring = true;
      })
```

That will avoid the creation of a `gdm-fingerprint` pam config with a default deny if we do not need it.

JohnRTitor: Done.

JohnRTitor:
@LeSuisse can we merge this?

As for fingerprint authentication, I plan to cherry-pick https://github.com/NixOS/nixpkgs/pull/306338 in a different PR

LeSuisse (approved):
Yum the pam config looks reasonable to me at this stage

PR #315908, opened 2024-05-30
nixos/stalwart-mail: rocksdb as default storage

pacien:
## Description of changes

This sets RocksDB as the default storage backend for `stateVersion` >=
24.11. For previous `stateVersion`s, the structured data and blobs
remain on SQLite and the filesystem respectively.

This is closer to the suggested upstream configuration for fully local
storage.

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [x] Tested, as applicable:

---

### DIFF
--- a/nixos/doc/manual/release-notes/rl-2411.section.md
+++ b/nixos/doc/manual/release-notes/rl-2411.section.md
@@ -23,6 +23,10 @@
   before changing the package to `pkgs.stalwart-mail` in
   [`services.stalwart-mail.package`](#opt-services.stalwart-mail.package).
 
+- The `stalwart-mail` module now uses RocksDB as the default storage backend
+  for `stateVersion` ≥ 24.11. (It was previously using SQLite for structured
+  data and the filesystem for blobs).
+
 ## Other Notable Changes {#sec-release-24.11-notable-changes}
 
 <!-- To avoid merge conflicts, consider adding your item at an arbitrary place in the list instead. -->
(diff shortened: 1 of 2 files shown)


Shawn8901:
is there a specific reason that we are preventing users easily to upgrade to rocksdb with `stateVersion` older than 24.11? 
I mean preventing an accidental upgrade via `stateVersion` is absolute reasonable, but atm its not possible to use rocksdb when having a machine that is older. As we do not expect people to change their `stateVersion`, i have the feeling we are missing an option to flip to no longer use the legacy storage.

Shawn8901:
I had problems setting this as it still tried to set up the legacy blob storage (as its defined in the module like that) and also the preStart does not work correctly in such cases.

I would have expected something like that 
```nix
    useLegacyStorage = mkOption {
      type = types.bool;
      description = ''
        Defines if the legacy storage should be used instead rocksdb
      '';
      default = versionOlder config.system.stateVersion "24.11";
    };
```

Which would to basically the same, but one could override that externally, in case one knows better, thats the reason why i am asking, if there was a specific reason not to use an option for that.

Is there a way to prevent the modules setting of the legacy storage with your code that i dont see atm?

Shawn8901:
Yeah the option would definitely not ideal, but as stalwart itself is not very stable yet, i personally would be fine in having breaking changes in the module, when its easier to migrate for the user(s).
 
but thx for the feedback. Ill try a bit more around, possibly i get it working anyways with the upstream (so nixpkgs) module.

PR #315909, opened 2024-05-30
doc: add stdenv passthru chapter

abathur:
Reviewer on another PR queried about documenting a passthru attribute it adds, and looking into it led me to see how thin and scattered existing passthru docs are.

I'm hoping to keep this fairly narrow (reworking the skeleton and moving existing coverage) to focus review on what I see as the big/fundamental questions here:
- whether we should have such a section
- whether that section should list common passthru attrs
- what razors we should apply to decide when a given passthru attr deserves mention here as opposed to a more-specific part of the manual, an in-repo README, code comments, etc.

@NixOS/documentation-team and @infinisil (given ~ownership of repo-internal READMEs, where a large fraction of the more in-depth passthru.tests coverage was moved last year).

---


doc/stdenv/passthru.chapter.md
```diff
@@ -0,0 +1,253 @@
...
+  };
+}
+```
+
+Much of the passthru's power comes from two features:
+- You can access passthru attributes as if they were set inside a derivation (e.g. `hello.baz.value1`).
+- These values are not passed to the builder, so you can change them without triggering a rebuild.
+
+One example of the passthru's power is how it enables us to associate the package for an interpreter Python or Ruby with a set of modules built for that specific interpreter (e.g., `python.pkgs` and `ruby.gems`) and to build runtime environments with a collection of those modules (e.g., (`python.withPackages`)[#python.withpackages-function] and `ruby.withPackages`).
+<!-- python.pkgs, ruby.gems, and ruby.withPackages are mentioned but don't have dedicated sections to link -->
```
abathur, line 24: Comment here is just a reminder for why I only linked one of these :)

doc/stdenv/passthru.chapter.md
```diff
@@ -0,0 +1,253 @@
...
+
+::: {.note}
+Since `passthru` is meant for values that are useful outside the derivation (e.g. in other derivations), it has no strict schema.
+:::
+
+<!-- 
+I'm extracting this statement from the old text because I feel like it isn't very clearly-phrased and doesn't help connect with something most people know already.
+I still think we should try to make sure the pattern it models is represented in the list of common attributes, though:
+
+convey a specific dependency of your derivation which contains a program with plugins support.
```
abathur, line 34: Still undecided on this. Roughly a TODO, but I didn't want to go off searching for a real example of this and writing it up if all of this is off-base in the first place.

doc/stdenv/passthru.chapter.md
```diff
@@ -0,0 +1,253 @@
...
+Since `passthru` is meant for values that are useful outside the derivation (e.g. in other derivations), it has no strict schema.
+:::
+
+<!-- 
+I'm extracting this statement from the old text because I feel like it isn't very clearly-phrased and doesn't help connect with something most people know already.
+I still think we should try to make sure the pattern it models is represented in the list of common attributes, though:
+
+convey a specific dependency of your derivation which contains a program with plugins support. 
+Later, others who make derivations with plugins can use passed-through dependency to ensure that their plugin would be binary-compatible with built program. 
+-->
```
abathur, line 36: Still undecided on this. Roughly a TODO, but I didn't want to go off searching for a real example of this and writing it up if all of this is off-base in the first place.

pkgs/development/mobile/androidenv/README.md
```diff
@@ -6,7 +6,7 @@
 4. Update the relevant argument defaults in `compose-android-packages.nix`
 
 # How to run tests
-You may need to make yourself familiar with [tests](https://nixos.org/manual/nixpkgs/stable/#var-meta-tests), and [Writing larger package tests](https://nixos.org/manual/nixpkgs/stable/#ssec-package-tests-writing) in the Manual, then run tests locally with:
+You may need to make yourself familiar with [package tests](https://github.com/NixOS/nixpkgs/blob/master/pkgs/README.md#package-tests), and [Writing larger package tests](https://github.com/NixOS/nixpkgs/blob/master/pkgs/README.md#writing-larger-package-tests), then run tests locally with:
```
hsjobeki, line 9: Those kind of links are brittle. Because it doesn't link to a specific commit it can happen that somebody moves or renames the heading in that file and will break the link without anyone noticing.

doc/stdenv/passthru.chapter.md
```diff
@@ -0,0 +1,253 @@
...
+      value1 = 4;
+      value2 = 5;
+    };
+  };
+}
+```
+
+Much of the passthru's power comes from two features:
+- You can access passthru attributes as if they were set inside a derivation (e.g. `hello.baz.value1`).
+- These values are not passed to the builder, so you can change them without triggering a rebuild.
```
AndersonTorres, line 21: A  better wording:

```suggestion
The interesting properties of `passthru` are:
- Similar to `meta` attributes, they convey information about the derivation without triggering a rebuild.
- Attributes inside `passthru` are directly accessible as if they were set inside the derivation. E.g. in the example above, `hello.baz.value1 + hello.baz.value2` returns `9`.

The properties above make `passthru` a powerful tool to provide information about the derivation to the outside world.
```

pkgs/development/mobile/androidenv/README.md
abathur: Thanks for poking at this. Forgot to leave a comment on it.

Roughly intending these two link changes as a query to @infinisil about what internal READMEs should be doing here. 

- The first link could technically keep pointing to the manual, but the existing links from the repo to the stable manual basically mean that I can't update the underlying anchors that the README links use without breaking them until the stable manual catches up. 

    (Maybe it's fine to change it to change it to the unstable manual. Maybe it's fine to just let the legacy anchor catch these, but it did feel like I should go ahead and update them. In the specific context of _this_ README, it felt like sending them to the internal contributor docs in both cases is probably a small improvement.)
- I agree the links can be fragile, but the `ssec-package-tests-writing` section was roughly hollowed out and replaced with equivalent links into the master branch of the repo in the course of #245243, and this is just removing a click from the reader's path to reach the information they need:

    https://github.com/NixOS/nixpkgs/pull/245243/files#diff-99f4eb67c3def9d78bbb9ec86a7dced190f46ee7151c77dcbb15a6a08727a330L565-R39
- I guess I could technically make these relative file links so that it'd be easier to notice that they're broken during work on some other branch. (I'm skeptical it would often get noticed unless we have tooling to check for it.)

    ```suggestion
    You may need to make yourself familiar with [package tests](../../../README.md#package-tests), and [Writing larger package tests](../../../README.md#writing-larger-package-tests), then run tests locally with:
    ```
    
    I'll save actually applying that fix until infinisil weighs in, though.

fricklerhandwerk:
> whether we should have such a section

Absolutely

> whether that section should list common passthru attrs

Yes

> what razors we should apply to decide when a given passthru attr deserves mention here as opposed to a more-specific part of the manual, an in-repo README, code comments, etc.

If it's commonly used, it should be in the common section. Otherwise the specific build helpers should document their passthrus and refer back to the common section.

doc/stdenv/passthru.chapter.md
kjeremy: Is this trying to say that `python.pkgs` and `python.withPackages` are defined in `python.passthru`? If so it should be explicit.

abathur: @kjeremy Can you clarify how that could be more explicit? (Perhaps in the form of a rephrasing? I feel like the text is already explicitly stating that the passthru is where these are defined.)

abathur:
Force-push to incorporate changes that landed a few hours ago via #315616. 

I did reline the changes there and make some small phrasing/consistency edits.

doc/stdenv/passthru.chapter.md
```diff
@@ -0,0 +1,253 @@
...
+Very broadly, passthru attributes listed here fall in one of these categories:
+<!-- 
+Very broadly, an attribute is worth documenting here if it fits either description below.
+Attributes that apply only to a single builder or language ecosystem are best documented alongside the builder or related docs.
+-->
+
+- Attributes that are (at least in theory) globally relevant--you could find them on any derivation in nixpkgs (generally because they don't entail any special support be built into the derivation they're added to). 
+  Common examples of this type are `passthru.tests` and `passthru.updateScript`.
+- Common patterns for adding extra functionality to a derivation (which do tend to entail building support into the derivation or its passthru). 
+  Common examples of this type are `passthru.optional-dependencies`, `passthru.withPlugins`, and `passthru.withPackages`.
```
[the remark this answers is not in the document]
abathur: @fricklerhandwerk Could use guidance on this implementation of the razor and whether you think it belongs in the primary text or in a comment.

AndersonTorres:
What is needed to merge this?

fricklerhandwerk:
I will properly review and merge on Monday or Tuesday.

pkgs/development/mobile/androidenv/README.md
infinisil: Some notes:
- Fully agree that it's a problem that markdown links could become invalid. We should use https://github.com/serokell/xrefcheck to fix this, would be super easy to run in CI actually
- Fully agree that it's a problem that anchors could become invalid when the sections are renamed. There's no good fix here, other than trying not to or to manually insert backwards compatible aliases.
- The proper fix really is to create some infrastructure for building all Markdown files throughout Nixpkgs into a separate contributors manual, with links between them, some custom syntax for stable anchors, etc.
- Yes at that point it's almost like the user reference manual, but with notably differences (see also https://github.com/NixOS/nixpkgs/issues/244056):
  - There should only be one rendered version of the contributors manual, there's no "stable" contributor docs, because you can't contribute to a past version
  - We can make "breaking" changes to the contributor docs at any time
  - The contributor docs aren't useful to a non-contributor
  - The contributor docs should mirror the repo structure (which itself also isn't stable) and be easily discoverable from that.

And the relative links are definitely the way to go imo, because they also work on different branches :)

doc/stdenv/passthru.chapter.md
```diff
@@ -0,0 +1,280 @@
...
+
+<!--
+I'm extracting this statement from the old text because I feel like it isn't very clearly-phrased and doesn't help connect with something most people know already.
+I still think we should try to make sure the pattern it models is represented in the list of common attributes, though:
+
+convey a specific dependency of your derivation which contains a program with plugins support.
+Later, others who make derivations with plugins can use passed-through dependency to ensure that their plugin would be binary-compatible with built program.
+-->
+
+## Common passthru-attributes {#sec-common-passthru-attributes}
```
infinisil, line 38: With the [new convention](https://github.com/NixOS/nixpkgs/pull/245243), I think this entire section should go into `pkgs/README.md` instead, because it's only useful to Nixpkgs contributors. Though one could also argue that the attributes are _exposed_ to users, so they could use it, but that's really not the intention behind these.

And it seems like this would combine well with https://github.com/NixOS/nixpkgs/tree/master/pkgs#package-tests. I never got around to doing this in the original PR, but it would be great to continue going that route!

abathur: This PR is roughly intended to force this sort of decision. (My primary goal here is that we find clarity on whether and to what extent the passthru should be documented in the manual and/or repo-internal docs and end up with a razor sharp enough to look at a given attr and decide whether/where to document it). 

As long as we leave with that razor in hand, I won't put out if the result is that we close this PR and do something else.

CC: @roberth, since this PR results from discussion with him.

fricklerhandwerk:
@abathur here's a sharper razor: if it is only ever needed by contributors, which is likely if links refer to the latest revision of the source code, then it's for the contributor guide. if examples won't go stale if the underlying code changes or moves, it's for the user's manual.

fricklerhandwerk (approved):

Issue #204303, opened 2022-12-03
Discourage/prohibit inheriting from package sets in callPackage


Artturin:
in all-packages.nix when something is inherited (ex, {inherit (pkgs.darwin.apple_sdk.frameworks) Security}) in a callPackage invocation the inherited attr will not have __spliced and so won't be cross-compilation compatible.

one way to fix this would be to change the `with pkgs;` in all-packages to `with pkgs.__splicedPackages` like tried in #58327
however that will result in other issues like infinite recursion.
```
$ nix build ".#pkgsCross.aarch64-multiplatform.bash"
error: infinite recursion encountered
```
other way to work around it would be to to inherit from the __splicedPackages set (ex, {inherit (pkgs.__splicedPackages.darwin.apple_sdk.frameworks) Security})

the way i prefer would be to just not inherit from package sets

Demo: `show-splices.nix`

```nix
let
  pkgs = import ./. { };
  pkgsCross = pkgs.pkgsCross.aarch64-multiplatform;
  workingSplices = pkgsCross.callPackage
    ({ darwin, xorg }: {
      d = darwin.apple_sdk.frameworks.Security.__spliced;
      x = xorg.libX11.__spliced;
    })
    { };

  notWorkingSplices = pkgsCross.callPackage
    ({ Security, libX11 }: {
      d = Security.__spliced;
      x = libX11.__spliced;
    })
    {
      inherit (pkgsCross.darwin.apple_sdk.frameworks) Security;
      inherit (pkgsCross.xorg) libX11;
    };
in
{
  # for --json to work
  workingSplices =
    removeAttrs workingSplices [ "override" "overrideDerivation" ];
  notWorkingSplices =
    removeAttrs notWorkingSplices [ "override" "overrideDerivation" ];
}
```

working:
```
$ NIXPKGS_ALLOW_UNSUPPORTED_SYSTEM=1 nix eval -f show-splices.nix workingSplices --json | jq
{
  "d": {
    "buildBuild": "/nix/store/ls83bixygn2bzqq1w2z66dyccfn61wic-apple-framework-Security",
    "buildHost": "/nix/store/ls83bixygn2bzqq1w2z66dyccfn61wic-apple-framework-Security",
    "buildTarget": "/nix/store/ls83bixygn2bzqq1w2z66dyccfn61wic-apple-framework-Security",
[...]

PR #315949, opened 2024-05-30
emacs: migrate pkgs/build-support/emacs to pkgs/applications/editors/…

AndersonTorres:
Waiting #316726 
https://nixpk.gs/pr-tracker.html?pr=316726

## Description of changes

As a consequence of restrictions imposed by RFC 140 - Simple Package Paths -,
files related to a package should be confined on the package directory.

Certainly this restriction does not apply to packages outside by-name hierarchy.

Nonetheless, this is an interesting organization heuristics: things that affect
Emacs should be confined inside Emacs directory.

Indeed a similar task was done before - namely, the migration of emacs-modes to
elisp-packages: https://github.com/NixOS/nixpkgs/pull/123859

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---


jian-lin (commented):
What is the motivation?

I think `pkgs/build-support/emacs` is pretty clear.  In addition, renaming files makes it harder to use `git blame`.

AndersonTorres:
`pkgs/by-name` requires all files used by a package should be in a same directory.
Yes, I know Emacs cannot be migrated yet, however keeping related things together (instead of scattered around the codebase) conveys a better organization and avoids hopping over otherwise unrelated directory tress.

And moving files (without editing them) don't affect the blame mechanics.
Indeed this blaming issue was bring at the discussions of the by-name massive migration already.

jian-lin:
> `pkgs/by-name` requires all files used by a package should be in a same directory. Yes, I know Emacs cannot be migrated yet

What else prevent Emacs from migrating to by-name if this PR is merged?

> 
> And moving files (without editing them) don't affect the blame mechanics.

Good to know that.  TIL: `git log --follows` can also follow file renaming.

AndersonTorres:
> What else prevent Emacs from migrating to by-name if this PR is merged?
> 

The most obvious is the use of `darwin.apple_sdk.callPackage` instead of pure `callPackage`. 
This is very hard, given [that whole banzé about splicing](https://github.com/NixOS/nixpkgs/issues/204303#issuecomment-1383282367) and other dark corners of Nixpkgs.

The second, and conceptually most complicated to deal with, is the use of `recurseIntoAttrs`, or, more accurately, the fact we tailored a single function in order to build four Emacs versions (namely, 28, 29 and their Macport counterparts) and a convenient bunch of flavors (gtk, pgtk, nox...).

This is a similar situation to some other package sets like the multiple versions of `nv-codec-headers`.

jian-lin:
I am slightly against this PR because
- `pkgs/build-support/` is a good fit for these files.  There are over 400 other "build-support" files like these in that directory.
- whether Emacs can be migrated to "by-name" is unclear to me, which makes the potential benefit of this PR rather small.

AndersonTorres:
>     * `pkgs/build-support/` is a good fit for these files.  There are over 400 other "build-support" files like these in that directory.

Almost 500 files inside less than 200 directories? It looks like a mess if you ask me.
This is why I do not believe the old hierarchy is a good fit for anything at all.

Further,
- We can easily throw everything at `misc` and conceptually nothing will change into the old hierarchy.
- The fetchers can - and certainly will - be migrate to by-name. This will remove most of the raison d'ètre of build-support directory
- many of the files inside are independent packages themselves. They can be migrated to by-name, defating even more the purposes of this directory.

Further
- `pkgs/build-support/` was a serious hurdle when I tried to migrate gettext to by-name, because gettext expression tries to access a file outside gettext directory - namely, files inside build-support.

Further
- emacs-modes was migrated to elisp-packages. It certainly improved Emacs Nix ecosystem maintenance, even more after  that dreadful debasement of Emacs to staging.
  - About this, remember that [a mere 22-line modification that affects only Emacs Lisp packages](https://github.com/NixOS/nixpkgs/pull/316726) causes mass rebuilds.

> whether Emacs can be migrated to "by-name" is unclear to me, which makes the potential benefit of this PR rather small.

The organization provided by this is way bigger, I would say,  given what I wrote above.

jian-lin:
I am convinced by you.

One way to avoid conflict with changes targeting the staging branch is to wait for another staging cycle before merging this PR.  It is OK for you?


> * way before RFC 140 be a thing, emacs-modes floated around pkgs and it was migrated to elisp-packages. The benefits of this migration were rather small initially, nonetheless it certainly improved Emacs Nix ecosystem maintenance along the Nixpkgs life line, even more after that dreadful debasement of Emacs to staging.
>   
>   * About this, remember that [a mere 22-line modification that affects only Emacs Lisp packages](https://github.com/NixOS/nixpkgs/pull/316726) causes mass rebuilds.
> 

I do not follow.  That massive rebuild is caused by the changed attributes, such as `buildPhase` and `installPhase`, of derivations, which is orthogonal to file locations.

AndersonTorres:
> One way to avoid conflict with changes targeting the staging branch is to wait for another staging cycle before merging this PR. It is OK for you?

Yes, it is OK to me.

> I do not follow. That massive rebuild is caused by the changed attributes, such as `buildPhase` and `installPhase`, of derivations, which is orthogonal to file locations.

I was not talking about the migration, apologies for not being clear.

I was saying that this mass-rebuild is very localized: while a modification at (say) meson setup hook propagates over all Nixpkgs, the modification above affects only the Elisp ecosystem.
Therefore, it makes sense to confine these files inside Emacs subdir.

jian-lin:
Hi, since the staging-next branch is just merged, it's time to do the last rebase!

jian-lin (approved):
I try to eval some related packages and find no issue.

AndersonTorres:
> This is a similar situation to some other package sets like the multiple versions of `nv-codec-headers`.

Answering myself, this example is outdated - and it can be a good news!
 
https://github.com/NixOS/nixpkgs/pull/324199

PR #315966, opened 2024-05-30
hterm: init at 0.8.9

zebreus:
## Description of changes

[hterm](https://www.der-hammer.info/pages/terminal.html) is a GUI terminal program for serial communication.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [x] i686-linux

---

### DIFF
--- a/pkgs/by-name/ht/hterm/package.nix
+++ b/pkgs/by-name/ht/hterm/package.nix
@@ -1,0 +1,72 @@
+{
+  stdenv,
+  lib,
+  fetchurl,
+  cairo,
+  pango,
+  libpng,
+  expat,
+  fontconfig,
+  gtk2,
+  xorg,
+  autoPatchelfHook,
+}:
+stdenv.mkDerivation (finalAttrs: {
+  pname = "hterm";
+  version = "0.8.9";
+
+  src =
+    let
+      versionWithoutDots = builtins.replaceStrings [ "." ] [ "" ] finalAttrs.version;
+    in
+    if stdenv.targetPlatform.is64bit then
+      fetchurl {
+        url = "https://www.der-hammer.info/terminal/hterm${versionWithoutDots}-linux-64.tgz";
+        hash = "sha256-DY+X7FaU1UBbNf/Kgy4TzBZiocQ4/TpJW3KLW1iu0M0=";
+      }
+    else
+      fetchurl {
+        url = "https://www.der-hammer.info/terminal/hterm${versionWithoutDots}-linux-32.tgz";
+        hash = "sha256-7wJFCpeXNMX94tk0QVc0T22cbv3ODIswFge5Cs0JhI8=";
+      };
+  sourceRoot = ".";
+
+  nativeBuildInputs = [ autoPatchelfHook ];
+
+  buildInputs = [
+    cairo
+    pango
+    libpng
+    expat
+    fontconfig.lib
+    gtk2
+    xorg.libSM
+  ];
+
+  installPhase = ''
+    runHook preInstall
+    install -m755 -D hterm $out/bin/hterm
+    install -m644 -D desktop/hterm.png $out/share/pixmaps/hterm.png
+    install -m644 -D desktop/hterm.desktop $out/share/applications/hterm.desktop
+    runHook postInstall
+  '';
+
+  passthru = {
+    updateScript = ./update.sh;
+  };
+
+  meta = {
+    homepage = "https://www.der-hammer.info/pages/terminal.html";
+    changelog = "https://www.der-hammer.info/terminal/CHANGELOG.txt";
+    description = "A terminal program for serial communication";
+    # See https://www.der-hammer.info/terminal/LICENSE.txt
+    license = lib.licenses.unfree;
+    sourceProvenance = with lib.sourceTypes; [ binaryNativeCode ];
+    platforms = [
+      "x86_64-linux"
+      "i686-linux"
+    ];
+    maintainers = with lib.maintainers; [ zebreus ];
+    mainProgram = "hterm";
+  };
+})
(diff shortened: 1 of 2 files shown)


pkgs/by-name/ht/hterm/package.nix
```diff
@@ -0,0 +1,65 @@
...
+, expat
+, fontconfig
+, gtk2
+, xorg
+, autoPatchelfHook
+, copyDesktopItems
+}:
+
+stdenv.mkDerivation
+  (finalAttrs: {
```
AndersonTorres, line 16:
```suggestion
stdenv.mkDerivation (finalAttrs: {
```

pkgs/by-name/ht/hterm/package.nix
```diff
@@ -0,0 +1,65 @@
...
+    pname = "hterm";
+    version = "0.8.9";
+
+    src = fetchurl {
+      url = "https://www.der-hammer.info/terminal/hterm${builtins.replaceStrings ["."] [""] finalAttrs.version}-linux-${if stdenv.targetPlatform.is64bit then "64" else "32"}.tgz";
+      hash = "sha256-DY+X7FaU1UBbNf/Kgy4TzBZiocQ4/TpJW3KLW1iu0M0=";
```
AndersonTorres, line 22: The same hash for both 64 and 32?

pkgs/by-name/ht/hterm/package.nix
```diff
@@ -0,0 +1,65 @@
...
+      description = "A terminal program for serial communication";
+      # See https://www.der-hammer.info/terminal/LICENSE.txt
+      license = licenses.unfree;
+      sourceProvenance = with sourceTypes; [ binaryNativeCode ];
+      platforms = [ "x86_64-linux" "i686-linux" ];
+      maintainers = with maintainers; [ zebreus ];
+      mainProgram = "hterm";
+    };
+
+    passthru = {
```
AndersonTorres, line 62: usually passthru comes before meta

pkgs/by-name/ht/hterm/package.nix
```diff
@@ -0,0 +1,65 @@
...
+      xorg.libSM
+    ];
+
+    installPhase = ''
+      runHook preInstall
+      install -m755 -D hterm $out/bin/hterm
+      runHook postInstall
+    '';
+
+    desktopItems = [ "desktop/hterm.desktop" "desktop/hterm.png" ];
```
AndersonTorres, line 47: PNG is not a desktop item.
Install these files manually

pkgs/by-name/ht/hterm/package.nix
zebreus: No, I just forgot to add the hash for the 32-bit archive.

pkgs/by-name/ht/hterm/package.nix
zebreus: Thanks, didn't know that.

pkgs/by-name/ht/hterm/package.nix
```diff
@@ -0,0 +1,77 @@
...
+  ];
+
+  installPhase = ''
+    runHook preInstall
+    install -m755 -D hterm $out/bin/hterm
+    install -m644 -D desktop/hterm.png $out/share/pixmaps/hterm.png
+    runHook postInstall
+  '';
+
+  desktopItems = [ "desktop/hterm.desktop" ];
```
AndersonTorres, line 57: Install this manually.

pkgs/by-name/ht/hterm/package.nix
```diff
@@ -0,0 +1,77 @@
...
+
+  passthru = {
+    updateScript = ./update.sh;
+  };
+
+  meta = with lib; {
```
AndersonTorres, line 63:
```suggestion
  meta = {
```

pkgs/by-name/ht/hterm/package.nix
```diff
@@ -0,0 +1,77 @@
...
+    homepage = "https://www.der-hammer.info/pages/terminal.html";
+    changelog = "https://www.der-hammer.info/terminal/CHANGELOG.txt";
+    description = "A terminal program for serial communication";
+    # See https://www.der-hammer.info/terminal/LICENSE.txt
+    license = licenses.unfree;
+    sourceProvenance = with sourceTypes; [ binaryNativeCode ];
```
AndersonTorres, line 69:
```suggestion
    license = lib.licenses.unfree;
    sourceProvenance = with lib.sourceTypes; [ binaryNativeCode ];
```

pkgs/by-name/ht/hterm/package.nix
```diff
@@ -0,0 +1,77 @@
...
+    platforms = [
+      "x86_64-linux"
+      "i686-linux"
+    ];
+    maintainers = with maintainers; [ zebreus ];
```
AndersonTorres, line 74:
```suggestion
    maintainers = with lib.maintainers; [ zebreus ];
```

AndersonTorres (changes requested):

pkgs/by-name/ht/hterm/package.nix
zebreus: Thanks for the suggestion, I will install it manually.

However I am wondering why I should not use the copyDesktopItems hook, that was made for the purpose of installing desktop files with the correct permissions to the correct location.

AndersonTorres: Usually, when the upstream project provides a file, we either 
- use it as is 
- patch it, in cases like inexistent paths (say, `/usr/bin/hterm` being cited)
- ignore completely and create our own if the file is utterly broken :)

in this order of preference.

The `copyDesktopItems` hook is intended to be used in conjunction with `makeDesktopItem`, a template that receives inputs, checks them and emit a file in the Nix Store.
Since they are meant to generate the files and deal with corner cases like interactions with other hooks and functions from Nixpkgs, they are naturally more complex.

There is no need to bring all this complexity just to copy a file already provided by upstream.

This would be similar to use rsync to copy a local file to a local directory. Too much complexity with no discernible benefit.

AndersonTorres (approved):

AndersonTorres (approved):

PR #316081, opened 2024-05-31
Promote cinnamon to highlights in release notes

jacg:
Last minute change emerging from discussion on Release Management matrix


jacg:
Looks like this has already been addressed in #316026.

PR #316107, opened 2024-05-31
melpaBuild: improve UX

jian-lin:
## Description of changes

This PR improves the UX of `melpaBuild` by providing doc and default value for `commit` and `recipe` and adding a `files` parameter to set `:files` of recipe conveniently.

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/build-support/emacs/melpa.nix
+++ b/pkgs/build-support/emacs/melpa.nix
@@ -35,8 +35,9 @@ in
   pname
   /*
     ename: Original Emacs package name, possibly containing special symbols.
+    Default: pname
   */
-, ename ? null
+, ename ? pname
 , version
   /*
     commit: Optional package history commit.
@@ -45,22 +46,22 @@ in
     the build process.
   */
 , commit ? (args.src.rev or "unknown")
-, recipe
+  /*
+    recipe: Optional MELPA recipe.
+    Default: a minimal but not accurate recipe
+    You need to provide your own recipe if you want to set :files.
+  */
+, recipe ? (writeText "${pname}-recipe" ''(${ename} :fetcher git :url "")'')
 , meta ? {}
 , ...
 }@args:
 
 genericBuild ({
 
-  ename =
-    if ename == null
-    then pname
-    else ename;
-
   elpa2nix = ./elpa2nix.el;
   melpa2nix = ./melpa2nix.el;
 
-  inherit packageBuild commit;
+  inherit packageBuild commit ename recipe;
 
   preUnpack = ''
     mkdir -p "$NIX_BUILD_TOP/recipes"
--- a/pkgs/build-support/emacs/melpa.nix
+++ b/pkgs/build-support/emacs/melpa.nix
@@ -38,6 +38,13 @@ in
   */
 , ename ? null
 , version
+  /*
+    commit: Optional package history commit.
+    Default: src.rev or "unknown"
+    This will be written into the generated package but it is not needed during
+    the build process.
+  */
+, commit ? (args.src.rev or "unknown")
 , recipe
 , meta ? {}
 , ...
@@ -53,7 +60,7 @@ genericBuild ({
   elpa2nix = ./elpa2nix.el;
   melpa2nix = ./melpa2nix.el;
 
-  inherit packageBuild;
+  inherit packageBuild commit;
 
   preUnpack = ''
     mkdir -p "$NIX_BUILD_TOP/recipes"
--- a/pkgs/build-support/emacs/melpa.nix
+++ b/pkgs/build-support/emacs/melpa.nix
@@ -47,11 +47,19 @@ in
   */
 , commit ? (args.src.rev or "unknown")
   /*
-    recipe: Optional MELPA recipe.
-    Default: a minimal but not accurate recipe
-    You need to provide your own recipe if you want to set :files.
+    files: Optional recipe property specifying the files used to build the package.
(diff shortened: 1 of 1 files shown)


pkgs/build-support/emacs/melpa.nix
```diff
@@ -35,25 +35,31 @@ in
...
   /*
     ename: Original Emacs package name, possibly containing special symbols.
+    Default: pname
   */
-, ename ? null
+, ename ? pname
 , version
-, recipe
+  /*
+    commit: Optioal package history commit.  Default: src.rev or "unknown"
```
AndersonTorres, line 43: Typo:

```suggestion
    commit: Optional package history commit.  Default: src.rev or "unknown"
```

pkgs/build-support/emacs/melpa.nix
```diff
@@ -35,25 +35,31 @@ in
...
+    This will be written into the generated package but it not needed during
+    the build process.
+  */
+, commit ? (args.src.rev or "unknown")
+  /*
+    recipe: Optional MELPA recipe.  Default: a minimal but not accurate recipe
+    You need to provide your own recipe if you want to set :file.
+  */
+, recipe ? (writeText "${pname}-recipe" ''(${ename} :fetcher git :url "")'')
```
AndersonTorres, line 52: Well, I understand this is not accurate, but there are other fetchers too. 
The recipe description is a bit free-form:

https://github.com/melpa/melpa?tab=readme-ov-file#recipe-format

(I am cogitating something like a `writeMelpaRecipe` function that converts an input set to the format above.)

jian-lin: I think the current implementation is good enough and reaches a good balance between complexity and functionality.

After reading related code of `melpaBuild`, [`package-recipe`][2] and [`package-build`][3], I find that only `:name`, `:version`, `:commit` and `:files` can affect the build process of `melpaBuild`.  `:fetcher` and `:url` (or `:repo`) are merely used to pass the [package-recipe--validate][1].  They cannot affect the build process of `melpaBuild`.  So I do not think extra complexity to provide a more accurate recipe is justified.

[1]: https://github.com/melpa/package-build/blob/d9ce37b82529827ce720ad3ccf4af887d19fcfb0/package-recipe.el#L162
[2]: https://github.com/melpa/package-build/blob/master/package-recipe.el
[3]: https://github.com/melpa/package-build/blob/master/package-build.el

jian-lin: > I am cogitating something like a `writeMelpaRecipe` function that converts an input set to the format above.

This is useful if users want to provide their own recipe instead of using the default value.

I think the only case they want to do so is to set `:files`.  To improve this UX of this use case, my proposal is to add another argument `files ? null` for `melpaBuild` and use it for the default value of `recipe`.

```nix
, files ? null
, recipe ? (writeText "${pname}-recipe" ''
    (${ename} :fetcher git :url ""
              ${if files != null then ":files " + files else ""})
  '')
```

AndersonTorres: Well, given that it will not be worse than the current situation, then fine.

AndersonTorres (approved):

pkgs/build-support/emacs/melpa.nix
jian-lin: > To improve UX of this use case, my proposal is to add another argument `files ? null` for `melpaBuild` and use it for the default value of `recipe`.

Implemented.

pkgs/build-support/emacs/melpa.nix
```diff
@@ -35,25 +35,38 @@ in
...
   /*
     ename: Original Emacs package name, possibly containing special symbols.
+    Default: pname
   */
-, ename ? null
+, ename ? pname
 , version
-, recipe
+  /*
+    commit: Optional package history commit.  Default: src.rev or "unknown"
```
AndersonTorres, line 43:
```suggestion
    commit: Optional package history commit.
    Default: src.rev or "unknown"
```

pkgs/build-support/emacs/melpa.nix
```diff
@@ -35,25 +35,38 @@ in
...
+    This will be written into the generated package but it not needed during
+    the build process.
+  */
+, commit ? (args.src.rev or "unknown")
+  /*
+    files: Optional recipe property specifying the files used to build the package.
+    Default: null.  If null, the upstream default value will be used.
```
AndersonTorres, line 50:
```suggestion
    If null, do not set it at `recipe`, keeping the default upstream behavior.
    Default: null.  
```

pkgs/build-support/emacs/melpa.nix
```diff
@@ -35,25 +35,38 @@ in
...
+  */
+, files ? null
+  /*
+    recipe: Optional MELPA recipe.  Default: a minimal but not accurate recipe
```
AndersonTorres, line 54:
```suggestion
    recipe: Optional MELPA recipe.
    Default: a minimally functional recipe.
```

AndersonTorres (commented):
Some nits.

AndersonTorres (approved):

jian-lin:
Thanks for your review!

PR #316139, opened 2024-05-31
bind: make systemd service wait for BIND to be ready

Luflosi:
## Description of changes
Without this change, the systemd unit will be marked as ready even though BIND has not finished starting yet.
This causes other units that depend on BIND to start even though BIND is not ready yet.
From https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=900788: "Bind9 will daemonize itself _when it is ready_."

Also modify the NixOS test. With this change, waiting for the unit alone will ensure that BIND is ready to accept queries. I would have expected to see the test failing without this commit but with the `machine.wait_for_open_port(53)` line removed but I found this to not be the case most of the time. This is probably the case because the situation is inherently racy and on my machine BIND happens to start in time most of the time.

Cc @xzfc @mrobbetts @mattchrist @yu-re-ka

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [x] Tested, as applicable:

---

### DIFF
--- a/nixos/modules/services/networking/bind.nix
+++ b/nixos/modules/services/networking/bind.nix
@@ -271,7 +271,8 @@ in
       '';
 
       serviceConfig = {
-        ExecStart = "${bindPkg.out}/sbin/named -u ${bindUser} ${optionalString cfg.ipv4Only "-4"} -c ${cfg.configFile} -f";
+        Type = "forking"; # Set type to forking, see https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=900788
+        ExecStart = "${bindPkg.out}/sbin/named -u ${bindUser} ${optionalString cfg.ipv4Only "-4"} -c ${cfg.configFile}";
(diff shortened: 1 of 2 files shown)


peti (approved):
Looks good to me.

I'm kinda surprised that `named` appears to lack a command-line option for "no detaching", which would IMHO be a nicer solution than having `systemd` track the fork. But alas, there seems to be no way to keep `named` running in the foreground.

Luflosi:
Sure there is a way to keep `named` running in the foreground (the `-f` option that I'm removing in this PR) but then there is no way to know when it finished starting. As far as I know waiting for it to fork is the only way to know when it's ready.

fpletz (approved):

PR #316230, opened 2024-05-31
nixos/emacs: add alternateEditor option

acid-bong:
## Description of changes

1. ~~Originally, this weird `emacseditor` script created a GUI client, when no arguments were provided, or a TUI one, when there was at least one arg (like `$EDITOR /tmp/crontab-2183912`). Not so convenient, when you're expecting the consistent behaviour. My proposition: replace this script and `EDITOR`'s value with `emacsclient -ca '${alternativeEditor}'`, so that, if `cfg.package` is set to a GUI-capable edition of Emacs, the command always started a graphical window.~~
2. ~~`cfg.alternativeEditor` is described in the patch itself.~~

This PR replaces `--alternate-editor` flag in the script with an envvar and also makes its value configurable.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/nixos/modules/services/editors/emacs.nix
+++ b/nixos/modules/services/editors/emacs.nix
@@ -9,9 +9,9 @@ let
   editorScript = pkgs.writeScriptBin "emacseditor" ''
     #!${pkgs.runtimeShell}
     if [ -z "$1" ]; then
-      exec ${cfg.package}/bin/emacsclient --create-frame --alternate-editor ${cfg.package}/bin/emacs
+      exec ${cfg.package}/bin/emacsclient --create-frame
     else
-      exec ${cfg.package}/bin/emacsclient --alternate-editor ${cfg.package}/bin/emacs "$@"
+      exec ${cfg.package}/bin/emacsclient "$@"
     fi
   '';
 
@@ -55,6 +55,20 @@ in
       '';
     };
 
+    alternateEditor = mkOption {
+      type = types.str;
+      default = "emacs";
+      example = "vi";
+      description = ''
+        If the Emacs server is not running, the shell command in this
+        environment variable runs instead. If set to the empty string, Emacs
+        starts in daemon mode, and the client tries to connect to it. Will be
+        overridden by the `--alternate-editor` option, if present.
+
+        Here it defaults to a non-daemon instance of Emacs.
+      '';
+    };
+
     startWithGraphical = mkOption {
       type = types.bool;
       default = config.services.xserver.enable;
@@ -85,7 +99,10 @@ in
 
     environment.systemPackages = [ cfg.package editorScript ];
 
-    environment.variables.EDITOR = mkIf cfg.defaultEditor (mkOverride 900 "emacseditor");
+    environment.variables = {
+      EDITOR = mkIf cfg.defaultEditor (mkOverride 900 "emacseditor");
+      ALTERNATE_EDITOR = cfg.alternateEditor;
+    };
   };
 
   meta.doc = ./emacs.md;


jian-lin:
> this weird `emacseditor` script created a GUI client, when no arguments were provided, or a TUI one, when there was at least one arg (like `$EDITOR /tmp/crontab-2183912`)

It seems it is necessary to not have any existing GUI frame to reproduce your second case.  If so, one workaround for you is to keep at least one GUI frame always open.

nixos/modules/services/editors/emacs.nix
```diff
@@ -83,9 +84,9 @@ in
       wantedBy = if cfg.startWithGraphical then [ "graphical-session.target" ] else [ "default.target" ];
     };
 
-    environment.systemPackages = [ cfg.package editorScript ];
+    environment.systemPackages = [ cfg.package ];
 
-    environment.variables.EDITOR = mkIf cfg.defaultEditor (mkOverride 900 "emacseditor");
+    environment.variables.EDITOR = mkIf cfg.defaultEditor (mkOverride 900 "emacsclient -ca '${cfg.alternativeEditor}'");
```
jian-lin, line 89: Here are some issues I notice:
- Always passing `-c` prevents users from re-using existing frame(s).
- An empty string of `cfg.alternativeEditor` starts **another** Emacs daemon which is not managed by systemd.  I think it is a bad idea.  Users of this module probably want their daemon to be managed by system.
- The single quote in `-a '...'` causes issues if `cfg.alternativeEditor` contains environment variables.

jian-lin (commented):
My proposal is to add an option to set flags passed to `emacsclient`.

nixos/modules/services/editors/emacs.nix
acid-bong: 1. Imma be honest, I don't have enough experience with Emacs, but why reusing existing frames?
2. It only starts the daemon if it's not running (for example, when a standalone window manager user doesn't start `graphical-session.target` manually), otherwise nothing happens. 
3. Will it? Cuz the entire string will be in double quotes, won't it interpolate instantly?

acid-bong:
> one workaround for you is to keep at least one GUI frame always open.

Weird workaround imho. I don't want Emacs to be TUI, otherwise i'd choose emacs-nox as `emacs.package`.

Forgot to clarify, another reason why I wanna remove that script is to change the logic behind the package behaviour: 
- the behaviour will be consistent: if `emacs.package` a GUI edition (and it is one by default), a graphical window will always appear
- if a user desires to use emacs strictly in the terminal, they better set `emacs.package = pkgs.emacs-nox`
- otherwise they can always set EDITOR and VISUAL to the desired values explicitly

nixos/modules/services/editors/emacs.nix
jian-lin: > why reusing existing frames?

For example, it is easier to move between buffers, windows and tabs in one frame.

> It only starts the daemon if it's not running

In that case, the old behaviour, i.e., launching a non-daemon Emacs, make more sense.  If you want a daemon, you should run `systemctl start emacs`.  Just my two cents.

> Will it?

My experiment says yes.

jian-lin:
> Weird workaround imho.

Not so weird actually.  The first thing of many Emacs users after their system boots is to open an Emacs frame.  I am one of them.  That's the reason why I cannot reproduce your TUI issue initially.  Only after some thoughs, I find it requires no existence of GUI frames.

> if a user desires to use emacs strictly in the terminal, they better set emacs.package = pkgs.emacs-nox

One disadvantage is that cache is not provided for `emacs-nox.pkgs` by cache.nixos.org.  So even if a user only use TUI Emacs, he may still choose `pkgs.emacs`.

> My proposal is to add an option to set flags passed to `emacsclient`.

Does this proposal satisfy you?

You never how an Emacs users uses Emacs.

nixos/modules/services/editors/emacs.nix
acid-bong: 2. Fair enough, I'll set the default altEditor value back to "emacs"

acid-bong:
>  The first thing of many Emacs users after their system boots is to open an Emacs frame.

Okay, I get it now. If i'm using vterm in Emacs instead of a separate terminal, i'd also expect the editor window to be reused (like, a new tab or buffer). That's understandable.

> options to set flags

Yup, that also sounds fair.

---

Okay, while i was typing, i decided i'll reduce the PR to only having ALTERNATIVE_EDITOR toggleable

nixos/modules/services/editors/emacs.nix
```diff
@@ -85,7 +99,10 @@ in
 
     environment.systemPackages = [ cfg.package editorScript ];
 
-    environment.variables.EDITOR = mkIf cfg.defaultEditor (mkOverride 900 "emacseditor");
+    environment.variables = {
+      EDITOR = mkIf cfg.defaultEditor (mkOverride 900 "emacseditor");
+      ALTERNATE_EDITOR = cfg.alternateEditor;
```
jian-lin, line 104: There is an issue of using a environment variable `ALTERNATE_EDITOR` instead of setting `--alternate-editor` in the script.  If a user changes`alternateEditor` and runs `nixos-rebuild test (or switch)`, `ALTERNATE_EDITOR` is not updated in the current shell.

acid-bong: Welp, that's an issue with all envvars, no? I would imagine an Emacs user rather test the alteditor with `emacsclient -a "new value"` or, at worst, with `ALTERNATE_EDITOR="new value" emacsclient`.

(this assumes that a user did read the option's description)

jian-lin: > that's an issue with all envvars, no?

Yes, and that issue can be avoided here if we do not use env var in the first place.

So what is the benefit of not using `-a` in the script?

acid-bong:
Ight, I give up. Apologies for bothering yall

PR #299843, opened 2024-03-28
qtile: Fix QTile restart environment

Rutherther:
### Issue

I stumbled across a bug in the QTile module where if qtile is restarted via their restart mechanism (https://github.com/qtile/qtile/blob/master/libqtile/core/lifecycle.py#L35), and there are extra packages configured in the module, the Python modules won't be in the new environment, and thus lead to `ModuleNotFoundError` if the modules are used in `config.py`.

### Notes

The cause of this is that the executable that is being used for restart is the unwrapped package itself, not the python environment that contains the extra packages.

I would welcome a discussion about possible solutions. My solution is sort of an ad-hoc solution, that works, but doesn't seem ideal to me. It seems to me these are basically the options we have:
- Change restart behavior inside of QTile to call the correct executable
- Make sure the argv[0] for QTile is the python environment wrapper file instead of the QTile unwrapped package 
- pass packages to the unwrapped package to make module available inside of QTile unwrapped package (what I did)

The other issue is good implementation of one of these options.

I am open to suggestions and willing to implement them myself.

### Current solution

Currently I implemented a solution where the package going to the module is overriden, and `extraPackages` are passed into the `propagatedBuildInputs` of the unwrapped package. This seems to be working fine. I have also removed the wrapping python environment since it seems to have become obsolete by the change I introduced.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
- For non-Linux: Is sandboxing enabled in `nix.conf`? (See [Nix manual](https://nixos.org/manual/nix/stable/command-ref/conf-file.html))
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:
  - [NixOS test(s)](https://nixos.org/manual/nixos/unstable/index.html#sec-nixos-tests) (look inside [nixos/tests](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests))
  - and/or [package tests](https://nixos.org/manual/nixpkgs/unstable/#sec-package-tests)
  - or, for functions and "core" functionality, tests in [lib/tests](https://github.com/NixOS/nixpkgs/blob/master/lib/tests) or [pkgs/test](https://github.com/NixOS/nixpkgs/blob/master/pkgs/test)
  - made sure NixOS tests are [linked](https://nixos.org/manual/nixpkgs/unstable/#ssec-nixos-tests-linking) to the relevant packages
[...]

PR #316239, opened 2024-05-31
qtile: install proper session .desktop files for both xorg and wayland

arjan-s:
## Description of changes
This PR modifies the qtile package to install the provided session desktop files for login managers.
That in turn allows us to add the package to `services.displayManager.sessionPackages` when the NixOS module is enabled.
After that, both the Xorg and Wayland sessions will be available from login (display) managers, instead of just one session that's hard coded to a specific backend (via the now removed option `backend`).

Additionally, this fixes the `services.xserver.windowManager.qtile.finalPackage` assignment to use the qtile package instead of building a whole new env.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [x] Tested, as applicable:

---

### DIFF
--- a/nixos/modules/services/x11/window-managers/qtile.nix
+++ b/nixos/modules/services/x11/window-managers/qtile.nix
@@ -7,6 +7,10 @@ let
 in
 
 {
+  imports = [
+    (mkRemovedOptionModule [ "services" "xserver" "windowManager" "qtile" "backend" ] "The qtile package now provides separate display sessions for both X11 and Wayland.")
+  ];
+
   options.services.xserver.windowManager.qtile = {
     enable = mkEnableOption "qtile";
 
@@ -19,14 +23,6 @@ in
       description = ''
           Path to the qtile configuration file.
           If null, $XDG_CONFIG_HOME/qtile/config.py will be used.
-      '';
-    };
-
-    backend = mkOption {
-      type = types.enum [ "x11" "wayland" ];
-      default = "x11";
-      description = ''
-          Backend to use in qtile: `x11` or `wayland`.
       '';
     };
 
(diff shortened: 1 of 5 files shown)


Sigmanificient:
image

Sigmanificient:
```
error: Default graphical session, 'none+qtile', not found.
       Valid names for 'services.displayManager.defaultSession' are:
         qtile
         none+icewm
```

Sigmanificient:
build my whole system with your branch, seems to work:
image

Sigmanificient:
@ofborg build qtile-unwrapped.passthru.tests.qtile

arjan-s:
Thanks! This PR changed the session names, that's why the test is failing.
Since these changes work on both our systems, I'll just fix the test.

Another todo I thought of after submitting is marking the now-removed backend option deprecated, so people having that in their configs get a nice message.

Sigmanificient:
> Another todo I thought of after submitting is marking the now-removed backend option deprecated, so people having that in their configs get a nice message.

Yes this would be nice

arjan-s:
Both done!

Sigmanificient (approved):

Sigmanificient (approved):

arjan-s:
Looking at this again, I'm afraid the `configFile` option should be deprecated too. But that could break the way some people configure their system, especially those who don't use home-manager (which can write `~/.config/qtile/config.py`).
We can't easily modify the session files that are now installed by the package, so the 'old' method isn't feasible I think.
I'm hoping others have ideas about this...? Perhaps a good solution would be a link to the provided config in the system-wide config dir ([qtile docs here](https://docs.qtile.org/en/stable/manual/config/index.html#configuration-lookup-order))?

Sigmanificient:
it look after into the `.desktop` pr, indead would break the `configFile` option.

Setting to `/etc/xdg` would work, but for users that used previous `configFile` options, it may not be the cleanest especially one multiple users system.

I'm a bit new to options, could the desktop file be wrapped in it's own derivation, with a substitubeInPlace `qtile start` -> `qtile start --config ${cfg.configFile}`?

arjan-s:
Done. I believe this is ready to test (again).

arjan-s:
Converted to draft, because after testing I found that the logic regarding the wrapper and `systemPackages` isn't right. I'll have to dive deeper to get to a point where it just works.

arjan-s:
Okay, I believe I have now (finally) arrived at a working and much cleaner solution!
* I removed the wrapper since we weren't actually using it anyway (we built a separate new env for display managers and we added the unwrapped package to `environment.systemPackages`).
* The `extraPackages` are now nicely added to the dependencies of Qtile itself, making https://github.com/NixOS/nixpkgs/pull/299843 a thing of the past (I hope).
* The `finalPackage` (which optionally contains `qtile-extras` and other dependencies) is used for both display managers and added to $PATH (the latter makes commands like `qtile check` work).
* If a `configFile` is provided, it is symlinked to `/etc/xdg/qtile/config.py`, which is automatically used by Qtile if `~/.config/qtile/config.py` isn't found.

drupol (approved):
LGTM

nixos/modules/services/x11/window-managers/qtile.nix
```diff
@@ -26,7 +26,7 @@ in
       type = types.enum [ "x11" "wayland" ];
       default = "x11";
       description = ''
-          Backend to use in qtile: `x11` or `wayland`.
+          This option is deprecated. The qtile package now provides separate display sessions for both X11 and Wayland.
```
SuperSandro2000, line 29: We should add an assetion or probably even better use mkRemovedOption

arjan-s: Much better indeed, fixed that now.

drupol (approved):

Sigmanificient (approved):

jwijenbergh (approved):
Thanks a lot for making these changes. Looks good to me. Should we maybe move qtile later to `programs.qtile`? Most wayland compositors do that, but I'm not sure for qtile as it also supports x11

acid-bong:
Should've made a warning for `qtile` instead of removing it tho 😕

Nebucatnetzer:
> Should've made a warning for `qtile` instead of removing it tho 😕

Yeah was quite surprised today when I tested the update to 24.11.

Issue #149692, opened 2021-12-08
cctools ld segfaults when linking haskellPackages.Agda (2.6.2 only?)


sternenseemann:
[Example Log](https://hydra.nixos.org/build/160900296/nixlog/1):

```
Building executable 'agda' for Agda-2.6.2..
[1 of 1] Compiling Main             ( src/main/Main.hs, dist/build/agda/agda-tmp/Main.o )
Linking dist/build/agda/agda ...
/nix/store/bp55vzlmcyqcx9n08pkzslvks10zybwc-clang-wrapper-11.1.0/bin/ld: line 256: 79105 Segmentation fault: 11  /nix/store/3fl5z9yfnz08kjb4cnhw2w8a2zsm5qim-cctools-binutils-darwin-949.0.1/bin/ld ${extraBefore+"${extraBefore[@]}"} ${params+"${params[@]}"} ${extraAfter+"${extraAfter[@]}"}
clang-11: error: linker command failed with exit code 139 (use -v to see invocation)
`cc' failed in phase `Linker'. (Exit code: 139)
builder for '/nix/store/fyzdyk7a79jqfh4r2gvjc65d9f7w22wd-Agda-2.6.2.drv' failed with exit code 1
```

### Steps To Reproduce

`nix-build -A haskellPackages.Agda` on `aarch64-darwin`

cc @NixOS/darwin-maintainers

prusnak:
Does adding `autoSignDarwinBinariesHook` to `nativeBuildInputs` help?

prusnak:
> Does adding `autoSignDarwinBinariesHook` to `nativeBuildInputs` help?

Ah, sorry. This is about fixing "Killed: 9" error.

sternenseemann:
Yeah, `Agda` is a normal haskell package which [work for the most part](https://hydra.nixos.org/eval/1729112?filter=haskellPackages&compare=1728918&full=#tabs-still-succeed).

veprbl:
Would help to post a crash dump from Other -> Console -> User Reports

sternenseemann:
Indeed, would be nice if someone could have a look at this locally, I can only check on Hydra logs, really.

olebedev:
Getting:

```
Linking dist/build/agda/agda ...
[...]

PR #316277, opened 2024-05-31
 haskellPackages.jsaddle-hello: fix on darwin 

alexfmpe:
Strictly speaking, `ghcjs-dom-hello` fails with a linker error on my mac but I suspect that's just for aarch64-darwin since I've often hit https://github.com/NixOS/nixpkgs/issues/149692. Should be fixed when the cctools bump is merged.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [x] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/development/haskell-modules/configuration-darwin.nix
+++ b/pkgs/development/haskell-modules/configuration-darwin.nix
@@ -323,6 +323,10 @@ self: super: ({
   # Tests fail on macOS https://github.com/mrkkrp/zip/issues/112
   zip = dontCheck super.zip;
 
+  warp = super.warp.overrideAttrs (drv: {
+    __darwinAllowLocalNetworking = true;
+  });
+
   jsaddle-wkwebview = overrideCabal (drv: {
     libraryFrameworkDepends = with pkgs.buildPackages.darwin.apple_sdk.frameworks; [ Cocoa WebKit ];
     libraryHaskellDepends = with self; [ aeson data-default jsaddle ]; # cabal2nix doesn't add darwin-only deps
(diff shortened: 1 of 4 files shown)


maralorn:
sterni apparently wants to comment on this as well. I think this largely makes sense.

pkgs/development/haskell-modules/configuration-darwin.nix
```diff
@@ -323,6 +323,20 @@ self: super: ({
   # Tests fail on macOS https://github.com/mrkkrp/zip/issues/112
   zip = dontCheck super.zip;
 
+  warp = super.warp.overrideAttrs (drv: {
+    __darwinAllowLocalNetworking = true;
```
sternenseemann, line 327: How did you hit this? Seems like building `warp` hasn't been a problem so far.

alexfmpe: Might be some combination of sandbox + firewall. I'll try to isolate the trigger

alexfmpe: This builds for me with `--option sandbox false`, does not build with `relaxed` or `true`.
Test suite just hangs after the `"range requests"` group log and there's no more output:
https://github.com/yesodweb/wai/blob/d50e1184600631a114e3d2ad119abdf0ef08834b/warp/test/ResponseSpec.hs#L106
I suspect this is the first case in the suite that actually performs a network request (apparently in `127.0.0.1` flavor)

sternenseemann: Do you hit this with just GHC 9.8? warp does seem to succeed with GHC 9.6 which is confusing to me.

alexfmpe: On further testing it looks like it only happens when both firewall AND sandboxing are active. 
Didn't try 9.8.

PR #316362, opened 2024-06-01
buildDotnetModule: split fetch-deps script to new file

Smaug123:
This makes the minimal change to put the script in a new file. It does not fix the resulting ShellCheck warnings (which would have been present in the previous version; I've just exposed them).

## Description of changes

So that we can run tooling (primarily ShellCheck) against the shell script, I've pulled it out into a new file.

The changes I've made to the Nix file are compliant with `nixfmt`.

Testing: I took [a random one of my projects](https://github.com/Smaug123/WoofWare.Myriad/blob/04ecbe60023c528506ac26c134d52efc4e14a8ee/flake.nix#L48-L57). This is the diff that results between the two scripts on that project. The resulting script correctly generates a lockfile. Lines with `<` are in the new version; lines with `>` are before this change.

```
< #! /nix/store/spyn2n8l9ca1pkqn20l40l424xmqwda4-bash-5.2p26/bin/bash
< # shellcheck shell=bash
---
> #!/nix/store/spyn2n8l9ca1pkqn20l40l424xmqwda4-bash-5.2p26/bin/bash
5c4
< export PATH="/nix/store/pldbvp0p2z2pa76h17g797vvk201155r-coreutils-9.5/bin:/nix/store/spyn2n8l9ca1pkqn20l40l424xmqwda4-bash-5.2p26/bin:/nix/store/nfsin5ji34p53977fpkn68zkavrxah7y-dotnet-sdk-8.0.301/bin:/nix/store/pwwk6vzi1jczh6hcbn6gc348148hyjlq-nuget-to-nix/bin"
---
> export PATH="/nix/store/pldbvp0p2z2pa76h17g797vvk201155r-coreutils-9.5/bin:/nix/store/spyn2n8l9ca1pkqn20l40l424xmqwda4-bash-5.2p26/bin:/nix/store/nfsin5ji34p53977fpkn68zkavrxah7y-dotnet-sdk-8.0.301/bin:/nix/store/kzjfv0cjdiyxbvfbnl33wlvjd8rj2azw-nuget-to-nix/bin"
75c74
< storeSrc="/nix/store/2icddcz9vlyaqrwmjlibggdbqbgzl9kv-WoofWare.Myriad.Plugins-source"
---
> storeSrc="/nix/store/m9jnvfj6f8wvdq2mc8ghs66ip2bh1ydd-WoofWare.Myriad.Plugins-source"
83c82
< "/nix/store/nfsin5ji34p53977fpkn68zkavrxah7y-dotnet-sdk-8.0.301/bin/dotnet" tool restore
---
> /nix/store/nfsin5ji34p53977fpkn68zkavrxah7y-dotnet-sdk-8.0.301/bin/dotnet tool restore
86,88c85
< runtimeIds=('osx-arm64' 'linux-arm64' 'osx-x64' 'linux-x64')
<
< for rid in "${runtimeIds[@]}"; do
---
> for rid in "osx-arm64" "linux-arm64" "osx-x64" "linux-x64"; do
102c99
< excluded_sources=( '/nix/store/82z2qmr1lcds3vlh1ypa2dkhxmzxx10g-dotnet-sdk-8.0.301-deps-nuget-deps' )
---
> excluded_sources="/nix/store/82z2qmr1lcds3vlh1ypa2dkhxmzxx10g-dotnet-sdk-8.0.301-deps-nuget-deps"
```

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [x] aarch64-darwin
  - [x] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---


Smaug123:
(I've rebased `master` and am re-checking my test assertions. Will update this comment when I'm done.)

Smaug123:
Hashbang question resolved; I consider this ready for review.

pkgs/build-support/dotnet/build-dotnet-module/fetchDeps.sh
```diff
@@ -0,0 +1,110 @@
+@shell@
```
corngood, line 1: Minor nitpick, but it seems to be much more common to do something like `#! @shell@`.  I think I'd prefer it that way because currently it's not actually the shell path, it's a shebang containing the shell path.

corngood (changes requested):

pkgs/build-support/dotnet/build-dotnet-module/fetchDeps.sh
Smaug123: I'd strongly prefer not to do that, because then there's no way to prevent Shellcheck's integration into my editor from complaining.

corngood: So I assume you get those errors on e.g. `nixos/modules/system/boot/stage-1-init.sh`?

I'm not sure I like the idea of going against the grain like this.  Couldn't any of the template substitutions confuse the checker?

Smaug123: Yep, stage-1-init.sh contains a whole bunch of Shellcheck errors, the first of which is the invalid hashbang. (Most of those errors look legit: unquoted interpolations, `cd`s that can silently fail, unused variables, `read` which mangles backslashes, etc.)

I've (hopefully) carefully phrased the substitutions so that they are all "valid Bash" both before and after substitution. I consider it a bug in this PR if any of them are invalid. (There are nine Shellcheck errors in fetch-deps.sh, all of which would have been errors before this PR except that it was all hidden within a Nix string; none of the errors are related to the substitutions.)

Smaug123: I could certainly go further: in one of my projects somewhere, I've written a `makeShellScript` of some kind which does substitutions and then runs Shellcheck against the result post-substitution. That would be the gold standard, although fixing errors under that workflow *with the shell script template as an inline Nix string* would still be more annoying vs fixing errors under that workflow *with the shell script template in a separate .sh file*.

corngood:
```
In [...]/nixos/modules/system/boot/stage-1-init.sh line 1:
#! @shell@
^-- SC1008 (error): This shebang was unrecognized. ShellCheck only supports sh/bash/dash/ksh/'busybox sh'. Add a 'shell' directive to specify.
^--------^ SC2239 (error): Ensure the shebang uses an absolute path to the interpreter.
```

I think you just need to add `# shellcheck shell=bash` on the next line to suppress it.  This will likely also help it check more accurately than if it couldn't find a shebang at all.

Smaug123: Oh that does work! Sorry, I thought those directives had to go right at the top of the file, but of course on second thoughts that's a silly idea (how could that possibly work with any existing hashbangs). I'll push that up now.

Smaug123: Done, and I've updated the diff in the original description.

corngood:
Result of `nixpkgs-review pr 316362` run on x86_64-linux [1](https://github.com/Mic92/nixpkgs-review)

pkgs/build-support/dotnet/build-dotnet-module/default.nix
```diff
@@ -202,131 +204,46 @@ stdenvNoCC.mkDerivation (args // {
...
+      path = lib.makeBinPath [
+        coreutils
+        runtimeShellPackage
+        dotnet-sdk
+        (nuget-to-nix.override { inherit dotnet-sdk; })
+      ];
+      dotnetSdkPath = toString dotnet-sdk;
+      excludedSources =
+        lib.concatStringsSep " " (map lib.escapeShellArg sdkDeps);
+    in substituteAll {
```
corngood, line 241: This seems to be resulting in the script not having the executable bit set, so `$(nix-build -A omnisharp-roslyn.fetch-deps)` fails with

> bash: /nix/store/8l6i3lrgy1ima071gvnqd25r73zsmbd2-fetchDeps.sh: Permission denied

pkgs/build-support/dotnet/build-dotnet-module/default.nix
```diff
@@ -202,131 +204,46 @@ stdenvNoCC.mkDerivation (args // {
...
+      src = ./fetchDeps.sh;
```
corngood, line 242: I think naming this `fetch-deps.sh` would be more consistent with the naming of other shell scripts in nixpkgs, and would also match the attribute name.

corngood (changes requested):
Thanks!  Just a couple more things I noticed...

pkgs/build-support/dotnet/build-dotnet-module/default.nix
Smaug123: Oh no, sorry about that - that's pretty inexcusable for me not to have actually *tried* to execute the script, I've only invoked it with `bash` :/

Smaug123: OK, I've now actually done an end-to-end test on omnisharp (NuGet restore failed because of a recent security advisory against System.Text.Json) and on my personal project.

corngood: I tested it here too and it seems fine now.

corngood (approved):

corngood:
FYI @NixOS/dotnet in case anyone has any concerns before I merge this

GGG-KILLER (approved):
Looks good to me

Issue #83884, opened 2020-03-31
Consider building unfreeRedistributable Packages


mkg20001:
Per definition those packages are redistributable without any legal problems (or at least should be, otherwise I feel like the license is wrongly described)

So we could build those as well, because there doesn't seem to be a reason not to

mkg20001:
The required change would be probably this one

```
nixpkgs.config.allowUnfreePredicate = (x: if x ? meta.license then x.meta.license.shortName == "unfreeRedistributable" else false);
```

mkg20001:
Still not resolved

domenkozar:
As per https://github.com/NixOS/nixpkgs/issues/15162 we might actually also want to whitelist what drivers to install into an ISO.

petrosagg:
I hit the same problem today after getting the package for `materialize` which is BSL licensed merged. The software license "grants you the right to copy, modify, create derivative works, redistribute, and make non-production use of the Licensed Work." so it should be perfectly fine to build and distribute binaries through Hydra.

I'd like to help resolve this problem, is the direction set forth in this issue and #97789 still the way to go here?

mkg20001:
I think so, I'm just not sure which way to implement the flag to enable it (ref https://github.com/NixOS/nixpkgs/pull/97789#issuecomment-691194274)

K900:
Any updates here?

mkg20001:
Redistributable is now a thing

```
nix-repl> lib.licenses.free
{ deprecated = false; free = true; fullName = "Unspecified free software license"; redistributable = true; shortName = "free"; }

nix-repl> lib.licenses.free.redistributable
true
```

Hydra building and caching redistributable unfree things is still pending

mkg20001:
The quick fix for hydra would be this:

```
nixpkgs.config.allowUnfreePredicate = (x: if x ? meta.license then x.meta.license.redistributable else false);
```

It only checks redistributable but this is true automatically for all free licenses and being redistributable is all that matters in the case of hydra builds, I suppose.

K900:
Is anything preventing us from making that change on Hydra, like, now?

mkg20001:
Not really, it could just be done

grahamc:
We'd need sign-off from the NixOS Foundation which is responsible for supporting the infrastructure, and might have ethical objections. One concern they've raised is NixOS should be fully usable without any unfree software, and allowing it in Hydra could make NixOS an inadvertently unfree Linux distribution.

[...]

PR #316431, opened 2024-06-01
 tigerjython: init at 2.39

rcmlz:
## Description of changes

Adding new package TigerJython https://tigerjython.ch/

TigerJython offers everything you need to go from programming beginner to professional. You will find a wide variety of tutorials and can get started right away in programming environments specially developed for you. 

TigerJython runs on the Java Virtual Machine (JVM), although we also offer the browser-based online solution [WebTigerJython](https://webtigerjython.ethz.ch) and [WebTigerPython](https://webtigerpython.ethz.ch). The design and development of TigerJython follow the guidelines of simplicity, understandability, and—naturally—Python's battery included philosophy with an extensive set of libraries.

## Things done

Taking the existing .tar.gz file from https://tigerjython.ch/en/products/download wrapping it to run the .jar file using the current stable JRE.

There are no dependencies introduced - just wrapping a jar for convenience.

I run ```nix-build -A tigerjython``` and then ```/nix/store/...-tigerjython-2.39/bin/tigerjython``` successfully.

### DIFF
--- a/pkgs/by-name/ti/tigerjython/package.nix
+++ b/pkgs/by-name/ti/tigerjython/package.nix
@@ -1,0 +1,99 @@
+{
+  stdenvNoCC,
+  lib,
+  fetchurl,
+  makeWrapper,
+  makeDesktopItem,
+  copyDesktopItems,
+  jre,
+}:
+let
+  icon = fetchurl {
+    # In case tigerjython/tjinstall becomes unavailable, use this url - see comment for src in MkDerivation
+    #url = "https://web.archive.org/web/20240623120114/https://raw.githubusercontent.com/tigerjython/tjinstall/master/tjlogo64.png";
+    url = "https://raw.githubusercontent.com/tigerjython/tjinstall/611c56d4e765731883656a5c4b71209d72b5ab74/tjlogo64.png";
+    hash = "sha256-tw3uDWLtcMHYmN6JGsEvVKLgI09v5DF27V2+OF9Z5tA=";
+  };
+in
+stdenvNoCC.mkDerivation (finalAttrs: {
+  pname = "tigerjython";
+
+  # UPDATE instructions
+  #
+  # We cache potentially unstable upstream input (.tar.gz file) via https://web.archive.org - this is a common procedure in Nixpkgs.
+  #
+  # - Open https://tigerjython.ch/en/products/download and identify the new version string for "TigerJython IDE for Linux"
+  version = "2.39";
+
+  # - and copy download link (most likely https://tigerjython.ch/user/pages/download/TigerJython.tar.gz) to clipboard.
+  # - Open http://web.archive.org and paste download link from clipboard into "Save Page Now" field and hit the "Save Page" button.
+  # - Unselect "Save Error Pages" and hit "Save Page" again.
+  # - Wait for the archive link to be generated and copy it to the url field - adjust hash accordingly.
+  src = fetchurl {
+    url = "http://web.archive.org/web/20240119124245/https://tigerjython.ch/user/pages/download/TigerJython.tar.gz";
+    hash = "sha256-PdoAOjr19aLmXYrLtMCq/tZ2Fqq7pINTuhFyMMiC0yM=";
+  };
+
+  nativeBuildInputs = [
+    makeWrapper
+    copyDesktopItems
+  ];
+
+  desktopItems = [
+    (makeDesktopItem {
+      name = "TigerJython";
+      desktopName = "TigerJython";
+      comment = "The Python IDE for beginners";
+      type = "Application";
+      categories = [ "Education" ];
+      terminal = false;
+      startupNotify = false;
+      exec = "tigerjython";
+      icon = "tigerjython";
+      mimeTypes = [ "text/x-python" ];
+    })
+  ];
+
+  dontConfigure = true;
+  dontBuild = true;
+
+  installPhase = ''
+    runHook preInstall
+
+    export CUSTOM_LIBS=$out/share/java
+    export JAR=$CUSTOM_LIBS/tigerjython2.jar
+    export EXAMPLES_DIR=$CUSTOM_LIBS/Examples
+
+    install -Dm444 bin/tigerjython2.jar $JAR
+    install -Dm444 bin/Lib/* --target-directory=$CUSTOM_LIBS
+    install -Dm444 bin/TestSamples/* --target-directory=$EXAMPLES_DIR
+
+    makeWrapper ${jre}/bin/java $out/bin/tigerjython \
+      --add-flags "-Duser.dir=$CUSTOM_LIBS/" \
+      --add-flags "-Xmx512M" \
+      --add-flags "-jar $JAR" \
+      --set _JAVA_OPTIONS '-Dawt.useSystemAAFontSettings=lcd'
+
+    runHook postInstall
+  '';
+
+  postInstall = ''
+    install -Dm444 ${icon} $out/share/icons/hicolor/64x64/apps/tigerjython.png
+  '';
+
+  meta = {
+    homepage = "https://www.tigerjython.ch";
+    downloadPage = "https://tigerjython.ch/en/products/download";
+    description = "Simple development environment for programming in Python";
+    longDescription = ''
+      Designing, coding, and amazing. TigerJython offers everything you need
+      to go from Python programming beginner to professional.
+      You will find a wide variety of tutorials and can get started right away
+      in programming environments specially developed for you.
+    '';
+    license = lib.licenses.unfreeRedistributable;
+    maintainers = with lib.maintainers; [ rcmlz ];
+    platforms = lib.platforms.all;
+    mainProgram = "tigerjython";
+  };
+})
--- a/maintainers/maintainer-list.nix
+++ b/maintainers/maintainer-list.nix
@@ -16839,6 +16839,12 @@
     githubId = 52847440;
     name = "Ryan Burns";
   };
+  rcmlz = {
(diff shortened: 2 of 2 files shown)


Noodlez1232:
Result of `nixpkgs-review pr 316431` run on x86_64-linux [1](https://github.com/Mic92/nixpkgs-review)
<details>
  <summary>1 package failed to build:</summary>
  <ul>
    <li>tigerjython</li>
  </ul>
</details>

Noodlez1232:
The failure here was a hash mismatch. Probably the solution here is to update the hash in the source for TigerJython.tar and for the tjlogo64.png files.

pkgs/by-name/ti/tigerjython/package.nix
```diff
@@ -0,0 +1,88 @@
...
+in
+
+stdenv.mkDerivation (finalAttrs: {
+  pname = "tigerjython";
+  version = "2.39";
+
+  src = fetchurl {
+    url = "https://tigerjython.ch/user/pages/download/TigerJython.tar.gz";
+    hash = "sha256-rfksd3faLHaXS79PvbQEYLdVaO5/6/4Ytp23amya5zI=";
+  };
```
TomaSajt, line 25: Hey, thanks for taking your time to contribute!

I don't think a package having an unstable download link would be accepted inside nixpkgs. Once the website updates to a later version of the package, that will be fetched instead, leading to some hash mismatch errors.
If there is a way to download specific versions, then great, please do that. Otherwise, I don't know what can be done.

TomaSajt:
I realized that this problem is not that uncommon across nixpkgs,
and a good solution I found was using http://web.archive.org/ to cache a download link.


I did that here: http://web.archive.org/web/20240119124245/https://tigerjython.ch/user/pages/download/TigerJython.tar.gz

You'll have to generate another snapshot when there's a new release, but this is much easier to maintain than a custom git repo.

emilazy:
web.archive.org is used extensively throughout the tree (112 files in `pkgs/` matching it), and doesn’t carry as many of the supply chain security risks as using a binary from an unreviewed maintainer‐controlled source. The Internet Archive regularly serves files of far greater size than that through their collections, and you can mark this package as `unfreeRedistributable` ([“Any such service or product may be shared and distributed freely, provided the TJ Group is identified as the original copyright holder.”](https://tigerjython.ch/en/legal/terms)) to get nixos.org to cache it, so I think that’s a far preferable solution.

TomaSajt:
> and you can mark this package as `unfreeRedistributable` ([“Any such service or product may be shared and distributed freely, provided the TJ Group is identified as the original copyright holder.”](https://tigerjython.ch/en/legal/terms)) to get nixos.org to cache it, so I think that’s a far preferable solution.

AFAICT unfreeRedistributable packages are not built on hydra...
https://github.com/NixOS/nixpkgs/issues/83884

emilazy:
Sorry, you’re right; I got my wires crossed. Won’t tarballs.nixos.org still cache the source derivation though? I admit I haven’t spent much time working with non‐free packages, so I could be totally wrong here. In any case I’m pretty sure the Internet Archive can take the load.

TomaSajt (changes requested):
I've requested some changes.

Also, could you squash your commits into 2 commits?
- one for adding yourself as a maintainer
- and one for adding the package itself

https://github.com/NixOS/nixpkgs/blob/master/CONTRIBUTING.md#commit-conventions
https://github.com/NixOS/nixpkgs/blob/master/pkgs/README.md#commit-conventions

TomaSajt:
> Thank you. I incorporated the change requests. Can you give me a hint how I can squash the 3 commits into 2?

You can run `git reset HEAD~3` to undo the last 3 commits, but keep the changes. You can then create the two commits from there.
Follow the commit conventions I linked previously.

pkgs/by-name/ti/tigerjython/package.nix
```diff
@@ -0,0 +1,87 @@
...
+  '';
+
+  postInstall = ''
+    install -Dm444 ${icon} $out/share/icons/hicolor/64x64/apps/tigerjython.png
+  '';
+
+  # Tests need a DISPLAY
+  doCheck = false;
+
+  meta = with lib; {
```
vigress8, line 72:
```suggestion
  meta = {
```
Generally better to keep usage of `with` to a minimum

pkgs/by-name/ti/tigerjython/package.nix
```diff
@@ -0,0 +1,87 @@
...
+    description = "Simple development environment for programming in Python";
+    longDescription = ''
+      Designing, coding, and amazing. TigerJython offers everything you need
+      to go from Python programming beginner to professional.
+      You will find a wide variety of tutorials and can get started right away
+      in programming environments specially developed for you.
+    '';
+    license = licenses.unfreeRedistributable;
+    maintainers = with maintainers; [ rcmlz ];
+    platforms = platforms.all;
```
vigress8, line 84:
```suggestion
    license = lib.licenses.unfreeRedistributable;
    maintainers = with lib.maintainers; [ rcmlz ];
    platforms = lib.platforms.all;
```

ghost (changes requested):
Per https://github.com/NixOS/rfcs/pull/166, format the file with
```
nix run nixpkgs#nixfmt-rfc-style -- pkgs/by-name/ti/tigerjython/package.nix
```

vigress8:
Since updating this seems somewhat cumbersome, I wonder if an updateScript could be written to handle it.

vigress8:
> My current view is that the risk of failures in webscraping is higher than the effort of a yearly (before Autumn term) manual update.

Fair enough, it's not quite fit for automation. But perhaps a script meant to run manually, and the result checked by a human, would be useful.

ghost (approved):

pkgs/by-name/ti/tigerjython/package.nix
```diff
@@ -0,0 +1,97 @@
...
+}:
+let
+  icon = fetchurl {
+    # In case tigerjython/tjinstall becomes unavailable, use this url - see comment for src in MkDerivation
+    #url = "https://web.archive.org/web/20240623120114/https://raw.githubusercontent.com/tigerjython/tjinstall/master/tjlogo64.png";
+    url = "https://raw.githubusercontent.com/tigerjython/tjinstall/611c56d4e765731883656a5c4b71209d72b5ab74/tjlogo64.png";
+    hash = "sha256-tw3uDWLtcMHYmN6JGsEvVKLgI09v5DF27V2+OF9Z5tA=";
+  };
+in
+stdenv.mkDerivation (finalAttrs: {
```
TomaSajt, line 18:
```suggestion
stdenvNoCC.mkDerivation (finalAttrs: {
```
Doing this removes the C/C++ toolchain from being available in the build environment.
This *might* prevent some unnecessary rebuilds.
However, usually, if `gcc` changes, almost everything else gets rebuilt, so this probably doesn't help that much.

pkgs/by-name/ti/tigerjython/package.nix
```diff
@@ -0,0 +1,97 @@
...
+      desktopName = "TigerJython";
+      comment = "The Python IDE for beginners";
+      categories = [ "Education" ];
+    })
+  ];
+
+  dontConfigure = true;
+  dontBuild = true;
+  # Tests need a DISPLAY
+  doCheck = false;
```
TomaSajt, line 56:
```suggestion
```

What was the original reason behind adding `doCheck = false`?
`doCheck` is false by default if you use `mkDerivation`, and there are no tests to be run anyway...

pkgs/by-name/ti/tigerjython/package.nix
```diff
@@ -0,0 +1,97 @@
...
+
+    runHook postInstall
+  '';
+
+  postInstall = ''
+    install -Dm444 ${icon} $out/share/icons/hicolor/64x64/apps/tigerjython.png
+  '';
+
+  meta = {
+    homepage = "https://www.tigerjython.com";
```
TomaSajt, line 83:
```suggestion
    homepage = "https://www.tigerjython.ch";
```

Might be better to always use the same TLD. `.com`, `.de` and `.ch` all seem to point to the same site.

TomaSajt (changes requested):
The PR looks pretty complete now, though I still have some small suggestions.

TomaSajt:
Also I'm a bit concerned about the long download times.
If someone used `nix-collect-garbage`, the tarball would probably get cleared out, since it's a build-time only dependency. Then, if the package needed to be rebuilt, it would need to fetch the tarball again, which is painfully long.
Let's hope whoever uses this package uses the stable branches, where things don't get rebuilt so often.

*It might be possible to hack together something where we symlink the tarball's files instead of copying them into the main package. This way they should not get cleared. What do you think, is it worth trying?*

TomaSajt:
I've though about the download speed a bit more and I think it's fine as it is for now.

If someone has an issue with long download times when rebuilding, we can ask them to just manually download the tar.gz file and do `(tigerjython.overrideAttrs { src = ./TigerJython.tar.gz; })`

TomaSajt (approved):
LGTM for the most part.
If some issue comes up in the future, we can fix it then.

JohnRTitor:
Result of `nixpkgs-review pr 316431` run on x86_64-linux [1](https://github.com/Mic92/nixpkgs-review)
<details>
  <ul>
    <li>tigerjython</li>
  </ul>
</details>

emilazy (approved):
I’m sorry this stalled out. I couldn’t merge things back when I left my original comments but I can now. This looks good to me and I’m happy to add it as‐is. Thank you for your contribution!

PR #316452, opened 2024-06-01
stash: init at 0.25.1

Golo300:
## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/by-name/st/stash/package.nix
+++ b/pkgs/by-name/st/stash/package.nix
@@ -1,0 +1,50 @@
+{
+  lib,
+  stdenv,
+  fetchurl,
+}:
+let
+
+  version = "0.25.1";
+
+  sources = {
+    x86_64-linux = {
+      url = "https://github.com/stashapp/stash/releases/download/v${version}/stash-linux";
+      hash = "sha256-Rb4x6iKx6T9NPuWWDbNaz+35XPzLqZzSm0psv+k2Gw4=";
+    };
+    aarch64-linux = {
+      url = "https://github.com/stashapp/stash/releases/download/v${version}/stash-linux-arm64v8";
+      hash = "sha256-6qPyIYKFkhmBNO47w9E91FSKlByepBOnl0MNJighGSc=";
+    };
+    x86_64-darwin = {
+      url = "https://github.com/stashapp/stash/releases/download/v${version}/stash-macos";
+      hash = "sha256-W8+rgqWUDTOB8ykGO2GL9tKEjaDXdx9LpFg0TAtJsxM=";
+    };
+  };
+in
+stdenv.mkDerivation (finalAttrs: {
+  inherit version;
+
+  pname = "stash";
+
+  src = fetchurl { inherit (sources.${stdenv.system}) url hash; };
+
+  dontUnpack = true;
+
+  installPhase = ''
+    runHook preInstall
+
+    install -Dm755 $src $out/bin/stash
+
+    runHook postInstall
+  '';
+
+  meta = with lib; {
+    description = "Stash is a self-hosted porn app.";
+    homepage = "https://github.com/stashapp/stash";
+    license = licenses.agpl3Only;
+    maintainers = with maintainers; [ Golo300 ];
+    platforms = builtins.attrNames sources;
+    mainProgram = "stash";
+  };
+})
--- a/maintainers/maintainer-list.nix
+++ b/maintainers/maintainer-list.nix
@@ -7709,6 +7709,12 @@
       fingerprint = "0BAF 2D87 CB43 746F 6237  2D78 DE60 31AB A0BB 269A";
     }];
   };
+  Golo300 = {
+    email = "lanzingertm@gmail.com";
+    github = "Golo300";
+    githubId = 58785758;
+    name = "Tim Lanzinger";
+  };
   Gonzih = {
     email = "gonzih@gmail.com";
     github = "Gonzih";


pkgs/by-name/st/stash/package.nix
```diff
@@ -0,0 +1,30 @@
+{ lib, stdenv, fetchurl, }:
+
+stdenv.mkDerivation rec {
```
drupol, line 3: I think we should use the `finalAttrs` pattern here.

```nix
mkDerivation rec {
   foo = 1;
   bar = foo + 1;
}
```

pkgs/by-name/st/stash/package.nix
```diff
@@ -0,0 +1,30 @@
...
+  pname = "stash";
+  version = "0.25.1";
+
+  src = fetchurl {
+    url = "https://github.com/stashapp/stash/releases/download/v${version}/stash-linux";
+    sha256 = "sha256-Rb4x6iKx6T9NPuWWDbNaz+35XPzLqZzSm0psv+k2Gw4=";
+  };
+
+  unpackPhase = "true";
```
drupol, line 12: Not needed

pkgs/by-name/st/stash/package.nix
```diff
@@ -0,0 +1,30 @@
+{ lib, stdenv, fetchurl, }:
+
+stdenv.mkDerivation rec {
+  pname = "stash";
+  version = "0.25.1";
+
+  src = fetchurl {
+    url = "https://github.com/stashapp/stash/releases/download/v${version}/stash-linux";
+    sha256 = "sha256-Rb4x6iKx6T9NPuWWDbNaz+35XPzLqZzSm0psv+k2Gw4=";
```
drupol, line 9: Prefer `hash = ...` to `sha256 = ...` Both currently work, but the former will make it easier if the algorithm ever needs to change.

pkgs/by-name/st/stash/package.nix
```diff
@@ -0,0 +1,30 @@
...
+
+  installPhase = ''
+    runHook preInstall
+
+    install -Dm755 $src $out/bin/stash
+
+    runHook postInstall
+  '';
+
+  meta = with lib; {
```
drupol, line 22: Missing `meta.mainProgram`

drupol (changes requested):
Hey hi !

Congrats for your first PR.

I left some comments, let me know if this is clear.

Please split the PR in 2 commits. One commit to add you in the maintainers list, the second commit to add `stash`.

Also, I will ask to use `nixfmt-rfc-style` to format the new files you're introducing.

pkgs/by-name/st/stash/package.nix
Golo300: Thanks for the input. The sources helped a lot. A adapted everything according to the given example.

pkgs/by-name/st/stash/package.nix
Golo300: It failed because the unpack phase is not needed. I added the `dontUnpack` attribute now and set it to `true` since there is no unpacking needed. Is that a better approach or do you suggest something else?

pkgs/by-name/st/stash/package.nix
Golo300: Done

pkgs/by-name/st/stash/package.nix
Golo300: Done

drupol (approved):

drupol (changes requested):
I think the current state of thing is wrong. It is not possible that the same linux binary works on x86 and aarch64.

By looking further, I notice that upstream publish a binary for each platforms, and also MacOS !

I think the current state of thing is wrong. It is not possible that the same linux binary works on x86 and aarch64.

By looking further, I notice that upstream publish a binary for each platforms, and also MacOS !

image

Therefore, I will kindly ask you to make the changes so that it is available for all the platforms.

Here's an example on how you could do that: https://github.com/NixOS/nixpkgs/blob/518daacfe2ee732432b4a7dac19b72eda037d44d/pkgs/applications/networking/cluster/lens/default.nix#L11

Let me know if you need some help.

drupol (approved):
Thanks!

pkgs/by-name/st/stash/package.nix
```diff
@@ -0,0 +1,52 @@
...
+    };
+  };
+in
+stdenv.mkDerivation (finalAttrs: {
+  inherit version;
+
+  pname = "stash";
+
+  src = fetchurl {
+    inherit (sources.${stdenv.system} or (throw "Unsupported system: ${stdenv.system}")) url hash;
```
isabelroses, line 31: You shouldn't need to throw an error here since `meta.platforms` will prevent platforms that are unsupported from trying to build this.
```suggestion
    inherit (sources.${stdenv.system}) url hash;
```

drupol: Indeed, nice.

Golo300: Thanks, Done

isabelroses (approved):

drupol:
Congrats with your first PR ! Keep them coming!

Golo300:
Thanks for your great support! It helped me a lot.

PR #316456, opened 2024-06-01
k3s_1_27: remove

superherointj:
As current stable supports `k3s_1_30`:
* https://github.com/NixOS/nixpkgs/tree/release-24.05/pkgs/applications/networking/cluster/k3s

Removing: k3s_1_27

CC @yajo @euank @wrmilling @wegank @Mic92 

k3s_1_27 is EOL at 2024-06-28

(This PR should not be back-ported.)

Scheduled for merge on 2024-06-28


Aleksanaa:
> And yes, I know these versions are currently supported in upstream. As unstable is the new 24.11-stable. These versions have to be removed because when 24.11-stable is released they're deprecated.

I kind of don't understand this argument, since it hasn't been deprecated by upstream yet, why do we need to remove them early because it will be deprecated in the next release (six months later)? Can't we just remove them when it's deprecated?

superherointj:
> I kind of don't understand this argument, since it hasn't been deprecated by upstream yet, why do we need to remove them early because it will be deprecated in the next release (six months later)? Can't we just remove them when it's deprecated?

* https://github.com/NixOS/nixpkgs/tree/master/pkgs/applications/networking/cluster/k3s#k3s-versions

* https://github.com/NixOS/nixpkgs/issues/313252#issuecomment-2121309843

wrmilling (approved):
Changes and merge plan LGTM.

wrmilling (changes requested):
Reviewed too quick first time, just need to remove the 1_27 block from default.nix as well. Otherwise LGTM

```
# 1_27 can be built with the same builder as 1_30
  k3s_1_27 = common (
  ...
  ) extraArgs;
```

wrmilling (approved):

Issue #308324, opened 2024-05-01
Using absolute path in desktop entries could cause confusion with xdg autostart


Aleksanaa:
Most upstream provided or auto generated desktop entries uses `Exec=foo` instead of `Exec=$out/bin/foo`. But when manually modifying or creating desktop entries, `Exec` is often set to an absolute path, like `$out/bin/foo`, which seems reasonable.

However, applications that are set to be launched when login, by xdg autostart, will have their desktop entries moved to `~/.config/autostart`. This will cause a problem: when users remove the package from their configuration, they will be surprised to find that the desktop file will still run the next time they log in, because the `Exec` in it remains valid, unless they manually remove the desktop entry or perform garbage collection.

Another problem that can arise with `Exec=$out/bin/foo` is that when people use `symlinkJoin` to wrap a package, they have to rewrite a desktop file, as the old one still points to the unwrapped binary.

I would like to open an issue here to first discuss whether we should set `Exec=foo` instead of `Exec=$out/bin/foo`. If this is the case, I'll rename it to `Tracking: ...`.

Thanks @VergeDX for reporting the issue and @OPNA2608 for pointing out the case, and @getchoo for giving me the inspiration in https://github.com/NixOS/nixpkgs/pull/308113#discussion_r1585819842 😄 

ping @jtojnar for opinions

-----
Add a :+1: [reaction] to [issues you find important].

[reaction]: https://github.blog/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/
[issues you find important]: https://github.com/NixOS/nixpkgs/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions-%2B1-desc

jtojnar:
Yeah, we generally do not want absolute paths in application launcher desktop files precisely for that reason. All methods that install a package so that the launcher is `XDG_DATA_DIRS` will also put the programs on `PATH` so having to do a `PATH` lookup is not an issue.

The only potentially problematic case I can imagine is a GUI app package with a `bin` output. In that case installing `<package>` will only link the `bin` output but the launcher in `out` would not be installed but that is really the opposite issue. We could fix that by moving `share/applications` to `${!outputBin}`  in `multiple-outputs.sh` hook but it is probably extremely rare anyway.

Aleksanaa:
[...]

PR #316498, opened 2024-06-01
session-desktop: file desktop exec path

joshlupo:
## Description of changes
Session messenger installs and runs from the terminal but the desktop file for Gnome etc. was broken since it referred to a file that did not exist.

### DIFF
--- a/pkgs/applications/networking/instant-messengers/session-desktop/default.nix
+++ b/pkgs/applications/networking/instant-messengers/session-desktop/default.nix
@@ -33,7 +33,7 @@ stdenvNoCC.mkDerivation {
       name = "Session";
       desktopName = "Session";
       comment = "Onion routing based messenger";
-      exec = "${appimage}/bin/session-desktop-${version}";
+      exec = "session-desktop";
       icon = "${appimage-contents}/session-desktop.png";
       terminal = false;
       type = "Application";


pkgs/applications/networking/instant-messengers/session-desktop/default.nix
```diff
@@ -33,7 +33,7 @@ stdenvNoCC.mkDerivation {
       name = "Session";
       desktopName = "Session";
       comment = "Onion routing based messenger";
-      exec = "${appimage}/bin/session-desktop-${version}";
+      exec = "${appimage}/bin/session-desktop";
```
Aleksanaa, line 36:
```suggestion
      exec = "session-desktop";
```

See https://github.com/NixOS/nixpkgs/issues/308324

pkgs/applications/networking/instant-messengers/session-desktop/default.nix
```diff
@@ -33,7 +33,7 @@ stdenvNoCC.mkDerivation {
       name = "Session";
       desktopName = "Session";
       comment = "Onion routing based messenger";
-      exec = "${appimage}/bin/session-desktop-${version}";
+      exec = "session-desktop";
```
drupol, line 36: How about keeping the same structure as before? Something like this?

```suggestion
      exec = "${appimage}/bin/session-desktop";
```

drupol (changes requested):

pkgs/applications/networking/instant-messengers/session-desktop/default.nix
Aleksanaa: I requested this change above

drupol: Oooh sorry !!! Ignore me then :)

drupol (approved):

drupol (dismissed):
Please update the PR title and also the commit log message.

It should be:

```
session-desktop: file desktop exec path
```

drupol:
Thank you!

PR #316591, opened 2024-06-02
doc/languages-frameworks/python: update

natsukium:
## Description of changes

Removed old description of build/test in documentation and added description of build according to PEP 517 and test using pytestCheckHook.

Also updated the `contributing` section.
This section is something I check when reviewing new packages or package updates and is often a boring and unproductive suggestion.
I think that almost all simple packages can be addressed.
It should be useful to contributors as well as reviewers who do not specialize in python packages.

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/doc/languages-frameworks/python.section.md
+++ b/doc/languages-frameworks/python.section.md
@@ -78,6 +78,7 @@ The following is an example:
 , fetchPypi
 
 # build-system
+, setuptools
 , setuptools-scm
 
 # dependencies
@@ -107,6 +108,7 @@ buildPythonPackage rec {
   '';
 
   build-system = [
+    setuptools
     setuptools-scm
   ];
 
@@ -294,7 +296,6 @@ python3Packages.buildPythonApplication rec {
 
   build-system = with python3Packages; [
     setuptools
-    wheel
   ];
 
   dependencies = with python3Packages; [
@@ -831,7 +832,6 @@ building Python libraries is [`buildPythonPackage`](#buildpythonpackage-function
 , buildPythonPackage
 , fetchPypi
 , setuptools
-, wheel
 }:
 
 buildPythonPackage rec {
@@ -846,7 +846,6 @@ buildPythonPackage rec {
 
   build-system = [
     setuptools
-    wheel
   ];
 
   # has no tests
@@ -905,7 +904,6 @@ with import <nixpkgs> {};
 
       build-system = [
         python311.pkgs.setuptools
-        python311.pkgs.wheel
       ];
 
       # has no tests
@@ -958,7 +956,7 @@ order to build [`datashape`](https://github.com/blaze/datashape).
 , fetchPypi
 
 # build dependencies
-, setuptools, wheel
+, setuptools
 
 # dependencies
 , numpy, multipledispatch, python-dateutil
@@ -979,7 +977,6 @@ buildPythonPackage rec {
 
   build-system = [
     setuptools
-    wheel
   ];
 
   dependencies = [
@@ -1016,7 +1013,6 @@ when building the bindings and are therefore added as [`buildInputs`](#var-stden
 , buildPythonPackage
 , fetchPypi
 , setuptools
-, wheel
 , libxml2
 , libxslt
 }:
@@ -1033,7 +1029,6 @@ buildPythonPackage rec {
 
   build-system = [
     setuptools
-    wheel
   ];
 
   buildInputs = [
@@ -1076,7 +1071,6 @@ therefore we have to set `LDFLAGS` and `CFLAGS`.
 
 # build dependencies
 , setuptools
-, wheel
 
 # dependencies
 , fftw
@@ -1098,7 +1092,6 @@ buildPythonPackage rec {
 
   build-system = [
     setuptools
-    wheel
   ];
 
   buildInputs = [
@@ -1465,7 +1458,6 @@ We first create a function that builds `toolz` in `~/path/to/toolz/release.nix`
 , buildPythonPackage
 , fetchPypi
 , setuptools
-, wheel
 }:
 
 buildPythonPackage rec {
@@ -1480,7 +1472,6 @@ buildPythonPackage rec {
 
   build-system = [
     setuptools
-    wheel
   ];
 
   meta = {
--- a/doc/languages-frameworks/python.section.md
+++ b/doc/languages-frameworks/python.section.md
@@ -869,7 +869,7 @@ buildPythonPackage rec {
 What happens here? The function [`buildPythonPackage`](#buildpythonpackage-function) is called and as argument
 it accepts a set. In this case the set is a recursive set, `rec`. One of the
 arguments is the name of the package, which consists of a basename (generally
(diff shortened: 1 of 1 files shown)


doc/languages-frameworks/python.section.md
```diff
@@ -1201,9 +1199,11 @@ been removed, in this case, it's recommended to use `pytestCheckHook`.
 
 #### Using pytestCheckHook {#using-pytestcheckhook}
 
-`pytestCheckHook` is a convenient hook which will substitute the setuptools
-`test` command for a [`checkPhase`](#ssec-check-phase) which runs `pytest`. This is also beneficial
+`pytestCheckHook` is a convenient hook which will substitute
+a [`checkPhase`](#ssec-check-phase) which runs `pytest`. This is also beneficial
```
mweinelt, line 1207: > which will set up (or configure) `checkPhase` to run `pytest`.

I think that's simpler.

doc/languages-frameworks/python.section.md
```diff
@@ -1929,16 +1931,15 @@ Both are also exported in `nix-shell`.
 It is recommended to test packages as part of the build process.
 Source distributions (`sdist`) often include test files, but not always.
 
-By default the command `python setup.py test` is run as part of the
-[`checkPhase`](#ssec-check-phase), but often it is necessary to pass a custom [`checkPhase`](#ssec-check-phase). An
-example of such a situation is when `py.test` is used.
+Previously, by default the command `python setup.py test` is run as part of the
+[`checkPhase`](#ssec-check-phase), but now it is necessary to pass a custom [`checkPhase`](#ssec-check-phase).
+In almost all cases, `pytestCheckHook` is used.
```
mweinelt, line 1941: Were still defaulting to that, aren"t we? Also you don't need to pass a custom `checkPhase`, but rather a hook to set up a certain test framework.

doc/languages-frameworks/python.section.md
```diff
@@ -1041,6 +1041,14 @@ buildPythonPackage rec {
     libxslt
   ];
 
+  # tests are meant to be ran "in-place" in the same directory as src
+  doCheck = false;
```
mweinelt, line 1041: Commit message says "enable check phase" :thinking:

doc/languages-frameworks/python.section.md
```diff
@@ -1983,14 +1985,24 @@ The following rules are desired to be respected:
 * Python applications live outside of `python-packages.nix` and are packaged
   with [`buildPythonApplication`](#buildpythonapplication-function).
 * Make sure libraries build for all Python interpreters.
-* By default we enable tests. Make sure the tests are found and, in the case of
+  If it fails to build on some Python versions, consider disabling them by setting `disable = pythonAtLeast "3.x"` along with a comment.
+* The two parameters, `pyproject` and `build-system` are set to avoid the legacy setuptools/distutils build.
+* Packages with versioning like `pydantic_1` are not included in the `dependencies`.
+  Such packages should not be used in libraries, because propagation causes conflicts.
```
mweinelt, line 1991: I think the takeaway could be something like this:

Only unversioned attributes (e.g. `pydantic`, but not `pypdantic_1`)  can be included in `dependencies`, since due to `PYTHONPATH` limitations we can only ever support a single version for libraries without running into duplicate module name conflicts.

doc/languages-frameworks/python.section.md
```diff
@@ -1983,14 +1985,24 @@ The following rules are desired to be respected:
...
+* The version restrictions of the dependencies are relaxed by [`pythonRelaxDepsHook`](#using-pythonrelaxdepshook).
```
mweinelt, line 1992:
```suggestion
* The version restrictions of `dependencies` can be relaxed by [`pythonRelaxDepsHook`](#using-pythonrelaxdepshook).
```

doc/languages-frameworks/python.section.md
```diff
@@ -1983,14 +1985,24 @@ The following rules are desired to be respected:
...
+* Make sure the tests are enabled with [`pytestCheckHook`](#using-pytestcheckhook) or something and, in the case of
```
mweinelt, line 1993:
```suggestion
* Make sure the tests are enabled using for example [`pytestCheckHook`](#using-pytestcheckhook) and, in the case of
```

doc/languages-frameworks/python.section.md
```diff
@@ -1983,14 +1985,24 @@ The following rules are desired to be respected:
...
   libraries, are passing for all interpreters. If certain tests fail they can be
   disabled individually. Try to avoid disabling the tests altogether. In any
   case, when you disable tests, leave a comment explaining why.
+* `pythonImportsCheck` is set. This is still a valid check even if `pytestCheckHook` is set.
```
mweinelt, line 1997:
```suggestion
* `pythonImportsCheck` is set. This is still a good smoke test even if `pytestCheckHook` is set.
```

https://en.wikipedia.org/wiki/Smoke_testing_(software)

mweinelt (commented):
Thanks for taking a stab at the docs, they don't receive enough attention sadly.

doc/languages-frameworks/python.section.md
natsukium: It is still `doCheck = true` by default, but the hook (python setup.py test) is not provided unless `format = "setuptools"`. This means that unless we add `pytestCheckHook` to the input or provide checkPhase, test will not run in the current builder.

How about this?
> Previously, by default the command `python setup.py test` is run as part of the [`checkPhase`](#ssec-check-phase), but now it is necessary to pass a test framework hook like `pytestCheckHook`.
It may be necessary to provide a custom checkPhase for packages that do not use a common framework.

doc/languages-frameworks/python.section.md
natsukium: I guess the test ran in my imaginary machine.🤷

doc/languages-frameworks/python.section.md
mweinelt: > By default the `checkPhase` executes the deprecated `python setup.py test` command.

This is the status quo, tests will still try to run `python setup.py test`, when no hook is given. We *could* however change that in a huge treewide change.

> The best practice today is to pass a test hook (e.g. `pytestCheckHook`, `unittestCheckHook`) into `nativeCheckInputs`, that will reconfigure the `checkPhase` to use that particular test framework. Occasionaly it may  be required to provide a custom `checkPhase`, when the package does not rely on a common framework.

natsukium: Do you mean that `python setup.py test` will be executed since `format = "setuptools"` is currently the default?
If so, I agree with you.

mweinelt: Oh right, that is depended on the format :facepalm:

natsukium:
There are no other comments, so I'll merge it.
Thanks for your check.

PR #316668, opened 2024-06-02
lib, treewide: introduce `repoRevToName` and use it to cleanup most `fetch*` functions

oxij:
This a first half of #49862. Basically, this is a noop cleanup that keeps all the `fetch*` names as they are now, while introducing the new `config.fetchedSourceNameDefault` option.

# Changes

This patch adds `lib.repoRevToName` function that generalizes away most of the code used for derivation name generation by `fetch*` functions (`fetchzip`, `fetchFromGitHub`, etc, except those which are delayed until latter commits for mass-rebuild reasons).

It's first argument controls how the resulting name will look (see below).

Since `lib` has no equivalent of Nixpkgs' `config`, this patch adds `config.fetchedSourceNameDefault` option to Nixpkgs and then re-exposes `lib.repoRevToName config.fetchedSourceNameDefault` expression as `pkgs.repoRevToNameMaybe` which is then used in `fetch*` derivations.

The result is that different values of `config.fetchedSourceNameDefault` now control how the `src` derivations produced by `fetch*` functions are to be named, e.g.:

- `fetchedSourceNameDefault = "source"` (the default):

        $ nix-instantiate -A fuse.src
        /nix/store/<hash>-source.drv

- `fetchedSourceNameDefault = "versioned"`:

        $ nix-instantiate -A fuse.src
        /nix/store/<hash>-libfuse-2.9.9-source.drv

- `fetchedSourceNameDefault = "full"`:

        $ nix-instantiate -A fuse.src
        /nix/store/<hash>-libfuse-2.9.9-github-source.drv

See documentation of `config.fetchedSourceNameDefault` for more info.

### DIFF
--- a/lib/sources.nix
+++ b/lib/sources.nix
@@ -3,7 +3,7 @@
 
 # Tested in lib/tests/sources.sh
 let
-  inherit (builtins)
+  inherit (lib.strings)
     match
     split
     storeDir
@@ -403,6 +403,101 @@ let
       };
     };
 
+  # urlToName : (URL | Path | String) -> String
+  #
+  # Transform a URL (or path, or string) into a clean package name.
+  urlToName =
+    url:
+    let
+      inherit (lib.strings) stringLength;
+      base = baseNameOf (lib.removeSuffix "/" (lib.last (lib.splitString ":" (toString url))));
+      # chop away one git or archive-related extension
+      removeExt =
+        name:
+        let
+          matchExt = match "(.*)\\.(git|tar|zip|gz|tgz|bz|tbz|bz2|tbz2|lzma|txz|xz|zstd)$" name;
+        in
+        if matchExt != null then lib.head matchExt else name;
+      # apply function f to string x while the result shrinks
+      shrink =
+        f: x:
+        let
+          v = f x;
+        in
+        if stringLength v < stringLength x then shrink f v else x;
+    in
+    shrink removeExt base;
+
+  # shortRev : (String | Integer) -> String
+  #
+  # Given a package revision (like "refs/tags/v12.0"), produce a short revision ("12.0").
+  shortRev =
+    rev:
+    let
+      baseRev = baseNameOf (toString rev);
+      matchHash = match "[a-f0-9]+" baseRev;
+      matchVer = match "([A-Za-z]+[-_. ]?)*(v)?([0-9.]+.*)" baseRev;
+    in
+    if matchHash != null then
+      builtins.substring 0 7 baseRev
+    else if matchVer != null then
+      lib.last matchVer
+    else
+      baseRev;
+
+  # revOrTag : String -> String -> String
+  #
+  # Turn git `rev` and `tag` pair into a revision usable in `repoRevToName*`.
+  revOrTag =
+    rev: tag:
+    if tag != null then
+      tag
+    else if rev != null then
+      rev
+    else
+      "HEAD";
+
+  # repoRevToNameFull : (URL | Path | String) -> (String | Integer | null) -> (String | null) -> String
+  #
+  # See `repoRevToName` below.
+  repoRevToNameFull =
+    repo_: rev_: suffix_:
+    let
+      repo = urlToName repo_;
+      rev = if rev_ != null then "-${shortRev rev_}" else "";
+      suffix = if suffix_ != null then "-${suffix_}" else "";
+    in
+    "${repo}${rev}${suffix}-source";
+
+  # repoRevToName : String -> (URL | Path | String) -> (String | Integer | null) -> String -> String
+  #
+  # Produce derivation.name attribute for a given repository URL/path/name and (optionally) its revision/version tag.
+  #
+  # This is used by fetch(zip|git|FromGitHub|hg|svn|etc) to generate discoverable
+  # /nix/store paths.
+  #
+  # This uses a different implementation depending on the `pretty` argument:
+  #  "source" -> name everything as "source"
+  #  "versioned" -> name everything as "${repo}-${rev}-source"
+  #  "full" -> name everything as "${repo}-${rev}-${fetcher}-source"
+  repoRevToName =
+    kind:
+    # match on `kind` first to minimize the thunk
+    if kind == "source" then
+      (
+        repo: rev: suffix:
+        "source"
+      )
+    else if kind == "versioned" then
+      (
+        repo: rev: suffix:
+        repoRevToNameFull repo rev null
+      )
+    else if kind == "full" then
+      repoRevToNameFull
+    else
+      throw "repoRevToName: invalid kind";
+
 in
 {
 
@@ -430,6 +525,11 @@ in
     cleanSourceFilter
     pathHasContext
     canCleanSource
+
+    urlToName
+    shortRev
+    revOrTag
+    repoRevToName
 
     sourceByRegex
     sourceFilesBySuffices
--- a/pkgs/build-support/fetchgitiles/default.nix
+++ b/pkgs/build-support/fetchgitiles/default.nix
@@ -1,11 +1,15 @@
-{ fetchzip, lib }:
+{
+  fetchzip,
+  repoRevToNameMaybe,
+  lib,
+}:
 
 lib.makeOverridable (
   {
     url,
     rev ? null,
     tag ? null,
-    name ? "source",
+    name ? repoRevToNameMaybe url (lib.revOrTag rev tag) "gitiles",
     ...
   }@args:
 
--- a/pkgs/top-level/config.nix
+++ b/pkgs/top-level/config.nix
@@ -59,6 +59,52 @@ let
       default = false;
     };
 
+    fetchedSourceNameDefault = mkOption {
+      type = types.uniq (
+        types.enum [
+          "source"
+          "versioned"
+          "full"
+        ]
+      );
+      default = "source";
+      description = ''
+        This controls the default derivation `name` attribute set by the
+        `fetch*` (`fetchzip`, `fetchFromGitHub`, etc) functions.
+
+        Possible values and the resulting `.name`:
+
+        - `"source"` -> `"source"`
+        - `"versioned"` -> `"''${repo}-''${rev}-source"`
+        - `"full"` -> `"''${repo}-''${rev}-''${fetcherName}-source"`
+
+        The default `"source"` is the best choice for minimal rebuilds, it
+        will ignore any non-hash changes (like branches being renamed, source
+        URLs changing, etc) at the cost of `/nix/store` being easily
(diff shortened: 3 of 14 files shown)


lib/sources.nix
```diff
@@ -254,6 +254,64 @@ let
...
     };
 
+  # urlToName : (URL | Path | String) -> String
+  #
+  # Transform a URL (or path, or string) into a clean package name.
+  urlToName = url_:
+    let
+      inherit (builtins) stringLength;
+      url = toString url_;
+      base = baseNameOf (lib.removeSuffix "/" (lib.last (lib.splitString ":" url)));
```
AndersonTorres, line 264:
```suggestion
      base = baseNameOf (lib.removeSuffix "/" (lib.last (lib.splitString ":" (toString url))));
```

lib/sources.nix
```diff
@@ -254,6 +254,64 @@ let
       outPath = builtins.path { inherit filter name; path = origSrc; };
     };
 
+  # urlToName : (URL | Path | String) -> String
+  #
+  # Transform a URL (or path, or string) into a clean package name.
```
AndersonTorres, line 259: Maybe I can be wrong, but the general term is URI.

lib/sources.nix
```diff
@@ -254,6 +254,64 @@ let
...
+      shrink = f: x: let v = f x; in if stringLength v < stringLength x then shrink f v else x;
+    in shrink removeExt base;
+
+  # shortRev : (String | Integer) -> String
+  #
+  # Given a package revision (like "refs/tags/v12.0"), produce a short revision ("12.0").
+  shortRev = rev_:
+    let
+      rev = toString rev_;
+      baseRev = baseNameOf rev;
```
AndersonTorres, line 279: No reason to do this other way:
```suggestion
      baseRev = baseNameOf (toString rev);
```

lib/sources.nix
```diff
@@ -254,6 +254,64 @@ let
...
+
+  # repoRevToName : (Bool | String) -> (URL | Path | String) -> (String | Integer | null) -> String -> String
+  #
+  # Produce derivation.name attribute for a given repository URL/path/name and (optionally) its revision/version tag.
+  #
+  # This is used by fetch(zip|git|FromGitHub|hg|svn|etc) to generate discoverable
+  # /nix/store paths.
+  #
+  # This uses a different implementation depending on the `pretty` argument:
+  #  false  -> name everything as "source"
```
AndersonTorres, line 305: To be honest I do not like dual-type attributes. This is a serious footgun in a duck-typed language.

And let's remember Nix is not stellar on the error report department:
https://discourse.nixos.org/t/rant-finding-errors-in-module-configs-after-incompatible-upgrades-is-dreadful/46375

Either is everything a string, say ` format ?  "short", "medium", "long"` or we stick to only two options (and then we can use a Boolean).

AndersonTorres (changes requested):

pkgs/top-level/config.nix
```diff
@@ -50,6 +50,39 @@ let
       default = false;
     };
 
+    nameSourcesPrettily =  mkOption {
```
AndersonTorres, line 53:
```suggestion
    usePrettierSourceNames =  mkOption {
```

lib/sources.nix
oxij: As a thing more general than a URL? Yes. But here it is literally "a URL, or a path, or a string (a package name, usually)".

lib/sources.nix
oxij: But it is not checked by Nix but by the `config.nix` and NixOS typing infra, and the errors it generates are pretty nice, e.g. with `nameSourcesPrettily = "error"`:

```
$ nix-instantiate ./default.nix -A xlife
error: A definition for option `nameSourcesPrettily' is not of type `boolean or value "full" (singular enum)'. Definition values:
- In `nixpkgs.config': "error"
(use '--show-trace' to show detailed location information)
```

pkgs/top-level/config.nix
oxij: Hm, but why? `nameSourcesPrettily` is shorter and means the same thing.

lib/sources.nix
AndersonTorres: Hum, less bad. 

Nonetheless, semantically it is a clear case for an enumeration with three possible states, instead of a weird Boolean-plus-an-extra.

pkgs/top-level/config.nix
AndersonTorres: It's shorter because it is more obscure.
verb-adjective-name is clearer than name-name-adverb.

On the other hand, it depends on the previous tristate vs boolean-plus-state discussion.

philiptaron (commented):
A few curious questions as I read through this PR.

lib/sources.nix
```diff
@@ -254,6 +254,62 @@ let
       outPath = builtins.path { inherit filter name; path = origSrc; };
     };
 
+  # urlToName : (URL | Path | String) -> String
+  #
+  # Transform a URL (or path, or string) into a clean package name.
+  urlToName = url:
+    let
+      inherit (builtins) stringLength;
+      base = baseNameOf (lib.removeSuffix "/" (lib.last (lib.splitString ":" (toString url))));
```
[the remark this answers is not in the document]
oxij: No, the approximation of the basename, even in some really weird cases.

lib/sources.nix
```diff
@@ -254,6 +254,62 @@ let
...
+      # chop away one git or archive-related extension
+      removeExt = name: let
+          matchExt = builtins.match "(.*)\\.(git|tar|zip|gz|tgz|bz|tbz|bz2|tbz2|lzma|txz|xz|zstd)" name;
```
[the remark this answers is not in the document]
oxij: Agreed. Done.

lib/sources.nix
```diff
@@ -254,6 +254,62 @@ let
       outPath = builtins.path { inherit filter name; path = origSrc; };
     };
 
+  # urlToName : (URL | Path | String) -> String
```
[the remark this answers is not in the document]
oxij: Well, with #49862 applied, whole of `nixpkgs` is a set of test for this, kind of.

I would not be against adding them in the future, but adding them now, when the default format is likely to change after (if ever) this gets into master and people start using it, would just ossify these specific decisions. And they are pretty random.

lib/sources.nix
```diff
@@ -254,6 +254,62 @@ let
...
+      shrink = f: x: let v = f x; in if stringLength v < stringLength x then shrink f v else x;
+    in shrink removeExt base;
+
+  # shortRev : (String | Integer) -> String
+  #
+  # Given a package revision (like "refs/tags/v12.0"), produce a short revision ("12.0").
+  shortRev = rev:
+    let
+      baseRev = baseNameOf (toString rev);
+      matchHash = builtins.match "[a-f0-9]+" baseRev;
```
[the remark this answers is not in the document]
oxij: AFAIK, yes.

lib/sources.nix
```diff
@@ -254,6 +254,62 @@ let
...
+      matchVer = builtins.match "([A-Za-z]+[-_. ]?)*(v)?([0-9.]+.*)" baseRev;
+    in
+      if matchHash != null then builtins.substring 0 7 baseRev
```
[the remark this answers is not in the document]
oxij: No idea, this is just the convention `fetchgit` uses.

lib/sources.nix
```diff
@@ -254,6 +254,62 @@ let
...
+      else if matchVer != null then lib.last matchVer
+      else baseRev;
+
+  # repoRevToNameFull : (URL | Path | String) -> (String | Integer | null) -> (String | null) -> String
+  #
+  # See `repoRevToName` below.
+  repoRevToNameFull = repo_: rev_: suffix_:
+    let
+      repo = urlToName repo_;
+      rev = if rev_ != null then "-${shortRev rev_}" else "";
```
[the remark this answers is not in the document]
oxij: Well, yes, but this is **a good default** which covers 99.99% of nixpkgs without any changes to specific packages needed. A packages that needs something else will just set `src.name` to something else.

pkgs/top-level/config.nix
oxij: Changed to `fetchedSourceNameDefault` in the latest version.

lib/sources.nix
```diff
@@ -254,6 +254,62 @@ let
...
+    in shrink removeExt base;
+
+  # shortRev : (String | Integer) -> String
+  #
+  # Given a package revision (like "refs/tags/v12.0"), produce a short revision ("12.0").
+  shortRev = rev:
+    let
+      baseRev = baseNameOf (toString rev);
+      matchHash = builtins.match "[a-f0-9]+" baseRev;
+      matchVer = builtins.match "([A-Za-z]+[-_. ]?)*(v)?([0-9.]+.*)" baseRev;
```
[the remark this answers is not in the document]
oxij: Then, `llama.cpp` will have to set `src.name` explicitly.

pkgs/build-support/fetchgitiles/default.nix
mweinelt, line 1: nixfmt 'pkgs/build-support/fetchgitiles/default.nix'

oxij: Indeed. Thanks!

oxij:
ping @AndersonTorres specifically, since your review requested changes that were since resolved, IMHO.

Also, somewhat relatedly, https://github.com/NixOS/nixpkgs/pull/391322 is a follow-up to https://github.com/NixOS/nixpkgs/commit/66209649883659e071b686d348de491774f57f23 which changed the version without changing the hash, which would not have happened if @mweinelt used this PR with `fetchedSourceNameDefault = "versioned"` set. :]

Such errors, how frequent are they? :thinking:

Just saying.

AndersonTorres (approved):
I would complain about nixfmt not having its own commit. But nothing relevant is being lost.

So, LGTM.

mweinelt:
> Such errors, how frequent are they? 🤔

The cause is limitations in the python bulk updater. These commits are under my name, but `update-python-libraries` created them.

oxij:
@AndersonTorres Thanks! Also, mass-reformat now ate those `nixfmt` changes anyway.

@mweinelt Well, okay, but it would have been caught by this infra.

Also, rebased onto `master` to fix conflicts.

oxij:
Ping. What's holding this up?

oxij:
Thanks, rebase LGTM.

oxij:
Rebased again to fix conflicts. And it is now stuck at Darwin check again. And the previous time it was stuck right there for a month. So, I don't think it's likely this will get any better anytime soon.

arianvp (approved):

oxij:
Thanks @Lassulus!

PR #316680, opened 2024-06-02
release: disallow aliases

jtojnar:
## Description of changes

Having multiple attributes point to the same derivation [makes debugging harder](https://discourse.nixos.org/t/rant-finding-errors-in-module-configs-after-incompatible-upgrades-is-dreadful/46375/3) because you often cannot just grep the canonical attribute name. It is even annoying when multiple aliases are chained.

Having Hydra build aliases is [recognized as redundant](https://github.com/NixOS/nixpkgs/commit/2d0a7c4eeecd22f26eb37f6077a0397c32250375) so let’s do the same with their evaluation.

This commit will make Hydra ignore aliases when evaluating.

A nice benefit of this is that it will allow us to warn users when an attribute is renamed, to assist them with early migration. Since tracing messages during evaluation are not allowed because of Hydra, we can currently only choose between having a silent alias and throwing.

This was last attempted in 2018 but ended up being [reverted](https://github.com/NixOS/nixpkgs/commit/8c025c67d5a151055ef63d2b0d94921604ff0f62) because of widespread use of aliases that was not caught by CI. CI has since been [improved](https://github.com/NixOS/ofborg/pull/645).

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/top-level/aliases.nix
+++ b/pkgs/top-level/aliases.nix
@@ -27,13 +27,6 @@ let
     then removeAttrs alias [ "recurseForDerivations" ]
     else alias;
 
-  # Disabling distribution prevents top-level aliases for non-recursed package
-  # sets from building on Hydra.
-  removeDistribute = alias: with lib;
-    if isDerivation alias then
-      dontDistribute alias
-    else alias;
-
   # Make sure that we are not shadowing something from all-packages.nix.
   checkInPkgs = n: alias:
     if builtins.hasAttr n super
@@ -43,9 +36,8 @@ let
   mapAliases = aliases:
     lib.mapAttrs
       (n: alias:
-        removeDistribute
-          (removeRecurseForDerivations
-            (checkInPkgs n alias)))
+        removeRecurseForDerivations
+          (checkInPkgs n alias))
       aliases;
 in
 
--- a/pkgs/top-level/release-cross.nix
+++ b/pkgs/top-level/release-cross.nix
@@ -17,7 +17,13 @@
 , # Strip most of attributes when evaluating to spare memory usage
   scrubJobs ? true
 , # Attributes passed to nixpkgs. Don't build packages marked as unfree.
-  nixpkgsArgs ? { config = { allowUnfree = false; inHydra = true; }; }
+  nixpkgsArgs ? {
+    config = {
+      allowAliases = false;
+      allowUnfree = false;
+      inHydra = true;
+    };
+  }
 }:
 
 let
--- a/pkgs/top-level/release-cuda.nix
+++ b/pkgs/top-level/release-cuda.nix
@@ -15,7 +15,13 @@
     "x86_64-linux"
   ]
 , # Attributes passed to nixpkgs.
-  nixpkgsArgs ? { config = { allowUnfree = true; inHydra = true; }; }
+  nixpkgsArgs ? {
+    config = {
+      allowAliases = false;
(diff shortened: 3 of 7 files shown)


pkgs/top-level/aliases.nix
```diff
@@ -27,13 +27,6 @@ let
     then removeAttrs alias [ "recurseForDerivations" ]
     else alias;
 
-  # Disabling distribution prevents top-level aliases for non-recursed package
```
jtojnar, line 30: I tried to [remove the `mapAttrs`](https://github.com/jtojnar/nixpkgs/commit/08daaaf2b3d118880165e9ca6336e46420b81322) to avoid having Nix forget the alias values containing `throw` come from the attribute set but it still does not remember:

```ShellSession
$ nix-build -A nixUnstable --show-trace
error:
       … while evaluating the attribute 'unstable'

         at /home/jtojnar/Projects/nixpkgs/pkgs/tools/package-management/nix/default.nix:220:5:

          219|   // {
          220|     unstable = throw "nixVersions.unstable has been removed. For bleeding edge (Nix master, roughly weekly updated) use nixVersions.git, otherwise use nixVersions.latest.";
             |     ^
          221|   }

       … while calling the 'throw' builtin

         at /home/jtojnar/Projects/nixpkgs/pkgs/tools/package-management/nix/default.nix:220:16:

          219|   // {
          220|     unstable = throw "nixVersions.unstable has been removed. For bleeding edge (Nix master, roughly weekly updated) use nixVersions.git, otherwise use nixVersions.latest.";
             |                ^
          221|   }

       error: nixVersions.unstable has been removed. For bleeding edge (Nix master, roughly weekly updated) use nixVersions.git, otherwise use nixVersions.latest.
```

So this will not help with https://discourse.nixos.org/t/rant-finding-errors-in-module-configs-after-incompatible-upgrades-is-dreadful/46375. But shorter stack traces are probably still better.

pkgs/top-level/release-cross.nix
```diff
@@ -17,7 +17,12 @@
 , # Strip most of attributes when evaluating to spare memory usage
   scrubJobs ? true
 , # Attributes passed to nixpkgs. Don't build packages marked as unfree.
-  nixpkgsArgs ? { config = { allowUnfree = false; inHydra = true; }; }
+  nixpkgsArgs ? {
+    config = {
+      allowUnfree = false;
```
lilyinstarlight, line 22: Is there any reason `allowAliases = false` wasn't added for this one?

jtojnar: Oops, must have gotten distracted.

Artturin:
@ofborg eval

Evail fail `       error: cannot connect to socket at '/nix/var/nix/gc-socket/socket': No such file or directory`

Artturin:
@ofborg eval

philiptaron:
I would very much like this PR to land.

vcunat:
So these nixpkgs jobs seem dropped by this PR:
https://hydra.nixos.org/eval/1816553#tabs-removed
EDIT: also the same in NixOS jobset
https://hydra.nixos.org/eval/1816581#tabs-removed

PR #316695, opened 2024-06-02
nixos/bcachefs: skip loading key if we already have it

lheckemann:
## Description of changes

This approach isn't ideal, but I'm not aware of any way of checking if the correct key for a filesystem is already loaded. I've asked about this on the bcachefs IRC and maybe we'll have something at some point, but until then maybe this approach makes sense.

This also still produces multiple prompts if the same filesystem is mounted in multiple places, though unlocking the filesystem succeeds if the passphrase is typed in the first time and the remaining prompts are dismissed with a blank entry. Maybe it would make more sense to have an unlock service per device rather than per mountpoint?

Thoughts?

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/nixos/modules/tasks/filesystems/bcachefs.nix
+++ b/nixos/modules/tasks/filesystems/bcachefs.nix
@@ -90,6 +90,7 @@ let
         ExecCondition = "${pkgs.bcachefs-tools}/bin/bcachefs unlock -c \"${device}\"";
         Restart = "on-failure";
         RestartMode = "direct";
+        RuntimeDirectory = "unlock";
         # Ideally, this service would lock the key on stop.
         # As is, RemainAfterExit doesn't accomplish anything.
         RemainAfterExit = true;
@@ -97,15 +98,20 @@ let
       script = let
         unlock = ''${pkgs.bcachefs-tools}/bin/bcachefs unlock "${device}"'';
         unlockInteractively = ''${config.boot.initrd.systemd.package}/bin/systemd-ask-password --timeout=0 "enter passphrase for ${name}" | exec ${unlock}'';
-      in if useClevis fs then ''
+      in ''
+        # try mounting it temporarily; if that succeeds, the key is
+        # already loaded and we don't need to do any more
+        if unshare -m mount -t bcachefs ${device} -o X-mount.mkdir /run/unlock/tmp ; then
+           exit 0
+        fi
+      '' + (lib.optionalString (useClevis fs) ''
         if ${config.boot.initrd.clevis.package}/bin/clevis decrypt < "/etc/clevis/${device}.jwe" | ${unlock}
         then
           printf "unlocked ${name} using clevis\n"
-        else
-          printf "falling back to interactive unlocking...\n"
-          ${unlockInteractively}
+          exit 0
         fi
-      '' else ''
+        printf "falling back to interactive unlocking...\n"
+      '') + ''
         ${unlockInteractively}
       '';
     };
@@ -154,6 +160,7 @@ in
         # do we need this? boot/systemd.nix:566 & boot/systemd/initrd.nix:357
         "bcachefs" = "${pkgs.bcachefs-tools}/bin/bcachefs";
         "mount.bcachefs" = "${pkgs.bcachefs-tools}/bin/mount.bcachefs";
+        "unshare" = "${pkgs.util-linux}/bin/unshare";
       };
       boot.initrd.extraUtilsCommands = lib.mkIf (!config.boot.initrd.systemd.enable) ''
         copy_bin_and_libs ${pkgs.bcachefs-tools}/bin/bcachefs


JohnRTitor:
This ^ and I believe on non-encrypted systems, this should not be trying to unlock it in the first place.

```
❯  journalctl -b | grep stage-1-init
May 29 00:25:18 Ainz-NIX stage-1-init: [Tue May 28 18:55:08 UTC 2024] Starting systemd-udevd version 255.6
May 29 00:25:18 Ainz-NIX stage-1-init: [Tue May 28 18:55:08 UTC 2024] starting device mapper and LVM...
May 29 00:25:18 Ainz-NIX stage-1-init: [Tue May 28 18:55:09 UTC 2024] waiting for device to appear UUID=7594aea3-de1f-466c-8420-3c7e4997fb34
May 29 00:25:18 Ainz-NIX stage-1-init: [Tue May 28 18:55:09 UTC 2024] Cannot unlock device 7594aea3-de1f-466c-8420-3c7e4997fb34 with path /dev/nvme0n1p7
May 29 00:25:18 Ainz-NIX stage-1-init: [Tue May 28 18:55:09 UTC 2024] mounting UUID=7594aea3-de1f-466c-8420-3c7e4997fb34 on /...
```

lheckemann:
@JohnRTitor indeed, it won't prompt for unencrypted filesystems due to the `ExecCondition` on the unit. This PR is meant to address filesystems that are encrypted, but mounted to multiple mount points. These currently result in a prompt per mount point, when one for the whole filesystem would be enough.

Also, your logs suggest that you're not using systemd stage-1, and I've only touched the systemd stage-1 implementation here.

JohnRTitor:
Mount bcachefs root in systemd-stage-1 does not even work on my system. Is it working for you?

I am using a pretty [simple setup](
https://github.com/JohnRTitor/nix-conf/blob/hyprland/default-host/hardware-configuration.nix#L20-L23) with `defaults` and `noatime` mount options.

JohnRTitor:
Identified the problem, [changing the fstab](https://github.com/JohnRTitor/nix-conf/commit/70df9aa2825839c9c6af0411af859152d9fa6979) to use `/dev/disk/by-uuid/UUID` format successfully lets systemd-initrd mount a bcachefs partition.

```patch
diff --git a/default-host/hardware-configuration.nix b/default-host/hardware-configuration.nix
index 0305bb69..ce093690 100644
--- a/default-host/hardware-configuration.nix
+++ b/default-host/hardware-configuration.nix
@@ -18,7 +18,7 @@
   boot.extraModulePackages = [];
 
   fileSystems."/" = {
-    device = "UUID=7594aea3-de1f-466c-8420-3c7e4997fb34";
+    device = "/dev/disk/by-uuid/7594aea3-de1f-466c-8420-3c7e4997fb34";
     fsType = "bcachefs";
   };
```

This might not be related to this particular PR, but I hope affected bcachefs users will find this fix helpful.

lheckemann:
That's a systemd bug which will be fixed in v256: https://github.com/systemd/systemd/issues/28989

ElvishJerricco:
@lheckemann Kind of... I just made an issue about why that's not really going to fix things: https://github.com/NixOS/nixpkgs/issues/317901

ElvishJerricco (changes requested):
> but I'm not aware of any way of checking if the correct key for a filesystem is already loaded.

Do you think we could check the kernel keyring somehow? I'm *very* inexperienced with the keyring. If not, this is a case where I honestly think we should fix it upstream and add the necessary flag / command to bcacehfs.

JohnRTitor:
Related PR which got merged upstream: https://github.com/koverstreet/bcachefs-tools/pull/263

ElvishJerricco:
Oh, well there we go. I was about to say that you *can* actually check with `keyctl`, but it seemed frustrating to parse out and all that so I'd rather just fix it upstream. But we don't have to! Nice.

@JohnRTitor Any chance of getting a subcommand or `unlock` flag to do this check? It'd be great to have that in the `ExecCondition`. As is we'll still needlessly re-invoke clevis, which might be troublesome.

lheckemann:
> > but I'm not aware of any way of checking if the correct key for a filesystem is already loaded.
> 
> Do you think we could check the kernel keyring somehow? I'm _very_ inexperienced with the keyring. If not, this is a case where I honestly think we should fix it upstream and add the necessary flag / command to bcacehfs.

We could, using keyctl, but I'm not sure the key description (~name), which is something like `bcachefs:bde20fb2-d752-4610-8147-83fb54a81146`, is a "public API". This is probably worth discussing with the bcachefs people as well; it would also be nice if `bcachefs unlock` could [optionally] be a no-op if the key is already loaded.

JohnRTitor:
> @JohnRTitor Any chance of getting a subcommand or `unlock` flag to do this check? 

Good idea, perhaps Tom (upstream contributor) can chip in @tmuehlbacher

tmuehlbacher:
Yes, I plan to add something like `unlock --dry-run` or as a separate command when I convert the unlock subcommand to Rust.
(It's not like I am a huge contributor though, just btw. But hoping I'll get to it soon.)

PR #316730, opened 2024-06-02
asfa: init at 0.10.0

obreitwi:
## Description of changes

* add new package: [asfa](https://crates.io/crates/asfa) at latest version

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [x] `sandbox = true`
- [ ] Tested, as applicable:
- [x] Tested basic functionality of all binary files (usually in `./result/bin/`), also [working my personal setup](f06e5d3592d8e7d061067acf61613832ef2750b9)

---

### DIFF
--- a/pkgs/by-name/as/asfa/package.nix
+++ b/pkgs/by-name/as/asfa/package.nix
@@ -1,0 +1,72 @@
+{
+  lib,
+  rustPlatform,
+  fetchFromGitHub,
+  stdenv,
+  darwin,
+  help2man,
+  openssl,
+  pkg-config,
+  installShellFiles,
+}:
+let
+  version = "0.10.0";
+  src = fetchFromGitHub {
+    owner = "obreitwi";
+    repo = "asfa";
+    rev = "v${version}";
+    hash = "sha256-MnhnwtZmPFhOuiqNiaxJnPu88JOdlpvyVy0YGphblBc=";
+  };
+in
+rustPlatform.buildRustPackage {
+  pname = "asfa";
+  inherit version src;
+
+  cargoHash = "sha256-/bRBP/NzcNOXl/nANeOYouUAo3NNbtbV9fxIJrNajYQ=";
+
+  outputs = [
+    "out"
+    "man"
+  ];
+
+  buildInputs = [
+    openssl
+  ] ++ lib.optionals stdenv.isDarwin [ darwin.apple_sdk.frameworks.SystemConfiguration ];
+
+  nativeBuildInputs = [
+    help2man
+    installShellFiles
+    pkg-config
+  ];
+
+  # checks disabled because tests need to be run against (docker-based) ephemeral ssh-server
+  doCheck = false;
+
+  postInstall = ''
+    mkdir -p "man/man1"
+    help2man -o "man/man1/asfa.1" "$out/bin/asfa"
+
+    # Generate info about all subcommands except for 'help' (which leads to error)
+    "$out/bin/asfa" --help | awk 'enabled && $1 != "help" { print $1 } /^SUBCOMMANDS:$/ { enabled=1 }' \
+        | while read -r cmd; do
+        help2man \
+            "--version-string=${version}" \
+            -o "man/man1/asfa-$cmd.1" \
+            "$out/bin/asfa $cmd"
+    done
+
+    installManPage man/man1/*.1
+  '';
+
+  meta = {
+    description = "Avoid sending file attachments by uploading them via SSH to a remote site and sharing a publicly-accessible URL with non-guessable (hash-based) prefix instead";
+    homepage = "https://github.com/obreitwi/asfa";
+    changelog = "https://github.com/obreitwi/asfa/blob/${src.rev}/CHANGELOG.md";
+    license = with lib.licenses; [
+      asl20
+      mit
+    ];
+    maintainers = with lib.maintainers; [ obreitwi ];
+    mainProgram = "asfa";
+  };
+}


AndersonTorres:
We do not use Semantic Versioning.

Reword the commit according to the Contributing Guidelines - the one you checked.

TomaSajt:
Hey! Thanks for taking your time to contribute.

Could you update the PR title/commit message to use the commit message conventions listed inside [pkgs/README.md](https://github.com/NixOS/nixpkgs/blob/master/pkgs/README.md#commit-conventions)?
It would be something like: `<package>: init at <version>`

pkgs/by-name/as/asfa/package.nix
```diff
@@ -0,0 +1,59 @@
...
+    buildInputs = [
+      pkgs.libssh2
+    ];
+
+    nativeBuildInputs = with pkgs; [
+      gawk
+      help2man
+    ];
+
+    OPENSSL_DIR = lib.attrsets.getOutput "dev" pkgs.openssl;
```
TomaSajt, line 32: By adding `pkg-config` to `nativeBuildInputs` and adding `openssl` to `buildInputs` the rust build can automatically discover the `openssl` package, making this part not needed.

(But if you really needed this, I'd recommend using `lib.getDev` and `lib.getLib` instead of `lib.attrsets.getOutput`)

pkgs/by-name/as/asfa/package.nix
```diff
@@ -0,0 +1,59 @@
+{
+  lib,
+  rustPlatform,
+  fetchFromGitHub,
+  pkgs,
+  ...
```
TomaSajt, line 6: Packages inside nixpkgs are usually imported using `pkgs.callPackage package.nix {}` which populates this top attrset with the values of `pkgs`, e.g. `rustPlatform` or `fetchFromGitHub`. This means you can also use this for the dependencies.

```suggestion
  gawk,
  help2man,
  openssl,
  libssh2,
```

pkgs/by-name/as/asfa/package.nix
```diff
@@ -0,0 +1,59 @@
...
+      owner = "obreitwi";
+      repo = "asfa";
+      rev = "v${version}";
+      hash = "sha256-MnhnwtZmPFhOuiqNiaxJnPu88JOdlpvyVy0YGphblBc=";
+    };
+
+    cargoHash = "sha256-/bRBP/NzcNOXl/nANeOYouUAo3NNbtbV9fxIJrNajYQ=";
+
+    buildInputs = [
+      pkgs.libssh2
```
TomaSajt, line 24: I saw that you recently added this dependency to the upstream repo flake as well. The package seems to build without this.
Could you provide the reason behind why this is needed?

pkgs/by-name/as/asfa/package.nix
obreitwi: Thanks, that is indeed much more straight forward :+1: 

For some reason it did not work like that in my first attempt at packaging in the original repo using [crane](https://github.com/ipetkov/crane); which, in hindsight, might have been overkill.

pkgs/by-name/as/asfa/package.nix
obreitwi: Really appreciate the detailed feedback, learning a lot! Adapted.

pkgs/by-name/as/asfa/package.nix
obreitwi: Same as above, it was needed to get [crane](https://github.com/ipetkov/crane)-integration working. But I will have to recheck.

After manual verification, it's definitely not needed for this PR… :+1:

pkgs/by-name/as/asfa/package.nix
```diff
@@ -0,0 +1,61 @@
+{
+  lib,
+  rustPlatform,
+  fetchFromGitHub,
+  gawk,
+  help2man,
+  libssh2,
+  openssl,
+  pkg-config,
+  ...
```
TomaSajt, line 10:
```suggestion
```

`pkgs.callPackage` is pretty smart and doesn't try to populate the attrset with everything if `...` is not used, so `...` can be removed`

pkgs/by-name/as/asfa/package.nix
```diff
@@ -0,0 +1,61 @@
...
+      mkdir -p "$out/man/man1"
+      help2man -o "$out/man/man1/asfa.1" "$out/bin/asfa"
+
+      # Generate info about all subcommands except for 'help' (which leads to error)
+      "$out/bin/asfa" --help | awk 'enabled && $1 != "help" { print $1 } /^SUBCOMMANDS:$/ { enabled=1 }' \
+          | while read -r cmd; do
+          help2man \
+              "--version-string=${version}" \
+              -o "$out/man/man1/asfa-''${cmd}.1" \
+              "$out/bin/asfa ''${cmd}"
```
TomaSajt, line 49: You could use `$var` instead of `''${var}`
This might be better, since less escaping is needed

```suggestion
              -o "$out/man/man1/asfa-$cmd.1" \
              "$out/bin/asfa $cmd"
```

TomaSajt (changes requested):
I've left some additional change suggestions.

Also, consider formatting the file using `nixfmt-rfc-style`.
It's the soon-to-be default across `nixpkgs`.
Doing this helps avoid unnecessarily big diffs when the tree-wide formatting happens sometime in the future.

pkgs/by-name/as/asfa/package.nix
```diff
@@ -0,0 +1,61 @@
+{
+  lib,
+  rustPlatform,
+  fetchFromGitHub,
+  gawk,
+  help2man,
+  libssh2,
```
TomaSajt, line 7:
```suggestion
```
This seems to no longer be used.

(btw I'm using `nil` as the language server, which marks unused variables, which is pretty useful)

pkgs/by-name/as/asfa/package.nix
obreitwi: Good point, adapted.

pkgs/by-name/as/asfa/package.nix
obreitwi: Good to know, adapted :+1:

pkgs/by-name/as/asfa/package.nix
```diff
@@ -0,0 +1,65 @@
...
+        | while read -r cmd; do
+        help2man \
+            "--version-string=${version}" \
+            -o "$out/man/man1/asfa-$cmd.1" \
+            "$out/bin/asfa $cmd"
+    done
+  '';
+
+  meta = with lib; {
+    description = "Avoid sending file attachments by uploading them via SSH to a remote site and sharing a publicly-accessible URL with non-guessable (hash-based) prefix instead.";
```
TomaSajt, line 55: IIRC `meta.description` shouldn't end with a period

```suggestion
    description = "Avoid sending file attachments by uploading them via SSH to a remote site and sharing a publicly-accessible URL with non-guessable (hash-based) prefix instead";
```

pkgs/by-name/as/asfa/package.nix
```diff
@@ -0,0 +1,65 @@
...
+  };
+
+  cargoHash = "sha256-/bRBP/NzcNOXl/nANeOYouUAo3NNbtbV9fxIJrNajYQ=";
+
+  buildInputs = [
+    openssl
+  ] ++ lib.optionals stdenv.isDarwin [ darwin.apple_sdk.frameworks.SystemConfiguration ];
+
+  nativeBuildInputs = [
+    gawk
```
TomaSajt, line 33:
```suggestion
```

According to [this](https://github.com/NixOS/nixpkgs/blob/master/doc/stdenv/stdenv.chapter.md#tools-provided-by-stdenv-sec-tools-of-stdenv) `awk` is already a part of the default `stdenv`.

(Don't forget to remove it from the top attrset too)

obreitwi: Thanks, adapted (& bookmarked).

pkgs/by-name/as/asfa/package.nix
obreitwi: Statistics definitely support this → adapted
```
❯ grep -r 'description.*[^\.]";' pkgs | wc -l
48133
❯ grep -r 'description.*\.";' pkgs | wc -l
931
```

TomaSajt (approved):
LGTM

pkgs/by-name/as/asfa/package.nix
```diff
@@ -0,0 +1,63 @@
...
+  buildInputs = [
+    openssl
+  ] ++ lib.optionals stdenv.isDarwin [ darwin.apple_sdk.frameworks.SystemConfiguration ];
+
+  nativeBuildInputs = [
+    help2man
+    pkg-config
+  ];
+
+  doCheck = false;
```
AndersonTorres, line 36:
```suggestion
  # <explain why doCheck is false>
  doCheck = false;
```

pkgs/by-name/as/asfa/package.nix
```diff
@@ -0,0 +1,63 @@
...
+
+  postInstall = ''
+    mkdir -p "$out/man/man1"
+    help2man -o "$out/man/man1/asfa.1" "$out/bin/asfa"
```
AndersonTorres, line 40: Since you are creating manpages, do these things:

1.  Modify outputs:

   ```nix
      outputs = [ "out" "man" ];
   ```

2. create the manpages at `postBuild` instead of `postInstall` - yes, do not create them directly at the final output dir;
3. install them at `postInstall` via `installManpage *.1` (installManpage comes from `installShellFiles` package).

pkgs/by-name/as/asfa/package.nix
```diff
@@ -0,0 +1,63 @@
...
+            "--version-string=${version}" \
+            -o "$out/man/man1/asfa-$cmd.1" \
+            "$out/bin/asfa $cmd"
+    done
+  '';
+
+  meta = with lib; {
+    description = "Avoid sending file attachments by uploading them via SSH to a remote site and sharing a publicly-accessible URL with non-guessable (hash-based) prefix instead";
+    homepage = "https://github.com/obreitwi/asfa";
+    changelog = "https://github.com/obreitwi/asfa/blob/v${version}/CHANGELOG.md";
```
AndersonTorres, line 55:
```suggestion
    changelog = "https://github.com/obreitwi/asfa/blob/${src.rev}/CHANGELOG.md";
```

It is more stable, since it works with hash-style commits.

AndersonTorres (changes requested):

pkgs/by-name/as/asfa/package.nix
obreitwi: Thanks, adjusted.

obreitwi: > create the manpages at postBuild instead of postInstall - yes, do not create them directly at the final output dir;

Turns out help2man hard-codes the name supplied binary so you can't replace `$out/bin/asfa` with `cargo run --release --` during the build phase. 

Found no easy & stable way of determining where the build products are stored (the cargo install hook seems to jump through [quite some hoops](https://github.com/NixOS/nixpkgs/blob/7fc3a682e8ae80681af91b84eea9a54583f21543/pkgs/build-support/rust/hooks/cargo-install-hook.sh#L17-L43) which I did not want to couple to).

Hence, the more straight-forward solution is to still create the the manpages in the `postInstall`-hook but then install them via `installManPage`. :+1:

pkgs/by-name/as/asfa/package.nix
```diff
@@ -0,0 +1,72 @@
...
+        help2man \
+            "--version-string=${version}" \
+            -o "man/man1/asfa-$cmd.1" \
+            "$out/bin/asfa $cmd"
+    done
+
+    installManPage man/man1/*.1
+  '';
+
+  meta = with lib; {
```
AndersonTorres, line 61:
```suggestion
  meta = {
```

Memo: https://github.com/NixOS/nixpkgs/issues/208242

pkgs/by-name/as/asfa/package.nix
```diff
@@ -0,0 +1,72 @@
...
+    description = "Avoid sending file attachments by uploading them via SSH to a remote site and sharing a publicly-accessible URL with non-guessable (hash-based) prefix instead";
+    homepage = "https://github.com/obreitwi/asfa";
+    changelog = "https://github.com/obreitwi/asfa/blob/${src.rev}/CHANGELOG.md";
+    license = with licenses; [
```
AndersonTorres, line 65:
```suggestion
    license = with lib.licenses; [
```

pkgs/by-name/as/asfa/package.nix
```diff
@@ -0,0 +1,72 @@
...
+      asl20
+      mit
+    ];
+    maintainers = with maintainers; [ obreitwi ];
```
AndersonTorres, line 69:
```suggestion
    maintainers = with lib.maintainers; [ obreitwi ];
```

AndersonTorres (changes requested):
Besides, LGTM

pkgs/by-name/as/asfa/package.nix
AndersonTorres: NixOS provides a test framework that fires QEMU virtual machines. But we can do this another day.

AndersonTorres (approved):

pkgs/by-name/as/asfa/package.nix
obreitwi: Yeah, been meaning to look into that. In other projects, I would like to run integration tests against ephemeral databases in pure nix derivations. Currently, those are set up by running docker images.

PR #316794, opened 2024-06-03
yazi: fix plugins and flavors config

viserctul:
Tried to install yazi with plugins, failed, discovered a typo of ```concatMapStringsSep``` instead of ```concatStringsSep``` in the package that makes build fail when plugins or flavors options are not default empty sets

### DIFF
--- a/pkgs/by-name/ya/yazi/package.nix
+++ b/pkgs/by-name/ya/yazi/package.nix
@@ -58,14 +58,14 @@ let
 
     mkdir $out/plugins
     ${lib.optionalString (plugins != { }) ''
-        ${lib.concatMapStringsSep
+        ${lib.concatStringsSep
         "\n"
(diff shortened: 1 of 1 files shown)


eclairevoyant:
Thanks for the contribution. In the future, please ensure to keep the PR template and follow CONTRIBUTING.md for the commit guidelines.

Also, can you explain the issue and why you believe this fixes it? Perhaps an example would be helpful here.

viserctul:
The issue is that when evaluated ```concatMapStringsSep``` was called with too few arguments and the build would crash because of failing to convert a function to a string. Most likely this was simply a typo, but because with the default arguments that part of code was skipped, the issue went unnoticed.

linsui (approved):

eclairevoyant:
Thanks. Please update the commit message to match the PR title, or at least stick to that format.

Aleksanaa (approved):

PR #316812, opened 2024-06-03
mattermost: fix nix-update-script version-regex and webapp auto-update

numinit:
## Description of changes

nixpkgs-update script wasn't working because we need ^ and $ anchors due to versions looking something like `v1.2.3-rc1` at times.

We also need to update the webapp. As it turns out, the nix-update-script looks at the `offlineCache` fixed-output derivation as well as the vendorHash and src hash, so use that.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [x] Tested, as applicable:

---

### DIFF
--- a/pkgs/servers/mattermost/default.nix
+++ b/pkgs/servers/mattermost/default.nix
@@ -37,6 +37,10 @@ buildGoModule rec {
     hash = "sha256-ZlvO/7kdMopIHBDdFp6wSQCR+NobGdDC6PcVd1iG16E=";
   };
 
+  # Makes nix-update-script pick up the fetchurl for the webapp.
+  # https://github.com/Mic92/nix-update/blob/1.3.1/nix_update/eval.py#L179
+  offlineCache = webapp;
+
   vendorHash = "sha256-TJCtgNf56A1U0EbV5gXjTro+YudVBRWiSZoBC3nJxnE=";
 
   modRoot = "./server";
@@ -69,7 +73,7 @@ buildGoModule rec {
 
   passthru = {
     updateScript = nix-update-script {
-      extraArgs = [ "--version-regex" "v(9\.5\.[0-9]+)" ];
+      extraArgs = [ "--version-regex" "^v(9\.5\.[0-9]+)$" ];
     };
     tests.mattermost = nixosTests.mattermost;
   };


stuebinm:
I have tested this locally by invoking `nix-update` with the new regex by hand. It works fine for updating from 9.5.5 to 9.5.6, but still fails to update the hash for the frontend webapp.

So I'm not sure how useful it'll be in practice, since if we ever forget to then manually update the webapp and ship out-of-sync versions of frontend and server, that'd probably not be good (I have never tried running different versions of frontend & server together, but afaik it is not supported by upstream in any way).

numinit:
Hmm. That is a good point. I've converted to a draft in the meantime. I wonder if there's some configuration we can make to the nix-update-script that fixes this.

numinit:
@stuebinm Figured it out, should be good now.

pkgs/servers/mattermost/default.nix
```diff
@@ -37,6 +37,10 @@ buildGoModule rec {
     hash = "sha256-ZlvO/7kdMopIHBDdFp6wSQCR+NobGdDC6PcVd1iG16E=";
   };
 
+  # Makes nix-update-script pick up the fetchurl for the webapp.
+  # https://github.com/Mic92/nix-update/blob/master/nix_update/eval.py#L190
```
stuebinm, line 41:
```suggestion
  # https://github.com/Mic92/nix-update/blob/dacee6fb5792ff7b671e8dee855772606c91dfc6/nix_update/eval.py#L190
```

nit: might be good to make this a permalink, since it refers to a specific line (and i feel linking to a potentially outdated version of `nix-update`, but to the correct place inside, is probably better than linking to the current version of `nix-update` in some random unhelpful place in case the code there has moved)

stuebinm (approved):
good news first: I have tested this by simply running `nix-update mattermost` on the nixpkgs checkout, and your change works as expected & it does update everything to the latest version now.

bad news: this currently only works without the version regex. When doing `nix-update mattermost -u` to use the script defined via `passthru`, `nix-update` fails to find any suitable version. Afaict, this is simply because the `https://github.com/mattermost/mattermost/releases.atom` file is truncated too aggressively (so that the `v9.5.6` release has already been pushed out again by newer point releases for the `9.6` and newer release series), and is not actually a problem with our update script. But it does mean we can't test it as thoroughly as I'd like.

… actually, if upstream regularly pushes out enough point releases to crowd out that atom feed, it might also explain at least some of the cases where the `r-ryantm` bot failed to pick up on new updates?

numinit:
To your point, I tested this by making the version regex point to something less specific than 9.5.

pkgs/servers/mattermost/default.nix
numinit: Fixed by linking to a tag.

stuebinm (approved):

numinit:
Can we get this merged prior to needing to roll this by hand?

RossComputerGuy (approved):

numinit:
@stuebinm FYI

https://github.com/Mic92/nix-update/issues/265

I'll see if I can roll this manually tonight, but looks like your worst case scenario for this happened.

numinit:
@stuebinm 

Alright, finally resolved! We have a successful auto update. 🔥🔥🔥

#335034

PR #316852, opened 2024-06-03
sdr-j-fm: init at 3.16-unstable-2023-12-07

doronbehar:
###### Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
- [ ] Tested, as applicable:

### DIFF
--- a/pkgs/applications/radio/sdr-j-fm/default.nix
+++ b/pkgs/applications/radio/sdr-j-fm/default.nix
@@ -1,0 +1,77 @@
+{
+  lib,
+  stdenv,
+  fetchFromGitHub,
+  cmake,
+  wrapQtAppsHook,
+  pkg-config,
+  qtbase,
+  qwt,
+  fftwFloat,
+  libsamplerate,
+  portaudio,
+  libusb1,
+  libsndfile,
+  featuresOverride ? { },
+}:
+
+stdenv.mkDerivation (finalAttrs: {
+  pname = "sdr-j-fm";
+  # The stable release doen't include the commit the came after 3.16 which
+  # added support for cmake options instead of using cmake set() commands. See
+  # also: https://github.com/JvanKatwijk/sdr-j-fm/pull/25
+  version = "3.16-unstable-2023-12-07";
+
+  src = fetchFromGitHub {
+    owner = "JvanKatwijk";
+    repo = "sdr-j-fm";
+    rev = "8e3a67f8fbf72dd6968cbeb2e3d7d513fd107c71";
+    hash = "sha256-l9WqfhDp2V01lhleYZqRpmyL1Ww+tJj10bjkMMlvyA0=";
+  };
+
+  nativeBuildInputs = [
+    cmake
+    wrapQtAppsHook
+    pkg-config
+  ];
+
+  buildInputs = [
+    qtbase
+    qwt
+    fftwFloat
+    libsamplerate
+    portaudio
+    libusb1
+    libsndfile
+  ];
+  cmakeFlags = lib.mapAttrsToList lib.cmakeBool finalAttrs.passthru.features;
+
+  passthru = {
+    features = {
+      # All of these features don't require an external depencies, althought it
+      # may be implied - upstraem bundles everything they need in their repo.
+      AIRSPY =     true;
+      SDRPLAY =    true;
+      SDRPLAY_V3 = true;
+      HACKRF =     true;
+      PLUTO =      true;
+      # Some more cmake flags are mentioned in upstream's CMakeLists.txt file
+      # but they don't actually make a difference.
+    } // featuresOverride;
+  };
+
+  postInstall = ''
+    # Weird default of upstream
+    mv $out/linux-bin $out/bin
+  '';
+
+  meta = with lib; {
+    description = "SDR based FM radio receiver software";
+    homepage = "https://github.com/JvanKatwijk/sdr-j-fm";
+    license = licenses.gpl2Only;
+    maintainers = with maintainers; [ doronbehar ];
+    # Upstream doesn't find libusb1 on Darwin. Upstream probably doesn't
+    # support it officially.
+    platforms = platforms.linux;
+  };
+})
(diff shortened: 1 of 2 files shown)


pkgs/applications/radio/sdr-j-fm/default.nix
```diff
@@ -0,0 +1,65 @@
...
+    librtlsdr
+    sdrplay
+  ];
+  postInstall = ''
+    # Weird default of upstream
+    mv $out/linux-bin $out/bin
+  '';
+
+  meta = with lib; {
+    description = "Sdr-j fm software";
```
markuskowa, line 60:
```suggestion
    description = "SDR based FM radio receiver software";
```
(is a bit more descriptive and does not contain the name of the package)

markuskowa (changes requested):
Please move the package to pkgs/by-name/sd/sdr-j-fm.
(see https://github.com/NixOS/rfcs/blob/master/rfcs/0140-simple-package-paths.md)
Please also consider to format the code with `nixfmt-rfc-style`.

pkgs/applications/radio/sdr-j-fm/default.nix
doronbehar: > (is a bit more descriptive and does not contain the name of the package)

Thanks!

doronbehar:
> Please move the package to pkgs/by-name/sd/sdr-j-fm. (see [NixOS/rfcs@`master`/rfcs/0140-simple-package-paths.md](https://github.com/NixOS/rfcs/blob/master/rfcs/0140-simple-package-paths.md?rgh-link-date=2024-06-03T11%3A45%3A22Z)) Please also consider to format the code with `nixfmt-rfc-style`.

Done!

doronbehar:
Thanks for the review.

doronbehar:
Oh I didn't use `pkgs/by-name` because I wanted to use `libsForQt5.callPackage`...

doronbehar:
I think that in this case not using `libsForQt5` is preferred, as it makes it possible to cross compile it. What do you think?

markuskowa:
> Oh I didn't use `pkgs/by-name` because I wanted to use `libsForQt5.callPackage`...

You can put it under `pkgs/by-name`, but then call it from `all-packages.nix` with `libsForQt5.callPackage`, or override parameters (at least that was my understanding how it should be done).

markuskowa:
> I think that in this case not using `libsForQt5` is preferred, as it makes it possible to cross compile it. What do you think?

I am not super familiar with the cross compiling intricacies. I would tend to recommend `libsForQt5` since this is the standard way to package a Qt-GUI package (?).

doronbehar:
> You can put it under `pkgs/by-name`, but then call it from `all-packages.nix` with `libsForQt5.callPackage`

Hmm OK haven't thought about that. Done.

markuskowa:
> > You can put it under `pkgs/by-name`, but then call it from `all-packages.nix` with `libsForQt5.callPackage`
> 
> On a 2nd thought, are you sure that's the correct thing to do? It fails currently:
> 
> https://github.com/NixOS/nixpkgs/actions/runs/9349951140/job/25732383548

Yes, you are right. This seems to make it an exception. 
This is the 1st time I run this issue since by-name is actively implemented.

doronbehar:
Apparently also using `pkgs/by-name` and a `default.nix` is also disallowed... Switched back to regular old `pkgs/applications/radio`.

pkgs/applications/radio/sdr-j-fm/default.nix
```diff
@@ -0,0 +1,66 @@
...
+  pkg-config,
+  qtbase,
+  qwt,
+  fftwFloat,
+  libsamplerate,
+  portaudio,
+  libusb,
+  libsndfile,
+  airspy,
+  librtlsdr,
```
markuskowa, line 16:
```suggestion
  rtl-sdr-librtlsdr,
```
I guess this got renamed at some point.

markuskowa: Ah, yes we have 3 different implementations for the library by now.

markuskowa (approved):

pkgs/applications/radio/sdr-j-fm/default.nix
```diff
@@ -0,0 +1,66 @@
...
+  qtbase,
+  qwt,
+  fftwFloat,
+  libsamplerate,
+  portaudio,
+  libusb1,
+  libsndfile,
+  airspy,
+  rtl-sdr,
+  sdrplay,
```
markuskowa, line 17: This package is `unfree`. One could consider to make it optional.

doronbehar: > This package is `unfree`. One could consider to make it optional. (I have no strong opinion on this)

Thanks for noticing this :) Apparently, the 3 dependencies `airspy`, `rtl-sdr` and `sdrplay` weren't really needed (I don't remember why I added them in the first place). So they are removed now.

doronbehar:
OK Apparently I misunderstood completely the `cmakeFlags` that upstream supports and how are they related to the devices' buildInputs I added initially. Now I read upstream's cmake files and it should be good.

markuskowa:
It fails to build on Darwin. Restrict the platform to Linux?

doronbehar:
> It fails to build on Darwin. Restrict the platform to Linux?

Done.

markuskowa (approved):

markuskowa:
> Apparently also using `pkgs/by-name` and a `default.nix` is also disallowed... Switched back to regular old `pkgs/applications/radio`.

Yes, under this tree the main file should be called `package.nix`.

doronbehar:
Thanks @markuskowa for the review. Here are a few more SDR related packages I recently added PRs for:

- https://github.com/NixOS/nixpkgs/pull/316920
- https://github.com/NixOS/nixpkgs/pull/316841

If they are of interest to you your review is welcome.

PR #316862, opened 2024-06-03
doc: improve lib.options reference documentation

hsjobeki:
## Description of changes

- Migrate documentation of lib.options to rfc145
- Init documentation for many undocumented option functions

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/lib/options.nix
+++ b/lib/options.nix
@@ -1,4 +1,6 @@
-/* Nixpkgs/NixOS option handling. */
+/**
+  Nixpkgs/NixOS option handling.
+*/
 { lib }:
 
 let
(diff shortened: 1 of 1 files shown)


lib/options.nix
```diff
@@ -45,158 +47,262 @@ let
...
+    isOption 1             // => false
+    isOption (mkOption {}) // => true
+    ```
+
+    :::
+
+    # Type
+
+    ```
+    isOption :: a -> bool
```
roberth, line 72: `bool` is also a module system type; and Nix types don't have a standardized syntax yet.
I'd suggest to differentiate them, because a _`types.bool` definition_ is not necessarily a valid boolean.

```suggestion
    isOption :: a -> Bool
```

This is also more consistent with the Haskell-inspired type syntax you're using.

lib/options.nix
```diff
@@ -45,158 +47,262 @@ let
...
 
-  /* Creates an Option attribute set. mkOption accepts an attribute set with the following keys:
+  /**
+    Creates an Option attribute set. mkOption accepts an attribute set with the following keys:
+
+    All keys default to `null` when not given.
+
+    # Inputs
+
+    Structured function argument
```
roberth, line 85: This looks like it's trying to say something complicated, so my mind went to function-valued argument instead.

```suggestion
    Structured attribute set
```

lib/options.nix
```diff
@@ -45,158 +47,262 @@ let
...
+    : Attribute set containing the following attributes.
```
roberth, line 86: This is incorrect. It does not need to contain any attributes.

```suggestion
    : Attribute set containing none or some of the following attributes.
```

lib/options.nix
```diff
@@ -45,158 +47,262 @@ let
...
     attrs // { _type = "option"; };
 
-  /* Creates an Option attribute set for a boolean value option i.e an
-     option to be toggled on or off:
+  /**
+    Creates an Option attribute set for a boolean value option i.e an
+    option to be toggled on or off.
+
+    default
+    : `false`
```
roberth, line 153: This looks structured, but isn't.
These docs are supposed to be in a general Nix-level syntax, so we shouldn't be inventing module system-level doc syntax.

lib/options.nix
roberth: There's more occurrences of this below.

lib/options.nix
```diff
@@ -220,16 +326,28 @@ rec {
...
+    Alias of mkPackageOption.
+
+    Previously used to create options with markdown
+    documentation, which is no longer required.
   */
   mkPackageOptionMD = mkPackageOption;
 
-  /* This option accepts anything, but it does not produce any result.
+  /**
+    This option accepts anything, but it does not produce any result.
```
roberth, line 338:
```suggestion
    This option accepts arbitrary definitions, but it does not produce an option value.
```

lib/options.nix
```diff
@@ -220,16 +326,28 @@ rec {
...
 
-     This is useful for sharing a module across different module sets
-     without having to implement similar features as long as the
-     values of the options are not accessed. */
+
+    # Inputs
+
+    `attrs`
+
+    : 1\. Function argument
```
roberth, line 349: What is `1\.` for?
Why are we repeating `attrs`, which is just an implementation detail, and not actually referenced in the documentation?

```suggestion
    : Attribute set whose attributes override the argument to `mkOption`.
```

lib/options.nix
```diff
@@ -243,6 +361,35 @@ rec {
     apply = x: throw "Option value is not readable because the option is not declared.";
   } // attrs);
 
+  /**
+    Merges multiple definitions of an option into a single value
+    by checking that they are all equal.
```
roberth, line 366:
```suggestion
    A merge function that merges multiple definitions of an option into a single value
    by checking that they are all equal.
    Merge functions are typically used for constructing option types.
```

lib/options.nix
```diff
@@ -243,6 +361,35 @@ rec {
...
+
+    `loc`
+    : location of the option in the configuration as a list of strings.
+
+      e.g. ["boot" "loader "grub" "enable"]
+
+    `defs`
+    : list of definition values and locations.
+
+      e.g. [ { file = "/foo.nix"; value = 1; } { file = "/bar.nix"; value = 2 } ]
```
roberth, line 378: This should simply refer to `mkOptionType`. That's where all the merge functions are passed to, so it's the best place to describe the type of those functions.
Implementations don't have to repeat that, because they aren't actually called by users.

lib/options.nix
```diff
@@ -254,18 +401,49 @@ rec {
     else if all isInt list && all (x: x == head list) list then head list
```
roberth, line 401: Note to self: resume review here.

roberth (commented):
Reviewed the first part. Ok overall, but noticed some quirks.

lib/options.nix
hsjobeki: Sorry i dont get it.

This is definition list syntax for commonmark. Which gets rendered nicely in the documentation.

Wouldn't a user want to know the default of `mkEnableOption` ? What is your proposed alternative here?

lib/options.nix
hsjobeki: > What is 1\. for?

Escaped argument list. This is autogenerated if there is no documentation.
Since the initial goal was to migrate to commonmark my migration tool creates such lists, which need to be escaped in order to not conflict with definition list syntax.

It is essentially a placeholder until somebody documents it.

lib/options.nix
hsjobeki: Yep had the same though. Gona fix the example.

hsjobeki: Just took a look into the nixpkgs manual. It seems we don't explain the concept of option types. At all....

lib/options.nix
```diff
@@ -45,158 +47,262 @@ let
...
-       isOption (mkOption {}) // => true
   */
   isOption = lib.isType "option";
 
-  /* Creates an Option attribute set. mkOption accepts an attribute set with the following keys:
+  /**
+    Creates an Option attribute set. mkOption accepts an attribute set with the following keys:
+
+    All keys default to `null` when not given.
+
```
infinisil, line 82: That's more of an implementation detail, I'd leave it out
```suggestion
```

lib/options.nix
```diff
@@ -45,158 +47,262 @@ let
...
+    # Inputs
+
+    Structured attribute set
+    : Attribute set containing none or some of the following attributes.
+
+      `default`
+      : Optional default value used when no definition is given in the configuration.
+
```
infinisil, line 90:
```suggestion
      `default`
      : Optional default value used when no definition is given in the configuration.
      
        _Default:_ None, when the option is evaluated an error is thrown.


```

lib/options.nix
```diff
@@ -45,158 +47,262 @@ let
...
+
+      `defaultText`
+      : Optional textual representation of the default, for the manual.
```
infinisil, line 93: Can mention that it supports `lib.literalExpression`

lib/options.nix
```diff
@@ -45,158 +47,262 @@ let
...
 
-     All keys default to `null` when not given.
+      `example`
+      : Optional example value used in the manual.
```
infinisil, line 96: Also supports `literalExpression`

lib/options.nix
```diff
@@ -45,158 +47,262 @@ let
...
 
-     Example:
-       mkOption { }  // => { _type = "option"; }
-       mkOption { default = "foo"; } // => { _type = "option"; default = "foo"; }
+      `description`
+      : Optional string describing the option.
```
infinisil, line 99:
```suggestion
      : Optional string describing the option. This is required if option documentation is generated.
```

infinisil (commented):
Just a small review, but note that this is a bit of a duplicate with https://nixos.org/manual/nixos/stable/#sec-option-declarations

We should really move those docs to the Nixpkgs manual, they don't belong into the NixOS one.

hsjobeki:
>We should really move those docs to the Nixpkgs manual, they don't belong into the NixOS one.

Yes. We should do a follow up PR, that moves them from the nixos manual into the nixpkgs manual and revise them with what we have here.

lib/options.nix
```diff
@@ -45,158 +47,263 @@ let
...
     attrs // { _type = "option"; };
 
-  /* Creates an Option attribute set for a boolean value option i.e an
-     option to be toggled on or off:
+  /**
+    Creates an Option attribute set for a boolean value option i.e an
+    option to be toggled on or off.
+
+    default
+    : `false`
```
hsjobeki, line 154:
```suggestion
   Create an option declaration with a default value of ´false´, and can be defined to ´true´.
```

infinisil (approved):

roberth:
Review header messages get lost in the noise sometimes (or just the PR `/files` page)

> Just a small review, but note that this is a bit of a duplicate with https://nixos.org/manual/nixos/stable/#sec-option-declarations
> 
> We should really move those docs to the Nixpkgs manual, they don't belong into the NixOS one.

My thoughts exactly. I hope this could be done as a next step.

PR #316901, opened 2024-06-03
autotools-language-server: init at 0.0.19

doronbehar:
###### Motivation for this change

Numerous other Python updates, and this new package.

###### Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
- [ ] Tested, as applicable:


pkgs/by-name/au/autotools-language-server/package.nix
```diff
@@ -0,0 +1,49 @@
...
+  propagatedBuildInputs = with python3.pkgs; [
+    tree-sitter-languages
+    lsp-tree-sitter
+  ];
+  nativeCheckInputs = [
+    python3.pkgs.pytestCheckHook
+  ];
+  #preCheck = ''
+  #  cd $out
+  #'';
```
Moraxyc, line 40: shoud it be removed?

pkgs/by-name/au/autotools-language-server/package.nix
```diff
@@ -0,0 +1,49 @@
...
+
+python3.pkgs.buildPythonApplication rec {
+  pname = "autotools-language-server";
+  version = "0.0.19";
+  pyproject = true;
+
+  src = fetchFromGitHub {
+    owner = "Freed-Wu";
+    repo = "autotools-language-server";
+    rev = version;
```
Moraxyc, line 15:
```suggestion
    rev = "refs/tags/${version}";
```
Help to avoid confusion.

pkgs/by-name/au/autotools-language-server/package.nix
doronbehar: > shoud it be removed?

Yes!

pkgs/by-name/au/autotools-language-server/package.nix
```diff
@@ -0,0 +1,49 @@
...
+      revert = true;
+    })
+  ];
+
+  nativeBuildInputs = [
+    python3.pkgs.setuptools-generate
+    python3.pkgs.setuptools-scm
+  ];
+
+  propagatedBuildInputs = with python3.pkgs; [
```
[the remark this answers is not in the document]
doronbehar: > It's better to use `dependencies` to conform `pyproject.toml`.

Thanks for the suggestion, I'm not used yet to these new attributes. Also switched `nativeBuildInputs` to `build-system`.

pkgs/by-name/au/autotools-language-server/package.nix
```diff
@@ -0,0 +1,49 @@
...
+  version = "0.0.19";
+  pyproject = true;
+
+  src = fetchFromGitHub {
+    owner = "Freed-Wu";
+    repo = "autotools-language-server";
+    rev = version;
+    hash = "sha256-V0EOV1ZmeC+4svc2fqV6AIiL37dkrxUJAnjywMZcENw=";
+  };
+  patches = [
```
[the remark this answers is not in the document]
doronbehar: > Each patch in nixpkgs applied to the upstream source should be documented, please explain the purpose of the patch.

Thanks, I forgot :). I commented on that commit in upstream's repo requesting the reason for that commit which I revert, and they replied a few days ago, but GitHub didn't notify me.. Now there is more info there, and I forced pushed a fix related to one of the dependencies related to the info there.

doronbehar:
Thanks for the review!

pkgs/by-name/au/autotools-language-server/package.nix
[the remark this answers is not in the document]
doronbehar: > It's better to use `dependencies` to conform `pyproject.toml`.

Did that for all other packages as well..

pkgs/development/python-modules/tree-sitter-languages/default.nix
```diff
@@ -0,0 +1,68 @@
...
+    wheel
+    cython
+  ];
+  dependencies = [
+    # https://github.com/grantjenks/py-tree-sitter-languages/issues/67
+    tree-sitter0_21
+  ];
+  # Generate languages.so file (build won't fail without this, but tests will).
+  preBuild = ''
+    ${python.pythonOnBuildForHost}/bin/python build.py
```
SuperSandro2000, line 46:
```suggestion
    ${python.pythonOnBuildForHost.interpreter} build.py
```

pkgs/development/python-modules/tree-sitter-languages/default.nix
```diff
@@ -0,0 +1,68 @@
...
+    hash = "sha256-wKU2c8QRBKFVFqg+DAeH5+cwm5jpDLmPZG3YBUsh/lM=";
+    # Use git, to also fetch tree-sitter repositories that upstream puts their
+    # hases in the repository as well, in repos.txt.
+    forceFetchGit = true;
+    postFetch = ''
+      cd $out
+      substitute build.py get-repos.py \
+        --replace-fail "from tree_sitter import Language" "" \
+        --replace-fail 'print(f"{sys.argv[0]}: Building", languages_filename)' "exit(0)"
+      ${python.pythonOnBuildForHost}/bin/python get-repos.py
```
SuperSandro2000, line 30:
```suggestion
      ${python.pythonOnBuildForHost.interpreter} get-repos.py
```

pkgs/development/python-modules/tree-sitter-languages/default.nix
doronbehar: Nice!

pkgs/development/python-modules/tree-sitter-languages/default.nix
doronbehar: Thanks!

pkgs/top-level/python-packages.nix
```diff
@@ -15402,6 +15404,20 @@ self: super: with self; {
 
   tree-sitter = callPackage ../development/python-modules/tree-sitter { };
 
+  tree-sitter0_21 = callPackage ../development/python-modules/tree-sitter0_21 { };
```
Janik-Haag, line 15407: I know this is a bit late, but why are you copying and pasting the file instead of using overrideAttrs and just modifying the original package, that would make this a lot more maintainable and would ensure that the two versions nix expressions don't diverge from each other?

doronbehar: > I know this is a bit late, but why are you copying and pasting the file instead of using overrideAttrs and just modifying the original package, that would make this a lot more maintainable and would ensure that the two versions nix expressions don't diverge from each other?

The argument to make the expressions different is exactly related to the potential divergence: I anticipate that in the future the two expressions will keep on diverging and diverging, especially since the developers of `tree-sitter` don't officially support this outdated `0.21` version.

Janik-Haag: I guess you should at least make yourself the maintainer rather then fabaff. I also don't know how to feel about keeping the older version alive, I guess it's fine as long as their is still upstream support, but once upstream drops support we kinda wann drop it as well I think.

doronbehar: > I guess you should at least make yourself the maintainer rather then fabaff. I also don't know how to feel about keeping the older version alive, I guess it's fine as long as their is still upstream support, but once upstream drops support we kinda wann drop it as well I think.

You are correct, this is not ideal, but this is related to the choice of `python3.pkgs.tree-sitter-languages` v.s `python3.pkgs.tree-sitter-make`. The later is not packaged in Nixpkgs yet, and [for a good reason](https://github.com/Freed-Wu/autotools-language-server/commit/f149843becfcfd6b2bb4a98eb1f3984c01d5fd33#r142659163) and the former has [this issue](https://github.com/grantjenks/py-tree-sitter-languages/issues/67) with the latest version of `python3.pkgs.tree-sitter`.

TBH only now I managed to explain it good enough :) The situation perhaps should have been explained in some comments.

PR #316920, opened 2024-06-03
LPCNet: init at 0.5

doronbehar:
###### Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
- [ ] Tested, as applicable:

### DIFF
--- a/pkgs/by-name/lp/LPCNet/package.nix
+++ b/pkgs/by-name/lp/LPCNet/package.nix
@@ -1,0 +1,56 @@
+{
+  lib,
+  stdenv,
+  fetchFromGitHub,
+  fetchurl,
+  cmake,
+  codec2,
+  # for tests
+  octave,
+}:
+
+stdenv.mkDerivation (finalAttrs: {
+  pname = "LPCNet";
+  version = "0.5";
+
+  src = fetchFromGitHub {
+    owner = "drowe67";
+    repo = "LPCNet";
+    rev = "v${finalAttrs.version}";
+    hash = "sha256-tHZLKXmuM86A6OpfS3CRRjhFbqj1Q/w1w56msdgLHb0=";
+  };
+  passthru = {
+    # Prebuilt neural network model that is needed during the build - can be overrwritten
+    nnmodel = fetchurl {
+      url = "http://rowetel.com/downloads/deep/lpcnet_191005_v1.0.tgz";
+      hash = "sha256-UJRAkkdR/dh/+qVoPuPd3ZN69cgzuRBMzOZdUWFJJsg=";
+    };
+  };
+  preConfigure = ''
+    mkdir build
+    cp \
+      ${finalAttrs.finalPackage.passthru.nnmodel} \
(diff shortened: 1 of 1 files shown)


pkgs/by-name/lp/LPCNet/package.nix
```diff
@@ -0,0 +1,51 @@
...
+
+stdenv.mkDerivation (finalAttrs: {
+  pname = "LPCNet";
+  version = "0.5";
+
+  src = fetchFromGitHub {
+    owner = "drowe67";
+    repo = "LPCNet";
+    rev = "v${finalAttrs.version}";
+    sha256 = "sha256-9VxPD5GAIymr9TOAZmL69y+bcGAkdIJ6//EIeHDX9AQ=";
```
markuskowa, line 20:
```suggestion
    hash = "sha256-9VxPD5GAIymr9TOAZmL69y+bcGAkdIJ6//EIeHDX9AQ=";
```

pkgs/by-name/lp/LPCNet/package.nix
```diff
@@ -0,0 +1,51 @@
...
+    # Also download prebuilt neural network model that is needed during the build.
+    postFetch = ''
+      mkdir $out/build
+      curl \
```
markuskowa, line 24: Wouldn't it be cleaner to use a separate `fetchurl` for the training data tarball (and link into the build dir)? That would decouple the code and the dataset.

markuskowa (changes requested):

pkgs/by-name/lp/LPCNet/package.nix
doronbehar: > Wouldn't it be cleaner to use a separate `fetchurl` for the training data tarball (and link into the build dir)? That would decouple the code and the dataset.

I was thinking about it, but I find it cleaner the fact that we have only 1 hash to maintain, and not 2... However choosing one or the other won't be significant in terms of version bumps to the package, as this tarball hasn't changed for a few years.

doronbehar:
Thanks for the review, pushed all changes besides the `fetchurl` dilemma, haven't decided what do with it yet.

pkgs/by-name/lp/LPCNet/package.nix
markuskowa: curl should probably never be used directly in a package. The one hash solution means that the dataset also gets stored again when the package is updated. This would also make it easier to override the data set if necessary. 
I would recommend using `fetchurl`.

doronbehar: > curl should probably never be used directly in a package. The one hash solution means that the dataset also gets stored again when the package is updated. This would also make it easier to override the data set if necessary. I would recommend using `fetchurl`.

Indeed being able to override it is a good reason to use `fetchurl` - in cases like the `rowetel.com` server is down.. Thanks :)

markuskowa (approved):
LGTM.
The Darwin failure is related to octave.

PR #231165, opened 2023-05-10
nixos/sshd: Remove algorithms that do MAC-then-encrypt

mweinelt:
Algorithms with the -etm suffix calculate the MAC after encryption, which is generally considered safer.

###### Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
- [ ] For non-Linux: Is `sandbox = true` set in `nix.conf`? (See [Nix manual](https://nixos.org/manual/nix/stable/command-ref/conf-file.html))
- [ ] Tested, as applicable:
  - [NixOS test(s)](https://nixos.org/manual/nixos/unstable/index.html#sec-nixos-tests) (look inside [nixos/tests](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests))
  - and/or [package tests](https://nixos.org/manual/nixpkgs/unstable/#sec-package-tests)
  - or, for functions and "core" functionality, tests in [lib/tests](https://github.com/NixOS/nixpkgs/blob/master/lib/tests) or [pkgs/test](https://github.com/NixOS/nixpkgs/blob/master/pkgs/test)
  - made sure NixOS tests are [linked](https://nixos.org/manual/nixpkgs/unstable/#ssec-nixos-tests-linking) to the relevant packages
- [ ] Tested compilation of all packages that depend on this change using `nix-shell -p nixpkgs-review --run "nixpkgs-review rev HEAD"`. Note: all changes have to be committed, also see [nixpkgs-review usage](https://github.com/Mic92/nixpkgs-review#usage)
- [ ] Tested basic functionality of all binary files (usually in `./result/bin/`)
- [23.05 Release Notes (or backporting 22.11 Release notes)](https://github.com/NixOS/nixpkgs/blob/master/CONTRIBUTING.md#generating-2305-release-notes)
  - [ ] (Package updates) Added a release notes entry if the change is major or breaking
  - [ ] (Module updates) Added a release notes entry if the change is significant
  - [ ] (Module addition) Added a release notes entry if adding a new NixOS module
- [x] Fits [CONTRIBUTING.md](https://github.com/NixOS/nixpkgs/blob/master/CONTRIBUTING.md).

PR #172393, opened 2022-05-10
nixos/openssh: Add sntrup761x25519-sha512 kexAlgo

mweinelt:
###### Description of changes

Introduced in OpenSSH 9.0 it became the part of the default kexAlgorithm
selection, visibile in sshd_config(5).

It is also enabled by default in the OpenSSH client, as can be seen from

$ ssh -Q KexAlgorithms

Also clarifies that we use the referenced documents as the lower bound,
given that they haven't been updated for 5-7y.

###### Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
- [ ] For non-Linux: Is `sandbox = true` set in `nix.conf`? (See [Nix manual](https://nixos.org/manual/nix/stable/command-ref/conf-file.html))
- [x] Tested, as applicable:
  - [NixOS test(s)](https://nixos.org/manual/nixos/unstable/index.html#sec-nixos-tests) (look inside [nixos/tests](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests))
  - and/or [package tests](https://nixos.org/manual/nixpkgs/unstable/#sec-package-tests)
  - or, for functions and "core" functionality, tests in [lib/tests](https://github.com/NixOS/nixpkgs/blob/master/lib/tests) or [pkgs/test](https://github.com/NixOS/nixpkgs/blob/master/pkgs/test)
  - made sure NixOS tests are [linked](https://nixos.org/manual/nixpkgs/unstable/#ssec-nixos-tests-linking) to the relevant packages
- [ ] Tested compilation of all packages that depend on this change using `nix-shell -p nixpkgs-review --run "nixpkgs-review rev HEAD"`. Note: all changes have to be committed, also see [nixpkgs-review usage](https://github.com/Mic92/nixpkgs-review#usage)
- [ ] Tested basic functionality of all binary files (usually in `./result/bin/`)
- [22.05 Release Notes (or backporting 21.11 Release notes)](https://github.com/NixOS/nixpkgs/blob/master/CONTRIBUTING.md#generating-2205-release-notes)
  - [ ] (Package updates) Added a release notes entry if the change is major or breaking
  - [ ] (Module updates) Added a release notes entry if the change is significant
  - [ ] (Module addition) Added a release notes entry if adding a new NixOS module
  - [ ] (Release notes changes) Ran `nixos/doc/manual/md-to-db.sh` to update generated release notes
[...]

PR #316934, opened 2024-06-03
nixos/ssh: use upstream's default cryptographic algorithms

tomfitzhenry:
## Motivation

Hardening SSH algorithms, which typically means dropping all-but-the-strongest is of questionable value, given SSH's downgrade protection[0]. We pay in compatibility, and maintenance.

Further, as noted in
https://github.com/NixOS/nixpkgs/pull/172393/files#r871727289 , both the guidelines that we follow have not been updated in years.

The costs of having/maintaining these defaults:

1. The burden of having a larger module that deviates from upstream. We've slowly been reducing the upstream diff, to reduce maintenance burden.
2. Difficult for users to opt-out of these defaults. For example, when using a "no OpenSSL" build of OpenSSH, having these defaults means having to manually overriding NixOS's defaults. Upstream's defaults, meanwhile, gracefully only use available algorithms, if OpenSSL is not linked.
    * For users seeking to reduce attack surfaces that are fortunate enough to only have modern clients, they could choose to use `pkgs.opensshPackages.openssh.override { linkOpenssl = false; }`, which only supports chacha20-poly1305 and curve25519-sha256. This has the added benefit of removing openssl from the memory of sshd.
3. https://github.com/NixOS/nixpkgs/pull/231165 unexpectedly broke many clients.
4. The time in discussing/reviewing these defaults.
5. Anecdotally, a friend trying NixOS for the first time with a ssh_config supporting only ecdh-* key exchanges was unable to SSH in after install.

There's a certain level of enjoyment that comes from researching and selecting a favourite suite of ciphers, but as a distro, it's not our core competancy, and best left for upstream who are active in advances/attacks/compatibility.

0. https://eprint.iacr.org/2016/072.pdf

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/nixos/doc/manual/release-notes/rl-2411.section.md
+++ b/nixos/doc/manual/release-notes/rl-2411.section.md
@@ -19,6 +19,9 @@
 - `openssh` and `openssh_hpn` are now compiled without Kerberos 5 / GSSAPI support in an effort to reduce the attack surface of the components for the majority of users. Users needing this support can
   use the new `opensshWithKerberos` and `openssh_hpnWithKerberos` flavors (e.g. `programs.ssh.package = pkgs.openssh_gssapi`).
 
+- The `openssh` module now uses upstream's default cryptographic algorithms, in an effort to reduce
+  required maintenance.
+
 - `nvimpager` was updated to version 0.13.0, which changes the order of user and
   nvimpager settings: user commands in `-c` and `--cmd` now override the
   respective default settings because they are executed later.
--- a/nixos/modules/services/networking/ssh/sshd.nix
+++ b/nixos/modules/services/networking/ssh/sshd.nix
@@ -417,36 +417,13 @@ in
             };
             KexAlgorithms = mkOption {
               type = types.nullOr (types.listOf types.str);
-              default = [
-                "sntrup761x25519-sha512@openssh.com"
-                "curve25519-sha256"
-                "curve25519-sha256@libssh.org"
-                "diffie-hellman-group-exchange-sha256"
-              ];
-              description = ''
-                Allowed key exchange algorithms
-
-                Uses the lower bound recommended in both
-                <https://stribika.github.io/2015/01/04/secure-secure-shell.html>
-                and
-                <https://infosec.mozilla.org/guidelines/openssh#modern-openssh-67>
-              '';
+              default = null;
+              description = "Allowed key exchange algorithms. `null` uses upstream's default.";
             };
             Macs = mkOption {
               type = types.nullOr (types.listOf types.str);
-              default = [
-                "hmac-sha2-512-etm@openssh.com"
-                "hmac-sha2-256-etm@openssh.com"
-                "umac-128-etm@openssh.com"
-              ];
-              description = ''
-                Allowed MACs
-
-                Defaults to recommended settings from both
-                <https://stribika.github.io/2015/01/04/secure-secure-shell.html>
-                and
-                <https://infosec.mozilla.org/guidelines/openssh#modern-openssh-67>
-              '';
+              default = null;
+              description = "Allowed MACs. `null` uses upstream's default.";
             };
             StrictModes = mkOption {
               type = types.nullOr (types.bool);
@@ -457,22 +434,8 @@ in
             };
             Ciphers = mkOption {
               type = types.nullOr (types.listOf types.str);
-              default = [
-                "chacha20-poly1305@openssh.com"
-                "aes256-gcm@openssh.com"
-                "aes128-gcm@openssh.com"
-                "aes256-ctr"
-                "aes192-ctr"
-                "aes128-ctr"
-              ];
-              description = ''
-                Allowed ciphers
-
-                Defaults to recommended settings from both
-                <https://stribika.github.io/2015/01/04/secure-secure-shell.html>
-                and
-                <https://infosec.mozilla.org/guidelines/openssh#modern-openssh-67>
-              '';
+              default = null;
+              description = "Allowed ciphers. `null` uses upstream's default.";
             };
             AllowUsers = mkOption {
               type = with types; nullOr (listOf str);
--- a/nixos/modules/system/boot/initrd-ssh.nix
+++ b/nixos/modules/system/boot/initrd-ssh.nix
@@ -150,9 +150,13 @@ in
         HostKey ${initrdKeyPath path}
       '')}
 
-      KexAlgorithms ${concatStringsSep "," sshdCfg.settings.KexAlgorithms}
-      Ciphers ${concatStringsSep "," sshdCfg.settings.Ciphers}
-      MACs ${concatStringsSep "," sshdCfg.settings.Macs}
+      '' + lib.optionalString (sshdCfg.settings.KexAlgorithms != null) ''
+        KexAlgorithms ${concatStringsSep "," sshdCfg.settings.KexAlgorithms}
+      '' + lib.optionalString (sshdCfg.settings.Ciphers != null) ''
+        Ciphers ${concatStringsSep "," sshdCfg.settings.Ciphers}
+      '' + lib.optionalString (sshdCfg.settings.Macs != null) ''
+        MACs ${concatStringsSep "," sshdCfg.settings.Macs}
+      '' + ''
 
       LogLevel ${sshdCfg.settings.LogLevel}
 
--- a/nixos/tests/openssh.nix
+++ b/nixos/tests/openssh.nix
@@ -119,14 +119,6 @@ in {
           hostKeys = [
             { type = "ed25519"; path = "/etc/ssh/ssh_host_ed25519_key"; }
           ];
-          settings = {
-            # Must not specify the OpenSSL provided algorithms.
-            Ciphers = [ "chacha20-poly1305@openssh.com" ];
-            KexAlgorithms = [
-              "curve25519-sha256"
-              "curve25519-sha256@libssh.org"
-            ];
-          };
         };
         users.users.root.openssh.authorizedKeys.keys = [
           snakeOilEd25519PublicKey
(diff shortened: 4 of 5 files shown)


aszlig:
I agree that we shouldn't deviate too much from upstream, but we could alternatively move those defaults to something like [`profiles/hardened.nix`](https://github.com/NixOS/nixpkgs/blob/c5b6c986b3c071de25a192be7989e1fbb9e9fef9/nixos/modules/profiles/hardened.nix) or even a dedicated module for people to opt-in.

SuperSandro2000 (commented):
We should document in the options that null means upstreams defaults. Also we should add a release note entry. 

Did you do a comparison which algorithms are now enabled again? Is there maybe some upstream guidance we could use instead like those algorithms are soon to be disabled by default and if you can, you should already disable them?

nixos/doc/manual/release-notes/rl-2411.section.md
```diff
@@ -19,6 +19,8 @@
 - `openssh` and `openssh_hpn` are now compiled without Kerberos 5 / GSSAPI support in an effort to reduce the attack surface of the components for the majority of users. Users needing this support can
   use the new `opensshWithKerberos` and `openssh_hpnWithKerberos` flavors (e.g. `programs.ssh.package = pkgs.openssh_gssapi`).
 
+- The `openssh` module now uses upstream's default cryptographic algorithms, as part of a broad effort to reduce the upstream diff in response to the 2024 xz incident.
```
SuperSandro2000, line 22:
```suggestion
- The `openssh` module now uses upstream's default cryptographic algorithms in an effort to reduce required maintenance.
```
I think mentioning xz here will just trigger unnecessary questions and lead people into a wrong direction.

nixos/modules/services/networking/ssh/sshd.nix
```diff
@@ -417,36 +417,13 @@ in
...
-              description = ''
-                Allowed key exchange algorithms
-
-                Uses the lower bound recommended in both
-                <https://stribika.github.io/2015/01/04/secure-secure-shell.html>
-                and
-                <https://infosec.mozilla.org/guidelines/openssh#modern-openssh-67>
-              '';
+              default = null;
+              description = "Allowed key exchange algorithms. null will use upstream's defaults.";
```
SuperSandro2000, line 421:
```suggestion
              description = "Allowed key exchange algorithms. `null` uses upstream's default.";
```

nixos/modules/services/networking/ssh/sshd.nix
```diff
@@ -417,36 +417,13 @@ in
...
-              description = ''
-                Allowed MACs
-
-                Defaults to recommended settings from both
-                <https://stribika.github.io/2015/01/04/secure-secure-shell.html>
-                and
-                <https://infosec.mozilla.org/guidelines/openssh#modern-openssh-67>
-              '';
+              default = null;
+              description = "Allowed MACs. null will use upstream's defaults.";
```
SuperSandro2000, line 426:
```suggestion
              description = "Allowed MACs. `null` uses upstream's default.";
```

nixos/modules/services/networking/ssh/sshd.nix
```diff
@@ -457,22 +434,8 @@ in
...
-              description = ''
-                Allowed ciphers
-
-                Defaults to recommended settings from both
-                <https://stribika.github.io/2015/01/04/secure-secure-shell.html>
-                and
-                <https://infosec.mozilla.org/guidelines/openssh#modern-openssh-67>
-              '';
+              default = null;
+              description = "Allowed ciphers. null will use upstream's defaults.";
```
SuperSandro2000, line 438:
```suggestion
              description = "Allowed ciphers. `null` uses upstream's default.";
```

nixos/modules/services/networking/ssh/sshd.nix
```diff
@@ -417,36 +417,13 @@ in
             };
             KexAlgorithms = mkOption {
               type = types.nullOr (types.listOf types.str);
-              default = [
-                "sntrup761x25519-sha512@openssh.com"
-                "curve25519-sha256"
-                "curve25519-sha256@libssh.org"
-                "diffie-hellman-group-exchange-sha256"
```
SuperSandro2000, line 424: The new ones are

```
sntrup761x25519-sha512@openssh.com,
curve25519-sha256,curve25519-sha256@libssh.org,
ecdh-sha2-nistp256,ecdh-sha2-nistp384,ecdh-sha2-nistp521,
diffie-hellman-group-exchange-sha256,
diffie-hellman-group16-sha512,diffie-hellman-group18-sha512,
diffie-hellman-group14-sha256
```

https://github.com/openssh/openssh-portable/blob/V_9_7_P1/sshd_config.5#L1053C1-L1058C30

nixos/modules/services/networking/ssh/sshd.nix
```diff
@@ -417,36 +417,13 @@ in
...
-              '';
+              default = null;
+              description = "Allowed key exchange algorithms. null will use upstream's defaults.";
             };
             Macs = mkOption {
               type = types.nullOr (types.listOf types.str);
-              default = [
-                "hmac-sha2-512-etm@openssh.com"
-                "hmac-sha2-256-etm@openssh.com"
-                "umac-128-etm@openssh.com"
```
SuperSandro2000, line 440:
```
umac-64-etm@openssh.com,umac-128-etm@openssh.com,
hmac-sha2-256-etm@openssh.com,hmac-sha2-512-etm@openssh.com,
hmac-sha1-etm@openssh.com,
umac-64@openssh.com,umac-128@openssh.com,
hmac-sha2-256,hmac-sha2-512,hmac-sha1
```

https://github.com/openssh/openssh-portable/blob/V_9_7_P1/sshd_config.5#L1206-L1210

nixos/modules/services/networking/ssh/sshd.nix
```diff
@@ -457,22 +434,8 @@ in
             };
             Ciphers = mkOption {
               type = types.nullOr (types.listOf types.str);
-              default = [
-                "chacha20-poly1305@openssh.com"
-                "aes256-gcm@openssh.com"
-                "aes128-gcm@openssh.com"
-                "aes256-ctr"
-                "aes192-ctr"
-                "aes128-ctr"
```
SuperSandro2000, line 466: https://github.com/openssh/openssh-portable/blob/V_9_7_P1/sshd_config.5#L578-L580

```
chacha20-poly1305@openssh.com,
aes128-ctr,aes192-ctr,aes256-ctr,
aes128-gcm@openssh.com,aes256-gcm@openssh.com
```

SuperSandro2000 (commented):
I think the last point from the review thread is, to move the old defaults to the hardened profile.

SuperSandro2000 (commented):
To test this out I just picked this PR into my fork and wanted to deploy it but it failed because I use openssh in the initrd 

I think we need to add some conditions to that config file

https://github.com/NixOS/nixpkgs/blob/master/nixos/modules/system/boot/initrd-ssh.nix#L153-L155

SuperSandro2000:
I fixed evaluation with https://github.com/SuperSandro2000/nixpkgs/commit/02c08fcc8355c02b639cfc81a5be779aa466f8db but I didn't test it yet e2e

SuperSandro2000:
> * keeping the overridden defaults in `hardened.nix` does still offer improved ergonomics for those that want to opt-out of those overrides, but not for `hardened.nix` users (who have to do the un-ergonomic "Ciphers = lib.mkForce [ ... ]".

We could fix that by using lib.mkDefault

I don#t think we need an extra option for that

@aszlig opinions?

aszlig:
> Alternatives are to have an option `enableMozillaRecommendations` (or whoever we're trusting, and I'd like us to select _one_ guide, if any).

That would be similar to eg. `services.nginx.recommendedTlsSettings`, which might probably be even preferable to the `hardened.nix` module, since if we want this in the `hardened` module it could still be included there via `services.openssh.enableMozillaRecommendations` (if the maintainers (ping @joachifm and @emilazy) of `hardened.nix` feel inclined to do).

> We could fix that by using lib.mkDefault

With the `enableMozillaRecommendations` option it would also address that issue since overriding is a single `mkForce` away.

mweinelt:
Mozilla has fired their security team who was in charge of config recommendations back in 2020.

LeSuisse:
Hello,

The given arguments seem fair to me however I'm a bit worried of re-introducing weak cryptography primitives by default. This seems like a regression and I would not be surprised if we end up getting issues from people getting suddenly a worse reports from <INSERT_YOUR_FAVORITE_COMMERCIAL_AUTOMATIC_PENTEST_TOOL> (or things like `ssh-audit` as shown in this PR).

I'm mostly worried about 2 things:
* re-introducing things like `hmac-sha1` or `umac-64@openssh.com` to the list of MACs
* (to a lesser extend) adding additional kex algos based on DH

both are arguably not something that is future proof.

We might have a middle ground solution that makes things easier for users wanting to harden their systems while not making it worse for users relying on the default values: `sshd` allows you to add or remove items from its default values using the `+` or `-` characters at the beginning of the list.
For example, right now at the top of your PR I can have `services.openssh.settings.Macs = ["-hmac-sha1" "hmac-sha1-etm@openssh.com" "umac-64@openssh.com"];` and `hmac-sha1` / `hmac-sha1-etm@openssh.com` / `umac-64@openssh.com` will be removed from the available MACs (I agree it feels a bit like a hack :smile: and maybe we should expose an nicer way to control that?).

This approach would allow us to exclude the crypto primitives with the lowest security margins and pick up the new ones proposed by `openssh` (I'm assuming they will be sane...) when we upgrade and also not cause issues with the `linkOpenssl = false;` use case.

PR #278126, opened 2024-01-01
wivrn: init at 0.10.1

MatthewCroughan:
## Description of changes

There's a few problems such as needing to override monado, and `callPackage` tinygltf since draco doesn't provide tiny_gtlf.h, if anyone can recommend ways around that I'd love to hear it, but I'm putting my work in a PR so it doesn't go to waste and so that others can continue it.

Adding cuda is also another TODO.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [x] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
- For non-Linux: Is sandboxing enabled in `nix.conf`? (See [Nix manual](https://nixos.org/manual/nix/stable/command-ref/conf-file.html))
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:
  - [NixOS test(s)](https://nixos.org/manual/nixos/unstable/index.html#sec-nixos-tests) (look inside [nixos/tests](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests))
  - and/or [package tests](https://nixos.org/manual/nixpkgs/unstable/#sec-package-tests)
  - or, for functions and "core" functionality, tests in [lib/tests](https://github.com/NixOS/nixpkgs/blob/master/lib/tests) or [pkgs/test](https://github.com/NixOS/nixpkgs/blob/master/pkgs/test)
  - made sure NixOS tests are [linked](https://nixos.org/manual/nixpkgs/unstable/#ssec-nixos-tests-linking) to the relevant packages
- [ ] Tested compilation of all packages that depend on this change using `nix-shell -p nixpkgs-review --run "nixpkgs-review rev HEAD"`. Note: all changes have to be committed, also see [nixpkgs-review usage](https://github.com/Mic92/nixpkgs-review#usage)
- [x] Tested basic functionality of all binary files (usually in `./result/bin/`)
- [24.05 Release Notes](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2405.section.md) (or backporting [23.05](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2305.section.md) and [23.11](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2311.section.md) Release notes)
  - [ ] (Package updates) Added a release notes entry if the change is major or breaking
[...]

PR #316975, opened 2024-06-03
wivrn: init at 0.19

PassiveLemon:
## Description of changes

Adds WiVRn https://github.com/NixOS/nixpkgs/pull/278126#issuecomment-1975239007

Continuation of https://github.com/NixOS/nixpkgs/pull/299830 because I goofed.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/nixos/modules/services/video/wivrn.nix
+++ b/nixos/modules/services/video/wivrn.nix
@@ -1,0 +1,226 @@
+{
+  config,
+  pkgs,
+  lib,
+  ...
+}:
+let
+  inherit (lib)
+    mkIf
+    mkEnableOption
+    mkPackageOption
+    mkOption
+    optionalString
+    optionalAttrs
+    isDerivation
+    recursiveUpdate
+    getExe
+    literalExpression
+    types
+    maintainers
+    ;
+  cfg = config.services.wivrn;
+  configFormat = pkgs.formats.json { };
+
+  # For the application option to work with systemd PATH, we find the store binary path of
+  # the package, concat all of the following strings, and then update the application attribute.
+  # Application can either be a package or a list that has a package as the first element.
+  applicationExists = builtins.hasAttr "application" cfg.config.json;
+  applicationListNotEmpty = (
+    if builtins.isList cfg.config.json.application then
+      (builtins.length cfg.config.json.application) != 0
+    else
+      true
+  );
+  applicationCheck = applicationExists && applicationListNotEmpty;
+
+  applicationBinary = (
+    if builtins.isList cfg.config.json.application then
+      builtins.head cfg.config.json.application
+    else
+      cfg.config.json.application
+  );
+  applicationStrings = builtins.tail cfg.config.json.application;
+
+  applicationPath = mkIf applicationCheck applicationBinary;
+
+  applicationConcat = (
+    if builtins.isList cfg.config.json.application then
+      builtins.concatStringsSep " " ([ (getExe applicationBinary) ] ++ applicationStrings)
+    else
+      (getExe applicationBinary)
+  );
+  applicationUpdate = recursiveUpdate cfg.config.json (
+    optionalAttrs applicationCheck { application = applicationConcat; }
+  );
+  configFile = configFormat.generate "config.json" applicationUpdate;
+in
+{
+  options = {
+    services.wivrn = {
+      enable = mkEnableOption "WiVRn, an OpenXR streaming application";
+
+      package = mkPackageOption pkgs "wivrn" { };
+
+      openFirewall = mkEnableOption "the default ports in the firewall for the WiVRn server";
+
+      defaultRuntime = mkEnableOption ''
+        WiVRn Monado as the default OpenXR runtime on the system.
+        The config can be found at `/etc/xdg/openxr/1/active_runtime.json`.
+
+        Note that applications can bypass this option by setting an active
+        runtime in a writable XDG_CONFIG_DIRS location like `~/.config`
+      '';
+
+      autoStart = mkEnableOption "starting the service by default";
+
+      monadoEnvironment = mkOption {
+        type = types.attrs;
+        description = "Environment variables to be passed to the Monado environment.";
+        default = {
+          XRT_COMPOSITOR_LOG = "debug";
+          XRT_PRINT_OPTIONS = "on";
+          IPC_EXIT_ON_DISCONNECT = "off";
+        };
+      };
+
+      extraPackages = mkOption {
+        type = types.listOf types.package;
+        description = "Packages to add to the wivrn-application service $PATH.";
+        default = [ ];
+        example = literalExpression "[ pkgs.bash pkgs.procps ]";
+      };
+
+      config = {
+        enable = mkEnableOption "configuration for WiVRn";
+        json = mkOption {
+          type = configFormat.type;
+          description = ''
+            Configuration for WiVRn. The attributes are serialized to JSON in config.json.
+
+            Note that the application option must be either a package or a
+            list with package as the first element.
+
+            See https://github.com/Meumeu/WiVRn/blob/master/docs/configuration.md
+          '';
+          default = { };
+          example = literalExpression ''
+            {
+              scale = 0.8;
+              bitrate = 100000000;
+              encoders = [
+                {
+                  encoder = "nvenc";
+                  codec = "h264";
+                  width = 1.0;
+                  height = 1.0;
+                  offset_x = 0.0;
+                  offset_y = 0.0;
+                }
+              ];
+              application = [ pkgs.wlx-overlay-s ];
+              tcp_only = true;
+            }
+          '';
+        };
+      };
+    };
+  };
+
+  config = mkIf cfg.enable {
+    assertions = [
+      {
+        assertion = !applicationCheck || isDerivation applicationBinary;
+        message = "The application in WiVRn configuration is not a package. Please ensure that the application is a package or that a package is the first element in the list.";
+      }
+    ];
+
+    systemd.user = {
+      services = {
+        # The WiVRn server runs in a hardened service and starts the applications in a different service
+        wivrn = {
+          description = "WiVRn XR runtime service";
+          environment = {
+            # Default options
+            # https://gitlab.freedesktop.org/monado/monado/-/blob/598080453545c6bf313829e5780ffb7dde9b79dc/src/xrt/targets/service/monado.in.service#L12
+            XRT_COMPOSITOR_LOG = "debug";
+            XRT_PRINT_OPTIONS = "on";
+            IPC_EXIT_ON_DISCONNECT = "off";
+          } // cfg.monadoEnvironment;
+          serviceConfig = {
+            ExecStart = (
+              (getExe cfg.package) + " --systemd" + optionalString cfg.config.enable " -f ${configFile}"
+            );
+            # Hardening options
+            CapabilityBoundingSet = [ "CAP_SYS_NICE" ];
+            AmbientCapabilities = [ "CAP_SYS_NICE" ];
+            LockPersonality = true;
+            NoNewPrivileges = true;
+            PrivateTmp = true;
+            ProtectClock = true;
+            ProtectControlGroups = true;
+            ProtectKernelLogs = true;
+            ProtectKernelModules = true;
+            ProtectKernelTunables = true;
+            ProtectProc = "invisible";
+            ProtectSystem = "strict";
+            RemoveIPC = true;
+            RestrictNamespaces = true;
+            RestrictSUIDSGID = true;
+          };
+          wantedBy = mkIf cfg.autoStart [ "default.target" ];
+          restartTriggers = [
+            cfg.package
+            configFile
+          ];
+        };
+        wivrn-application = mkIf applicationCheck {
+          description = "WiVRn application service";
+          requires = [ "wivrn.service" ];
+          serviceConfig = {
+            ExecStart = (
+              (getExe cfg.package) + " --application" + optionalString cfg.config.enable " -f ${configFile}"
+            );
+            Restart = "on-failure";
+            RestartSec = 0;
+            PrivateTmp = true;
+          };
+          # We need to add the application to PATH so WiVRn can find it
+          path = [ applicationPath ] ++ cfg.extraPackages;
+        };
+      };
+    };
+
+    services = {
+      # WiVRn can be used with some wired headsets so we include xr-hardware
+      udev.packages = with pkgs; [
+        android-udev-rules
+        xr-hardware
+      ];
+      avahi = {
+        enable = true;
+        publish = {
+          enable = true;
+          userServices = true;
+        };
+      };
+    };
+
+    networking.firewall = mkIf cfg.openFirewall {
+      allowedTCPPorts = [ 9757 ];
+      allowedUDPPorts = [ 9757 ];
+    };
+
+    environment = {
+      systemPackages = [
+        cfg.package
+        applicationPath
+      ];
+      pathsToLink = [ "/share/openxr" ];
+      etc."xdg/openxr/1/active_runtime.json" = mkIf cfg.defaultRuntime {
+        source = "${cfg.package}/share/openxr/1/openxr_wivrn.json";
+      };
+    };
+  };
+  meta.maintainers = with maintainers; [ passivelemon ];
+}
--- a/pkgs/by-name/wi/wivrn/package.nix
+++ b/pkgs/by-name/wi/wivrn/package.nix
@@ -1,0 +1,149 @@
+{
+  config,
+  lib,
+  stdenv,
+  fetchFromGitHub,
+  fetchFromGitLab,
+  applyPatches,
+  autoAddDriverRunpath,
+  avahi,
+  boost,
+  cli11,
+  cmake,
+  cudaPackages ? { },
+  cudaSupport ? config.cudaSupport,
+  eigen,
+  ffmpeg,
+  freetype,
+  git,
+  glm,
+  glslang,
+  harfbuzz,
+  libdrm,
+  libGL,
+  libva,
+  libpulseaudio,
+  libX11,
+  libXrandr,
+  nix-update-script,
+  nlohmann_json,
+  onnxruntime,
+  openxr-loader,
+  pipewire,
+  pkg-config,
+  python3,
+  shaderc,
+  spdlog,
+  systemd,
+  udev,
+  vulkan-headers,
+  vulkan-loader,
+  vulkan-tools,
+  x264,
+}:
+stdenv.mkDerivation (finalAttrs: {
+  pname = "wivrn";
+  version = "0.19";
+
+  src = fetchFromGitHub {
+    owner = "wivrn";
+    repo = "wivrn";
+    rev = "v${finalAttrs.version}";
+    hash = "sha256-DYV+JUWjjhLZLq+4Hv7jxOyxDqQut/mU1X0ZFMoNkDI=";
+  };
+
+  monado = applyPatches {
+    src = fetchFromGitLab {
+      domain = "gitlab.freedesktop.org";
+      owner = "monado";
+      repo = "monado";
+      rev = "bcbe19ddd795f182df42051e5495e9727db36c1c";
+      hash = "sha256-sh5slHROcuC3Dgenu1+hm8U5lUOW48JUbiluYvc3NiQ=";
+    };
+
+    patches = [
+      "${finalAttrs.src}/patches/monado/0001-c-multi-disable-dropping-of-old-frames.patch"
+      "${finalAttrs.src}/patches/monado/0002-ipc-server-Always-listen-to-stdin.patch"
+      "${finalAttrs.src}/patches/monado/0003-c-multi-Don-t-log-frame-time-diff.patch"
+      "${finalAttrs.src}/patches/monado/0005-distortion-images.patch"
+      "${finalAttrs.src}/patches/monado/0008-Use-mipmaps-for-distortion-shader.patch"
+      "${finalAttrs.src}/patches/monado/0009-convert-to-YCbCr-in-monado.patch"
+    ];
+  };
+
+  strictDeps = true;
+
+  postUnpack = ''
+    # Let's make sure our monado source revision matches what is used by WiVRn upstream
+    ourMonadoRev="${finalAttrs.monado.src.rev}"
+    theirMonadoRev=$(grep "GIT_TAG" ${finalAttrs.src.name}/CMakeLists.txt | awk '{print $2}')
+    if [ ! "$theirMonadoRev" == "$ourMonadoRev" ]; then
+      echo "Our Monado source revision doesn't match CMakeLists.txt." >&2
+      echo "  theirs: $theirMonadoRev" >&2
+      echo "    ours: $ourMonadoRev" >&2
+      return 1
+    fi
+  '';
+
+  nativeBuildInputs = [
+    cmake
+    git
+    glslang
+    pkg-config
+    python3
+  ] ++ lib.optionals cudaSupport [ autoAddDriverRunpath ];
+
+  buildInputs = [
+    avahi
+    boost
+    cli11
+    eigen
+    ffmpeg
+    freetype
+    glm
+    harfbuzz
+    libdrm
+    libGL
+    libva
+    libX11
+    libXrandr
+    libpulseaudio
+    nlohmann_json
+    onnxruntime
+    openxr-loader
+    pipewire
+    shaderc
+    spdlog
+    systemd
+    udev
+    vulkan-headers
+    vulkan-loader
+    vulkan-tools
+    x264
+  ] ++ lib.optionals cudaSupport [ cudaPackages.cudatoolkit ];
+
+  cmakeFlags = [
+    (lib.cmakeBool "WIVRN_USE_VAAPI" true)
+    (lib.cmakeBool "WIVRN_USE_X264" true)
+    (lib.cmakeBool "WIVRN_USE_NVENC" cudaSupport)
+    (lib.cmakeBool "WIVRN_USE_SYSTEMD" true)
+    (lib.cmakeBool "WIVRN_USE_PIPEWIRE" true)
+    (lib.cmakeBool "WIVRN_USE_PULSEAUDIO" true)
+    (lib.cmakeBool "WIVRN_BUILD_CLIENT" false)
+    (lib.cmakeBool "WIVRN_OPENXR_INSTALL_ABSOLUTE_RUNTIME_PATH" true)
+    (lib.cmakeBool "FETCHCONTENT_FULLY_DISCONNECTED" true)
+    (lib.cmakeFeature "FETCHCONTENT_SOURCE_DIR_MONADO" "${finalAttrs.monado}")
+  ];
+
+  passthru.updateScript = nix-update-script { };
+
+  meta = with lib; {
+    description = "An OpenXR streaming application to a standalone headset";
+    homepage = "https://github.com/Meumeu/WiVRn/";
+    changelog = "https://github.com/Meumeu/WiVRn/releases/";
+    license = licenses.gpl3Only;
+    maintainers = with maintainers; [ passivelemon ];
+    platforms = platforms.linux;
+    mainProgram = "wivrn-server";
+  };
+})
--- a/nixos/doc/manual/release-notes/rl-2405.section.md
+++ b/nixos/doc/manual/release-notes/rl-2405.section.md
@@ -201,6 +201,8 @@ The pre-existing `services.ankisyncd` has been marked deprecated and will be dro
 
 - [watchdogd](https://troglobit.com/projects/watchdogd/), a system and process supervisor using watchdog timers. Available as [services.watchdogd](#opt-services.watchdogd.enable).
 
+- [WiVRn](https://github.com/Meumeu/WiVRn), an OpenXR streaming application. Available as [services.wivrn](#opt-services.wivrn.enable).
+
 - [Workout-tracker](https://github.com/jovandeginste/workout-tracker), a workout tracking web application for personal use.
 
 - [wyoming-satellite](https://github.com/rhasspy/wyoming-satellite), a voice assistant satellite for Home Assistant using the Wyoming protocol. Available as [services.wyoming.satellite](#opt-services.wyoming.satellite.enable).
--- a/nixos/modules/module-list.nix
+++ b/nixos/modules/module-list.nix
@@ -1373,6 +1373,7 @@
   ./services/video/mediamtx.nix
   ./services/video/unifi-video.nix
   ./services/video/v4l2-relayd.nix
+  ./services/video/wivrn.nix
   ./services/wayland/cage.nix
   ./services/wayland/hypridle.nix
   ./services/web-apps/akkoma.nix


PassiveLemon:
@Scrumplex @ckiee @Pandapip1 @technobaboo @xytovl

ckiee:
Welcome back.. It's a right of passage sometimes.

Pandapip1 (approved):

Scrumplex:
Both `nixpkgs-review` and ofborg fail to evaluate. While I can evaluate it when built by itself, FDO GitLab is currently under maintenance, so I can't pull the Monado source.

Pandapip1:
@grahamcofborg eval

PassiveLemon:
From what Ofborg says, it looks like the monado CMake feature is involved but I don't really know why it fails. It builds on my system and it built in the other PR before

Scrumplex:
I think it just can't build the monado derivation currently because of the downtime of FDO GitLab

Scrumplex:
I have refactored the derivation a little, utilizing `finalAttrs` and moving the sources into the derivation itself, so overriding is much easier.

```patch
From 49cf6b5a420e347c1edcfc18f7e9787091fc1434 Mon Sep 17 00:00:00 2001
From: Sefa Eyeoglu <contact@scrumplex.net>
Date: Tue, 4 Jun 2024 23:23:01 +0200
Subject: [PATCH] wivrn: refactor derivation

Signed-off-by: Sefa Eyeoglu <contact@scrumplex.net>
---
 pkgs/by-name/wi/wivrn/package.nix | 59 ++++++++++++++-----------------
 1 file changed, 26 insertions(+), 33 deletions(-)

diff --git a/pkgs/by-name/wi/wivrn/package.nix b/pkgs/by-name/wi/wivrn/package.nix
index 67ec086ab758..81c798ee8b6b 100644
--- a/pkgs/by-name/wi/wivrn/package.nix
+++ b/pkgs/by-name/wi/wivrn/package.nix
@@ -3,7 +3,7 @@
 , stdenv
 , fetchFromGitHub
 , fetchFromGitLab
-, fetchpatch
+, applyPatches
 , avahi
 , boost
 , cmake
@@ -38,55 +38,36 @@
 , vulkan-tools
 , x264
 }:
-let
-  wivrnVersion = "0.15";
+  pname = "wivrn";
+  version = "0.15";
 
-  wivrnSrc = fetchFromGitHub {
+  src = fetchFromGitHub {
     owner = "meumeu";
     repo = "wivrn";
-    rev = "v${wivrnVersion}";
     hash = "sha256-RVRbL9hqy9pMKjvzwaP+9HGEfdpAhmlnnvqZsEGxlCw=";
   };
 
-  monadoVersion = builtins.head (builtins.elemAt (builtins.split
-    "monado\n +GIT_TAG +([A-Za-z0-9]+)"
-    (builtins.readFile (wivrnSrc + "/CMakeLists.txt"))
-  ) 1);
-
-  monado = stdenv.mkDerivation {
-    pname = "monado";
-    version = monadoVersion;
-
+  monadoSrc = applyPatches {
     src = fetchFromGitLab {
       domain = "gitlab.freedesktop.org";
       owner = "monado";
       repo = "monado";
-      rev = monadoVersion;
+      rev = "ffb71af26f8349952f5f820c268ee4774613e200";
       hash = "sha256-+RTHS9ShicuzhiAVAXf38V6k4SVr+Bc2xUjpRWZoB0c=";
     };
 
     patches = [
-      (wivrnSrc + "/patches/monado/0001-c-multi-disable-dropping-of-old-frames.patch")
-      (wivrnSrc + "/patches/monado/0002-ipc-server-Always-listen-to-stdin.patch")
-      (wivrnSrc + "/patches/monado/0003-c-multi-Don-t-log-frame-time-diff.patch")
+      ("${finalAttrs.src}/patches/monado/0001-c-multi-disable-dropping-of-old-frames.patch")
+      ("${finalAttrs.src}/patches/monado/0002-ipc-server-Always-listen-to-stdin.patch")
+      ("${finalAttrs.src}/patches/monado/0003-c-multi-Don-t-log-frame-time-diff.patch")
     ];
 
     postPatch = ''
       substituteInPlace CMakeLists.txt --replace "add_subdirectory(doc)" ""
     '';
-
-    dontBuild = true;
-
-    installPhase = ''
-      cp -r . $out
-    '';
   };
-in
-stdenv.mkDerivation {
-  pname = "wivrn";
-  version = wivrnVersion;
-
-  src = wivrnSrc;
 
   nativeBuildInputs = [
     cmake
@@ -128,6 +109,18 @@ stdenv.mkDerivation {
     cudaPackages.cudatoolkit
   ];
 
+  postUnpack = ''
+    # Let's make sure our monado source revision matches what is used by WiVRn upstream
+    ourMonadoRev="${finalAttrs.monadoSrc.src.rev}"
+    theirMonadoRev=$(grep -A1 "https://gitlab.freedesktop.org/monado/monado" ${finalAttrs.src.name}/CMakeLists.txt | tail -n1 | sed 's/.*GIT_TAG\s*//')
+    if [ ! "$theirMonadoRev" == "$ourMonadoRev" ]; then
+      echo "Our Monado source revision doesn't match CMakeLists.txt." >&2
+      echo "  theirs: $theirMonadoRev" >&2
+      echo "    ours: $ourMonadoRev" >&2
+      return 1
+    fi
+  '';
+
   cmakeFlags = [
     (lib.cmakeBool "WIVRN_USE_VAAPI" true)
     (lib.cmakeBool "WIVRN_USE_X264" true)
@@ -138,7 +131,7 @@ stdenv.mkDerivation {
     (lib.cmakeBool "WIVRN_BUILD_CLIENT" false)
     (lib.cmakeBool "WIVRN_OPENXR_INSTALL_ABSOLUTE_RUNTIME_PATH" true)
     (lib.cmakeBool "FETCHCONTENT_FULLY_DISCONNECTED" true)
-    (lib.cmakeFeature "FETCHCONTENT_SOURCE_DIR_MONADO" "${monado}")
+    (lib.cmakeFeature "FETCHCONTENT_SOURCE_DIR_MONADO" "${finalAttrs.monadoSrc}")
   ];
 
   passthru.updateScript = nix-update-script { };
@@ -146,10 +139,10 @@ stdenv.mkDerivation {
   meta = with lib; {
     description = "An OpenXR streaming application to a standalone headset";
     homepage = "https://github.com/Meumeu/WiVRn/";
-    changelog = "https://github.com/Meumeu/WiVRn/releases/tag/v${wivrnVersion}";
+    changelog = "https://github.com/Meumeu/WiVRn/releases/tag/v${finalAttrs.wivrnVersion}";
     license = licenses.gpl3Only;
     maintainers = with maintainers; [ passivelemon ];
     platforms = platforms.linux;
     mainProgram = "wivrn-server";
   };
-}
+})
-- 
2.44.1
```

Pandapip1:
Actually, might it make sense to patch `wivrn` to use the latest version of `monado` instead?

xytovl:
> Actually, might it make sense to patch `wivrn` to use the latest version of `monado` instead?

I do not recommend it, updating monado often requires changes in WiVRn itself, see latest updates on dev branch:
https://github.com/Meumeu/WiVRn/commit/27235d3ea79497c284fcd0961d2e7ea822f1a70a
https://github.com/Meumeu/WiVRn/commit/eaf6efda72a6315f39e36f9083e5b5b3bf37ec7a
https://github.com/Meumeu/WiVRn/commit/41044fe0125df226c12e6dd0956eb712187f9562 (which didn't require any change).

You may backport those changes if you wish, but you may not always have a compatible version.

pkgs/by-name/wi/wivrn/package.nix
```diff
@@ -0,0 +1,155 @@
...
+    owner = "meumeu";
+    repo = "wivrn";
+    rev = "v${wivrnVersion}";
+    hash = "sha256-RVRbL9hqy9pMKjvzwaP+9HGEfdpAhmlnnvqZsEGxlCw=";
+  };
+
+  monadoVersion = builtins.head (builtins.elemAt (builtins.split
+    "monado\n +GIT_TAG +([A-Za-z0-9]+)"
+    (builtins.readFile (wivrnSrc + "/CMakeLists.txt"))
+  ) 1);
```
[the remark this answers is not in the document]
PassiveLemon: Yeah I like that approach much better

PassiveLemon:
I believe it currently is best to target the official releases rather than backporting changes. I like the idea case of checking if the versions match so if WiVRn is automatically updated but the Monado source isn't, a broken package isn't released.

PassiveLemon:
Also I think it would be of interest to add WiVRn configuration to the module as well. It's in json so there's probably some parts of other modules I can use to get that working. I'll probably look into it tomorrow.

Pandapip1:
Fair enough. Is it possible to have an update script update something other than the derivation's `src`?

PassiveLemon:
Config file doesn't appear to be that hard to implement, however it's loaded from the users config dir by default and I don't think there's a flag to specify a config file to load. Not sure how this could be done otherwise. Maybe a suggestion for upstream.

PassiveLemon:
Result of `nixpkgs-review` run on x86_64-linux [1](https://github.com/Mic92/nixpkgs-review)
<details>
  <summary>1 package blacklisted:</summary>
  <ul>
    <li>nixos-install-tools</li>
  </ul>
</details>
<details>
  <ul>
    <li>wivrn</li>
  </ul>
</details>

Pandapip1:
I guess you could abuse the logic and override `XDG_CONFIG_HOME` / `XDG_CACHE_HOME`

Scrumplex:
https://github.com/Meumeu/WiVRn/releases/tag/v0.16 this should be updated to 0.16

PassiveLemon:
Looks like 0.16 got configuration file support as well so we can define the config in this module

PassiveLemon:
New push:
Bumped to 0.16
Config option added to the module which must be enabled to be used. Otherwise, it will default to user configurations.

nixos/modules/services/video/wivrn.nix
```diff
@@ -0,0 +1,130 @@
...
+            Configuration for WiVRn. The attributes are serialized to JSON in config.json.
+            See https://github.com/Meumeu/WiVRn/blob/master/docs/configuration.md
+          '';
+          default = {
+            scale = 0.8;
+            bitrate = 50000000;
+            encoders = [
+              {
+                encoder = "nvenc";
+                codec = "h265";
```
Pandapip1, line 40: Since NVIDIA's stuff is proprietary and only works with an NVIDIA GPU, it can't be the default. Additionally, the h265 codec itself is proprietary, so it _also_ can't be the default.

Almost all CPUs post-2011 have hardware h264 acceleration, so h264 + VAAPI makes sense as an open-source engine / encoder combination that will work on _most_ systems.

Pandapip1 (changes requested):

xytovl:
Why set a configuration file at all?
Most of the values match the default, so there is no difference. If some defaults seem to be incorrect, please raise an issue upstream so that we can discuss.

For instance, `tcp_only` adds some additional locking and prevents video data to be prioritized on the network, therefore I don't recommend it when UDP can be used.

PassiveLemon:
I'm only setting it to the defaults that are upstream. I accidentally left `tcp_only` true so I'll fix that when I can.

nixos/modules/services/video/wivrn.nix
PassiveLemon: It's what's set to default upstream so that's what I put here, but we can always change it. By default the config option is not even enabled.

xytovl:
The default is not static for encoder:
> Default value: nvenc if Nvidia GPU and compiled with cuda, vaapi for all other GPU when compiled with ffmpeg, else x264.

PassiveLemon:
Ah I must have missed that, I will make changes when I have time

PassiveLemon:
Fixed the defaults

nixos/modules/services/video/wivrn.nix
```diff
@@ -0,0 +1,130 @@
...
+          description = ''
+            Configuration for WiVRn. The attributes are serialized to JSON in config.json.
+            See https://github.com/Meumeu/WiVRn/blob/master/docs/configuration.md
+          '';
+          default = {
+            scale = 0.8;
+            bitrate = 50000000;
+            encoders = [
+              {
+                encoder = "x264";
```
xytovl, line 39: This forces software encoding, most users will want to use hardware accelerated encoders isntead

nixos/modules/services/video/wivrn.nix
```diff
@@ -0,0 +1,130 @@
...
+                codec = "h265";
```
xytovl, line 40: x264 encoder can not encode into h265, WiVRn will ignore the codec value.

PassiveLemon: What should the default be then? I think using software encoding as a default is fine as it should work for everyone. We can't use Nvenc because its proprietary and Vaapi doesn't cover all GPUs (unless that does include Nvidia?)

ckiee:
@PassiveLemon i'm unsubscribing for now (84 days), let me know when you're ready to freeze new changes to go for merge (:

nixos/modules/services/video/wivrn.nix
xytovl: I may not know NixOS enough, if it is meant as a template for user to modify, you can leave x264 encoder with h264 codec.
If it is meant as a safe default, it is better to not have a configuration file at all and let the application decide at run-time.

PassiveLemon: With it needing to be enabled, I believe it functions more as a template.

Pandapip1: +0 to no config file by default. Alternatively, you could add a systemd unit that checks the hardware and generates a config file at runtime.

PassiveLemon: That feels a bit out of the scope of this module. No config sounds fine, I will push a change if that's what we agree on

LuNeder:
Hey! I apologize if this is not the correct place for this, but did you get to try opencomposite apps - more specifically on nvidia? 

I was testing this out and found that opencomposite programs make wivrn crash on connection. 

After some troubleshooting with the lvra discord folks, someone there mentioned that monado-vulkan-layers is a dependency for that on nvidia and that that might be the cause (tbh I'm yet to test it tho). Could that be added here or should this be a separate package and PR instead? (if it's the latter I could try doing it I think, don't wanna bother you \^-^)

PassiveLemon:
There is a separate PR that plans to add it but it hasn't gotten any attention. You can find it at https://github.com/NixOS/nixpkgs/pull/301044 but I also have a flake package at https://github.com/PassiveLemon/lemonake

LuNeder:
> There is a separate PR that plans to add it but it hasn't gotten any attention. You can find it at #301044 but I also have a flake package at [PassiveLemon/lemonake](https://github.com/PassiveLemon/lemonake)

Ohhh, I’ll take a look at those! Thanks!!

LuNeder:
Yep, that works. Thanks!

nixos/modules/services/video/wivrn.nix
PassiveLemon: No default config is defined so the server can use its own default

nixos/modules/services/video/wivrn.nix
```diff
@@ -0,0 +1,137 @@
...
+
+    services = {
+      wivrn.monadoEnvironment = {
+        # Default options
+        # https://gitlab.freedesktop.org/monado/monado/-/blob/4548e1738591d0904f8db4df8ede652ece889a76/src/xrt/targets/service/monado.in.service#L12
+        XRT_COMPOSITOR_LOG = mkDefault "debug";
+        XRT_PRINT_OPTIONS = mkDefault "on";
+        IPC_EXIT_ON_DISCONNECT = mkDefault "off";
+      };
+      udev.packages = with pkgs; [ xr-hardware ]; # WiVRn can be used with some wired headsets
```
Pandapip1, line 113:
```suggestion
      udev.packages = with pkgs; [ xr-hardware android-udev-rules ]; # WiVRn can be used with some wired headsets
```

Nit: it *probably* makes sense to install `android-udev-rules`. If you disagree, just mark this as resolved.

nixos/modules/services/video/wivrn.nix
```diff
@@ -0,0 +1,137 @@
...
+      owner = "root";
+      group = "root";
+      # cap_sys_nice needed for asynchronous reprojection
+      capabilities = "cap_sys_nice+eip";
+      source = getExe' cfg.package "wivrn-server";
+    };
+
+    systemd.user = {
+      services.wivrn = {
+        description = "WiVRn XR runtime service module";
```
Pandapip1, line 77: Nit: Is there any hardening that can be done here?

Pandapip1 (approved):
Other than the two nits, LGTM

nixos/modules/services/video/wivrn.nix
PassiveLemon: Yeah I didn't even think about that. Will add.

Like in the monado-vulkan-layers package in the other PR, would like you like me to remove the `with lib;` from meta?

nixos/modules/services/video/wivrn.nix
PassiveLemon: Personally I don't know. Xytovl can probably provide some input since they develop WiVRn, and maybe Scrumplex too since they know more about Monado than I do.

Pandapip1: It probably makes sense to set `ProtectSystem`, `InaccessiblePaths`, `ReadWritePaths`, and `ReadOnlyPaths`. For capabilities, use `CapabilityBoundingSet` and `AmbientCapabilities`.

See https://www.freedesktop.org/software/systemd/man/latest/systemd.exec.html

xytovl: The unit file is shipped upstream: https://github.com/Meumeu/WiVRn/blob/4c9f24499eb9bbea3c8949e026b6177e19e7c1a3/server/dist/wivrn.service.in
There is no sandboxing yet, I suggest raising an issue on https://github.com/Meumeu/WiVRn if you want it.

Pandapip1: Nixpkgs tries to minimally deviate from upstream. NixOS *doesn't*, and hardening options are actually explicitly called out as one of the things that are frequently adjusted: https://discourse.nixos.org/t/re-tensions-between-upstream-developers-and-package-maintainers/19556/6?u=pandapip1

nixos/modules/services/video/wivrn.nix
```diff
@@ -0,0 +1,141 @@
...
+    security.wrappers."wivrn-server" = mkIf cfg.highPriority {
+      setuid = false;
+      owner = "root";
+      group = "root";
+      # cap_sys_nice needed for asynchronous reprojection
+      capabilities = "cap_sys_nice+eip";
+      source = getExe' cfg.package "wivrn-server";
+    };
+
+    systemd.user = {
```
Pandapip1, line 76:
```suggestion
    systemd = {
```

nixos/modules/services/video/wivrn.nix
PassiveLemon: How secure do we want the service to be? I did a bit of reading and got the security from like a 9.8 down to a 5 in systemd-analyze security, all while still being functional.

Pandapip1: The goal is to provide the minimum attack surface, but even a 5 is *much* better compared to a 9.8. I would consider that more than enough hardening for a new nixpkgs module.

Pandapip1 (approved):
LGTM

PassiveLemon:
I fixed up the application portion of the config so it can find the binary and will also properly concatenate the binary path and any flags/options into a string.
It has a check for a list so the user may provide just the package, or a list with a package and flags/options.

Pandapip1 (changes requested):
The manual build is failing

PassiveLemon:
```
at /nix/store/cv4xysa8n4mh9cij9vsw8r1ydqrj7r9x-nixos/modules/services/video/wivrn.nix:67:29:
       >
       >            66|             ];
       >            67|             application = [ pkgs.wlx-overlay-s ];
       >              |                             ^
       >            68|             tcp_only = true;
       > Cacheable portion of option doc build failed.
       > Usually this means that an option attribute that ends up in documentation (eg `default` or `description`) depends on the restricted module arguments `config` or `pkgs`.
       >
```
It has something to do with `pkgs.wlx-overlay-s` in the example for the json config. I don't exactly know what that means, are packages not supposed to be used in there or something?

nixos/modules/services/video/wivrn.nix
```diff
@@ -0,0 +1,178 @@
...
+          description = ''
+            Configuration for WiVRn. The attributes are serialized to JSON in config.json.
+
+            Note that the application option must be either a package or a
+            list with package as the first element.
+
+            See https://github.com/Meumeu/WiVRn/blob/master/docs/configuration.md
+          '';
+          default = { };
+          example = literalExpression ''
```
Scrumplex, line 54:
```suggestion
          example = literalExpression ''
            {
```

nixos/modules/services/video/wivrn.nix
```diff
@@ -0,0 +1,178 @@
...
+                codec = "h264";
+                width = 1.0;
+                height = 1.0;
+                offset_x = 0.0;
+                offset_y = 0.0;
+              }
+            ];
+            application = [ pkgs.wlx-overlay-s ];
+            tcp_only = true;
+          '';
```
Scrumplex, line 69:
```suggestion
            }
          '';
```

PassiveLemon:
I've encountered a problem.
It seems like literally every single security option causes asynchronous reprojection with the security wrapper to not work.

So its either hardened but no async repro, or not hardened but you get async repro.

I can disable all of the options if highPriority is enabled but do we really want to do that?

Pandapip1:
Just remove the hardening options for now. We can add them back in once https://github.com/Meumeu/WiVRn/pull/80 is merged.

Scrumplex:
Most hardening options are probably gonna rely on user namespaces where file capabilities will not be effective anymore. This is a similar issue to #217119.

Systemd might have a way to set effective capabilities for the process during startup instead but that might not be available for user services.

PassiveLemon:
> Just remove the hardening options for now. We can add them back in once [Meumeu/WiVRn#80](https://github.com/Meumeu/WiVRn/pull/80) is merged.

Aren't we just going to add them back once it's merged? I don't see why we need to remove them

Scrumplex:
> Aren't we just going to add them back once it's merged? I don't see why they should be removed for now

This might need more investigation before the upstream PR can even be merged.

PassiveLemon:
> > Aren't we just going to add them back once it's merged? I don't see why they should be removed for now
> 
> This might need more investigation before the upstream PR can even be merged.

That's fair, I will remove them for now

Pandapip1 (approved):
Diff LGTM

PassiveLemon:
That was just a simple assertion if the application is not actually a package

Pandapip1 (approved):

xytovl (approved):

LuNeder:
Building the nixos module seems to be broken 🤔

PassiveLemon:
> Building the nixos module seems to be broken 🤔

What seems to be the issue?

LuNeder:
> > Building the nixos module seems to be broken 🤔
> 
> What seems to be the issue?

I apologize if this is just me being dumb, I'm still quite new to nix, but it seems to complain about error: attribute 'application' missing

<details>
<summary>Full error</summary>
<br>

````

error:
       … while calling anonymous lambda

         at /nix/store/bi5zxc0v5g6ylygdwyqzh280sccg3ykb-source/lib/attrsets.nix:1571:24:

         1570|     let f = attrPath:
         1571|       zipAttrsWith (n: values:
             |                        ^
         1572|         let here = attrPath ++ [n]; in

       … while calling anonymous lambda

         at /nix/store/bi5zxc0v5g6ylygdwyqzh280sccg3ykb-source/lib/attrsets.nix:1205:18:

         1204|         mapAttrs
         1205|           (name: value:
             |                  ^
         1206|             if isAttrs value && cond value

       … from call site

         at /nix/store/bi5zxc0v5g6ylygdwyqzh280sccg3ykb-source/lib/attrsets.nix:1208:18:

         1207|             then recurse (path ++ [ name ]) value
         1208|             else f (path ++ [ name ]) value);
             |                  ^
         1209|     in

       … while calling anonymous lambda

         at /nix/store/bi5zxc0v5g6ylygdwyqzh280sccg3ykb-source/lib/modules.nix:242:72:

          241|           # For definitions that have an associated option
          242|           declaredConfig = mapAttrsRecursiveCond (v: ! isOption v) (_: v: v.value) options;
             |                                                                        ^
          243|

       … while evaluating the option `system.build.toplevel':

       … while calling anonymous lambda

         at /nix/store/bi5zxc0v5g6ylygdwyqzh280sccg3ykb-source/lib/modules.nix:824:28:

          823|         # Process mkMerge and mkIf properties.
          824|         defs' = concatMap (m:
             |                            ^
          825|           map (value: { inherit (m) file; inherit value; }) (builtins.addErrorContext "while evaluating definitions from `${m.file}':" (dischargeProperties m.value))

       … while evaluating definitions from `/nix/store/bi5zxc0v5g6ylygdwyqzh280sccg3ykb-source/nixos/modules/system/activation/top-level.nix':

       … from call site

         at /nix/store/bi5zxc0v5g6ylygdwyqzh280sccg3ykb-source/lib/modules.nix:825:137:

          824|         defs' = concatMap (m:
          825|           map (value: { inherit (m) file; inherit value; }) (builtins.addErrorContext "while evaluating definitions from `${m.file}':" (dischargeProperties m.value))
             |                                                                                                                                         ^
          826|         ) defs;

       … while calling 'dischargeProperties'

         at /nix/store/bi5zxc0v5g6ylygdwyqzh280sccg3ykb-source/lib/modules.nix:896:25:

          895|   */
          896|   dischargeProperties = def:
             |                         ^
          897|     if def._type or "" == "merge" then

       … from call site

         at /nix/store/bi5zxc0v5g6ylygdwyqzh280sccg3ykb-source/nixos/modules/system/activation/top-level.nix:71:12:

           70|   # Replace runtime dependencies
           71|   system = foldr ({ oldDependency, newDependency }: drv:
             |            ^
           72|       pkgs.replaceDependency { inherit oldDependency newDependency drv; }

       … while calling 'foldr'

         at /nix/store/bi5zxc0v5g6ylygdwyqzh280sccg3ykb-source/lib/lists.nix:121:20:

          120|   */
          121|   foldr = op: nul: list:
             |                    ^
          122|     let

       … from call site

         at /nix/store/bi5zxc0v5g6ylygdwyqzh280sccg3ykb-source/lib/lists.nix:128:8:

          127|         else op (elemAt list n) (fold' (n + 1));
          128|     in fold' 0;
             |        ^
          129|

       … while calling 'fold''

         at /nix/store/bi5zxc0v5g6ylygdwyqzh280sccg3ykb-source/lib/lists.nix:124:15:

          123|       len = length list;
          124|       fold' = n:
             |               ^
          125|         if n == len

       … while calling anonymous lambda

         at /nix/store/bi5zxc0v5g6ylygdwyqzh280sccg3ykb-source/nixos/modules/system/activation/top-level.nix:64:50:

           63|
           64|   failedAssertions = map (x: x.message) (filter (x: !x.assertion) config.assertions);
             |                                                  ^
           65|

       … from call site

         at /nix/store/a5ng1p2gcwnr5f72irpi50sgchg2kzjj-source/nixos/modules/services/video/wivrn.nix:84:21:

           83|       {
           84|         assertion = isDerivation applicationBinary;
             |                     ^
           85|         message = "The application in WiVRn configuration is not a package. Please ensure that the application is a package or that a package is the first element in the list.";

       … while calling 'isDerivation'

         at /nix/store/bi5zxc0v5g6ylygdwyqzh280sccg3ykb-source/lib/attrsets.nix:1283:5:

         1282|   isDerivation =
         1283|     value: value.type or null == "derivation";
             |     ^
         1284|

       error: attribute 'application' missing

       at /nix/store/a5ng1p2gcwnr5f72irpi50sgchg2kzjj-source/nixos/modules/services/video/wivrn.nix:10:24:

            9|   applicationBinary = (
           10|     if builtins.isList cfg.config.json.application
             |                        ^
           11|     then (builtins.head cfg.config.json.application)
````
</details>

**Comment by PassiveLemon** [comment]
Ah i didn't have a check for if application was set at all...

**Comment by PassiveLemon** [comment]
Updated the application check.

All cases I have checked and validated:
- Application attribute is not set -> application not set in config and not added to path
- Application attribute is empty list -> application not set in config and not added to path
- Application is a package -> application set in config and added to service path
- Application is not a package -> asserts and notifies user
- Application list first element is package (no extra strings) -> application set in config and added to service path
- Application list first element is package (extra strings) -> application set in config with extra strings and added to service path
- Application list first element is not package -> asserts and notifies user

**Review by LuNeder — APPROVED** [review]

File: pkgs/by-name/wi/wivrn/package.nix [hunk 10]
```diff
@@ -0,0 +1,145 @@
...
+    (lib.cmakeBool "FETCHCONTENT_FULLY_DISCONNECTED" true)
+    (lib.cmakeFeature "FETCHCONTENT_SOURCE_DIR_MONADO" "${finalAttrs.monado}")
+  ];
+
+  passthru.updateScript = nix-update-script { };
+
+  meta = with lib; {
+    description = "An OpenXR streaming application to a standalone headset";
+    homepage = "https://github.com/Meumeu/WiVRn/";
+    changelog = "https://github.com/Meumeu/WiVRn/releases/tag/v${finalAttrs.version}";
```
Pandapip1, line 139: Nit: should be

```suggestion
    changelog = "https://github.com/Meumeu/WiVRn/releases";
```

By default, the tag view does *not* show the full changelog. Additionally, if the `version` is overwritten to a tag that does not exist, then this link should not update.

pkgs/by-name/wi/wivrn/package.nix
```diff
@@ -0,0 +1,145 @@
...
+    (lib.cmakeBool "WIVRN_USE_SYSTEMD" true)
+    (lib.cmakeBool "WIVRN_USE_PIPEWIRE" true)
+    (lib.cmakeBool "WIVRN_USE_PULSEAUDIO" true)
+    (lib.cmakeBool "WIVRN_BUILD_CLIENT" false)
+    (lib.cmakeBool "WIVRN_OPENXR_INSTALL_ABSOLUTE_RUNTIME_PATH" true)
+    (lib.cmakeBool "FETCHCONTENT_FULLY_DISCONNECTED" true)
+    (lib.cmakeFeature "FETCHCONTENT_SOURCE_DIR_MONADO" "${finalAttrs.monado}")
+  ];
+
+  passthru.updateScript = nix-update-script { };
```
Pandapip1, line 134: This will fail to update the hardcoded monado version.

nixos/modules/services/video/wivrn.nix
```diff
@@ -0,0 +1,170 @@
...
+  # the package, concat all of the following strings, and then update the application attribute.
+  # Application can either be a package or a list that has a package as the first element.
+  applicationExists = builtins.hasAttr "application" cfg.config.json;
+  applicationListNotEmpty = (
+    if builtins.isList cfg.config.json.application
+    then if builtins.length cfg.config.json.application == 0
+      then false
+      else true
+    else true
+  );
```
Pandapip1, line 17: How about

```suggestion
  applicationListNotEmpty = (
    if builtins.isList cfg.config.json.application
    then builtins.length cfg.config.json.application != 0
    else true
  );
```

pkgs/by-name/wi/wivrn/package.nix
```diff
@@ -0,0 +1,145 @@
...
+    ourMonadoRev="${finalAttrs.monado.src.rev}"
+    theirMonadoRev=$(grep -A1 "https://gitlab.freedesktop.org/monado/monado" ${finalAttrs.src.name}/CMakeLists.txt | tail -n1 | sed 's/.*GIT_TAG\s*//')
+    if [ ! "$theirMonadoRev" == "$ourMonadoRev" ]; then
+      echo "Our Monado source revision doesn't match CMakeLists.txt." >&2
+      echo "  theirs: $theirMonadoRev" >&2
+      echo "    ours: $ourMonadoRev" >&2
+      return 1
+    fi
+  '';
+
```
Pandapip1, line 80: Might be worth setting

```suggestion

  strictDeps = true;

```

nixos/modules/services/video/wivrn.nix
```diff
@@ -0,0 +1,170 @@
...
+  configFile = configFormat.generate "config.json" applicationUpdate;
+in
+{
+  options = {
+    services.wivrn = {
+      enable = mkEnableOption "WiVRn, an OpenXR streaming application";
+
+      package = mkPackageOption pkgs "wivrn" { };
+
+      openFirewall = mkEnableOption "the default ports in the firewall for the WiVRn server";
```
Pandapip1, line 44: Shouldn't this be true by default?

nixos/modules/services/video/wivrn.nix
```diff
@@ -0,0 +1,170 @@
...
+
+      defaultRuntime = mkEnableOption ''
+        WiVRn Monado as the default OpenXR runtime on the system.
+        The config can be found at `/etc/xdg/openxr/1/active_runtime.json`.
+
+        Note that applications can bypass this option by setting an active
+        runtime in a writable XDG_CONFIG_DIRS location like `~/.config`
+      '' // { default = true; };
```
Pandapip1, line 52: Shouldn't this *not* be true by default?

pkgs/by-name/wi/wivrn/package.nix
```diff
@@ -0,0 +1,145 @@
...
+      owner = "monado";
+      repo = "monado";
+      rev = "598080453545c6bf313829e5780ffb7dde9b79dc";
+      hash = "sha256-9LsKvIXAQpr+rpv8gDr4YfoNN+MSkXfccbIwLrWcIXg=";
+    };
+
+    patches = [
+      ("${finalAttrs.src}/patches/monado/0001-c-multi-disable-dropping-of-old-frames.patch")
+      ("${finalAttrs.src}/patches/monado/0002-ipc-server-Always-listen-to-stdin.patch")
+      ("${finalAttrs.src}/patches/monado/0003-c-multi-Don-t-log-frame-time-diff.patch")
```
SuperSandro2000, line 65:
```suggestion
      "${finalAttrs.src}/patches/monado/0001-c-multi-disable-dropping-of-old-frames.patch"
      "${finalAttrs.src}/patches/monado/0002-ipc-server-Always-listen-to-stdin.patch"
      "${finalAttrs.src}/patches/monado/0003-c-multi-Don-t-log-frame-time-diff.patch"
```

nixos/modules/services/video/wivrn.nix
```diff
@@ -0,0 +1,170 @@
...
+          '';
+        };
+      };
+    };
+  };
+
+  imports = [
+    (mkAliasOptionModule [ "services" "wivrn" "monadoEnvironment" ] [ "systemd" "user" "services" "wivrn" "environment" ])
+  ];
+
```
SuperSandro2000, line 95:
```suggestion
```
we do not want to add aliases for a new module

nixos/modules/services/video/wivrn.nix
```diff
@@ -0,0 +1,170 @@
...
+    };
+
+    services = {
+      wivrn.monadoEnvironment = {
+        # Default options
+        # https://gitlab.freedesktop.org/monado/monado/-/blob/598080453545c6bf313829e5780ffb7dde9b79dc/src/xrt/targets/service/monado.in.service#L12
+        XRT_COMPOSITOR_LOG = mkDefault "debug";
+        XRT_PRINT_OPTIONS = mkDefault "on";
+        IPC_EXIT_ON_DISCONNECT = mkDefault "off";
+      };
```
SuperSandro2000, line 141:
```suggestion
```

nixos/modules/services/video/wivrn.nix
```diff
@@ -0,0 +1,170 @@
...
+      owner = "root";
+      group = "root";
+      # We need cap_sys_nice for asynchronous reprojection
+      capabilities = "cap_sys_nice+eip";
+      source = getExe cfg.package;
+    };
+
+    systemd.user = {
+      services.wivrn = {
+        description = "WiVRn XR runtime service module";
```
SuperSandro2000, line 115:
```suggestion
        description = "WiVRn XR runtime service module";
        environment = {
          # Default options
          # https://gitlab.freedesktop.org/monado/monado/-/blob/598080453545c6bf313829e5780ffb7dde9b79dc/src/xrt/targets/service/monado.in.service#L12
          XRT_COMPOSITOR_LOG = mkDefault "debug";
          XRT_PRINT_OPTIONS = mkDefault "on";
          IPC_EXIT_ON_DISCONNECT = mkDefault "off";
        };
```

nixos/modules/services/video/wivrn.nix
SuperSandro2000: nope, we don't open firewalls by default.

nixos/modules/services/video/wivrn.nix
```diff
@@ -0,0 +1,170 @@
...
+  applicationStrings = builtins.tail cfg.config.json.application;
+
+  applicationPath = mkIf applicationCheck applicationBinary;
+
+  applicationConcat = (
+    if builtins.isList cfg.config.json.application
+    then builtins.concatStringsSep " " ([ (getExe applicationBinary) ] ++ applicationStrings)
+    else (getExe applicationBinary)
+  );
+  applicationUpdate = cfg.config.json // optionalAttrs applicationCheck { application = applicationConcat; };
```
SuperSandro2000, line 34:
```suggestion
  applicationUpdate = lib.recursiveUpdate cfg.config.json (optionalAttrs applicationCheck { application = applicationConcat; });
```

nixos/modules/services/video/wivrn.nix
SuperSandro2000, line 17:
```suggestion
  applicationListNotEmpty = builtins.isList cfg.config.json.application
    || builtins.length cfg.config.json.application != 0;
```

nixos/modules/services/video/wivrn.nix
Pandapip1: This appears to be true. Do you have a link to the documentation explaining why this is the case, because that doesn't make sense to me.

PassiveLemon:
Should I add an option for starting by default through "wantedBy = [ "multi-user.target" ];"?

Pandapip1:
> Should I add an option for starting by default through "wantedBy = [ "multi-user.target" ];"? I would think it makes sense for it to be started by automatically.

Since it's a user service, shouldn't it be `wantedBy = [ "default.target" ];`?

PassiveLemon:
> Since it's a user service, shouldn't it be `wantedBy = [ "default.target" ];`?

It think it should be, though I'm not 100% sure. We could also make it depend on graphical.target instead but I also don't know if WiVRn requires that. I'll have to check and see

Pandapip1:
WiVRn shouldn't need `graphical.target`

PassiveLemon:
Pushed something. Still waiting for the previous reviews to be resolved

Pandapip1:
> Still waiting for the previous reviews to be resolved

Only the pull request author and committers can resolve reviews (not even the review author)! You should manually resolve any resolved discussions yourself.

PassiveLemon:
> Only the pull request author and committers can resolve reviews (not even the review author)! You should manually resolve any resolved discussions yourself.

I did not know that. We still didn't come to a conclusion so i'm unsure if those discussions are resolved or not

Pandapip1:
https://github.com/NixOS/nixpkgs/pull/316975#discussion_r1661525091 and https://github.com/NixOS/nixpkgs/pull/316975#discussion_r1660421243 are unresolved nits.

https://github.com/NixOS/nixpkgs/pull/316975#discussion_r1665704811 is also an unresolved nit. The code it's pointing to is outdated. It refers to these updated lines:

https://github.com/NixOS/nixpkgs/blob/7fa26b141fd611e22acd96b5cb5a442282f83488/nixos/modules/services/video/wivrn.nix#L11-L15

Pandapip1:
https://github.com/NixOS/nixpkgs/pull/316975#discussion_r1661525091 and https://github.com/NixOS/nixpkgs/pull/316975#discussion_r1660421243 are unresolved nits.

https://github.com/NixOS/nixpkgs/pull/316975#discussion_r1665704811 is also an unresolved nit. The code it's pointing to is outdated. It refers to these updated lines:

https://github.com/NixOS/nixpkgs/blob/7fa26b141fd611e22acd96b5cb5a442282f83488/nixos/modules/services/video/wivrn.nix#L11-L15

PassiveLemon:
Updated to release 0.17

pkgs/by-name/wi/wivrn/package.nix
```diff
@@ -0,0 +1,147 @@
...
+}:
+stdenv.mkDerivation (finalAttrs: {
+  pname = "wivrn";
+  version = "0.17";
+
+  src = fetchFromGitHub {
+    owner = "meumeu";
+    repo = "wivrn";
+    rev = "v${finalAttrs.version}";
+    hash = "";
```
LuNeder, line 50: Unsure if you forgot it or will just do this later, but the hash is missing here :)

pkgs/by-name/wi/wivrn/package.nix
```diff
@@ -0,0 +1,147 @@
...
+  };
+
+  monado = applyPatches {
+    src = fetchFromGitLab {
+      domain = "gitlab.freedesktop.org";
+      owner = "monado";
+      repo = "monado";
+      rev = "28cb225be12a8a55c221f22103c235aff6de3883";
+      hash = "";
```
LuNeder, line 59: Same here

nixos/modules/services/video/wivrn.nix
```diff
@@ -0,0 +1,226 @@
...
+        mkEnableOption ''
+          WiVRn Monado as the default OpenXR runtime on the system.
+          The config can be found at `/etc/xdg/openxr/1/active_runtime.json`.
+
+          Note that applications can bypass this option by setting an active
+          runtime in a writable XDG_CONFIG_DIRS location like `~/.config`
+        ''
+        // {
+          default = true;
+        };
```
Pandapip1, line 80: Nit: I still don't think this should be true by default

nixos/modules/services/video/wivrn.nix
```diff
@@ -0,0 +1,226 @@
...
+
+      autoStart = mkEnableOption "start the service by default";
```
Pandapip1, line 82: Nit: Using mkEnableOption means that the description for this will be "Enable start the service by default", which is wrong. Override the description:

```suggestion
      autoStart = mkEnableOption "start the service by default" // {
        description = "Start the service by default";
      };
```

Or, use `mkOption` with `types.bool`.

nixos/modules/services/video/wivrn.nix
```diff
@@ -0,0 +1,226 @@
...
+      };
+    };
+  };
+
+  config = mkIf cfg.enable {
+    assertions = [
+      (mkIf applicationCheck {
+        assertion = isDerivation applicationBinary;
+        message = "The application in WiVRn configuration is not a package. Please ensure that the application is a package or that a package is the first element in the list.";
+      })
```
Pandapip1, line 135: Nit: Why not

```suggestion
      {
        assertion = !applicationCheck || isDerivation applicationBinary;
        message = "The application in WiVRn configuration is not a package. Please ensure that the application is a package or that a package is the first element in the list.";
      }
```

Pandapip1 (approved):
LGTM, despite the nits

Pandapip1 (approved):

Pandapip1 (approved):

PassiveLemon:
With so many applications that can run in the wivrn-application service, I occasionally run into PATH issues. Should I add an option so the user can add additional packages/paths into the services environment?

PassiveLemon:
Added the option anyways. extraPackages just adds the packages to the path of wivrn-application so they can be exposed to programs like wlx-overlay-s

Pandapip1 (approved):
Still LGTM. Can we a committer merge this?

Scrumplex (approved):

PassiveLemon:
Bumped to v0.18.
I haven't tested this because I recently moved and do not have my VR stuff set up.

Scrumplex:
Result of `nixpkgs-review pr 316975` run on x86_64-linux [1](https://github.com/Mic92/nixpkgs-review)
<details>
  <ul>
    <li>nixos-install-tools</li>
  </ul>
</details>
<details>
  <summary>1 package failed to build:</summary>
  <ul>
    <li>wivrn</li>
  </ul>
</details>

Scrumplex:
```
Our Monado source revision doesn't match CMakeLists.txt.
  theirs: dfc602288ab05131584a3f2be18031a13fccd061
    ours: 28cb225be12a8a55c221f22103c235aff6de3883
```

Our check works ^^

PassiveLemon:
Oh nice. forgot to update that when copying the hash from my personal package to the PR package

Scrumplex:
Result of `nixpkgs-review pr 316975` run on x86_64-linux [1](https://github.com/Mic92/nixpkgs-review)
<details>
  <ul>
    <li>nixos-install-tools</li>
  </ul>
</details>
<details>
  <summary>1 package failed to build:</summary>
  <ul>
    <li>wivrn</li>
  </ul>
</details>

Scrumplex:
```
CMake Error at /nix/store/p9hrg1d47czzsqr4zasyf3wnfdbj74zr-source-patched/cmake/SPIR-V.cmake:8 (message):
  glslangValidator required - source maintained at
  https://github.com/KhronosGroup/glslang
```

PassiveLemon:
I do not know what I need to do with this lol

Scrumplex:
Result of `nixpkgs-review pr 316975` run on x86_64-linux [1](https://github.com/Mic92/nixpkgs-review)
<details>
  <ul>
    <li>nixos-install-tools</li>
  </ul>
</details>
<details>
  <summary>1 package failed to build:</summary>
  <ul>
    <li>wivrn</li>
  </ul>
</details>

Scrumplex:
Okay, I wanted to confirm the issue. It only happens when built against master. Could you rebase your PR, as it is about 32000 commits behind?

pkgs/by-name/wi/wivrn/package.nix
```diff
@@ -0,0 +1,146 @@
...
+
+  buildInputs = [
+    avahi
+    boost
+    cli11
+    eigen
+    ffmpeg
+    freetype
+    glm
+    glslang
```
Scrumplex, line 100: You need to move `glslang` to `nativeBuildInputs` to fix the build failure.

`glslangValidator` is used during configure/build time by Monado so to facilitate cross compiling it needs to be a native build input.

```diff
diff --git a/pkgs/by-name/wi/wivrn/package.nix b/pkgs/by-name/wi/wivrn/package.nix
index 59a8a15d45e6..6c1241c5e2e4 100644
--- a/pkgs/by-name/wi/wivrn/package.nix
+++ b/pkgs/by-name/wi/wivrn/package.nix
@@ -85,6 +85,7 @@ stdenv.mkDerivation (finalAttrs: {
   nativeBuildInputs = [
     cmake
     git
+    glslang
     pkg-config
     python3
   ] ++ lib.optionals cudaSupport [ autoAddDriverRunpath ];
@@ -97,7 +98,6 @@ stdenv.mkDerivation (finalAttrs: {
     ffmpeg
     freetype
     glm
-    glslang
     harfbuzz
     libdrm
     libGL
```

PassiveLemon:
Seems to have done it

Pandapip1:
Result of `nixpkgs-review pr 316975` run on x86_64-linux [1](https://github.com/Mic92/nixpkgs-review)
<details>
  <ul>
    <li>nixos-install-tools</li>
  </ul>
</details>
<details>
  <summary>1 package failed to build:</summary>
  <ul>
    <li>wivrn</li>
  </ul>
</details>

PassiveLemon:
My nixpkgs-review succeeds, evaluated on the same commit as this PR

Pandapip1:
Weird. Mine failed to fetch sources.

Scrumplex:
Result of `nixpkgs-review pr 316975` run on x86_64-linux [1](https://github.com/Mic92/nixpkgs-review)
<details>
  <ul>
    <li>nixos-install-tools</li>
  </ul>
</details>
<details>
  <ul>
    <li>wivrn</li>
  </ul>
</details>

Pandapip1:
Result of `nixpkgs-review pr 316975` run on x86_64-linux [1](https://github.com/Mic92/nixpkgs-review)
<details>
  <ul>
    <li>nixos-install-tools</li>
  </ul>
</details>
<details>
  <summary>1 package failed to build:</summary>
  <ul>
    <li>wivrn</li>
  </ul>
</details>

Pandapip1:
Result of `nixpkgs-review pr 316975` run on x86_64-linux [1](https://github.com/Mic92/nixpkgs-review)
<details>
  <ul>
    <li>nixos-install-tools</li>
  </ul>
</details>
<details>
  <summary>1 package failed to build:</summary>
  <ul>
    <li>wivrn</li>
  </ul>
</details>

PassiveLemon:
interesting

LuNeder:
I get a 404 here too. Entering the link, I see WiVRn has moved repositories.

From WiVRn’s readme: WiVRn has moved to https://github.com/WiVRn/WiVRn

So yeah, your nix is probably using a cached tar or something (so much for reproducibility…)

PassiveLemon:
Ooooh haha that makes sense, I will update the package

LuNeder:
Ye

(I’m skeptical about how safe it is to drive a truck while wearing a vr headset tho, I’m glad the moving went fine xD)

LuNeder:
> How can I make sure I don't have any caching affecting the results like in this case?

I assume running the garbage collector could do the job? Unsure tho, never used nixpkgs-review  

(I was about to say there might be an obscure nearly-undocumented flag/option for that too as often happens, but from that issue panda just opened I assume that’s not the case)

SuperSandro2000:
This can be easily solved by running eg. `nix-build -K wivrn.src` / `nix-build --rebuild wivrn.src`

PassiveLemon:
Any update on this? Xytovl said they will make a new release probably in a week or 2 so maybe we can wait until then

PassiveLemon:
Updated the package postUnpack to fix an issue where "theirMonadoRev" becomes incorrect due to a change in the CMakeLists.txt file.

From `grep -A1 "https://gitlab.freedesktop.org/monado/monado" ${finalAttrs.src.name}/CMakeLists.txt | tail -n1 | sed 's/.*GIT_TAG\s*//'`
To `grep "GIT_TAG" ${finalAttrs.src.name}/CMakeLists.txt | awk '{print $2}'`

PassiveLemon:
Would there happen to be an easier way to apply patches that doesn't involve me listing each one?

Pandapip1:
> Would there happen to be an easier way to apply patches that doesn't involve me listing each one?

You could use `runCommand` and make a script that iterates through wivrn's patches directory and applies them to monado.

Pandapip1:
Also, just a heads up: every time you update the PR non-trivially, it requires a re-review. This is a pretty big PR, so reviews take a while. Once you're happy with it, it might make sense to just leave it alone, even letting the PR lag a couple versions behind, and only making changes requested by reviews. Once that's done, subsequent PRs to update wivrn will be able to be merged much quicker.

PassiveLemon:
I'm happy with it now, but maybe in the future, if enough patches need to be applied, I might consider refactoring the package. I'm only going to update versions until it gets merged, which hopefully won't take much longer than it already is

Pandapip1:
Result of `nixpkgs-review pr 316975` run on x86_64-linux [1](https://github.com/Mic92/nixpkgs-review)
<details>
  <ul>
    <li>nixos-install-tools</li>
  </ul>
</details>
<details>
  <ul>
    <li>disko</li>
    <li>wivrn</li>
  </ul>
</details>

Pandapip1 (approved):

Titaniumtown:
0.20 released! https://github.com/WiVRn/WiVRn/releases/tag/v0.20

Titaniumtown (approved):

PassiveLemon:
> 0.20 released! https://github.com/WiVRn/WiVRn/releases/tag/v0.20

We are waiting for this to be merged before any more commits are added. Once it's merged, it can be updated faster

Titaniumtown:
Very fair. I hope it gets merged soon. I want to try and move off of ALVR and all the steamvr specific stuff I use as steamvr is very buggy.

tomodachi94 (approved):

tomodachi94:
Sorry for the wait! Changes look good. It looks like a few of the reviewers have a lot of experience with modules, so I'm making the assumption that they caught most potential problems with this module.

Titaniumtown:
@PassiveLemon Thanks for your hard work getting this working. I've fully moved off of ALVR now in favor of WiVRn, and I'm loving it.

NovaViper:
Can't wait to give this a try! Is anyone aware if WiVRn can be used with SteamVR and the Quest 2?

Titaniumtown:
@NovaViper I just wrote docs on it! https://wiki.nixos.org/wiki/VR#WiVRn

Here's how i implemented it:
https://git.gardling.com/titaniumtown/dotfiles/src/commit/4e4f69484f68277cded571bac0daf7142c3bd6b3/home-manager/system-nixos.nix#L31
and
https://git.gardling.com/titaniumtown/dotfiles/src/commit/4e4f69484f68277cded571bac0daf7142c3bd6b3/etcnixos/vr.nix

Hope that helps!

Titaniumtown:
But you actually ditch SteamVR entirely with WiVRn. You launch the game from the desktop then put on the headset :p

NovaViper:
> @NovaViper I just wrote docs on it! https://wiki.nixos.org/wiki/VR#WiVRn
> 
> 
> 
> Here's how i implemented it:
> 
> https://git.gardling.com/titaniumtown/dotfiles/src/commit/4e4f69484f68277cded571bac0daf7142c3bd6b3/home-manager/system-nixos.nix#L31
> 
> and
> 
> https://git.gardling.com/titaniumtown/dotfiles/src/commit/4e4f69484f68277cded571bac0daf7142c3bd6b3/etcnixos/vr.nix
> 
> 
> 
> Hope that helps!

Thank you! Will take a look at this in the morning!

Titaniumtown:
Happy to help. Feel free to ping me on the matrix or here if you have any questions!

NovaViper:
> But you actually ditch SteamVR entirely with WiVRn. You launch the game from the desktop then put on the headset :p

🤔 I wonder if that would work with Beat Saber and VRChat (in particular Beat Saber because it's a bit hectic to get running because of its recently added OpenXR dependency). Hopefully I can get the NVEC stuff working (is actually what's currently holding up the current nixpkgs PR for the latest version for ALVR).

Titaniumtown:
I can confirm it works with Beat Saber! It was the first game I tested.

Compared to ALVR. I get way better visual fidelity on my Quest 2. ALVR always seemed a bit blurry and weird no matter how much i tweaked it.

Only downside over WiVRn is that you don't get all of the steam niceties like launching your games *from* VR, because that's all SteamVR stuff. I guess you could use [wlx-overlay-s](https://github.com/galister/wlx-overlay-s) and use steam from there though haha.

LuNeder:
> Only downside over WiVRn is that you don't get all of the Steam niceties like launching your games _from_ VR

I think you might be able to use StardustXR for that

NovaViper:
@Titaniumtown Hey sorry for the lengthy delay in my reply! I finally got around to installing it but I can't figure out how to get the games to launch.. Beat Saber oddly crashes WiVRn with Segmentation Fault errors and VRChat just screams at me that it couldn't initialize EasyAntiCheat 😭

NovaViper:
Hey thanks @PassiveLemon! Adding monado-vulkan-layers to the `hardware.graphics.extraPackages` made it actually stop segfaulting! I hadn't been able to get VRChat working, I can get it to go past the EAC issues but then it just immedately closes with `GLib-GObject-CRITICAL **: 00:44:54.893: g_object_unref: assertion 'G_IS_OBJECT (object)' failed` in the journal. 😭 

I'm also looking at setting up StardustXR, is there currently anyway to set it up in NixOS along with WiVRn?

Pandapip1:
> is there currently anyway to set it up in NixOS along with WiVRn?

*Theoretically*, all you need to do is install stardust-xr-server, run it, and install and run stardust-xr-protostar (once it lands in unstable).

I haven't had much luck with getting stardust to work back when I had a headset to test with. I don't currently have access to a working headset, which makes testing significantly harder, so bug reports are *very* welcome.

NovaViper:
@Pandapip1 I gave it a try and uh.... it immediately crashes with my system 😅 . It tries to connect to WiVRn but then immediately disconnects and spams this in the terminal (completely floods it actually!):
```
ERROR [ipc_send] sendmsg(136) failed: '32' 'Broken pipe'!
ERROR [ipc_client_session_poll_events] ipc_call_session_poll_events failed: XRT_ERROR_IPC_FAILURE [/nix/store/BFF96Z4FKNP9SY4RK6JN084IPWD8QYK4-source-patched/src/xrt/ipc/client/ipc_client_session.c:56]
XR_ERROR_INSTANCE_LOST in xrPollEvent: Call to "xrt_session_poll_events" failed
XR_ERROR_SESSION_LOST in xrSyncActions: Session is lost
XR_ERROR_SESSION_LOST in xrLocateSpace: Session is lost
XR_ERROR_SESSION_LOST in xrGetActionStatePose: Session is lost
XR_ERROR_SESSION_LOST in xrLocateSpace: Session is lost
XR_ERROR_SESSION_LOST in xrGetActionStatePose: Session is lost
XR_ERROR_SESSION_LOST in xrLocateSpace: Session is lost
XR_ERROR_SESSION_LOST in xrGetActionStatePose: Session is lost
XR_ERROR_SESSION_LOST in xrLocateSpace: Session is lost
XR_ERROR_SESSION_LOST in xrGetActionStatePose: Session is lost
XR_ERROR_SESSION_LOST in xrLocateSpace: Session is lost
XR_ERROR_SESSION_LOST in xrGetActionStateFloat: Session is lost
XR_ERROR_SESSION_LOST in xrGetActionStateFloat: Session is lost
XR_ERROR_SESSION_LOST in xrGetActionStateFloat: Session is lost
XR_ERROR_SESSION_LOST in xrGetActionStateFloat: Session is lost
XR_ERROR_SESSION_LOST in xrGetActionStateBoolean: Session is lost
XR_ERROR_SESSION_LOST in xrGetActionStateBoolean: Session is lost
XR_ERROR_SESSION_LOST in xrGetActionStateBoolean: Session is lost
XR_ERROR_SESSION_LOST in xrGetActionStateBoolean: Session is lost
XR_ERROR_SESSION_LOST in xrGetActionStateFloat: Session is lost
XR_ERROR_SESSION_LOST in xrGetActionStateFloat: Session is lost
XR_ERROR_SESSION_LOST in xrGetActionStateFloat: Session is lost
XR_ERROR_SESSION_LOST in xrGetActionStateFloat: Session is lost
XR_ERROR_SESSION_LOST in xrGetActionStateBoolean: Session is lost
XR_ERROR_SESSION_LOST in xrGetActionStateBoolean: Session is lost
XR_ERROR_SESSION_LOST in xrGetActionStateBoolean: Session is lost
XR_ERROR_SESSION_LOST in xrGetActionStateBoolean: Session is lost
XR_ERROR_SESSION_LOST in xrLocateHandJointsEXT: Session is lost
XR_ERROR_SESSION_LOST in xrLocateHandJointsEXT: Session is lost
```

And the only way to kill the process is by running `killall stardust-xr-server -9v`

Pandapip1:
From the StardustXR discord:

image

technobaboo:
btw for stardust you should run the flake at https://github.com/StardustXR/telescope because that has a full stardust setup for you ready to go :)

NovaViper:
@technobaboo Hey does the telescope flake work with the systemd service that the wivrn module creates?

technobaboo:
uhhhhhhhhhhhh i have no idea, you can easily launch the executable though?

NovaViper:
Uh... it dies still even with telescope 👀 
```
❯ nix run github:StardustXR/telescope
[SK info] Before init >>>
libEGL warning: egl: failed to create dri2 screen
LOG in xrCreateInstance: Instance created
        createInfo->applicationInfo.applicationName: Stardust XR
        createInfo->applicationInfo.applicationVersion: 1
        createInfo->applicationInfo.engineName: StereoKit
        createInfo->applicationInfo.engineVersion: 16384
        createInfo->applicationInfo.apiVersion: 1.1.38
        appinfo.detected.engine.name: (null)
        appinfo.detected.engine.version: 0.0.0
        quirks.disable_vulkan_format_depth_stencil: false
        quirks.no_validation_error_in_create_ref_space: true
        quirks.skip_end_session: false
        quirks.parallel_views: false

LOG in xrCreateInstance: Selected devices
        Head: 'WiVRn HMD'
        Eyes: '<none>'
        Left: 'WiVRn HMD left hand controller'
        Right: 'WiVRn HMD right hand controller'
        Gamepad: '<none>'
        Hand-Tracking Left: 'WiVRn HMD left hand controller'
        Hand-Tracking Right: 'WiVRn HMD right hand controller'
ERROR [ipc_compositor_get_reference_bounds_rect] ipc_call_compositor_get_reference_bounds_rect failed: XRT_ERROR_NOT_IMPLEMENTED [/nix/store/BFF96Z4FKNP9SY4RK6JN084IPWD8QYK4-source-patched/src/xrt/ipc/client/ipc_client_compositor.c:868]
ERROR [ipc_compositor_get_reference_bounds_rect] ipc_call_compositor_get_reference_bounds_rect failed: XRT_ERROR_NOT_IMPLEMENTED [/nix/store/BFF96Z4FKNP9SY4RK6JN084IPWD8QYK4-source-patched/src/xrt/ipc/client/ipc_client_compositor.c:868]
xkbcommon: ERROR: Unrecognized RMLVO option "terminate:ctrl_alt_bksp" was ignored
xkbcommon: ERROR: Unrecognized RMLVO option "caps:ctrl_modifier" was ignored
[src/main.rs:27:13] Args::parse() = Args {
    x: 0.0,
    y: 0.0,
    z: -0.5,
    yaw: None,
    command: [
        "hexagon_launcher",
    ],
}
Setting connection env var STARDUST_INSTANCE to stardust-0
Setting connection env var QT_QPA_PLATFORM to wayland
Setting connection env var SDL_VIDEODRIVER to wayland
Setting connection env var GDK_BACKEND to wayland
Setting connection env var MOZ_ENABLE_WAYLAND to 1
Setting connection env var WAYLAND_DISPLAY to wayland-1
Setting connection env var FLAT_WAYLAND_DISPLAY to wayland-0
Setting connection env var CLUTTER_BACKEND to wayland
Io error: Connection reset by peer (os error 104)
thread 'main' panicked at /build/source/src/server/mod.rs:713:39:
called `Result::unwrap()` on an `Err` value: Io(Os { code: 104, kind: ConnectionReset, message: "Connection reset by peer" })
Error: IO Error: Connection reset by peer (os error 104)

Location:
    src/main.rs:44:27
The application panicked (crashed).
Message:  called `Result::unwrap()` on an `Err` value: MessengerError { e: ReceiverDropped }
Location: hexagon_launcher/src/main.rs:47

Backtrace omitted. Run with RUST_BACKTRACE=1 environment variable to display it.
Run with RUST_BACKTRACE=full to include source snippets.
 2024-11-14T17:17:22.918Z INFO  xwayland_process > (EE) could not connect to wayland server
thread 'main' panicked at src/flatland.rs:43:37:
called `Option::unwrap()` on a `None` value
/nix/store/g0455djm5pz6vb6zik744blzpaw6vi76-telescope/bin/telescope: line 8: 1536969 Segmentation fault      (core dumped) stardust-xr-server -d -o 1 -e "/nix/store/yvlc4yl3dhq1na7hv9qsykawsh48lacs-startup_script/bin/startup_script" "$@"
```

[wivrn-with-telescope.log](https://github.com/user-attachments/files/17753919/wivrn-with-telescope.log)

technobaboo:
what in the world is going on here, it looks like literally nothing can connect via unix domain sockets somehow?

LuNeder:
> what in the world is going on here, it looks like literally nothing can connect via unix domain sockets somehow? is this in a sandbox?

The systemd service is hardened, that might be related

```nix
            # Hardening options
            CapabilityBoundingSet = [ "CAP_SYS_NICE" ];
            AmbientCapabilities = [ "CAP_SYS_NICE" ];
            LockPersonality = true;
            NoNewPrivileges = true;
            PrivateTmp = true;
            ProtectClock = true;
            ProtectControlGroups = true;
            ProtectKernelLogs = true;
            ProtectKernelModules = true;
            ProtectKernelTunables = true;
            ProtectProc = "invisible";
            ProtectSystem = "strict";
            RemoveIPC = true;
            RestrictNamespaces = true;
            RestrictSUIDSGID = true;
```

technobaboo:
yeah idk, the files are all in `$XDG_RUNTIME_DIR` so if that's affected that'd be why this is all catastrophically failing

NovaViper:
Is there a way to fix it on my end? I wonder if there needs to be a module made available on NixOS to make it function properly with the wivrn NixOS module 🤔

technobaboo:
i honestly have no clue what's happening here, the way stardust works is that the server creates a unix domain socket at `$XDG_RUNTIME_DIR/stardust-0` and `$XDG_RUNTIME_DIR/wayland-<lowest free number>` and monado/wivrn make `$XDG_RUNTIME_DIR/monado_comp_ipc` there and then the stardust clients, wayland apps, openxr runtime libraries, etc. connect to those and send messages.

PR #316993, opened 2024-06-03
webcord: pass --enable-wayland-ime to electron under wayland session with NIXOS_OZONE_WL=1

alyraffauf:
## Description of changes

Pass `--enable-wayland-ime` flag to electron for better virtual keyboard and input method support under wayland sessions when NIXOS_OZONE_WL=1.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/by-name/we/webcord/package.nix
+++ b/pkgs/by-name/we/webcord/package.nix
@@ -55,7 +55,7 @@ buildNpmPackage rec {
       # Add xdg-utils to path via suffix, per PR #181171
       makeWrapper '${lib.getExe electron}' $out/bin/webcord \
         --suffix PATH : "${binPath}" \
-        --add-flags "\''${NIXOS_OZONE_WL:+\''${WAYLAND_DISPLAY:+--ozone-platform-hint=auto --enable-features=WaylandWindowDecorations}}" \
+        --add-flags "\''${NIXOS_OZONE_WL:+\''${WAYLAND_DISPLAY:+--ozone-platform-hint=auto --enable-features=WaylandWindowDecorations --enable-wayland-ime}}" \
         --add-flags $out/lib/node_modules/webcord/
 
       runHook postInstall


eclairevoyant:
I gave my thoughts here: https://github.com/NixOS/nixpkgs/pull/311937#issuecomment-2112756777

alyraffauf:
oops, totally missed that! Had the idea weeks ago before that draft was posted but only authored this commit today.

my main rationale for this is to be able to easily use --enable-wayland-ime, which is crucial for getting virtual keyboard support on electron apps, and makes it consistent with the way I do it with chrome, brave, and all the other electron apps I use. Without that override, the virtual keyboard is never summoned, making the app totally inaccessible for anyone who needs a touch keyboard.

In my opinion it's a far more flexible and consistent solution when you are doing small overrides like this for several applications, or want to easily configure it with options in your own modules.

huantianad:
I wonder if instead of introducing this argument which could be a bit of extra work in the future if we need to remove, we should just copy vesktop and always add the wayland ime flag...

eclairevoyant:
I'm fine with adding the wayland-ime flag as long as it's gated behind some arg or envvar. Most likely `NIXOS_OZONE_WL` itself, same as the existing flags.

alyraffauf:
> I'm fine with adding the wayland-ime flag as long as it's gated behind some arg or envvar. Most likely `NIXOS_OZONE_WL` itself, same as the existing flags.
> That being said I don't really have a setup to test it, as I only use sway

I can update my PR after testing this. That should work for tablet and virtual keyboard use (tested on sway, hyprland, and plasma), and should also work for fctx users as well, per their wiki.

eclairevoyant:
afaik `--enable-wayland-ime` only works with `zwp_text_input_v1`, while sway was specifically called out [as an example here](https://issues.chromium.org/issues/40113488) of a compositor that only supports v3, and chromium devs have explicitly said (due to some design decisions and usecases around ChromiumOS) that they intend to stick to v1 for the foreseeable future.

I don't know how up-to-date this other (unofficial) site (wayland.app) is, but you can see the support tables for [`zwp_input_method_v1`](https://wayland.app/protocols/input-method-unstable-v1#compositor-support) and [`zwp_text_input_manager_v3`](https://wayland.app/protocols/text-input-unstable-v3#compositor-support). If that's still the case, I don't see how this would work in sway.

alyraffauf:
> afaik `--enable-wayland-ime` only works with `zwp_text_input_v1`, while sway was specifically called out [as an example here](https://issues.chromium.org/issues/40113488) of a compositor that only supports v3, and chromium devs have explicitly said (due to some design decisions and usecases around ChromiumOS) that they intend to stick to v1 for the foreseeable future.
> 
> I don't know how up-to-date this other (unofficial) site (wayland.app) is, but you can see the support tables for [`zwp_input_method_v1`](https://wayland.app/protocols/input-method-unstable-v1#compositor-support) and [`zwp_text_input_manager_v3`](https://wayland.app/protocols/text-input-unstable-v3#compositor-support). If that's still the case, I don't see how this would work in sway.
> 
> And fwiw, I've never gotten Fcitx5 to work in any Ozone Wayland app, ever (again, only tested on sway).

This works perfectly fine for me with squeekboard running in the background, it summons the keyboard as expected and I'm able to type text. It also works with kwin, and it's been part of my workflow on sway for a couple weeks now. That said, I haven't got it working with Qt apps on Sway (just GTK/Electron), but maliit does work with Qt/GTK/Electron apps on plasma.

PR has been updated to pass `--enable-wayland-ime` if NIXOS_OZONE_WL=1 is set.

eclairevoyant (approved):

PR #317026, opened 2024-06-03
nixos/grub: add new implementation of `install-grub.pl`

pluiedev:
## Description of changes

Following in the footsteps of Perlless Activation (#270727) and Perlless `switch-to-configuration` (#308801), it's now time for Perlless `install-grub`, which can be enabled by simply setting `boot.loader.grub.useInstallNg` to `true`.

This is still a draft PR for now since I don't use GRUB myself, and I could do with some help from GRUB users who could actually test this :) The GRUB config generation logic seems to work pretty well, though.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---


pluiedev:
Found a couple of (serious) bugs, fortunately there's a test I can use to catch them... :D

pluiedev:
Ready for review

nyabinary:
Should probably post this in the discourse thread of prs ready to review :P

pkgs/by-name/in/install-grub-ng/src/builder/entries.rs
```diff
@@ -0,0 +1,424 @@
...
+
+		let xen = path.join("xen.gz");
+		let xen = if xen.exists() {
+			Some((
+				self.copy_to_kernels_dir(&xen)?,
+				fs::read_to_string(path.join("xen-params")).unwrap_or_default(),
+			))
+		} else {
+			None
+		};
```
SigmaSquadron, line 261: Has the Xen stuff been tested against #324911? The Perl implementation was broken so we simply dropped BIOS support there.

pluiedev: To be honest, I have no idea as I've never used Xen myself. It's probably a good idea to drop this as it is after all a direct port of a purportedly broken script

SigmaSquadron: Agreed. The team would rather just support `systemd-boot` and Lanzaboote, so this can be dropped.

Scrumplex:
@ofborg test grub grub-ng

pkgs/by-name/in/install-grub-ng/src/config.rs
```diff
@@ -0,0 +1,292 @@
...
+      impl<'a, 'input: 'a> FromNode<'a, 'input> for $ty {
+      	fn from_node(node: Node<'a, 'input>) -> Result<Self> {
+		      check_tag_name(node, "int", |node| {
+		      	value(node)?.parse::<$ty>().map_err(|e| eyre!("Invalid int: {e}"))
+		      })
+      	}
+      }
+    )*
+  }
+}
```
Scrumplex, line 269: Mixed tabs/spaces

Could you run rustfmt over all your changes?

KiaraGrouwstra:
## `nixpkgs-review` result

Command: `nixpkgs-review pr 317026`

---
### `x86_64-linux`
<details>
  <summary>:fast_forward: 2 packages blacklisted:</summary>
  <ul>
    <li>nixos-install-tools</li>
  </ul>
</details>
<details>
  <ul>
    <li>install-grub-ng</li>
  </ul>
</details>

pkgs/by-name/in/install-grub-ng/src/config.rs
```diff
@@ -0,0 +1,292 @@
...
+
+use eyre::{Context, Result, bail, eyre};
+use roxmltree::{Document, Node};
+
+use crate::grub::FsIdentifier;
+
+config! {
+  'a;
+  grub: Option<&'a Path> => grub,
+  grub_target: Option<&'a Path> => grubTarget,
```
ThinkChaos, line 11: Why not use serde?  
We could also use something else than XML when using NG.

pkgs/by-name/in/install-grub-ng/src/main.rs
```diff
@@ -0,0 +1,38 @@
...
+
+use crate::{builder::Builder, config::Config};
+
+fn main() -> Result<()> {
+    let mut args = std::env::args();
+    let Some(config_file) = args.nth(1) else {
+        bail!("Config file not given: expected it to be the first argument")
+    };
+    let Some(default_config) = args.next() else {
+        bail!("Default config not given: expected it to be the second argument")
```
ThinkChaos, line 18: Short and (hopefully) clearer messages:
```suggestion
        bail!("Missing argv[1]: config file")
    };
    let Some(default_config) = args.next() else {
        bail!("Missing argv[2]: default config file")
```

pkgs/by-name/in/install-grub-ng/src/main.rs
```diff
@@ -0,0 +1,38 @@
...
+use std::path::Path;
+
+use eyre::{Result, bail};
+use roxmltree::Document;
+
+use crate::{builder::Builder, config::Config};
+
+fn main() -> Result<()> {
+    let mut args = std::env::args();
+    let Some(config_file) = args.nth(1) else {
```
ThinkChaos, line 14: style nit: For easier changing:
```suggestion
    let mut args = std::env::args().skip(1);

    let Some(config_file) = args.next() else {
```

pkgs/by-name/in/install-grub-ng/src/main.rs
```diff
@@ -0,0 +1,38 @@
...
+    };
+
+    let document_file = std::fs::read_to_string(config_file)?;
+    let document = Document::parse(&document_file)?;
+
+    let config = Config::new(&document)?;
+
+    eprintln!("updating GRUB 2 menu...");
+
+    std::env::set_var("PATH", config.path);
```
ThinkChaos, line 28:
```suggestion
    // TODO: move this to the wrapper script
    std::env::set_var("PATH", config.path);
```

pkgs/by-name/in/install-grub-ng/src/grub.rs
```diff
@@ -0,0 +1,325 @@
...
+            Some("by-uuid") => Self::Uuid,
+            _ => return None,
+        };
+
+        Some(format!(
+            "{} {}",
+            ty.to_flag(),
+            components.as_path().display()
+        ))
+    }
```
ThinkChaos, line 290: Don't need to mess with components here:
```suggestion
    fn provided_search(device: &Path) -> Option<String> {
        // If the provided dev is identifying the partition using a label or uuid,
        // we should get the label / uuid and ask Grub to do a proper search
        let flag = match device.to_str() {
            Some(s) if s.starts_with("/dev/disk/by-label/") => "--label",
            Some(s) if s.starts_with("/dev/disk/by-uuid/") => "--fs-uuid",
            _ => return None,
        };
    
        Some(format!(
            "{}={}",
            flag,
            device.file_name()?.to_str().expect("path basename must be valid UTF-8")
        ))
    }
```

pkgs/by-name/in/install-grub-ng/src/builder/install.rs
```diff
@@ -0,0 +1,456 @@
...
+        };
+
+        Ok(self)
+    }
+
+    fn append_prepare_config(&self) -> Result<()> {
+        let extra_prepare_config = self
+            .config
+            .extra_prepare_config
+            .replace("@bootPath@", &self.config.boot_path.to_string_lossy());
```
ThinkChaos, line 58: The "lossy" `Path` funcs shouldn't be used, especially when we're already in a function returning a `Result`.
And we don't need to allocate a `String` to then use it as a `&str`:
```suggestion
        let boot_path = self
                .config
                .boot_path
                .to_str()
                .ok_or_else(|| eyre!("boot path is not valid UTF-8: {}", ))?;

        let extra_prepare_config = self
            .config
            .extra_prepare_config
            .replace("@bootPath@", boot_path);
```

pkgs/by-name/in/install-grub-ng/src/builder/entries.rs
```diff
@@ -0,0 +1,401 @@
...
+                    .canonicalize()?
+                    .parent()
+                    .ok_or_else(|| {
+                        eyre!("Somehow {}/kernel doesn't have a parent..?", link.display())
+                    })?
+                    .join("lib/modules");
+
+                let Some(version) = fs::read_dir(&modules)?.find_map(|m| {
+                    m.ok()
+                        .and_then(|p| Some(p.path().file_name()?.to_string_lossy().into_owned()))
```
ThinkChaos, line 197: Please replace `lossy` with an error since it's easy here.

pkgs/by-name/in/install-grub-ng/src/builder/entries.rs
```diff
@@ -0,0 +1,401 @@
...
+                };
+
+                version
+            };
+
+            let entry_name =
+                fs::read_to_string(link.join("configuration-name")).unwrap_or_else(|_| {
+                    format!(
+                        "({} - {date} - {version})",
+                        link.file_name().unwrap_or_default().to_string_lossy()
```
ThinkChaos, line 209: Please replace `lossy` with an error since it's easy here.

pkgs/by-name/in/install-grub-ng/src/builder/entries.rs
```diff
@@ -0,0 +1,401 @@
...
+            bail!("Path {} is not in /nix/store!", path.display())
+        };
+
+        // GRUB store exists, which means the kernels and initrds are on the same
+        // filesystem as / and /nix/store. No need to copy!
+        if let Some(store) = &self.grub_store {
+            return Ok(store.path.join(path));
+        }
+
+        let name = path.to_string_lossy().replace('/', "-");
```
ThinkChaos, line 374: Please replace `lossy` with an error since it's easy here.

pkgs/by-name/in/install-grub-ng/src/builder.rs
```diff
@@ -0,0 +1,153 @@
...
+            let grub = config.boot_path.join("grub");
+            fs::create_dir_all(&grub)?;
+            fs::set_permissions(&grub, fs::Permissions::from_mode(0o700))?;
+        }
+
+        // Discover whether the bootPath is on the same filesystem as / and
+        // /nix/store.  If not, then all kernels and initrds must be copied
+        // to the bootPath.
+        if let Ok(metadata) = config.boot_path.metadata() {
+            if metadata.st_dev() != Path::new("/nix/store").metadata()?.st_dev() {
```
ThinkChaos, line 47: The nix store can technically be at a different path, ideally it would be passed in via the config using `builtins.storeDir`.
Just a TODO would be ok for now.

pkgs/by-name/in/install-grub-ng/src/builder/appearance.rs
```diff
@@ -0,0 +1,185 @@
...
+            return Ok(());
+        };
+
+        let mut target = PathBuf::from("background");
+
+        let ext = if let Some(ext) = splash_image.extension() {
+            if ext == "jpg" {
+                "jpeg".into()
+            } else {
+                ext.to_string_lossy()
```
ThinkChaos, line 78: Please replace `lossy` with an error since it's easy here.

pkgs/by-name/in/install-grub-ng/src/builder/appearance.rs
```diff
@@ -0,0 +1,185 @@
...
+            ..
+        } = &self.config;
+
+        let Some(splash_image) = splash_image else {
+            return Ok(());
+        };
+
+        let mut target = PathBuf::from("background");
+
+        let ext = if let Some(ext) = splash_image.extension() {
```
ThinkChaos, line 74: style nit: This whole expression can be a `match`.

pkgs/by-name/in/install-grub-ng/src/builder/entries.rs
```diff
@@ -0,0 +1,401 @@
...
+
+        Ok(self)
+    }
+
+    fn append_default_entries(&mut self) -> Result<()> {
+        // extraEntries could refer to @bootRoot@, which we have to substitute
+        let extra_entries = self
+            .config
+            .extra_entries
+            .replace("@bootRoot@", &self.grub_boot.path.to_string_lossy());
```
ThinkChaos, line 29: Please replace `lossy` with an error since it's easy here.

pkgs/by-name/in/install-grub-ng/src/builder/entries.rs
```diff
@@ -0,0 +1,401 @@
...
+                "Profile `{}` somehow does not have a file name!",
+                profile.display()
+            )
+        };
+
+        let mut links = fs::read_dir(parent)?
+            .filter_map(|m| {
+                let m = m.ok()?;
+                let filename = m.file_name();
+                let file = filename.to_string_lossy();
```
ThinkChaos, line 98: Please replace `lossy` with an error since it's easy here.

nixos/modules/system/boot/loader/grub/grub.nix
```diff
@@ -738,13 +746,23 @@ in
...
-      '' + flip concatMapStrings cfg.mirroredBoots (args: ''
-        ${perl}/bin/perl ${install-grub-pl} ${grubConfig args} $@
-      '') + cfg.extraInstallCommands);
+          ng = pkgs.install-grub-ng.override {
+            inherit (config.system.nixos) distroName;
+          };
+          genRun = args: if cfg.useInstallNg then
+            "${lib.getExe ng} ${grubConfig args} $@\n"
+          else
+            "${lib.getExe perl} ${install-grub-pl} ${grubConfig args} $@\n";
```
ThinkChaos, line 755: Might as well fix the shell quoting issue:
```suggestion
            ''${lib.getExe ng} ${grubConfig args} "$@"\n''
          else
            ''${lib.getExe perl} ${install-grub-pl} ${grubConfig args} "$@"\n'';
```

pkgs/by-name/in/install-grub-ng/src/builder.rs
ThinkChaos, line 1: Having a `builder.rs` file and `builder/` dir seems weird. Maybe move this to `builder/mod.rs`?

pkgs/by-name/in/install-grub-ng/src/builder.rs
```diff
@@ -0,0 +1,153 @@
...
+    inner: String,
+}
+impl<'conf> Builder<'conf> {
+    pub fn new(mut config: Config<'conf>, default_config: &'conf Path) -> Result<Self> {
+        let dry_run = std::env::var("DRY_RUN").as_deref() == Ok("true");
+
+        if !dry_run {
+            let grub = config.boot_path.join("grub");
+            fs::create_dir_all(&grub)?;
+            fs::set_permissions(&grub, fs::Permissions::from_mode(0o700))?;
```
ThinkChaos, line 40: Any parent dirs will have unexpected permissions here.

pkgs/by-name/in/install-grub-ng/src/builder.rs
```diff
@@ -0,0 +1,153 @@
...
+};
+
+pub struct Builder<'conf> {
+    config: Config<'conf>,
+
+    grub_boot: Grub,
+    grub_store: Option<Grub>,
+
+    default_config: &'conf Path,
+    pub copied: HashSet<PathBuf>,
```
ThinkChaos, line 27: Might've missed something, but I don't think this needs to be `pub`.

pkgs/by-name/in/install-grub-ng/src/builder.rs
```diff
@@ -0,0 +1,153 @@
...
+
+    dry_run: bool,
```
ThinkChaos, line 29: It seems error prone to have `!dry_run` every where. Could be nice to have a wrapper for each `fs` function we use that checks the flag automatically.
A TODO is fine by me if you like the idea:
```suggestion
    // TODO: create wrappers for fs functions we use that check `$DRY_RUN` automatically
    dry_run: bool,
```

pkgs/by-name/in/install-grub-ng/src/builder/entries.rs
```diff
@@ -0,0 +1,401 @@
...
+
+        let kernel_dir = self.copy_to_kernels_dir(&kernel_dir)?;
+        let initrd_dir = self.copy_to_kernels_dir(&initrd_dir)?;
+
+        // Include second initrd with secrets
+        let secrets_dir = self
+            .append_initrd_secrets(name, path, current)?
+            .unwrap_or_default();
+
+        // FIXME: $confName
```
ThinkChaos, line 245: What does this mean? Maybe put more details in the comment.

ThinkChaos (commented):
Thank you for working on this!
It's what's blocking me from going perl-less. I'll give it a try when I have time, though my setup is not likely to trigger many edge cases.

Generally it seems like we should use `eyre` a lot more: most code just uses `?` without adding any context. That'll lead to errors users can't understand :slightly_frowning_face: 
IMO, it's fine for a prototype to do that, but the best time for error handling is when you write the code, and the second best is when there's a full review. Otherwise you end up having to go over every thing again and it's very easy to miss some places.

I haven't done a full review but already spent enough time one this now. Here's the comments I wrote for now.
Don't hesitate to tell me I'm wrong or whatever in the comments :)

pkgs/by-name/in/install-grub-ng/src/builder.rs
```diff
@@ -0,0 +1,153 @@
+mod appearance;
+mod entries;
+mod install;
+
+use std::{
+    collections::HashSet,
+    fmt::Write as _,
```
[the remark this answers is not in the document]
pluiedev: This is required for `write(ln)!` unfortunately as otherwise it introduces a name conflict with `std::io::Write`

pkgs/by-name/in/install-grub-ng/src/builder.rs
pluiedev: I prefer new-style (mod.rs-less) modules trees

pkgs/by-name/in/install-grub-ng/src/builder/entries.rs
pluiedev: This is from the original script, and I have no idea what it means either

pkgs/by-name/in/install-grub-ng/src/builder.rs
ThinkChaos: I forgot that part of the RFC, fine by me as is!

pkgs/by-name/in/install-grub-ng/src/config.rs
pluiedev: Serde doesn't really have a standard API that parses *normal* XML well, let alone the weird format that `builtins.toXML` generates. I think we should really just switch to a saner format like JSON

pkgs/by-name/in/install-grub-ng/src/main.rs
pluiedev: IMO this is too brief — we don't live in an era constrained by storage sizes anymore so might as well give the user longer but more helpful error messages

pkgs/by-name/in/install-grub-ng/src/builder/entries.rs
ThinkChaos: It seems to date back to https://github.com/NixOS/nixpkgs/commit/f07f221f0ee46e64bb8a51e04dec6bade8645cfa. I say we just remove it. If it still makes sense, I'm sure we'll figure it out again at some point, no use keeping a 12 year old FIXME no one understands :stuck_out_tongue:

pluiedev: Geez it's *that* old? ...Guess the Perl parts of Nixpkgs are the most ancient after all

pkgs/by-name/in/install-grub-ng/src/builder.rs
pluiedev: I've been thinking about it and I think fs wrappers are not really enough since it also runs a lot of extra logic (e.g. running os-prober) when not in dry run mode

symphorien:
(I didn't read the PR but the os prober test passes, great!)

pkgs/by-name/in/install-grub-ng/src/builder.rs
ThinkChaos: If `create_dir_all` creates any parent directories, they won't use the permission we set just below.

I'm not sure what's the best thing to do here.

xddxdd:
I tested this and the `grub.cfg` generated contains errors that would prevent boot. The kernel/initrd references `($drive1)/kernels/name` instead of the actual path to kernel/initrd.

Example entry from this PR:

```
menuentry "NixOS - Configuration 730 (2024-11-25 - 25.05pre-git)" --class nixos {
  savedefault
search --set=drive1 --fs-uuid 5573-574D
  linux ($drive1)/kernels/name init=/nix/store/myp185i80v1ibjicj33bv1jqm8vwzk98-nixos-system-lt-hp-omen-25.05pre-git/init [args...]
  initrd ($drive1)/kernels/name 
}
```

Example entry from old script:

```
menuentry "NixOS - Configuration 730 (2024-11-24 - 25.05pre-git)" --class nixos {
  savedefault
search --set=drive1 --fs-uuid 5573-574D
  linux ($drive1)//kernels/i52qc8wmsbldpzh5sqy3591c9lzwa6pz-linux-cachyos-latest-6.12.1-bzImage init=/nix/store/myp185i80v1ibjicj33bv1jqm8vwzk98-nixos-system-lt-hp-omen-25.05pre-git/init [args...]
  initrd ($drive1)//kernels/hh0dadclbbfnbrd3xhc1nbdy8ppskrfz-initrd-linux-cachyos-latest-6.12.1-initrd
}
```

I have an impermanence setup. This is my partition setup for reference:
```
Device           Start        End    Sectors  Size Type
/dev/nvme2n1p1    2048    1050623    1048576  512M EFI System - /boot partition, FAT32
/dev/nvme2n1p2 1050624 7814037134 7812986511  3.6T Linux filesystem - /nix partition, Btrfs
```

pluiedev:
Might have missed a variable reference there when porting? (Note to self: never touch Perl ever again)

rnhmjoj:
So, you could say I'm one of the maintainers of grub: I use it on both EFI and non-EFI systems and I wrote the NixOS tests.
I'll try to test this when I have some spare time.

As a first impression: can we try to keep it simpler?
This implementation is almost 2000 lines split over 7 files: it looks quite impenetrable to me.
Compared, install-grub.pl is just 800 lines of linear imperative code: I understand it well enough even without knowing perl.
Say, recently I was extending the script to generate [boot loader entries](https://github.com/NixOS/nixpkgs/pull/95901), here I wouldn't even know to where start.

pluiedev:
Most of the complexity is simply handling the error cases that the Perl code just skips over — more rigorous input parsing, more rigorous subprocess spawning, etc. Rust is in general a much more explicit language than Perl (which is infamous for being so terse that it's impenetrable).

nyabinary:
What's necessary for this to move forward?

pluiedev:
Mostly to fix some bugs with the code output (see comments above). I'm a bit burnt out from Nix work at the moment but I'll come back to this in due time

pkgs/by-name/in/install-grub-ng/src/builder/appearance.rs
```diff
@@ -0,0 +1,189 @@
...
+			splash_mode,
+			..
+		} = &self.config;
+
+		let Some(splash_image) = splash_image else {
+			return Ok(());
+		};
+
+		let Some(ext) = splash_image.extension() else {
+			bail!("Splash image has no extension - could not decide which module to load!")
```
Shados, line 75: Given you're using `eyre`<sup>1</sup> and returning its `Result` type already, you can avoid these rather verbose `Option` matches and replace them with this arguably more readable construct, converting the `Option` to a `Result` with error context and opening it with the try operator:
```rust
let ext = splash_image.extension().ok_or_eyre("Splash image has no extension - could not decide which module to load!")?;
```

1: Also, what's the motivation for using `eyre` instead of just `anyhow`?

pkgs/by-name/in/install-grub-ng/src/grub.rs
```diff
@@ -0,0 +1,305 @@
...
+			};
+			let Some(device) = fields.next() else {
+				bail!("Device not found in mountinfo entry: {line}")
+			};
+			let Some(mut super_options) = fields.next().map(|s| s.split(',')) else {
+				bail!("Super options not found in mountinfo entry: {line}")
+			};
+
+			// Skip the bind-mount for the Nix store.
+			if mount_point == store_dir && super_options.any(|s| s == "rw") {
```
Shados, line 205: I believe it should be checking `mount_options` rather than `super_options` here, as the latter is *always* going to contain `rw`, for both the bind mount and non-bind mount, at least on the systems I've checked.

That would also explain why `mount_options` appears to be unused in the Perl version: they accidentally used the wrong variable here. Although I guess that never mattered enough to cause a bug in practice, or at least not one that anyone tracked down...

pkgs/by-name/in/install-grub-ng/src/builder/appearance.rs
pluiedev: I preferred `eyre` when I first wrote this, but I've since switched to anyhow — haven't gotten to porting this over yet.

pkgs/by-name/in/install-grub-ng/src/grub.rs
pluiedev: Ah, that makes *so* much sense actually. I was confused by this as well xD

pkgs/by-name/in/install-grub-ng/src/grub.rs
```diff
@@ -0,0 +1,305 @@
...
+impl Fs {
+	fn new(dir: &Path, store_dir: &Path) -> Result<Self> {
+		let mut best = Self::default();
+		let mount_info = BufReader::new(fs::File::open("/proc/self/mountinfo")?);
+
+		for line in mount_info.lines() {
+			let line = line?;
+
+			let mut fields = line.split(' ');
+			let Some(mount_point) = fields.nth(4).map(Path::new) else {
```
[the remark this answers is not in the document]
pluiedev: Do they always use octal escapes and never other escape formats (e.g. hexadecimal escapes)?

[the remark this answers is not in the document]
Shados: Took a look at the kernel code generating the `mountinfo` pseudo-files, and [yep](https://www.kernel.org/doc/html/latest/filesystems/seq_file.html#formatted-output). It is also only those three fields which get this escaping/mangling, too, looks like.

osnyx (changes requested):

pkgs/by-name/in/install-grub-ng/src/builder/appearance.rs
xddxdd, line 104: `ext` here should have the same `jpg => jpeg` mapping as line 82-85, otherwise incorrect module name might be used.

pkgs/by-name/in/install-grub-ng/src/builder/entries.rs
xddxdd, line 129: Should this be using the `DISTRO_NAME` variable instead of compile time env var?

xddxdd, line 421: Should use the `name` variable instead of literal `name`

pkgs/by-name/in/install-grub-ng/src/builder/install.rs
xddxdd, line 60: This section does not handle `NIXOS_INSTALL_BOOTLOADER` correctly which forces GRUB reinstall. The `NIXOS_INSTALL_GRUB -> NIXOS_INSTALL_BOOTLOADER` conversion should happen before the `grub_state.update` check, and installation should happen if either `grub_state.update` is true or `NIXOS_INSTALL_BOOTLOADER` is set.

pkgs/by-name/in/install-grub-ng/src/config.rs
xddxdd, line 15: These entries may contain escaped characters (e.g. when `boot.loader.grub.memtest86.enable = true`) and breaks `serde` zero copy JSON parsing. Use regular `String` to handle escaped chars, or just use `String` for everything since the config JSON is small and memory efficiency isn't a huge concern.

xddxdd, line 25: Same escaped character concerns.

xddxdd (changes requested):

pkgs/by-name/in/install-grub-ng/src/builder/entries.rs
xddxdd, line 383: This makes `path` a relative path in `/nix/store`, which causes the copy command in line 412 to use `[cwd]/[relative path]` as source path. Since cwd is very unlikely to be `/nix/store` the copy will fail.

xddxdd (changes requested):

PR #316878, opened 2024-06-03
syslinux: fix submodule clone

SuperSandro2000:
## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
- For non-Linux: Is sandboxing enabled in `nix.conf`? (See [Nix manual](https://nixos.org/manual/nix/stable/command-ref/conf-file.html))
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:
  - [NixOS test(s)](https://nixos.org/manual/nixos/unstable/index.html#sec-nixos-tests) (look inside [nixos/tests](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests))
  - and/or [package tests](https://github.com/NixOS/nixpkgs/blob/master/pkgs/README.md#package-tests)
  - or, for functions and "core" functionality, tests in [lib/tests](https://github.com/NixOS/nixpkgs/blob/master/lib/tests) or [pkgs/test](https://github.com/NixOS/nixpkgs/blob/master/pkgs/test)
  - made sure NixOS tests are [linked](https://nixos.org/manual/nixpkgs/unstable/#ssec-nixos-tests-linking) to the relevant packages
- [ ] Tested compilation of all packages that depend on this change using `nix-shell -p nixpkgs-review --run "nixpkgs-review rev HEAD"`. Note: all changes have to be committed, also see [nixpkgs-review usage](https://github.com/Mic92/nixpkgs-review#usage)
- [ ] Tested basic functionality of all binary files (usually in `./result/bin/`)
- [24.11 Release Notes](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2411.section.md) (or backporting [23.11](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2311.section.md) and [24.05](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2405.section.md) Release notes)
  - [ ] (Package updates) Added a release notes entry if the change is major or breaking
  - [ ] (Module updates) Added a release notes entry if the change is significant
  - [ ] (Module addition) Added a release notes entry if adding a new NixOS module
- [x] Fits [CONTRIBUTING.md](https://github.com/NixOS/nixpkgs/blob/master/CONTRIBUTING.md).

---

[...]

PR #317030, opened 2024-06-03
syslinux: Use Nixpkgs' gnu-efi

samueldr:
## Description of changes

#316878 aimed to change how `gnu-efi` was fetched. My understanding (after futzing about) is that sourceforge's git endpoint now doesn't provide direct commit access.

Instead of awkwardly fetching an older `gnu-efi`, prefer using Nixpkgs' `gnu-efi`.

Note that *within Nixpkgs* only the "legacy bios" syslinux is used. We could also drop the EFI build instead.

Closes #316878.

## Things done

```
 $ nix-build -A syslinux -A pkgsi686Linux.syslinux
```

Ran the `.efi` output, from before the change, and after. Same result.

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/os-specific/linux/syslinux/default.nix
+++ b/pkgs/os-specific/linux/syslinux/default.nix
@@ -68,12 +68,12 @@ stdenv.mkDerivation {
   ];
 
   postPatch = ''
-    substituteInPlace Makefile --replace /bin/pwd $(type -P pwd)
-    substituteInPlace utils/ppmtolss16 --replace /usr/bin/perl $(type -P perl)
+    substituteInPlace Makefile --replace-fail /bin/pwd $(type -P pwd)
+    substituteInPlace utils/ppmtolss16 --replace-fail /usr/bin/perl $(type -P perl)
 
     # fix tests
     substituteInPlace tests/unittest/include/unittest/unittest.h \
-      --replace /usr/include/ ""
+      --replace-fail /usr/include/ ""
 
     # Hack to get `gcc -m32' to work without having 32-bit Glibc headers.
     mkdir gnu-efi/inc/ia32/gnu
@@ -126,7 +126,7 @@ stdenv.mkDerivation {
   '';
 
   meta = with lib; {
-    homepage = "http://www.syslinux.org/";
+    homepage = "https://www.syslinux.org/";
     description = "A lightweight bootloader";
     license = licenses.gpl2Plus;
     maintainers = [ maintainers.samueldr ];
--- a/pkgs/os-specific/linux/syslinux/default.nix
+++ b/pkgs/os-specific/linux/syslinux/default.nix
@@ -1,7 +1,9 @@
 { lib
 , stdenv
-, fetchgit
+, fetchFromRepoOrCz
+, gnu-efi
 , fetchurl
+, fetchpatch
 , libuuid
 , makeWrapper
 , mtools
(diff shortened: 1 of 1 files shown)


pkgs/os-specific/linux/syslinux/default.nix
```diff
@@ -111,8 +115,22 @@ stdenv.mkDerivation {
     "MANDIR=$(out)/share/man"
     "PERL=perl"
     "HEXDATE=0x00000000"
+    # Works around confusing (unrelated) error messages when upx is not made availab.e
```
raboof, line 118:
```suggestion
    # Works around confusing (unrelated) error messages when upx is not made available
```

raboof (commented):
Looks great to me, I can't really judge how risky it is to use a different version of gnu-efi than is referenced by the syslinux submodule...

samueldr:
> Looks great to me, I can't really judge how risky it is to use a different version of gnu-efi than is referenced by the syslinux submodule...

The main risk would be that it stops building with it due to backwards-incompatible changes.

Technically the added patch works around a backwards-incompatible change, but *barely* so. (A feature was split off out into its own header.)

If it causes issues down the line, I believe it should be fine to disable building the EFI outputs entirely. Anyways they are not used within Nixpkgs nor NixOS. To do so would be relatively simple, removing the references to `gnu-efi`, and then removing the EFI targets.

I actually hesitated and almost went with just removing the EFI outputs.

raboof (approved):

K900:
Somehow, and I genuinely have no idea how, this breaks the ISO boot: https://cache.nixos.org/log/z24ffw9730kzaaljzy18g67jjnn2ix9h-vm-test-run-boot-bios-cdrom.drv

raboof:
oof, thanks for the quick revert :/

raboof:
created #317158

pkgs/os-specific/linux/syslinux/default.nix
```diff
@@ -16,11 +18,10 @@ stdenv.mkDerivation {
 
   # This is syslinux-6.04-pre3^1; syslinux-6.04-pre3 fails to run.
   # Same issue here https://www.syslinux.org/archives/2019-February/026330.html
-  src = fetchgit {
-    url = "https://repo.or.cz/syslinux";
-    rev = "b40487005223a78c3bb4c300ef6c436b3f6ec1f7";
-    sha256 = "sha256-GqvRTr9mA2yRD0G0CF11x1X0jCgqV4Mh+tvE0/0yjqk=";
-    fetchSubmodules = true;
+  src = fetchFromRepoOrCz {
```
SuperSandro2000, line 21: Didn't know that existed. fetchgit defaults to fetching submodules, I assume this doesnt

pkgs/os-specific/linux/syslinux/default.nix
```diff
@@ -65,19 +66,21 @@ stdenv.mkDerivation {
       "0018-prevent-pow-optimization.patch"
       "26f0e7b2"
       "sha256-dVzXBi/oSV9vYgU85mRFHBKuZdup+1x1BipJX74ED7E=")
+    (fetchpatch {
+      # Fixes build with "modern" gnu-efi
+      # https://github.com/samueldr/syslinux/commit/68defee52f4eba82eefaeea17f21c7498448dd6b
+      url = "https://github.com/samueldr/syslinux/commit/68defee52f4eba82eefaeea17f21c7498448dd6b.patch";
+      hash = "sha256-5xIqM8Gq8D86HL1i7fBTHRB/ZR2XtCz5VsT4bI5vneo=";
+    })
```
SuperSandro2000, line 74: tbh we would have been probably better off to just vendor that tiny patch to not rely on your repo 😅

pkgs/os-specific/linux/syslinux/default.nix
samueldr: It also uses a zip instead of git.

PR #317033, opened 2024-06-03
rPackages.covidsymptom: fix build

Kupac:
As of May 2024, there will be [no more data updates](https://github.com/hugofitipaldi/covidsymptom/?tab=readme-ov-file#update-data).

The data included in the package are final, so I removed the update script, which was breaking the nix build.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/development/r-modules/default.nix
+++ b/pkgs/development/r-modules/default.nix
@@ -1411,6 +1411,10 @@ let
       ];
     });
 
+    covidsymptom = old.covidsymptom.overrideAttrs (attrs: {
+      preConfigure = "rm R/covidsymptomdata.R";
+    });
+
     cubature = old.cubature.overrideAttrs (attrs: {
       enableParallelBuilding = false;
     });


Kupac:
I need to double check if the included rda files are identical to the latest csv sources. Should be though

b-rodrigues:
I was also trying to fix it :smile: 

The thing I don't get though: why was the package trying to download data during installation? How could the script was getting called?

Kupac:
They are identical, I just checked.

Kupac:
`covidsymptomdata.R` is a script describing how to generate the files in `data/`. In fact, when you run it, it will recreate the identical rda files. The byte-compiling step would compile all the `*.R` files in the `R/` folder, including this one. I fails, because it assigns the result of a function that's trying to read url-s, so in order to byte compile the variable, it has to download the file, which it can't.

It's good practice to include the script that generates your data, but I believe it shouldn't be in `R/`. I remember something like inst/data or something similar. Anyway, I just remove that script, and it's good, I won't miss it.

b-rodrigues:
> `covidsymptomdata.R` is a script describing how to generate the files in `data/`. In fact, when you run it, it will recreate the identical rda files. The byte-compiling step would compile all the `*.R` files in the `R/` folder, including this one. 

Ah yes I see now why this is happening !


> It's good practice to include the script that generates your data, but I believe it shouldn't be in `R/`. I remember something like inst/data or something similar. 

I believe it should be going into `data-raw`.

>Anyway, I just remove that script, and it's good, I won't miss it.

Great!

b-rodrigues (approved):
working as intended
LGTM

pkgs/development/r-modules/default.nix
```diff
@@ -1402,6 +1402,10 @@ let
       ];
     });
 
+    covidsymptom = old.covidsymptom.overrideAttrs (attrs: {
+      patches = [ ./patches/covidsymptom.patch ];
```
jbedo, line 1406:
```suggestion
      preConfigure = "rm R/covidsymptomdata.R";
```
Would this work? It's smaller than bundling a patch.

Kupac: :facepalm:

Kupac:
OK, done. Works well for me

PR #317153, opened 2024-06-04
Revert "syslinux: Use Nixpkgs' gnu-efi"

K900:
Reverts NixOS/nixpkgs#317030

Somehow, this breaks ISO boot, and Samuel is probably asleep by now, so let's just revert and fix later.


alyssais:
Please include reasons for reverts in the commit message!

PR #233301, opened 2023-05-21
Emacs refactor

AndersonTorres:
###### Description of changes

Cosmetical and structural changes to the code.

###### Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
- [ ] For non-Linux: Is `sandbox = true` set in `nix.conf`? (See [Nix manual](https://nixos.org/manual/nix/stable/command-ref/conf-file.html))
- [ ] Tested, as applicable:
  - [NixOS test(s)](https://nixos.org/manual/nixos/unstable/index.html#sec-nixos-tests) (look inside [nixos/tests](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests))
  - and/or [package tests](https://nixos.org/manual/nixpkgs/unstable/#sec-package-tests)
  - or, for functions and "core" functionality, tests in [lib/tests](https://github.com/NixOS/nixpkgs/blob/master/lib/tests) or [pkgs/test](https://github.com/NixOS/nixpkgs/blob/master/pkgs/test)
  - made sure NixOS tests are [linked](https://nixos.org/manual/nixpkgs/unstable/#ssec-nixos-tests-linking) to the relevant packages
- [ ] Tested compilation of all packages that depend on this change using `nix-shell -p nixpkgs-review --run "nixpkgs-review rev HEAD"`. Note: all changes have to be committed, also see [nixpkgs-review usage](https://github.com/Mic92/nixpkgs-review#usage)
- [ ] Tested basic functionality of all binary files (usually in `./result/bin/`)
- [23.05 Release Notes (or backporting 22.11 Release notes)](https://github.com/NixOS/nixpkgs/blob/master/CONTRIBUTING.md#generating-2305-release-notes)
  - [ ] (Package updates) Added a release notes entry if the change is major or breaking
  - [ ] (Module updates) Added a release notes entry if the change is significant
  - [ ] (Module addition) Added a release notes entry if adding a new NixOS module
- [ ] Fits [CONTRIBUTING.md](https://github.com/NixOS/nixpkgs/blob/master/CONTRIBUTING.md).

PR #317155, opened 2024-06-04
csound-plugins: init at 1.0.2

rvl:
## Description of changes

Some features which were previously built in to Csound have been moved to an [external plugins repository](https://github.com/csound/plugins).

Examples of removed functionality includes the Jacko opcodes, FLTK, and the virtual MIDI keyboard.

This patch allows the lost opcodes to be used again.

## Things done

- [x] Built on platform(s): `x86_64-linux`, `x86_64-linux.pkgsCross.aarch64-multiplatform`, `x86_64-linux.pkgsCross.raspberryPi`
- [x] Tested: has a package test to check for plugin opcodes
- [x] Tested basic functionality of all binary files

---

### DIFF
--- a/pkgs/applications/audio/csound/default.nix
+++ b/pkgs/applications/audio/csound/default.nix
@@ -25,7 +25,7 @@ stdenv.mkDerivation rec {
     sha256 = "sha256-O7s92N54+zIl07eIdK/puoSve/qJ3O01fTh0TP+VdZA=";
   };
 
-  cmakeFlags = [ "-DBUILD_CSOUND_AC=0" ] # fails to find Score.hpp
+  cmakeFlags = lib.optional stdenv.isAarch32 (lib.cmakeFeature "CMAKE_C_FLAGS" "-DPFFFT_SIMD_DISABLE")
     ++ lib.optional stdenv.isDarwin "-DCS_FRAMEWORK_DEST=${placeholder "out"}/lib"
     # Ignore gettext in CMAKE_PREFIX_PATH on cross to prevent find_program picking up the wrong gettext
     ++ lib.optional (stdenv.hostPlatform != stdenv.buildPlatform) "-DCMAKE_IGNORE_PATH=${lib.getBin gettext}/bin";
--- a/pkgs/applications/audio/csound/default.nix
+++ b/pkgs/applications/audio/csound/default.nix
@@ -9,12 +9,7 @@
 , libjack2 ? null
 , liblo ? null
 , ladspa-sdk ? null
-, fluidsynth ? null
-# , gmm ? null  # opcodes don't build with gmm 5.1
-, eigen ? null
 , curl ? null
-, tcltk ? null
-, fltk ? null
 }:
 
 stdenv.mkDerivation rec {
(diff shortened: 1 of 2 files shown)


pkgs/applications/audio/csound/plugins.nix
```diff
@@ -0,0 +1,78 @@
+{ lib, stdenv, fetchFromGitHub, cmake, gettext, runCommand
+, csound
+, libjack2 ? null
```
AndersonTorres, line 3: Never use null as default. It can have nasty consequences.

If you need to control what to include or exclude, use guard flags, like

```suggestion
, enableLibjack2 ? true
, libjack2
```

pkgs/top-level/all-packages.nix
```diff
@@ -30116,6 +30116,10 @@ with pkgs;
     inherit (pkgs.darwin.apple_sdk.frameworks) Accelerate AudioUnit CoreAudio CoreMIDI;
   };
 
+  csound-plugins = callPackage ../applications/audio/csound/plugins.nix {
```
AndersonTorres, line 30119: Given that by-name is mandatory for new packages, please migrate it.

Hint: use passthru.

pkgs/applications/audio/csound/default.nix
```diff
@@ -30,20 +25,18 @@ stdenv.mkDerivation rec {
     sha256 = "sha256-O7s92N54+zIl07eIdK/puoSve/qJ3O01fTh0TP+VdZA=";
   };
 
-  cmakeFlags = [ "-DBUILD_CSOUND_AC=0" ] # fails to find Score.hpp
-    ++ lib.optional stdenv.isDarwin "-DCS_FRAMEWORK_DEST=${placeholder "out"}/lib"
+  cmakeFlags = lib.optional stdenv.isDarwin "-DCS_FRAMEWORK_DEST=${placeholder "out"}/lib"
     # Ignore gettext in CMAKE_PREFIX_PATH on cross to prevent find_program picking up the wrong gettext
     ++ lib.optional (stdenv.hostPlatform != stdenv.buildPlatform) "-DCMAKE_IGNORE_PATH=${lib.getBin gettext}/bin";
```
AndersonTorres, line 30: Can you use `canExecute` instead?

AndersonTorres (changes requested):

pkgs/applications/audio/csound/plugins.nix
rvl: What are the nasty consequences?

pkgs/top-level/all-packages.nix
rvl: OK, will do.

pkgs/applications/audio/csound/default.nix
rvl: OK, will have a look, thanks.

pkgs/applications/audio/csound/plugins.nix
rvl: I have added `enableX` flags for each plugin.

pkgs/top-level/all-packages.nix
rvl: Done.

pkgs/applications/audio/csound/default.nix
rvl: `!(stdenv.buildPlatform.canExecute stdenv.hostPlatform)` has a slightly different meaning to `stdenv.hostPlatform != stdenv.buildPlatform`. I don't want to change this and break something.

@lilyinstarlight added it in #227135.

I have tried building `.#legacyPackages.x86_64-linux.pkgsCross.aarch64-multiplatform.{csound,csound-plugins}` to test it. Both can build with that cmake flag completely removed.

rvl:
@AndersonTorres, thanks for the review - I have addressed your comments. Can you please take another look?

AndersonTorres:
Well, the csound expression has too many `pkg ? null`. This is a dangerous anti-pattern, and I would suggest to get rid of it.

bue it is better to do this in another PR.

pkgs/applications/audio/csound/plugins.nix
AndersonTorres: This is hard to explain. A concrete example is better:

https://github.com/NixOS/nixpkgs/pull/233301#issuecomment-1567661913

PR #317185, opened 2024-06-04
platformsh: add updateScript

spk:
## Description of changes

Handle auto update for platformsh, ref https://github.com/NixOS/nixpkgs/pull/309681#issuecomment-2146399565

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/by-name/pl/platformsh/update.sh
+++ b/pkgs/by-name/pl/platformsh/update.sh
@@ -1,0 +1,29 @@
+#!/usr/bin/env nix-shell
+#!nix-shell -i bash -p curl jq
+#shellcheck shell=bash
+
+set -eu -o pipefail
+
+version=$(curl -s ${GITHUB_TOKEN:+-u ":$GITHUB_TOKEN"} \
+    https://api.github.com/repos/platformsh/cli/releases/latest | jq -e -r ".tag_name")
+
+linux_arm64_url=https://github.com/platformsh/cli/releases/download/$version/platform_${version}_linux_arm64.tar.gz
+linux_amd64_url=https://github.com/platformsh/cli/releases/download/$version/platform_${version}_linux_amd64.tar.gz
+darwin_all_url=https://github.com/platformsh/cli/releases/download/$version/platform_${version}_darwin_all.tar.gz
+linux_arm64_hash=$(nix hash to-sri --type sha256 $(nix-prefetch-url --type sha256 "$linux_arm64_url"))
+linux_amd64_hash=$(nix hash to-sri --type sha256 $(nix-prefetch-url --type sha256 "$linux_amd64_url"))
+darwin_all_hash=$(nix hash to-sri --type sha256 $(nix-prefetch-url --type sha256 "$darwin_all_url"))
+jq -n \
+    --arg version "$version" \
+    --arg darwin_all_hash "$darwin_all_hash" \
+    --arg darwin_all_url "$darwin_all_url" \
+    --arg linux_amd64_hash "$linux_amd64_hash" \
+    --arg linux_amd64_url "$linux_amd64_url" \
+    --arg linux_arm64_hash "$linux_arm64_hash" \
+    --arg linux_arm64_url "$linux_arm64_url" \
+    '{ "version": $version,
+    "darwin-amd64": { "hash": $darwin_all_hash, "url": $darwin_all_url },
+    "darwin-arm64": { "hash": $darwin_all_hash, "url": $darwin_all_url },
+    "linux-amd64": { "hash": $linux_amd64_hash, "url": $linux_amd64_url },
+    "linux-arm64": { "hash": $linux_arm64_hash, "url": $linux_arm64_url }
+}' > pkgs/by-name/pl/platformsh/versions.json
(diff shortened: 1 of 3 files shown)


pkgs/by-name/pl/platformsh/update.sh
```diff
@@ -0,0 +1,16 @@
...
+version=$(curl -s ${GITHUB_TOKEN:+-u ":$GITHUB_TOKEN"} \
+    curl -s https://api.github.com/repos/platformsh/cli/releases/latest | jq -e -r ".tag_name")
+old_version=$(nix-instantiate --eval -A platformsh.version | jq -e -r)
+
+if [[ $version == "$old_version" ]]; then
+    echo "New version same as old version, nothing to do." >&2
+    exit 0
+fi
+
+update-source-version platformsh "$version"
```
infinisil, line 16: This can be tested by invalidating the old hashes, changing the version and running

```
nix-shell maintainers/scripts/update.nix --argstr package platformsh
```

(or I guess the script directly)

It looks like only the `x86_64-linux` hash is being updated, and it doesn't look like update-source-version can easily be adapted for others (there's `--system`, but it fails if you call it multiple times)

My recommendation is to make this script generate a JSON file containing something like
```json
{
  "version": "5.0.13",
  "hashes": {
     "linux_arm64": "...",
     "linux_amd64": "...",
     "darwin_all": "..."
  }
}
```

And then change the Nix code to read the version and hashes from the file instead. This way there's no need to rewrite Nix source code.

You can get the hashes using

```
linux_amd64_hash=$(nix-prefetch-url https://github.com/platformsh/cli/releases/download/5.0.14/platform_$version_linux_amd64.tar.gz)
```

pkgs/by-name/pl/platformsh/update.sh
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env nix-shell
+#!nix-shell -i bash -p curl jq common-updater-scripts
+#shellcheck shell=bash
+
+set -eu -o pipefail
+
+version=$(curl -s ${GITHUB_TOKEN:+-u ":$GITHUB_TOKEN"} \
+    curl -s https://api.github.com/repos/platformsh/cli/releases/latest | jq -e -r ".tag_name")
```
infinisil, line 8: This looks like a mistake (two curl's) :)

pkgs/by-name/pl/platformsh/update.sh
```diff
@@ -0,0 +1,16 @@
...
+old_version=$(nix-instantiate --eval -A platformsh.version | jq -e -r)
+
+if [[ $version == "$old_version" ]]; then
+    echo "New version same as old version, nothing to do." >&2
+    exit 0
+fi
```
infinisil, line 14: With the JSON approach mentioned above, this shouldn't be necessary anymore.

infinisil (dismissed):

pkgs/by-name/pl/platformsh/update.sh
spk: Thanks look like my copy paste skills should improve https://github.com/NixOS/nixpkgs/pull/319781

pkgs/by-name/pl/platformsh/update.sh
spk: Got it thanks for the explanations, I though the update script was run on all architectures but look like I was wrong

pkgs/by-name/pl/platformsh/update.sh
spk: Yes removed

pkgs/by-name/pl/platformsh/update.sh
jtojnar: > (there's `--system`, but it fails if you call it multiple times)

`--system` does not work as one would expect – typically, this is implemented by moving the sources to `passthru.sources` attribute and then looping `update-source-version` over the items ([example](https://github.com/NixOS/nixpkgs/blob/3d37622cffe5ee2f30a624087830374af8fe33f8/pkgs/applications/networking/browsers/palemoon/bin.nix#L186-L188)). Running it multiple times is now supported: https://github.com/NixOS/nixpkgs/pull/318589

shyim (approved):

pkgs/by-name/pl/platformsh/update.sh
spk: is it the preferred way to do those auto-updates? can we go with the current solution since I've already put some effort implementing it?

spk: I've tried to setup the update-source-version but got an error with darwin `update-source-version: error: Couldn't locate old source hash 'sha256-G5/T6ZPcvC7dvw82p2CEMEOb7GCTyCAB8xJ2lxV2UXk=' (or it appeared more than once) in '/home/spk/dev/nixpkgs/pkgs/by-name/pl/platformsh/package.nix'!` I think its because its the same package for both x86-64 and aarch64, how can I resolve this?

PR #86948, opened 2020-05-05
haskell: tune generic-builder for more parallel build performance

peti:
Raise the maximum number of cores used for parallel building from 4 to 16. Increase the size of the allocation area for GHC's garbage collector from 1 MiB to 64 MiB.

See https://www.twitch.tv/videos/611899011 for the motivation for this change.

PR #317251, opened 2024-06-04
haskellPackages: build with RTS -A64M options

wolfgangwalther:
While working on #317181 I realized that the `+RTS -A64M -RTS` options in `parallelBuildingFlags` were never passed to GHC, because the arguments are not passed quoted. This means those arguments are actually passed to `Setup.hs`. This can be verified by removing `-rtsopts` from `setupCompileFlags` - after this change the configure command will throw an error, because it doesn't understand the `+RTS ...` argument anymore. Once you pass it properly, the error is gone again and GHC receives the arguments.

Passing the RTS flags to `Setup.hs`  (i.e. cabal) - makes little to zero sense. The same is true for the entire `parallelBuildingFlags` being passed to `setupCompileFlags`: If there was any custom `Setup.hs` script in any package anywhere that benefits from heavy parallelization like this... there would be something wrong with that script! ;)

Those flags were introduced in #86948. The related twitch live stream uses the build of git-annex as a measurement. I get the following numbers when building `git-annex` with `doCheck = false;`:

- current master: 1:40 wall clock / 340s user
- without -A64M: 1:40 wall clock / 340s user
- with this fix: 1:13 wall clock / 280s user

This 27% improvement is essentially the result from that live stream mentioned above. Let's use that, shall we?

## Things done

- Built on platform(s)
  - [x] x86_64-linux: `haskell.lib.overrideCabal haskellPackages.git-annex { doCheck = false; }`
[...]

---

### DIFF
--- a/pkgs/development/haskell-modules/generic-builder.nix
+++ b/pkgs/development/haskell-modules/generic-builder.nix
@@ -235,7 +235,7 @@ let
   ] ++ optional (allPkgconfigDepends != [])
     "--with-pkg-config=${pkg-config.targetPrefix}pkg-config";
 
-  parallelBuildingFlags = "-j$NIX_BUILD_CORES" + optionalString stdenv.isLinux " +RTS -A64M -RTS";
+  makeGhcOptions = opts: lib.concatStringsSep " " (map (opt: "--ghc-option=${opt}") opts);
 
   crossCabalFlagsString =
     lib.optionalString isCross (" " + lib.concatStringsSep " " crossCabalFlags);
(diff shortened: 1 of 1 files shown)


pkgs/development/haskell-modules/generic-builder.nix
```diff
@@ -256,8 +256,9 @@ let
     "--package-db=$packageConfDir"
     (optionalString (enableSharedExecutables && stdenv.isLinux) "--ghc-option=-optl=-Wl,-rpath=$out/${ghcLibdir}/${pname}-${version}")
     (optionalString (enableSharedExecutables && stdenv.isDarwin) "--ghc-option=-optl=-Wl,-headerpad_max_install_names")
-    (optionalString enableParallelBuilding "--ghc-options=${parallelBuildingFlags}")
-    (optionalString useCpphs "--with-cpphs=${cpphs}/bin/cpphs --ghc-options=-cpp --ghc-options=-pgmP${cpphs}/bin/cpphs --ghc-options=-optP--cpp")
+    (optionalString enableParallelBuilding "--ghc-option=-j$NIX_BUILD_CORES")
+    (optionalString (enableParallelBuilding && stdenv.isLinux) (makeGhcOptions [ "+RTS" "-A64M" "-RTS" ]))
+    (optionalString useCpphs ("--with-cpphs=${cpphs}/bin/cpphs" + (makeGhcOptions [ "-cpp"  "-pgmP${cpphs}/bin/cpphs" "-optP--cpp" ])))
```
sternenseemann, line 261: Aren't we missing a space  here?

```suggestion
    (optionalString useCpphs ("--with-cpphs=${cpphs}/bin/cpphs " + (makeGhcOptions [ "-cpp"  "-pgmP${cpphs}/bin/cpphs" "-optP--cpp" ])))
```

sternenseemann (commented):
Otherwise this makes sense to me. `-A64M` is beneficial in multi-threaded situations. As far as I'm aware, `Cabal` is single-threaded for our usecase while ghc (even with our current escaping mess) receives a `-j` flag.

sternenseemann:
My testing with `lens` seems to confirm what you are saying (though I got very noisy data).

sternenseemann (approved):

PR #270221, opened 2023-11-26
nixos/acme: default to lets encrypt production URL instead of null, mention lets encrypt staging URI

SuperSandro2000:
## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
- For non-Linux: Is sandboxing enabled in `nix.conf`? (See [Nix manual](https://nixos.org/manual/nix/stable/command-ref/conf-file.html))
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:
  - [NixOS test(s)](https://nixos.org/manual/nixos/unstable/index.html#sec-nixos-tests) (look inside [nixos/tests](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests))
  - and/or [package tests](https://nixos.org/manual/nixpkgs/unstable/#sec-package-tests)
  - or, for functions and "core" functionality, tests in [lib/tests](https://github.com/NixOS/nixpkgs/blob/master/lib/tests) or [pkgs/test](https://github.com/NixOS/nixpkgs/blob/master/pkgs/test)
  - made sure NixOS tests are [linked](https://nixos.org/manual/nixpkgs/unstable/#ssec-nixos-tests-linking) to the relevant packages
- [ ] Tested compilation of all packages that depend on this change using `nix-shell -p nixpkgs-review --run "nixpkgs-review rev HEAD"`. Note: all changes have to be committed, also see [nixpkgs-review usage](https://github.com/Mic92/nixpkgs-review#usage)
- [ ] Tested basic functionality of all binary files (usually in `./result/bin/`)
- [23.11 Release Notes](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2311.section.md) (or backporting [23.05 Release notes](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2305.section.md))
  - [ ] (Package updates) Added a release notes entry if the change is major or breaking
  - [ ] (Module updates) Added a release notes entry if the change is significant
  - [ ] (Module addition) Added a release notes entry if adding a new NixOS module
- [x] Fits [CONTRIBUTING.md](https://github.com/NixOS/nixpkgs/blob/master/CONTRIBUTING.md).

### Priorities

Add a :+1: [reaction] to [pull requests you find important].

[...]

PR #317257, opened 2024-06-04
nixos/acme: allow setting security.acme.defaults.server = null to keep old accounts directory

### LINKED ISSUE #316608: NixOS 24.05 creates a new letsencrypt account and breaks CAA
After updating to 24.05, it appears that the ACME client (with the default infrastructure: security.acme/services.nginx.xxx.enableACME etc.) registers a new account for obtaining letsencrypt certificates. If you had CAA with account binding enabled on your domains, the certificate renewals will now fail.

If you are affected by the problem, here is how you can figure out the new account URI in order to update your CAA entries:
1. ls -l /var/lib/acme/.lego/accounts/ and take the most recent directory
2. cat /var/lib/acme/.lego/accounts/xxxxx/acme-v02.api.letsencrypt.org/email@example.com/account.json

This should probably be added to the release notes.

arianvp:
## Description of changes

The accounts directory is based on the hash of the settings.

https://github.com/NixOS/nixpkgs/pull/270221 changed the  default of security.acme.defaults.server from null to the default letsencrypt URL however as an unwanted side effect this means the accounts directory changes and the ACME module will create a new a new account.

This can cause issues with people using CAA records that pin the account ID or people who have datacenter-scale NixOS deployments

We allow setting this option to null again for people who want to keep the old account and migrate at their own leisure.

Fixes https://github.com/NixOS/nixpkgs/issues/316608

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/nixos/doc/manual/release-notes/rl-2405.section.md
+++ b/nixos/doc/manual/release-notes/rl-2405.section.md
@@ -429,6 +429,16 @@ Use `services.pipewire.extraConfig` or `services.pipewire.configPackages` for Pi
 
 - `screen`'s module has been cleaned, and will now require you to set `programs.screen.enable` in order to populate `screenrc` and add the program to the environment.
 
+- `security.acme.defaults.server` now has a default value instead of `null`.
+  This effectively uses the same server, the Let's Encrypt production server,
+  but makes the default explicit, instead of relying on the Lego default.
+
+  A side effect of this is that the directory in which account data is stored
+  changes and the ACME module will request a new account and new certificates
+  for all domains.  This may cause issues if you pin an `acccounturl` in a CAA
+  DNS record.  To avoid this, you
+  may set `security.acme.defaults.server = null` to keep the old hashes.
+
 - `security.pam.sshAgentAuth.enable` now requires `services.openssh.authorizedKeysFiles` to be non-empty,
   which is the case when `services.openssh.enable` is true. Previously, `pam_ssh_agent_auth` silently failed to work.
 
--- a/nixos/modules/security/acme/default.nix
+++ b/nixos/modules/security/acme/default.nix
@@ -545,7 +545,7 @@ let
       };
 
       server = mkOption {
-        type = types.str;
+        type = types.nullOr types.str;
         inherit (defaultAndText "server" "https://acme-v02.api.letsencrypt.org/directory") default defaultText;
         example = "https://acme-staging-v02.api.letsencrypt.org/directory";
         description = ''


arianvp:
@GrahamcOfBorg test acme

nixos/doc/manual/release-notes/rl-2405.section.md
```diff
@@ -429,7 +429,17 @@ Use `services.pipewire.extraConfig` or `services.pipewire.configPackages` for Pi
 
 - `screen`'s module has been cleaned, and will now require you to set `programs.screen.enable` in order to populate `screenrc` and add the program to the environment.
 
-- `security.pam.sshAgentAuth.enable` now requires `services.openssh.authorizedKeysFiles` to be non-empty,
```
arianvp, line 432: hmm this looks weird. let me fix

nixos/modules/security/acme/default.nix
```diff
@@ -545,7 +545,7 @@ let
       };
 
       server = mkOption {
-        type = types.str;
+        type = types.nullOr types.str;
```
SuperSandro2000, line 548: I think we should add a warning to nudge people to set this explicitly to safeguard us against future changes in lego.

SuperSandro2000:
otherwise looks good to me.

Not sure if we should add a diverged test

nixos/modules/security/acme/default.nix
m1cr0man: By the same logic, I doubt that lego would ever make such a change due to the implications for their end users. I also think that users who now set the value to null understand that they adopt the default from Lego and won't need a warning to remind them.

m1cr0man (approved):
Looks good. Change log entry makes sense and is clear.

nixos/modules/security/acme/default.nix
osnyx: When we're talking about catching future changes, we should probably also add a test case for `null` to https://github.com/NixOS/nixpkgs/pull/286999 to catch if someone decides to drop this special case *again*.

flokli (approved):

arianvp:
The backport failed due to a conflict due to https://github.com/NixOS/nixpkgs/pull/316884 which seems to not have been backported. (IDK what the normal etiquette is with backporting typos in release notes)

My git foo is not good enough to fix this. Help would be appreciated.

Issue #312345, opened 2024-05-17
Lint on broken recursion structure


lf-:
## Issue description

It's possible to write something like:

```nix
{somePackageName, stdenv}:
stdenv.mkDerivation {
  pname = "some-package-name";
  # ...
  passthru.someOtherThing = pkgs.callPackage ./something.nix { inherit somePackageName; };
}
```

where something in `passthru` doesn't follow the containing derivation's `overrideAttrs` properly due to broken recursion structure. The correct code here would be:

```nix
{somePackageName, stdenv}:
stdenv.mkDerivation (finalAttrs: {
  pname = "some-package-name";
  # ...
  passthru.someOtherThing = pkgs.callPackage ./something.nix { somePackageName = finalAttrs.finalPackage; };
})
```

I would like to have an automated lint that rejects code that exhibits this broken recursion structure: `passthru` referring to the parent derivation but the passthru attrs get the parent derivation from other means than `finalAttrs.finalPackage`.

eclairevoyant:
Why? Also, shouldn't we ensure all the builders support the `finalAttrs` pattern in the first place?

lf-:
> Also, shouldn't we ensure all the builders support the `finalAttrs` pattern in the first place?

yeah, there's a lot of very broken builders that have very broken overrides in general. we should at least fix the standard mkDerivation though.

lf-:
not stale! bad bot

philiptaron:
Now that extendMkDerivation has landed, and the various build helpers are all gaining good overrideAttrs and finalAttrs support, I think nixpkgs-vet is now in a fine place to make this sort of assertion.

Issue #257446, opened 2023-09-26
`fetchpatch2` allows lengths of commit hashes to change


NiklasGollenstede:
### Describe the bug

The point of `fetchpatch` and `fetchpatch2` is to prevent changes in API-generated patches that don't affect the patched tree from affecting the hash of the fetched patch.

GitHub dynamically chooses the length of abbreviated commit hashes in a number of places, and (I guess) also in the generated patches.

Some time around August, a commit hash in the following patch changed from `5d7ffe2d91910` to `5d7ffe2d91910a`.

```nix
/* was: */ (pkgs.fetchpatch2 { url = "https://github.com/NixOS/nixpkgs/commit/eeff6c493373d3fff11421b55309fab6a1d4ec7d.diff"; sha256 = "sha256-Sqt1FQG+0j9dJMtZiBX+dK3UW5SH5Gtg8zVFTs0F23s="; })
/* now: */ (pkgs.fetchpatch2 { url = "https://github.com/NixOS/nixpkgs/commit/eeff6c493373d3fff11421b55309fab6a1d4ec7d.diff"; sha256 = "sha256-Z5vENYxMNBmTUuhzug9Kgi7/21J1X6d7BxqovyOv9fg="; })
```
<details><summary>Old patch output</summary>

```diff
diff --git a/pkgs/os-specific/linux/systemd/default.nix b/pkgs/os-specific/linux/systemd/default.nix
index 5d7ffe2d91910..14d5e29c372a88 100644
--- a/pkgs/os-specific/linux/systemd/default.nix
+++ b/pkgs/os-specific/linux/systemd/default.nix
@@ -178,6 +178,13 @@
     # need (AFAICT).
     # See https://github.com/systemd/systemd/pull/20479 for upstream discussion.
     ./0019-core-handle-lookup-paths-being-symlinks.patch
+
+    # fixes reproducability of dbus xml files
+    # Should no longer be necessary with v251.
+    (fetchpatch {
+      url = "https://github.com/systemd/systemd/pull/22174.patch";
+      sha256 = "sha256-RVhxUEUiISgRlIP/AhU+w1VHfDQw2W16cFl2TXXyxno=";
+    })
   ] ++ lib.optional stdenv.hostPlatform.isMusl (
     let
       oe-core = fetchzip {
```
</details>

I still had that file in a local store. The new output is identical, except it uses `5d7ffe2d91910a` as the commit hash.

### Steps To Reproduce

[...]

PR #317275, opened 2024-06-04
libgit2: 1.7.2 -> 1.8.1; python312Packages.pygit2: 1.14.1 -> 1.15.0

eclairevoyant:
## Description of changes

- 1.7.2 -> 1.8.1
  - This also replaces the dependency on the abandoned `http-parser` with `llhttp`.
- fix incorrect recursive structure in tests (tracking issue: #312345)

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/development/libraries/libgit2-glib/default.nix
+++ b/pkgs/development/libraries/libgit2-glib/default.nix
@@ -13,6 +13,7 @@
 , libgit2
 , glib
 , python3
+, fetchpatch
 }:
 
 stdenv.mkDerivation rec {
@@ -25,6 +26,15 @@ stdenv.mkDerivation rec {
     url = "mirror://gnome/sources/${pname}/${lib.versions.majorMinor version}/${pname}-${version}.tar.xz";
     sha256 = "EzHa2oOPTh9ZGyZFnUQSajJd52LcPNJhU6Ma+9/hgZA=";
   };
+
+  patches = [
+    (fetchpatch {
+      name = "support-libgit2-1.8.patch";
+      # https://gitlab.gnome.org/GNOME/libgit2-glib/-/merge_requests/40
+      url = "https://gitlab.gnome.org/GNOME/libgit2-glib/-/commit/a76fdf96c3af9ce9d21a3985c4be8a1aa6eea661.patch";
+      hash = "sha256-ysU8pAixyftensfEC9bE0RUFMPMei0jYT26WKN5uOFE=";
+    })
+  ];
 
   nativeBuildInputs = [
     meson
--- a/pkgs/development/libraries/libgit2/default.nix
+++ b/pkgs/development/libraries/libgit2/default.nix
@@ -8,7 +8,6 @@
 , libssh2
 , openssl
 , pcre
-, http-parser
 , libiconv
 , Security
 , staticBuild ? stdenv.hostPlatform.isStatic
@@ -16,11 +15,14 @@
 , libgit2-glib
 , python3Packages
 , gitstatus
+, llhttp
+, withGssapi ? false
+, krb5
 }:
 
 stdenv.mkDerivation rec {
   pname = "libgit2";
-  version = "1.7.2";
+  version = "1.8.1";
   # also check the following packages for updates: python3Packages.pygit2 and libgit2-glib
 
   outputs = ["lib" "dev" "out"];
@@ -29,12 +31,13 @@ stdenv.mkDerivation rec {
     owner = "libgit2";
     repo = "libgit2";
     rev = "v${version}";
-    hash = "sha256-fVPY/byE2/rxmv/bUykcAbmUFMlF3UZogVuTzjOXJUU=";
+    hash = "sha256-J2rCxTecyLbbDdsyBWn9w7r3pbKRMkI9E7RvRgAqBdY=";
   };
 
   cmakeFlags = [
     "-DUSE_HTTP_PARSER=system"
     "-DUSE_SSH=ON"
+    (lib.cmakeBool "USE_GSSAPI" withGssapi)
     "-DBUILD_SHARED_LIBS=${if staticBuild then "OFF" else "ON"}"
   ] ++ lib.optionals stdenv.hostPlatform.isWindows [
     "-DDLLTOOL=${stdenv.cc.bintools.targetPrefix}dlltool"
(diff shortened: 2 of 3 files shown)


pkgs/development/libraries/libgit2/default.nix
SuperSandro2000, line 1: Please wait for the coordinated effort to format files instead of doing that in individual PRs, to reduce chores and noise in the commit history.

pkgs/development/libraries/libgit2/default.nix
```diff
@@ -34,13 +35,14 @@ stdenv.mkDerivation rec {
...
     rev = "v${version}";
-    hash = "sha256-fVPY/byE2/rxmv/bUykcAbmUFMlF3UZogVuTzjOXJUU=";
+    hash = "sha256-J2rCxTecyLbbDdsyBWn9w7r3pbKRMkI9E7RvRgAqBdY=";
   };
 
   cmakeFlags =
     [
       "-DUSE_HTTP_PARSER=system"
       "-DUSE_SSH=ON"
+      "-DUSE_GSSAPI=ON"
```
SuperSandro2000, line 45: I am not sure if we should turn this on by default. Openssh has it off IIRC

SuperSandro2000 (approved):
LGTM, please test the reverse deps enough to not cause to many breakges

pkgs/development/libraries/libgit2-glib/default.nix
```diff
@@ -26,6 +27,14 @@ stdenv.mkDerivation rec {
     sha256 = "EzHa2oOPTh9ZGyZFnUQSajJd52LcPNJhU6Ma+9/hgZA=";
   };
 
+  patches = [
+    (fetchpatch2 {
+      name = "support-libgit2-1.8.patch";
+      url = "https://github.com/GNOME/libgit2-glib/pull/4/commits/304a8c3fae1784d665e92c1ce2143c66d7ecf47f.patch";
```
jtojnar, line 33: This patch looks insufficient compared to https://gitlab.gnome.org/GNOME/libgit2-glib/-/merge_requests/40

Also it is unmerged so we are not certain the URL will persist when/if the commit garbage collected.

And `fetchpatch2` should not be used for dynamically-generated patches – it produces unstable FODs because short git commit ids in the patch will be lengthened eventually.

eclairevoyant: > This patch looks insufficient

it builds as is

> if the commit garbage collected

gh doesn't gc ever, unless you file a support ticket

> And `fetchpatch2` should not be used for dynamically-generated patches – it produces unstable FODs because short git commit ids in the patch will be lengthened eventually.

Not sure what you're asking for here

eclairevoyant:
> please test the reverse deps enough to not cause to many breakges

unfortunately there are lots of breakages, I'm trying to go through them when I get a chance, hence left the PR in draft for now

pkgs/development/libraries/libgit2-glib/default.nix
jtojnar: > it builds as is

It lacks the casts, which I suspect might be necessary on some configurations (maybe CLang).

> gh doesn't gc ever, unless you file a support ticket

Is that true of unmerged PRs? I have seen some such PRs, whose content has disappeared (though it was a long time ago).

> Not sure what you're asking for here

`fetchpath` should be used instead of `fetchpatch2` when fetching dynamically generated patch. `fetchpatch2` is unfortunately only suitable for transforming static patches

SuperSandro2000: > Is that true of unmerged PRs? I have seen some such PRs, whose content has disappeared (though it was a long time ago).

I think they can if a PR is force pushed and maybe never accessed but that's GitLab.

fetchpatch2 is required for renames. Also I never heard of those issues. Some people are using fetchpatch2 by default to get it into the tree more.

SuperSandro2000:
I meant the key reverse deps listed in passthru 😅

pkgs/development/libraries/libgit2-glib/default.nix
jtojnar: https://github.com/NixOS/nixpkgs/issues/257446

eclairevoyant: I've changed the url to use the actual commit and added `full_index=1` as a param to forcibly use the full hash in the patch itself; this should obviate concerns about the patch being dynamic now.

pkgs/development/libraries/libgit2-glib/default.nix
```diff
@@ -26,6 +27,15 @@ stdenv.mkDerivation rec {
     sha256 = "EzHa2oOPTh9ZGyZFnUQSajJd52LcPNJhU6Ma+9/hgZA=";
   };
 
+  patches = [
+    (fetchpatch2 {
+      name = "support-libgit2-1.8.patch";
+      # https://github.com/GNOME/libgit2-glib/pull/4
+      url = "https://github.com/GNOME/libgit2-glib/commit/304a8c3fae1784d665e92c1ce2143c66d7ecf47f.patch?full_index=1";
```
SuperSandro2000, line 34: https://gitlab.gnome.org/GNOME/libgit2-glib/-/merge_requests/40

eclairevoyant: updated

trofi:
Bisect claims that 602760ce42b83bf9071f52d97adcbebc046f5fe1 `libgit2: fix incorrect recursive structure in tests` broke the eval of `libgit2.tests` in `staging` as:

```
$ nix build --no-link -f. libgit2.tests

error:
       … while calling a functor (an attribute set with a '__functor' attribute)
         at pkgs/development/libraries/libgit2/default.nix:73:40:
           72|
           73|   passthru.tests = lib.mapAttrs (_: v: v.override { libgit2 = finalAttrs.finalPackage; }) {
             |                                        ^
           74|     inherit libgit2-glib;

       … while calling a functor (an attribute set with a '__functor' attribute)
         at lib/customisation.nix:144:43:
          143|       # Re-call the function but with different arguments
          144|       overrideArgs = mirrorArgs (newArgs: makeOverridable f (overrideWith newArgs));
             |                                           ^
          145|       # Change the result of the function call by applying g to it

       error: function 'anonymous lambda' called with unexpected argument 'libgit2'
       at pkgs/applications/version-management/gitstatus/default.nix:1:1:
            1| { callPackage, lib, stdenv, fetchFromGitHub, git, zsh }:
             | ^
            2|
```

I think it says that `gitstatus` does not have an immediate `libgit2` input.

eclairevoyant:
Thanks for catching.
And yikes, how would we even override that? it's coming in via `buildInputs = [ (callPackage ./romkatv_libgit2.nix { }) ];` which in turn has a `libgit2` input...

SuperSandro2000:
I try to take a look at this later today

PR #317281, opened 2024-06-04
postgresql: add readme with eol-policy

wolfgangwalther:
This was discussed and agreed on in https://github.com/NixOS/nixpkgs/pull/310580#discussion_r1597284693.

@Ma27 @mweinelt

Feedback welcome!

---

### DIFF
--- a/nixos/modules/services/databases/postgresql.md
+++ b/nixos/modules/services/databases/postgresql.md
@@ -244,6 +244,27 @@ The upgrade process is:
      $ ./delete_old_cluster.sh
      ```
 
+## Versioning and End-of-Life {#module-services-postgres-versioning}
+
+PostgreSQL's versioning policy is described [here](https://www.postgresql.org/support/versioning/). TLDR:
+
+- Each major version is supported for 5 years.
+- Every three months there will be a new minor release, containing bug and security fixes.
+- For criticial/security fixes there could be more minor releases inbetween. This happens *very* infrequently.
+- After five years, a final minor version is released. This usually happens in early November.
+- After that a version is considered end-of-life (EOL).
+- Around February each year is the first time an EOL-release will not have received regular updates anymore.
+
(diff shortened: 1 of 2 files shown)


pkgs/servers/sql/postgresql/README.md
```diff
@@ -0,0 +1,22 @@
...
+PostgreSQL's versioning policy is described [here](https://www.postgresql.org/support/versioning/). TLDR:
+
+- Each major version is supported for 5 years.
+- Every three months there will be a new minor release, containing bug and security fixes.
+- For criticial/security fixes there could be more minor releases inbetween. This happens *very* infrequently.
+- After five years, a final minor version is released. This usually happens in early November.
+- After that a version is considered end-of-life (EOL).
+- Around February each year is the first time an EOL-release will not have received regular updates anymore.
+
+Technically, we'd not want to have EOL'ed packages in a stable NixOS release, which is to be supported until one month after the previous release. Thus, with NixOS' release schedule in May and November, the oldest PostgreSQL version in nixpkgs would have to be supported until December. It could be argued that a soon-to-be-EOL-ed version should thus be removed in May for the .05 release already.
```
Ma27, line 12: There are two aspects I'd like to see reflected here:

* with the last release coming in November, we technically have coverage until February because vulnerabilities are disclosed for the other versions in February and not earlier. For me this was the primary reason why I agreed IIRC.
* if there are "emergency releases" before and within the lifetime of the old release (in our case 24.05), then we'll mark the package as insecure unless we can _trivially_ patch the issue.

Ma27 (commented):
May I suggest to place this into the nixpkgs manual somewhere (`doc/`)? I'd consider this a little more discoverable than having it in a README in the random directory postgresql happens to be in.

pkgs/servers/sql/postgresql/README.md
Ma27: > This is addressed in the very last paragraph already. I can rephrase that to use "trivially".

Eh yes, sorry. I define "trivially" as "patch applies without conflicts" here and <=5 patches to apply. Everything else seems a little too error-prone to me then. WDYT?

Ma27:
> Putting it in manual will make it less visible for the random package updater, though.

I can only speak for me here, but I usually just do `nix edit -f. postgresql` to edit the expressions (or use `find`) and I can't recall having checked for a README once (also because I'm not aware of READMEs in pkgs-subdirs being a thing).

I think it's where I would've put it and having it in a _manual_ kinda makes sense. But at the end of the day, I think we'll have to link it to anyone who asks anyways.

Ma27 (approved):

PR #317385, opened 2024-06-05
 libeduvpn-common: 1.2.1 -> 2.0.1 and eduvpn-client: 4.2.1 -> 4.3.0 

benneti:
## Description of changes

Packaging of upstream changed and needed some manual tweaks, replaces https://github.com/NixOS/nixpkgs/pull/317353

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---


jwijenbergh:
> I am not sure why the client now needs the postInstall thing, but the folders are not copied to the right place else, maybe you @jwijenbergh have a better idea what is happening there.

Yeah I think my setup.cfg was not completely correct here, does it work without changes with this patch? https://github.com/eduvpn/python-eduvpn-client/pull/595

benneti:
yes this works, and also get's rid of one of the nix specific patches 👍

pkgs/by-name/ed/eduvpn-client/package.nix
```diff
@@ -38,7 +40,12 @@ python3Packages.buildPythonApplication rec {
     setuptools
   ];
 
-  patches = [ ./nix-python-prefix.patch ];
```
jwijenbergh, line 41: I think it's best to keep this patch in

benneti: I can revert this change, however, I was thinking that a more streamlined package with less nix specific patches would make future updates more convenient (and keep nixpkgs a bit slimmer)

jwijenbergh:
> yes this works, and also get's rid of one of the nix specific patches 👍 EDIT: let me know if you tag a 4.3.1 and I'll update the PR accordingly and remove the manual patch

4.3.1 has been tagged, thank you for testing this!

pkgs/by-name/ed/eduvpn-client/package.nix
jwijenbergh: I agree with you, but this ensures the nix package would never clash with any other installs of eduvpn. E.g. through pip. For e.g. 4.4.0 I will think about how I can solve this better

benneti: This may simplify stuff to much, but could you not put the prefix as the first element, such that the relative path to the running installation is prefered?

jwijenbergh: I am thinking to do that but that breaks pip and local installs (git checkouts running `python3 -m eduvpn.ui`). So I would have to patch the pip package and find add a script to run it locally with a patch

jwijenbergh: Just another note I will probably need to use `pkg_resources` module or so, I will try this for the next version.

benneti: fair, do you think it is worth it to readd it for now or can we leave it as is, and simply hope that nothing bad happens until the new versions comes out  (also accounting for that probably only very few people install the same package using nix and pip)?

jwijenbergh: as the intention is to remove it in the future, I would go for adding it. But my opinions on this are not super strong so if you feel like this is also good then we will go for this. Let me build the package and then approve

jwijenbergh (approved):
Works perfectly, thanks again

pkgs/by-name/ed/eduvpn-client/package.nix
benneti: great, then let's leave it like this. I am not an expert but I have a feeling that it is preferred to avoid unnecessary patches.

jwijenbergh:
@benneti i tagged another common version for some translations and some tiny changes

https://github.com/eduvpn/eduvpn-common/releases/tag/2.0.1

benneti:
sorry for the delay, bumped and squashed

jwijenbergh:
Thanks! @h7x4 @benneti

PR #317392, opened 2024-06-05
doc:  add a generic section on exploring package sets

fricklerhandwerk:
## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/doc/languages-frameworks/haskell.section.md
+++ b/doc/languages-frameworks/haskell.section.md
@@ -38,8 +38,7 @@ For packages that are part of [Stackage] (a curated set of known to be
 compatible packages), we use the version prescribed by a Stackage snapshot
 (usually the current LTS one) as the default version. For all other packages we
 use the latest version from [Hackage](https://hackage.org) (the repository of
-basically all open source Haskell packages). See [below](#haskell-available-
-versions) for a few more details on this.
+basically all open source Haskell packages). See [below](#haskell-available-versions) for a few more details on this.
 
 Roughly half of the 16K packages contained in `haskellPackages` don’t actually
 build and are [marked as broken semi-automatically](https://github.com/NixOS/nixpkgs/blob/haskell-updates/pkgs/development/haskell-modules/configuration-hackage2nix/broken.yaml).
--- a/doc/languages-frameworks/haskell.section.md
+++ b/doc/languages-frameworks/haskell.section.md
@@ -21,41 +21,14 @@ Many “normal” user facing packages written in Haskell, like `niv` or `cachix
 are also exposed at the top level, and there is nothing Haskell specific to
 installing and using them.
 
-All of these packages are originally defined in the `haskellPackages` package
-set and are re-exposed with a reduced dependency closure for convenience.
-(see `justStaticExecutables` or `separateBinOutput` below)
+All of these packages are originally defined in the `haskellPackages` package set.
+The same packages are re-exposed with a reduced dependency closure for convenience (see `justStaticExecutables` or `separateBinOutput` below).
+
+:::{.note}
+See [](#chap-language-support) for techniques to explore package sets.
+:::
 
 The `haskellPackages` set includes at least one version of every package from [Hackage](https://hackage.haskell.org/) as well as some manually injected packages.
-Haskell packages can be searched on [search.nixos.org](https://search.nixos.org/packages?query=haskellPackages) or listed on the command line with [`nix-instantiate`](https://nixos.org/manual/nix/stable/command-ref/nix-instantiate).
-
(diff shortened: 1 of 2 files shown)


doc/languages-frameworks/haskell.section.md
```diff
@@ -25,21 +25,12 @@ All of these packages are originally defined in the `haskellPackages` package
...
+The `haskellPackages` set includes at least one version of every package from [Hackage](https://hackage.haskell.org/) as well as some manually injected packages.
+Haskell packages can be searched on [search.nixos.org](https://search.nixos.org/packages?query=haskellPackages) or listed on the command line with [`nix-instantiate`](https://nixos.org/manual/nix/stable/command-ref/nix-instantiate):
 
-```console
-$ nix-env -f '<nixpkgs>' -qaP -A haskellPackages
-haskellPackages.a50                                                         a50-0.5
-haskellPackages.AAI                                                         AAI-0.2.0.1
-haskellPackages.aasam                                                       aasam-0.2.0.0
-haskellPackages.abacate                                                     abacate-0.0.0.0
-haskellPackages.abc-puzzle                                                  abc-puzzle-0.2.1
```
Atemu, line 39: The new method does not show version information which might be critical.

It'd be great to retain the example output in any case though.

Atemu (approved):

sternenseemann (dismissed):
This just seems silly to me. There's nothing wrong with using `nix-env` in this way and the presented alternative is strictly worse?

Atemu:
Listing packages given a (possibly nested) attrset while ignoring any eval errors is one of the only sensible uses of nix-env. What we want to get away from is using nix-env for imperative package management because it really sucks quite a lot at that.

fricklerhandwerk:
@Atemu would I be missing something crucial if I claim this can be done in the 4-liner I added as an example?

Atemu:
Yes, it'd fall on its nose on the first attribute that doesn't eval or doesn't have a version. haskellPackages is quite well behaved in this regard but go I dare you to this on `pkgs`... The only way to get this sort of information reliably on `pkgs` is `nix-env`; I've been there.

fricklerhandwerk:
Incredible...

https://github.com/NixOS/nix/blob/master/src/nix-env/nix-env.cc#L195
https://github.com/NixOS/nix/blob/master/src/libexpr/get-drvs.cc#L311-L330

How about we all forget that code ever existed, merge this PR, and focus on something else?

Atemu:
I don't have a strong opinion on this and I don't maintain this part of Nixpkgs. I also don't have anything more to contribute to this discussion, so I'm going to unsubscribe here.

fricklerhandwerk:
@maralorn @ncfavier opinions?

maralorn:
I have no strong opinion on this and I frankly don’t feel like arbitrating this on my free time. So I will just leave some thoughts:

I empathize with the desire to get rid of references to nix-env. I mean in general even mentioning that packages won’t get listed by `nix-env -qa` seems quite irrelevant because that workflow seems rather alien to me. I wonder if anyone ever actually needs this on the command line.

On the other hand the fact that this command becomes so much more complicated without it is annoying, eval errors will slip in from time to time and this usage of nix-env seems harmless in contrast to the usual critique of that command.

 I think the biggest advantage of having the listing in the manual is that it makes the package set a bit more graspable.

My usual workflow is search.nixos.org or nix repl + tabcompletion, I don‘t know if that’s representative but if it is it might be the way to go also for the manual …

sternenseemann:
> Incredible... just ignore assertions?

You fundamentally have to do this when dealing with how nixpkgs works. `haskellPackages` is a *very* tame corner of nixpkgs, but even here you'll need to deal with assertions if you want to know `outPath`s.

> the only stable, reliable, and (mostly) sane interface to Nix is the Nix language.

It's frankly bizarre to state that to list packages you should

1. Install `jq`,
2. Copy and paste some code snippet that is specific to `haskellPackages`

instead of using a subcommand of `nix-env(1)` that is not problematic.

> I thought it was widely agreed-upon that using `nix-env` should not be encouraged

Yes, `nix-env` for package management is to be discouraged, but that concerns `nix-env -u` and `nix-env -i`. I just don't see any alternative to `nix-env -qP` in this case.

maralorn:
I like this in general.

I still think that having a few example packages listed in the Haskell section makes it easier to quickly grasp the structure of the package sets in question. But overall this change is already an improvement.

PR #317397, opened 2024-06-05
nixos/nextcloud: Cleanup old deprecation warnings

onny:
## Description of changes

- Remove old deprecation and migration warnings

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [x] Tested, as applicable:

---


nixos/modules/services/web-apps/nextcloud.nix
```diff
@@ -811,31 +777,9 @@ in {
...
-      in (optional (cfg.poolConfig != null) ''
-          Using config.services.nextcloud.poolConfig is deprecated and will become unsupported in a future release.
-          Please migrate your configuration to config.services.nextcloud.poolSettings.
-        '')
-        ++ (optional (versionOlder cfg.package.version "25") (upgradeWarning 24 "22.11"))
-        ++ (optional (versionOlder cfg.package.version "26") (upgradeWarning 25 "23.05"))
-        ++ (optional (versionOlder cfg.package.version "27") (upgradeWarning 26 "23.11"))
-        ++ (optional (versionOlder cfg.package.version "28") (upgradeWarning 27 "24.05"))
+      in (optional (versionOlder cfg.package.version "28") (upgradeWarning 27 "24.05"))
         ++ (optional (versionOlder cfg.package.version "29") (upgradeWarning 28 "24.11"));
```
SuperSandro2000, line 781:
```suggestion
      in optional (versionOlder cfg.package.version "28") (upgradeWarning 27 "24.05")
        ++ optional (versionOlder cfg.package.version "29") (upgradeWarning 28 "24.11");
```

nixos/modules/services/web-apps/nextcloud.nix
```diff
@@ -811,31 +777,9 @@ in {
...
-            then throw ''
-              The `pkgs.nextcloud`-attribute has been removed. If it's supposed to be the default
-              nextcloud defined in an overlay, please set `services.nextcloud.package` to
-              `pkgs.nextcloud`.
-            ''
-          else if versionOlder stateVersion "23.05" then nextcloud25
-          else if versionOlder stateVersion "23.11" then nextcloud26
-          else if versionOlder stateVersion "24.05" then nextcloud27
-          else nextcloud29
-        );
```
SuperSandro2000, line 837: Shouldn't we keep a default for package based on stateVersion?

SuperSandro2000 (commented):
I think you missed some entry in top-level about nextcloud27.

Also not sure if we want an alias here or not

pkgs/top-level/all-packages.nix
```diff
@@ -10892,20 +10892,24 @@ with pkgs;
 
   grocy = callPackage ../servers/grocy { };
 
-  inherit (callPackages ../servers/nextcloud {})
-    nextcloud27 nextcloud28 nextcloud29;
-
-  nextcloud27Packages = callPackage ../servers/nextcloud/packages {
-    apps = lib.importJSON ../servers/nextcloud/packages/27.json;
+  nextcloud29 = nextcloud;
+  nextcloud28 = nextcloud.overrideAttrs rec {
```
Ma27, line 10896: So for the sake of having `by-name`, we... make all-packages longer than it was before?

The RFC was supposed to achieve the exact opposite.

:-1: on that.

nixos/doc/manual/release-notes/rl-2411.section.md
```diff
@@ -19,6 +19,12 @@
 - `openssh` and `openssh_hpn` are now compiled without Kerberos 5 / GSSAPI support in an effort to reduce the attack surface of the components for the majority of users. Users needing this support can
   use the new `opensshWithKerberos` and `openssh_hpnWithKerberos` flavors (e.g. `programs.ssh.package = pkgs.openssh_gssapi`).
 
+- `nextcloud27` has been removed since it's not maintained anymore by upstream. The latest available version of Nextcloud is now v29 (available as `pkgs.nextcloud29`). The installation logic is as follows:
+  - If [`services.nextcloud.package`](#opt-services.nextcloud.package) is specified explicitly, this package will be installed (**recommended**).
+  - If [`system.stateVersion`](#opt-system.stateVersion) is >=24.11, `pkgs.nextcloud29` will be installed by default.
```
Ma27, line 24: That's wrong. This seems to be the 24.05 release notes copied, but with altered state versions.

pkgs/top-level/aliases.nix
```diff
@@ -869,18 +869,18 @@ mapAliases ({
   netbox_3_3 = throw "netbox 3.3 series has been removed as it was EOL"; # Added 2023-09-02
   netbox_3_5 = throw "netbox 3.5 series has been removed as it was EOL"; # Added 2024-01-22
   netease-music-tui = throw "netease-music-tui has been removed due to unmaintained by upstream and broken functionality"; # Added 2024-03-03
-  nextcloud26 = throw ''
-    Nextcloud v26 has been removed from `nixpkgs` as the support for is dropped
-    by upstream in 2024-04. Please upgrade to at least Nextcloud v27 by declaring
+  nextcloud27 = throw ''
```
Ma27, line 872: To me it seems too early do that. Especially given that each of these is needed to upgrade, I think it's good to have major releases as long as they're supported in tree.

Ma27 (changes requested):

pkgs/top-level/all-packages.nix
SuperSandro2000: We can keep the old layout and just ignore the failing github action for this PR 😉

Ma27: > We can keep the old layout and just ignore the failing github action for this PR 😉

I think I'm missing something which GH action is failing on this PR?

pkgs/top-level/aliases.nix
onny: Agree with that but probably we could drop it anyways since it's officially EOL in 1 week https://endoflife.date/nextcloud

pkgs/top-level/all-packages.nix
onny: I like the Nextcloud package in `pkgs/by-name/ne/nextcloud` a lot since the `package.nix` file now is very clean and easy to read. Is there any other place where we could define the overrides for the old version 28?

pkgs/top-level/aliases.nix
Ma27: Uhm, I think I confused the nextcloud EOL with the EOL date from a different software, apologies!

Will file a PR that removes it on master now and updates it (still marking it as eol) along with the other updates on 24.05.

Ma27: See https://github.com/NixOS/nixpkgs/pull/322410, https://github.com/NixOS/nixpkgs/pull/322409, https://github.com/NixOS/nixpkgs/pull/322397

onny:
Rebased the pull request :)

pkgs/top-level/all-packages.nix
Ma27: > a lot since the package.nix file now is very clean and easy to read

and the way it was before isn't? This is a pretty common pattern in nixpkgs fwiw.

Also, the consequence is the plain ugly solution with `overrideAttrs`: now we're messing around with derivation-level attributes instead of having a clear interface for building Nextcloud derivations.

> where we could define the overrides for the old version 28?

See https://github.com/NixOS/nixpkgs/blob/5ea6f34ee541b37e8a5abe4d7259dfbd8b8170af/pkgs/by-name/README.md?plain=1#L124

A structure outside of by-name is used even then.

PR #316519, opened 2024-06-01
nixos/espanso: fix espanso options

michaelBelsanti:
## Description of changes

Broke package option in #285138 causing it to be unable to build.

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
- For non-Linux: Is sandboxing enabled in `nix.conf`? (See [Nix manual](https://nixos.org/manual/nix/stable/command-ref/conf-file.html))
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:
  - [NixOS test(s)](https://nixos.org/manual/nixos/unstable/index.html#sec-nixos-tests) (look inside [nixos/tests](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests))
  - and/or [package tests](https://github.com/NixOS/nixpkgs/blob/master/pkgs/README.md#package-tests)
  - or, for functions and "core" functionality, tests in [lib/tests](https://github.com/NixOS/nixpkgs/blob/master/lib/tests) or [pkgs/test](https://github.com/NixOS/nixpkgs/blob/master/pkgs/test)
  - made sure NixOS tests are [linked](https://nixos.org/manual/nixpkgs/unstable/#ssec-nixos-tests-linking) to the relevant packages
- [ ] Tested compilation of all packages that depend on this change using `nix-shell -p nixpkgs-review --run "nixpkgs-review rev HEAD"`. Note: all changes have to be committed, also see [nixpkgs-review usage](https://github.com/Mic92/nixpkgs-review#usage)
- [ ] Tested basic functionality of all binary files (usually in `./result/bin/`)
- [24.11 Release Notes](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2411.section.md) (or backporting [23.11](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2311.section.md) and [24.05](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2405.section.md) Release notes)
  - [ ] (Package updates) Added a release notes entry if the change is major or breaking
  - [ ] (Module updates) Added a release notes entry if the change is significant
  - [ ] (Module addition) Added a release notes entry if adding a new NixOS module
[...]

PR #285138, opened 2024-01-31
nixos/espanso: add wayland and package option

michaelBelsanti:
This adds two options to the espanso module. One that enables wayland support by using the espanso-wayland package, and a package option.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
- For non-Linux: Is sandboxing enabled in `nix.conf`? (See [Nix manual](https://nixos.org/manual/nix/stable/command-ref/conf-file.html))
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:
  - [NixOS test(s)](https://nixos.org/manual/nixos/unstable/index.html#sec-nixos-tests) (look inside [nixos/tests](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests))
  - and/or [package tests](https://nixos.org/manual/nixpkgs/unstable/#sec-package-tests)
  - or, for functions and "core" functionality, tests in [lib/tests](https://github.com/NixOS/nixpkgs/blob/master/lib/tests) or [pkgs/test](https://github.com/NixOS/nixpkgs/blob/master/pkgs/test)
  - made sure NixOS tests are [linked](https://nixos.org/manual/nixpkgs/unstable/#ssec-nixos-tests-linking) to the relevant packages
- [ ] Tested compilation of all packages that depend on this change using `nix-shell -p nixpkgs-review --run "nixpkgs-review rev HEAD"`. Note: all changes have to be committed, also see [nixpkgs-review usage](https://github.com/Mic92/nixpkgs-review#usage)
- [ ] Tested basic functionality of all binary files (usually in `./result/bin/`)
- [24.05 Release Notes](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2405.section.md) (or backporting [23.05](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2305.section.md) and [23.11](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2311.section.md) Release notes)
  - [ ] (Package updates) Added a release notes entry if the change is major or breaking
  - [ ] (Module updates) Added a release notes entry if the change is significant
  - [ ] (Module addition) Added a release notes entry if adding a new NixOS module
[...]

PR #317457, opened 2024-06-05
nixos/espanso: fix wayland option

rawkode:
## Description of changes

The espanso module recently added a wayland option which appears to be missing from the config.

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/nixos/modules/services/desktops/espanso.nix
+++ b/nixos/modules/services/desktops/espanso.nix
@@ -8,6 +8,7 @@ in {
   options = {
     services.espanso = {
       enable = mkEnableOption "Espanso";
+      wayland = mkEnableOption "use the Wayland compatible espanso package";
       package = mkPackageOption pkgs "espanso" {
         example = "pkgs.espanso-wayland";
       };


nixos/modules/services/desktops/espanso.nix
```diff
@@ -8,6 +8,13 @@ in {
   options = {
     services.espanso = {
       enable = mkEnableOption "Espanso";
+      wayland = mkOption {
+        type = types.bool;
+        default = false;
+        description = ''
+          Use the Wayland compatible espanso package
+        '';
```
jtbx, line 16: You can just make this a one liner
```suggestion
        description = "Whether to use the Wayland compatible Espanso package";
```

jtbx (changes requested):
What is this supposed to do? If it's setting the package option, one can already do that manually. In any case it just looks like it's adding an option without any logic. :)

nixos/modules/services/desktops/espanso.nix
rawkode: The module already uses this option and is currently broken because of it. I could have also removed the `mkIf` and we could allow people to set the wayland package manually, but there is a growing convention across other packages to support the wayland option.

jtbx (approved):
Ah I didn't see that down below, thanks.

nixos/modules/services/desktops/espanso.nix
```diff
@@ -8,6 +8,11 @@ in {
   options = {
     services.espanso = {
       enable = mkEnableOption "Espanso";
+      wayland = mkOption {
```
Aleksanaa, line 11: You can directly `mkEnableOption` with "using the Wayland compatible espanso package".

rawkode:
@jtbx @Aleksanaa Can we merge?

nixos/modules/services/desktops/espanso.nix
```diff
@@ -8,6 +8,7 @@ in {
   options = {
     services.espanso = {
       enable = mkEnableOption "Espanso";
+      wayland = mkEnableOption "Use the Wayland compatible espanso package";
```
Aleksanaa, line 11:
```suggestion
      wayland = mkEnableOption "use the Wayland compatible espanso package";
```

```nix
  mkEnableOption =
    # Name for the created option
    name: mkOption {
    default = false;
    example = true;
    description = "Whether to enable ${name}.";
    type = lib.types.bool;
  };
```

Aleksanaa:
Also please squash your commits into `nixos/espanso: fix wayland option`

rawkode:
@Aleksanaa Done

Aleksanaa:
Please squash your commits into `nixos/espanso: fix wayland option`.

rawkode:
Sorry, I missed read your message. This has been resolved.

pitkling:
Was there a race condition between this PR and commit 7f447ce0f88323cc61ce43e538a62379baa2b608 (by @michaelBelsanti)? The latter removed the `mkIf` that was referencing the (at the time of the commit non-existent) `wayland` option. Shortly after, this PR introduced the option, which now is not used at all in the current `master` branch.

pitkling:
Just for completeness: The PR for the commit mentioned above seems to be PR #316519. Also pinging @n8henrie and @SuperSandro2000 who were involved in that PR.

n8henrie:
I thought we had determined not to use a separate `wayland` option since the idiomatic approach is just to allow one to set `package`.

Sorry about missing this, I meant to add myself to espanso maintainers at some point.

n8henrie:
I'll take a closer look later today, currently on mobile.

n8henrie:
@Aleksanaa @rawkode @pitkling Here was [the review](https://github.com/NixOS/nixpkgs/pull/285138#discussion_r1478367044) where @SuperSandro2000 had recommended to just use the `package` option (without a separate `wayland` option), and we had worked towards removing it. Was there a reason to go back the other direction with this?

Should we have just removed the `wayland` option (and all references to it), since it is already [configured to use the `package` option](https://github.com/rawkode/nixpkgs/blob/f13e08ac0e6964fc0087987b08ffc05a979fa548/nixos/modules/services/desktops/espanso.nix#L23)?

PR #317477, opened 2024-06-05
libvpx: fix cross compiling

### LINKED ISSUE #208746: libvpx fails to cross-compile to armv7l-hf-multiplatform
### Steps To Reproduce
1. Make sure that the system is set up for cross-compiling (extra-platforms).
2. `nix-build '<nixpkgs>' -A pkgsCross.armv7l-hf-multiplatform.libvpx`

### Build log
```
Configuring for target 'armv7l-linux-gcc'
  enabling armv7l
# ...
    [LD] libvpx.so.7.1.0
    [AR] libvpx_g.a
    [AR] libvpxrc_g.a
    [STRIP] libvpxrc.a < libvpxrc_g.a
    [STRIP] libvpx.a < libvpx_g.a
/nix/store/y7bs7f9j931b2hps70dc1d165m45cpib-armv7l-unknown-linux-gnueabihf-binutils-2.39/bin/armv7l-unknown-linux-gnueabihf-ld: vpx_dsp/vpx_dsp_rtcd.c.o: in function `setup_rtcd_internal':
vpx_dsp_rtcd.c:(.text+0x1014): undefined reference to `vpx_avg_4x4_neon'
# ...
```

### Additional context
The following patch fixes the cross-build I am looking for (but it might break other builds, I don't know):
```
diff --git a/pkgs/development/libraries/libvpx/default.nix b/pkgs/development/libraries/libvpx/default.nix
index d78a0885..999ef381 100644
--- a/pkgs/development/libraries/libvpx/default.nix
+++ b/pkgs/development/libraries/libvpx/default.nix
@@ -148,7 +148,7 @@ stdenv.mkDerivation rec {
     # libvpx darwin targets include darwin version (ie. ARCH-darwinXX-gcc, XX being the darwin version)
     # See all_platforms: https://github.com/webmproject/libvpx/blob/master/configure
     # Darwin versions: 10.4=8, 10.5=9, 10.6=10, 10.7=11, 10.8=12, 10.9=13, 10.10=14
-    "--force-target=${stdenv.hostPlatform.parsed.cpu.name}-${stdenv.hostPlatform.parsed.kernel.name}${
+    "--target=${stdenv.hostPlatform.parsed.cpu.arch}-${stdenv.hostPlatform.parsed.kernel.name}${
             if stdenv.hostPlatform.isDarwin then
               if      stdenv.hostPlatform.osxMinVersion == "10.10" then "14"
               else if stdenv.hostPlatform.osxMinVersion == "10.9"  then "13"
```
Issue seems to be that the configure script of libvpx does not know the target (as a result, we need to force it), which results in a mess. By using `arch` instead of `name`, the target is understood and everything works for me.

### Notify maintainers
@codyopel 

### Metadata

```console
 - system: `"x86_64-linux"`
 - host os: `Linux 5.15.85, NixOS, 22.11 (Raccoon), 22.11.1198.913a47cd064`
 - multi-user?: `yes`
 - sandbox: `yes`
 - version: `nix-env (Nix) 2.11.1`
 - channels(markus): `""`
 - channels(root): `"home-manager-22.11.tar.gz, nixos-22.11"`
```

RossComputerGuy:
## Description of changes

Fixes #208746

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [x] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/development/libraries/libvpx/default.nix
+++ b/pkgs/development/libraries/libvpx/default.nix
@@ -54,12 +54,26 @@ let
     else if stdenv.hostPlatform.osxMinVersion == "10.5"  then "9"
     else "8";
 
+  cpu =
+    /**/ if stdenv.hostPlatform.isArmv7 then "armv7"
+    else if stdenv.hostPlatform.isAarch64 then "arm64"
+    else if stdenv.hostPlatform.isx86_32 then "x86"
+    else stdenv.hostPlatform.parsed.cpu.name;
+
   kernel =
     # Build system doesn't understand BSD, so pretend to be Linux.
     /**/ if stdenv.isBSD then "linux"
     else if stdenv.isDarwin then "darwin${darwinVersion}"
     else stdenv.hostPlatform.parsed.kernel.name;
 
+  isGeneric =
+    /**/ (stdenv.hostPlatform.isPower && stdenv.hostPlatform.isLittleEndian)
+    || stdenv.hostPlatform.parsed.cpu.name == "armv6l";
+
+  target =
+    /**/ if (stdenv.isBSD || stdenv.hostPlatform != stdenv.buildPlatform) then
+      (if isGeneric then "generic-gnu" else "${cpu}-${kernel}-gcc")
+    else null;
 in
 
 assert vp8DecoderSupport || vp8EncoderSupport || vp9DecoderSupport || vp9EncoderSupport;
@@ -163,8 +177,8 @@ stdenv.mkDerivation rec {
     (enableFeature (experimentalSpatialSvcSupport ||
                     experimentalFpMbStatsSupport ||
                     experimentalEmulateHardwareSupport) "experimental")
-  ] ++ optionals (stdenv.isBSD || stdenv.hostPlatform != stdenv.buildPlatform) [
-    "--force-target=${stdenv.hostPlatform.parsed.cpu.name}-${kernel}-gcc"
+  ] ++ optionals (target != null) [
+    "--target=${target}"
     (lib.optionalString stdenv.hostPlatform.isCygwin "--enable-static-msvcrt")
   ] # Experimental features
     ++ optional experimentalSpatialSvcSupport "--enable-spatial-svc"


pkgs/development/libraries/libvpx/default.nix
```diff
@@ -164,7 +168,7 @@ stdenv.mkDerivation rec {
                     experimentalFpMbStatsSupport ||
                     experimentalEmulateHardwareSupport) "experimental")
   ] ++ optionals (stdenv.isBSD || stdenv.hostPlatform != stdenv.buildPlatform) [
-    "--force-target=${stdenv.hostPlatform.parsed.cpu.name}-${kernel}-gcc"
+    "--target=${cpu}-${kernel}-gcc"
```
uninsane, line 171:
```suggestion
    "--force-target=${cpu}-${kernel}-gcc"
```

the change from `--force-target` to `--target` seems to break `nix-build -A pkgsCross.aarch64-multiplatform.libvpx`. if i use `--force-target` it builds for aarch64 and `pkgsCross.armv7l-hf-multiplatform.libvpx`. i haven't checked what the difference is, the error was during configure:
```
Unrecognized toolchain 'aarch64-linux-gcc'
```

uninsane (changes requested):

pkgs/development/libraries/libvpx/default.nix
RossComputerGuy: `aarch64-linux-gnu` isn't a valid target, apparently it is `arm64`.

uninsane: it's not just aarch64, unfortunately. somehow all these systems work on master (via `pkgsCross.$system.libvpx`), but break with this patch (not exhaustive):
- gnu32
- mips-linux-gnu
- mipsel-linux-gnu
- musl32
- musl64
- ppc64
- raspberryPi (armv6l)
- riscv32
- riscv64
- sheevaplug  (armv5tel)

RossComputerGuy: Interesting, looking at the supported systems according to the configure script. These are the only supported targets:
```
Supported targets:
    arm64-android-gcc        arm64-darwin-gcc         arm64-darwin20-gcc
    arm64-darwin21-gcc       arm64-darwin22-gcc       arm64-darwin23-gcc
    arm64-linux-gcc          arm64-win64-gcc          arm64-win64-vs15
    arm64-win64-vs16         arm64-win64-vs16-clangcl arm64-win64-vs17
    arm64-win64-vs17-clangcl
    armv7-android-gcc        armv7-darwin-gcc         armv7-linux-rvct
    armv7-linux-gcc          armv7-none-rvct          armv7-win32-gcc
    armv7-win32-vs14         armv7-win32-vs15         armv7-win32-vs16
    armv7-win32-vs17
    armv7s-darwin-gcc
    armv8-linux-gcc
    loongarch32-linux-gcc
    loongarch64-linux-gcc
    mips32-linux-gcc
    mips64-linux-gcc
    ppc64le-linux-gcc
    sparc-solaris-gcc
    x86-android-gcc          x86-darwin8-gcc          x86-darwin8-icc
    x86-darwin9-gcc          x86-darwin9-icc          x86-darwin10-gcc
    x86-darwin11-gcc         x86-darwin12-gcc         x86-darwin13-gcc
    x86-darwin14-gcc         x86-darwin15-gcc         x86-darwin16-gcc
    x86-darwin17-gcc         x86-iphonesimulator-gcc  x86-linux-gcc
    x86-linux-icc            x86-os2-gcc              x86-solaris-gcc
    x86-win32-gcc            x86-win32-vs14           x86-win32-vs15
    x86-win32-vs16           x86-win32-vs17
    x86_64-android-gcc       x86_64-darwin9-gcc       x86_64-darwin10-gcc
    x86_64-darwin11-gcc      x86_64-darwin12-gcc      x86_64-darwin13-gcc
    x86_64-darwin14-gcc      x86_64-darwin15-gcc      x86_64-darwin16-gcc
    x86_64-darwin17-gcc      x86_64-darwin18-gcc      x86_64-darwin19-gcc
    x86_64-darwin20-gcc      x86_64-darwin21-gcc      x86_64-darwin22-gcc
    x86_64-darwin23-gcc      x86_64-iphonesimulator-gcc x86_64-linux-gcc
    x86_64-linux-icc         x86_64-solaris-gcc       x86_64-win64-gcc
    x86_64-win64-vs14        x86_64-win64-vs15        x86_64-win64-vs16
    x86_64-win64-vs17
    generic-gnu
```

uninsane: i think we can either use `--target` and map to libvpx's own (bespoke?) database of target triples, or we can use `--force-target` and use the nixpkgs flavor of triples.

- [here's](https://github.com/webmproject/libvpx/blob/main/configure#L98) their target database
- [here's](https://github.com/webmproject/libvpx/blob/main/build/make/configure.sh#L869) the primary way in which `--force-target` differs from ordinary `--target` (at a glance)

RossComputerGuy: Ideally, we should be using `--target` exclusively. It sounds like `--force-target` is like `--target` except it bypasses the toolchain check.

uninsane: i'd guess they do some feature guessing (like e.g. neon fpu) based on the target, which seems to be what caused the original aarch64-multiplatform failure.

i don't see armv5tel (sheevaplug) or armv6l (raspberryPi) in their platform list though, so we may still need to `--force-target` for nixpkgs stdenv triples which can't be mapped onto that list you shared.

RossComputerGuy: I've did some changes that I'll push here in a minute. I've tuned it to where there's an `isGeneric` variable that determines if we should be using `generic-gnu`. On all of the other cases, we will be using the specific target.

tomberek (approved):
built on x86_64-linux via `nix-build -A pkgsCross.aarch64-multiplatform.libvpx`

uninsane:
the solution which was merged breaks some of the targets i listed [here](https://github.com/NixOS/nixpkgs/pull/317477#discussion_r1628200526), e.g. `nix-build -A pkgsCross.riscv64.libvpx`. sorry, i should have noted that -- i thought i'd be able to explore a fix that leverages some of the target mappings in `lib/system/` instead of asking Ross to special-case _every_ system in that list, but i got busy elsewhere :-(

i think those platforms are less used than aarch64, so we're probably still in a better situation than before, and no need to revert it? but apologies if this does break anyone's config.

PR #317530, opened 2024-06-05
nixos/flood: init

thiagokokada:
## Description of changes

Add service module for [`flood`](https://flood.js.org/).

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [x] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [x] Tested, as applicable:

---

### DIFF
--- a/nixos/modules/services/torrent/flood.nix
+++ b/nixos/modules/services/torrent/flood.nix
@@ -1,0 +1,85 @@
+{ config, lib, pkgs, utils, ... }:
+
+let
+  cfg = config.services.flood;
+in
+{
+  meta.maintainers = with lib.maintainers; [ thiagokokada ];
+
+  options.services.flood = {
+    enable = lib.mkEnableOption "flood";
+    package = lib.mkPackageOption pkgs "flood" { };
+    openFirewall = lib.mkEnableOption "" // {
+      description = "Whether to open the firewall for the port in {option}`services.flood.port`.";
+    };
+    port = lib.mkOption {
+      type = lib.types.int;
+      description = "Port to bind webserver.";
+      default = 3000;
+      example = 3001;
+    };
+    host = lib.mkOption {
+      type = lib.types.str;
+      description = "Host to bind webserver.";
+      default = "localhost";
+      example = "::";
+    };
+    extraArgs = lib.mkOption {
+      type = with lib.types; listOf str;
+      description = "Extra arguments passed to `flood`.";
+      default = [ ];
+      example = [ "--baseuri=/" ];
+    };
+  };
+
+  config = lib.mkIf cfg.enable {
+    systemd.services.flood = {
+      description = "A modern web UI for various torrent clients.";
+      after = [ "network.target" ];
+      wantedBy = [ "multi-user.target" ];
+      unitConfig = {
+        Documentation = "https://github.com/jesec/flood/wiki";
+      };
+      serviceConfig = {
+        Restart = "on-failure";
+        RestartSec = "3s";
+        ExecStart = utils.escapeSystemdExecArgs ([
+          (lib.getExe cfg.package)
+          "--host"
+          cfg.host
+          "--port"
+          (toString cfg.port)
+          "--rundir=/var/lib/flood"
+        ] ++ cfg.extraArgs);
+
+        CapabilityBoundingSet = [ "" ];
+        DynamicUser = true;
+        LockPersonality = true;
+        NoNewPrivileges = true;
+        PrivateDevices = true;
+        PrivateTmp = true;
+        ProtectClock = true;
+        ProtectControlGroups = true;
+        ProtectHome = true;
+        ProtectHostname = true;
+        ProtectKernelLogs = true;
+        ProtectKernelModules = true;
+        ProtectKernelTunables = true;
+        ProtectProc = "invisible";
+        ProtectSystem = "strict";
+        RestrictAddressFamilies = [ "AF_UNIX" "AF_INET" "AF_INET6" ];
+        RestrictNamespaces = true;
+        RestrictRealtime = true;
+        RestrictSUIDSGID = true;
+        StateDirectory = "flood";
+        SystemCallArchitectures = "native";
+        SystemCallFilter = [ "@system-service" "@pkey" "~@privileged" ];
+      };
+    };
+
+    networking.firewall.allowedTCPPorts = lib.mkIf cfg.openFirewall [
+      cfg.port
+    ];
+  };
+}
+
--- a/nixos/doc/manual/release-notes/rl-2411.section.md
+++ b/nixos/doc/manual/release-notes/rl-2411.section.md
@@ -15,6 +15,8 @@
 
 - [Quickwit](https://quickwit.io), sub-second search & analytics engine on cloud storage. Available as [services.quickwit](options.html#opt-services.quickwit).
 
+- [Flood](https://flood.js.org/), a beautiful WebUI for various torrent clients. Available as [services.flood](options.html#opt-services.flood).
+
 - [Renovate](https://github.com/renovatebot/renovate), a dependency updating tool for various git forges and language ecosystems. Available as [services.renovate](#opt-services.renovate.enable).
 
 ## Backward Incompatibilities {#sec-release-24.11-incompatibilities}
--- a/nixos/modules/module-list.nix
+++ b/nixos/modules/module-list.nix
@@ -1316,6 +1316,7 @@
   ./services/system/zram-generator.nix
   ./services/torrent/deluge.nix
   ./services/torrent/flexget.nix
+  ./services/torrent/flood.nix
   ./services/torrent/magnetico.nix
   ./services/torrent/opentracker.nix
   ./services/torrent/peerflix.nix
--- a/nixos/tests/all-tests.nix
+++ b/nixos/tests/all-tests.nix
@@ -328,6 +328,7 @@ in {
   firewall-nftables = handleTest ./firewall.nix { nftables = true; };
   fish = handleTest ./fish.nix {};
   flannel = handleTestOn ["x86_64-linux"] ./flannel.nix {};
+  flood = handleTest ./flood.nix {};
   floorp = handleTest ./firefox.nix { firefoxPackage = pkgs.floorp; };
   fluentd = handleTest ./fluentd.nix {};
   fluidd = handleTest ./fluidd.nix {};
--- a/nixos/tests/flood.nix
+++ b/nixos/tests/flood.nix
@@ -1,0 +1,27 @@
+import ./make-test-python.nix ({ pkgs, ... }:
+let
+  port = 3001;
+in
+{
+  name = "flood";
+  meta = {
+    maintainers = with pkgs.lib.maintainers; [ thiagokokada ];
+  };
+
+  nodes.machine = { pkgs, ... }: {
+    services.flood = {
+      inherit port;
+      enable = true;
+      openFirewall = true;
+      extraArgs = [ "--baseuri=/" ];
+    };
+  };
+
+  testScript = /* python */ ''
+    machine.start()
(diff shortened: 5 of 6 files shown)


thiagokokada:
@ofborg test flood

thiagokokada (commented):
CC @winterqt since you maintain the `flood` package.

nixos/modules/services/torrent/flood.nix
```diff
@@ -0,0 +1,115 @@
...
+let
+  cfg = config.services.flood;
+in
+{
+  meta.maintainers = with lib.maintainers; [ thiagokokada ];
+
+  options.services.flood = {
+    enable = lib.mkEnableOption "flood";
+    package = lib.mkPackageOption pkgs "flood" { };
+    openFirewall = lib.mkEnableOption "" // {
```
AndersonTorres, line 12: What does this line do?

thiagokokada: It opens the firewall ports: https://github.com/NixOS/nixpkgs/pull/317530/files#diff-9cac63135951b34a3d402ec244c73ab992337c7eec80645de5a755b982c25371R110-R112

thiagokokada:
@ofborg test flood

nixos/modules/services/torrent/flood.nix
```diff
@@ -0,0 +1,115 @@
...
+    systemd.services.flood = {
+      description = "A web UI for rTorrent with a Node.js backend and React frontend.";
+      after = [ "network.target" ];
+      wantedBy = [ "multi-user.target" ];
+      unitConfig = {
+        Documentation = "https://github.com/jesec/flood/wiki";
+      };
+      serviceConfig = {
+        User = cfg.user;
+        Group = cfg.group;
```
mweinelt, line 71:
```suggestion
    systemd.services.flood = {
      description = "A web UI for rTorrent with a Node.js backend and React frontend.";
      after = [ "network.target" ];
      wantedBy = [ "multi-user.target" ];
      unitConfig = {
        Documentation = "https://github.com/jesec/flood/wiki";
      };
      serviceConfig = {
        DynamicUser = true;
```

There should be no reason for a static user/group allocation.

nixos/modules/services/torrent/flood.nix
```diff
@@ -0,0 +1,115 @@
...
+    user = lib.mkOption {
+      type = lib.types.str;
+      description = "User account used to run flood.";
+      default = "flood";
+    };
+    group = lib.mkOption {
+      type = lib.types.str;
+      description = "Group account used to run flood.";
+      default = "flood";
+    };
```
mweinelt, line 25:
```suggestion
```

nixos/modules/services/torrent/flood.nix
```diff
@@ -0,0 +1,115 @@
...
+    port = lib.mkOption {
+      type = lib.types.int;
+      description = "Port to bind webserver.";
+      default = 3000;
+    };
+    host = lib.mkOption {
+      type = lib.types.str;
+      description = "Host to bind webserver.";
+      default = "0.0.0.0";
```
mweinelt, line 34: - Can this bind dualtstacked? For example using `::`
- Can we default to localhost, just so people don't accidentally expose this?

nixos/modules/services/torrent/flood.nix
```diff
@@ -0,0 +1,115 @@
...
+    };
+    runDir = lib.mkOption {
+      type = lib.types.path;
+      description = "Directory to store Flood configuration. Needs to be readable/writable by Flood's user/group.";
+      default = "/var/lib/flood";
+    };
```
mweinelt, line 40:
```suggestion
```

Use `StateDirectory = "flood";` instead in the systemd unit. There should be no relevant state here.

nixos/modules/services/torrent/flood.nix
```diff
@@ -0,0 +1,115 @@
...
+        RestartSec = "3s";
+        ExecStart = lib.concatStringsSep " " ([
+          "${lib.getExe cfg.package}"
+          "--host"
+          cfg.host
+          "--port"
+          (toString cfg.port)
+          "--rundir"
+          cfg.runDir
+        ] ++ cfg.extraArgs);
```
mweinelt, line 84:
```suggestion
        ExecStart = utils.escapeSystemdExecArgs ([
          (lib.getExe cfg.package)
          "--host"
          cfg.host
          "--port"
          (toString cfg.port)
          "--rundir"
          cfg.runDir
        ] ++ cfg.extraArgs);
```

Make sure we properly escape the exec args.

mweinelt (commented):
I think we can make this service much more lean with a dynamic user and statedir allocation, since it does not actually have relevant state that goes beyond its configuration.

nixos/modules/services/torrent/flood.nix
mweinelt: And actually the project description is

>  A modern web UI for various torrent clients with a Node.js backend and React frontend.

nixos/modules/services/torrent/flood.nix
thiagokokada: - Not sure.
- Yes, good idea.

nixos/modules/services/torrent/flood.nix
thiagokokada: > There should be no reason for a static user/group allocation.

One issue is that in some cases you want a static user/group allocation, for example to reuse a `rtorrent` user to access its permission. In this case you can't use `DynamicUser`.

So I would prefer to keep the service as currently is.

mweinelt: The proper solution for this is `SupplementaryGroup = [ "rtorrent" ]` on the unit IMO.

thiagokokada: @mweinelt Did all the proposed changes. I think in the worst case I can override the service for my own needs. Can you re-review?

nixos/modules/services/torrent/flood.nix
thiagokokada: Done.

nixos/modules/services/torrent/flood.nix
thiagokokada: Done.

nixos/modules/services/torrent/flood.nix
```diff
@@ -0,0 +1,86 @@
...
+      unitConfig = {
+        Documentation = "https://github.com/jesec/flood/wiki";
+      };
+      serviceConfig = {
+        Type = "simple";
+        KillMode = "process";
+        Restart = "on-failure";
+        RestartSec = "3s";
+        ExecStart = utils.escapeSystemdExecArgs ([
+          "${lib.getExe cfg.package}"
```
mweinelt, line 48:
```suggestion
          (lib.getExe cfg.package)
```

nixos/modules/services/torrent/flood.nix
```diff
@@ -0,0 +1,86 @@
...
+  config = lib.mkIf cfg.enable {
+    systemd.services.flood = {
+      description = "A modern web UI for various torrent clients.";
+      after = [ "network.target" ];
+      wantedBy = [ "multi-user.target" ];
+      unitConfig = {
+        Documentation = "https://github.com/jesec/flood/wiki";
+      };
+      serviceConfig = {
+        Type = "simple";
```
mweinelt, line 43:
```suggestion
```

This is the default.

nixos/modules/services/torrent/flood.nix
```diff
@@ -0,0 +1,86 @@
...
+        KillMode = "process";
```
mweinelt, line 44: This is not recommended. Do you have a specific reason for this setting?

> Note that it is not recommended to set KillMode= to process or even none, as this allows processes to escape the service manager's lifecycle and resource management, and to remain running even while their service is considered stopped and is assumed to not consume any resources.

https://www.freedesktop.org/software/systemd/man/latest/systemd.kill.html#KillMode=

nixos/modules/services/torrent/flood.nix
```diff
@@ -0,0 +1,86 @@
...
+        ExecStart = utils.escapeSystemdExecArgs ([
+          "${lib.getExe cfg.package}"
+          "--host"
+          cfg.host
+          "--port"
+          (toString cfg.port)
+          "--rundir=/var/lib/flood"
+        ] ++ cfg.extraArgs);
+
+        CapabilityBoundingSet = "";
```
mweinelt, line 56:
```suggestion
        CapabilityBoundingSet = [ "" ];
```

This is better, since lists merge, but the default remains an explicitly empty set.

nixos/modules/services/torrent/flood.nix
```diff
@@ -0,0 +1,86 @@
...
+        ProtectKernelTunables = true;
+        ProtectProc = "invisible";
+        ProtectSystem = "strict";
+        RestrictAddressFamilies = [ "AF_UNIX" "AF_INET" "AF_INET6" ];
+        RestrictNamespaces = true;
+        RestrictRealtime = true;
+        RestrictSUIDSGID = true;
+        StateDirectory = "flood";
+        SystemCallArchitectures = "native";
+        SystemCallFilter = [ "~@privileged" ];
```
mweinelt, line 77:
```suggestion
        SystemCallFilter = [ "@system-service" "~@privileged" ];
```

Set an explicity allowed scope first, then deny everything you want to exclude. Else it is `everything, except @privileged`, which is a much broader scope.

nixos/modules/services/torrent/flood.nix
thiagokokada: Recommendation from upstream: https://github.com/jesec/flood/wiki/Install-Flood-as-a-service-(systemd).

But maybe this is just something random and it doesn't actually need it. I will remove it.

nixos/modules/services/torrent/flood.nix
thiagokokada: I am almost sure that I did this on purpose, because I was playing with hardening for this service and what I got above was pretty much the most I could harden without breaking it.

But let me try again.

thiagokokada: @mweinelt Adding `@system-service` causes the service to instant crash.

mweinelt: That is because the scope is now narrower. Check dmesg why it was killed.

mweinelt: Suggestion updated.

thiagokokada: Done.

thiagokokada:
@ofborg test flood

mweinelt:
I wish flood worked for Deluge :weary:

AndersonTorres (approved):

nixos/modules/services/torrent/flood.nix
```diff
@@ -0,0 +1,84 @@
...
+{
+  meta.maintainers = with lib.maintainers; [ thiagokokada ];
+
+  options.services.flood = {
+    enable = lib.mkEnableOption "flood";
+    package = lib.mkPackageOption pkgs "flood" { };
+    openFirewall = lib.mkEnableOption "" // {
+      description = "Whether to automatically open the necessary ports in the firewall.";
+      default = true;
+    };
```
mweinelt, line 15: It is convention that `openFirewall` options default to false except in the case of sshd.

thiagokokada: Done, also added examples for `port` and `host` and updated the description in `openFirewall` to be more informative.

thiagokokada:
@ofborg test flood

thiagokokada:
@mweinelt Are there any other concerns in this PR? Can I get an approval here?

mweinelt (approved):

PR #317618, opened 2024-06-06
python311Packages.nireports: init at 23.2.1

bcdarwin:
## Description of changes

Add `python311Packages.nireports`, a package for generating reports on medical image processing pipelines.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/development/python-modules/nireports/default.nix
+++ b/pkgs/development/python-modules/nireports/default.nix
@@ -1,0 +1,97 @@
+{ lib,
+  buildPythonPackage,
+  fetchFromGitHub,
+  pythonOlder,
+  hatch-vcs,
+  hatchling,
+  nipreps-versions,
+  importlib-resources,
+  matplotlib,
+  nibabel,
+  nilearn,
+  nipype,
+  numpy,
+  pandas,
+  pybids,
+  pyyaml,
+  seaborn,
+  svgutils,
+  templateflow,
+  pytestCheckHook,
+  pytest-xdist,
+}:
+
+buildPythonPackage rec {
+  pname = "nireports";
+  version = "23.2.1";
+  pyproject = true;
+
+  disabled = pythonOlder "3.8";
+
+  src = fetchFromGitHub {
+    owner = "nipreps";
+    repo = "nireports";
+    rev = version;
+    hash = "sha256-kr6nWsCA4GcPLEYV/IPm0N3VRz+Mho209/iNVd+59Ik=";
+  };
+
+  build-system = [
+    hatch-vcs
+    hatchling
+    nipreps-versions
+  ];
+
+  dependencies = [
+    matplotlib
+    nibabel
+    nilearn
+    nipype
+    numpy
+    pandas
+    pybids
+    pyyaml
+    seaborn
+    svgutils
+    templateflow
+  ] ++ lib.optionals (pythonOlder "3.12") [
+    importlib-resources
+  ];
+
+  env.SETUPTOOLS_SCM_PRETEND_VERSION = version;
+
+  nativeCheckInputs = [
+    matplotlib
+    pytestCheckHook
+    pytest-xdist
+  ];
+
+  pytestFlagsArray = [ "nireports" ];
+
+  preCheck = ''
+    export HOME=$(mktemp -d)
+  '';
+
+  disabledTests = [
+    # try to download data:
+    "test_cifti_surfaces_plot"
+    "test_mriqc_plot_mosaic"
+    # fails when run under pytest-xdist
+    "test_reportlets.py"
+  ];
+
+  pythonImportsCheck = [
+    "nireports"
+    "nireports.assembler"
+    "nireports.interfaces"
+    "nireports.reportlets"
+    "nireports.tools"
+  ];
+
+  meta = with lib; {
+    description = "The NiPreps Reporting and Visualization system - report templates and 'reportlets'";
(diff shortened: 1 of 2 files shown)


pkgs/development/python-modules/nireports/default.nix
dani0854, line 1: Tests on python311 fail 50% of the time, while creating different tmp files. Looks like they don't work asynchronously. Specifically `test_reportlets.py` while importing `templateflow` on line 34. I have `NIX_BUILD_CORE = 8`.

dani0854 (changes requested):
##### Reviewed points

- [x] package path fits guidelines
- [x] package name fits guidelines
- [x] package version fits guidelines
- [ ] package builds on x86_64-linux
- [x] `meta.description` is set and fits guidelines
- [x] `meta.license` fits upstream license
- [x] source is fetched using the appropriate function

dani0854:
Also, `python312` won't build due to `nose` dependency, #322103 should fix this.

pkgs/development/python-modules/nireports/default.nix
bcdarwin: Disabled this test.

pkgs/development/python-modules/nireports/default.nix
```diff
@@ -0,0 +1,97 @@
...
+    pyyaml
+    seaborn
+    svgutils
+    templateflow
+  ] ++ lib.optionals (pythonOlder "3.12") [
+    importlib-resources
+  ];
+
+  env.SETUPTOOLS_SCM_PRETEND_VERSION = version;
+
```
natsukium, line 61:
```suggestion
```
IIRC, this variable should be set automatically now.

pkgs/development/python-modules/nireports/default.nix
```diff
@@ -0,0 +1,95 @@
...
+  pname = "nireports";
+  version = "23.2.1";
+  pyproject = true;
+
+  disabled = pythonOlder "3.8";
+
+  src = fetchFromGitHub {
+    owner = "nipreps";
+    repo = "nireports";
+    rev = "refs/tags/${version}";
```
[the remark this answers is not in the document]
natsukium: Well, it is a preference, but it has the advantage that we can fetch the source from the tag, even if there is a branch with the same name.

pkgs/development/python-modules/nireports/default.nix
bcdarwin: Didn't work for me even rebasing on `master` and explicitly adding `setuptools-scm` to `build-system`:

```
LookupError: Error getting the version from source `vcs`: setuptools-scm was unable to detect version for /build/source.
 ...
```

natsukium: I didn't notice that this package uses hatch-vsc.
Thanks for trying.

natsukium (approved):

pkgs/development/python-modules/nireports/default.nix
```diff
@@ -0,0 +1,97 @@
...
+  preCheck = ''
+    export HOME=$(mktemp -d)
+  '';
+
+  disabledTests = [
+    # try to download data:
+    "test_cifti_surfaces_plot"
+    "test_mriqc_plot_mosaic"
+    # fails when run under pytest-xdist
+    "test_reportlets.py"
```
dani0854, line 79: That won't disable the test, it still being run. I think to disable all tests in a file you have to use `disabledTestPaths` and specify path from repo root.

Another way this problem can be solved, is by forcing pytest-xdist to run things synchronously. Since there aren't that many tests, and they are quick, distribution is not that important. And running more test is a good thing.
But to do that you will have to play with pytest flags to override:
https://github.com/NixOS/nixpkgs/blob/0158c93ad0e497575058ea7553547665c2a27450/pkgs/development/python-modules/pytest-xdist/default.nix#L45

However, specifying whole file in `disabledTestPaths` also is fine.

dani0854: Played around with flags and option. 
Turns out that if you want to run it on 1 worker, you have to **both** disable the hook using `dontUsePytestXdist = true;`, since it will append `--numprocesses=$NIX_BUILD_CORES` overriding the one in `pytestFlagsArray`.
And add a flag `--numprocesses=1` to `pytestFlagsArray`, to override `-n auto` which is specified in their `pyproject.toml` (I think).

If only one them test still run on amount of cores in `/etc/nix/nix.conf`.
When using only 1 core, it took 89 seconds for me, so not too long.

pkgs/development/python-modules/nireports/default.nix
```diff
@@ -0,0 +1,97 @@
...
+
+  env.SETUPTOOLS_SCM_PRETEND_VERSION = version;
+
+  nativeCheckInputs = [
+    matplotlib
+    pytestCheckHook
+    pytest-xdist
+  ];
+
+  pytestFlagsArray = [ "nireports" ];
```
ShamrockLee, line 68:
```suggestion
  enabledTestPaths = [ "nireports" ];
```

BTW, the relevant attribute to specify extra pytest command-line arguments is `pytestFlags`.

bcdarwin:
Superseded by #509584.

PR #317646, opened 2024-06-06
erlang: patch sh path

liamdiprose:
## Description of changes

RabbitMQ 3.13 crashes because it requires Erlang/OTP 26 and hits [this call to `sh`](https://github.com/erlang/otp/blob/maint-26/lib/os_mon/src/disksup.erl#L243). This patches the call to `sh` with its full path, `${stdenv.shell}/bin/sh`.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

## Questions for the reviewers

- [x] Uses `substituteInPlace`, would you prefer a patch file instead?
- [ ] This issue only started happening in OTP 26, is generic-builder.nix the best place for it?

---

### DIFF
--- a/pkgs/development/interpreters/erlang/generic-builder.nix
+++ b/pkgs/development/interpreters/erlang/generic-builder.nix
@@ -11,6 +11,7 @@
 , ncurses
 , openssl
 , perl
+, runtimeShell
 , autoconf
 , openjdk11 ? null # javacSupport
 , unixODBC ? null # odbcSupport
@@ -115,6 +116,9 @@ stdenv.mkDerivation ({
     patchShebangs make
 
     ${postPatch}
+  '' + optionalString (lib.versionOlder "25" version) ''
+    substituteInPlace lib/os_mon/src/disksup.erl \
+      --replace-fail '"sh ' '"${runtimeShell} '
   '';
 
   # For OTP 27+ we need ex_doc to build the documentation


yurrriq:
Thanks! Should we use pkgs.sh or whatever instead? Otherwise this lgtm

turion:
@ofborg test rabbitmq

liamdiprose:
I've changed `${stdenv.shell}/bin/sh` to `${bash}/bin/sh`, since it is run on the host and isn't related to the build environment.

pkgs/development/interpreters/erlang/generic-builder.nix
```diff
@@ -113,6 +114,8 @@ stdenv.mkDerivation ({
 
   postPatch = ''
     patchShebangs make
+    substituteInPlace lib/os_mon/src/disksup.erl \
+      --replace '"sh ' '"${bash}/bin/sh '
```
superherointj, line 118:
```suggestion
    substituteInPlace lib/os_mon/src/disksup.erl \
      --replace '"sh ' '"${bash}/bin/sh '
```
Currently `substituteInPlace` has `--replace` deprecated.
Suggested usage: `--replace-fail`, `--replace-warn` and `--replace-quiet`

superherointj:
> I've changed `${stdenv.shell}/bin/sh` to `${bash}/bin/sh`, since it is run on the host and isn't related to the build environment.

When I attempted to use NixNG (which uses nixpkgs) [a Nix distribution for containers, like alpine], one important goal was to replace `bash` with `busybox ash` because it is much smaller. I think, using it with default shell makes sense. But I'm also not certain of the nuanced consequences of this. I hear you want to think of the `host` but I saw this as the build of the `erlang` package.

Why is it a problem to keep `${stdenv.shell}/bin/sh`? Wouldn't it offer better compatibility for other shells like `ash`?

superherointj:
>  Uses substituteInPlace, would you prefer a patch file instead?

No. The usage of a patch file should be avoided for simple things like this. As long as the text matching is done properly.

pkgs/development/interpreters/erlang/generic-builder.nix
superherointj: > This issue only started happening in OTP 26, is generic-builder.nix the best place for it?

Considering this, will it fail for other Erlang versions? Humm...

liamdiprose:
> Why is it a problem to keep ${stdenv.shell}/bin/sh? Wouldn't it offer better compatibility for other shells like ash?

I've worked on cross compiling where the `buildInputs` and `nativeBuildInputs` are separate from each other. The `sh` will be ran on the host system, so it doesn't make sense to use the `sh` used to build the package.

The original commit patched `/bin/sh` because I thought it would be a tiny efficient blob of bytes, but its actually a symlink to `${bash}/bin/sh`. I would have liked to see ash there as well.

Since bash is probably already loaded, maybe it's still efficient?

superherointj:
I suspect `erlang/28.nix` will need same patch. I think would be best to assume all forward versions will need this patch.
And then, making it conditional for previous versions at `generic.nix` might be best.

superherointj:
> Since bash is probably already loaded, maybe it's still efficient?

As it is runtime, I think we should use: "${runtimeShell}"

pkgs/development/interpreters/erlang/generic-builder.nix
```diff
@@ -47,6 +48,7 @@
 , odbcPackages ? [ unixODBC ]
 , opensslPackage ? openssl
 , wxPackages ? [ libGL libGLU wxGTK xorg.libX11 wrapGAppsHook3 ]
+, patchShell ? true
```
superherointj, line 51:
```suggestion
, patchShell ? true
```
Sorry, I did not mean this.
I don't think the user needs to tweak this.
I meant just to check for versions directly.

pkgs/development/interpreters/erlang/24.nix
```diff
@@ -3,4 +3,5 @@
 mkDerivation {
   version = "24.3.4.17";
   sha256 = "sha256-V26pZEyFo+c+ztDDkjDNFK6LTw5xzF8gQYepWGNlGKg=";
+  patchShell = false;
```
superherointj, line 6:
```suggestion
```

pkgs/development/interpreters/erlang/25.nix
```diff
@@ -3,4 +3,5 @@
 mkDerivation {
   version = "25.3.2.12";
   sha256 = "sha256-ZqyJtKr0QdzuuzKnJ/S6Z4cGLWolVMVWcMuAHyOGzkQ=";
+  patchShell = false;
```
superherointj, line 6:
```suggestion
```

pkgs/development/interpreters/erlang/generic-builder.nix
```diff
@@ -11,6 +11,7 @@
 , ncurses
 , openssl
 , perl
+, bash
```
superherointj, line 14:
```suggestion
```

pkgs/development/interpreters/erlang/generic-builder.nix
liamdiprose: I think the only users of those options are the sibling `24.nix`, `25.nix` files?

I can't see a good way to check the versions directly, other than using the major-version nix files themselves.

i.e. `25.nix` specifies `version = "25.3.2.12";`.

pkgs/development/interpreters/erlang/generic-builder.nix
```diff
@@ -115,6 +117,9 @@ stdenv.mkDerivation ({
     patchShebangs make
 
     ${postPatch}
+  '' + optionalString patchShell ''
```
superherointj, line 120:
```suggestion
  '' + optionalString patchShell ''
```
I think you should test version here replacing `patchShell`. using `lib.versionOlder`. Does it make sense?

pkgs/development/interpreters/erlang/generic-builder.nix
```diff
@@ -115,6 +117,9 @@ stdenv.mkDerivation ({
...
+    substituteInPlace lib/os_mon/src/disksup.erl \
+      --replace-fail '"sh ' '"${bash}/bin/sh '
```
superherointj, line 122:
```suggestion
      --replace-fail '"sh ' '"${runtimeShell} '
```
(Remember to add `runtimeShell` at top arguments too.)

liamdiprose:
Thanks for the help, I've made the changes.

superherointj:
@ofborg test rabbitmq
@ofborg build pkgsMusl.erlang

superherointj:
Result of `nixpkgs-review pr 317646` run on x86_64-linux [1](https://github.com/Mic92/nixpkgs-review)
<details>
  <summary>37 packages built:</summary>
  <ul>
    <li>akkoma</li>
    <li>cl</li>
    <li>couchdb3</li>
    <li>elixir</li>
    <li>elixir-ls</li>
    <li>elixir_1_10</li>
    <li>elixir_1_11</li>
    <li>elixir_1_12</li>
    <li>elixir_1_13</li>
    <li>elixir_1_14</li>
    <li>elixir_1_15</li>
    <li>elvis-erlang</li>
    <li>erlang</li>
    <li>erlang-ls</li>
    <li>erlang_26</li>
    <li>erlang_27</li>
    <li>erlang_nox</li>
    <li>erlfmt</li>
    <li>ex_doc</li>
    <li>lexical</li>
    <li>lfe</li>
    <li>livebook</li>
    <li>mercury</li>
    <li>mix2nix</li>
    <li>mobilizon</li>
    <li>notmuch-bower</li>
    <li>plausible</li>
    <li>pleroma</li>
    <li>rabbitmq-server</li>
    <li>rabbitmq-server.doc</li>
    <li>rabbitmq-server.man</li>
    <li>rebar</li>
    <li>rebar3</li>
    <li>sonic-pi</li>
    <li>tsung</li>
    <li>wings</li>
    <li>yaws</li>
  </ul>
</details>

superherointj (approved):

Issue #294725, opened 2024-03-10
`runNixOSTest` fails on Darwin (` tests.testers.runNixOSTest-example` fails to build)


ShamrockLee:
### Describe the bug
#282401 should have made NixOS Tests available on Darwin. This should include package tests produced via tester `runNixOSTest`, but the OfBorg build result shows it isn't. An example is `tests.testers.runNixOSTest-example`

See the OfBorg test result in #292759.

### Steps To Reproduce
1. Set up Nix on MacOS
2. `git clone https://github.com/NixOS/nixpkgs.git`
3. `cd nixpkgs`
4. Checkout commit
5. Build `tests.testers.runNixOSTest-example`
6. See build failure

### Expected behavior
Package tests written by `testers.runNixOSTest`, including `tests.testers.runNixOSTest-example`, should run on MacOS. Related OfBorg build should be green.

### Metadata
Please run `nix-shell -p nix-info --run "nix-info -m"` and paste the result.

```console
[user@system:~]$ nix-shell -p nix-info --run "nix-info -m"
output here
```

---

Add a :+1: [reaction] to [issues you find important].

[reaction]: https://github.blog/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/
[issues you find important]: https://github.com/NixOS/nixpkgs/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions-%2B1-desc

Gabriella439:
The macOS support for NixOS tests only adds support for running them but not building them.  You still need access to a Linux builder to build those tests.

ShamrockLee:
Thanks for your quick response!

> The macOS support for NixOS tests only adds support for running them but not building them. You still need access to a Linux builder to build those tests.

What's the difficulty to build NixOS tests on macOS? Isn't it just a QEMU virtual machine?

Gabriella439:
It is possible to build Linux products on NixOS using something like `darwin.linux-builder` but we don't yet have a way to do that in the course of a Nix build.  See:

https://nixos.org/manual/nixpkgs/unstable/#sec-darwin-builder

lilyinstarlight:
[...]

PR #282401, opened 2024-01-20
macOS support for NixOS tests

Gabriella439:
Closes #193336
Closes #261694
Related to #108984 

This pull request is a composite of the changes from #193336 and also #261694, with some changes of my own that I made.

The goal here was to get the following flake to build and run on `aarch64-darwin`:

```nix
{ inputs.nixpkgs.url = "github:Gabriella439/nixpkgs/gabriella/macOS_NixOS_test";

  outputs = { nixpkgs, ... }: {
    checks.aarch64-darwin.default =
      nixpkgs.legacyPackages.aarch64-darwin.nixosTest {
        name = "test";

        nodes.machine = { };

        testScript = ''
          print(machine.succeed('echo hello'));
        '';
      };
  };
}
```

… and after this change it does.  There's no longer a need for the user to set `nodes.*.nixpkgs.pkgs` or
`nodes.*.virtualisation.host.pkgs` as the correct values are inferred from the host system.

This change is spiritually much closer to the approach in #261694 than the approach in #193336.  However, I still made a few changes compared to #261694:

- I didn't include the change to increase `ulimit`

  I think this change was questionable because I feel like a script provided by Nixpkgs shouldn't be tinkering with the user's ambient `ulimit` settings.

- I named the required system feature `hvf` instead of `apple-virt`

  I preferred the `hvf` system feature name chosen by #193336.

- I didn't use the `node.pkgs` setting to set `nodes.*.nixpkgs.pkgs`

  That does not work, based on my testing, but setting `nixpkgs.system` in `nixos/lib/testing/nodes.nix` does work with some supporting changes to `pkgs/build-support/testers/default.nix`.

## Things done

- Built on platform(s)
  - [x] aarch64-darwin
- [ ] Tested, as applicable:
  - [NixOS test(s)](https://nixos.org/manual/nixos/unstable/index.html#sec-nixos-tests) (look inside [nixos/tests](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests))
  - and/or [package tests](https://nixos.org/manual/nixpkgs/unstable/#sec-package-tests)
  - or, for functions and "core" functionality, tests in [lib/tests](https://github.com/NixOS/nixpkgs/blob/master/lib/tests) or [pkgs/test](https://github.com/NixOS/nixpkgs/blob/master/pkgs/test)
  - made sure NixOS tests are [linked](https://nixos.org/manual/nixpkgs/unstable/#ssec-nixos-tests-linking) to the relevant packages
[...]

PR #317651, opened 2024-06-06
pkgsLinux: set crossSystem instead of localSystem

tie:
## Description of changes

As the doc implies, we should be using crossSystem to build *for* the Linux platform. Using localSystem just happens to occasionally hit cached builds. See also https://github.com/NixOS/nixpkgs/pull/316659#issuecomment-2149539101 and https://github.com/NixOS/nixpkgs/runs/25855763929

See also #282401, #294725

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [x] aarch64-linux
  - [ ] x86_64-darwin
  - [x] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/top-level/stage.nix
+++ b/pkgs/top-level/stage.nix
@@ -249,7 +249,7 @@ let
       if stdenv.hostPlatform.isLinux
       then self
       else nixpkgsFun {
-        localSystem = lib.systems.elaborate "${stdenv.hostPlatform.parsed.cpu.name}-linux";
+        crossSystem = lib.systems.elaborate "${stdenv.hostPlatform.parsed.cpu.name}-linux";
       };
 
     # Extend the package set with zero or more overlays. This preserves
--- a/pkgs/development/libraries/tpm2-tss/default.nix
+++ b/pkgs/development/libraries/tpm2-tss/default.nix
@@ -1,8 +1,8 @@
-{ stdenv, lib, fetchFromGitHub
+{ stdenv, lib, fetchFromGitHub, runCommand
 , autoreconfHook, autoconf-archive, pkg-config, doxygen, perl
 , openssl, json_c, curl, libgcrypt
 , cmocka, uthash, ibm-sw-tpm2, iproute2, procps, which
-, shadow, libuuid
+, libuuid
 }:
 let
   # Avoid a circular dependency on Linux systems (systemd depends on tpm2-tss,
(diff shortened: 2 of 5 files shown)


tie:
The latest commit would likely cause a lot of rebuilds since it’s systemd dependency. I’ll rebase to the staging branch later.

tie:
OK, apparently some Linux-specific packages are still getting pulled from `hostPkgs` (mostly for full systemd, for some reason, e.g. bpftool). Marking the PR as ready for review so we’d at least get an evaluation failure instead of confusing build failure for now.

Gabriella439 (approved):

tie:
Ah, I see #293573. That’s still seems like a really weird and niche use case to me — I mean, if it requires linux builder, so why not run tests there? Instead, I think we should be able to cross-compile NixOS from macOS with some effort.

baloo (approved):
This will need to retarget to `staging` (because of tpm2-tss), but for the tpm2 changes I don't see issues.

pkgs/top-level/stage.nix
```diff
@@ -249,7 +249,7 @@ let
       if stdenv.hostPlatform.isLinux
       then self
       else nixpkgsFun {
-        localSystem = lib.systems.elaborate "${stdenv.hostPlatform.parsed.cpu.name}-linux";
+        crossSystem = lib.systems.elaborate "${stdenv.hostPlatform.parsed.cpu.name}-linux";
```
roberth, line 252: This is _meant_ to be a natively built Linux package set, ie a non-cross build that's potentially done on a different machine. Do you think we _non-cross_ would be a clearer description than _natively built_ in the doc above?
Note that the "machine" is easy to set up with for example [`linux-builder`](https://daiderd.com/nix-darwin/manual/#opt-nix.linux-builder.enable).

By changing to cross compilation, we make the user experience worse and increase cost.
- Cross compilation, despite how well it works in Nixpkgs, does not permit tests to be run in package builds
- We don't have a cache for darwin->linux cross builds, afaik. These builds would be unnecessary extra builds, costing compute and cache storage (ofborg, hydra, cache.nixos.org, etc)

Also note that if you're in a team using a mix of Linux and Darwin and you're doing deployment with cross builds, your performing unnecessary redeployments depending on who's initiating it, causing unnecessary disruptions as system services are "updated" to an equivalent package on a different store path. You could solve this by deploying from dedicated infrastructure, but then you might as well use that for remote builds as well.
More importantly though, non-cross builds are more likely to work.
All in all, non-cross deployments are simpler and more robust, so I would not default to cross compilation anywhere, including here.

Perhaps your goal could be achieved by adding `pkgsLinuxCross`, but then I still wouldn't make the test framework use that for the reasons above.

roberth (changes requested):
See comment.
To summarize, building in a VM is preferable over cross compilation for various reasons.

roberth:
I'd previously asked the infra team to provide a Linux builder, but they were concerned about a deadlock issue. I've now opened an issue with alternatives that avoid the problem (and it's also actually properly visible, unlike chat in Matrix).
- https://github.com/NixOS/infra/issues/439

pkgs/top-level/stage.nix
tie: That sounds reasonable given the current state of cross-compilation of Nixpkgs (though I wouldn’t say it works well, rather it works for well-tested {local,cross}System pairs, but that doesn’t cover glibc → glibc static cross-build, NixOS built from macOS, and a lot of other cases).

I think we’ve had a similar discussion in https://github.com/NixOS/nix/pull/10291 about this use case. W.r.t. tests, these do not change the resulting package output. So it doesn’t really matter where the tests are run as long the closures are byte-for-byte identical and reproducible. That is, as long as we can test that a certain set of packages is identical independent of the `localSystem`, it doesn’t really matter which derivation runs the tests (as long as at least one does). Sure, that requires content-addressed derivations to get this right, but then there are a lot of other places that should be fixed before that.

I understand your point, but I do want to improve the current state of affairs. I’ll open separate PR for package-specific commits I’ve pushed to this branch, without this particular change.

PR #276408, opened 2023-12-24
gst_all_1.gst-plugins-good: dlopen libsoup_3 with an absolute path

ghost:
## Description of changes
fixes #271960 

the libgstsoup plugin in the gstreamer gst-pluggins-good loads the libsoup (2 or 3) via a call to g_module_open -- g_module_open is implemented by glib as a dlopen in the libgmodule.so library. This call fails for both darwin and linux as libsoup is not found in the runpath of the libgmodule.so object. This change patches [gstsouploader.c](https://gitlab.freedesktop.org/gstreamer/gstreamer/-/blob/e19428a802c2f4ee9773818aeb0833f93509a1c0/subprojects/gst-plugins-good/ext/soup/gstsouploader.c#L184) to fall back on an absolute path for libsoup_3 if libsoup (2 or 3) is not already loaded so the subsequent g_module_open succeeds.

consumers of gst-pluggins-good need to add glib-networking to enable TLS support

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [x] x86_64-darwin
  - [ ] aarch64-darwin
- For non-Linux: Is sandboxing enabled in `nix.conf`? (See [Nix manual](https://nixos.org/manual/nix/stable/command-ref/conf-file.html))
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:
  - [NixOS test(s)](https://nixos.org/manual/nixos/unstable/index.html#sec-nixos-tests) (look inside [nixos/tests](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests))
  - and/or [package tests](https://nixos.org/manual/nixpkgs/unstable/#sec-package-tests)
  - or, for functions and "core" functionality, tests in [lib/tests](https://github.com/NixOS/nixpkgs/blob/master/lib/tests) or [pkgs/test](https://github.com/NixOS/nixpkgs/blob/master/pkgs/test)
  - made sure NixOS tests are [linked](https://nixos.org/manual/nixpkgs/unstable/#ssec-nixos-tests-linking) to the relevant packages
- [ ] Tested compilation of all packages that depend on this change using `nix-shell -p nixpkgs-review --run "nixpkgs-review rev HEAD"`. Note: all changes have to be committed, also see [nixpkgs-review usage](https://github.com/Mic92/nixpkgs-review#usage)
- [x] Tested basic functionality of all binary files (usually in `./result/bin/`)
[...]

PR #317737, opened 2024-06-06
gst_all_1.gst-plugins-good: restore support for dlopen libsoup on Darwin

reckenrode:
I ran into this issue when trying to enable and use GStreamer in Wine. Wine because gst-plugins-bad links libsoup 2.4 while gst-plugins-good is linked against libsoup 3.

Upstream removed support for using dlopen on platforms other than Linux to fix https://gitlab.freedesktop.org/gstreamer/gstreamer/-/issues/1171, but Darwin in nixpkgs also needs to `dlopen` libsoup because gst-plugins-bad depends on libsoup 2.4 indirectly via libnice.

https://github.com/NixOS/nixpkgs/pull/276408 fixed the issue for Linux, but this PR expands the patch to work with Darwin as well by restoring support for using `dlopen` on Darwin. I have tried to maintain the spirit of preferring libsoup3 on Linux when no other libsoup is in use, but I have not tested it. On Darwin, libsoup 2.4 is preferred.

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [x] x86_64-darwin
  - [x] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/development/libraries/glib/gmodule-rtld_local.patch
+++ b/pkgs/development/libraries/glib/gmodule-rtld_local.patch
@@ -1,0 +1,13 @@
+diff --git a/gmodule/gmodule-dl.c b/gmodule/gmodule-dl.c
+index 6d1c5fab7..b21773bca 100644
+--- a/gmodule/gmodule-dl.c
++++ b/gmodule/gmodule-dl.c
+@@ -136,7 +136,7 @@ _g_module_open (const gchar *file_name,
+   
+   lock_dlerror ();
+   handle = dlopen (file_name,
+-		   (bind_local ? 0 : RTLD_GLOBAL) | (bind_lazy ? RTLD_LAZY : RTLD_NOW));
++		   (bind_local ? RTLD_LOCAL : RTLD_GLOBAL) | (bind_lazy ? RTLD_LAZY : RTLD_NOW));
+   if (!handle)
+     {
+       const gchar *message = fetch_dlerror (TRUE);
--- a/pkgs/development/libraries/gstreamer/good/souploader.diff
+++ b/pkgs/development/libraries/gstreamer/good/souploader.diff
@@ -1,8 +1,17 @@
 diff --git a/ext/soup/gstsouploader.c b/ext/soup/gstsouploader.c
-index 85048ce303..d7d818cf95 100644
+index 85a033668e..8082b2614d 100644
 --- a/ext/soup/gstsouploader.c
 +++ b/ext/soup/gstsouploader.c
-@@ -181,7 +181,7 @@ gst_soup_load_library (void)
+@@ -197,7 +197,7 @@ gst_soup_load_library (void)
+ 
+     if (!libsoup_sonames[0]) {
+       GST_DEBUG ("Trying all libsoups");
+-      libsoup_sonames[0] = LIBSOUP_3_SONAME;
++      libsoup_sonames[0] = "@nixLibSoup3Path@/" LIBSOUP_3_SONAME;
+       libsoup_sonames[1] = LIBSOUP_2_SONAME;
+     }
+   }
+@@ -216,7 +216,7 @@ gst_soup_load_library (void)
        GST_DEBUG ("LibSoup 2 found");
      } else {
        GST_DEBUG ("Trying all libsoups");
--- a/pkgs/development/libraries/glib/default.nix
+++ b/pkgs/development/libraries/glib/default.nix
@@ -130,6 +130,9 @@ stdenv.mkDerivation (finalAttrs: {
     # and by default meson installs in to $out/share/gdb/auto-load
     # which does not help
     ./gdb_script.patch
+
+    # glib assumes that `RTLD_LOCAL` is defined to `0`, which is true on Linux and FreeBSD but not on Darwin.
+    ./gmodule-rtld_local.patch
   ];
 
   outputs = [ "bin" "out" "dev" "devdoc" ];
--- a/pkgs/development/libraries/gstreamer/good/default.nix
+++ b/pkgs/development/libraries/gstreamer/good/default.nix
@@ -68,6 +68,8 @@ stdenv.mkDerivation rec {
   };
 
   patches = [
+    # Reenable dynamic loading of libsoup on Darwin and use a different approach to do it.
+    ./souploader-darwin.diff
     # dlopen libsoup_3 with an absolute path
     (substituteAll {
       src = ./souploader.diff;
--- a/pkgs/development/libraries/gstreamer/good/souploader-darwin.diff
+++ b/pkgs/development/libraries/gstreamer/good/souploader-darwin.diff
@@ -1,0 +1,128 @@
+diff --git a/ext/adaptivedemux2/meson.build b/ext/adaptivedemux2/meson.build
+index 711b38a2a9..67a789f664 100644
+--- a/ext/adaptivedemux2/meson.build
++++ b/ext/adaptivedemux2/meson.build
+@@ -82,7 +82,7 @@ soup_link_args = []
+ soup_link_deps = []
+ 
+ default_library = get_option('default_library')
+-if host_system != 'linux' or default_library in ['static', 'both']
++if default_library in ['static', 'both']
+   if soup_ver_opt in ['auto', '3']
+     libsoup3_dep = dependency('libsoup-3.0', allow_fallback: true,
+                               required: soup_ver_opt == '3' and soup_opt.enabled())
+@@ -120,7 +120,7 @@ adaptive_deps = [gmodule_dep, gst_dep, gsttag_dep, gstnet_dep, gstbase_dep, gstp
+ adaptive_args = [gst_plugins_good_args, soup_loader_args, hls_cargs,
+     '-DGST_ISOFF_API=G_GNUC_INTERNAL']
+ 
+-if host_system != 'linux'
++if false
+     adaptivedemux2 = library('gstadaptivedemux2',
+       c_args: [adaptive_args, soup_link_args],
+       dependencies: [adaptive_deps, soup_link_deps],
+diff --git a/ext/soup/gstsouploader.c b/ext/soup/gstsouploader.c
+index 9192e4dac5..8082b2614d 100644
+--- a/ext/soup/gstsouploader.c
++++ b/ext/soup/gstsouploader.c
+@@ -34,12 +34,18 @@ GST_DEBUG_CATEGORY (gst_soup_debug);
+ 
+ #ifndef LINK_SOUP
+ 
+-#if defined(__APPLE__) || defined(G_OS_WIN32)
+-#error "dlopen of libsoup is only supported on Linux"
++#if defined(G_OS_WIN32)
++#error "dlopen of libsoup is only supported on Darwin and Linux"
+ #endif
+ 
++#if defined(__APPLE__)
++#define LIBSOUP_3_SONAME "libsoup-3.0.0.dylib"
++#define LIBSOUP_2_SONAME "libsoup-2.4.1.dylib"
++#define LIBSOUP_COMMON_SYMBOL "soup_get_major_version"
++#else
+ #define LIBSOUP_3_SONAME "libsoup-3.0.so.0"
+ #define LIBSOUP_2_SONAME "libsoup-2.4.so.1"
++#endif
+ 
+ #define LOAD_SYMBOL(name) G_STMT_START {                                \
+     if (!g_module_symbol (module, G_STRINGIFY (name), (gpointer *) &G_PASTE (vtable->_, name))) { \
+@@ -156,7 +162,46 @@ gst_soup_load_library (void)
+   GST_DEBUG_CATEGORY_INIT (gst_soup_debug, "soup", 0, "soup");
+ #endif
+ 
+-#ifdef HAVE_RTLD_NOLOAD
++#if defined(__APPLE__)
++  g_autofree gchar* libsoup_path = NULL;
++  {
++    /* In order to avoid causing conflicts we detect if libsoup 2 or 3 is loaded already.
++     * Darwin has to probe by checking for symbols because `dlopen` called with a dylib
++     * basename will try to locate it via the process’s rpath stack (and not find it). */
++
++    gpointer func = NULL;
++    Dl_info info = { 0 };
++
++    GModule* module = g_module_open (NULL, 0);
++    if (g_module_symbol (module, LIBSOUP_COMMON_SYMBOL, &func)
++        && dladdr (func, &info) && info.dli_sname) {
++      libsoup_path = g_strndup(info.dli_fname, PATH_MAX - 1);
++      g_autofree gchar* image_name = g_path_get_basename (libsoup_path);
++
++      /* Make sure `libsoup_path` points to a dylib that actually exists and
++       * contains the libsoup symbol that was queried. */
++      gpointer handle = dlopen (libsoup_path, RTLD_NOW | RTLD_NOLOAD);
++      gboolean has_symbol = dlsym (handle, LIBSOUP_COMMON_SYMBOL) != NULL;
++      if (handle && has_symbol && g_str_equal (image_name, LIBSOUP_3_SONAME)) {
++        libsoup_sonames[0] = libsoup_path;
++        GST_DEBUG ("LibSoup 3 found");
++      } else if (handle && has_symbol && g_str_equal (image_name, LIBSOUP_2_SONAME)) {
++        libsoup_sonames[0] = libsoup_path;
++        GST_DEBUG ("LibSoup 2 found");
++      } else {
++        g_clear_pointer (&libsoup_path, g_free);
++      }
++      g_clear_pointer (&handle, dlclose);
++    }
++    g_module_close (module);
++
++    if (!libsoup_sonames[0]) {
++      GST_DEBUG ("Trying all libsoups");
++      libsoup_sonames[0] = LIBSOUP_3_SONAME;
++      libsoup_sonames[1] = LIBSOUP_2_SONAME;
++    }
++  }
++#elif defined(HAVE_RTLD_NOLOAD)
+   {
+     gpointer handle = NULL;
+ 
+diff --git a/ext/soup/meson.build b/ext/soup/meson.build
+index aaa01dbcf6..83b7cf3fbf 100644
+--- a/ext/soup/meson.build
++++ b/ext/soup/meson.build
+@@ -20,8 +20,8 @@ soup_link_deps = []
+ libsoup2_dep = disabler()
+ libsoup3_dep = disabler()
+ default_library = get_option('default_library')
+-soup_lookup_dep = get_option('soup-lookup-dep') and host_system == 'linux'
+-if host_system != 'linux' or default_library in ['static', 'both'] or soup_lookup_dep
++soup_lookup_dep = get_option('soup-lookup-dep')
++if default_library in ['static', 'both'] or soup_lookup_dep
+   if soup_ver_opt in ['auto', '3']
+     libsoup3_dep = dependency('libsoup-3.0', allow_fallback: true,
+                               required: soup_ver_opt == '3' and soup_opt.enabled())
+@@ -33,7 +33,7 @@ if host_system != 'linux' or default_library in ['static', 'both'] or soup_looku
+   endif
+ endif
+ 
+-if host_system != 'linux' or default_library in ['static', 'both']
++if default_library in ['static', 'both']
+   if libsoup3_dep.found()
+     soup_link_deps += libsoup3_dep
+     soup_link_args += '-DLINK_SOUP=3'
+@@ -60,7 +60,7 @@ soup_library_kwargs = {
+ soup_library_deps = [gst_dep, gstbase_dep, gsttag_dep, gmodule_dep, gio_dep, libdl_dep]
+ soup_library_c_args = gst_plugins_good_args
+ 
+-if host_system != 'linux'
++if false
+   gstsouphttpsrc = library('gstsoup',
+     c_args : soup_library_c_args + soup_link_args,
+     dependencies : soup_library_deps + soup_link_deps,


paparodeo (commented):
there seems to be a subtle difference in dlopen behavior on linux which could cause issues when a libsoup from a different derivation is loaded than the one specified in the full path.

why does darwin prefer libsoup-2?

pkgs/development/libraries/gstreamer/good/souploader.diff
```diff
@@ -1,13 +1,164 @@
...
+@@ -163,16 +168,16 @@ gst_soup_load_library (void)
+     /* In order to avoid causing conflicts we detect if libsoup 2 or 3 is loaded already.
+      * If so use that. Otherwise we will try to load our own version to use preferring 3. */
+ 
+-    if ((handle = dlopen (LIBSOUP_3_SONAME, RTLD_NOW | RTLD_NOLOAD))) {
+-      libsoup_sonames[0] = LIBSOUP_3_SONAME;
+-      GST_DEBUG ("LibSoup 3 found");
+-    } else if ((handle = dlopen (LIBSOUP_2_SONAME, RTLD_NOW | RTLD_NOLOAD))) {
+-      libsoup_sonames[0] = LIBSOUP_2_SONAME;
++    if ((handle = dlopen ("@nixLibSoup2Path@/" LIBSOUP_2_SONAME, RTLD_NOW | RTLD_NOLOAD))) {
```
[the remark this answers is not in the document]
reckenrode: Linux doesn’t use the full path for libsoup 2.4. It only uses it for libsoup3, which is duplicating the behavior of https://github.com/NixOS/nixpkgs/pull/276408.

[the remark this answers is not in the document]
paparodeo: which libsoup the testcode is linking is irrelevant as the same issue happens when linking to libsoup_3 and then trying to dlopen libsoup_3 using the fullpath of a copy.

pkgs/development/libraries/gstreamer/good/souploader.diff
```diff
@@ -1,13 +1,164 @@
...
+-  endif
+-  if default_library in ['shared', 'both']
+-    gstsouphttpsrc_shared = shared_library('gstsoup',
+-      c_args : soup_library_c_args,
+-      dependencies : soup_library_deps,
+-      kwargs: soup_library_kwargs,
+-    )
+-  endif
++endif
++if default_library in ['shared', 'both']
```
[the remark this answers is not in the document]
reckenrode: It’s set on line 133 of the patch. Might be more obvious if I switch the `if host_system != 'linux'` lines to `if false`.

pkgs/development/libraries/gstreamer/good/souploader.diff
```diff
@@ -1,13 +1,164 @@
...
+-if host_system != 'linux' or default_library in ['static', 'both']
++if default_library in ['static', 'both']
+   if libsoup3_dep.found()
+     soup_link_deps += libsoup3_dep
+     soup_link_args += '-DLINK_SOUP=3'
+@@ -60,29 +60,19 @@ soup_library_kwargs = {
+ soup_library_deps = [gst_dep, gstbase_dep, gsttag_dep, gmodule_dep, gio_dep, libdl_dep]
+ soup_library_c_args = gst_plugins_good_args
+ 
+-if host_system != 'linux'
```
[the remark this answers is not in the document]
reckenrode: I changed the patch to use this approach. It should also help with clarity.

pkgs/development/libraries/gstreamer/good/souploader.diff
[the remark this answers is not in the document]
reckenrode: In the update, I populate the trailing slash in the substitution. Should I have separate patches for both Linux and Darwin (to preserve the libsoup3 first checks for Linux)?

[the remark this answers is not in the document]
paparodeo: i don't think switching the order is an issue as on linux libsoup_3 will get loaded using the full path if no other libsoups are detected.

to fix the subtle linux issue could do something like

```c
#ifdef __APPLE__
#define MAKE_LIB_PATH(dirname, basename) dirname basename
#else
#define MAKE_LIB_PATH(dirname, basename) basename
#endif
[...]
if ((handle = dlopen (MAKE_LIB_PATH("@nixLibSoup2Path@/", LIBSOUP_2_SONAME), RTLD_NOW | RTLD_NOLOAD))) {
```

should work so linux uses the basename and apple uses the full path.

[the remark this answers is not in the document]
reckenrode: @paparodeo I pushed an update taking a different approach. On Darwin, it uses dyld APIs to check the loaded images for libsoup. If it finds one, it uses that. Otherwise, it defaults to libsoup3 from the store.

Linux should work exactly as it did before.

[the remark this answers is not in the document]
reckenrode: Technically, this approach isn’t thread safe. An image could be added or removed while the search is being performed. Unfortunately, I don’t know enough about how gstreamer loads plugins whether it is safe to assume that once loaded they remain loaded, and the thread safe APIs that are available require a newer SDK than is available in nixpkgs. However, I was able to successfully use GStreamer with Wine with this patch.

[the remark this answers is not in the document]
paparodeo: that looks good! i was playing around with `dladdr(dlsym(RTLD_DEFAULT, "soup_get_major_version"), &info)` earlier which seems to also provide the image name, though has downside of breaking if the symbol name gets changed or removed.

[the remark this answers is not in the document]
paparodeo: > Technically, this approach isn’t thread safe. An image could be added or removed while the search is being performed. 

this seems like an unlikely event that we can live with as the current version (linux) isn't thread safe either. the image can be unloaded after detection and then will fail to load in the g_module_open call.

I don't think the gstreamer plugins will get unloaded. i thought that at init time the plugin list is searched and loaded and then that's it. tho i could be misremembering as it was 6-7 months ago since i was looking at this code.

pkgs/development/libraries/gstreamer/good/souploader.diff
```diff
@@ -1,8 +1,83 @@
...
+   GST_DEBUG_CATEGORY_INIT (gst_soup_debug, "soup", 0, "soup");
+ #endif
+ 
+-#ifdef HAVE_RTLD_NOLOAD
++#if defined(__APPLE__)
++  {
++    /* Use Darwin’s dyld APIs to check the loaded images for one that matches
++     * since `dlopen` won’t find a match with just a bare soname. */
++    for (uint32_t idx = 0, count = _dyld_image_count(); idx < count; ++idx) {
++      const gchar* image_path = _dyld_get_image_name(idx);
```
paparodeo, line 58: perhaps overly cautious but might want to check that `image_path` is non-null

reckenrode: That should only happen when an image is removed, and `idx` is now out of bounds. I assume by your edit that GStreamer plugins can actually get unloaded?

I force-pushed an implementation of your idea of using `dladdr(dlsym(RTLD_DEFAULT, "soup_get_major_version"), &info)` to probe for libsoup. If libsoup drops `soup_get_major_version`, we probably have more problems than just not finding the dylib anymore. It’s only for Darwin because of the glibc bug note at the end of https://linux.die.net/man/3/dladdr.

reckenrode: https://gitlab.gnome.org/GNOME/glib/-/blob/601b8b45fb4ad5bc4c9f59df883bcd4918f3ecd0/gmodule/gmodule-dl.c#L139

I don’t think that’s right for Darwin. glibc defines `RTLD_LOCAL` as `0x0`, but Darwin defines it as `0x4`.

https://sourceware.org/git/?p=glibc.git;a=blob;f=bits/dlfcn.h;h=dd5bb46a182d93a6d7217cb1e575456b4b64bf70;hb=HEAD#l38
https://github.com/apple-oss-distributions/dyld/blob/d552c40cd1de105f0ec95008e0e0c0972de43456/include/dlfcn.h#L87

According to POSIX, if you don’t specify `RLTD_LOCAL` or `RTLD_GLOBAL`, the behavior is unspecified. On Linux (and FreeBSD), the default is `RTLD_LOCAL`. On Darwin, it’s apparently `RTLD_GLOBAL`.

https://pubs.opengroup.org/onlinepubs/009696799/functions/dlopen.html

I’m testing a glib patch to use `RTLD_LOCAL` explicitly to see if it fits the plugin ordering problem I was still observing.

reckenrode: See the latest force push. I’m pretty sure that’s a long-standing glib bug with `g_module_open` on Darwin. I included a fix for that and cleaned up the probing. Darwin now loads libsoup without issue regardless of the order of folders specified on `GST_PLUGIN_PATH`. Notice how it uses libsoup2 when gst-plugins-bad is first and libsoup3 (with `RTLD_LOCAL`) when gst-plugins-good is first.

```shell
$ rm -f foo.bin && GST_REGISTRY=foo.bin GST_DEBUG='*soup*:6' GST_PLUGIN_PATH=gst-plugins-bad/lib/gstreamer-1.0:gst-plugins-good/lib/gstreamer-1.0 result-gst-2-bin/bin/gst-inspect-1.0 adaptivedemux2
0:00:00.340118125 88901 0x60000077ca50 DEBUG    adaptivedemux2-soup gstsouploader.c:184:gst_soup_load_library: LibSoup 2 found
0:00:00.340414708 88901 0x60000077ca50 DEBUG    adaptivedemux2-soup gstsouploader.c:229:gst_soup_load_library: Loaded /nix/store/3kp87dcj8r9zrsz6kfnr7ds2y3s4h7zs-libsoup-2.74.3/lib/libsoup-2.4.1.dylib
0:00:00.349393375 88901 0x60000077ca50 DEBUG              souputils gstsoupelement.c:38:soup_element_init: binding text domain gst-plugins-good-1.0 to locale dir /nix/store/6vinyky0acyww6a767xq7n1w7zzxm5m9-gst-plugins-good-1.24.2/share/locale
0:00:00.349510500 88901 0x60000077ca50 DEBUG                   soup gstsouploader.c:184:gst_soup_load_library: LibSoup 2 found
0:00:00.349515500 88901 0x60000077ca50 DEBUG                   soup gstsouploader.c:229:gst_soup_load_library: Loaded /nix/store/3kp87dcj8r9zrsz6kfnr7ds2y3s4h7zs-libsoup-2.74.3/lib/libsoup-2.4.1.dylib
Plugin Details:
  Name                     adaptivedemux2
  Description              Adaptive Streaming 2 plugin
  Filename                 gst-plugins-good/lib/gstreamer-1.0/libgstadaptivedemux2.dylib
  Version                  1.24.2
  License                  LGPL
  Source module            gst-plugins-good
  Documentation            https://gstreamer.freedesktop.org/documentation/adaptivedemux2/
  Source release date      2024-04-09
  Binary package           GStreamer Good Plug-ins source release
  Origin URL               Unknown package origin

  dashdemux2: DASH Demuxer
  hlsdemux2: HLS Demuxer
  mssdemux2: Smooth Streaming demuxer (v2)

  3 features:
  +-- 3 elements

$ rm -f foo.bin && GST_REGISTRY=foo.bin GST_DEBUG='*soup*:6' GST_PLUGIN_PATH=gst-plugins-good/lib/gstreamer-1.0:gst-plugins-bad/lib/gstreamer-1.0 result-gst-2-bin/bin/gst-inspect-1.0 adaptivedemux2
0:00:00.016612208 92549 0x60000048a1f0 DEBUG    adaptivedemux2-soup gstsouploader.c:190:gst_soup_load_library: Trying all libsoups
0:00:00.018981792 92549 0x60000048a1f0 DEBUG    adaptivedemux2-soup gstsouploader.c:229:gst_soup_load_library: Loaded /nix/store/fq4a5slwr3m9vyprw6cwhfy4hq0b1ns7-libsoup-3.4.4/lib/libsoup-3.0.0.dylib
0:00:00.023456417 92549 0x60000048a1f0 DEBUG              souputils gstsoupelement.c:38:soup_element_init: binding text domain gst-plugins-good-1.0 to locale dir /nix/store/6vinyky0acyww6a767xq7n1w7zzxm5m9-gst-plugins-good-1.24.2/share/locale
0:00:00.023546292 92549 0x60000048a1f0 DEBUG                   soup gstsouploader.c:190:gst_soup_load_library: Trying all libsoups
0:00:00.023550083 92549 0x60000048a1f0 DEBUG                   soup gstsouploader.c:229:gst_soup_load_library: Loaded /nix/store/fq4a5slwr3m9vyprw6cwhfy4hq0b1ns7-libsoup-3.4.4/lib/libsoup-3.0.0.dylib
Plugin Details:
  Name                     adaptivedemux2
  Description              Adaptive Streaming 2 plugin
  Filename                 gst-plugins-good/lib/gstreamer-1.0/libgstadaptivedemux2.dylib
  Version                  1.24.2
  License                  LGPL
  Source module            gst-plugins-good
  Documentation            https://gstreamer.freedesktop.org/documentation/adaptivedemux2/
  Source release date      2024-04-09
  Binary package           GStreamer Good Plug-ins source release
  Origin URL               Unknown package origin

  dashdemux2: DASH Demuxer
  hlsdemux2: HLS Demuxer
  mssdemux2: Smooth Streaming demuxer (v2)

  3 features:
  +-- 3 elements

```

paparodeo: i thought i was mixing up scanning vs loading and i'd need to look at the code, thus the edit. the reason to check for null is just because the number of loaded objects can decrease even if gstreamer objects only increase.

if protecting against the race when the libsoup module gets unloaded we're not going to be able to avoid the case where we are reading the image path string and the page it is on gets unmapped and we segfault but could limit the damage from reading a garbage string. we also can't do anything if a different libsoup gets loaded after the detection code and before the final g_module_open call.

a couple of minor modifications with your current approach will protect against loading a bad module.

1. perform the strdup before the basename / strcmp. this way the path name will not change from under us
2. if the name matches do a `dlopen(image_path, RTLD_NOW | RTLD_NOLOAD)` to verify that the path we got points to a loaded object -- this mainly is to protect against somehow getting a path that looks like `/evil/location/libsoup-2.4.1.dylib`.
3. maybe rather than using strdup do a malloc / strlcpy so if we are reading garbage we don't end up allocating way too much -- this is somewhat paranoid.

this is my hacked version. it uses memcmp rather than basename just cause of testcode not using glib but should get the gist.

<details><summary>somewhat-resiliant-dlopen</summary>

```c
bool
somewhat_resilant_thing(void)
{
    _Static_assert(sizeof LIBSOUP_3_SONAME > sizeof(char *), "LIBSOUP_3_SONAME is not a char[]");
    _Static_assert(sizeof LIBSOUP_2_SONAME > sizeof(char *), "LIBSOUP_2_SONAME is not a char[]");

    Dl_info info = { 0 };
    char *image_path = NULL;
    libsoup_sonames[0] = /* "@nixLibSoup3Path@/" */ LIBSOUP_3_SONAME;
    libsoup_sonames[1] = LIBSOUP_2_SONAME;

    if (dladdr(dlsym(RTLD_DEFAULT, "soup_get_major_version"), &info) && info.dli_fname != NULL) {
        // excessively large path
        const size_t bufSize = 4096;
        image_path = malloc(bufSize);
        if (image_path == NULL) {
            return false;
        }
        // copy path (could segfault if info.dli_fname ends up getting unmapped
        strlcpy(image_path, info.dli_fname, bufSize);

        const struct {
            const char * const name;
            const size_t size;
        } libs[] = {
            { LIBSOUP_3_SONAME, sizeof LIBSOUP_3_SONAME },
            { LIBSOUP_2_SONAME, sizeof LIBSOUP_2_SONAME },
        };

        for (size_t i = 0; i < sizeof libs / sizeof *libs; i++) {
            // can switch to basename / strcmp here rather than using sizes and memcmp
            const char * const end = strchr(image_path, '\0');
            const size_t pathSize  = end - image_path;
            if (pathSize >= libs[i].size
                && memcmp(end - (libs[i].size - 1), libs[i].name, libs[i].size) == 0) {
                //  ok. now that basename(inage_path) == LIBSOUP_X_SONAME
                // verify the image is still loaded to verify it is valid
                // and doesn't point somewhere evil and contains the key symbol
                void *hndl = dlopen(image_path, RTLD_NOW | RTLD_NOLOAD);
                if (hndl != NULL && dlsym(hndl, "soup_get_major_version") != NULL) {
                    libsoup_sonames[0] = image_path;
                }
                if (hndl != NULL) {
                    dlclose(hndl);
                }
                break;
            }
        }
    }
    printf("soname 0 = %s\n", libsoup_sonames[0]);
    printf("soname 1 = %s\n", libsoup_sonames[1]);
    free(image_path);
    return true;
}
```
</details>

paparodeo: > See the latest force push. I’m pretty sure that’s a long-standing glib bug with `g_module_open` on Darwin. I included a fix for that and cleaned up the probing. Darwin now loads libsoup without issue regardless of the order of folders specified on `GST_PLUGIN_PATH`. Notice how it uses libsoup2 when gst-plugins-bad is first and libsoup3 (with `RTLD_LOCAL`) when gst-plugins-good is first.

was it crashing / giving an error when using 0 rather than `RTLD_LOCAL`? i'm not clear what's happening with gst-inspect, why does it load so many things?  on linux i looked at the loaded modules `pldd` when gst-play is streaming

<details><summary>pldd `pgrep gst-play`|sort</summary>

```
/nix/store/fjqvwg4hxszzna6znkran8zqcj3h6a6a-gst-plugins-base-1.24.2/bin/gst-play-1.0
linux-vdso.so.1
/nix/store/05n35k0p8l969kfskxqhbqhy8il3a3cq-util-linux-minimal-2.40.1-lib/lib/libblkid.so.1
/nix/store/05n35k0p8l969kfskxqhbqhy8il3a3cq-util-linux-minimal-2.40.1-lib/lib/libmount.so.1
/nix/store/0vcrqaqf5sk158xy77mcl52374c1dlbi-keyutils-1.6.3-lib/lib/libkeyutils.so.1
/nix/store/2vqgx0si98m8c3yd2fhspmikavf3nvmd-libpsl-0.21.5/lib/libpsl.so.5
/nix/store/36kr91f8s509z2r8hqvpayw5bqyqgyzr-curl-8.7.1/lib/libcurl.so.4
/nix/store/5frkz5kp1p3i6wfgcjgbba38pypg4app-libselinux-3.6/lib/libselinux.so.1
/nix/store/5pcy995yarjy7bfdswgkfggbzg48dlq0-libkrb5-1.21.2/lib/libcom_err.so.3
/nix/store/5pcy995yarjy7bfdswgkfggbzg48dlq0-libkrb5-1.21.2/lib/libgssapi_krb5.so.2
/nix/store/5pcy995yarjy7bfdswgkfggbzg48dlq0-libkrb5-1.21.2/lib/libk5crypto.so.3
/nix/store/5pcy995yarjy7bfdswgkfggbzg48dlq0-libkrb5-1.21.2/lib/libkrb5.so.3
/nix/store/5pcy995yarjy7bfdswgkfggbzg48dlq0-libkrb5-1.21.2/lib/libkrb5support.so.0
/nix/store/5rplqz6dwgfhrfazyr4bv45j14iddzlb-orc-0.4.38/lib/liborc-0.4.so.0
/nix/store/771mz0x86qzhlphd22awf3s601mhwpkv-p11-kit-0.25.3/lib/libp11-kit.so.0
/nix/store/7i6p08gsvpblaj962h469fadlc080b7b-gst-plugins-good-1.24.2/lib/gstreamer-1.0/libgstautodetect.so
/nix/store/7i6p08gsvpblaj962h469fadlc080b7b-gst-plugins-good-1.24.2/lib/gstreamer-1.0/libgstpulseaudio.so
/nix/store/7i6p08gsvpblaj962h469fadlc080b7b-gst-plugins-good-1.24.2/lib/gstreamer-1.0/libgstsoup.so
/nix/store/7v85vgw01j852y0i1s4hgdjkda4d0561-libproxy-0.5.6/lib/libproxy/libpxbackend-1.0.so
/nix/store/7v85vgw01j852y0i1s4hgdjkda4d0561-libproxy-0.5.6/lib/libproxy.so.1
/nix/store/a1vdbxij9xn9i2qs695rlcn3jkwpldd2-flac-1.4.3/lib/libFLAC.so.12
/nix/store/bp74x3bkh8979gxz72wc9hq3jb0hdaq7-systemd-minimal-255.6/lib/libsystemd.so.0
/nix/store/by853zqacqy647v0z7l70xm77xl9ifal-zstd-1.5.6/lib/libzstd.so.1
/nix/store/c2v6ycn0sjcpx9ww8x7j4ima6xnpssry-glib-2.80.2/lib/libgio-2.0.so.0
/nix/store/c2v6ycn0sjcpx9ww8x7j4ima6xnpssry-glib-2.80.2/lib/libglib-2.0.so.0
/nix/store/c2v6ycn0sjcpx9ww8x7j4ima6xnpssry-glib-2.80.2/lib/libgmodule-2.0.so.0
/nix/store/c2v6ycn0sjcpx9ww8x7j4ima6xnpssry-glib-2.80.2/lib/libgobject-2.0.so.0
/nix/store/cfgpxq64il1sfsjqbblghds82jf2a185-nettle-3.9.1/lib/libhogweed.so.6
/nix/store/cfgpxq64il1sfsjqbblghds82jf2a185-nettle-3.9.1/lib/libnettle.so.8
/nix/store/d1g6mz24zwx4pcl1wfwja6lgbf8x0m8d-libvorbis-1.3.7/lib/libvorbisenc.so.2
/nix/store/d1g6mz24zwx4pcl1wfwja6lgbf8x0m8d-libvorbis-1.3.7/lib/libvorbis.so.0
/nix/store/d857vkjzr4rf68ycyfkcpdiax6v4d3c0-libtasn1-4.19.0/lib/libtasn1.so.6
/nix/store/ds2h2v0frms447yhg9cgyk78mrp5240a-libmpg123-1.32.6/lib/libmpg123.so.0
/nix/store/fjqvwg4hxszzna6znkran8zqcj3h6a6a-gst-plugins-base-1.24.2/lib/gstreamer-1.0/libgstaudioconvert.so
/nix/store/fjqvwg4hxszzna6znkran8zqcj3h6a6a-gst-plugins-base-1.24.2/lib/gstreamer-1.0/libgstaudioresample.so
/nix/store/fjqvwg4hxszzna6znkran8zqcj3h6a6a-gst-plugins-base-1.24.2/lib/gstreamer-1.0/libgstogg.so
/nix/store/fjqvwg4hxszzna6znkran8zqcj3h6a6a-gst-plugins-base-1.24.2/lib/gstreamer-1.0/libgstopus.so
/nix/store/fjqvwg4hxszzna6znkran8zqcj3h6a6a-gst-plugins-base-1.24.2/lib/gstreamer-1.0/libgstplayback.so
/nix/store/fjqvwg4hxszzna6znkran8zqcj3h6a6a-gst-plugins-base-1.24.2/lib/gstreamer-1.0/libgsttypefindfunctions.so
/nix/store/fjqvwg4hxszzna6znkran8zqcj3h6a6a-gst-plugins-base-1.24.2/lib/gstreamer-1.0/libgstvolume.so
/nix/store/fjqvwg4hxszzna6znkran8zqcj3h6a6a-gst-plugins-base-1.24.2/lib/libgstaudio-1.0.so.0
/nix/store/fjqvwg4hxszzna6znkran8zqcj3h6a6a-gst-plugins-base-1.24.2/lib/libgstpbutils-1.0.so.0
/nix/store/fjqvwg4hxszzna6znkran8zqcj3h6a6a-gst-plugins-base-1.24.2/lib/libgstriff-1.0.so.0
/nix/store/fjqvwg4hxszzna6znkran8zqcj3h6a6a-gst-plugins-base-1.24.2/lib/libgsttag-1.0.so.0
/nix/store/fjqvwg4hxszzna6znkran8zqcj3h6a6a-gst-plugins-base-1.24.2/lib/libgstvideo-1.0.so.0
/nix/store/g19314860ijffhbksw25pl2rcagk7wcx-dconf-0.40.0-lib/lib/gio/modules/libdconfsettings.so
/nix/store/g3vi60zgyjsvij7xkk6dxky1hkwh0ynd-pcre2-10.43/lib/libpcre2-8.so.0
/nix/store/gp504m4dvw5k2pdx6pccf1km79fkcwgf-openssl-3.0.13/lib/libcrypto.so.3
/nix/store/gp504m4dvw5k2pdx6pccf1km79fkcwgf-openssl-3.0.13/lib/libssl.so.3
/nix/store/ha8i7q6xf4794idzrzh35vsycb5lmjzn-libpulseaudio-17.0/lib/libpulse.so.0
/nix/store/ha8i7q6xf4794idzrzh35vsycb5lmjzn-libpulseaudio-17.0/lib/pulseaudio/libpulsecommon-17.0.so
/nix/store/ic63ay0py10fyryaw7345k4ps32da33w-libidn2-2.3.7/lib/libidn2.so.0
/nix/store/jd41k79l3nxq4b7b7yvc0kmcjd3lq7sa-dbus-1.14.10-lib/lib/libdbus-1.so.3
/nix/store/jpjmh2325b6k3svvsmhj57wqlph9srdx-brotli-1.1.0-lib/lib/libbrotlicommon.so.1
/nix/store/jpjmh2325b6k3svvsmhj57wqlph9srdx-brotli-1.1.0-lib/lib/libbrotlidec.so.1
/nix/store/k7zgvzp2r31zkg9xqgjim7mbknryv6bs-glibc-2.39-52/lib/ld-linux-x86-64.so.2
/nix/store/k7zgvzp2r31zkg9xqgjim7mbknryv6bs-glibc-2.39-52/lib/libc.so.6
/nix/store/k7zgvzp2r31zkg9xqgjim7mbknryv6bs-glibc-2.39-52/lib/libdl.so.2
/nix/store/k7zgvzp2r31zkg9xqgjim7mbknryv6bs-glibc-2.39-52/lib/libm.so.6
/nix/store/k7zgvzp2r31zkg9xqgjim7mbknryv6bs-glibc-2.39-52/lib/libmvec.so.1
/nix/store/k7zgvzp2r31zkg9xqgjim7mbknryv6bs-glibc-2.39-52/lib/libpthread.so.0
/nix/store/k7zgvzp2r31zkg9xqgjim7mbknryv6bs-glibc-2.39-52/lib/libresolv.so.2
/nix/store/kbg5m7fyi1w23fyfmxjhhzcbd577rpg0-libffi-3.4.6/lib/libffi.so.8
/nix/store/kw1nc4sg6hp550z8g8pdjb3d3ixz4m3l-nghttp2-1.61.0-lib/lib/libnghttp2.so.14
/nix/store/l738rl0x75xhna5qy2g2nww6i8nwzacf-lame-3.100-lib/lib/libmp3lame.so.0
/nix/store/lfpszg9pqbl4g5gwif16kq5sshv1qcd2-libogg-1.3.5/lib/libogg.so.0
/nix/store/lphbn1va4i43fj7f3m9xskf9y86khzf3-xz-5.4.6/lib/liblzma.so.5
/nix/store/lv6nackqis28gg7l2ic43f6nk52hb39g-zlib-1.3.1/lib/libz.so.1
/nix/store/mxg79hhk4dgv0lnifhj6k9jgk7xrkcsd-libopus-1.5.2/lib/libopus.so.0
/nix/store/n90qk04jw71dk42g7c1ifwdswmbpm3ml-gstreamer-1.24.2/lib/gstreamer-1.0/libgstcoreelements.so
/nix/store/n90qk04jw71dk42g7c1ifwdswmbpm3ml-gstreamer-1.24.2/lib/libgstbase-1.0.so.0
/nix/store/n90qk04jw71dk42g7c1ifwdswmbpm3ml-gstreamer-1.24.2/lib/libgstreamer-1.0.so.0
/nix/store/npb9hyshsbnfwx89jfa1ykrb18ncmw7s-libsoup-3.4.4/lib/libsoup-3.0.so.0
/nix/store/r20kl3djvc3r5w5a0gwy4d896mn6b22a-elfutils-0.191/lib/libdw.so.1
/nix/store/r20kl3djvc3r5w5a0gwy4d896mn6b22a-elfutils-0.191/lib/libelf.so.1
/nix/store/r2jjxkphbxfyrawn8qvyc6r2wrkyvjwn-libunwind-1.8.1/lib/libunwind.so.8
/nix/store/rv8x9609d5nh5z8v7v9gmhy6z1wnrhkk-libsndfile-1.2.2/lib/libsndfile.so.1
/nix/store/rx6nkd40819acppajq29g1hxa4d9r35f-gmp-with-cxx-6.3.0/lib/libgmp.so.10
/nix/store/ryp6f4v7xfcjsdrwn9dy7ppz2ida360l-glib-networking-2.80.0/lib/gio/modules/libgiognomeproxy.so
/nix/store/ryp6f4v7xfcjsdrwn9dy7ppz2ida360l-glib-networking-2.80.0/lib/gio/modules/libgiognutls.so
/nix/store/ryp6f4v7xfcjsdrwn9dy7ppz2ida360l-glib-networking-2.80.0/lib/gio/modules/libgiolibproxy.so
/nix/store/vrqy1mn0nw6g6y0w22gipbspknnwr9q5-libssh2-1.11.0/lib/libssh2.so.1
/nix/store/wf7w1vbis4fphasyn4z3vad6336rxc64-gnutls-3.8.5/lib/libgnutls.so.30
/nix/store/wxhfv9zglm74a05k54n34pxnsz6g7r7w-duktape-2.7.0/lib/libduktape.so.207
/nix/store/xvzz97yk73hw03v5dhhz3j47ggwf1yq1-gcc-13.2.0-lib/lib/libgcc_s.so.1
/nix/store/xwcf1rw3ackqp2vxms9myf9jq2ny6ynv-bzip2-1.0.8/lib/libbz2.so.1
/nix/store/y5p181lnbh8rrwd3ljp83k0cn3xklyk2-sqlite-3.45.3/lib/libsqlite3.so.0
/nix/store/yfp7dr8m7zi7kxk49wd714gwvhb105hf-libunistring-1.1/lib/libunistring.so.5
/nix/store/yvhyhcfhc98wm86pw4ygk5jdr804iwrw-libcap-2.69-lib/lib/libcap.so.2
```
</details>

<details><summary>make-wrapper</summary>

```
makeCWrapper '/nix/store/9sq9438rcks6z82ih4ixl0swjr3waj67-gstreamer-env/bin/.gst-play-1.0-wrapped' \
    --inherit-argv0 \
    --prefix 'GIO_EXTRA_MODULES' ':' '/nix/store/g19314860ijffhbksw25pl2rcagk7wcx-dconf-0.40.0-lib/lib/gio/modules' \
    --prefix 'GIO_EXTRA_MODULES' ':' '/nix/store/ryp6f4v7xfcjsdrwn9dy7ppz2ida360l-glib-networking-2.80.0/lib/gio/modules' \
    --prefix 'GIO_EXTRA_MODULES' ':' '/nix/store/g19314860ijffhbksw25pl2rcagk7wcx-dconf-0.40.0-lib/lib/gio/modules' \
    --prefix 'GIO_EXTRA_MODULES' ':' '/nix/store/ryp6f4v7xfcjsdrwn9dy7ppz2ida360l-glib-networking-2.80.0/lib/gio/modules' \
    --prefix 'XDG_DATA_DIRS' ':' '/nix/store/9sq9438rcks6z82ih4ixl0swjr3waj67-gstreamer-env/share' \
    --prefix 'GST_PLUGIN_SYSTEM_PATH_1_0' ':' '/nix/store/n90qk04jw71dk42g7c1ifwdswmbpm3ml-gstreamer-1.24.2/lib/gstreamer-1.0:/nix/store/fjqvwg4hxszzna6znkran8zqcj3h6a6a-gst-plugins-base-1.24.2/lib/gstreamer-1.0:/nix/store/v6l1g6q0jf726b92c6jq1wm408ga5wdb-gst-plugins-bad-1.24.2/lib/gstreamer-1.0:/nix/store/rdxfllwr6ls191lwxn0y7yh07v769k4x-gst-rtsp-server-1.24.2/lib/gstreamer-1.0:/nix/store/g78hf1fsrw74bxna1p1djb8xbw0a8jvq-gst-devtools-1.24.2/lib/gstreamer-1.0:/nix/store/xicyhvid38bfq5n43cb7j8xarpvqhnkn-gst-editing-services-1.24.2/lib/gstreamer-1.0:/nix/store/8pdhws5vkgzvyzs9idmjxls51lyaabg8-gst-libav-1.24.2/lib/gstreamer-1.0:/nix/store/7i6p08gsvpblaj962h469fadlc080b7b-gst-plugins-good-1.24.2/lib/gstreamer-1.0:/nix/store/2799bnwagv54mh12fsvk4q5ij4fbys78-gst-plugins-rs-0.12.4/lib/gstreamer-1.0:/nix/store/yghx4yqip2nn69r81fj07x5wcagjrnsg-gst-plugins-ugly-1.24.2/lib/gstreamer-1.0:/nix/store/v62vssssnrizyr861y994jfrn7r7s7kw-gst-plugins-viperfx-unstable-2020-9-20/lib/gstreamer-1.0:/nix/store/ncfq5fyddc2nb5mldvj9hw4lc1mafd73-gstreamer-vaapi-1.24.2/lib/gstreamer-1.0'
```
</details>

reckenrode: I incorporated some of the feedback. I switched to using [`g_strndup`](https://docs.gtk.org/glib/func.strndup.html) with a limit of `PATH_MAX - 1` instead of using `malloc`. I also added the `dlopen` and `dlsym` checks. I chose `soup_uri_new` as the symbol to check because that’s what libsoup checks.

I also updated the formatting to follow upstream’s style. Once the glib bug is fixed upstream, I’d like to try upstreaming this as well since it should make loading libsoup work reliably on Darwin.

reckenrode: > was it crashing / giving an error when using 0 rather than RTLD_LOCAL? i'm not clear what's happening with gst-inspect, why does it load so many things? on linux i looked at the loaded modules pldd when gst-play is streaming

The gst-inspect command is the test case from the upstream [issue](https://gitlab.freedesktop.org/gstreamer/gstreamer/-/issues/1171) that prompted them to restrict loading libsoup to Linux. It doesn’t cause a crash, though it breaks Wine, but it causes libsoup to issue the following error.

```
(gst-plugin-scanner:65699): libsoup-ERROR **: 15:11:40.167: libsoup3 symbols detected. Using libsoup2 and libsoup3 in the same process is not supported.
```

While working on a fix, I noticed that whether the error triggers on Darwin is dependent on the load order of plugins in `GST_PLUGIN_PATH`. That’s why I initially had it defaulting to libsoup 2.4 on Darwin, which ended up being a workaround for the actual problem (see below). Note that this does not happen on Linux (also see below).

* If gst-bad-plugins is loaded first, which includes the webrtc plugin that indirectly links against libsoup 2.4 via libnice, then libsoup 2.4 will be detected by the soup loader. When the adaptivedemux2 and soup plugins are loaded, they will use libsoup 2.4 instead of libsoup 3.
* If gst-good-plugins is loaded first, which includes the adaptivedemux2 and soup plugins, then libsoup will raise the error when webrtc is loaded because adaptivedemux and soup did not detect libsoup 2.4 and defaulted to libsoup 3. Because of the Glib bug, this libsoup 3 is loaded globally when it should be loaded locally.

Linux does not experience this problem because `0` is `RTLD_LOCAL`. Darwin defines `RTLD_LOCAL` as `0x4`, so when it gets `0`, it defaults to `RTLD_GLOBAL`, which is the opposite of what’s wanted by GStreamer. Linux can default to libsoup3 because the symbols it loads are not global. Darwin (without the fix) breaks because they are.

In essence, [`G_MODULE_BIND_LOCAL`](https://docs.gtk.org/gmodule/flags.ModuleFlags.html#local) does not work on Darwin. GStreamer tries to use it to load libsoup, but it ends up accidentally loading libsoup globally instead. The fix is trivial, but I’d like to go through a staging cycle and see if anything breaks once Glib starts working correctly on Darwin before trying to upstream the fix.

paparodeo: >     * If gst-good-plugins is loaded first, then libsoup will raise the error when webrtc is loaded because adaptivedemux and soup did not detect libsoup 2.4 and defaulted to libsoup 3. Due to the Glib bug, this libsoup 3 is loaded globally when it should be loaded locally.

got it, it think -- thanks.  so libsoup-2.4 is linked in at compile time and isn't loading using g_module_open then, so RTLD_LOCAL matters. and now that makes sense that you're using the g_find_symbol so it is run in the context of the gmodule.dylib.

seems like a good bugfix. nice work!

reckenrode: > so libsoup-2.4 is linked in at compile time and isn't loading using g_module_open then

Right. Plugins (and applications) can link directly to libsoup, so those need to work. The `g_find_symbol` check I’m doing is more or less how libsoup does its check. If it fails to detect anything, then libsoup shouldn’t either.

paparodeo: > I chose `soup_uri_new` as the symbol to check because that’s what libsoup checks.

i think `soup_uri_new` is only in libsoup-2 https://gitlab.freedesktop.org/gstreamer/gstreamer/-/blob/main/subprojects/gst-plugins-good/ext/soup/gstsouploader.c?ref_type=heads#L199 (it is part of versioned symbols). also not detected on libsoup-3 looking at `nm` output.

reckenrode: You’re right. libsoup2 and libsoup3 check different symbols. I’ll change it back to checking `soup_get_major_version`.

* libsoup2: https://gitlab.gnome.org/GNOME/libsoup/-/blob/0f6c259f5b98dacdffc1e4a35edbba354f5d4518/libsoup/soup-init.c#L30
* libsoup3: https://gitlab.gnome.org/GNOME/libsoup/-/blob/8afa87883e71fc589b46431df320d1a818433a2f/libsoup/soup-init.c#L30

paparodeo (approved):
change looks good to me sans merge conflict. tested on master for x64 linux / darwin and seems good. minor nit would be to zero initialize `info = {0};` and `func = NULL;` for paranoia.

<details><summary>tested on x64 linux/darwin (linux output shown below)</summary>

```console
$ nix-build -I nixpkgs=. - <<\ENV
let  
  inherit (import <nixpkgs> {})
    buildEnv
    glib-networking
    gst_all_1
    wrapGAppsNoGuiHook
    ;
in
buildEnv rec {
  name = "gstreamer-env";
  paths = with gst_all_1; [
    gstreamer
    gst-plugins-base
    gst-plugins-bad
    gst-plugins-good
  ];
  nativeBuildInputs = [ wrapGAppsNoGuiHook ];
  buildInputs = paths ++ [ glib-networking ];
  pathsToLink = [ "/bin" "/share/man" ];
  postBuild = ''
    rm -- $out/bin/.*
    gappsWrapperArgsHook
    wrapGAppsHook
  '';
}
ENV
$ rm -rf ~/.cache/gstream* ; GST_DEBUG='*soup*:6' result/bin/gst-inspect-1.0 adaptivedemux2
0:00:00.131779254 2205998      0x2629af0 DEBUG    adaptivedemux2-soup gstsouploader.c:216:gst_soup_load_library: LibSoup 2 found
0:00:00.131958380 2205998      0x2629af0 DEBUG    adaptivedemux2-soup gstsouploader.c:238:gst_soup_load_library: Loaded libsoup-2.4.so.1
0:00:00.137940138 2205998      0x2629af0 DEBUG              souputils gstsoupelement.c:37:soup_element_init: binding text domain gst-plugins-good-1.0 to locale dir /nix/store/iwx4012karfxfc18laafdzpym613rky1-gst-plugins-good-1.24.2/share/locale
0:00:00.138040111 2205998      0x2629af0 DEBUG                   soup gstsouploader.c:216:gst_soup_load_library: LibSoup 2 found
0:00:00.138049501 2205998      0x2629af0 DEBUG                   soup gstsouploader.c:238:gst_soup_load_library: Loaded libsoup-2.4.so.1
Plugin Details:
  Name                     adaptivedemux2
  Description              Adaptive Streaming 2 plugin
  Filename                 /nix/store/iwx4012karfxfc18laafdzpym613rky1-gst-plugins-good-1.24.2/lib/gstreamer-1.0/libgstadaptivedemux2.so
  Version                  1.24.2
  License                  LGPL
  Source module            gst-plugins-good
  Documentation            https://gstreamer.freedesktop.org/documentation/adaptivedemux2/
  Source release date      2024-04-09
  Binary package           GStreamer Good Plug-ins source release
  Origin URL               Unknown package origin

  dashdemux2: DASH Demuxer
  hlsdemux2: HLS Demuxer
  mssdemux2: Smooth Streaming demuxer (v2)

  3 features:
  +-- 3 elements
$ ./result/bin/gst-play-1.0 "https://ia803206.us.archive.org/9/items/odnt_20200806/02. Marlena Shaw - California Soul.mp3"
Press 'k' to see a list of keyboard shortcuts.
Now playing https://ia803206.us.archive.org/9/items/odnt_20200806/02. Marlena Shaw - California Soul.mp3
Prerolling...
Redistribute latency...
Redistribute latency...
0:00:45.2 / 0:02:57.6       

```

</details>

reckenrode:
> change looks good to me sans merge conflict. tested on master for x64 linux / darwin and seems good. minor nit would be to zero initialize `info = {0};` and `func = NULL;` for paranoia.

I fixed the conflict and updated the patch to zero-initialize `info` and `func`.

reckenrode:
I also added the backport label since this issue would affect 24.05, and it’s just released.

paparodeo (approved):

paparodeo:
broken eval seems due to https://github.com/NixOS/nixpkgs/pull/318447 and fix by 6d0ba08b970248d207addab47dfd09c0dc3b0895

reckenrode:
@ofborg eval

paparodeo:
> I also added the backport label since this issue would affect 24.05, and it’s just released.

label needs to target  staging-release-24.05 (or whatever the staging label is named)

pkgs/development/libraries/glib/gmodule-rtld_local.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/gmodule/gmodule-dl.c b/gmodule/gmodule-dl.c
+index 6d1c5fab7..b21773bca 100644
+--- a/gmodule/gmodule-dl.c
++++ b/gmodule/gmodule-dl.c
+@@ -136,7 +136,7 @@ _g_module_open (const gchar *file_name,
+   
+   lock_dlerror ();
+   handle = dlopen (file_name,
+-		   (bind_local ? 0 : RTLD_GLOBAL) | (bind_lazy ? RTLD_LAZY : RTLD_NOW));
++		   (bind_local ? RTLD_LOCAL : RTLD_GLOBAL) | (bind_lazy ? RTLD_LAZY : RTLD_NOW));
```
jtojnar, line 10: This should really be upstreamed. Looks okay to me at a glance but I am not confident about possible effects.

pkgs/development/libraries/gstreamer/good/souploader.diff
```diff
@@ -1,8 +1,99 @@
...
++#endif
+ 
+ #define LOAD_SYMBOL(name) G_STMT_START {                                \
+     if (!g_module_symbol (module, G_STRINGIFY (name), (gpointer *) &G_PASTE (vtable->_, name))) { \
+@@ -156,7 +162,46 @@ gst_soup_load_library (void)
+   GST_DEBUG_CATEGORY_INIT (gst_soup_debug, "soup", 0, "soup");
+ #endif
+ 
+-#ifdef HAVE_RTLD_NOLOAD
++#if defined(__APPLE__)
```
jtojnar, line 53: I am not really comfortable with large patches like this in-tree. Is it possible to split it into an upstreamable part and keep only a small patch that substitutes path?

pkgs/development/libraries/glib/gmodule-rtld_local.patch
reckenrode: I agree, but I want to go through a staging cycle before doing that given that it’s a big semantic change (because I think, “What does this break?” is an obvious question, and having a successful cycle helps answer that).

pkgs/development/libraries/gstreamer/good/souploader.diff
reckenrode: Upstream removed support for dynamically loading libsoup on non-Linux platforms under the assumption only Linux distributions ship both libsoup 2 and libsoup 3. While that’s true (more or less) of Homebrew and MacPorts, nixpkgs is still shipping both even on Darwin.

I plan to try to upstream this patch after the Glib patch is merged, but there’s no guarantee they’ll accept it.

> Is it possible to split it into an upstreamable part and keep only a small patch that substitutes path?

Just to make sure I’m understanding correctly: keep `souploader.diff` as-is and provide a separate patch with the changes I plan to upstream (after a staging cycle and the Glib patch is merged)?

jtojnar: > Just to make sure I’m understanding correctly: keep `souploader.diff` as-is (with the substitution patch) and provide a separate patch with the changes I plan to upstream (after the Glib patch is merged)?

Yes.

reckenrode: Sorry about the delay, but I split out the Darwin-specific change to a separate patch in the latest force push.

reckenrode:
@lilyinstarlight It was suggested I ping you for a review. This fixed GStreamer on Darwin with plugins that use libsoup.

pkgs/development/libraries/gstreamer/good/souploader.diff
jtojnar: I still only see a single patch.

lilyinstarlight (approved):
Finally getting a chance to look at this, I'm broadly okay with the change and it does appear like it should work around (at least for nixpkgs) the issue outlined at <https://gitlab.freedesktop.org/gstreamer/gstreamer/-/issues/1171#note_2290789>

I will echo @jtojnar that the gst-plugins-good patch should probably be split into one for the functionality (that could at least be submitted upstream for review) and another for the substitution to embed an absolute fallback path. But I'm not going to personally block this PR on that

pkgs/development/libraries/gstreamer/good/souploader.diff
reckenrode: Ah, right. I was counting the existing patch for Linux as the second patch, but there also needs to be a separate one for Darwin. I’ll fix that.

reckenrode:
@jtojnar @lilyinstarlight I updated the PR to isolate the absolute path patch to `souploader.diff`. Please let me know if I missed anything else or need to make any other changes.

paparodeo:
FYI: the backport tag still needs to be changed from `backport release-24.05` -> `backport staging-24.05`

reckenrode:
> FYI: the backport tag still needs to be changed from `backport release-24.05` -> `backport staging-24.05`

Thanks! I’m going to merge this, so it makes the next staging cycle. I’m at a gaming convention this weekend, but I can start working on the upstreaming stuff next week.

reckenrode:
I created https://gitlab.gnome.org/GNOME/glib/-/merge_requests/4154 to upstream the GLib patch.

PR #317746, opened 2024-06-06
sbcl: make deriv options overridable attributes

hraban:
## Description of changes

This also makes the options available for reading by other derivations which depend on SBCL. This is useful for derivations who want to evaluate differently based on the capabilities of the SBCL derivation they're passed. Eexample: bordeaux-threads depends on a specific feature of SBCL (threads): having the `.threadSupport` value readable from the derivation at runtime allows bordeaux-thread to set `meta.broken = true` for an SBCL without thread support.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [x] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/development/compilers/sbcl/default.nix
+++ b/pkgs/development/compilers/sbcl/default.nix
@@ -1,13 +1,4 @@
 { lib, stdenv, callPackage, ecl, coreutils, fetchurl, strace, texinfo, which, writeText, zstd
-, threadSupport ? (stdenv.hostPlatform.isx86 || "aarch64-linux" == stdenv.hostPlatform.system || "aarch64-darwin" == stdenv.hostPlatform.system)
-, linkableRuntime ? stdenv.hostPlatform.isx86
-, disableImmobileSpace ? false
-  # Meant for sbcl used for creating binaries portable to non-NixOS via save-lisp-and-die.
-  # Note that the created binaries still need `patchelf --set-interpreter ...`
-  # to get rid of ${glibc} dependency.
-, purgeNixReferences ? false
-, coreCompression ? true
-, markRegionGC ? threadSupport
 , version
   # Set this to a lisp binary to use a custom bootstrap lisp compiler for SBCL.
   # Leave as null to use the default.  This is useful for local development of
@@ -65,12 +56,14 @@ let
 
 in
 
-stdenv.mkDerivation (self: rec {
+stdenv.mkDerivation (self: {
   pname = "sbcl";
   inherit version;
 
   src = fetchurl {
-    url = "mirror://sourceforge/project/sbcl/sbcl/${version}/${pname}-${version}-source.tar.bz2";
+    # Changing the version shouldn’t change the source for the
+    # derivation. Override the src entirely if desired.
+    url = "mirror://sourceforge/project/sbcl/sbcl/${version}/sbcl-${version}-source.tar.bz2";
     inherit (versionMap.${version}) sha256;
   };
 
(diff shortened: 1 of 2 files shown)


kini (approved):
LGTM, thanks!

Uthar:
Hmm - I would put the options in passthru - this is how the Emacs package does it for checking "withNativeCompilation".

hraban:
> Hmm - I would put the options in passthru - this is how the Emacs package does it for checking "withNativeCompilation".

Interesting. Any particular reason? Passthru isn't available to the actual derivation in the end which makes sense for non-fundamental types but I'm curious what the motivation is for putting regular flags on passthru. Maybe I'm personally a bit too jaded by the emacs derivation to take much stock in it as prior art 😂 I looked around at other derivations but I have a feeling it's quite haphazard across the board.

Uthar:
It's only my taste :-)

We could make all assertions at the start (kind of like check-type/declare in lisp) - which feels more "organized" - assertions are separate from build logic. For example:

```
{

...

# Build options  
, threadSupport ? (stdenv.hostPlatform.isx86 || "aarch64-linux" == stdenv.hostPlatform.system || "aarch64-darwin" == stdenv.hostPlatform.system)
, linkableRuntime ? stdenv.hostPlatform.isx86
, disableImmobileSpace ? false
, purgeNixReferences ? false
, coreCompression ? true
, markRegionGC ? threadSupport
}:

assert coreCompression -> !purgeNixReferences;
assert markRegionGC -> threadSupport;

let ...
```

Similarly - build-time env variables inside mkDerivation  are separate from build options. These options can be used to decide on those flags/vars, but themselves are "metadata". Then it's more immediately visible, what is build-internal and what is "exported". (in passthru attrset).

Uthar:
SO perhaps a reason is to keep overrideAttrs for attrs specific to mkDerivation , and use override for customizations on top of that. But it's hard to find the one true convention in nixpkgs (-:

> Feels like we want the opposite for these values, right? Definitely trigger a rebuild--that's kinda the point. Fundamentally different derivation.

I think it's for cases with `.pkgs` style sets, so as not to cause havoc. In the case if it's referenced inside mkDerivation it will re-build correctly.

hraban:
> > Feels like we want the opposite for these values, right? Definitely trigger a rebuild--that's kinda the point. Fundamentally different derivation.
> 
> I think it's for cases with `.pkgs` style sets, so as not to cause havoc. In the case if it's referenced inside mkDerivation it will re-build correctly.

That's a totally fair point!

Re the assertions sprinkled throughout: I've kept them close to the point of use so they don't trip people up who want to override them, e.g.:

https://github.com/NixOS/nixpkgs/blob/1acef6e636e4111933dfd0213ee20ffaafe997c7/pkgs/development/compilers/sbcl/default.nix#L86-L93

I thought this would strike a nice balance between failing loud, but leaving the ultimate decision with the user. I like the idea of leaving the user in charge if they insist on using incompatible flags, e.g. to test a build which does in fact support that combination. What do you think?

Re the derivation: I think I see why you prefer passthru. I realize I see derivations as more bespoke and case-by-case, where `mkDerivation` (and on a lower level `stdenv`) is just a framework _on top_ of which you build your derivation. I also thought that any value stored as a direct property on the derivation itself (i.e. passed as a env to the builder) would also be available after an `import my-derivation.drvPath`, which was already abstruse enough but also turns out to be entirely wrong. So there goes that.

I guess all I have to go off of is a "feeling" that it's nice to have the flags which control the build process live directly on the derivation rather than incidentally in an `override` tacked on to the object which represents the derivation. It's a very weak intuition so I'm happy to go your way!

Uthar:
ah - I see - yes like this the asserts can be overridden out, unlike on top-level - quite nice.

>  I also thought that any value stored as a direct property on the derivation itself (i.e. passed as a env to the builder) would also be available after an import my-derivation.drvPath

Theoretically one could `builtins.readFile x.drvPath` and parse it - I don't know about "pure evaluation mode" though. (One might get dark side points for this).

Hmm a little bit out of scope - but regarding the zstd case - for linux we could allow compression + purge by using a static libzstd.a - I don't know about macos.

> I guess all I have to go off of is a "feeling" that it's nice to have the flags which control the build process live directly on the derivation 

heh, they say one should follow their gut feeling - go for it (-:

hraban:
@7c6f434c what are your thoughts? Ready to merge?

7c6f434c:
I am honestly a bit confused that `mkDerivation` now allows this! But yeah, any other option is prone to forgetting to inherit something…

PR #317772, opened 2024-06-06
Make NixOS/nixpkgs Flake pure

RossComputerGuy:
## Description of changes

Discovered with Garnix, the checks are executed on push and it is currently impure. This PR makes the Flake pure for `nix flake check`.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [x] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -44,13 +44,14 @@
           );
       });
 
-      checks.x86_64-linux = {
-        tarball = jobs.x86_64-linux.tarball;
+      checks = forAllSystems (system: {
+        tarball = jobs.${system}.tarball;
+      } // lib.optionalAttrs (self.legacyPackages.${system}.stdenv.isLinux) {
         # Test that ensures that the nixosSystem function can accept a lib argument
         # Note: prefer not to extend or modify `lib`, especially if you want to share reusable modules
         #       alternatives include: `import` a file, or put a custom library in an option or in `_module.args.<libname>`
         nixosSystemAcceptsLib = (self.lib.nixosSystem {
-          pkgs = self.legacyPackages.x86_64-linux;
+          pkgs = self.legacyPackages.${system};
           lib = self.lib.extend (final: prev: {
             ifThisFunctionIsMissingTheTestFails = final.id;
           });
@@ -66,13 +67,13 @@
             })
           ];
         }).config.system.build.toplevel;
-      };
+      });
 
       htmlDocs = {
-        nixpkgsManual = jobs.x86_64-linux.manual;
+        nixpkgsManual = builtins.mapAttrs (_: jobSet: jobSet.manual) jobs;
         nixosManual = (import ./nixos/release-small.nix {
           nixpkgs = self;
-        }).nixos.manual.x86_64-linux;
+        }).nixos.manual;
       };
 
       # The "legacy" in `legacyPackages` doesn't imply that the packages exposed
--- a/flake.nix
+++ b/flake.nix
@@ -5,14 +5,15 @@
 
   outputs = { self }:
     let
-      jobs = import ./pkgs/top-level/release.nix {
-        nixpkgs = self;
-      };
-
       libVersionInfoOverlay = import ./lib/flake-version-info.nix self;
       lib = (import ./lib).extend libVersionInfoOverlay;
 
       forAllSystems = lib.genAttrs lib.systems.flakeExposed;
+
+      jobs = forAllSystems (system: import ./pkgs/top-level/release.nix {
+        nixpkgs = self;
+        inherit system;
+      });
     in
     {
       lib = lib.extend (final: prev: {
@@ -44,11 +45,12 @@
       });
 
       checks.x86_64-linux = {
-        tarball = jobs.tarball;
+        tarball = jobs.x86_64-linux.tarball;
         # Test that ensures that the nixosSystem function can accept a lib argument
         # Note: prefer not to extend or modify `lib`, especially if you want to share reusable modules
         #       alternatives include: `import` a file, or put a custom library in an option or in `_module.args.<libname>`
         nixosSystemAcceptsLib = (self.lib.nixosSystem {
+          pkgs = self.legacyPackages.x86_64-linux;
           lib = self.lib.extend (final: prev: {
             ifThisFunctionIsMissingTheTestFails = final.id;
           });
@@ -67,7 +69,7 @@
       };
 
       htmlDocs = {
-        nixpkgsManual = jobs.manual;
+        nixpkgsManual = jobs.x86_64-linux.manual;
         nixosManual = (import ./nixos/release-small.nix {
           nixpkgs = self;
         }).nixos.manual.x86_64-linux;
--- a/lib/tests/release.nix
+++ b/lib/tests/release.nix
@@ -24,7 +24,10 @@ in
       #
       #   https://github.com/NixOS/nixpkgs/issues/272591
       #
-      [(import ../../pkgs/test/release {})]
+      [(import ../../pkgs/test/release {
+        inherit pkgs;
+        pkgs-path = pkgs.path;
+      })]
     ;
 
   }
--- a/lib/tests/release.nix
+++ b/lib/tests/release.nix
@@ -25,8 +25,7 @@ in
       #   https://github.com/NixOS/nixpkgs/issues/272591
       #
       [(import ../../pkgs/test/release {
-        inherit pkgs;
-        pkgs-path = pkgs.path;
+        inherit pkgs lib nix;
       })]
     ;
 
--- a/pkgs/top-level/release-lib.nix
+++ b/pkgs/top-level/release-lib.nix
@@ -1,4 +1,5 @@
 { supportedSystems
+, system ? builtins.currentSystem
 , packageSet ? (import ../..)
 , scrubJobs ? true
 , # Attributes passed to nixpkgs. Don't build packages marked as unfree.
@@ -33,7 +34,7 @@ let
     systems
     ;
 
-  pkgs = packageSet (recursiveUpdate { system = "x86_64-linux"; config.allowUnsupportedSystem = true; } nixpkgsArgs);
+  pkgs = packageSet (recursiveUpdate { inherit system; config.allowUnsupportedSystem = true; } nixpkgsArgs);
 
   hydraJob' = if scrubJobs then hydraJob else id;
 
--- a/pkgs/top-level/release.nix
+++ b/pkgs/top-level/release.nix
@@ -9,6 +9,7 @@
    $ nix-build pkgs/top-level/release.nix -A coreutils.x86_64-linux
 */
 { nixpkgs ? { outPath = (import ../../lib).cleanSource ../..; revCount = 1234; shortRev = "abcdef"; revision = "0000000000000000000000000000000000000000"; }
+, system ? builtins.currentSystem
 , officialRelease ? false
   # The platform doubles for which we build Nixpkgs.
 , supportedSystems ? [ "x86_64-linux" "x86_64-darwin" "aarch64-linux" "aarch64-darwin" ]
@@ -54,7 +55,7 @@
 
 let
   release-lib = import ./release-lib.nix {
-    inherit supportedSystems scrubJobs nixpkgsArgs;
+    inherit supportedSystems scrubJobs nixpkgsArgs system;
   };
 
   inherit (release-lib) mapTestOn pkgs;


drupol (approved):
LGTM! I like the removal of the `system` hardcoded value and abstract it away.

pkgs/top-level/release-lib.nix
```diff
@@ -33,7 +34,7 @@ let
     systems
     ;
 
-  pkgs = packageSet (recursiveUpdate { system = "x86_64-linux"; config.allowUnsupportedSystem = true; } nixpkgsArgs);
+  pkgs = packageSet (recursiveUpdate { inherit system; config.allowUnsupportedSystem = true; } nixpkgsArgs);
```
infinisil, line 37: This and the other change in this file doesn't seem to be making it purer, but rather makes the system not hardcoded. I guess it wasn't possible to run this on non-x86_64-linux systems before?

infinisil (commented):
Is there a command that can be used to test this?

RossComputerGuy:
> Is there a command that can be used to test this?

`nix flake check`. Though I found this issue with Garnix. Looking at the Garnix logs with this PR, the checks pass.

pkgs/top-level/release-lib.nix
RossComputerGuy: Before, `builtins.currentSystem` was being used in certain places. This PR passes system around and uses `forAllSystems` so multiple platforms can do the release jobs and such.

RossComputerGuy: It is pure changes because we're passing system instead of inferring from the builtins.

infinisil: But the value here didn't come from `builtins.currentSystem`, it was hardcoded before, which is technically also pure. So the change here doesn't make it any more pure, but rather just avoids the hardcoding.

RossComputerGuy: Yeah, I just used `inherit system;` since the system would already need to be passed to this file.

flake.nix
```diff
@@ -5,10 +5,6 @@
 
   outputs = { self }:
     let
-      jobs = import ./pkgs/top-level/release.nix {
-        nixpkgs = self;
-      };
```
infinisil, line 10: You're now evaluating the release.nix file twice in the flake. Something like

```nix
jobs = forAllSystems (system: import ./pkgs/top-level/release.nix {
  nixpkgs = self;
  inherit system;
});
```

And then

```nix
checks = forAllSystems (system: {
  tarball = jobs.${system}.tarball
# ...
```

avoids that :)

RossComputerGuy: Woops, well that is fixed. I did the same for the other occurrence.

pkgs/top-level/release-lib.nix
infinisil: It looks very sensible, but I think I'm still not understanding a piece of the puzzle: Why was this hardcoded before without ever causing problems? I can't test it locally right now, but I presume `nix flake check` on e.g. Darwin fails if this is left hardcoded? Did nobody on Darwin ever evaluate `release-lib.nix`, or is the `pkgs` it imports never really used perhaps? Perhaps it's hardcoded for a reason, such that you get reproducible results on all systems? Looking at the git blame might be useful

Would be great to figure this out, especially because nobody seems to really know the code here well!

RossComputerGuy: Yeah, it was originally written before I joined in so I am not sure why it was never an issue before.

pkgs/top-level/release.nix
```diff
@@ -8,7 +8,16 @@
...
-{ nixpkgs ? { outPath = (import ../../lib).cleanSource ../..; revCount = 1234; shortRev = "abcdef"; revision = "0000000000000000000000000000000000000000"; }
+{ nixpkgs ? {
+  outPath = (import ../../lib).cleanSource ../..;
+  revCount = 1234;
+  shortRev = "abcdef";
+  revision = "0000000000000000000000000000000000000000";
+  legacyPackages.${system} = import ../.. {
+    inherit system;
+  };
+}
```
infinisil, line 19: This change doesn't seem to be necessary anymore (I see you used it only for an intermediate commit)

RossComputerGuy: It's necessary to evaluate with the `nix-build` tool. I added `legacyPackages` to help retain compatibility.

infinisil: Not sure what you mean, I just tested `nix-build pkgs/top-level/release.nix -A manual` and it works fine without this change (removing the `legacyPackages`). It doesn't appear to be used anywhere

RossComputerGuy: Also, I'm pretty sure Hydra uses `nix-build` instead of Flakes.

RossComputerGuy: > Not sure what you mean, I just tested `nix-build pkgs/top-level/release.nix -A manual` and it works fine without this change (removing the `legacyPackages`). It doesn't appear to be used anywhere

Huh, must've been a leftover change when I was working on trying to get things working.

RossComputerGuy: Oh wait, it is necessary. `release-lib` takes `packageSet` which is using `legacyPackages`.

infinisil: So what fails without it exactly? Because neither the above command nor `nix flake check` fail for me

RossComputerGuy: I have no idea. It might not be necessary? Kinda hard to tell with how `release-lib` and `release` works. I'll remove it and see what happens. Could be like what I mentioned earlier, leftovers from trying to fix stuff.

pkgs/top-level/release-lib.nix
infinisil: At least [this commit](https://github.com/NixOS/nixpkgs/commit/3e9f76774ab03c0aa06267dcfa1c7d7f6c984cf2) gives a bit of context

pkgs/top-level/release.nix
```diff
@@ -55,6 +56,7 @@
 let
   release-lib = import ./release-lib.nix {
     inherit supportedSystems scrubJobs nixpkgsArgs;
+    packageSet = nixpkgs.legacyPackages.${system};
```
infinisil: This intermediate commit is not correct because of this (which you're undoing again in the next commit).

RossComputerGuy: Yeah, I'm trying to remove that but git problems.

RossComputerGuy: Fixed

pkgs/top-level/release.nix
RossComputerGuy: Removed the change.

pkgs/top-level/release-lib.nix
RossComputerGuy: Interesting. I wonder why specifically `x86_64-linux` was chosen as the "base system". Probably because not many people use ARM? I'm on aarch64 so at least this PR makes this work for me.

infinisil: For reference, what failure with which command are you getting on aarch64 before this PR?

RossComputerGuy: I would get this:
```
error:
       … while checking flake output 'checks'

         at /nix/store/nryiv97qvy45sm4pxr0kz42rkq74bd42-source/flake.nix:46:7:

           45|
           46|       checks.x86_64-linux = {
             |       ^
           47|         tarball = jobs.tarball;

       … while checking the derivation 'checks.x86_64-linux.tarball'

         at /nix/store/nryiv97qvy45sm4pxr0kz42rkq74bd42-source/flake.nix:47:9:

           46|       checks.x86_64-linux = {
           47|         tarball = jobs.tarball;
             |         ^
           48|         # Test that ensures that the nixosSystem function can accept a lib argument

       error: attribute 'currentSystem' missing

       at /nix/store/nryiv97qvy45sm4pxr0kz42rkq74bd42-source/pkgs/top-level/impure.nix:17:43:

           16|   # (build, in GNU Autotools parlance) platform.
           17|   localSystem ? { system = args.system or builtins.currentSystem; }
             |                                           ^
           18|
```

infinisil: No I mean if you leave the hardcoded `x86_64-linux` value here

RossComputerGuy:
```
error:
       … while checking flake output 'checks'

         at /nix/store/k4bwf1ilnd60pv7vz294ghgrzmazm1am-source/flake.nix:47:7:

           46|
           47|       checks = forAllSystems (system: {
             |       ^
           48|         inherit (jobs.${system}) tarball;

       … while checking the derivation 'checks.x86_64-darwin.nixosSystemAcceptsLib'

         at /nix/store/k4bwf1ilnd60pv7vz294ghgrzmazm1am-source/flake.nix:52:9:

           51|         #       alternatives include: `import` a file, or put a custom library in an option or in `_module.args.<libname>`
           52|         nixosSystemAcceptsLib = (self.lib.nixosSystem {
             |         ^
           53|           pkgs = self.legacyPackages.${system};

       error: assertion '(stdenv).isLinux' failed

       at /nix/store/k4bwf1ilnd60pv7vz294ghgrzmazm1am-source/pkgs/os-specific/linux/kernel/generic.nix:87:1:

           86|
           87| assert stdenv.isLinux;
             | ^
           88|
```

infinisil: Oh so you're on `aarch64-darwin` I guess. But `nix flake check --system aarch64-darwin` fails evaluation no matter whether the value here is hardcoded to `x86_64-linux` or not (tested it locally). I'm looking for a command that didn't work previously, but now does because it's not hardcoded anymore.

RossComputerGuy: No, I'm on `aarch64-linux`. With `nix flake check` and this PR, the checks pass.

RossComputerGuy: But it also makes it where the checks that are in the Flake get executed on Garnix since I use Garnix for CI. Before, I would get the `builtins.currentSystem` missing error and now there is no error.

infinisil: I feel like I'm missing something. I can't reproduce your eval failures as you describe them. [This failure](https://github.com/NixOS/nixpkgs/pull/317772#discussion_r1630569831) I can only get with `--system aarch64-darwin`, but you're saying you're on `aarch64-linux`?

And I'm specifically asking whether removing the _hardcoded_ part here actually helps making any command work. Please try this:

```
❯ nix-env --version
nix-env (Nix) 2.18.2

❯ nix flake check github:ExpidusOS/nixpkgs/fix/flake-check --no-build --system aarch64-linux
warning: unknown flake output 'htmlDocs'
warning: The check omitted these incompatible systems: aarch64-darwin, armv5tel-linux, armv6l-linux, armv7l-linux, i686-linux, mipsel-linux, powerpc64le-linux, riscv64-linux, x86_64-darwin, x86_64-linux
Use '--all-systems' to check all.

❯ nix flake check github:infinisil/nixpkgs/hardcode --no-build --system aarch64-linux
warning: unknown flake output 'htmlDocs'
warning: The check omitted these incompatible systems: aarch64-darwin, armv5tel-linux, armv6l-linux, armv7l-linux, i686-linux, mipsel-linux, powerpc64le-linux, riscv64-linux, x86_64-darwin, x86_64-linux
Use '--all-systems' to check all.
```

The last command uses the hardcoded `x86_64-linux` on top of your branch: https://github.com/infinisil/nixpkgs/commit/1af88e1976e5e08f86b775686c46825515d2d9d3. And evaluation works just fine with it with `--system aarch64-linux`. So unless you noticed something else that was broken, it doesn't seem to be necessary to remove the hardcoding.

RossComputerGuy: > And I'm specifically asking whether removing the _hardcoded_ part here actually helps making any command work.

Both commands seem to work until Darwin fails with the `stdenv.isLinux` assert fails. I've pushed a fix to not include the NixOS stuff outside on Darwin. With the hardcoded value, it tries calling a binfmt I have set up for x86. On NixOS on Apple Silicon, getting x86 emulation working is a pain. Making it `inherit system` makes the build work for the Flake check.

> So unless you noticed something else that was broken, it doesn't seem to be necessary to remove the hardcoded value.

Why do we even need hardcoded values in the first place? Imo, it's better to abstract that out to `builtins.currentSystem` or whatever system-specific attribute set something is a part of.

RossComputerGuy: Since my system has bintfmt set up, a better look into this issue is on a system without binfmt that is also ARM. Using the hardcode, I get this error:
```
error: a 'x86_64-linux' with features {} is required to build '/nix/store/lmx4147d1l51j1qvwhy7wh3kll5ssyq7-maintainer-teams.json.drv', but I am a 'aarch64-linux' with features {benchmark, big-parallel, gccarch-armv8-a, kvm, nixos-test}
```

Upstream only does:
```
warning: The check omitted these incompatible systems: aarch64-darwin, armv5tel-linux, armv6l-linux, armv7l-linux, i686-linux, mipsel-linux, powerpc64le-linux, riscv64-linux, x86_64-darwin, x86_64-linux
Use '--all-systems' to check all.
```

While with my PR builds the checks.

infinisil: > With the hardcoded value (and removing --all-systems), it tries calling a binfmt I have set up for x86

Ah that's great to know, that's what I'm trying to get at. So the hardcoded value _does_ end up causing problem on Darwin. If it wouldn't have caused problems, removing it would be barking up the wrong tree.

I think this is a separate enough change that it deserves its own commit, with the commit message saying that it fixes building tests on non-x86_64-linux systems.

RossComputerGuy: > _does_ end up causing problem on Darwin

Yeah, on any system outside `x86_64-linux`.

RossComputerGuy: > I think removing the hardcoded value and the `optionalAttrs` is a separate enough change that it deserves its own commit, with the commit message saying that it fixes building tests on non-x86_64-linux systems.

Would that be too little of changes for their own commits?

infinisil: It definitely doesn't fit into any "make pure" commit, because it's arguably the complete opposite of that, since it removes a hardcoded value and replacing it with something that could be different on everybody's system. So yeah if you don't mind, I think it really deserves its own commit.

infinisil:
Gotta sleep now, but please also test
```
$ nix-build lib/tests/release.nix
```

I can't run it myself right now, but I suspect it's broken, and CI wouldn't catch that.

pkgs/top-level/release-lib.nix
RossComputerGuy: Done

RossComputerGuy:
> Gotta sleep now, but please also test
> 
> ```
> $ nix-build lib/tests/release.nix
> ```
> 
> I can't run it myself right now, but I suspect it's broken, and CI wouldn't catch that.

It very much so is:
```
error: pkgs.lib accessed, but the lib tests should use nixpkgs' lib path directly!
```

RossComputerGuy:
I fixed `lib/tests/release.nix`

lib/tests/release.nix
```diff
@@ -24,7 +24,10 @@ in
       #
       #   https://github.com/NixOS/nixpkgs/issues/272591
       #
-      [(import ../../pkgs/test/release {})]
+      [(import ../../pkgs/test/release {
+        inherit pkgs lib;
+        pkgs-path = pkgs.path;
+      })]
```
infinisil, line 30: I don't think `pkgs-path` has to be set anymore (and `pkgs.path` is a bit of an odd-ball anyways). Furthermore, I'm just seeing that @lf- opened https://github.com/NixOS/nixpkgs/pull/318654 which also changes this part, notably also setting Nix, which I think is good, so:

```suggestion
      [(import ../../pkgs/test/release {
        inherit pkgs lib nix;
      })]
```

RossComputerGuy: Alright, removed `pkgs-path`.

flake.nix
```diff
@@ -64,13 +67,13 @@
...
+      });
 
       htmlDocs = {
-        nixpkgsManual = jobs.manual;
+        nixpkgsManual = builtins.mapAttrs (_: jobSet: jobSet.manual) jobs;
         nixosManual = (import ./nixos/release-small.nix {
           nixpkgs = self;
-        }).nixos.manual.x86_64-linux;
+        }).nixos.manual;
```
infinisil, line 76: Note that before this PR, `nixpkgsManual`/`nixosManual` were derivations (hardcoded to `x86_64-linux`), but now they're an attribute set of derivations, so it would break for any tooling that doesn't expect that. We're getting bitten by Flakes poor system handling here :sweat_smile:

In this particular case though I don't think it's a problem: Flakes doesn't specify any inherent stability promises to output attributes, and these attributes are not even mentioned anywhere in Nixpkgs. If this Flake attribute was consumed by automated tooling for example, we'd have a problem.

So, no change necessary here imo, but just wanted to point it out, because it's very easy to break consumers with this kind of thing.

lf-:
Thanks for submitting this! We have been [suffering in Lix](https://gerrit.lix.systems/c/lix/+/1429) because of the regression this fixes, so I'm very excited for it to get fixed. It will need a backport to 24.05 once you're done :)

infinisil (approved):
Thanks for being so responsive, looking good to me now!

flake.nix
RossComputerGuy: Yeah, I'm not sure why we weren't already having these attributes be attribute sets of derivations before. My guess is most people use `x86_64-linux` so why bother for others systems. I mainly wanted to fix the issue I had with this causing failed jobs on Garnix with forks. But now this opens the door to using these things outside of `x86_64-linux` and not rely on binfmt. Really is the best way to go even if compatibility has to be broken. More people are going to transition to non-x86 systems (including myself) in the future so having proper support is key.

infinisil:
Oh, looks like you didn't include the passing of `nix` as I suggested in https://github.com/NixOS/nixpkgs/pull/317772#discussion_r1632464811, but let's merge for now and improve as necessary with future PRs

RossComputerGuy:
> Oh, looks like you didn't include the passing of `nix`

Woops, didn't catch that. I added it anyway :smile_cat:

RossComputerGuy:
> It will need a backport to 24.05 once you're done :)

I added the backport tag so this will be backported immediately after merge.

lf- (approved):
Yep, this looks to me like it makes things overall uniformly less busted. Thanks!

infinisil:
Although I'm not sure about backporting because of https://github.com/NixOS/nixpkgs/pull/317772#discussion_r1632475729. Everything else sounds good to backport though.

lf-:
> Although I'm not sure about backporting because of [#317772 (comment)](https://github.com/NixOS/nixpkgs/pull/317772#discussion_r1632475729). Everything else sounds good to backport though.

Hm, fair. I wouldn't personally care, because our flake outputs *definitely* should not be stable API, and anyhow, it is certainly in the window of time where people are very much actively migrating and so I would have very little shame about just doing it.

RossComputerGuy:
> Although I'm not sure about backporting because of [#317772 (comment)](https://github.com/NixOS/nixpkgs/pull/317772#discussion_r1632475729).

We can let the backport PR be made and do any checks with CI and whatnot to ensure it is capable of handling this change and then merge.

flake.nix
infinisil: Ohh, we can do this to keep `nix build .#htmlDocs.nixpkgsManual` working:

```nix
      packages = forAllSystems (system: {
        htmlDocs = {
          nixpkgsManual = jobs.${system}.manual;
          nixosManual = (import ./nixos/release-small.nix {
            nixpkgs = self;
          }).nixos.manual.${system};
        };
      });
```

infinisil (changes requested):
> our flake outputs _definitely_ should not be stable API

`legacyPackages`, `lib` and `nixosModules` definitely need to be considered stable, but yeah not so much the others. (Side note: I think flakes has contributed to poorer understanding of stable interface surface for Nix code)

> We can let the backport PR be made and do any checks with CI and whatnot to ensure it is capable of handling this change and then merge.

I'm not worried about CI, that won't be a problem for it. I'm worried about people doing something like `nix build github:NixOS/nixpkgs/nixos-24.05#htmlDocs.nixpkgsManual`, which would break if https://github.com/NixOS/nixpkgs/pull/317772#discussion_r1632475729 was backported. Even if we don't hear from anybody depending on this, and even if there's nobody at all depending on this, I think this is a great opportunity to preach interface stability.

My general thoughts: If an interface isn't explicitly marked as internal, we should assume it's public by default and therefore should be kept stable.

---

But, having written this, I also found a way to not break stability :)

RossComputerGuy:
> > 
> 
> I'm not worried about CI, that won't be a problem for it. I'm worried about people doing something like `nix build github:NixOS/nixpkgs/nixos-24.05#htmlDocs.nixpkgsManual`

Is it possible to edit a backport PR? We could always revert the specific change and then let people change to the new way with the next NixOS release.

flake.nix
RossComputerGuy: Does `packages.${system}` even accept multi-level attribute sets? I haven't seen anyone done it so I am not sure.

RossComputerGuy: Just tried it and nope, I get an error (`error: expected a derivation`) so I don't think making `htmlDocs` inside of `packages` would be a good solution unless we remove the `htmlDocs` attribute set and just make it `packages.${system}.nix{pkgs,os}Manual`.

lf-:
> > > 
> > 
> > 
> > I'm not worried about CI, that won't be a problem for it. I'm worried about people doing something like `nix build github:NixOS/nixpkgs/nixos-24.05#htmlDocs.nixpkgsManual`
> 
> Is it possible to edit a backport PR? We could always revert the specific change and then let people change to the new way with the next NixOS release.

Committers have such powers, but I would just do a manual cherry-pick backport.

flake.nix
infinisil: Oh damn, sorry I apparently wasn't very careful when testing :sweat_smile:

infinisil (approved):

infinisil:
Please check out the backport PR, I removed the above-mentioned changes from it

pbsds:
PSA for those who came here via `git blame`:

Use
```
nix build github:nixos/nixpkgs/refs/pull/<PR_NUMBER>/merge#htmlDocs.nixosManual.x86_64-linux
```
Instead of 
```
nix build github:nixos/nixpkgs/refs/pull/<PR_NUMBER>/merge#htmlDocs.nixosManual
```

grahamc:
It looks like this has a regression, in that `nix flake show --all-systems --json --no-write-lock-file` fails:

> error: missing bootstrap url for platform armv5te-unknown-linux-gnueabi

lf-:
I'm not convinced that exposing new attributes and then those performing an unguarded recursive evaluation encountering an error should be considered a regression. Nevertheless, do you have more details like which attribute is causing it?

RossComputerGuy:
> Nevertheless, do you have more details like which attribute is causing it?

Probably with legacy packages. Though I'm not sure why we've got `armv5te-linux` in the Flake systems of we don't have a bootstrap for it.

RossComputerGuy:
Just checked, it fails at the check.

RossComputerGuy:
@grahamc #320639 fixes the issue, I am now able to get:
```
{"checks":{"aarch64-darwin":{"tarball":{"description":"Source distribution","name":"nixpkgs-tarball-24.11pre20240617.dirty","type":"derivation"}},"aarch64-linux":{"nixosSystemAcceptsLib":{"name":"nixos-system-nixos-24.11.20240617.dirty","type":"derivation"},"tarball":{"description":"Source distribution","name":"nixpkgs-tarball-24.11pre20240617.dirty","type":"derivation"}},"armv6l-linux":{"nixosSystemAcceptsLib":{"name":"nixos-system-nixos-24.11.20240617.dirty","type":"derivation"},"tarball":{"description":"Source distribution","name":"nixpkgs-tarball-24.11pre20240617.dirty","type":"derivation"}},"armv7l-linux":{"nixosSystemAcceptsLib":{"name":"nixos-system-nixos-24.11.20240617.dirty","type":"derivation"},"tarball":{"description":"Source distribution","name":"nixpkgs-tarball-24.11pre20240617.dirty","type":"derivation"}},"i686-linux":{"nixosSystemAcceptsLib":{"name":"nixos-system-nixos-24.11.20240617.dirty","type":"derivation"},"tarball":{"description":"Source distribution","name":"nixpkgs-tarball-24.11pre20240617.dirty","type":"derivation"}},"powerpc64le-linux":{"tarball":{"description":"Source distribution","name":"nixpkgs-tarball-24.11pre20240617.dirty","type":"derivation"}},"riscv64-linux":{"nixosSystemAcceptsLib":{"name":"nixos-system-nixos-24.11.20240617.dirty","type":"derivation"},"tarball":{"description":"Source distribution","name":"nixpkgs-tarball-24.11pre20240617.dirty","type":"derivation"}},"x86_64-darwin":{"tarball":{"description":"Source distribution","name":"nixpkgs-tarball-24.11pre20240617.dirty","type":"derivation"}},"x86_64-linux":{"nixosSystemAcceptsLib":{"name":"nixos-system-nixos-24.11.20240617.dirty","type":"derivation"},"tarball":{"description":"Source distribution","name":"nixpkgs-tarball-24.11pre20240617.dirty","type":"derivation"}}},"htmlDocs":{"type":"unknown"},"legacyPackages":{"aarch64-darwin":{},"aarch64-linux":{},"armv6l-linux":{},"armv7l-linux":{},"i686-linux":{},"powerpc64le-linux":{},"riscv64-linux":{},"x86_64-darwin":{},"x86_64-linux":{}},"lib":{"type":"unknown"},"nixosModules":{"notDetected":{"type":"nixos-module"},"readOnlyPkgs":{"type":"nixos-module"}}}
```

PR #317786, opened 2024-06-06
tpm2-tss: remove shadow dependency

tie:
## Description of changes

Fixes `tpm2-tss` (cross-)build on macOS since shadow package is Linux-specific and therefore cannot be used in nativeBuildInputs.

Split from #317651

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [x] aarch64-linux
  - [ ] x86_64-darwin
  - [x] aarch64-darwin → x86_64-linux
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/development/libraries/tpm2-tss/default.nix
+++ b/pkgs/development/libraries/tpm2-tss/default.nix
@@ -2,7 +2,7 @@
 , autoreconfHook, autoconf-archive, pkg-config, doxygen, perl
 , openssl, json_c, curl, libgcrypt
 , cmocka, uthash, ibm-sw-tpm2, iproute2, procps, which
-, shadow, libuuid
+, libuuid
 }:
 let
   # Avoid a circular dependency on Linux systems (systemd depends on tpm2-tss,
@@ -28,7 +28,6 @@ stdenv.mkDerivation rec {
 
   nativeBuildInputs = [
     autoreconfHook autoconf-archive pkg-config doxygen perl
-    shadow
   ];
 
   buildInputs = [
@@ -53,6 +52,17 @@ stdenv.mkDerivation rec {
     # Do not rely on dynamic loader path
     # TCTI loader relies on dlopen(), this patch prefixes all calls with the output directory
     ./no-dynamic-loader-path.patch
+
+    # Configure script expects tools from shadow (e.g. useradd) but they are
+    # actually optional (and we can’t use them in Nix sandbox anyway). Make the
+    # check in configure.ac a warning instead of an error so that we can run
+    # configure phase on platforms that don’t have shadow package (e.g. macOS).
+    # Note that *on platforms* does not mean *for platform* i.e. this is for
+    # cross-compilation, tpm2-tss does not support macOS, see upstream issue:
+    # https://github.com/tpm2-software/tpm2-tss/issues/2629
+    # See also
+    # https://github.com/tpm2-software/tpm2-tss/blob/6c46325b466f35d40c2ed1043bfdfcfb8a367a34/Makefile.am#L880-L898
+    ./no-shadow.patch
   ];
 
   postPatch = ''
--- a/pkgs/development/libraries/tpm2-tss/no-shadow.patch
+++ b/pkgs/development/libraries/tpm2-tss/no-shadow.patch
@@ -1,0 +1,16 @@
+diff --git a/configure.ac b/configure.ac
+index e2d579b8..0eac4ff3 100644
+--- a/configure.ac
++++ b/configure.ac
+@@ -672,9 +672,9 @@ AS_IF([test "$HOSTOS" = "Linux" && test "x$systemd_sysusers" != "xyes"],
+       AC_CHECK_PROG(adduser, adduser, yes)
+       AC_CHECK_PROG(addgroup, addgroup, yes)
+       AS_IF([test "x$addgroup" != "xyes" && test "x$groupadd" != "xyes" ],
+-         [AC_MSG_ERROR([addgroup or groupadd are needed.])])
++         [AC_MSG_WARN([addgroup or groupadd are needed.])])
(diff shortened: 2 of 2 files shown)


tie:
@baloo, gentle ping. You’ve approved these changes on the original PR (https://github.com/NixOS/nixpkgs/pull/317651#pullrequestreview-2102323254), are you still OK with them?

baloo:
Ha, I missed that notification sorry. Thanks for the ping.

baloo (approved):

baloo:
Would you mind changing the `meta.platforms` as well?

tie:
>Would you mind changing the meta.platforms as well?

This PR is for building tmp2-tss for Linux on macOS, e.g.
```
(import <nixpkgs> {
  localSystem = "aarch64-darwin";
  crossSystem = "x86_64-linux";
}).tpm2-tss
```

I’m not really sure if it builds _for_ macOS as-is, I’ll try building tpm2-tss once my macOS builder finishes building another PR.

baloo:
Ha good point. Sorry I got a bit too excited haha.
I had the need to compile tpm2-tss on macos (not to interact with a TPM directly but because it's required for makecredentials) for coworkers to be able to help, but I don't have access to a macos myself.

tie:
Yes, I don’t think tpm2-tss supports macOS. It’s trivial to convince build system to actually build for macOS, but the code uses `uchar.h` header (that is not available on macOS) in a few places and I gave up at that point.

There seems to be an upstream issue for macOS support, https://github.com/tpm2-software/tpm2-tss/issues/2629

tie:
@JohnRTitor, hi, can you take a look at this please (and hopefully merge if everything is OK)?

JohnRTitor:
Thanks for the review request, however as I am not knowledgeable in this topic, I am not confident that replacing this with a dummy package won't cause any issues. 

If in future the package ends up relying more on this dependency, this change would break the package then. 

I'll let others handle this.

pkgs/development/libraries/tpm2-tss/default.nix
```diff
@@ -11,6 +11,20 @@ let
...
   procpsWithoutSystemd = procps.override { withSystemd = false; };
   procps_pkg = if stdenv.isLinux then procpsWithoutSystemd else procps;
+
+  # Configure script expects tools from shadow but we are not actually using
+  # them. Instead of patching the build scripts, just add fake executables
+  # to native build inputs. These would fail the build if called from both POSIX
+  # shell and via native exec system call (albeit for different reasons).
+  # https://github.com/tpm2-software/tpm2-tss/blob/6c46325b466f35d40c2ed1043bfdfcfb8a367a34/configure.ac#L666-L675
+  # https://github.com/tpm2-software/tpm2-tss/blob/6c46325b466f35d40c2ed1043bfdfcfb8a367a34/Makefile.am#L880-L898
+  fakeShadow = runCommand "fake-shadow" { } ''
```
jtojnar, line 21: Looks to me like it will just [fail with a warning](https://github.com/tpm2-software/tpm2-tss/blob/6c46325b466f35d40c2ed1043bfdfcfb8a367a34/Makefile.am#L988), which we should be able to safely ignore. So we should just be able to remove shadow altogether. Maybe also patch the warning out to be explicit and avoid the noise.

Or maybe even better, switch to sysusers.d.

tie: I don’t think `tpm2-tss` package runs `install-dirs` so we don’t even reach this warning. However, configure script fails if we don’t have shadow in PATH. Perhaps we can patch that instead and run autoreconfHook? 🤔
https://github.com/tpm2-software/tpm2-tss/blob/6c46325b466f35d40c2ed1043bfdfcfb8a367a34/configure.ac#L667-L675

numinit: I don't think removing these dependencies is a problem at all. Nix manages these users. If anything, this should be able to be disabled upstream.

numinit: Easier to fake:

https://github.com/tpm2-software/tpm2-tss/blob/6c46325b466f35d40c2ed1043bfdfcfb8a367a34/configure.ac#L661

Just create a noop `systemd-sysusers` script on Linux that does nothing, and it'll use that instead of shadow.

tie: Actually, we do indeed run `install-dirs` and get the warning because of the Nix sandbox.
```console
$ nix log nixpkgs/nixos-24.05#tpm2-tss | fgrep -C3 'Failed to create the tss user and group'
call make_tss_user_and_group
groupadd: Permission denied.
groupadd: cannot lock /etc/group; try again later.
WARNING Failed to create the tss user and group
(call make_fapi_dirs) && (call set_fapi_permissions)
chown: invalid user: 'tss:tss'
WARNING Failed to create the FAPI directories with the correct permissions
```

I’ve update the PR to patch configure.ac so we print a warning instead of aborting configure phase 😅
I’m not sure why I didn’t do that from the beginning.

tie:
I’ve removed the need for dummy package by converting relevant AC_MSG_ERROR to AC_MSG_WARN in configure.ac (see https://github.com/NixOS/nixpkgs/pull/317786#discussion_r1701963935). As pointed out in https://github.com/NixOS/nixpkgs/pull/317786#discussion_r1696021955, this is already a warning instead of an error at install phase, so it shouldn’t be an error at configure phase.

baloo (approved):

numinit (approved):

SuperSandro2000 (approved):

vcunat:
At a glance the patch looks like something worth trying to upstream.

numinit:
Totally agreed, it somewhat seems like a bug.

PR #289652, opened 2024-02-18
wstunnel-rust: init at 9.2.2

Aleksanaa:
## Description of changes

The Rust rewrite of [wstunnel](https://search.nixos.org/packages?channel=unstable&show=wstunnel&from=0&size=50&sort=relevance&type=packages&query=wstunnel) from the same author. The  Haskell version may no longer be maintained. The command line arguments are not compatible though, so I dare not replace it directly.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
- For non-Linux: Is sandboxing enabled in `nix.conf`? (See [Nix manual](https://nixos.org/manual/nix/stable/command-ref/conf-file.html))
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:
  - [NixOS test(s)](https://nixos.org/manual/nixos/unstable/index.html#sec-nixos-tests) (look inside [nixos/tests](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests))
  - and/or [package tests](https://nixos.org/manual/nixpkgs/unstable/#sec-package-tests)
  - or, for functions and "core" functionality, tests in [lib/tests](https://github.com/NixOS/nixpkgs/blob/master/lib/tests) or [pkgs/test](https://github.com/NixOS/nixpkgs/blob/master/pkgs/test)
  - made sure NixOS tests are [linked](https://nixos.org/manual/nixpkgs/unstable/#ssec-nixos-tests-linking) to the relevant packages
- [ ] Tested compilation of all packages that depend on this change using `nix-shell -p nixpkgs-review --run "nixpkgs-review rev HEAD"`. Note: all changes have to be committed, also see [nixpkgs-review usage](https://github.com/Mic92/nixpkgs-review#usage)
- [x] Tested basic functionality of all binary files (usually in `./result/bin/`)
- [24.05 Release Notes](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2405.section.md) (or backporting [23.05](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2305.section.md) and [23.11](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2311.section.md) Release notes)
  - [ ] (Package updates) Added a release notes entry if the change is major or breaking
[...]

PR #317840, opened 2024-06-06
wstunnel: 0.5.1.0 -> 9.6.2

NeverBehave:
## Description of changes

Add https://github.com/erebe/wstunnel , wstunnel rewritten in rust.

There is one exists in haskell but it seems the author abandon it: https://search.nixos.org/packages?channel=24.05&from=0&size=50&sort=relevance&type=packages&query=wstunnel

I renamed the package to `wstunnel-rust` for now, but maybe we should just replace it with the haskell one? @gebner

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [x] Tested, as applicable:

---

### DIFF
--- a/nixos/modules/services/networking/wstunnel.nix
+++ b/nixos/modules/services/networking/wstunnel.nix
@@ -7,6 +7,9 @@ let
     (name: value: if value == true then "--${name}" else "--${name}=${value}")
     attrs
   );
+
+  hostPortToString = { host, port }: "${host}:${builtins.toString port}";
+
   hostPortSubmodule = {
     options = {
       host = mkOption {
@@ -19,28 +22,7 @@ let
       };
     };
   };
-  localRemoteSubmodule = {
-    options = {
-      local = mkOption {
-        description = "Local address and port to listen on.";
-        type = types.submodule hostPortSubmodule;
-        example = {
-          host = "127.0.0.1";
-          port = 51820;
-        };
-      };
-      remote = mkOption {
-        description = "Address and port on remote to forward traffic to.";
-        type = types.submodule hostPortSubmodule;
-        example = {
-          host = "127.0.0.1";
-          port = 51820;
-        };
-      };
-    };
-  };
-  hostPortToString = { host, port }: "${host}:${builtins.toString port}";
-  localRemoteToString = { local, remote }: utils.escapeSystemdExecArg "${hostPortToString local}:${hostPortToString remote}";
+
   commonOptions = {
     enable = mkOption {
       description = "Whether to enable this `wstunnel` instance.";
@@ -66,10 +48,16 @@ let
       };
     };
 
-    verboseLogging = mkOption {
-      description = "Enable verbose logging.";
-      type = types.bool;
-      default = false;
+    loggingLevel = mkOption {
+      description = ''
+        Passed to --log-lvl
+
+        Control the log verbosity. i.e: TRACE, DEBUG, INFO, WARN, ERROR, OFF
+        For more details, checkout [EnvFilter](https://docs.rs/tracing-subscriber/latest/tracing_subscriber/filter/struct.EnvFilter.html#example-syntax)
+      '';
+      type = types.nullOr types.str;
+      example = "INFO";
+      default = null;
     };
 
     environmentFile = mkOption {
@@ -99,11 +87,12 @@ let
 
       restrictTo = mkOption {
         description = "Accepted traffic will be forwarded only to this service. Set to `null` to allow forwarding to arbitrary addresses.";
-        type = types.nullOr (types.submodule hostPortSubmodule);
-        example = {
+        type = types.listOf (types.submodule hostPortSubmodule);
+        default = [];
+        example = [{
           host = "127.0.0.1";
           port = 51820;
-        };
+        }];
       };
 
       enableHTTPS = mkOption {
@@ -134,59 +123,36 @@ let
       };
     };
   };
+
   clientSubmodule = { config, ... }: {
     options = commonOptions // {
       connectTo = mkOption {
         description = "Server address and port to connect to.";
-        type = types.submodule hostPortSubmodule;
-        example = {
-          host = "example.com";
-        };
-      };
-
-      enableHTTPS = mkOption {
-        description = "Enable HTTPS when connecting to the server.";
-        type = types.bool;
-        default = true;
+        type = types.str;
+        example = "https://wstunnel.server.com:8443";
       };
 
       localToRemote = mkOption {
-        description = "Local hosts and ports to listen on, plus the hosts and ports on remote to forward traffic to. Setting a local port to a value less than 1024 will additionally give the process the required CAP_NET_BIND_SERVICE capability.";
-        type = types.listOf (types.submodule localRemoteSubmodule);
+        description = ''Listen on local and forwards traffic from remote.'';
+        type = types.listOf (types.str);
         default = [];
-        example = [ {
-          local = {
-            host = "127.0.0.1";
-            port = 8080;
-          };
-          remote = {
-            host = "127.0.0.1";
-            port = 8080;
-          };
-        } ];
+        example = [
+          "tcp://1212:google.com:443"
+          "unix:///tmp/wstunnel.sock:g.com:443"
+        ];
       };
 
-      dynamicToRemote = mkOption {
-        description = "Host and port for the SOCKS5 proxy to dynamically forward traffic to. Leave this at `null` to disable the SOCKS5 proxy. Setting the port to a value less than 1024 will additionally give the service the required CAP_NET_BIND_SERVICE capability.";
-        type = types.nullOr (types.submodule hostPortSubmodule);
-        default = null;
-        example = {
-          host = "127.0.0.1";
-          port = 1080;
-        };
+      remoteToLocal = mkOption {
+        description = "Listen on remote and forwards traffic from local. Only tcp is supported";
+        type = types.listOf (types.str);
+        default = [];
+        example = [
+          "tcp://1212:google.com:443"
+          "unix://wstunnel.sock:g.com:443"
+        ];
       };
 
-      udp = mkOption {
-        description = "Whether to forward UDP instead of TCP traffic.";
-        type = types.bool;
-        default = false;
-      };
-
-      udpTimeout = mkOption {
-        description = "When using UDP forwarding, timeout in seconds after which the tunnel connection is closed. `-1` means no timeout.";
-        type = types.int;
-        default = 30;
-      };
+      addNetBind = mkEnableOption "Whether add CAP_NET_BIND_SERVICE to the tunnel service, this should be enabled if you want to bind port < 1024";
 
       httpProxy = mkOption {
         description = ''
@@ -214,12 +180,6 @@ let
         example = "wstunnel";
       };
 
-      hostHeader = mkOption {
-        description = "Use this as the HTTP host header instead of the real hostname. Useful for circumventing hostname-based firewalls.";
-        type = types.nullOr types.str;
-        default = null;
-      };
-
       tlsSNI = mkOption {
         description = "Use this as the SNI while connecting via TLS. Useful for circumventing hostname-based firewalls.";
         type = types.nullOr types.str;
@@ -234,7 +194,7 @@ let
 
       # The original argument name `websocketPingFrequency` is a misnomer, as the frequency is the inverse of the interval.
       websocketPingInterval = mkOption {
-        description = "Do a heartbeat ping every N seconds to keep up the websocket connection.";
+        description = "Frequency at which the client will send websocket ping to the server.";
         type = types.nullOr types.ints.unsigned;
         default = null;
       };
@@ -261,6 +221,7 @@ let
       };
     };
   };
+
   generateServerUnit = name: serverCfg: {
     name = "wstunnel-server-${name}";
     value = {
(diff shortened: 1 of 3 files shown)


SuperSandro2000:
Please follow the [contributing guide](https://github.com/NixOS/nixpkgs/blob/master/CONTRIBUTING.md#commit-conventions) when naming your commits.

pkgs/by-name/ws/wstunnel-rust/package.nix
```diff
@@ -0,0 +1,32 @@
...
+  version = "9.6.1";
+
+  src = fetchFromGitHub {
+    owner = "erebe";
+    repo  = "wstunnel";
+    rev = "v${version}";
+    sha256 = "sha256-+tNQf53KUNIz9RFVTHqgXEKLxlXiQKKPZtXqD3C/t7I=";
+  };
+
+  cargoHash = "sha256-NER7NIWUXyWFUGcdZ5HLFaKaO9FNLPlG7jql08karS4=";
```
SuperSandro2000, line 17:
```suggestion
  cargoHash = "sha256-NER7NIWUXyWFUGcdZ5HLFaKaO9FNLPlG7jql08karS4=";

```

pkgs/by-name/ws/wstunnel-rust/package.nix
```diff
@@ -0,0 +1,32 @@
...
+  meta = with lib; {
+    description = "Wstunnel uses the websocket protocol which is compatible with http in order to bypass firewalls and proxies. Wstunnel allows you to tunnel whatever traffic you want and access whatever resources/site you need.";
+    mainProgram = "wstunnel";
+    homepage    = "https://github.com/erebe/wstunnel";
+    license     = with licenses; [ bsd3Clear ];
+    maintainers = with maintainers; [ neverbehave ];
+    platforms   = platforms.linux;
+  };
+}
+
```
SuperSandro2000, line 32: there is one new line to much here

pkgs/by-name/ws/wstunnel-rust/package.nix
```diff
@@ -0,0 +1,32 @@
...
+  };
+
+  cargoHash = "sha256-NER7NIWUXyWFUGcdZ5HLFaKaO9FNLPlG7jql08karS4=";
+  checkFlags = [
+    # make use of network connection
+    "--skip=tcp::tests::test_proxy_connection"
+  ];
+
+  meta = with lib; {
+    description = "Wstunnel uses the websocket protocol which is compatible with http in order to bypass firewalls and proxies. Wstunnel allows you to tunnel whatever traffic you want and access whatever resources/site you need.";
```
SuperSandro2000, line 24:
```suggestion
    description = "Wstunnel uses the websocket protocol in order to bypass firewalls and proxies";
```
we neeed to shorten that or move some things to longDescription.

NeverBehave:
I will keep the name `wstunnel-rust` for now to avoid confusion with haskell package.

pkgs/by-name/ws/wstunnel-rust/package.nix
```diff
@@ -0,0 +1,32 @@
+{ fetchFromGitHub
+, rustPlatform
+, lib
+}:
+
+rustPlatform.buildRustPackage rec {
+  pname = "wstunnel-rust";
```
NickCao, line 7: `pname` should be identical to upstream package name, and can be different from the attribute name, so `wstunnel` is fine.

pkgs/by-name/ws/wstunnel-rust/package.nix
```diff
@@ -0,0 +1,32 @@
...
+  version = "9.6.1";
+
+  src = fetchFromGitHub {
+    owner = "erebe";
+    repo  = "wstunnel";
+    rev = "v${version}";
+    sha256 = "sha256-+tNQf53KUNIz9RFVTHqgXEKLxlXiQKKPZtXqD3C/t7I=";
```
NickCao, line 14:
```suggestion
    hash = "sha256-+tNQf53KUNIz9RFVTHqgXEKLxlXiQKKPZtXqD3C/t7I=";
```

pkgs/by-name/ws/wstunnel-rust/package.nix
```diff
@@ -0,0 +1,32 @@
...
+  };
+
+  cargoHash = "sha256-NER7NIWUXyWFUGcdZ5HLFaKaO9FNLPlG7jql08karS4=";
+
+  checkFlags = [
+    # make use of network connection
+    "--skip=tcp::tests::test_proxy_connection"
+  ];
+
+  meta = with lib; {
```
NickCao, line 24: Avoid using `with lib;` if possible.

pkgs/by-name/ws/wstunnel-rust/package.nix
```diff
@@ -0,0 +1,32 @@
...
+    description = "Wstunnel uses the websocket protocol which is compatible with http in order to bypass firewalls and proxies.";
```
NickCao, line 25: Description should not contain package name, should be short and concise, and shouldn't end with a period.

pkgs/by-name/ws/wstunnel-rust/package.nix
```diff
@@ -0,0 +1,32 @@
...
+    mainProgram = "wstunnel";
+    homepage    = "https://github.com/erebe/wstunnel";
+    license     = with licenses; [ bsd3Clear ];
```
NickCao, line 28: Doesn't seem to be `bsd3Clear`, not sure which variant of bsd license either.

h7x4:
> There is one exists in haskell but it seems the author abandon it: [search.nixos.org/packages?channel=24.05&from=0&size=50&sort=relevance&type=packages&query=wstunnel](https://search.nixos.org/packages?channel=24.05&from=0&size=50&sort=relevance&type=packages&query=wstunnel)
>
> I renamed the package to `wstunnel-rust` for now, but maybe we should just replace it with the haskell one?

> I will keep the name `wstunnel-rust` for now to avoid confusion with haskell package, but binary name intact as `wstunnel`.

According to the readme (https://github.com/erebe/wstunnel?tab=readme-ov-file#note), the project seems to have had a rewrite from haskell. This also checks out for hackage, with the author name being the same as the github project owner (https://hackage.haskell.org/package/wstunnel). Since this means that the haskell package is legacy code and this could be considered a "big update", I think the new package should just claim the `wstunnel` attribute, and then we drop the haskell package. There seems to be no haskell dependents, so there should not be any breakage.

pkgs/by-name/ws/wstunnel-rust/package.nix
NeverBehave: How to include the license here? It does sounds like a BSD

NickCao: There is a concerning line: `All rights reserved`, better ask author for clarification.

NeverBehave: I think this is from https://spdx.org/licenses/BSD-3-Clause-Clear.html ?

NickCao: But missing the `NO EXPRESS OR IMPLIED LICENSES TO ANY PARTY'S PATENT RIGHTS ARE GRANTED BY THIS LICENSE` wording about patent.

NeverBehave: Open a discussion at the project: https://github.com/erebe/wstunnel/discussions/293

NeverBehave: mark as unfree until we have update from the author

NeverBehave: Author makes clear that it is BSD3

h7x4:
There seems to be some cleanup needed for the `wstunnel` nixos module to function properly. Are you up to the task, or would you rather not touch it?

NickCao:
Hash mismatch on ofborg:
```
error: hash mismatch in fixed-output derivation '/nix/store/7jxg6yq98d22qxvh4qsc0n9y0fbkqig4-wstunnel-9.6.1-vendor.tar.gz.drv':
         specified: sha256-NER7NIWUXyWFUGcdZ5HLFaKaO9FNLPlG7jql08karS4=
            got:    sha256-a0k1j3LemSXHVZeU5+UgcIjrK9ppAgF65ojnOPlU3iw=
```

NeverBehave:
> There seems to be some cleanup needed for the `wstunnel` nixos module to function properly. Are you up to the task, or would you rather not touch it?

I think it only refer to the `pkgs.wstunnel` but CLI API should be the same. I didn't lookin to too much.
I am okay to work them thru but not in this PR if possible. @h7x4 how usually this should be handled.

h7x4:
> I think it only refer to the pkgs.wstunnel but CLI API should be the same. I didn't lookin to too much.

I have just barely looked at the cli flags, and the first thing that hits me is that they have changed from pascalCase to kebab-case. The available flags also did not seem to match up with the old ones completely. The package bump will probably break the module completely.

> I am okay to work them thru but not in this PR if possible. @h7x4 how usually this should be handled?

Tbh, I'm not sure what would be the usual way to handle it, if such a way even exists. It's not often that an update involves rewriting the program from the bottom up in another language. But in order to not render the module completely useless on the master branch, I think we either should introduce the package under another attribute name (e.g. `wstunnel-rust`, like you initially proposed) and keep the old package until we are able to fix the module, or we just wait with merging the PR until the module has been fixed in this same PR.

NeverBehave:
@h7x4 The original haskell package is actually marked broken and the module is not really usable. I could have the module package default pointed back the haskell one if that's okay, or I could include the module changes as well.

The reason I don't want to include that in this PR is 1) avoid bigger PR and testing headache 2) I am currently not familiar w/ testing nixos modules yet thus this will stale for a while until I put all pieces together..

NeverBehave:
Actually I just found there is another PR around this: https://github.com/NixOS/nixpkgs/pull/289652

NeverBehave:
The API changes a lot, I have to change the module a lot. E.g. structurally changed the host module definition. 
@alyaeanyx Would you open to that? Or you have better idea how this would proceed.

NeverBehave:
I have tried building the module with some different configuration.nix w/ `nixos-rebuild build-vm --fast -I nixos-config=./configuration.nix -I nixpkgs=nixpkgs --show-trace` and fix some typo.

Since there isn't much feedback from the original maintainer, @NickCao Would you be able to take another quick look about the changes here? Since it involve some module changes, I am not sure if I should add some extra docs etc. 
Any advise would be appreciated.

nixos/modules/services/networking/wstunnel.nix
```diff
@@ -66,10 +48,16 @@ let
       };
     };
 
-    verboseLogging = mkOption {
-      description = "Enable verbose logging.";
-      type = types.bool;
-      default = false;
+    loggingLevel = mkOption {
+      description = ''
+        passing to --log-lvl
```
NickCao, line 53:
```suggestion
        Passed to --log-lvl
```

nixos/modules/services/networking/wstunnel.nix
```diff
@@ -66,10 +48,16 @@ let
...
+
+        Control the log verbosity. i.e: TRACE, DEBUG, INFO, WARN, ERROR, OFF
+        for more details: https://docs.rs/tracing-subscriber/latest/tracing_subscriber/filter/struct.EnvFilter.html#example-syntax
```
NickCao, line 56: We can use markdown formatting here for the url. e.g. `[EnvFilter](https://docs.rs/tracing-subscriber/latest/tracing_subscriber/filter/struct.EnvFilter.html#example-syntax)`

nixos/modules/services/networking/wstunnel.nix
```diff
@@ -134,59 +123,36 @@ let
...
-        example = {
-          host = "example.com";
-        };
-      };
-
-      enableHTTPS = mkOption {
-        description = "Enable HTTPS when connecting to the server.";
-        type = types.bool;
-        default = true;
+        type = types.str;
```
NickCao, line 131: Why not the `hostPortSubmodule`?

NickCao (commented):
The package part looks good. But the inconsistent use of `str` and `localRemoteSubmodule` types in the module can use some rework.

nixos/modules/services/networking/wstunnel.nix
NeverBehave: Right now it has at least two protocol support: 

> <ws[s]|http[s]://wstunnel.server.com[:port]>
          Address of the wstunnel server
          You can either use websocket or http2 as transport protocol. Use websocket if you are unsure.
          Example: For websocket with TLS wss://wstunnel.example.com or without ws://wstunnel.example.com
                   For http2 with TLS https://wstunnel.example.com or without http://wstunnel.example.com

NeverBehave:
> `localRemoteSubmodule`

This submodule is now replaced with str since the number of protocol/combination has expand a lot. Similar to `services.autossh.sessions.*.extraArguments`, where user specify the entire forward options instead of predefined submodule, seems to be a better approach imo for this situation.

NickCao (approved):
diff LGTM, please format the second commit as `nixos/wstunnel: ...`.

NickCao:
And a few points for future improvements:
- The lack of `utils.escapeSystemdExecArgs` client unit.
- A nixos test.

PR #81135, opened 2020-02-26
nixos/version: validate system.stateVersion

tilpner:
###### Motivation for this change

Stripped down #80907 to validate only.

###### Things done

Correctly invalid (on branch nixos-20.03):

- "19", "19.3"
- "unstable", "nixos-unstable", "nixos.unstable"
- "19.04"

Questionably invalid (on branch nixos-20.03):

- "20.09": future value, not used yet, but there might still be reason for setting this

Correctly valid:

- "19.03"
- "16.09"

Incorrectly valid:

- "xy.03"

###### Open questions

- Is this the right place to encode how NixOS versions look like?
- Should future NixOS versions be valid?

PR #317858, opened 2024-06-06
nixos/version: validate system.stateVersion

tilpner:
## Description of changes

Continuation of #81135.

The location of the release type is open for discussion, it could go into lib.types instead.

Test with

```nix
nix-instantiate --eval -E '(import ./nixos { configuration.system.stateVersion = "20.05"; }).config.system.stateVersion'
```

Correctly invalid (with `lib.trivial.release == "24.11"`):

    "19", "19.3"
    "unstable", "nixos-unstable", "nixos.unstable"
    "19.04"
    "xy.03"
    "25.11"
    "20.05"
    "21.03"

Correctly valid:

    "19.03"
    "16.09"
    "20.09"
    "21.05"
    "24.11"

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [x] Tested, as applicable:

---

### DIFF
--- a/nixos/doc/manual/release-notes/rl-2411.section.md
+++ b/nixos/doc/manual/release-notes/rl-2411.section.md
@@ -45,6 +45,10 @@
 - `zx` was updated to v8, which introduces several breaking changes.
   See the [v8 changelog](https://github.com/google/zx/releases/tag/8.0.0) for more information.
 
+- `system.stateVersion` is now validated. If you never changed this yourself, you don't need to do anything. If your `stateVersion` is not a valid NixOS release version (e.g. "24.11" is valid),
+  your system was already at risk of experiencing silent incompatible state updates. If your previous value is a well-formed version but not a valid release (e.g. "23.12"),
+  round down to the nearest actual release. If it wasn't a well-formed version (e.g. "nixos-unstable"), set it to the version of NixOS that you originally installed.
+
 - The `portunus` package and service do not support weak password hashes anymore.
   If you installed Portunus on NixOS 23.11 or earlier, upgrade to NixOS 24.05 first to get support for strong password hashing.
   Then, follow the instructions on the [upstream release notes](https://github.com/majewsky/portunus/releases/tag/v2.0.0) to upgrade all existing user accounts to strong password hashes.
--- a/nixos/modules/misc/version.nix
+++ b/nixos/modules/misc/version.nix
@@ -44,6 +44,49 @@ let
   };
   initrdRelease = pkgs.writeText "initrd-release" (attrsToText initrdReleaseContents);
 
+  checkRelease = version:
+    let
+      parts = lib.versions.splitVersion version;
+      isVersion = lib.length parts == 2 && lib.all (p: lib.stringLength p == 2) parts;
+      majorVersion = lib.toIntBase10 (lib.elemAt parts 0);
+      minorVersion = lib.elemAt parts 1;
+
+      versionPatterns = [
+        # only 13.10
+        { fromMajor = 13; minor = [ "10" ]; }
+        # 14.04 and 14.12
+        { fromMajor = 14; minor = [ "04" "12" ]; }
+        # only 15.09
+        { fromMajor = 15; minor = [ "09" ]; }
+        # 16.03 to 20.09
+        { fromMajor = 16; minor = [ "03" "09" ]; }
+        # from 21.05
+        { fromMajor = 21; minor = [ "05" "11" ]; }
+      ];
+
+      # find the versioning pattern that applies by looking for the first
+      # major version newer than `majorVersion`, and picking the previous pattern
+      patternIndex = lib.lists.findFirstIndex
+        ({ fromMajor, ... }: fromMajor > majorVersion)
+        (lib.length versionPatterns)
+        versionPatterns;
+
+      validMinorVersions =
+        if patternIndex == 0
+        then []
+        else (lib.elemAt versionPatterns (patternIndex - 1)).minor;
+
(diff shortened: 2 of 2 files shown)


tilpner:
@jopejoe1 noticed that [`system.stateVersion` is occasionally set to versions earlier than it was introduced](https://github.com/NixOS/infra/blob/c8b2e1baf8a562bef4e103ba33ad28c5226e774b/build/haumea/default.nix#L52). The adjusted commit supports all versions from 13.10 on (as listed on https://channels.nixos.org/), but unfortunately the less regular minor version pattern make for a somewhat less compact check.

jopejoe1 (commented):
Invalid versions, that have a Major version between 10 and 12 and a minor version 2 digits long, currently return the following error instead of the in tended error message:

```error
error: expected a set but found null: null
```

This could be fixed by adding `{ fromMajor = 10; minor = [ ]; }` to the `versionPatterns` list so that it returns the intended error message:

```error
 error: A definition for option `system.stateVersion' is not of type `NixOS release version, e.g. "24.11"'. Definition values:
       - In `/home/jopejoe1/Documents/nixpkgs/nixos/default.nix': "10.11"

```

tilpner:
@jopejoe1 Good catch, thank you! My validation script was just checking that the set of accepted versions is correct, but didn't check the error message. The error message is now consistent over {09..26}.{00..12}.

jopejoe1 (approved):
Correctly identified all versions I tried to dump on it.

a-kenji (approved):

mweinelt:
Reverted as this broke the channel evaluation, see https://github.com/NixOS/nixpkgs/pull/339671#issue-2506577382.

tilpner:
Oh, sorry about that. I do remember seeing that style of version strings (24.11pre56789.gfedcba) from the ISO grub menu, but never considered it might be set as `system.stateVersion`. Or that `lib.versionAtLeast` will special-case the `pre` element in that version string (via `builtins.compareVersions`, following [these rules](https://nix.dev/manual/nix/2.18/command-ref/nix-env/upgrade#versions)).

It's not entirely clear where and how to fix this, as the intended format of `stateVersion` is not explicitly spelled out:

The format pre-filled by `nixos-generate-config` is probably the only way most people interact with `stateVersion`, and I've never seen that be anything but the `XX.YY` format.

The option description only mentions the format in this paragraph, but only by example rather than definition:

> For example, if NixOS version XX.YY ships with AwesomeDB version N by default, and is then upgraded to version XX.YY+1, which ships AwesomeDB version N+1, the existing databases may no longer be compatible, causing applications to fail, or even leading to data loss.

The default value comes from `system.nixos.release`, which is described simply as "The NixOS release (e.g. `16.03`)." and defaults to `lib.trivial.release`, which is read from `<nixpkgs/.version>`, which has never been anything but `XX.YY`.

From this I would conclude that the format was intended to be just `XX.YY`, but since it doesn't explicitly forbid appending to it, and the version comparison functions usually used with `stateVersion` do interpret extra version data, there could be *someone* actually using the extra flexibility to express the version constraint "installed from nixos-unstable after 24.05, but before 24.11".

Regardless, I don't think anything within nixpkgs/nixos relies on the `pre` (or much worse, the hash part), and the "pre56789.gfedcba" could reasonably have slipped into the `stateVersion` by accident, and gone unnoticed because it does mostly just continue to work (until you try to validate it). The release ISOs should probably be generated with `system.stateVersion = config.system.nixos.release;` instead.

emilazy:
BTW, is there a reason for the complex version logic in this PR rather than just having an `enum` of all the releases? I guess we might not have a list of all release versions anywhere, but it seems like it might be good to anyway.

tilpner:
I didn't want to force an update of the validation function for every new release, only for new release schedules. To be fair, the previous version of this PR looked quite a bit simpler, but didn't account for the earlier, more volatile, release schedules from before stateVersion was introduced. For those, a list would definitely be simpler, but I wanted to avoid having two competing ways to list a version (exhaustive list for early versions, then schedules for more recent ones), so I shoehorned them into the schedule list as well, even though a few of them only had a single release in their respective schedule. I'm not perfectly happy with it, but tried the split approach and liked it less. :/

emilazy:
Right, I was thinking we could maybe have a list of past releases in a central place. (Though I wonder how many conditionals from the irregular release cycle era are even still around?)

tilpner:
> Though I wonder how many conditionals from the irregular release cycle era are even still around?

The oldest in current nixpkgs seems to be "17.09", but the validation doesn't prevent you from comparing against invalid/irregular versions, it just prevents you from setting one as your `stateVersion`.
As @jopejoe1 found, NixOS infra is still using [`stateVersion = "14.12";`](https://github.com/NixOS/infra/blob/c8b2e1baf8a562bef4e103ba33ad28c5226e774b/build/haumea/default.nix#L52), and given the hassle and risk in upgrading `stateVersion`, it's not reasonable to ask anyone to upgrade just to make the check more readable.

> I was thinking we could maybe have a list of past releases in a central place.

An exhaustive list in a central place seems useful at first, and I definitely wanted one while making this change (I referred to https://channels.nixos.org/ instead), but I can't come up with many usecases (post-validation) where it helps a lot:

- while writing a conditional that includes `stateVersion` in some way
  - but for those you usually only need the current or previous release version. 
- maybe checking that an old conditional you found was ever valid

I don't want to unnecessarily add to the manual release steps, but if you and others think that the exhaustive list would be worth it, I'll rewrite it with a simple list.

PR #259275, opened 2023-10-05
caddy: Allow adding external plugins

jpds:
## Description of changes

Solves: #14671

Caddy has support for plugins that are added at compile time. This exposes an argument `externalPlugins` that will build caddy with the specified plugins.

Example:
```
caddy-custom = pkgs.caddy.override {
  externalPlugins = [
    {name = "coraza"; repo = "github.com/corazawaf/coraza-caddy/v2"; version = "v2.0.0-rc.3";}
    {name = "cache-handler"; repo = "github.com/caddyserver/cache-handler"; version = "v0.9.0";}
    {name = "ratelimit"; repo = "github.com/mholt/caddy-ratelimit"; version = "2dc0d586f0b87e983757c403bc0929ddeb84a537";}
  ];
  vendorHash = "<SRI hash>";
};
```

Adapted from Brendan Taylor's CoreDNS change in commit 95e66809debf42dbe1e4935fd31c8c275914d2eb.

## Things done

- Built on platform(s)
  - [X] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
- [ ] For non-Linux: Is `sandbox = true` set in `nix.conf`? (See [Nix manual](https://nixos.org/manual/nix/stable/command-ref/conf-file.html))
- [ ] Tested, as applicable:
  - [NixOS test(s)](https://nixos.org/manual/nixos/unstable/index.html#sec-nixos-tests) (look inside [nixos/tests](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests))
  - and/or [package tests](https://nixos.org/manual/nixpkgs/unstable/#sec-package-tests)
  - or, for functions and "core" functionality, tests in [lib/tests](https://github.com/NixOS/nixpkgs/blob/master/lib/tests) or [pkgs/test](https://github.com/NixOS/nixpkgs/blob/master/pkgs/test)
  - made sure NixOS tests are [linked](https://nixos.org/manual/nixpkgs/unstable/#ssec-nixos-tests-linking) to the relevant packages
- [ ] Tested compilation of all packages that depend on this change using `nix-shell -p nixpkgs-review --run "nixpkgs-review rev HEAD"`. Note: all changes have to be committed, also see [nixpkgs-review usage](https://github.com/Mic92/nixpkgs-review#usage)
- [X] Tested basic functionality of all binary files (usually in `./result/bin/`)
[...]

Issue #14671, opened 2016-04-13
caddy: add all addons


davidak:
compile caddy with all addons listed on the [download page](https://caddyserver.com/download).

> Only middleware handlers which are invoked from the Caddyfile will be chained in, so small Caddyfiles are very fast and efficient.

source: Directive section on https://caddyserver.com/docs/caddyfile

a later improvement could be to select just the addons you want like vim_configurable.

@angus-g

angus-g:
This does sound like a good idea to improve the usefulness of caddy, but it's quite non-trivial from what I can see. It looks like there's some code generation required (in https://github.com/caddyserver/caddydev/blob/master/caddybuild/build.go) to insert the custom directives into caddy.

If that's the case, it would probably be easier to start with desired addons and do the code generation at build time. I expect this would need a bit more Nix code than currently exists for the package in order to support hashes for caddy and all addons, then modifying and building the source after this. I'm not great with the language just yet, so I'm not even sure of the feasibility of this. (Anything's possible, right?)

davidak:
I found a Dockerfile where someone installed caddy with the git addon. That is also the one i want to use.

https://github.com/jumanjihouse/docker-caddy/blob/master/Dockerfile.build#L34

You could write the nix-expression like for a general program, but the go infrastructure is different.
You should ask in the IRC or Mailinglist how to realize that best. I'm also not that experienced with nix, never packaged a go program.

davidak:
you can override build phases like described here https://nixos.org/nixpkgs/manual/#ssec-controlling-phases

for example the preBuildPhases like here: https://github.com/NixOS/nixpkgs/blob/dfc6d42860b41042deeccb487c73adb138a36e48/pkgs/top-level/go-packages.nix#L642

then you can use `caddyext install git` in the build phase.

if you are experienced with go, you can do that. i could also figure it out and create a PR. how you like :)

davidak:
we can also create an additional package of [caddyext](https://github.com/caddyserver/caddyext) and use it as caddy buildInputs.

but i see 2 problems:

- the hash of the addon source is not checked
- reproducible builds are impossible since the source of the addons change over time

0xABAB:
If the source of the addons changes over time, I would try to convince upstream to change their ways and otherwise completely ignore this project, for a lack of professional software development practices. 

Sometimes "gratis" software is even too costly.

davidak:
[...]

PR #317881, opened 2024-06-07
caddy: add support for compiling with Caddy modules (plugins)

Nanotwerp:
## Description of changes
Solves: https://github.com/NixOS/nixpkgs/issues/14671

This PR continues the work of @jpds in #259275 toward adding support for building `caddy` with Caddy modules (plugins). A few recommendations given in code reviews of the PR were used, along with its previous changes being rebased. The PR is split in three commits: one adding support for compiling `caddy` with plugins (`caddy-with-plugins`), one formatting it with `nixfmt-rfc-style`, and one adding its documentation to the `24.11` release notes.

To support the compilation of `caddy` with plugins, a `passthru` function named `withPlugins` has been added that takes in a list of attribute sets and a string as arguments (`caddyModules` and `vendorHash`, respectively), resulting in a `caddy-with-plugins` package. `caddyModules` is the list of attribute sets that represents a list of `caddy` plugins. The valid attributes for a plugin are `repo` and `version`. Adding plugins changes the `buildGoModule` `vendorHash` from that of vanilla `caddy`, so the string argument also named `vendorHash` is used to change the former hash for `caddy-with-plugins` . 

The value of the `vendorHash` argument defaults to `lib.fakeHash`, which allows for one to find the correct hash by attempting to build `caddy.withPlugins [<attrset(s) of plugins>]` without setting the `vendorHash`. Using `caddy.withPlugins [<attrset(s) of plugins>]` on the result of overriding `caddy` with the correct `vendorHash` produces a `caddy-with-plugins` package, compiled with the inputted plugins.

Example:
  ```
  services.caddy = {
    enable = true;
    package = pkgs.caddy.withPlugins {
      caddyModules = [
        { repo = "github.com/corazawaf/coraza-caddy/v2"; version = "v2.0.0-rc.3"; }
        { repo = "github.com/caddyserver/cache-handler"; version = "v0.9.0"; }
        { repo = "github.com/mholt/caddy-ratelimit"; version = "2dc0d586f0b87e983757c403bc0929ddeb84a537"; }
      ];
      vendorHash = "<SHA256 hash>";
    };
  };
  ```
Adapted from Brendan Taylor's CoreDNS change in commit 95e66809debf42dbe1e4935fd31c8c275914d2eb.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/by-name/ca/caddy/package.nix
+++ b/pkgs/by-name/ca/caddy/package.nix
@@ -1,24 +1,26 @@
 { lib
 , buildGoModule
 , fetchFromGitHub
+, gnused
+, installShellFiles
 , nixosTests
 , caddy
 , testers
-, installShellFiles
 , stdenv
 }:
 let
+  attrsToModule = map (plugin: plugin.repo);
+  attrsToVersionedModule = map ({ repo, version, ... }: lib.escapeShellArg "${repo}@${version}");
+
+  pname = "caddy";
   version = "2.8.4";
+
   dist = fetchFromGitHub {
     owner = "caddyserver";
     repo = "dist";
     rev = "v${version}";
     hash = "sha256-O4s7PhSUTXoNEIi+zYASx8AgClMC5rs7se863G6w+l0=";
   };
-in
-buildGoModule {
-  pname = "caddy";
-  inherit version;
 
   src = fetchFromGitHub {
     owner = "caddyserver";
@@ -26,8 +28,6 @@ buildGoModule {
     rev = "v${version}";
     hash = "sha256-CBfyqtWp3gYsYwaIxbfXO3AYaBiM7LutLC7uZgYXfkQ=";
   };
-
-  vendorHash = "sha256-1Api8bBZJ1/oYk4ZGIiwWCSraLzK9L+hsKXkFtk6iVM=";
 
   subPackages = [ "cmd/caddy" ];
 
@@ -39,7 +39,7 @@ buildGoModule {
   # matches upstream since v2.8.0
   tags = [ "nobadger" ];
 
-  nativeBuildInputs = [ installShellFiles ];
+  nativeBuildInputs = [ gnused installShellFiles ];
 
   postInstall = ''
     install -Dm644 ${dist}/init/caddy.service ${dist}/init/caddy-api.service -t $out/lib/systemd/system
@@ -61,19 +61,80 @@ buildGoModule {
       --zsh <($out/bin/caddy completion zsh)
   '';
 
-  passthru.tests = {
-    inherit (nixosTests) caddy;
-    version = testers.testVersion {
-      command = "${caddy}/bin/caddy version";
-      package = caddy;
-    };
-  };
-
   meta = with lib; {
     homepage = "https://caddyserver.com";
     description = "Fast and extensible multi-platform HTTP/1-2-3 web server with automatic HTTPS";
     license = licenses.asl20;
     mainProgram = "caddy";
     maintainers = with maintainers; [ Br1ght0ne emilylange techknowlogick ];
+  };
+in
+buildGoModule {
+  inherit
+    pname
+    version
+    src
+    subPackages
+    ldflags
+    tags
+    nativeBuildInputs
+    postInstall
+    meta
+    ;
+
+  vendorHash = "sha256-1Api8bBZJ1/oYk4ZGIiwWCSraLzK9L+hsKXkFtk6iVM=";
+
+  passthru = {
+    withPlugins =
+      {
+        caddyModules,
+        vendorHash ? lib.fakeHash,
+      }:
+      buildGoModule {
+        pname = "${caddy.pname}-with-plugins";
+
+        inherit
+          version
+          src
+          subPackages
+          ldflags
+          tags
+          nativeBuildInputs
+          postInstall
+          meta
+          ;
+
+        modBuildPhase = ''
+          for module in ${toString (attrsToModule caddyModules)}; do
+            sed -i "/standard/a _ \"$module\"" ./cmd/caddy/main.go
+          done
+          for plugin in ${toString (attrsToVersionedModule caddyModules)}; do
+            go get $plugin
+          done
+          go mod vendor
+        '';
+
+        modInstallPhase = ''
+          mv -t vendor go.mod go.sum
+          cp -r vendor "$out"
+        '';
+
+        preBuild = ''
+          chmod -R u+w vendor
+          [ -f vendor/go.mod ] && mv -t . vendor/go.{mod,sum}
+          for module in ${toString (attrsToModule caddyModules)}; do
+            sed -i "/standard/a _ \"$module\"" ./cmd/caddy/main.go
+          done
+        '';
+
+        inherit vendorHash;
+      };
+    tests = {
+      inherit (nixosTests) caddy;
+      version = testers.testVersion {
+        command = "${caddy}/bin/caddy version";
+        package = caddy;
+      };
+    };
   };
 }
--- a/pkgs/by-name/ca/caddy/package.nix
+++ b/pkgs/by-name/ca/caddy/package.nix
@@ -1,12 +1,13 @@
-{ lib
-, buildGoModule
-, fetchFromGitHub
-, gnused
-, installShellFiles
-, nixosTests
-, caddy
-, testers
-, stdenv
+{
+  lib,
+  buildGoModule,
+  fetchFromGitHub,
+  gnused,
+  installShellFiles,
+  nixosTests,
+  caddy,
+  testers,
+  stdenv,
 }:
 let
   attrsToModule = map (plugin: plugin.repo);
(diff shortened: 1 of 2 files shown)


emilylange (changes requested):
Blocking due to https://github.com/NixOS/nixpkgs/pull/259275#issuecomment-1962974820.

Nanotwerp:
> Blocking due to [#259275 (comment)](https://github.com/NixOS/nixpkgs/pull/259275#issuecomment-1962974820).

Maybe I'm misunderstanding, but it seems that (to your point about `lib.fakeHash`), it *is* being used for `vendorHash`. When I try to compile `caddy` with plugin(s) with `vendorHash = ""`, it specifies `sha256-AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA=`, not the default `vendorHash` of `caddy`. Also, I thought `passthru` was for attributes that don't affect the build itself. Are you suggesting something like using `passthru` along with a wrapper for a package of `caddy` with external plugins? Something like [this](https://github.com/NixOS/nixpkgs/blob/573a75cb4ac63cce91d4e7d003f5779ff9c7cae9/pkgs/development/tools/gauge/wrapper.nix#L48)?

emilylange:
Sorry for taking so long to respond.
Will try to find time for this tomorrow.

emilylange:
> Maybe I'm misunderstanding, but it seems that (to your point about `lib.fakeHash`), it _is_ being used for `vendorHash`. When I try to compile `caddy` with plugin(s) with `vendorHash = ""`, it specifies `sha256-AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA=`, not the default `vendorHash` of `caddy`.

An empty FOD hash (e.g. `vendorHash = ""`) and `lib.fakeHash` result in the same here.
The former is implemented right in nixcpp, while the latter is part of nixpkgs.

So when I wrote

> This also comes with the added benefit, that the `vendorHash` can be left empty by default to use `lib.fakeHash` instead of the vanilla caddy-without-plugins one.

I was referring to the universal "dummy hash" that is `sha256-AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA=` that both nixcpp and nixpkgs use, saving me from having to copy it from somewhere.
You can use `""` if you want.

Furthermore, the important bit is

> [...] `vendorHash` can be left empty by default [...] instead of the vanilla caddy-without-plugins one.

I foresee a lot of users using the suggested `caddy.override { }` pattern from #259275 (and this PR here), but forgetting to specify `vendorHash`, which would result in the default vanilla variant.

We don't have any checks whether the specified plugins are actually part of the build and resulting binary yet.
We should add something like that.

If we had a dedicated entrypoint in `passthru` that would default to `vendorHash = ""`, accidentally building vanilla Caddy despite specifying plugins couldn't happen.
One foot gun less.

> Also, I thought `passthru` was for attributes that don't affect the build itself.

Yep, that's intentional. Changing something in `passthru` won't trigger rebuilds of the package itself. We want that.

> Are you suggesting something like using `passthru` along with a wrapper for a package of `caddy` with external plugins? Something like [this](https://github.com/NixOS/nixpkgs/blob/573a75cb4ac63cce91d4e7d003f5779ff9c7cae9/pkgs/development/tools/gauge/wrapper.nix#L48)?

Yes.

However, note that I suggested and requested Caddy with plugins to not replace the current vanilla Caddy in the [linked comment](https://github.com/NixOS/nixpkgs/pull/259275#issuecomment-1962974820):

> I do, however, not like, that this replaces current caddy package.
> [...]
> How about the following:
> We keep the current caddy drv as is, and add a with-external-plugins in passthru.
> [...]

Which this PR doesn't. This PR, just like #259275, replaces the current Caddy derivation.

And lastly, some time passed since my initial linked comment in #259275 and I changed my mind about one detail in the meantime.

I now heavily prefer the `xcaddy`+`XCADDY_SKIP_BUILD=1` approach over the custom `modBuildPhase` and `preBuild` phases.

This also has the added benefit of giving us `xcaddy`'s `--embed` feature for free (that has been released as part of https://github.com/caddyserver/xcaddy/releases/tag/v0.4.0).

I hope this clears up some questions.

Nanotwerp:
> I hope this clears up some questions.

Ah, I think I understand. Thank you! I'll try to convert the PR to use `xcaddy` and add a  `passthru` when I have the time.

Nanotwerp:
@emilylange I followed your suggestion and exposed the functionality of compiling `caddy` with plugins as a `passthru` function. These changes are documented in the latest edit of this PR's description. I did have issues with using `xcaddy`, though. Every method that I've seen which uses `xcaddy` seems to require bypassing the Nix sandbox to allow network connectivity during the build process, via the `outputHash` fixed-output derivation. Wouldn't it be more idiomatic to not use `xcaddy`? [This section](https://github.com/caddyserver/xcaddy?tab=readme-ov-file#warning-pro-tip) of `xcaddy`'s GitHub page suggests not using it if it conflicts too much with one's build system.

pkgs/by-name/ca/caddy/package.nix
```diff
@@ -61,11 +64,58 @@ buildGoModule {
...
-      package = caddy;
+  passthru = {
+    withPlugins =
+      {
+        caddyModules,
+        vendorHash ? lib.fakeHash,
+      }:
+      buildGoModule {
+        pname = "${caddy.pname}-with-plugins";
+        inherit (caddy)
```
nevivurn, line 75: One concern, this might result in confusing behavior with overrides, as `(caddy.override ...).withPlugins ...` would not work as expected. This could be resolved by inheriting from the outer scope instead of `(caddy)` I believe.

Nanotwerp: Thank you for the suggestion! I [restructured](https://github.com/NixOS/nixpkgs/pull/317881/commits/db5046f7caef06498835c47185a9035f80b446b3) the code to change its scoping so that `caddy.withPlugins` inherits from the outer scope. For some reason, though, this doesn't seem to fix `overrideAttrs` done on `caddy` when `caddy.withPlugins` is used. Would you suggest I add documentation that recommends using `overrideAttrs` on `caddy.withPlugins` instead of on `caddy` when deriving `caddy-with-plugins`?

flokli (approved):
LGTM!

pkgs/by-name/ca/caddy/package.nix
```diff
@@ -27,53 +30,124 @@ buildGoModule {
...
-
-  passthru.tests = {
-    inherit (nixosTests) caddy;
-    version = testers.testVersion {
-      command = "${caddy}/bin/caddy version";
-      package = caddy;
-    };
-  };
+  nativeBuildInputs = [
+    gnused
```
nevivurn, line 45: gnused shouldn't be necessary for non-extended caddy. Even in the `withPlugins` case, it doesn't seem strictly necessary (builds fine without it) though I don't know whether if it's best practice to include it anyway.

pkgs/by-name/ca/caddy/package.nix
```diff
@@ -27,53 +30,124 @@ buildGoModule {
...
+          ;
+
+        modBuildPhase = ''
+          for module in ${toString (attrsToModules caddyModules)}; do
+            sed -i "/standard/a _ \"$module\"" ./cmd/caddy/main.go
+          done
+          for plugin in ${toString (attrsToSources caddyModules)}; do
+            go get $plugin
+          done
+          go generate
```
nevivurn, line 125: Question, why is go generate needed here? Same in the `preBuild` below.

pkgs/by-name/ca/caddy/package.nix
nevivurn: Getting it to work correctly with `overrideAttrs` will probably require something like the `finalAttrs` pattern for `buildGoModule`, which we don't have yet. Unsure on the docs mention.

pkgs/by-name/ca/caddy/package.nix
Nanotwerp: `go generate` isn't needed at all! Thank you! I removed it.

pkgs/by-name/ca/caddy/package.nix
Nanotwerp: `gnused` is necessary to support Mac build targets, due to Mac using BSD `sed` by default, which has a different syntax for `sed -i`. See [this review of the previous PR](https://github.com/NixOS/nixpkgs/pull/259275/files/a33b02fa9d664f31dadc8a874eb1a5dbaa9f4ecf#r1463001643) and [this thread](https://stackoverflow.com/questions/4247068/sed-command-with-i-option-failing-on-mac-but-works-on-linux).

nevivurn (approved):

flokli:
@emilylange any final thoughts on this?

emilylange:
> @emilylange any final thoughts on this?

Yes. I don't think this should get merged as is.

It uglifies this derivation and exposes footguns to an extent, that I am not willing to maintain.
I will try to mention a few points, but ultimately, I won't be able to delegate much time into this any time soon.

- There are no checks in place. A check phase should be added to validate whether the specified modules are actually part of the resulting binary. This isn't as straight forward if one wants to take cross compilation into account. But still, some sort of check is absolutely necessary. There are many ways to approach such check.
- The `src` FOD hash is now more than 50 lines apart from the last hash in the file, the `vendorHash`. We usually try to avoid this in nixpkgs where possible. There are multiple reasons, one of them being that it's easy to overlook one of the hashes. Totally avoidable here. `vendorHash` should get moved somewhere closer to the other hashes.
- The derivation in general is a mess now. Things have been moved to a `let in` to then be inhered. Things could easily be inherited and deduplicated without the use of `let in`. It would even result less duplication.
- The approach how things are deduplicated is simply inconsistent. E.g. `pname` exists in the `let in` scope, is then inherited in the "vanilla" `buildGoModule`, but then in `passthru`, `caddy.pname` is used instead of `pname` from the `let in` scope.
- There are smaller imperfections, like `{ repo, version, ... }` should be `{ repo, version }`, because there are no unknown overflows.
- This should leverage FOD store path invalidation, or whatever this concept may be called. Some representation of `caddyModules` should end up in `goModules.pname`. As it stands, it's currently `/nix/store/<hash>-caddy-with-plugins-2.8.4-go-modules`. If it encoded `caddyModules` in there instead, a change in `caddyModules` would result in a missing store path, a FOD hash lookup at cache.nixos.org, which is essentially guaranteed to fail, a forceful local build of `goModules`, which then gets validated against the provided FOD hash, which may or may not be accurate after the `caddyModules` modification.

And of course, it does not use `xcaddy`. I advocate for `xcaddy` and don't quite agree with the reasoning against it:

> Every method that I've seen which uses `xcaddy` seems to require bypassing the Nix sandbox to allow network connectivity during the build process, via the `outputHash` fixed-output derivation.

The `xcaddy` approach I suggested uses a single FOD for `src` and then `null`s `vendorHash`.
Meaning it requires a single FOD just like this variant requires a single FOD, additional to the vanilla derivation, that is.

> Wouldn't it be more idiomatic to not use `xcaddy`? [This section](https://github.com/caddyserver/xcaddy?tab=readme-ov-file#warning-pro-tip) of `xcaddy`'s GitHub page suggests not using it if it conflicts too much with one's build system.

I don't see how `xcaddy` conflicts with our build system. Can you elaborate?

And finally, as said before, `xcaddy` provides an FS `--embed` feature out of the box. While niche, it would be fairly trivial to add and expose it in our abstraction, if it's based on `xcaddy`.

So no, I strongly believe this should not get merged as is.

@flokli, if you want to take responsibility for it, dismiss my blocking review and merge it. I won't stop you.

However, I would like to be removed from `caddy.meta.maintainers` in that case. No hard feelings. I am just not willing to maintain this. You can either add another commit dropping me to this PR, or I'll open a follow-up PR when I find time.

flokli:
Thanks for the thorough explanation ❤️

I was not sure your comments were sufficiently addressed. But now you left some valid, valuable and actionable feedback on how this can be less of a footgun.

I agree with the reasoning, and don't want to leave this in a less maintainable state than before.

I'm happy to offer some help on getting this into shape, @Nanotwerp let me know if you need some help.

Nanotwerp:
> Thanks for the thorough explanation ❤️
> 
> I was not sure your comments were sufficiently addressed. But now you left some valid, valuable and actionable feedback on how this can be less of a footgun.
> 
> I agree with the reasoning, and don't want to leave this in a less maintainable state than before.
> 
> I'm happy to offer some help on getting this into shape, @Nanotwerp let me know if you need some help.

Thank you so much for offering your help. I'll probably need all the help I can get!

flokli:
> Thank you so much for offering your help. I'll probably need all the help I can get!

Feel free to reach out on Matrix or any other communication medium :)

vincentbernat:
Here is an alternative using xcaddy: https://github.com/vincentbernat/caddy-nix. I can do a PR to integrate it if there is some interest.

thegatesdev:
Thanks for the great amount of work already done on this!

I wanted to note that certain plugins [like FrankenPHP](https://frankenphp.dev/docs/compile/) will require more setup to get working.
In case of FrankenPHP, it requires PHP (built with some extra flags) to be installed in order to compile.
See [this gist](https://gist.github.com/thegatesdev/016fbce6a176b43f0de02890f25a9b37) for how I got it building using xcaddy (excuse the workarounds for go module downloading, I didn't find a solution for that before).

I'm definitely interested in this, I wanted to build caddy with plugins for my weak VPS, turned out to be alot of trouble.

Nanotwerp:
Closing. Superseded by #358586!

PR #318047, opened 2024-06-07
julius: darwin build, add maintainer

matteo-pacini:
## Description of changes

Added Darwin support to Julius.

List of changes:
- Disabled `fixup_bundle` CMake directive - gave a few errors and doesn't really makes sense on Nix
- Disabled unit tests - they don't compile 
- Set min deployment target to macOS 11.0
- Cmake app bundle generation :
    - replaced `sips` with `convert` (`imagemagick`)
    - replaced `iconutil` with `png2icns` (`libicns`)
    - replaced `codesign` binary with the one from `darwin.sigtool`
- Added myself as second maintainer for Darwin

The screenshot below is from the compiled `julius.app` running on my M1 machine (macOS 14.5).

Screenshot 2024-06-07 at 17 52 43

Binaries tested both on `aarch64-darwin` and `aarch64-linux`.

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [X] aarch64-linux
  - [ ] x86_64-darwin
  - [X] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/games/julius/darwin-fixes.patch
+++ b/pkgs/games/julius/darwin-fixes.patch
@@ -1,0 +1,49 @@
+diff --git a/CMakeLists.txt b/CMakeLists.txt
+index 038a34ba..d46b9258 100644
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
+@@ -669,16 +669,14 @@ endif()
+ if(APPLE)
+     # generating a macOS icns file (see https://stackoverflow.com/a/20703594)
+     add_custom_command(OUTPUT ${PROJECT_SOURCE_DIR}/res/julius.icns
+-        COMMAND mkdir -p julius.iconset
+-        COMMAND sips -z 16 16    julius_256.png --out julius.iconset/icon_16x16.png
+-        COMMAND sips -z 32 32    julius_256.png --out julius.iconset/icon_16x16@2x.png
+-        COMMAND sips -z 32 32    julius_256.png --out julius.iconset/icon_32x32.png
+-        COMMAND sips -z 64 64    julius_256.png --out julius.iconset/icon_32x32@2x.png
+-        COMMAND sips -z 128 128  julius_256.png --out julius.iconset/icon_128x128.png
+-        COMMAND sips -z 256 256  julius_256.png --out julius.iconset/icon_128x128@2x.png
+-        COMMAND sips -z 256 256  julius_256.png --out julius.iconset/icon_256x256.png
+-        COMMAND sips -z 512 512  julius_512.png --out julius.iconset/icon_256x256@2x.png
+-        COMMAND iconutil -c icns julius.iconset
++        COMMAND mkdir julius.iconset
++        COMMAND convert julius_256.png -resize 16x16 julius.iconset/icon_16.png
++        COMMAND convert julius_256.png -resize 32x32 julius.iconset/icon_32.png
++        COMMAND convert julius_256.png -resize 64x64 julius.iconset/icon_64.png
++        COMMAND convert julius_256.png -resize 128x128 julius.iconset/icon_128.png
++        COMMAND cp julius_256.png julius.iconset/icon_256.png
++        COMMAND cp julius_512.png julius.iconset/icon_512.png
++        COMMAND png2icns julius.icns julius.iconset/icon_{16,32,64,128,256,512}.png
+         COMMAND rm -R julius.iconset
+         WORKING_DIRECTORY ${PROJECT_SOURCE_DIR}/res)
+     set_source_files_properties(${PROJECT_SOURCE_DIR}/res/julius.icns PROPERTIES
+@@ -687,7 +685,6 @@ if(APPLE)
+     # setting variables that will populate Info.plist
+     set(MACOSX_BUNDLE_GUI_IDENTIFIER "com.github.bvschaik.julius")
+     set(MACOSX_BUNDLE_BUNDLE_NAME ${USER_FRIENDLY_NAME})
+-    set(CMAKE_OSX_DEPLOYMENT_TARGET "10.10" CACHE STRING "Minimum OS X deployment version" FORCE)
+     set(MACOSX_BUNDLE_ICON_FILE "julius.icns")
+     set(MACOSX_BUNDLE_BUNDLE_VERSION
+         "${PROJECT_VERSION_MAJOR}.${PROJECT_VERSION_MINOR}.${PROJECT_VERSION_PATCH}${VERSION_REVISION}")
+diff --git a/cmake/FindSDL2_mixer.cmake b/cmake/FindSDL2_mixer.cmake
+index 29cc683c..1fa28a1f 100644
+--- a/cmake/FindSDL2_mixer.cmake
++++ b/cmake/FindSDL2_mixer.cmake
+@@ -93,7 +93,6 @@ ELSE()
+         PATH_SUFFIXES include include/SDL2
+         PATHS ${SDL2_SEARCH_PATHS}
+       )
+-      set(SDL2_MIXER_INCLUDE_DIR "${SDL2_MIXER_INCLUDE_DIR}/Headers")
+     endif()
+ 
+     if(NOT APPLE OR NOT EXISTS "${SDL2_MIXER_INCLUDE_DIR}/SDL_mixer.h")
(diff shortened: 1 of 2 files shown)


pkgs/games/julius/default.nix
```diff
@@ -18,16 +25,42 @@ stdenv.mkDerivation rec {
...
-  buildInputs = [ SDL2 SDL2_mixer libpng ];
+  patches = [
+    # This disables the broken fixup_bundle and
+    # also disables unit tests.
+    ./darwin-fixes.patch
+  ];
+
+  postPatch = lib.optionalString stdenv.isDarwin ''
+    substituteInPlace src/platform/sound_device.c \
+      --replace-fail "#include \"SDL_mixer.h\"" "#include \"SDL2/SDL_mixer.h\""
```
SuperSandro2000, line 36: Can we add the correct ifdefs here and make that unconditional?

pkgs/games/julius/default.nix
```diff
@@ -5,8 +5,15 @@
 , SDL2_mixer
 , cmake
 , libpng
+, runCommand
+, darwin ? null
```
SuperSandro2000, line 9:
```suggestion
, darwin
```
that will never be null, even on linux

pkgs/games/julius/default.nix
```diff
@@ -5,8 +5,15 @@
...
 }:
-
+let
+  darwinSymlinks = runCommand "julius-darwin-build-symlinks" { } ''
+    mkdir -p $out/bin
+    ln -s /usr/bin/sips /usr/bin/iconutil /usr/bin/codesign $out/bin
```
SuperSandro2000, line 14: Wouldn't it be easier to append /usr/bin to PATH?

matteo-pacini: Got inspired by a similar approach used in`macvim`: https://github.com/NixOS/nixpkgs/blob/nixos-24.05/pkgs/applications/editors/vim/macvim.nix#L18-L27

Happy to add it to the PATH if it's a better solution

pkgs/games/julius/default.nix
matteo-pacini: Good to know, thanks!

pkgs/games/julius/default.nix
matteo-pacini: Added this as part of the patch itself, using ifdefs as suggested

pkgs/games/julius/default.nix
```diff
@@ -18,16 +25,37 @@ stdenv.mkDerivation rec {
...
+  ] ++ lib.optionals stdenv.isDarwin [
+    darwinSymlinks
+  ];
+
+  buildInputs = [
+    SDL2
+    SDL2_mixer
+    libpng
+  ] ++ lib.optionals stdenv.isDarwin [
+    darwin.apple_sdk_11_0.frameworks.Cocoa
```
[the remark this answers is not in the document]
SuperSandro2000: or just ``apple_sdk_11_0.callPackage``

matteo-pacini:
@paparodeo @SuperSandro2000 ready for re-review, thanks for the feedback so far!

pkgs/games/julius/darwin-fixes.patch
```diff
@@ -0,0 +1,65 @@
...
+@@ -869,9 +867,4 @@ else()
+         install(FILES "res/julius_512.png" DESTINATION "share/icons/hicolor/512x512/apps" RENAME "com.github.bvschaik.julius.png")
+     endif()
+ 
+-    # Unit tests
+-    if(${TARGET_PLATFORM} STREQUAL "default")
+-        enable_testing()
+-        add_subdirectory(test)
+-    endif()
+ endif()
```
paparodeo, line 48:
```suggestion
```
i removed this and build still works on x64 darwin.

pkgs/games/julius/darwin-fixes.patch
```diff
@@ -0,0 +1,65 @@
...
+index 0423e58f..ab2aef39 100644
+--- a/src/platform/sound_device.c
++++ b/src/platform/sound_device.c
+@@ -6,7 +6,12 @@
+ #include "platform/vita/vita.h"
+ 
+ #include "SDL.h"
++
++#ifdef __APPLE__
++#include "SDL2/SDL_mixer.h"
```
paparodeo, line 59: this seems like something is off in cmake/FindSDL2mixer.cmake tho this seems to work too

matteo-pacini: I think it expects a framework (?)

paparodeo: for some reason it is adding `Headers` to the path it located, which seems incorrect. this patch fixes that.
```diff
diff --git a/cmake/FindSDL2_mixer.cmake b/cmake/FindSDL2_mixer.cmake
index 29cc683..1fa28a1 100644
--- a/cmake/FindSDL2_mixer.cmake
+++ b/cmake/FindSDL2_mixer.cmake
@@ -93,7 +93,6 @@ ELSE()
         PATH_SUFFIXES include include/SDL2
         PATHS ${SDL2_SEARCH_PATHS}
       )
-      set(SDL2_MIXER_INCLUDE_DIR "${SDL2_MIXER_INCLUDE_DIR}/Headers")
     endif()
 
     if(NOT APPLE OR NOT EXISTS "${SDL2_MIXER_INCLUDE_DIR}/SDL_mixer.h")
```

pkgs/games/julius/darwin-fixes.patch
```diff
@@ -0,0 +1,55 @@
...
+     set_source_files_properties(${PROJECT_SOURCE_DIR}/res/julius.icns PROPERTIES
+@@ -687,7 +685,7 @@ if(APPLE)
+     # setting variables that will populate Info.plist
+     set(MACOSX_BUNDLE_GUI_IDENTIFIER "com.github.bvschaik.julius")
+     set(MACOSX_BUNDLE_BUNDLE_NAME ${USER_FRIENDLY_NAME})
+-    set(CMAKE_OSX_DEPLOYMENT_TARGET "10.10" CACHE STRING "Minimum OS X deployment version" FORCE)
++    set(CMAKE_OSX_DEPLOYMENT_TARGET "11.00" CACHE STRING "Minimum OS X deployment version" FORCE)
+     set(MACOSX_BUNDLE_ICON_FILE "julius.icns")
+     set(MACOSX_BUNDLE_BUNDLE_VERSION
+         "${PROJECT_VERSION_MAJOR}.${PROJECT_VERSION_MINOR}.${PROJECT_VERSION_PATCH}${VERSION_REVISION}")
```
paparodeo, line 38:
```suggestion
```
Is this needed? this just sets the minimum deployment target.

paparodeo: or commented out. i think nix uses `stdenv.hostPlatform.darwinMinVersion` which defaults to `10.12` for x64-darwin.

pkgs/games/julius/darwin-fixes.patch
matteo-pacini: Much better, thanks!

pkgs/games/julius/darwin-fixes.patch
matteo-pacini: @paparodeo I am bit puzzled about the override at top level  now- should I keep `11.0` even though `x86-64-darwin` have a lower min. deployment target?

paparodeo: so i think you can use the newer SDK and still have a lower deployment target. most of the reason i've seen overrideSDK used is when the build breaks due to using features from the newer sdk, like alligned_alloc. however, this package seems to build fine without the top-level override so it does seem unnecessary.

so i guess remove overrideSDK in top-level and then comment out or remove the line `set(CMAKE_OSX_DEPLOYMENT_TARGET "11.00" CACHE STRING "Minimum OS X deployment version" FORCE)` in the cmake file and then should be all done.

matteo-pacini: Done

paparodeo (approved):
LGTM

matteo-pacini:
@SuperSandro2000 friendly ping to ask for merge, thank you 🙏🏻

SuperSandro2000 (approved):

PR #289664, opened 2024-02-18
gitbutler: init at 0.10.11

hacker1024:
## Description of changes

This PR adds a [GitButler](https://gitbutler.com) package that builds from source.

Unfortunately quite a lot needs to be vendored, including the Yarn lockfile (#231513), which needs to be regenerated and may differ from the upstream pnpm lockfile.

Alternative to #289131.
Closes #288567.

macOS is untested as webkitgtk has been broken for years.

If you're interested in maintaining as well, feel free to let me know or make a PR into my branch adding yourself.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [x] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
- For non-Linux: Is sandboxing enabled in `nix.conf`? (See [Nix manual](https://nixos.org/manual/nix/stable/command-ref/conf-file.html))
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:
  - [NixOS test(s)](https://nixos.org/manual/nixos/unstable/index.html#sec-nixos-tests) (look inside [nixos/tests](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests))
  - and/or [package tests](https://nixos.org/manual/nixpkgs/unstable/#sec-package-tests)
  - or, for functions and "core" functionality, tests in [lib/tests](https://github.com/NixOS/nixpkgs/blob/master/lib/tests) or [pkgs/test](https://github.com/NixOS/nixpkgs/blob/master/pkgs/test)
  - made sure NixOS tests are [linked](https://nixos.org/manual/nixpkgs/unstable/#ssec-nixos-tests-linking) to the relevant packages
- [ ] Tested compilation of all packages that depend on this change using `nix-shell -p nixpkgs-review --run "nixpkgs-review rev HEAD"`. Note: all changes have to be committed, also see [nixpkgs-review usage](https://github.com/Mic92/nixpkgs-review#usage)
- [x] Tested basic functionality of all binary files (usually in `./result/bin/`)
[...]

PR #318101, opened 2024-06-07
gitbutler: init at 0.12.2

### LINKED ISSUE #288567: Package request: GitButler
**Project description**

GitButler is a git client that lets you work on multiple branches at the same time. It allows you to quickly organize file changes into separate branches while still having them applied to your working directory. You can then push branches individually to your remote, or directly create pull requests.

In a nutshell, it's a more flexible version of git add -p and git rebase -i, allowing you to efficiently multitask across branches.

**Metadata**

* homepage URL: https://gitbutler.com/
* source URL: https://github.com/gitbutlerapp/gitbutler
* license: FSL-1.0-MIT
* platforms: linux, darwin

getchoo:
## Description of changes

https://gitbutler.com/

Closes #288567

This is meant to be an alternative to #289664, with some improvements being:

- Adopting the new pnpm tooling from #290715
- Tauri's native bundles are used to reduce maintainer overhead
- The project is contained in a single derivation, simplifying things
- Updating to the latest release
- Applying multiple suggestions from the previous PR

cc @hacker1024 as the author of the previous PR, @SuperSandro2000 as the one who suggested the move to the new pnpm tooling

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---


matteo-pacini:
Result of `nixpkgs-review pr 318101` run on aarch64-darwin [1](https://github.com/Mic92/nixpkgs-review)
<details>
  <ul>
    <li>gitbutler</li>
  </ul>
</details>

Works fine on `aarch64-darwin`.

isabelroses (approved):
Tested and working on aarch64-darwin :)

pkgs/by-name/gi/gitbutler/package.nix
```diff
@@ -0,0 +1,177 @@
...
+      desktop-file-edit \
+        --set-comment "A Git client for simultaneous branches on top of your existing workflow." \
+        --set-key="Keywords" --set-value="git;" \
+        --set-key="StartupWMClass" --set-value="GitButler" \
+        $out/share/applications/git-butler.desktop
+    ''
+    + ''
+      runHook postInstall
+    '';
+
```
0x61nas, line 153: the UI doesn't render correctly on Linux with an Nvidia card. you have to make a wrapper script and disable the WebKit composting mode.

```suggestion

  # WEBKIT_DISABLE_COMPOSITING_MODE essential in NVIDIA + compositor https://github.com/NixOS/nixpkgs/issues/212064#issuecomment-1400202079
  postFixup = ''
    wrapProgram "$out/bin/gitbulter" \
      --set WEBKIT_DISABLE_COMPOSITING_MODE 1
  '';

```

getchoo: i have actually only tested this myself on nixos and nvidia, so i'm pretty surprised to see others having issues with this

unless it's very common, i don't think we should be adding this workaround to each and every tauri and/or webkitgtk app - `modrinth-app` for example does fine without this, along with `cinny`. it'd be much better left to individual configurations where these bugs actually occur

isabelroses: If this PR is still open in ~4 days, I'm happy to test on a RTX3060 to confirm.

0x61nas: >  unless it's very common

i have GTX 1660 TI with 535.154.05 drivers, and it has happened to me with every tauri app i tried with nix 'till now, including this one.
 
> i don't think we should be adding this workaround to each and every tauri and/or webkitgtk app - `modrinth-app` for example does fine without this, along with `cinny`. it'd be much better left to individual configurations where these bugs actually occur

the problem is that the error messages are pretty (generic), actually, it doesn't even give an error is just spamming the `stdout` with `AcceleratedSurfaceDMABuf was unable to construct a complete framebuffer` and you have to do some search to figure out what this means and how to fix it, I have no problem to do this with other imperative distros, but with Nix, I expect to things to just work so I can use them real fast with `nix run`, `nix shell` to get the job done without all this hassle

hacker1024 (commented):
Does this still need to replace the `libayatana-appindicator` library as was done [here](https://github.com/NixOS/nixpkgs/pull/289664/files#diff-cb032ab6fc41224fad7f65d1dd881c476f2dba0f75651255b72d5b3d500bb8aaR56-R58)?

getchoo:
> Does this still need to replace the `libayatana-appindicator` library as was done [here](https://github.com/NixOS/nixpkgs/pull/289664/files#diff-cb032ab6fc41224fad7f65d1dd881c476f2dba0f75651255b72d5b3d500bb8aaR56-R58)?

doesn't seem like it, no. fairly certain this is something to do with newer versions of tauri - i noticed it in `modrinth-app` as well

pkgs/by-name/gi/gitbutler/package.nix
getchoo: > i have GTX 1660 TI with 535.154.05 drivers

i think that's your problem: your drivers are outdated. the 535 series is no longer supported, so you should be upgrading to either the feature, production, or beta branch (545, 550, and 555 respectively) of the driver. i have a 1660 super on 550 (which `config.boot.kernelPackages.nvidiaPackages.latest` currently refers to) and this works as expected

> the problem is that the error messages are pretty (generic), actually, it doesn't even give an error is just spamming the stdout with AcceleratedSurfaceDMABuf was unable to construct a complete framebuffer and you have to do some search to figure out what this means and how to fix it, I have no problem to do this with other imperative package managers, but with Nix, I expect to things to just work

i can get this. after all, i've dealt with a lot of bugs like these myself with my card over the years 😆 

but i don't think it should be expected for nixpkgs maintainers to keep up with all of these possible issues and adopt the workarounds (that may or may not affect most users and could give an unexpected experience) into their packages. in this case specifically, it seems like it doesn't affect many people, requires a drastic action of disabling composting entirely, can be fixed with one environment variable, and may only be the result of an outdated driver. i don't think this is a fair tradeoff

i think a better solution for this case (and others) would be to help make users aware and avoid endless searching by making an issue. as 535 is actually still available in nixpkgs (i didn't know this until i started writing this part lol) and this may affect all webkitgtk apps, this should probably have it's own dedicated thread that will be easier to find, provide these workarounds, and notify maintainers so they can look into possible fixes as well

> so I can use them real fast with nix run, nix shell to get the job done without all this hassle

it should work with `nix-shell` if `WEBKIT_DISABLE_COMPOSITING_MODE` is already set in your environment - you can even do this declaratively via `environment.sessionVariables`

with `nix shell` you may need to use `--impure`, though

0x61nas: > you should be upgrading to either the feature, production, or beta branch (545, 550, and 555 respectively) of the driver. i have a 1660 super on 550 (which config.boot.kernelPackages.nvidiaPackages.latest currently refers to) and this works as expected

I can't, the latest versions have problems with my card, plus am not sure that this problem has anything to do with the GPU driver version

0x61nas (approved):

pkgs/by-name/gi/gitbutler/package.nix
```diff
@@ -0,0 +1,177 @@
...
+  };
+
+  meta = {
+    description = "Git client for simultaneous branches on top of your existing workflow";
+    homepage = "https://gitbutler.com";
+    changelog = "https://github.com/gitbutlerapp/gitbutler/releases/tag/release/${version}";
+    mainProgram = "git-butler";
+    license = lib.licenses.fsl11Mit;
+    maintainers = with lib.maintainers; [ getchoo ];
+    platforms = lib.platforms.linux ++ lib.platforms.darwin;
```
0x61nas, line 175:
```suggestion
    platforms = with lib.platforms; linux ++ darwin;
```

getchoo: the lack of `with lib.platforms;` here was intentional, as applying this diff results in `lib.platforms.darwin` being overshadowed by the top level `darwin` argument

image

0x61nas: Oh, i didn't think of that, that's something new for me thx

pkgs/by-name/gi/gitbutler/package.nix
Aleksanaa: @isabelroses Could you test this?

isabelroses: Ahh yeah totally forgot about this PR. I'm happy to.

isabelroses: I can confirm without this suggestion it still works perfectly.

0x61nas: maybe it is a problem with my particular card

Aleksanaa (approved):

PR #318172, opened 2024-06-08
libpng: fix build on FreeBSD

rhelmot:
## Description of changes

As per comment.

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [x] x86_64-freebsd
  - [ ] `sandbox = relaxed`
  - [x] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/development/libraries/libpng/default.nix
+++ b/pkgs/development/libraries/libpng/default.nix
@@ -20,7 +20,11 @@ in stdenv.mkDerivation (finalAttrs: {
     url = "mirror://sourceforge/libpng/libpng-${finalAttrs.version}.tar.xz";
     hash = "sha256-alygZSOSotfJ2yrltAIQhDwLvAgcvUEIJasAzFnxSmw=";
   };
-  postPatch = whenPatched "gunzip < ${patch_src} | patch -Np1";
+  postPatch = whenPatched "gunzip < ${patch_src} | patch -Np1"
+  + lib.optionalString stdenv.isFreeBSD ''
+
+    sed -i 1i'int feenableexcept(int __mask);' contrib/libtests/pngvalid.c
+  '';
 
   outputs = [ "out" "dev" "man" ];
   outputBin = "dev";


Ericson2314:
https://sourceforge.net/p/png-mng/mailman/message/36390652/ did you try this?

rhelmot:
That patch is already applied to libpng upstream. It doesn't help...

pkgs/development/libraries/libpng/default.nix
```diff
@@ -22,6 +22,12 @@ in stdenv.mkDerivation (finalAttrs: {
   };
   postPatch = whenPatched "gunzip < ${patch_src} | patch -Np1";
 
+  # feenableexcept is locked behind BSD_VISIBLE on FreeBSD, but enabling it
+  # messes up basically everything else. Just punt on implicit declarations.
+  env.CFLAGS = lib.optionalString stdenv.hostPlatform.isFreeBSD ''
+    -Wno-implicit-function-declaration
+  '';
+
```
paparodeo, line 30: can avoid mass rebuilds by using `configureFlags` and then this change can go into master.

```suggestion
  configureFlags = lib.optionals stdenv.hostPlatform.isFreeBSD [
    # feenableexcept is locked behind BSD_VISIBLE on FreeBSD, but enabling it
    # messes up basically everything else. Just punt on implicit declarations.
    "CFLAGS=-Wno-implicit-function-declaration"
  ];

```

rhelmot: _crosses fingers that I'm doing the rebase right_

rhelmot: Looks like that worked :) Let's see what ofborg has to say!

paparodeo (approved):
no rebuilds for `nix-build -A libpng` and `nix-build -A pkgsCross.x86_64-freebsd.libpng` builds successfully.

pkgs/development/libraries/libpng/default.nix
paparodeo: nice work.

alternatively:

```diff
diff --git a/pkgs/development/libraries/libpng/default.nix b/pkgs/development/libraries/libpng/default.nix
index 4a8c52ae279d..11e3fb31604f 100644
--- a/pkgs/development/libraries/libpng/default.nix
+++ b/pkgs/development/libraries/libpng/default.nix
@@ -20,7 +20,11 @@ in stdenv.mkDerivation (finalAttrs: {
     url = "mirror://sourceforge/libpng/libpng-${finalAttrs.version}.tar.xz";
     hash = "sha256-alygZSOSotfJ2yrltAIQhDwLvAgcvUEIJasAzFnxSmw=";
   };
-  postPatch = whenPatched "gunzip < ${patch_src} | patch -Np1";
+  postPatch = whenPatched "gunzip < ${patch_src} | patch -Np1"
+  + lib.optionalString stdenv.isFreeBSD ''
+
+    sed -i 1i'int feenableexcept(int __mask);' contrib/libtests/pngvalid.c
+  '';
 
   outputs = [ "out" "dev" "man" ];
   outputBin = "dev";
```

works too with no rebuilds (the extra \n before sed is required as to not end up an argument to patch). tho perhaps too nitpicky. either way works for me.

vcunat (commented):
Honestly I'd rather see you adding lines like
```c
int feenableexcept(int excepts);
```
into one of the headers.  I think this particular function should be OK without a declaration visible at call site, but this flag is global and can lead to subtle *runtime* breakages, e.g. see https://stackoverflow.com/a/437763/587396

rhelmot:
I've pushed the @paparodeo patch which does basically what you've asked :)

vcunat:
Ah, OK.  I haven't read it previously, as it was a "resolved" thread.

paparodeo (approved):
looks good with latest patch. no rebuilds for nix-build -A libpng and nix-build -A pkgsCross.x86_64-freebsd.libpng builds successfully.

vcunat:
Yes, I like this.  Especially as the issue only happens in a test.

PR #318188, opened 2024-06-08
geoserver: add service

rolfschr:
## Description of changes

This PR adds a new `services.geoserver` module which allows to run `geoserver` as a systemd service.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [x] Tested, as applicable:

---

### DIFF
--- a/nixos/modules/services/web-apps/geoserver.nix
+++ b/nixos/modules/services/web-apps/geoserver.nix
@@ -1,0 +1,76 @@
+{
+  config,
+  lib,
+  pkgs,
+  ...
+}:
+
+with lib;
+
+let
+
+  cfg = config.services.geoserver;
+
+in
+{
+  options = {
+    services.geoserver = {
+      enable = mkEnableOption "Geoserver service";
+
+      package = lib.mkPackageOption pkgs "geoserver" { };
+
+      dataDir = mkOption {
+        type = types.path;
+        default = "/var/lib/geoserver";
+        description = ''
+          The data directory of the server.
+
+          When a non-default location is used, the systemd unit will get a bind
+          mount to it. Furthermore, the directory must exist, be writeable for
+          the dynamic user and outside of protected directories (e.g. /home).
+        '';
+      };
+
+      jvmOpts = mkOption {
+        type = types.lines;
+        default = "";
+        description = "Any options passed to the JVM.";
+      };
+
+      jettyOpts = mkOption {
+        type = types.lines;
+        default = "";
+        example = "jetty.http.port=1234";
+        description = "Any options passed to the Jetty web server.";
+      };
+    };
+  };
+
+  config = mkIf cfg.enable {
+
+    systemd.services.geoserver = {
+      description = "Geoserver";
+
+      wantedBy = [ "multi-user.target" ];
+
+      environment = {
+        GEOSERVER_HOME = "${cfg.package}/share/geoserver";
+        GEOSERVER_DATA_DIR = "/var/lib/geoserver";
+        JAVA_OPTS = "${cfg.jvmOpts}";
+        JETTY_OPTS = "${cfg.jettyOpts}";
+      };
+
+      serviceConfig = {
+        ExecStart = "${cfg.package}/bin/geoserver-startup";
+        ExecStop = "${cfg.package}/bin/geoserver-shutdown";
+        DynamicUser = true;
+        NoNewPrivileges = true;
+        ProtectHome = true; # true=deny access to /home, /root, /run/user
+        StateDirectory = "geoserver";
+        BindPaths = [ "${cfg.dataDir}:/var/lib/geoserver" ];
+      };
+    };
+  };
+
+  meta.maintainers = with lib.maintainers; [ teams.geospatial.members ];
+}
--- a/nixos/doc/manual/release-notes/rl-2411.section.md
+++ b/nixos/doc/manual/release-notes/rl-2411.section.md
@@ -181,6 +181,8 @@
 
 - [Fedimint](https://github.com/fedimint/fedimint), a module based system for building federated applications (Federated E-Cash Mint). Available as [services.fedimintd](#opt-services.fedimintd).
 
+- [Geoserver](https://geoserver.org/), a server for sharing geospatial data. Available as [services.geoserver](#opt-services.geoserver.enable).
+
 ## Backward Incompatibilities {#sec-release-24.11-incompatibilities}
 
 - The `sound` options have been removed or renamed, as they had a lot of unintended side effects. See [below](#sec-release-24.11-migration-sound) for details.
--- a/nixos/modules/module-list.nix
+++ b/nixos/modules/module-list.nix
@@ -1419,6 +1419,7 @@
   ./services/web-apps/freshrss.nix
   ./services/web-apps/galene.nix
   ./services/web-apps/gancio.nix
+  ./services/web-apps/geoserver.nix
   ./services/web-apps/gerrit.nix
   ./services/web-apps/glance.nix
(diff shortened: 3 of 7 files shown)


rolfschr:
@imincik Could you have a look into this PR?

nixos/modules/services/web-apps/geoserver.nix
```diff
@@ -0,0 +1,76 @@
...
+      wantedBy = [ "multi-user.target" ];
+
+      environment = {
+        GEOSERVER_HOME = "${cfg.package}/share/geoserver";
+        GEOSERVER_DATA_DIR = "${cfg.datadir}";
+        JAVA_OPTS = "${cfg.jvmOpts}";
+      };
+
+      serviceConfig = {
+        User = cfg.user;
```
imincik, line 65: * I suggest to use systemd's `DynamicUser` instead of real user/group.

See for example here: https://github.com/NixOS/nixpkgs/blob/d51f8579f03a76718d643dd184ae964e6f0dd751/nixos/modules/services/matrix/matrix-sliding-sync.nix#L96

* It is also worth to consider to use `StateDirectory` instead of manual `datadir` configuration.

imincik (changes requested):
Thanks for your work @rolfschr . I left some comments with suggestions.

nixos/modules/services/web-apps/geoserver.nix
rolfschr: Ah, I considered both but refrained from it because I thought people might want to copy stuff into the data dir.
But fine with me, I'll adapt the service.

imincik: @rolfschr , are you still interested in finishing this PR ?

rolfschr:
@imincik Sorry for the delay. Yes, I was still working on this. As per your idea to use `StateDirectory` I played around a lot to have a nice way to either allow the data dir to be stored in `/var/lib` (and thus use `StateDirectory`) *or* have it outside `/var/lib` (but then *not* use `StateDirectory`). Unfortunately, I haven't come up with a nice solution and now I have removed the possibility to choose the data dir location all together. (I had issues with permissions, and also had to look into workarounds like [this](https://github.com/NixOS/nixpkgs/blob/932390c76b7e0d44656a648fe336a706f659cd42/nixos/modules/services/continuous-integration/jenkins/default.nix#L239)).

I also wanted to have a mean to provide jetty args (say, change the port). Providing `-D...` as java opts doesn't work (IMHO). I have tried out several things, but here again, I am not satisfied and, at least for now, opted for not doing anything. For reference, this is how I changed the port on a (prod) server:

```nix
  geoserverWithExtensions = ...;
  geoserver = geoserverWithExtensions.overrideAttrs (finalAttrs: previousAttrs:
    let
      previousPostInstall = if builtins.hasAttr "postInstall" previousAttrs then previousAttrs.postInstall else "";
    in {
      postInstall = previousPostInstall + ''
        sed -i "s/jetty.http.port=8080/jetty.http.port=12345/" $out/share/geoserver/start.ini
        '';
    }
  );
```

This is not nice which is why I wanted to provide a better means when implementing `services.geoserver`. However, as I said, I haven't been able to find a good way.

Thus ... I just pushed my current version. It is now very bare bones and I'm not sure whether this is sufficient to justify a new `service`. I'm happy to further work on this but would need further guidance.

h7x4:
@rolfschr I have seen some people use bind mounts to get around the datadir-not-in-var-lib issue in a nice way, but I'm not sure how well it plays together with `DynamicUser`. Maybe have a look at [`nixos/modules/services/audio/gonic.nix`](https://github.com/NixOS/nixpkgs/blob/e8d0a51d9aa7cf6bb72c3bcc2a106de2d2e7ad46/nixos/modules/services/audio/gonic.nix#L51-L69)? They seem to have gotten it working.

rolfschr:
@h7x4 Thanks for your input. I have been able to use a non-/var/lib directory as you suggested using bind mounts. This works as expected. The drawback I see here is that the log files will always print `/var/lib/geoserver` as data dir (because that's what the systemd unit thinks it is). Feel free to have a look into this.

As for the jetty configuration, I spent further time reading the docs and figured one can modify the jetty properties on the command line. However, a little patch is needed. I have added it in this PR. I'm not so happy that we now have a dedicated `*.patch` file as well as a `postPatch`-`sed`-based approach. I didn't want to add a patch file (yet).

I have now several commits here which I haven't squashed yet. Please have a look into this (also @imincik) and let me know what you all think.

rolfschr:
@imincik Could you have a look here, please? I'd like to finish this PR.

rolfschr:
I fixed the formatting but I think ofBorg didn't realize that. I assume there will be change requests so the checks will rerun anyways.

Please have a look into this now.

rolfschr:
@NixOS/geospatial Can you please look into this? I am waiting for feedback here. I had to resolve the merge conflict repeatedly because the rl-2411.section.md keeps getting updated. This PR stems from June.

Note that I have added an additional commit to disable the check for the gdal extension. It does in fact not work but I somehow believe/hope it will work auto-magically one day again. It seems to be related to libxml2 having disabled HTTP support by default. This seems to be [known](https://github.com/NixOS/nixpkgs/pull/340707#issuecomment-2361894717) issue in nixpkgs but I haven't managed to fix it. Actually, the problem [should not occur in the first place](https://github.com/NixOS/nixpkgs/blob/54cce652cb480258b12f15b2915e7a6c79ac51ae/pkgs/development/libraries/gdal/default.nix#L183) so I wonder what is going on. I tried to override in at other locations (gdal, netcdf) but to no avail. Maybe someone has a fix to propose but I do have to say I spent so much time already that I have now decided to simply disable the test for now (to make the build green) and keep an eye on this.

--- 

Here is an extract of the test log output:
```bash
vm-test-run-geoserver> Native library load failed.
vm-test-run-geoserver> java.lang.UnsatisfiedLinkError: /nix/store/qpi9ac8rkp51wbx87f38nj3chcd84h89-gdal-3.9.2/lib/jni/libgdalalljni.so: /nix/store/8627wpd4mbygb4p5ikzy7whvqhclqnc4-libspatialite-5.1.0/lib/libspatialite.so.8: undefined symbol: xmlNanoHTTPCleanup, version LIBXML2_2.4.30
vm-test-run-geoserver> Oct 12, 2024 1:54:14 PM it.geosolutions.imageio.gdalframework.GDALUtilities loadGDAL
vm-test-run-geoserver> WARNING: Failed to load the GDAL native libs. This is not a problem unless you need to use the GDAL plugins: they won't be enabled.
```

---

> As for the jetty configuration, I spent further time reading the docs and figured one can modify the jetty properties on the command line. However, a little patch is needed. I have added it in this PR. I'm not so happy that we now have a dedicated `*.patch` file as well as a `postPatch`-`sed`-based approach. I didn't want to add a patch file (yet).
> 
> I have now several commits here which I haven't squashed yet. Please have a look into this (also @imincik) and let me know what you all think.

Commits squashed and patch file added.

autra:
>  It seems to be related to libxml2 having disabled HTTP support by default.

This will be resolved soon I hope, with https://github.com/NixOS/nixpkgs/pull/350694. Maybe you can try to rebase this PR on that and see if it fixes it?

Apart from that the code looks good, but I didn't test and I'm not a geoserver user, so I'd prefer someone with more exp than me to approve.

> This PR stems from June.

Once, it took me a year and a half to get a PR merged (not on nixpkgs though), you still have a long way to go before beating this 😜

PR #318204, opened 2024-06-08
nixos/looking-glass: init module

calebstewart:
### Description of changes
Add module for automatically configuring the Looking Glass client and associated shared memory and/or KVMFR devices for interacting with guest VMs. The steps done by this module are taken from the [official installation instructions](https://looking-glass.io/docs/B7-rc1/install/).

It's worth noting that I don't know much about nor do I use AppArmor beyond knowing that it exists and what it is. I *believe* that I configured the policy as recommended by the installation instructions, but due to my unfamiliarity with AppArmor paired with the abstractions placed on top by NixOS, I'm not certain and I haven't tested that part specifically.

I added myself as a maintainer for the module, and I'm happy to do that. However, I'm not sure about the etiquette for notifying or roping in the maintainers of the `looking-glass-client` package. if they would normally be added here as well, I'm happy to put them on the list as well (or instead if preferred).

This is my first module, so please let me know if I'm doing something in a weird way. I wrote this all up initially as an internal module for my own configuration, but thought it might be more broadly useful. :)

##### Reviewed points

- [x] module path fits the guidelines
- [ ] module tests succeed on ARCHITECTURE
- [ ] options have appropriate types
- [x] options have default
- [x] options have example
- [x] options have descriptions
- [x] No unneeded package is added to `environment.systemPackages`
- [x] module documentation is declared in `meta.doc`

##### Possible improvements
I'm unsure about how to handle the changes needed in `qemu.conf` for `libvirtd`. I've opted to have an option named `kvmfr.configureDeviceACLs` which is disabled by default, but will append to `virtualisation.libvirtd.qemu.verbatimConfig` when enabled.

Personally, I've never needed to modify `qemu.conf` aside from this specific instance, so I expect it to be fairly common for someone to enable this option. However, given that it could end up being added after a configuration from somewhere else, and inadvertently overriding those settings, I've chosen to make it disabled by default. I tried to be very clear in the docs, and in the description for the option what exactly it does and when/why you would want to enable it.

If anyone has a better suggestion, I'm all ears.


nixos/modules/programs/looking-glass/default.nix
```diff
@@ -0,0 +1,188 @@
...
+  # Generate the looking-glass-client configuration where boolean values
+  # are encoded as "yes" or "no".
+  generateConfig = settings: lib.generators.toINI {
+    mkKeyValue = lib.generators.mkKeyValueDefault {
+      mkValueString = v:
+        if (true == v) then "yes"
+        else if (false == v) then "no"
+        else (lib.generators.mkValueStringDefault {} v);
+    } "=";
+  } settings;
```
jian-lin, line 14: It is not recommended to add all config options of a specific program into its NixOS module due to issues such as maintenance burden.  For more info, [rfc 42][rfc] is a good read.

Only some important settings should be defined in the NixOS module.  This [doc] guides you how to do so and here is an [example][example].

[rfc]: https://github.com/NixOS/rfcs/blob/master/rfcs/0042-config-option.md
[doc]: https://nixos.org/manual/nixos/unstable/#sec-settings-options
[example]: https://nixos.org/manual/nixos/unstable/#sec-settings-attrs-options

calebstewart: I appreciate the feedback! I'll take a look at that documentation, and revisit.

Issue #317055, opened 2024-06-04
Build failure: zig-0.12.0


khaneliman:
### Steps To Reproduce

1. build zig-0.12.0 

### Build log

```
@nix { "action": "setPhase", "phase": "unpackPhase" }
Running phase: unpackPhase
unpacking source archive /nix/store/3ilc0q7g7253778mhyg3gcvb8i4f7mh2-source
source root is source
@nix { "action": "setPhase", "phase": "patchPhase" }
Running phase: patchPhase
substituteStream(): WARNING: '--replace' is deprecated, use --replace-{fail,warn,quiet}. (file 'lib/std/zig/system.zig')
@nix { "action": "setPhase", "phase": "updateAutotoolsGnuConfigScriptsPhase" }
Running phase: updateAutotoolsGnuConfigScriptsPhase
@nix { "action": "setPhase", "phase": "configurePhase" }
Running phase: configurePhase
fixing cmake files...
[...]

PR #318226, opened 2024-06-08
cmake:  fix strictDeps

paparodeo:
## Description of changes

patch cmake (and meson) to use a new `NIXPKGS_CMAKE_PREFIX_PATH` variable which is like `CMAKE_PREFIX_PATH` but doesn't provide a search path to `find_program`

cmake uses `CMAKE_PREFIX_PATH` to search for libraries, header files, binaries, cmake modules, etc. This is mostly fine except for binaries. We set `CMAKE_PREFIX_PATH` to the derivation output locations, which results in cmake finding and using binaries which are not set in PATH. when strictDeps is true, nativeBuildInputs is not added to `CMAKE_PREFIX_PATH` but buildInputs so binaries are first searched for in the buildInputs packages before falling back to PATH.

This change attempts to fix the cmake search path such that cmake will only locate binaries accessible via PATH.

the meson patch is required when it falls back on cmake to find modules, it needs to know about the `NIXPKGS_CMAKE_PREFIX_PATH` env variable.

the following PRs fix various packages that enable `strictDeps` but fail to build due to packages missing from `nativeBuildInputs`

- https://github.com/NixOS/nixpkgs/pull/324545
- https://github.com/NixOS/nixpkgs/pull/324951
- https://github.com/NixOS/nixpkgs/pull/329977

### testing
- 908 builds where cmake in nativeBuildInputs and  strictDeps set
- 54 builds where cmake + meson in nativeBuildInputs
- random sample of 865 builds where cmake in nativeBuildInputs and strictDeps not set (out of 3480 attributes)

tests were usually run on master with the cmake modifications contained in a new cmake2 package to skip unnecessary rebuilds. packages that required cmake were then built using an override to the modified cmake or manually edited/hacked when overriding was not supported, eg:

```
# breaks with clang when using strictDeps and unmodified cmake
nix-build -E 'with import ./.{}; castxml.override { stdenv = clangStdenv; cmake = cmake2; }'
# meson build but uses cmake to locate packages
nix-build -E 'with import ./.{}; wf-touch.override { cmake = cmake2; meson = meson2; }'
[...]
```
packages that require cmake were found using a variation of the following script run from the root of the source tree
<details><summary>example script to get a json list of packages using cmake but not strict deps</summary>
 
```sh
nix-instantiate --show-trace --eval --strict --json -A devs - <<EOF
{
  paths ? (import ./pkgs/top-level/release-attrpaths-superset.nix { }).paths,
  pkgs ? (import ./. { }),
}:

let
  inherit (pkgs) lib;
  inherit (builtins) currentSystem tryEval;

  usesCmake =
    let
      isCmakeOverridable = p: p ? override && lib.hasAttr "cmake" (lib.functionArgs p.override);

      hasCmakeNativeBuildInputs =
        p: p ? nativeBuildInputs && lib.lists.any (x: x == pkgs.cmake) p.nativeBuildInputs;

      usesCmake' =
        p:
        p != null
        && !(p.strictDeps or false)     # reverse logic to get list of only strict deps
        && !(p.meta.broken or false)
        && isCmakeOverridable p         # comment out remove overrideable filter 
        && lib.meta.availableOn p currentSystem
        && hasCmakeNativeBuildInputs p;
    in
    path: (tryEval (usesCmake' (lib.attrByPath path null pkgs))).value;
in
{
  devs = map (f: lib.concatStringsSep "." f) (lib.filter usesCmake paths);
}
EOF
```
</details>

then iterate over the packages (or a sample of) passing in the modified `cmake2`

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [x] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/by-name/cm/cmake/setup-hook.sh
+++ b/pkgs/by-name/cm/cmake/setup-hook.sh
@@ -151,22 +151,22 @@ makeCmakeFindLibs(){
   for flag in ${NIX_CFLAGS_COMPILE-} ${NIX_LDFLAGS-}; do
     if test -n "$isystem_seen" && test -d "$flag"; then
       isystem_seen=
-      export CMAKE_INCLUDE_PATH="${CMAKE_INCLUDE_PATH-}${CMAKE_INCLUDE_PATH:+:}${flag}"
+      addToSearchPath CMAKE_INCLUDE_PATH "${flag}"
     elif test -n "$iframework_seen" && test -d "$flag"; then
       iframework_seen=
-      export CMAKE_FRAMEWORK_PATH="${CMAKE_FRAMEWORK_PATH-}${CMAKE_FRAMEWORK_PATH:+:}${flag}"
+      addToSearchPath CMAKE_FRAMEWORK_PATH "${flag}"
     else
       isystem_seen=
       iframework_seen=
       case $flag in
         -I*)
-          export CMAKE_INCLUDE_PATH="${CMAKE_INCLUDE_PATH-}${CMAKE_INCLUDE_PATH:+:}${flag:2}"
+          addToSearchPath CMAKE_INCLUDE_PATH "${flag:2}"
           ;;
         -L*)
-          export CMAKE_LIBRARY_PATH="${CMAKE_LIBRARY_PATH-}${CMAKE_LIBRARY_PATH:+:}${flag:2}"
+          addToSearchPath CMAKE_LIBRARY_PATH "${flag:2}"
           ;;
         -F*)
-          export CMAKE_FRAMEWORK_PATH="${CMAKE_FRAMEWORK_PATH-}${CMAKE_FRAMEWORK_PATH:+:}${flag:2}"
+          addToSearchPath CMAKE_FRAMEWORK_PATH "${flag:2}"
           ;;
         -isystem)
           isystem_seen=1
--- a/pkgs/by-name/cm/cmake/setup-hook.sh
+++ b/pkgs/by-name/cm/cmake/setup-hook.sh
@@ -1,5 +1,7 @@
 addCMakeParams() {
-    addToSearchPath CMAKE_PREFIX_PATH $1
+    # NIXPKGS_CMAKE_PREFIX_PATH is like CMAKE_PREFIX_PATH except cmake
+    # will not search it for programs
+    addToSearchPath NIXPKGS_CMAKE_PREFIX_PATH $1
 }
 
 fixCmakeFiles() {
--- a/pkgs/by-name/cm/cmake/000-nixpkgs-cmake-prefix-path.diff
+++ b/pkgs/by-name/cm/cmake/000-nixpkgs-cmake-prefix-path.diff
@@ -1,0 +1,28 @@
+diff --git a/Source/cmFindBase.cxx b/Source/cmFindBase.cxx
+index 8840cdcb..c34b7ee9 100644
+--- a/Source/cmFindBase.cxx
++++ b/Source/cmFindBase.cxx
+@@ -280,6 +280,11 @@ void cmFindBase::FillCMakeEnvironmentPath()
+   // Add CMAKE_*_PATH environment variables
+   std::string var = cmStrCat("CMAKE_", this->CMakePathName, "_PATH");
+   paths.AddEnvPrefixPath("CMAKE_PREFIX_PATH");
++  if (this->CMakePathName != "PROGRAM") {
++    // Like CMAKE_PREFIX_PATH except when searching for programs. Programs need
++    // to be located via PATH
++    paths.AddEnvPrefixPath("NIXPKGS_CMAKE_PREFIX_PATH");
++  }
+   paths.AddEnvPath(var);
+ 
+   if (this->CMakePathName == "PROGRAM") {
+diff --git a/Source/cmFindPackageCommand.cxx b/Source/cmFindPackageCommand.cxx
+index 9b51b1ad..6acc676c 100644
+--- a/Source/cmFindPackageCommand.cxx
++++ b/Source/cmFindPackageCommand.cxx
+@@ -2039,6 +2039,7 @@ void cmFindPackageCommand::FillPrefixesCMakeEnvironment()
+ 
+   // And now the general CMake environment variables
+   paths.AddEnvPath("CMAKE_PREFIX_PATH");
++  paths.AddEnvPath("NIXPKGS_CMAKE_PREFIX_PATH");
+   if (this->DebugMode) {
+     debugBuffer = cmStrCat(debugBuffer,
+                            "CMAKE_PREFIX_PATH env variable "
--- a/pkgs/by-name/cm/cmake/package.nix
+++ b/pkgs/by-name/cm/cmake/package.nix
@@ -56,6 +56,9 @@ stdenv.mkDerivation (finalAttrs: {
   };
 
   patches = [
+    # Add NIXPKGS_CMAKE_PREFIX_PATH to cmake which is like CMAKE_PREFIX_PATH
+    # except it is not searched for programs
+    ./000-nixpkgs-cmake-prefix-path.diff
     # Don't search in non-Nix locations such as /usr, but do search in our libc.
     ./001-search-path.diff
     # Don't depend on frameworks.
--- a/pkgs/by-name/me/meson/000-nixpkgs-cmake-prefix-path.patch
+++ b/pkgs/by-name/me/meson/000-nixpkgs-cmake-prefix-path.patch
@@ -1,0 +1,12 @@
+diff --git a/mesonbuild/dependencies/data/CMakePathInfo.txt b/mesonbuild/dependencies/data/CMakePathInfo.txt
+index 662ec58..4d5f4e4 100644
+--- a/mesonbuild/dependencies/data/CMakePathInfo.txt
++++ b/mesonbuild/dependencies/data/CMakePathInfo.txt
+@@ -5,6 +5,7 @@ list(APPEND TMP_PATHS_LIST ${CMAKE_PREFIX_PATH})
+ list(APPEND TMP_PATHS_LIST ${CMAKE_FRAMEWORK_PATH})
+ list(APPEND TMP_PATHS_LIST ${CMAKE_APPBUNDLE_PATH})
+ list(APPEND TMP_PATHS_LIST $ENV{CMAKE_PREFIX_PATH})
++list(APPEND TMP_PATHS_LIST $ENV{NIXPKGS_CMAKE_PREFIX_PATH})
+ list(APPEND TMP_PATHS_LIST $ENV{CMAKE_FRAMEWORK_PATH})
+ list(APPEND TMP_PATHS_LIST $ENV{CMAKE_APPBUNDLE_PATH})
+ list(APPEND TMP_PATHS_LIST ${CMAKE_SYSTEM_PREFIX_PATH})
--- a/pkgs/by-name/me/meson/package.nix
+++ b/pkgs/by-name/me/meson/package.nix
@@ -31,6 +31,9 @@ python3.pkgs.buildPythonApplication rec {
   };
 
   patches = [
+    # Nixpkgs cmake uses NIXPKGS_CMAKE_PREFIX_PATH for the search path
+    ./000-nixpkgs-cmake-prefix-path.patch
+
     # In typical distributions, RPATH is only needed for internal libraries so
     # meson removes everything else. With Nix, the locations of libraries
     # are not as predictable, therefore we need to keep them in the RPATH.


paparodeo:
closes: https://github.com/NixOS/nixpkgs/issues/317055

paparodeo:
cc cmake maintainers: @ttuegel @LnL7 @AndersonTorres

paparodeo:
after a little matrix discussion it seems that we shouldn't be using [CMAKE_PREFIX_PATH](https://cmake.org/cmake/help/latest/envvar/CMAKE_PREFIX_PATH.html#envvar:CMAKE_PREFIX_PATH) to add the buildInputs path as that ends up adding the `bin` directory which seems incorrect.  so instead just add the `include` and `lib` directories to the paths seems better.

```diff
diff --git a/pkgs/by-name/cm/cmake/setup-hook.sh b/pkgs/by-name/cm/cmake/setup-hook.sh
index b28ed42b6896..d84ff6f99edd 100755
--- a/pkgs/by-name/cm/cmake/setup-hook.sh
+++ b/pkgs/by-name/cm/cmake/setup-hook.sh
@@ -1,5 +1,7 @@
 addCMakeParams() {
-    addToSearchPath CMAKE_PREFIX_PATH $1
+    [[ -d "$1/include" ]] && addToSearchPath CMAKE_INCLUDE_PATH "$1/include"
+    [[ -d "$1/lib" ]] && addToSearchPath CMAKE_LIBRARY_PATH "$1/lib"
+    return 0
 }
 
 fixCmakeFiles() {
```

paparodeo:
so it used to just use `CMAKE_{INCLUDE,LIBRARY}_PATH` but that was changed in https://github.com/NixOS/nixpkgs/commit/e0ff817c4071bdd1b4b8bc003110fd2b75c14e7a#diff-a923f95cc35b1a7b2a5853e553cca08989536827d1b7f39942e034c2627db7f7 to use `CMAKE_PREFIX_PATH` in 2010 but perhaps things have changed enough since then to make it work.

AndersonTorres:
> so it used to just use `CMAKE_{INCLUDE,LIBRARY}_PATH` but that was changed in [e0ff817#diff-a923f95cc35b1a7b2a5853e553cca08989536827d1b7f39942e034c2627db7f7](https://github.com/NixOS/nixpkgs/commit/e0ff817c4071bdd1b4b8bc003110fd2b75c14e7a#diff-a923f95cc35b1a7b2a5853e553cca08989536827d1b7f39942e034c2627db7f7) to use `CMAKE_PREFIX_PATH` in 2010 but perhaps things have changed enough since then to make it work.

This is clearly KDE-specific. Should it be "generalized" outside KDE?

paparodeo:
with a modified cmake setup-hooks only specifying CMAKE_LIBRARY_PATH and CMAKE_INCLUDE PATH i was getting failures with `openexr_3`

```
CMake Warning at cmake/OpenEXRSetup.cmake:272 (find_package):
  By not providing "FindImath.cmake" in CMAKE_MODULE_PATH this project has
  asked CMake to find a package configuration file provided by "Imath", but
  CMake did not find one.

  Could not find a package configuration file provided by "Imath" (requested
  version 3.1) with any of the following names:

    ImathConfig.cmake
    imath-config.cmake
```

and after a bit of trial and error i found that after adding `addToSearchPath CMAKE_PREFIX_PATH "$1/lib/cmake"` cmake found the module and the build succeeded.

```bash
addCMakeParams() {
    addToSearchPath CMAKE_INCLUDE_PATH "$1/include"
    addToSearchPath CMAKE_LIBRARY_PATH "$1/lib"
    addToSearchPath CMAKE_PREFIX_PATH "$1/lib/cmake"
} 
```
this configuration also worked for zig / castxml when build with clangStdenv and strictDeps enabled -- actually for all 3 packages only having `addToSearchPath CMAKE_PREFIX_PATH "$1/lib/cmake"` was sufficient.

will try to build `plasma-desktop` again and see what happens.

paparodeo:
was not finding pkgconfig so stiched strategy and just patched out using `bin` as a prefix if `NIX_CMAKE_DISABLE_PREFIX_BIN` is set, and set it when calling cmake as part of the configure. kind of gross but does the trick.

paparodeo:
last change removed the cmake patch in favor of modifying `addCMakeParams` with various paths found mostly via trial and error and looking over the cmake source.

AndersonTorres (commented):
Looks nicer!

SomeoneSerge:
> changed ... to use CMAKE_PREFIX_PATH in 2010 but perhaps things have changed enough since then to make it work.

This is mainly used for `find_package` (`*Config.cmake` files) and it was big surprise to us in https://github.com/NixOS/nixpkgs/pull/329452 that it also affects `find_program`

pkgs/by-name/cm/cmake/setup-hook.sh
SomeoneSerge, line 1: AFAIU strictDeps are the default, but maybe we should still do something like `[[ -n "${strictDeps-} ]] && old behaviour`

SomeoneSerge, line 1: > [cmFindPackageCommand.cxx#L2715](https://github.com/Kitware/CMake/blob/01a111d97f71cd9f5534e2cc36024209c6b34708/Source/cmFindPackageCommand.cxx#L2715)
looks like there are a few more paths. lib{32,64,x32} and lib/ARCH whatever ARCH expands to

I haven't seen many paths other than `{lib,share}/cmake` (except for python packages), but it seems like we're going to have to mirror a bit of cmake's logic and we already found some of it surprising/unpredictable. It could easily be that cmake abuses these prefices for more cursed things, but I think I still like the approach in this pr (limiting cmake's vision)

pkgs/by-name/cm/cmake/setup-hook.sh
```diff
@@ -1,5 +1,14 @@
 addCMakeParams() {
-    addToSearchPath CMAKE_PREFIX_PATH $1
+    addToSearchPath CMAKE_INCLUDE_PATH "$1/include"
+    addToSearchPath CMAKE_INCLUDE_PATH "$1"
```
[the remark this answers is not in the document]
paparodeo: because this replicates what `CMAKE_PREFIX_PATH` does. it adds `include` to the search path first and then the bare directory last. there is code which relies on this working. looking at the edit in https://github.com/NixOS/nixpkgs/pull/318226#issuecomment-2175370938 it seems like plasma5Packages.libksysguard  has some custom module finder that requires the bare directory to find headers because it adds `include` suffix to the search path...

https://github.com/KDE/libksysguard/blob/b1b4989cf630122f3d58868a884e624a620eaf03/cmake/FindNL.cmake#L36-L38

pkgs/by-name/cm/cmake/setup-hook.sh
paparodeo: strictDeps is not the default. i considered conditioning on strictDeps but it seemed better to just get the paths correct. using CMAKE_PREFIX_PATH is wrong both with and without strictDeps as cmake will pull executables from the there before it searches in PATH.

paparodeo: nixpkgs will move lib64 to lib. not sure about the others. when leaving out the lib32/64/x64 out of the search paths i reasoned that if they were required then the nixpkgs package was broken as it should just be installing things into lib.

i did have a version of this change where i just patched cmake so that it didn't search for bin in CMAKE_PREFIX_PATH but i opted to revert it and go the setup-hook route as changing the cmake behavior, while is what we want for builds within nixpkgs, is not what is wanted for builds using nixpkgs cmake not in nixpkgs.

thanks for looking at this and for the comments.

pkgs/by-name/cm/cmake/setup-hook.sh
```diff
@@ -1,5 +1,14 @@
...
+
+    addToSearchPath CMAKE_LIBRARY_PATH "$1/lib"
+    addToSearchPath CMAKE_LIBRARY_PATH "$1"
+
+    addToSearchPath CMAKE_PREFIX_PATH "$1/lib/cmake"
+    addToSearchPath CMAKE_PREFIX_PATH "$1/share/cmake"
+    addToSearchPath CMAKE_PREFIX_PATH "$1/lib"
+    addToSearchPath CMAKE_PREFIX_PATH "$1/share"
```
emilazy, line 11: There are some packages that put executables in `/lib` and `/share` as a hacky kind of `/libexec`; I’m a little worried that could cause issues? Still a big improvement over the status quo though obviously.

emilazy: We could probably fix `extra-cmake-modules` to install into `/lib/cmake` if that’s the only issue with `/lib`. Did you encounter anything that needed bare `/share`?

emilazy: (FWIW I don’t feel fully confident in the CMake hook stuff enough to get this through to the finish line myself, but if @SomeoneSerge is willing hopefully we can collaborate on getting this merged?)

paparodeo: > We could probably fix `extra-cmake-modules` to install into `/lib/cmake` if that’s the only issue with `/lib`. Did you encounter anything that needed bare `/share`?

in the early commit i have annotated that vulkan headers / vulkan-loader needed it. given that these are default search paths for cmake removing them would well, remove the default search paths

> There are some packages that put executables in /lib and /share as a hacky kind of /libexec; I’m a little worried that could cause issues? Still a big improvement over the status quo though obviously.

the change i had which patched cmake should deal with that, i think: https://github.com/NixOS/nixpkgs/pull/318226/commits/52fa78963bf3615fe5e677667a2a021ac628b1f9 it doesn't add any paths from CMAKE_PREFIX_PATH when a binary is requested.  I opted to not use that change as i was triggering the new nixpkgs behaviour with an env var but i thought that would confuse people if they were trying to do builds via nix-shell and invoking cmake by hand and not via runPhase.

pkgs/by-name/cm/cmake/000-nixpkgs-disable-bin-prefix.diff
```diff
@@ -0,0 +1,31 @@
...
++
+ void cmSearchPath::AddPrefixPaths(const std::vector<std::string>& paths,
+                                   const char* base)
+ {
+@@ -165,6 +175,9 @@ void cmSearchPath::AddPrefixPaths(const std::vector<std::string>& paths,
+     subdir = "lib";
+   } else if (this->FC->CMakePathName == "FRAMEWORK") {
+     subdir.clear(); // ? what to do for frameworks ?
++  } else if (IsBinPrefixDisabled()) {
++      // don't look at the bin component of CMAKE_PREFIX_PATH
```
paparodeo: this should read -- `don't use any component of CMAKE_PREFIX_PATH when a program is requested`

pkgs/by-name/cm/cmake/setup-hook.sh
emilazy: Thanks for elaborating. I think the patch behaviour seems like it’d get us closer to where we want to be overall. Maybe we could set the `NIXPKGS_CMAKE_DISABLE_BIN_PREFIX` variable in a hook instead of directly in the `cmake` invocation? Nix development shells have a bunch of impurities and weirdnesses anyway, so I think it’s okay if we just make the happy path work.

pkgs/by-name/cm/cmake/000-nixpkgs-disable-bin-prefix.diff
emilazy: Right, it’d be good to have a slightly clearer comment and maybe environment variable name, just for the sake of anyone who has to poke at this stuff later on.

pkgs/by-name/cm/cmake/setup-hook.sh
SomeoneSerge: > but if @SomeoneSerge is willing hopefully we can collaborate on getting this merged?

I'm very much interested in getting this merged. I expect a _lot_ of fall out, e.g. I'd expect there to be projects which abuse some kind of generic `find_file` which we'd break, but I'd rather we tightened the bolts and handled those as special cases

pkgs/by-name/cm/cmake/setup-hook.sh
```diff
@@ -1,5 +1,14 @@
 addCMakeParams() {
-    addToSearchPath CMAKE_PREFIX_PATH $1
+    addToSearchPath CMAKE_INCLUDE_PATH "$1/include"
```
SomeoneSerge, line 2: https://cmake.org/cmake/help/latest/variable/CMAKE_INCLUDE_PATH.html#variable:CMAKE_INCLUDE_PATH says "semi-colon separated", so I think we need

```suggestion
    addToSearchPathWithCustomDelimiter CMAKE_INCLUDE_PATH ";" "$1/include"
```

I wonder if it'd make sense to add an alias, something like `addCmakeSearchPath`, because this construct will already be repeated in `cmake/setup-hooks.h` and we're also using this in `pkgs/development/cuda-modules/setup-hooks/setup-cuda-hook.sh`

pkgs/by-name/cm/cmake/setup-hook.sh
SomeoneSerge: You're right, and this is disappointing. I then would like to look into https://github.com/NixOS/nixpkgs/issues/178468 after your PR

emilazy: It wouldn’t change impure workflows since it’s conditioned on a Nixpkgs‐specific environment variable, and it would avoid further potential fallout because packages do sometimes ship binaries in `/lib` and `/share`. What we want is just to separate out CMake paths from binary paths and the patch would accomplish that precisely. I worry about arcane failures down the line if we do a half‐measure. If we set it in a hook like I proposed in https://github.com/NixOS/nixpkgs/pull/318226#discussion_r1689604322 then it’d work properly for `nix develop` derivation debugging too. I lean towards doing the patch for that reason. That said, obviously not having to patch the code is an advantage in itself.

emilazy: If you want to work on long‐delayed default switches then check out https://github.com/NixOS/nixpkgs/pull/318614 too :)

pkgs/by-name/cm/cmake/setup-hook.sh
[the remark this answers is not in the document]
SomeoneSerge: I think the way `PATH_SUFFIXES` work is they are checked in addition to the prefix, but maybe I'm wrong

pkgs/by-name/cm/cmake/setup-hook.sh
emilazy: That said I’d also be okay if we just standardized on one specific CMake directory to add and fixed all the derivations that install `.cmake` files. There’s no good reason to have them fragmented.

pkgs/by-name/cm/cmake/000-nixpkgs-disable-bin-prefix.diff
SomeoneSerge: Additionally, this doesn't explain why anyone would consider disabling this: "prevent find_program from looking into CMAKE_PREFIX_PATH to avoid picking up the wrong platform's executables"

pkgs/by-name/cm/cmake/setup-hook.sh
SomeoneSerge: > If we set it in a hook like I proposed in https://github.com/NixOS/nixpkgs/pull/318226#discussion_r1689604322 then it’d work properly for nix develop derivation debugging too.

That actually sounds convincing. If the goal is to just fix cross-compilation of cmake projects in Nixpkgs, the patch + env variable are the "smaller" and cheaper change than overhauling the search paths.

From the point of view of long term maintenance, I think the blanket `addSearchPath CMAKE_PREFIX_PATH` is maybe something we do want to get rid of, since it has so many side-effects. To reiterate, I didn't realize up until now that it's used for anything other than `find_package()`

pkgs/by-name/cm/cmake/setup-hook.sh
paparodeo: > https://cmake.org/cmake/help/latest/variable/CMAKE_INCLUDE_PATH.html#variable:CMAKE_INCLUDE_PATH says "semi-colon separated", so I think we need
...

you're looking at the wrong page. that is not the environment variable but the cmake variable. https://cmake.org/cmake/help/latest/envvar/CMAKE_INCLUDE_PATH.html#envvar:CMAKE_INCLUDE_PATH

it is not semi-colon separated but normal unix colon separated -- unless on windows.

pkgs/by-name/cm/cmake/setup-hook.sh
paparodeo: > > but if @SomeoneSerge is willing hopefully we can collaborate on getting this merged?
> 
> I'm very much interested in getting this merged. I expect a _lot_ of fall out, e.g. I'd expect there to be projects which abuse some kind of generic `find_file` which we'd break, but I'd rather we tightened the bolts and handled those as special cases

i built all attributes that had cmake in the nativeBuildInputs and strict deps set and fixed the breakage
https://github.com/NixOS/nixpkgs/pull/324545
https://github.com/NixOS/nixpkgs/pull/324951

along with darwin up to, i forget how far i built. kdePackages builds 400 packages or something that all set strictDeps but many of them don't work with strictDeps so they need to be disabled. I did all of these builds more than a month ago so probably need to do another round.

paparodeo:
new commits switch back to patching cmake, and update the env name and comments.

pkgs/by-name/cm/cmake/000-nixpkgs-disable-bin-prefix.diff
```diff
@@ -0,0 +1,36 @@
+diff --git a/Source/cmSearchPath.cxx b/Source/cmSearchPath.cxx
```
AndersonTorres, line 1: What are the chances of upstreaming this patch (possibly with modifications)?

paparodeo: to upstream would want to remove env var and use a command line argument to enable/disable i think. then need an argument why users other than nixpkgs would want this. the patch as is i think works well for our purposes as it is minimal and non-intrusive. it might be worth while filing an issue upstream tho as perhaps it causes issues for other cross-compilation frameworks?

there are cmake variables  https://cmake.org/cmake/help/latest/variable/CMAKE_IGNORE_PATH.html#variable:CMAKE_IGNORE_PATH and https://cmake.org/cmake/help/latest/variable/CMAKE_IGNORE_PREFIX_PATH.html#variable:CMAKE_IGNORE_PREFIX_PATH which seem like they'd work but when set they also ignore things in PATH which doesn't work for us.

AndersonTorres: It doesn't work in an absolute sense, or in the sense of "using this CMAKE_IGNORE_PREFIX_PATH could theoretically work, however it would require too much extra work"?

Because, well, my first thought was "well, we set this var to throw all PATH out, then we rebuild PATH as we wish".

paparodeo: it does not work because it ignores the directories in PATH in addition to CMAKE_PREFIX_PATH. I think my first attempt was when we add a directory to CMAKE_PREFIX_PATH i added `dir/bin` to CMAKE_IGNORE_PATH. That has issues if the directory is also in the PATH. Then I thought i could just take the difference of CMAKE_PREFIX_PATH + bin - PATH and ignore that. But then that can change the lookup order. In the case of zig / castxml I think that the compiler wrapper needed to come first but there were still binaries that were required in the llvm directory that also contained the unwrapped compiler. so both directories were needed and preserving the order was important.

so your suggestion, which i think is to say ignore everything in CMAKE_PREFIX_PATH then build a PATH with aliases  seems like it would work. However, it would still have the issue @emilazy brought up about finding binaries outside of the bin directory, like share and lib, but i suppose those could be enumerated and just added to CMAKE_IGNORE_PATH as well.

though, i suppose we could use my previous, non cmake change, and just add `share` `lib` etc to the CMAKE_IGNORE_PATH too. I do prefer cmake patch because of it's simplicity and we don't need to enumerate all paths that cmake searches long hand and it seems exactly what we want. the downside is that we need to maintain a patch that changes cmake behavior.

pkgs/by-name/cm/cmake/setup-hook.sh
paparodeo: - added a PR to remove strictDeps from kdePackages https://github.com/NixOS/nixpkgs/pull/329822

paparodeo:
will squash commits once reviewed.

pkgs/by-name/cm/cmake/setup-hook.sh
SomeoneSerge: > > If we set it in a hook like I proposed in https://github.com/NixOS/nixpkgs/pull/318226#discussion_r1689604322 then it’d work properly for nix develop derivation debugging too.

> That actually sounds convincing.

Oh this was maybe a bit of a jump.
With the patch+env variable approach we do change the behaviour for cmake devShells, when used with "impure (system) dependencies" for general development as opposed to "debugging nix derivations". What's in the scope of this issue is to change how cmake  discovers other nixpkgs libraries, not how cmake interprets CMAKE_PREFIX_PATH when the user tries to use impure/fhs libraries.

I don't think this is a shortcoming of the patch+env approach, but rather of mkShell or of how we design setup hooks

pkgs/by-name/cm/cmake/000-nixpkgs-disable-bin-prefix.diff
```diff
@@ -0,0 +1,36 @@
...
+ }
+ 
++static bool IsBinPrefixDisabled(void)
++{
++  enum States { UNDEFINED, DISABLED, ENABLED };
++  static States state = UNDEFINED;
++  if (state == UNDEFINED) {
++    state = cmSystemTools::GetEnvVar("NIXPKGS_CMAKE_IGNORE_PREFIX_PATH_FOR_BIN_SEARCH") != std::nullopt ? DISABLED : ENABLED;
++  }
++  return state == DISABLED;
```
SomeoneSerge, line 16: I read `bool enabled = ...; return enabled`. Any particular motivation to write this the way it is?

pkgs/by-name/cm/cmake/setup-hook.sh
SomeoneSerge: Side note: we can't say "unset NIXPKGS_... to use fhs libraries" in the docs either, because this would break nixpkgs libraries back again

pkgs/by-name/cm/cmake/000-nixpkgs-disable-bin-prefix.diff
paparodeo: not sure i understand. there need to be 3 states. undefined, disabled and enabled. with a boolean there are only two states so it is not possible to know if the call to getenv took place or not.

the reason i didn't just do `getenv("NIXPKGS_...") != NULL` was to just avoid searching the env on every call and keep the change minimal impact. tho admittedly it probably doesn't really matter either way.

pkgs/by-name/cm/cmake/setup-hook.sh
paparodeo: I might be able to patch in a `CMAKE_PREFIX_NO_BIN_PATH` that works just like `CMAKE_PREFIX_PATH` except it doesn't search bins

https://github.com/Kitware/CMake/blob/262a89c92b4c736fd07268bf868779256afb0d9d/Source/cmFindBase.cxx#L282

and just do
```c++
if (this->CMakePathName != "PROGRAM") {
   paths.AddEnvPrefixPath("CMAKE_PREFIX_NO_BIN_PATH");
}
```

it think that would solve all problems. `CMAKE_PREFIX_PATH` behaviour does not change and nixpkgs will work with strictDeps. let me try it out and see if it actually works...

paparodeo: ok. got the patch to work to use `CMAKE_PREFIX_NO_BIN_PATH`. perhaps a better name is `CMAKE_PREFIX_PATH_NO_PROGRAM`. or `CMAKE_PREFIX_PATH_SANS_PROGRAM` or `NIXPKGS_PREFIX_PATH` or `NIXPKGS_CMAKE_PREFIX_PATH` or `CMAKE_PREFIX_PATH_NIXPKGS`. idk.

paparodeo:
updated cmake patch so to add `NIXPKGS_CMAKE_PREFIX_PATH` which is like `CMAKE_PREFIX_PATH` but not searched for programs. see https://github.com/NixOS/nixpkgs/pull/318226#discussion_r1692178772

commits need to be squashed

pkgs/by-name/cm/cmake/000-nixpkgs-disable-bin-prefix.diff
SomeoneSerge: I mean, `state` is just a local variable here, and `if (state == UNDEFINED)` always goes into the then-branch?

paparodeo: > I mean, `state` is just a local variable here, and `if (state == UNDEFINED)` always goes into the then-branch?

it is static so
```
static States state = UNDEFINED;
```
so `UNDEFINED` is only the value on the entrance to the first call to the function.

pkgs/by-name/cm/cmake/setup-hook.sh
SomeoneSerge: Super nice

pkgs/by-name/cm/cmake/000-nixpkgs-disable-bin-prefix.diff
SomeoneSerge: Aaah! I should wear glasses xD

pkgs/by-name/cm/cmake/setup-hook.sh
paparodeo: so it seems that when meson uses cmake to find dependencies it doesn't pass through the environment and calls cmake with ` -DCMAKE_PREFIX_PATH=/nix/store/6ibnp3wdspc36lgyrayajd4hnqipf9xy-meson-1.4.0;...` which is just passing in the CMAKE_PREFIX_PATH it got from the env, but converting the colons to semi-colons. so that's a problem.

paparodeo: so given the meson error and not passing thru the env i think that means we're back to this:

```
addCMakeParams() {
    addToSearchPath CMAKE_INCLUDE_PATH "$1/include"
    addToSearchPath CMAKE_INCLUDE_PATH "$1"

    addToSearchPath CMAKE_LIBRARY_PATH "$1/lib"
    addToSearchPath CMAKE_LIBRARY_PATH "$1"

    addToSearchPath CMAKE_PREFIX_PATH "$1/lib/cmake"
    addToSearchPath CMAKE_PREFIX_PATH "$1/share/cmake"
    addToSearchPath CMAKE_PREFIX_PATH "$1/lib"
    addToSearchPath CMAKE_PREFIX_PATH "$1/share"
}
```

and if there are executables in share or lib that conflict with PATH, tough.

SomeoneSerge: I'll try looking into whether we can cheaply fix meson

paparodeo: ok -- i restored this PR with the NIXPKGS_CMAKE_PREFIX_PATH and will leave it with changes un-squashed.

pkgs/by-name/cm/cmake/000-nixpkgs-disable-bin-prefix.diff
SomeoneSerge: Regarding upstreaming, I get a sense that maybe cmake should need separate `CMAKE_PREFIX_PATH_HOST` and `CMAKE_PREFIX_PATH_BUILD`, with the logic that `CMAKE_PREFIX_PATH` is appended to both

pkgs/by-name/cm/cmake/setup-hook.sh
emilazy: Sorry about the back and forth between the approaches here.

I’m surprised that Messon clears the environment like that. Surely that breaks other Nixpkgs things? I guess for things like `NIX_CFLAGS_COMPILE` it isn’t relevant because that mostly applies to after the CMake generation?

I agree that a separate variable seems nice and that patching Meson is probably preferable if it’s not too tricky. Apart from just programs, do we want the behaviour where `CMAKE_PREFIX_PATH` applies to things like libraries as well, rather than just CMake code? It seems there’s a `CMAKE_MODULE_PATH`, but it doesn’t apply to `*Config.cmake` files for whatever reason.

SomeoneSerge: Actually, wdyt about the names in https://github.com/NixOS/nixpkgs/pull/318226#discussion_r1692875910?

We could basically implement the logic that
- `CMAKE_PREFIX_PATH_BUILD` and `CMAKE_PREFIX_PATH` are added to `find_program`'s `paths`
- `CMAKE_PREFIX_PATH_HOST` and `CMAKE_PREFIX_PATH` are added to `find_library`'s and `find_path`'s `paths` (a big stretch, assuming that `find_path` is used to lookup host's stuff)
- setup-hooks add `pkgsBuildHost` to `CMAKE_PREFIX_PATH_BUILD` and `pkgsHost*` to `CMAKE_PREFIX_PATH_HOST`; nothing is added to `CMAKE_PREFIX_PATH`; users may set up a `shellHook = "export CMAKE_PREFIX_PATH=$CMAKE_PREFIX_PATH_BUILD:$CMAKE_PREFIX_PATH_HOST` if they wish

SomeoneSerge: Didn't find any issues about CMAKE_PREFIX_PATH and cross upstream, we _must_ be reinventing the wheel.

Looking at https://cmake.org/cmake/help/book/mastering-cmake/chapter/Cross%20Compiling%20With%20CMake.html#finding-external-libraries-programs-and-other-files, they have thought about controlling `find_{path,library,program}`'s behaviour in the cross setting.

Looking at [CMAKE_FIND_ROOT_PATH](https://cmake.org/cmake/help/latest/variable/CMAKE_FIND_ROOT_PATH.html#variable:CMAKE_FIND_ROOT_PATH), the variable is also used for `find_package` lookups.

This maybe means that we can
- populate `CMAKE_FIND_ROOT_PATH` with the prefixes from `pkgsHost{Host,Target}`
- set [`CMAKE_FIND_ROOT_PATH_MODE_PROGRAM`](https://cmake.org/cmake/help/latest/variable/CMAKE_FIND_ROOT_PATH_MODE_PROGRAM.html#variable:CMAKE_FIND_ROOT_PATH_MODE_PROGRAM) to `NEVER`
- populate `CMAKE_PREFIX_PATH` with `${!output}/{lib,share}/cmake` and `${!output}/bin` for `pkgsBuild{Build,Host}` so that `find_package` and `find_program` work for native inputs

paparodeo: I'm not sure how [CMAKE_FIND_ROOT_PATH](https://cmake.org/cmake/help/latest/variable/CMAKE_FIND_ROOT_PATH.html#variable:CMAKE_FIND_ROOT_PATH) is supposed to work. from the [docs](https://cmake.org/cmake/help/book/mastering-cmake/chapter/Cross%20Compiling%20With%20CMake.html#finding-external-libraries-programs-and-other-files) it indicates that this should be a list of package roots, so basically the same thing we were using for `CMAKE_PREFIX_PATH`.

so modify setup-hook.sh to:
```
addCMakeParams() {
    # NIXPKGS_CMAKE_PREFIX_PATH is like CMAKE_PREFIX_PATH except cmake
    # will not search it for programs
    addToSearchPath CMAKE_FIND_ROOT_PATH $1
}
[...]
    cmakeFlags="-DCMAKE_FIND_ROOT_PATH=${CMAKE_FIND_ROOT_PATH/:/;} $cmakeFlags"
    cmakeFlags="-DCMAKE_FIND_ROOT_PATH_MODE_PROGRAM=NEVER $cmakeFlags"
    cmakeFlags="-DCMAKE_PREFIX_PATH=/ $cmakeFlags"
```

then do a test build (castxml with clang will break with old cmake)
```
nix-build -E 'with import ./.{}; castxml.override { stdenv = clangStdenv; cmake = cmake3; }'
```

and find_package fails to find `LLVMConfig.cmake` so there is something i'm missing.

(cmake3 is just cmake with the modifications to setup-hook.sh)

paparodeo: ok. this works:

```
addCMakeParams() {
    addToSearchPath CMAKE_FIND_ROOT_PATH $1
}
[...]
    cmakeFlags="-DCMAKE_FIND_ROOT_PATH=${CMAKE_FIND_ROOT_PATH//:/;} $cmakeFlags"
    cmakeFlags="-DCMAKE_FIND_ROOT_PATH_MODE_PROGRAM=NEVER $cmakeFlags"
    cmakeFlags="-DCMAKE_PREFIX_PATH=/ $cmakeFlags"
```

```
nix-build -E 'with import ./.{}; castxml.override { stdenv = clangStdenv; cmake = cmake3; }'
```

the downside is needing to set `CMAKE_PREFIX_PATH` to `/` which is not great when trying to do out-of-sandbox builds on non-nixos systems. meson is still broken when using `cmake` to find packages.

paparodeo:
due to other commitments i no longer have time to work on this will leave the PRs open but someone else will need to move them forward.

eclairevoyant:
Thank you for the work thus far!

pkgs/by-name/cm/cmake/setup-hook.sh
paparodeo: tl;dr -- ranking of methods
1. https://github.com/NixOS/nixpkgs/pull/318226#discussion_r1692828567 modify `addCMakeParams` and use various CMAKE_XXX permutations. up sides -- strictDeps works better, meson still works. downside, avoids binaries in `bin` but can still pick up binaries that may exist in `share`, `lib` or `include`. However, this is an improvement over where we are today without adding further regressions.
2. patch cmake and use `NIXPKGS_CMAKE_PREFIX_PATH`. this is like `CMAKE_PREFIX_PATH` but doesn't search for binaries. fixes strictDeps, binaries are searched via PATH, however, meson breaks when using cmake to locate packages thus requiring a patch.
3. use CMAKE_FIND_ROOT_PATH. fixes strictDeps. out-of-sandbox builds pick up binaries from /bin, meson still broken. would need to patch cmake so doesn't seem better than option 2.

so i think should go with option 1. and open an issue and perhaps try to transition to a full solution. I just don't have the bandwidth to be searching / testing other solutions. even option 1 needs to run more tests as i did most of my tests using a patched cmake. there are 3600 cmake builds that are overridable without strict deps set, so need to do like 600 builds at least. could re-run all the strict dep builds as well and just skip the 30 or so already patched. anyway, this is just more time than i have available.

yeah, `CMAKE_FIND_ROOT_PATH` almost works except for falling back to search in `/{bin,include,lib,etc}` for programs before PATH which is the same issue we have with just using `CMAKE_PREFIX_PATH`. I verified doing a nix-shell build on ubuntu picks up the system compiler rather than the nix one. On darwin `/bin` is not linked to `/usr/bin` so there is less there, at least in my monterey VM but that still seems like a problem esp because darwin sandbox isn't enabled by default.

frustrating -- that seemed so promising.

> We could basically implement the logic that
> 
>     * `CMAKE_PREFIX_PATH_BUILD` and `CMAKE_PREFIX_PATH` are added to `find_program`'s `paths`

so this basically would just amount to `export CMAKE_PREFIX_PATH_BUILD=$PATH` (or i guess the paths stripped of the leaf directory). Given that it would require patching cmake I don't see the advantage over the current patching strategy -- ignoring prefix-path for `find_program`.  

>     * `CMAKE_PREFIX_PATH_HOST` and `CMAKE_PREFIX_PATH` are added to `find_library`'s and `find_path`'s `paths` (a big stretch, assuming that `find_path` is used to lookup host's stuff)

i think this is how `NIXPKGS_CMAKE_PREFIX_PATH` works in the current cmake patch. renaming seems fine I just tried to avoid `CMAKE` prefix to avoid a future collision.

paparodeo: https://github.com/NixOS/nixpkgs/pull/318226#discussion_r1692828567 breaks building `elastix`. one issue with cmake files in `hdf5-cpp-1.14.4.3-dev`. finds files but when a `lib` or `lib/cmake` is appended see a failure, otherwise no failure. when that passes can't find opengl.  neither failures make sense to me. effing ef.

configure runs fine with patched cmake.

paparodeo: ok. back to option 2, the patched cmake with `NIXPKGS_CMAKE_PREFIX_PATH`. meson does pass thru env vars, despite what i said earlier. the following meson patch makes `wl-touch` work.

<details><summary>meson patch for nixpkgs_cmake_prefix_path</summary>

```
diff --git a/mesonbuild/dependencies/data/CMakePathInfo.txt b/mesonbuild/dependencies/data/CMakePathInfo.txt
index 662ec58..ea79746 100644
--- a/mesonbuild/dependencies/data/CMakePathInfo.txt
+++ b/mesonbuild/dependencies/data/CMakePathInfo.txt
@@ -2,9 +2,11 @@ cmake_minimum_required(VERSION ${CMAKE_MAJOR_VERSION}.${CMAKE_MINOR_VERSION}.${C
 
 set(TMP_PATHS_LIST)
 list(APPEND TMP_PATHS_LIST ${CMAKE_PREFIX_PATH})
+list(APPEND TMP_PATHS_LIST ${NIXPKGS_CMAKE_PREFIX_PATH})
 list(APPEND TMP_PATHS_LIST ${CMAKE_FRAMEWORK_PATH})
 list(APPEND TMP_PATHS_LIST ${CMAKE_APPBUNDLE_PATH})
 list(APPEND TMP_PATHS_LIST $ENV{CMAKE_PREFIX_PATH})
+list(APPEND TMP_PATHS_LIST $ENV{NIXPKGS_CMAKE_PREFIX_PATH})
 list(APPEND TMP_PATHS_LIST $ENV{CMAKE_FRAMEWORK_PATH})
 list(APPEND TMP_PATHS_LIST $ENV{CMAKE_APPBUNDLE_PATH})
 list(APPEND TMP_PATHS_LIST ${CMAKE_SYSTEM_PREFIX_PATH})
```
</details>

tho, in the case of `CMAKE_PREFIX_PATH` meson takes from the env, removes duplicates and then passes it to cmake on the command line. so unclear to me what the negative consequences of this patch may be.

paparodeo:
> due to other commitments i no longer have time to work on this. will leave the PRs open but someone else will need to move them forward.

haven't quit quite yet. pushed the meson patch. builds successfully on 54 packages that use meson + cmake in nativeBuildInputs.

at 500 builds from sample of 3500 packages that use cmake but not strictDeps. strictDeps (400 packages) were already tested.

emilazy:
I’m happy with this approach. The cross stuff upstream in CMake seems potentially promising but likely not fully aligned to our needs. I support this as a simple, light solution that keeps full compatibility with standard CMake semantics. We can always revisit later if we figure out a nicer way or come up with something that can be upstreamed.

I can hit the merge button on this when you think it’s ready. @SomeoneSerge Are you planning to review #324545?

SomeoneSerge:
I agree with the NIXPKGS_CMAKE_PREFIX_PATH approach as an interim solution.

I'm still waiting for the results on `pkgsCross.aarch64-multiplatform.glfw` (I keep getting stuck on gnutls tests; maybe I should've just used the modified cmake with glfw from master) but it's ok to merge before.

> @SomeoneSerge Are you planning to review https://github.com/NixOS/nixpkgs/pull/324545?

Yes, but I might not be able to do that before tomorrow

paparodeo:
commits squashed.
- this should get merged after https://github.com/NixOS/nixpkgs/pull/324545 is merged and then merged into staging.

thanks!

emilazy (approved):
Let’s go!

Thank you for working on this. I’m very happy with where it’s ended up. Hopefully we can turn `strictDeps` on by default in not too long.

paparodeo:
> I'm still waiting for the results on `pkgsCross.aarch64-multiplatform.glfw` (I keep getting stuck on gnutls tests; maybe I should've just used the modified cmake with glfw from master) but it's ok to merge before.

```
nix-build -E 'with import ./.{}; pkgsCross.aarch64-multiplatform.glfw.override { cmake = cmake2; extra-cmake-modules = extra-cmake-modules.override { cmake = cmake2; }; }'
```
builds on master where cmake2 is the patched cmake from this PR. `extra-cmake-modules` adds cmake to its `propogatedBuildInputs` so need to thread cmake2 in there as well or else the `addCMakeParams` gets clobbered and the wrong env var gets set. with both getting cmake2 build succeeds finding the correct `wayland-scanner` binary.

SomeoneSerge:
>  so need to thread cmake2 in there as well or else the addCMakeParams gets clobbered and the wrong env var gets set.

M'hm this sounds like we want to put a guard around https://github.com/NixOS/nixpkgs/blob/55575065f591c8c000412c62bd612a10c9d571cc/pkgs/by-name/cm/cmake/setup-hook.sh#L146

Something like

```bash
if (( "$hostOffset" < 0 )); then 
    addEnvHooks "$targetOffset" addCMakeParams
fi
```

...would prevent the hook from running from buildInputs' cmake

SomeoneSerge:
One more thing, since we're bothering these dragons. How is `addEnvHooks "$targetOffset" $callback` any better than looping over `"${pkgsHostHost[@]}" "${pkgsHostTarget[@]}"`?

paparodeo:
i feel like `extra-cmake-modules` shouldn't be propogating cmake, and if it had to, shouldn't it be propogatedNativeBuildInputs? it puts `pkg-config` there too so seems like it is wrong.
https://github.com/NixOS/nixpkgs/blob/55575065f591c8c000412c62bd612a10c9d571cc/pkgs/development/libraries/kde-frameworks/extra-cmake-modules/default.nix#L16

i modified the nix script in the PR description to find packages that use `extra-cmake-modules` but don't have a cmake. there are 30 or so.

as for adding a guard to prevent the setup-hook.sh from running if in the build inputs. seems reasonable, would need to fix the 30 or so packages that aren't including cmake and getting it from `extra-cmake-modules` and fix any packages that are adding cmake to the buildInputs rather than nativeBuildInputs.

> One more thing, since we're bothering these dragons. How is addEnvHooks "$targetOffset" $callback any better than looping over "${pkgsHostHost[@]}" "${pkgsHostTarget[@]}"?

i am not sure what the difference is. TBH those functions are just like -- magic happens and the buildInputs get passed in and whatever appropriate depsXXYY

SomeoneSerge:
> i am not sure what the difference is. TBH those functions are just like -- magic happens and the buildInputs get passed in and whatever appropriate depsXXYY

Well I'm essentially reading ~~tea leaves~~ bash scripts here, but I think `pkgsXY` are kind of static and they are `X` and `Y` relative to the derivation being built, whereas `targetOffset` and `hostOffset` correspond to the current recursion level of `findInputs`... This is too complicated for me to reason about, but I was suspecting they're equivalent because the offset always stay within `-1 <= offset <= 1` and we don't really want the hook to run for cmake not at `host=-1`

> as for adding a guard to prevent the setup-hook.sh from running if in the build inputs. seems reasonable, would need to fix the 30 or so packages that aren't including cmake and getting it from extra-cmake-modules and fix any packages that are adding cmake to the buildInputs rather than nativeBuildInputs

Moving cmake to `propagatedNativeBuildInputs` (assuming `extra-cmake-modules` always go to `buildInputs`) would hide this issue?

emilazy:
i’ve also seen it [claimed/written in derivations](https://discourse.nixos.org/t/frustrations-about-splicing/49607/21) that when you have `extra-cmake-modules` you want `cmake` in `pkgsBuildBuild`. I can’t pretend to understand what’s going on there.

paparodeo:
yeah, idk https://github.com/NixOS/nixpkgs/blob/55575065f591c8c000412c62bd612a10c9d571cc/pkgs/development/libraries/kde-frameworks/extra-cmake-modules/default.nix#L16 looks wrong to me. cmake nor pkg-config should be in propogatedBuildInputs. and i don't feel like they should be propogated at all. do include files propogate a c compiler?

as for the discourse discussion. no idea. that makes no sense to me. cmake isn't building extra-cmake-modules in any module except for when creating the `extra-cmake-modules` so why would it need to be `depsBuildBuild-ing` anything.

> Moving cmake to propagatedNativeBuildInputs (assuming extra-cmake-modules always go to buildInputs) would hide this issue?
`extra-cmake-modules ` is in nativeBuildInputs from what i've seen, but you're correct, if `extra-cmake-modules` is still propagating cmake no downstream modules will need to get modified -- i guess i was thinking cmake should not get propogated at all.


will get back to the offset stuff...

SomeoneSerge:
> looks wrong to me. cmake nor pkg-config should be in propogatedBuildInputs. and i don't feel like they should be propogated at all. do include files propogate a c compiler?

Very implicit, but it could be to propagate the cmake and pkg-config hooks... (we do something like that with cuda which I'm very sorry about)

> extra-cmake-modules is in nativeBuildInputs from what i've seen,

We could also put ecm in buildInputs and cmake/pkg-config in propagatedNativeBuildInputs or maybe depsBuildBuildPropagated to the exact same effect (I always forget if there are any constants added to the offsets)

But I agree that propagating cmake/pkg-config from ECM is weird.

paparodeo:
so @SomeoneSerge, you might have suggested or implied and i missed it. so seems like we can use cmake's cross framework to avoid patching cmake. can just set `CMAKE_FIND_ROOT_PATH=${NIX_STORE}` and then `CMAKE_PREFIX_PATH` would contain the pkgs paths sans `/nix/store`. this should solve the out of sandbox problem as when cmake tries to search the the paths in the bare `CMAKE_PREFIX_PATH` eg: `/kp2j7yn0wzwq5piy494r54dafrh83s6s-gcc-wrapper-13.3.0` they will be useless without `/nix/store` prepended to them.

still would need to find a solution to the meson problem -- would need to patch in `CMAKE_FIND_ROOT_PATH=${NIX_STORE}` somewhere and maybe some other things.

but cmake would just require setup-hook.sh modifications like
```bash
addCMakeParams() {
    # just a bogus var name to collect paths
    addToSearchPath NIXPKGS_CMAKE_PREFIX_PATH "$1"
}

cmakeConfigurePhase() {
[...]
    # Remove the $NIX_STORE component from the paths so a valid path requires
    # appending CMAKE_FIND_ROOT_PATH to it. This allows us to use the
    # CMAKE_FIND_ROOT_PATH_MODE_XXX flags to prohibit cmake from looking up
    # programs in the conconated paths.
    export CMAKE_PREFIX_PATH="${NIXPKGS_CMAKE_PREFIX_PATH//${NIX_STORE}//}"
    cmakeFlags="-DCMAKE_FIND_ROOT_PATH=${NIX_STORE} $cmakeFlags"
    cmakeFlags="-DCMAKE_FIND_ROOT_PATH_MODE_PROGRAM=NEVER $cmakeFlags"
    cmakeFlags="-DCMAKE_FIND_ROOT_PATH_MODE_LIBRARY=ONLY $cmakeFlags"
    cmakeFlags="-DCMAKE_FIND_ROOT_PATH_MODE_INCLUDE=ONLY $cmakeFlags"
[...]
}
```

i suppose someone else can tackle that if they hate the cmake patch.

paparodeo:
> > looks wrong to me. cmake nor pkg-config should be in propogatedBuildInputs. and i don't feel like they should be propogated at all. do include files propogate a c compiler?
> 
> Very implicit, but it could be to propagate the cmake and pkg-config hooks... (we do something like that with cuda which I'm very sorry about). What's wrong is that there's no comment above propagatedBuildInputs. We should inspect the git-blame to find out why it's there
> 
> > extra-cmake-modules is in nativeBuildInputs from what i've seen,
> 
> We could also put ecm in buildInputs and cmake/pkg-config in propagatedNativeBuildInputs or maybe depsBuildBuildPropagated to the exact same effect (I always forget if there are any constants added to the offsets)
> 
> But I agree that propagating cmake/pkg-config from ECM is weird.

yeah, after writing a little [test code](https://gist.github.com/paparodeo/f40666a675cef4d7cae60c520b8a21ef) to see how propagated inputs of one package get mapped when adding it to the input of another I can't say that i am any less confused but I can at least put packages in the correct buckets to achieve the desired output.

so i guess that means that as long as ECM is in `nativeBuildInputs` if it is propagating cmake, that has to be in `propagatedBuildInputs`. still doesn't seem like it needs to propagate it -- looking at the hooks it seems like it is just a convenience to avoid needing to include both cmake and ECM. likely the same with pkg-config. which is basically what you already said, i believe.

the git blame  https://github.com/NixOS/nixpkgs/commit/252b36a28db1dd9036cfcbfbaa64b63e6ea5655a and https://github.com/NixOS/nixpkgs/pull/23239/commits/127bf18a350afa562cad3f8cb3b2f626edd23de0#diff-d6ab427e9da03c0f8d9663c67fe522712a04e32ef13859e23aada8693e77d674 are not very enlightening.

> One more thing, since we're bothering these dragons. How is `addEnvHooks "$targetOffset" $callback` any better than looping over `"${pkgsHostHost[@]}" "${pkgsHostTarget[@]}"`?

from the cmake perspective -- using `addEnvHooks "$targetOffset" $callback` seems worse as it depends on what input bucket cmake is dumped in -- it is only equivalent to looping over `"${pkgsHostHost[@]}" "${pkgsHostTarget[@]}"` when cmake is part of `nativeBuildInputs`

there are only 4 packages that use cmake in `buildInputs` rather than `nativeBuildInputs` so perhaps it doesn't matter much

SomeoneSerge:
> - it is only equivalent to looping over "${pkgsHostHost[@]}" "${pkgsHostTarget[@]}" when cmake is part of nativeBuildInputs

Yes precisely my point and I've no idea if that's intentional. The only use-cases I know use cmake setup hook exclusively in nativeBuildInputs so if I were writing this I'd go for the less generic logic

SomeoneSerge:
> still would need to find a solution to the meson problem -- would need to patch in CMAKE_FIND_ROOT_PATH=${NIX_STORE} somewhere and maybe some other things.

Damn I think I've completely misunderstood cmake's docs on CMAKE_FIND_ROOT_PATH, because I thought at first it was precisely like CMAKE_PREFIX_PATH, just more controllable thanks to the `DCMAKE_FIND_ROOT_PATH_MODE_*` bool options 😱

paparodeo:
> Damn I think I've completely misunderstood cmake's docs on CMAKE_FIND_ROOT_PATH, because I thought at first it was precisely like CMAKE_PREFIX_PATH, just more controllable thanks to the `DCMAKE_FIND_ROOT_PATH_MODE_*` bool options 😱

it took me a few tries to get it to work. https://github.com/NixOS/nixpkgs/pull/318226#issuecomment-2254491401 is the closest i got to what we want, except it breaks down if `NIX_STORE` ends up being `/` in which case we're back to where we started -- find_program pulling things from `CMAKE_PREFIX_PATH` rather than path.

anyway -- the patches to cmake and meson in this PR are small and hopefully easy to understand. Perhaps it is not obvious without grepping for `CMAKE_PREFIX_PATH` in the cmake code, but this PR only adds a new env var and not a cmake variable, so one can't pass in `NIXPKGS_CMAKE_PREFIX_PATH` on the cmd line or in a cmake file, needs to be the env. it is easy enough to add -- basically the same changes calling a slightly different function -- but i decided to leave it out since we weren't using it.

> > * it is only equivalent to looping over "${pkgsHostHost[@]}" "${pkgsHostTarget[@]}" when cmake is part of nativeBuildInputs
> 
> Yes precisely my point and I've no idea if that's intentional. The only use-cases I know use cmake setup hook exclusively in nativeBuildInputs so if I were writing this I'd go for the less generic logic

i agree -- the thing i like about the current hooks is that it is possible to drop into the derivation via nix-shell and `echo $NIXPKGS_CMAKE_PREFIX_PATH` without needed to run any build phases which is also required for interacting with meson. tho perhaps it is easy enough to run the loop at load time. looking at a few of the packages that put `cmake` in `buildInputs` it works because they aren't using `strictDeps` and they also are using meson, so keeping cmake out of `nativeBuildInputs` eliminates the risk of cmake taking over the configurePhase

eclairevoyant:
> keeping cmake out of `nativeBuildInputs` eliminates the risk of cmake taking over the configurePhase

yet we have `dontUseCmakeConfigure`, IMO propagation is overused in Qt and I wouldn't be surprised if it was related to ous problems with cross-compiling Qt apps

AndersonTorres:
> IMO propagation is overused in Qt

A long time ago it was recommended to me "avoid `propagated*`" in a Python package.

Is there anything documenting these things in a more practical way?

eclairevoyant:
Well it was just my opinion 😅 I found it very difficult to reason about dependencies when I was trying to fix `strictDeps` in Qt a while back. I don't know of it being documented.

SomeoneSerge:
> A long time ago it was recommended to me "avoid propagated*" in a Python package.

Not exactly on-topic, but pythonXPackages now use pyproject.toml-style `build-system` and `dependencies` attributes, which are temporarily mapped into `propagated*` but will be hopefully later replaced with the retained/`python3.withPackages`-time composition on the evaluation side, which should reduce # of rebuilds and allow better conflict handling

AndersonTorres (approved):

kjeremy:
We're pretty sure that this breaks nix-ros-overlay: https://github.com/lopsided98/nix-ros-overlay/issues/491

SomeoneSerge:
@sielicki I think these two messages reflect the knowledge we've had at the moment:

- https://github.com/NixOS/nixpkgs/pull/318226#discussion_r1693744252
- https://github.com/NixOS/nixpkgs/pull/318226#issuecomment-2254491401

We sought how other projects (cmake, vcpkg, connan) approach this problem but cut the search short and went for paparodeo's patch as interim solution to cross issues. I don't think we were aware of `CMAKE_SYSTEM_IGNORE_PREFIX_PATH` which seems relevant. A PR removing the patch, while keeping the build-host separation behave correctly, would be welcomed by all parties?

As you bring this up, I'll start listing some side considerations again:

<details>

- our approach to configuring CMake (wlog) is poorly compatible with dev shells,
  - e.g. because we don't separate input configuration from `CMAKE_INSTALL_*_PATH`s,
  - and because it's cumbersome to teach IDEs about our variables (I'm fine with `compile_commands.json`, but at some point I had to deal with people insisting on using VScode, and on making it "windows-like", whatever that means),
  - as we're moving towards `__structuredAttrs`, we're getting less compatible with `direnv`.
  - I think the way to approach this is by implementing a Vcpkg-like UX: accumulate all dependency configuration and flags (other than `*INSTALL*`) in a toolchain file instead of in environment variables.
- mutable `source` directory leads to `/source/build/` references in `-g` debug flags, breaking source listings
-

sielicki:
> I think the way to approach this [is to] accumulate all dependency configuration and flags (other than *INSTALL*) in a toolchain file instead of in environment variables.

Totally agree on this overall idea. Recent cmake releases have a lot of new features to make this easier.

1. [cmake_pkg_config](https://cmake.org/cmake/help/latest/command/cmake_pkg_config.html) landed in 3.31, which allows cmake to parse `.pc` files in a way that's totally native to cmake, ie: without relying on any 3rd-party `pkg-config` binary. You can also invoke this on the command line with [cmake --find-package](https://cmake.org/cmake/help/latest/manual/cmake.1.html#run-the-find-package-tool). We need to review the environment variables being used there to make sure the defaults are cross-safe.
2. [CMake dependency providers](https://cmake.org/cmake/help/latest/command/cmake_language.html#dependency-providers) landed in 3.24. This lets you intercept `FetchContent` and `find_package` calls at configure time, fulfilling them however you'd like. We should be able to create a fixed output representing the cache (similar to `cargoHash`/`pnpmHash`/etc.) by writing a nix dependency provider. I came across [cmake-tipi-provider](https://github.com/tipi-build/cmake-tipi-provider/blob/main/tipi_provider.cmake) which gives a good idea of what it would need to look like.
3. Along the same lines, there's now the [cmake-file-api](https://cmake.org/cmake/help/latest/manual/cmake-file-api.7.html#manual:cmake-file-api(7)) which lets you discover targets in a portable way, again at configure time. We just need a hook that creates a `.cmake/api/v1/query/client-nix_cmake_hook/query.json` in cmake build directories, and then write something which can walk the result and map it to nixisms. This could let us do a bunch of interesting things: detecting all the targets, detecting test targets, detecting configuration flags, discerning between generated files and project source files for better source filtering, cheaply constructing compile_commands.json for nix derivations without requiring a build to take place, etc.
4. the whole reason that I was looking at this PR in the first place was that I was thinking about [CPS](https://cps-org.github.io/cps/overview.html) and how it could be implemented in nixpkgs. CMake is now [capable of finding packages through cps](https://www.kitware.com/navigating-cmake-dependencies-with-cps/). I'm not sure how it fits into nixpkgs today, but it's worth thinking about.

> We sought how other projects (cmake, vcpkg, connan) approach this problem but cut the search short and went for paparodeo's patch as interim solution to cross issues. I don't think we were aware of CMAKE_SYSTEM_IGNORE_PREFIX_PATH which seems relevant. A PR removing the patch, while keeping the build-host separation behave correctly, would be welcomed by all parties?

I can try to put that together -- just having trouble understanding exactly how these variables are used, and what stands to break.

sielicki:
> yes, because they don't work.

Is there an upstream issue?

It's mentioned elsewhere in the cmake docs, but not on these specific doc pages, that these variables are semicolon separated and not colon-separated like normal search paths. So you need to use `addToSearchPathWithCustomDelimiter ";" CMAKE_SYSTEM_* "$@"`, etc.

Here's a patch I have locally, but I need to test it more to see what breaks:

```diff
9f3110649578b2f7c8bff356570e97138c3ffc10
Author:     Nicholas Sielicki <nix@opensource.nslick.com>
AuthorDate: Sun Apr 20 23:16:39 2025 -0700
Commit:     Nicholas Sielicki <nix@opensource.nslick.com>
CommitDate: Sun Apr 20 23:18:35 2025 -0700

remove a cmake patch

5 files changed, 22 insertions(+), 57 deletions(-)
.../cm/cmake/000-nixpkgs-cmake-prefix-path.diff    | 28 ------------------
pkgs/by-name/cm/cmake/package.nix                  |  3 --
pkgs/by-name/cm/cmake/setup-hook.sh                | 33 ++++++++++++++--------
.../me/meson/000-nixpkgs-cmake-prefix-path.patch   | 12 --------
pkgs/by-name/me/meson/package.nix                  |  3 --

deleted    pkgs/by-name/cm/cmake/000-nixpkgs-cmake-prefix-path.diff
@@ -1,28 +0,0 @@
-diff --git a/Source/cmFindBase.cxx b/Source/cmFindBase.cxx
-index 8840cdcb..c34b7ee9 100644
---- a/Source/cmFindBase.cxx
-+++ b/Source/cmFindBase.cxx
-@@ -280,6 +280,11 @@ void cmFindBase::FillCMakeEnvironmentPath()
-   // Add CMAKE_*_PATH environment variables
-   std::string var = cmStrCat("CMAKE_", this->CMakePathName, "_PATH");
-   paths.AddEnvPrefixPath("CMAKE_PREFIX_PATH");
-+  if (this->CMakePathName != "PROGRAM") {
-+    // Like CMAKE_PREFIX_PATH except when searching for programs. Programs need
-+    // to be located via PATH
-+    paths.AddEnvPrefixPath("NIXPKGS_CMAKE_PREFIX_PATH");
-+  }
-   paths.AddEnvPath(var);
- 
-   if (this->CMakePathName == "PROGRAM") {
-diff --git a/Source/cmFindPackageCommand.cxx b/Source/cmFindPackageCommand.cxx
-index 9b51b1ad..6acc676c 100644
---- a/Source/cmFindPackageCommand.cxx
-+++ b/Source/cmFindPackageCommand.cxx
-@@ -2039,6 +2039,7 @@ void cmFindPackageCommand::FillPrefixesCMakeEnvironment()
- 
-   // And now the general CMake environment variables
-   paths.AddEnvPath("CMAKE_PREFIX_PATH");
-+  paths.AddEnvPath("NIXPKGS_CMAKE_PREFIX_PATH");
-   if (this->DebugMode) {
-     debugBuffer = cmStrCat(debugBuffer,
-                            "CMAKE_PREFIX_PATH env variable "
modified   pkgs/by-name/cm/cmake/package.nix
@@ -59,9 +59,6 @@ stdenv.mkDerivation (finalAttrs: {
 
   patches =
     [
-      # Add NIXPKGS_CMAKE_PREFIX_PATH to cmake which is like CMAKE_PREFIX_PATH
-      # except it is not searched for programs
-      ./000-nixpkgs-cmake-prefix-path.diff
       # Don't search in non-Nix locations such as /usr, but do search in our libc.
       ./001-search-path.diff
     ]
modified   pkgs/by-name/cm/cmake/setup-hook.sh
@@ -1,8 +1,11 @@
-addCMakeParams() {
-    # NIXPKGS_CMAKE_PREFIX_PATH is like CMAKE_PREFIX_PATH except cmake
-    # will not search it for programs
-    addToSearchPath NIXPKGS_CMAKE_PREFIX_PATH $1
-}
+addCmakeSystemAppbundlePath() { addToSearchPathWithCustomDelimiter ";" CMAKE_SYSTEM_APPBUNDLE_PATH "$@" ; }
+addCmakeSystemFrameworkPath() { addToSearchPathWithCustomDelimiter ";" CMAKE_SYSTEM_FRAMEWORK_PATH "$@" ; }
+addCmakeSystemIgnorePath() { addToSearchPathWithCustomDelimiter ";" CMAKE_SYSTEM_IGNORE_PATH "$@" ; }
+addCmakeSystemIgnorePrefixPath() { addToSearchPathWithCustomDelimiter ";" CMAKE_SYSTEM_IGNORE_PREFIX_PATH "$@" ; }
+addCmakeSystemIncludePath() { addToSearchPathWithCustomDelimiter ";" CMAKE_SYSTEM_INCLUDE_PATH "$@" ; }
+addCmakeSystemLibraryPath() { addToSearchPathWithCustomDelimiter ";" CMAKE_SYSTEM_LIBRARY_PATH "$@" ; }
+addCmakeSystemPrefixPath() { addToSearchPathWithCustomDelimiter ";" CMAKE_SYSTEM_PREFIX_PATH "$@" ; }
+addCmakeSystemProgramPath() { addToSearchPathWithCustomDelimiter ";" CMAKE_SYSTEM_PROGRAM_PATH "$@" ; }
 
 fixCmakeFiles() {
     # Replace occurences of /usr and /opt by /var/empty.
@@ -147,7 +150,15 @@ if [ -z "${dontUseCmakeConfigure-}" -a -z "${configurePhase-}" ]; then
     configurePhase=cmakeConfigurePhase
 fi
 
-addEnvHooks "$targetOffset" addCMakeParams
+addEnvHooks "$targetOffset" \
+    addCmakeSystemAppbundlePath \
+    addCmakeSystemFrameworkPath \
+    addCmakeSystemIgnorePath \
+    addCmakeSystemIgnorePrefixPath \
+    addCmakeSystemIncludePath \
+    addCmakeSystemLibraryPath \
+    addCmakeSystemPrefixPath \
+    addCmakeSystemProgramPath
 
 makeCmakeFindLibs() {
     isystem_seen=
@@ -155,22 +166,22 @@ makeCmakeFindLibs() {
     for flag in ${NIX_CFLAGS_COMPILE-} ${NIX_LDFLAGS-}; do
         if test -n "$isystem_seen" && test -d "$flag"; then
             isystem_seen=
-            addToSearchPath CMAKE_INCLUDE_PATH "${flag}"
+            addToSearchPathWithCustomDelimiter ";" CMAKE_SYSTEM_INCLUDE_PATH "${flag}"
         elif test -n "$iframework_seen" && test -d "$flag"; then
             iframework_seen=
-            addToSearchPath CMAKE_FRAMEWORK_PATH "${flag}"
+            addToSearchPathWithCustomDelimiter ";" CMAKE_SYSTEM_FRAMEWORK_PATH "${flag}"
         else
             isystem_seen=
             iframework_seen=
             case $flag in
             -I*)
-                addToSearchPath CMAKE_INCLUDE_PATH "${flag:2}"
+                addToSearchPathWithCustomDelimiter ";" CMAKE_SYSTEM_INCLUDE_PATH "${flag:2}"
                 ;;
             -L*)
-                addToSearchPath CMAKE_LIBRARY_PATH "${flag:2}"
+                addToSearchPathWithCustomDelimiter ";" CMAKE_SYSTEM_LIBRARY_PATH "${flag:2}"
                 ;;
             -F*)
-                addToSearchPath CMAKE_FRAMEWORK_PATH "${flag:2}"
+                addToSearchPathWithCustomDelimiter ";" CMAKE_SYSTEM_FRAMEWORK_PATH "${flag:2}"
                 ;;
             -isystem)
                 isystem_seen=1
deleted    pkgs/by-name/me/meson/000-nixpkgs-cmake-prefix-path.patch
@@ -1,12 +0,0 @@
-diff --git a/mesonbuild/dependencies/data/CMakePathInfo.txt b/mesonbuild/dependencies/data/CMakePathInfo.txt
-index 662ec58..4d5f4e4 100644
---- a/mesonbuild/dependencies/data/CMakePathInfo.txt
-+++ b/mesonbuild/dependencies/data/CMakePathInfo.txt
-@@ -5,6 +5,7 @@ list(APPEND TMP_PATHS_LIST ${CMAKE_PREFIX_PATH})
- list(APPEND TMP_PATHS_LIST ${CMAKE_FRAMEWORK_PATH})
- list(APPEND TMP_PATHS_LIST ${CMAKE_APPBUNDLE_PATH})
- list(APPEND TMP_PATHS_LIST $ENV{CMAKE_PREFIX_PATH})
-+list(APPEND TMP_PATHS_LIST $ENV{NIXPKGS_CMAKE_PREFIX_PATH})
- list(APPEND TMP_PATHS_LIST $ENV{CMAKE_FRAMEWORK_PATH})
- list(APPEND TMP_PATHS_LIST $ENV{CMAKE_APPBUNDLE_PATH})
- list(APPEND TMP_PATHS_LIST ${CMAKE_SYSTEM_PREFIX_PATH})
modified   pkgs/by-name/me/meson/package.nix
@@ -25,9 +25,6 @@ python3.pkgs.buildPythonApplication rec {
   };
 
   patches = [
-    # Nixpkgs cmake uses NIXPKGS_CMAKE_PREFIX_PATH for the search path
-    ./000-nixpkgs-cmake-prefix-path.patch
-
     # In typical distributions, RPATH is only needed for internal libraries so
     # meson removes everything else. With Nix, the locations of libraries
     # are not as predictable, therefore we need to keep them in the RPATH.

```

paparodeo:
> > yes, because they don't work.
> 
> Is there an upstream issue?

I apologize -- I have no idea what you're trying to do. please test your change and open a new PR. the 80 some comments in this PR log the various methods I tried and why we ended up with the cmake patch. To test the various changes I looked at all the packages which had `cmake` in the `nativeBuildInputs` and then built most of them.

looking over your patch I am pretty sure it doesn't work given these are cmake variables you're modifying and not env vars and you're not passing them via `-D` on the command line to cmake.

I'd ask if you even tried your patch to see if anything works but I don't think that a response belongs in this PR comments and you need to open up a new PR.

PR #318306, opened 2024-06-08
grafana-alloy: add NixOS module

flokli:
cc @vunso @hbjydev @claudiiii 

## Description of changes

    This adds a NixOS module for grafana-alloy.
    
    I started from the grafana-agent, one but dropped all settings and
    config management whatsoever.
    
    grafana-alloy uses its own Alloy config format (similar to HCL), which
    is not really possible to express in Nix.
    Simply pointing to a path in `/etc`, and leaving it up to the user to configure
    it via `environment.etc` allows the user to arrange config files however
    it makes most sense for them.

    It also adds a VM test, ensuring it starts up healthy (with empty config),
    and adds it to the package passthru.
## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [x] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [x] Tested, as applicable:

---

### DIFF
--- a/nixos/modules/module-list.nix
+++ b/nixos/modules/module-list.nix
@@ -834,6 +834,7 @@
   ./services/misc/zoneminder.nix
   ./services/misc/zookeeper.nix
   ./services/monitoring/alerta.nix
+  ./services/monitoring/alloy.nix
   ./services/monitoring/apcupsd.nix
   ./services/monitoring/arbtt.nix
   ./services/monitoring/below.nix
--- a/nixos/modules/services/monitoring/alloy.nix
+++ b/nixos/modules/services/monitoring/alloy.nix
@@ -5,7 +5,7 @@ let
 in
 {
   meta = {
-    maintainers = with maintainers; [ flokli ];
+    maintainers = with maintainers; [ flokli hbjydev ];
   };
 
   options.services.alloy = {
--- a/nixos/modules/services/monitoring/alloy.nix
+++ b/nixos/modules/services/monitoring/alloy.nix
@@ -1,0 +1,80 @@
+{ lib, pkgs, config, ... }:
+with lib;
+let
+  cfg = config.services.alloy;
+in
+{
+  meta = {
+    maintainers = with maintainers; [ flokli ];
+  };
+
+  options.services.alloy = {
+    enable = mkEnableOption "Grafana Alloy";
+
+    package = mkPackageOption pkgs "grafana-alloy" { };
+
+    configPath = mkOption {
+      type = lib.types.path;
+      default = "/etc/alloy";
+      description = ''
+        Alloy configuration file/directory path.
+
+        We default to `/etc/alloy` here, and expect the user to configure a
+        configuration file via `environment.etc."alloy/config.alloy"`.
+
+        This allows config reload, contrary to specifying a store path.
+        A `reloadTrigger` for `config.alloy` is configured.
+
+        Other `*.alloy` files in the same directory (ignoring subdirs) are also
+        honored, but it's necessary to manually extend
+        `systemd.services.alloy.reloadTriggers` to enable config reload
+        during nixos-rebuild switch.
+
+        This can also point to another directory containing `*.alloy` files, or
+        a single Alloy file in the Nix store (at the cost of reload).
+
+        Component names must be unique across all Alloy configuration files, and
+        configuration blocks must not be repeated.
+
+        Alloy will continue to run if subsequent reloads of the configuration
+        file fail, potentially marking components as unhealthy depending on
+        the nature of the failure. When this happens, Alloy will continue
+        functioning in the last valid state.
+      '';
+    };
+
+    extraFlags = mkOption {
+      type = with lib.types; listOf str;
+      default = [ ];
+      example = [ "--server.http.listen-addr=127.0.0.1:12346" "--disable-reporting" ];
+      description = ''
+        Extra command-line flags passed to {command}`alloy run`.
+
+        See <https://grafana.com/docs/alloy/latest/reference/cli/run/>
+      '';
+    };
+  };
+
+
+  config = mkIf cfg.enable {
+    systemd.services.alloy = {
+      wantedBy = [ "multi-user.target" ];
+      reloadTriggers = [ config.environment.etc."alloy/config.alloy".source or null ];
+      serviceConfig = {
+        Restart = "always";
+        DynamicUser = true;
+        RestartSec = 2;
+        SupplementaryGroups = [
+          # allow to read the systemd journal for loki log forwarding
+          "systemd-journal"
+        ];
+        ExecStart = "${lib.getExe cfg.package} run ${cfg.configPath} ${escapeShellArgs cfg.extraFlags}";
+        ExecReload = "${pkgs.coreutils}/bin/kill -SIGHUP $MAINPID";
+        ConfigurationDirectory = "alloy";
+        StateDirectory = "alloy";
+        WorkingDirectory = "%S/alloy";
+        Type = "simple";
+      };
+    };
+  };
+}
--- a/nixos/tests/all-tests.nix
+++ b/nixos/tests/all-tests.nix
@@ -115,6 +115,7 @@ in {
   akkoma = handleTestOn [ "x86_64-linux" "aarch64-linux" ] ./akkoma.nix {};
   akkoma-confined = handleTestOn [ "x86_64-linux" "aarch64-linux" ] ./akkoma.nix { confined = true; };
   alice-lg = handleTest ./alice-lg.nix {};
+  alloy = handleTest ./alloy.nix {};
   allTerminfo = handleTest ./all-terminfo.nix {};
   alps = handleTest ./alps.nix {};
   amazon-init-shell = handleTest ./amazon-init-shell.nix {};
--- a/nixos/tests/alloy.nix
+++ b/nixos/tests/alloy.nix
@@ -1,0 +1,32 @@
+import ./make-test-python.nix ({ lib, pkgs, ... }:
+
+  let
+    nodes = {
+      machine = {
+        services.alloy = {
+          enable = true;
+        };
+        environment.etc."alloy/config.alloy".text = "";
+      };
+    };
+  in
+  {
+    name = "alloy";
+
+    meta = with lib.maintainers; {
+      maintainers = [ flokli ];
+    };
+
+    inherit nodes;
+
+    testScript = ''
+      start_all()
+
+      machine.wait_for_unit("alloy.service")
+      machine.wait_for_open_port(12345)
+      machine.succeed(
+          "curl -sSfN http://127.0.0.1:12345/-/healthy"
+      )
+      machine.shutdown()
+    '';
+  })
--- a/nixos/tests/alloy.nix
+++ b/nixos/tests/alloy.nix
@@ -14,7 +14,7 @@ import ./make-test-python.nix ({ lib, pkgs, ... }:
     name = "alloy";
 
     meta = with lib.maintainers; {
-      maintainers = [ flokli ];
+      maintainers = [ flokli hbjydev ];
     };
 
     inherit nodes;
--- a/pkgs/by-name/gr/grafana-alloy/package.nix
+++ b/pkgs/by-name/gr/grafana-alloy/package.nix
@@ -121,7 +121,7 @@ buildGoModule rec {
     mainProgram = "alloy";
     license = licenses.asl20;
     homepage = "https://grafana.com/oss/alloy";
-    maintainers = with maintainers; [ flokli emilylange ];
+    maintainers = with maintainers; [ flokli emilylange hbjydev ];
     platforms = lib.platforms.unix;
   };
 }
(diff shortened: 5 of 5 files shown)


nixos/modules/services/monitoring/grafana-alloy.nix
```diff
@@ -0,0 +1,80 @@
...
+  };
+
+  options.services.grafana-alloy = {
+    enable = mkEnableOption "grafana-alloy";
+
+    package = mkPackageOption pkgs "grafana-alloy" { };
+
+    configPath = mkOption {
+      type = lib.types.path;
+      default = "/etc/grafana-alloy";
```
hbjydev, line 18: This should probably be /etc/alloy to keep in line with upstream packaging. Or specifically /etc/alloy/config.alloy

hbjydev (changes requested):

nixos/modules/services/monitoring/grafana-alloy.nix
flokli: Good point! I made this default to `/etc/alloy`, and updated the descriptions and `reloadTrigger` to suggest putting a `config.alloy` file in there.

hbjydev (approved):
LGTM

flokli:
Deployed this to some of my infrastructure (using my old grafana flow configs). Also verified changes in the config file cause a reload (not restart), and are picked up successfully.

nixos/modules/services/monitoring/grafana-alloy.nix
```diff
@@ -0,0 +1,80 @@
+{ lib, pkgs, config, ... }:
```
vunso, line 1: @flokli Thanks for change, I was thinking if `grafana-alloy.nix` can go under `nixos/modules/services/logging` instead of `nixos/modules/services/monitoring` as previously [promtail.nix](https://github.com/NixOS/nixpkgs/blob/master/nixos/modules/services/logging/promtail.nix) was under same directory and alloy will replace promtail later on.

flokli: It can also act as a replacement for grafana-agent, prometheus, otlp-collector etc.

It can collect metrics, logs, traces and profiles, it's definitely not just about collecting logs.

vunso (approved):

nixos/modules/services/monitoring/grafana-alloy.nix
hbjydev: @vunso Promtail likely won't be replaced by Alloy; Alloy uses Promtail under the hood for all its `loki.source`, `loki.process` and `loki.relabel` blocks.

flokli: I meant to say, you can use alloy as a replacement for promtail on your machine (and other collectors for other things). It's not obsoleting Promtail upstream.

nixos/modules/services/monitoring/grafana-alloy.nix
```diff
@@ -0,0 +1,80 @@
...
+
+        We default to `/etc/alloy` here, and expect the user to configure a
+        configuration file via `environment.etc."alloy/config.alloy"`.
+
+        This allows config reload, contrary to specifying a store path.
+        A `reloadTrigger` for `config.alloy` is configured.
+
+        Other `*.alloy` files in the same directory (ignoring subdirs) are also
+        honored, but it's necessary to manually extend
+        `systemd.services.grafana-alloy.reloadTriggers to enable config reload
```
hbjydev, line 30:
```suggestion
        `systemd.services.grafana-alloy.reloadTriggers` to enable config reload
```

nixos/modules/services/monitoring/grafana-alloy.nix
```diff
@@ -0,0 +1,80 @@
...
+        file fail, potentially marking components as unhealthy depending on
+        the nature of the failure. When this happens, Alloy will continue
+        functioning in the last valid state.
+      '';
+    };
+
+    extraFlags = mkOption {
+      type = with lib.types; listOf str;
+      default = [ ];
+      example = [ "--server.http.listen-addr=127.0.0.1:12346" "-disable-reporting" ];
```
hbjydev, line 49:
```suggestion
      example = [ "--server.http.listen-addr=127.0.0.1:12346" "--disable-reporting" ];
```

nixos/modules/services/monitoring/grafana-alloy.nix
```diff
@@ -0,0 +1,80 @@
...
+      description = ''
+        Extra command-line flags passed to {command}`grafana-alloy run`.
```
hbjydev, line 51: This may be me being unfamiliar with NixOS standards; but should `{command}` be here?

nixos/modules/services/monitoring/grafana-alloy.nix
```diff
@@ -0,0 +1,80 @@
...
+      wantedBy = [ "multi-user.target" ];
+      reloadTriggers = [ config.environment.etc."alloy/config.alloy".source or null ];
+      serviceConfig = {
+        Restart = "always";
+        DynamicUser = true;
+        RestartSec = 2;
+        SupplementaryGroups = [
+          # allow to read the systemd journal for loki log forwarding
+          "systemd-journal"
+        ];
```
hbjydev, line 70: Should also grant the user the `adm` group too.

https://github.com/grafana/alloy/blob/c7998608fe162e025bf33ccaf4a16cbadf6d2d34/packaging/deb/control/postinst#L21-L24

nixos/modules/services/monitoring/grafana-alloy.nix
```diff
@@ -0,0 +1,80 @@
...
+        ExecStart = "${lib.getExe cfg.package} run ${cfg.configPath} ${escapeShellArgs cfg.extraFlags}";
+        ExecReload = "${pkgs.coreutils}/bin/kill -SIGHUP $MAINPID";
+        ConfigurationDirectory = "alloy";
+        StateDirectory = "grafana-alloy";
+        WorkingDirectory = "%S/grafana-alloy";
```
hbjydev, line 75:
```suggestion
        StateDirectory = "alloy";
        WorkingDirectory = "%S/alloy";
```

Again, keeping in line with upstream packaging.

nixos/modules/services/monitoring/grafana-alloy.nix
```diff
@@ -0,0 +1,80 @@
...
+        Extra command-line flags passed to {command}`grafana-alloy run`.
+
+        See <https://grafana.com/docs/alloy/latest/reference/cli/run/>
+      '';
+    };
+  };
+
+
+  config = mkIf cfg.enable {
+    systemd.services.grafana-alloy = {
```
hbjydev, line 60:
```suggestion
    systemd.services.alloy = {
```

nixos/tests/grafana-alloy.nix
```diff
@@ -0,0 +1,32 @@
...
+    meta = with lib.maintainers; {
+      maintainers = [ flokli ];
+    };
+
+    inherit nodes;
+
+    testScript = ''
+      start_all()
+
+      machine.wait_for_unit("grafana-alloy.service")
```
hbjydev, line 25:
```suggestion
      machine.wait_for_unit("alloy.service")
```

nixos/tests/grafana-alloy.nix
```diff
@@ -0,0 +1,32 @@
...
+        };
+        environment.etc."alloy/config.alloy".text = "";
+      };
+    };
+  in
+  {
+    name = "grafana-alloy";
+
+    meta = with lib.maintainers; {
+      maintainers = [ flokli ];
```
hbjydev, line 17: Happy to help maintain this one too, btw :)

```suggestion
      maintainers = [ flokli hbjydev ];
```

nixos/tests/grafana-alloy.nix
```diff
@@ -0,0 +1,32 @@
+import ./make-test-python.nix ({ lib, pkgs, ... }:
+
+  let
+    nodes = {
+      machine = {
+        services.grafana-alloy = {
```
hbjydev, line 6:
```suggestion
        services.alloy = {
```

hbjydev (changes requested):
So, on a second (more thorough) readthrough, I've got more stuff to add to this; as payment for being a PITA though I'm happy to also add myself to the maintainers list on this one :)

Apologies, just want to make sure it's got decent consistency with the upstream stuff, primarily for the sake of it being an easy transition for people coming from something like Ubuntu where the naming is all just `alloy`, not `grafana-alloy`.

Once these are addressed (even if they're dismissed), I'm happy for this to go through one more build-test cycle and then look to get it merged. I'll deploy it to my own machines later today if I can, but without binary cache it'll be a tad painful lol

nixos/modules/services/monitoring/grafana-alloy.nix
flokli: This is going too far, the systems service unit and nixos module is called "grafana-alloy", I don't want to have this diverge.

nixos/modules/services/monitoring/grafana-alloy.nix
flokli: See above.

nixos/modules/services/monitoring/grafana-alloy.nix
flokli: This applies additional markup to binaries / commands to execute, see other examples in the NixOS module system.

Unfortunately it's not documented in https://nixos.org/manual/nixos/unstable/#chap-contributing (hah, how meta!), but that's out of scope for this PR.

nixos/tests/grafana-alloy.nix
flokli: See above.

nixos/modules/services/monitoring/grafana-alloy.nix
flokli: Done

nixos/modules/services/monitoring/grafana-alloy.nix
flokli: Done

flokli:
Hmmh, okay, their debian package also calls this `alloy`. Let's rename this to `alloy` consistently (including the package).

I'll push a new version shortly.

nixos/modules/services/monitoring/grafana-alloy.nix
flokli: done

nixos/tests/grafana-alloy.nix
flokli: Done

flokli:
Everything done except for the `adm` group change. I propose we merge this in as-is.

if adding the group becomes necessary, we can still fix it in a followup. It wasn't necessary for grafana-agent to get systemd journal logs collected, so I'd be surprised it'd be necessary for alloy.

hbjydev:
@flokli it's not for journal, it's because adm has read access to most of /var/log, so it opens up syslog and stuff. I'd like to avoid removing groups from the user if it's not necessary personally.

nixos/modules/services/monitoring/grafana-alloy.nix
arianvp: Sounds to me the upstream packaging is incorrect.

flokli: Okay, so let's keep it like this. I'll open an upstream issue and link to this. Thanks!

flokli: https://github.com/grafana/alloy/pull/1012

flokli:
> @flokli it's not for journal, it's because adm has read access to most of /var/log, so it opens up syslog and stuff. I'd like to avoid removing groups from the user if it's not necessary personally.

There's nothing in the tmpfiles shipped with NixOS granting access to `/var/log`, only `/var/log/journal`, so this has no effect in terms of allowing more log files to be read out of the box.

I'd prefer this to be opt in, due to the security implications and principle of least privilege, and  `adm` being used to allow `sudo` in some cases.

After all, it's only a `systemd.services.alloy.serviceConfig.SupplementaryGroups` away, or even better, access be allowed on the logs themselves, by granting the `alloy` user access to log files through ACLs.

nixos/modules/services/monitoring/grafana-alloy.nix
flokli: @arianvp  you confused this with `admin`:

> adm: Group adm is used for system monitoring tasks. Members of this group can read many log files in /var/log, and can use xconsole. Historically, /var/log was /usr/adm (and later /var/adm), thus the name of the group.

flokli:
Okay, apparently it's `admin`, not to be confused with `adm`, which should not hand out sudo access. At least assuming stock `sudoers` files, and noone confusing it the other way round :scream: .

I still think there's very little on NixOS these days still logging to `/var/log` , which is not to the journal, so I'd prefer to not add this group to the alloy service by default.

Considering we don't configure ACLs for other files in `/var/log`, users need to do some work to make it available for the alloy service, so I'd rather keep it opt-in (and granting the alloy user access might be the cleaner choice).

arianvp (approved):

PR #317314, opened 2024-06-04
CONTRIBUTING.md: patch conventions

raboof:
## Description of changes

As discussed in the NixOS Security Discussion Matrix room at https://matrix.to/#/!NBBFPbiuttRgTqbrcY:nixos.org/$kRB7R9ksae-78H4GQOO7NjMurKh7hSHimYSGMVE3w8k?via=nixos.org&via=matrix.org&via=nixos.dev and #317169

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
- For non-Linux: Is sandboxing enabled in `nix.conf`? (See [Nix manual](https://nixos.org/manual/nix/stable/command-ref/conf-file.html))
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:
  - [NixOS test(s)](https://nixos.org/manual/nixos/unstable/index.html#sec-nixos-tests) (look inside [nixos/tests](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests))
  - and/or [package tests](https://github.com/NixOS/nixpkgs/blob/master/pkgs/README.md#package-tests)
  - or, for functions and "core" functionality, tests in [lib/tests](https://github.com/NixOS/nixpkgs/blob/master/lib/tests) or [pkgs/test](https://github.com/NixOS/nixpkgs/blob/master/pkgs/test)
  - made sure NixOS tests are [linked](https://nixos.org/manual/nixpkgs/unstable/#ssec-nixos-tests-linking) to the relevant packages
- [ ] Tested compilation of all packages that depend on this change using `nix-shell -p nixpkgs-review --run "nixpkgs-review rev HEAD"`. Note: all changes have to be committed, also see [nixpkgs-review usage](https://github.com/Mic92/nixpkgs-review#usage)
- [ ] Tested basic functionality of all binary files (usually in `./result/bin/`)
- [24.11 Release Notes](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2411.section.md) (or backporting [23.11](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2311.section.md) and [24.05](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2405.section.md) Release notes)
[...]

PR #318499, opened 2024-06-09
refind: make the build reproducible

raboof:
## Description of changes

Avoid leaking build timestamps into the output.

/cc @AndersonTorres @samueldr @chewblacka 

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/tools/bootloaders/refind/default.nix
+++ b/pkgs/tools/bootloaders/refind/default.nix
@@ -34,6 +34,9 @@ stdenv.mkDerivation rec {
   patches = [
     # Removes hardcoded toolchain for aarch64, allowing successful aarch64 builds.
     ./0001-toolchain.patch
+    # Avoid leaking the build timestamp
+    # https://sourceforge.net/p/refind/code/merge-requests/53/
+    ./reproducible.patch
   ];
 
   nativeBuildInputs = [ makeWrapper ];
--- a/pkgs/tools/bootloaders/refind/reproducible.patch
+++ b/pkgs/tools/bootloaders/refind/reproducible.patch
@@ -1,0 +1,67 @@
+diff --git a/Makefile b/Makefile
+index 4d07160..4ae2a7d 100644
+--- a/Makefile
++++ b/Makefile
+@@ -150,9 +150,11 @@ edk2: build_edk2
+ 	cp $(EDK2_BUILDLOC)/refind.efi ./refind/refind_$(FILENAME_CODE).efi
+ 	cp $(EDK2_BUILDLOC)/gptsync.efi ./gptsync/gptsync_$(FILENAME_CODE).efi
+ ifneq ($(OMIT_SBAT), 1)
+-	$(OBJCOPY) --set-section-alignment '.sbat=512' --add-section .sbat=$(REFIND_SBAT_CSV) \
++	$(OBJCOPY) --preserve-dates --set-section-alignment '.sbat=512' \
++		--add-section .sbat=$(REFIND_SBAT_CSV) \
+ 		--adjust-section-vma .sbat+10000000 ./refind/refind_$(FILENAME_CODE).efi
+-	$(OBJCOPY) --set-section-alignment '.sbat=512' --add-section .sbat=$(REFIND_SBAT_CSV) \
++	$(OBJCOPY) --preserve-dates --set-section-alignment '.sbat=512' \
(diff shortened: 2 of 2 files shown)


drupol (approved):
LGTM!

I guess you used a local patch because using the patch directly from upstream (SF) is not available.

chewblacka:
**Current Master**
Result of current `nix-build -A refind --check`:

`error: derivation '/nix/store/nif8vnzh5j65s3hlzjf88ca9af0kfvag-refind-0.14.2.drv' may not be deterministic: output '/nix/store/sq3rgaih1cmkpdj8qwi636q6cjr1yplm-refind-0.14.2' differs`

**PR 318499**

Result of `nixpkgs-review pr 318499`
```
1 package built:
refind
```

Result of `nix-build -A refind --check`
No errors (i.e. build is reproducible).

chewblacka (approved):

raboof:
> I guess you used a local patch because using the patch directly from upstream (SF) is not available.

I haven't really checked, but since the MR isn't merged yet that commit is still at risk of being garbage-collected, so this is safer (as per #317314)

pkgs/tools/bootloaders/refind/default.nix
```diff
@@ -34,6 +34,9 @@ stdenv.mkDerivation rec {
   patches = [
     # Removes hardcoded toolchain for aarch64, allowing successful aarch64 builds.
     ./0001-toolchain.patch
+    # Avoid leaking the build timestamp
+    # https://sourceforge.net/p/refind/code/merge-requests/53/
+    ./reproducible.patch
```
AndersonTorres, line 39:
```suggestion
    ./0002-preserve-dates.patch
```

raboof: Ah, sorry, I had merged already but am happy to follow up.

Is the number prefix really useful? The ordering is already specified by the list of patches, having the ordering in the filenames as well just seems to make things messy when patches are added or removed later. On the other hand, indeed it seems somewhat common: about 500 of the 3700 patches in nixpkgs seem to be numbered that way.

drupol: I think you just need to use git format-patch IIRC.

raboof: I mean, I can, but I'm not sure I understand the benefit.

(indeed `preserve-dates` might have been a better name though)

drupol: Yeah I don't really understand the benefits either.

raboof: Right, so it makes it easier to work with `git am` and rebase all patches in one go. Good to know - though indeed probably not universally necessary.

samueldr: We should prefer `git format-patch` patches all the time. Not for the filename, but for the additional information in its header, and tracking authorship information.

raboof: Good idea. Could y'all add those recommendations to https://github.com/NixOS/nixpkgs/tree/master/pkgs#patches ?

SuperSandro2000: > We should prefer `git format-patch` patches all the time. Not for the filename, but for the additional metadata in its header, and tracking authorship information.

I wouldn't want to write that as a hard requirement. It doesn't work when changes are done on top of a downloaded tarball and we already have most of the metadata like who wrote it, when it was written and there should be a short comment above the patch why we need it.

AndersonTorres: My main concern with the numbering and the name is that patches are meant to be applied in sequence. (Sometimes patches can be applied unordered, but sometimes not.)
(Except maybe in Pijul, but I digress.)

Yes, lists in Nix are positional, but since sometimes the patch can be useful outside Nixpkgs - by a question of principle:  we pick patches from other distros, and they can pick patches from us - it is a good idea to put an order on the patches.

P.S.: git-format-patch can be used outside git? Sometimes the patch comes from other VCSes...

PR #318599, opened 2024-06-09
nixos/fcgiwrap: refactor to fix permissions

pacien:
## Description of changes

This PR fixes mulitple issues with the fcgiwrap module:
- It's now possible to run multiple instances of fcgiwrap,
  solving the interference issue between its consumer modules.
- The service and the spawned scripts no longer run as root by default.
- The UNIX socket is no longer exposed to everyone by default.

With the previous defaults in the module,
any local user could execute scripts as root through the socket when any of
`services.{fcgiwrap,cgit,smokeping}.enable = true` was set.
The smokeping and cgit services were indeed partially and fully running as
root. This has also been fixed in this PR.

This changes the exposed options quite a bit, making this change backward
incompatible. The security benefits might however make it worth backporting.

See indidivual commit messages and modified release notes for details.

I tested these changes by running the NixOS tests of some consumers of this
module: `nix build .#nixosTests.{smokeping,zoneminder,gitolite-fcgiwrap,cgit}`

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [x] Tested, as applicable:

---

### DIFF
--- a/nixos/modules/services/web-servers/fcgiwrap.nix
+++ b/nixos/modules/services/web-servers/fcgiwrap.nix
@@ -13,7 +13,7 @@ in {
     default = { };
     type = types.attrsOf (types.submodule ({ config, ... }: { options = {
       process.prefork = mkOption {
-        type = types.int;
+        type = types.ints.positive;
         default = 1;
         description = "Number of processes to prefork.";
       };
--- a/nixos/modules/services/web-servers/fcgiwrap.nix
+++ b/nixos/modules/services/web-servers/fcgiwrap.nix
@@ -3,17 +3,15 @@
 with lib;
 
 let
-  cfg = config.services.fcgiwrap;
+  forEachInstance = f: flip mapAttrs' config.services.fcgiwrap (name: cfg:
+    nameValuePair "fcgiwrap-${name}" (f cfg)
+  );
+
 in {
-
-  options = {
-    services.fcgiwrap = {
-      enable = mkOption {
-        type = types.bool;
-        default = false;
-        description = "Whether to enable fcgiwrap, a server for running CGI applications over FastCGI.";
-      };
-
+  options.services.fcgiwrap = mkOption {
+    description = "Configuration for fcgiwrap instances.";
+    default = { };
+    type = types.attrsOf (types.submodule ({ config, ... }: { options = {
       preforkProcesses = mkOption {
         type = types.int;
         default = 1;
@@ -28,7 +26,7 @@ in {
 
       socketAddress = mkOption {
         type = types.str;
-        default = "/run/fcgiwrap.sock";
+        default = "/run/fcgiwrap-${config._module.args.name}.sock";
         example = "1.2.3.4:5678";
         description = "Socket address. In case of a UNIX socket, this should be its filesystem path.";
       };
@@ -44,11 +42,11 @@ in {
         default = null;
         description = "Group permissions for the socket.";
       };
-    };
+    }; }));
   };
 
-  config = mkIf cfg.enable {
-    systemd.services.fcgiwrap = {
+  config = {
+    systemd.services = forEachInstance (cfg: {
       after = [ "nss-user-lookup.target" ];
       wantedBy = optional (cfg.socketType != "unix") "multi-user.target";
 
@@ -60,13 +58,13 @@ in {
         User = cfg.user;
         Group = cfg.group;
       } else { } );
-    };
+    });
 
-    systemd.sockets = if (cfg.socketType == "unix") then {
-      fcgiwrap = {
-        wantedBy = [ "sockets.target" ];
-        socketConfig.ListenStream = cfg.socketAddress;
+    systemd.sockets = forEachInstance (cfg: mkIf (cfg.socketType == "unix") {
+      wantedBy = [ "sockets.target" ];
+      socketConfig = {
+        ListenStream = cfg.socketAddress;
       };
-    } else { };
+    });
   };
 }
--- a/nixos/modules/services/web-servers/fcgiwrap.nix
+++ b/nixos/modules/services/web-servers/fcgiwrap.nix
@@ -54,9 +54,13 @@ in {
       wantedBy = optional (cfg.socket.type != "unix") "multi-user.target";
 
       serviceConfig = {
-        ExecStart = "${pkgs.fcgiwrap}/sbin/fcgiwrap -c ${builtins.toString cfg.process.prefork} ${
-          optionalString (cfg.socket.type != "unix") "-s ${cfg.socket.type}:${cfg.socket.address}"
-        }";
+        ExecStart = ''
+          ${pkgs.fcgiwrap}/sbin/fcgiwrap ${cli.toGNUCommandLineShell {} ({
+            c = cfg.process.prefork;
+          } // (optionalAttrs (cfg.socket.type != "unix") {
+            s = "${cfg.socket.type}:${cfg.socket.address}";
+          }))}
+        '';
       } // (if cfg.process.user != null && cfg.process.group != null then {
         User = cfg.process.user;
         Group = cfg.process.group;
--- a/nixos/modules/services/web-servers/fcgiwrap.nix
+++ b/nixos/modules/services/web-servers/fcgiwrap.nix
@@ -12,35 +12,38 @@ in {
     description = "Configuration for fcgiwrap instances.";
     default = { };
     type = types.attrsOf (types.submodule ({ config, ... }: { options = {
-      preforkProcesses = mkOption {
+      process.prefork = mkOption {
         type = types.int;
         default = 1;
         description = "Number of processes to prefork.";
       };
 
-      socketType = mkOption {
+      process.user = mkOption {
+        type = types.nullOr types.str;
+        default = null;
+        description = "User as which this instance of fcgiwrap will be run.";
+      };
+
+      process.group = mkOption {
+        type = types.nullOr types.str;
+        default = null;
+        description = "Group as which this instance of fcgiwrap will be run.";
+      };
+
+      socket.type = mkOption {
         type = types.enum [ "unix" "tcp" "tcp6" ];
         default = "unix";
         description = "Socket type: 'unix', 'tcp' or 'tcp6'.";
       };
 
-      socketAddress = mkOption {
+      socket.address = mkOption {
         type = types.str;
         default = "/run/fcgiwrap-${config._module.args.name}.sock";
         example = "1.2.3.4:5678";
-        description = "Socket address. In case of a UNIX socket, this should be its filesystem path.";
-      };
-
-      user = mkOption {
-        type = types.nullOr types.str;
-        default = null;
-        description = "User permissions for the socket.";
-      };
-
-      group = mkOption {
-        type = types.nullOr types.str;
-        default = null;
-        description = "Group permissions for the socket.";
+        description = ''
+          Socket address.
+          In case of a UNIX socket, this should be its filesystem path.
+        '';
       };
     }; }));
   };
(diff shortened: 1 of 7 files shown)


LeSuisse (approved):
The approach looks sensible to me and I'm also find with the changes.

Not quite sure how to handle it on stable :/

minijackson (approved):
Changes looks quite good, and quite useful. Thanks! I'm interested in running `cgit` as non-root.

LeSuisse:
Moving to merge this, flagging for backport to 24.05 so we can discuss how we want to handle it.

bendlas:
Handling of this PR leaves a lot to be desired:

- removing options is a lot more than just "fixing permissions"
- making a service multi-instance is a lot more than just "fixing permissions"
- if existing rename/deprecation helpers cannot be used, please add custom warnings
- if a change to an important service like this cannot be made backward compatible, it's better to roll a new service name, instead of breaking the existing one

> Not quite sure how to handle it on stable :/

@LeSuisse easy: we don't. It's bad enough to break options between releases. Breaking options on a stable release is absolutely unacceptable. I'd usually not blame the author in this, but since @pacien is also a Member, you should also be held to a higher standard.

I'd also implore everybody involved in this mess to rethink a few options to move forward from this:

- keep the old options, and have an instance called `default`, to which the previous options map
- handle different features / refactors in different PRs
- think about our users. we have production deployments, depending on this. nobody cares if you think that this is better and we can just update our config.

bendlas:
On a way forward: I do see value in making fcgiwrap multi-instance, so I wouldn't like to have to outright revert all of this.

How about we introduce the multi-instance options as maybe `services.fcgiwraps.<name>` and deprecate `services.fcgiwrap`, forwarding it to `services.fcgiwraps.default`?

mweinelt:
> I'm interested in running `cgit` as non-root.

That should have never been a thing. Thanks for cleaning that up.

> - removing options is a lot more than just "fixing permissions"
> - making a service multi-instance is a lot more than just "fixing permissions"

I agree that the scope of this PR goes far beyond removing the obvious footgun and can't be backported in full.

> flagging for backport to 24.05 so we can discuss how we want to handle it.

But that was also an open discussion question.

> keep the old options, and have an instance called default, to which the previous options map

I think doing breaking changes to module interfaces is reasonable on unstable, and ideally we'd have escape hatches for functionality that was covered by removed options.

> think about our users. we have production deployments, depending on this. nobody cares if you think that this is better and we can just update our config. this is not NPM and many of us moved into nixos, exactly to escape the never-ending churn.

No, you don't get to make this "about our users", if your production deployments run on NixOS *Unstable*. You will have to deal with breaking changes that we usually announce in release notes every once in a while. Of course that'll hit you faster on NixOS Unstable.

> How about we introduce the multi-instance options as maybe services.fcgiwraps.<name> and deprecate services.fcgiwrap, forwarding it to services.fcgiwraps.default?

I think that is undesirable.

bendlas:
> No, you don't get to make this "about our users", if your production deployments run on NixOS _Unstable_. You will have to deal with breaking changes that we usually announce in release notes every once in a while. Of course that'll hit you faster on NixOS Unstable.

I do get to, I did, and I think the whole community should do more of that.

But speaking of that: I do see a release note update, but where are the upgrade instructions in that?

> > How about we introduce the multi-instance options as maybe services.fcgiwraps. and deprecate services.fcgiwrap, forwarding it to services.fcgiwraps.default?
> 
> I think that is undesirable.

How so?

LeSuisse:
Hello,

> Breaking options on a stable release is absolutely unacceptable. I'd usually not blame the author in this, but since @pacien is also a Member, you should also be held to a higher standard.

Thank you for the feedback and I can understand the frustration.

To be clear, I merged this because:
* I did not see a way to fix the footgun/LPE introduced by the existing module in a backward compatible way
* the bundle with the support of multi-instances made sense and something desirable, especially since users are going to need to check/adjust their config for the first bullet point anyway
* I did not have the intention to merge it as is in stable but it is easier to have a discussion on what we do and how we can do it if we have something concrete to discuss on

bendlas:
> * I did not see a way to fix the footgun/LPE introduced by the existing module in a backward compatible way

That's what I don't understand: Are you putting "tightening the socket permissions" on the same level of incompatibility as "changing nixos options in an incompatible way"? If so, is there a way to convince you that those aren't even in the same ballpark of breakage?

Or is there another security enhancement that I'm failing to see?

> * the bundle with the support of multi-instances made sense and something desirable, especially since users are going to need to check/adjust their config for the first bullet point anyway

Most wouldn't, because the only thing accessing the fcgiwrap socket will be nginx in many cases.

Besides that, I do support the multi-instance thing. And I'm still wondering what @mweinelt has against pulling up a new hierarchy and mapping the old one to that.

LeSuisse:
It has been an hectic week and I'm too tired to work efficiently on this tonight. I will get back to it tomorrow to see how we improve the migration phase/improve the warnings and provide more context around the whole thing.

Since I saw you were looking for my username on Discourse, it's [@tgerbet](https://discourse.nixos.org/u/tgerbet) (and I'm `@tgerbet:matrix.org` in the Matrix rooms if you want to reach me)

bendlas:
@LeSuisse thanks for the olive branch and for staying level-headed.

I'd like ask your forgiveness (as well as other participants in this PR) for me escalating this immediately and putting your nose to the grindstone to that degree.

In a sense, it was just unlucky for this particular PR to be the one where I've finally had enough, and me getting that "loud" wasn't exactly deserved for what happened here.

 I'm 100% willing to help sort this out in a productive and constructive way (including just letting it be), if you still want my help after this.

emilazy:
> While already problematic when taken separately, both of these flaws combined allowed any local user to execute any program as root on the machine.

Thank you for your work on fixing this vulnerability!

Are there any plans to issue a security advisory of some kind for users that outlines the risk they’ve been exposed to, perhaps after this is fixed on stable? I would be willing to help out if people think it’s desirable and nothing has been drafted yet.

LeSuisse:
> Are there any plans to issue a security advisory of some kind for users that outlines the risk they’ve been exposed to, perhaps after this is fixed on stable? I would be willing to help out if people think it’s desirable and nothing has been drafted yet.

At least it was my intention and yes I would have hopped to have something for stable ready.
That's said given the current situation we can probably issue something and update it once we have an agreement for 24.05.

emilazy:
Good to hear it’s in the pipeline, feel free to ping me on Matrix if I can be of any help :)

Is there a mitigation users on stable can apply to their configurations to lock down the control socket, or will it require changes on stable before anything can be done?

LeSuisse:
> I'm not familiar with how advisory are handled. In particular, who should write and publish them, and where?

We recently started to publish them in the following Discourse sub category: https://discourse.nixos.org/c/announcements/security/56

I was starting to craft something up but it's all your work so if you want to do it please help yourself (and thanks for your work on this :heart: ). Do not hesitate to post it in the [#security-discuss Matrix room](https://matrix.to/#/#security-discuss:nixos.org) if you want some reviews before publishing it.


> I don't think there's any way we can backport anything without breaking any existing setup.

I have the same conclusion but maybe we could at least add an assertion throwing a message when `services.fcgiwrap.user` and `services.fcgiwrap.group` are null linking to the advisory. Yes, it would break things but that would make people aware of the situation. If they still want to go with it they can set `services.fcgiwrap.user = "root";` and `services.fcgiwrap.group = "root";` explicitly.

emilazy:
I’m a bit worried about starting another unproductive discussion, but if there’s no way to protect people backwards‐compatibly and the manual mitigation is fairly complex, maybe it might make sense to backport this to stable after all. Silently leaving people with a known‐insecure configuration seems very bad, but making people add involved manual mitigations to their configuration, when they’ll have to change *again* upon upgrading to the next stable release, seems silly. As an analogy, Rust’s backwards‐compatibility guarantee is explicitly waived for soundness issues. Of course, we’d need to make sure we have good error messages in that case.

nh2:
My opinions on how this should be done, as an industrial user (meaning running servers that people depend on):

* The writeup https://github.com/NixOS/nixpkgs/pull/318599#issuecomment-2211515713 by @pacien is a great and exactly what's needed to make good decisions.
* The `security` tag is correct here. This was a big Local Privilege Escalation bug.
* This should _not_ be backported to stable if it requires manual user interaction:
  * Lots of people run stable with [`system.autoUpgrade`](https://search.nixos.org/options?channel=24.05&show=system.autoUpgrade.enable&from=0&size=50&sort=relevance&type=packages&query=autoUpgrade) to get 0-manual-involvement security updates (great value!). They rely on servers staying online, and getting automatic fixes for critical security bugs, such as the recent [OpenSSH Remote Code Execution vulnerability](https://discourse.nixos.org/t/security-advisory-openssh-cve-2024-6387-regresshion-update-your-servers-asap/48220/21).
  * Thus **stable changes should never break `autoUpgrade`**.
    Introducing e.g. a `throw` error during evaluation (e.g. telling the user that they should change their NixOS options) would stop automatic updates, preventing systems from automatically getting much more important fixes, like the OpenSSH fix.
  * Making a change stable that just turns the services in question _off_ would be wrong. Lots of servers are set up such that Local Privilege Escalation does not matter / is a very small threat. The services staying online is important. Stable should backport straightforward fixes; when re-archtitecturing is needed to fix wrong security architecture (as happens here), that should be in `unstable` and be part of the next release.
* What this PR does (multiple independent fcgiwrap instances) is the correct technical solution.
* This PR is acceptable for `nixos-unstable`.

But I think it would have been even better to:

* Do the fix as `services.fcgiwraps`, keep `services.fcgiwrap` unchanged.
  (As suggested by @bendlas.)
  * This allows trivial backporting of this change to stable without breakage.
  * Also reduces damage and allows users to switch back if we find that the fix is not quite correct.
  * Lets stable users choose whether fixing a Local Privilege Escalation is something they need to fix now or with the next release.
* Immediately wrap an eval warning around `services.fcgiwrap`, land that in unstable and stable.
  This will make all users that interactively upgrade NixOS notice and upgrade fast, while not breaking `autoUpgrade`.
* Make a security announcement (as proposed above, [here](https://discourse.nixos.org/c/announcements/security/56)).
* On unstable, add a warning that `services.fcgiwrap` is insecure and unmaintained, and will be removed soon.
  Remove the old module shortly before the release of NixOS 24.11, or remove it for NixOS 25.05 (removing it later would allow users to upgrade NixOS independently from switching to the new module architecture as a minor added convenience).

Doing that seems to have no drawbacks.

nixos/modules/services/web-servers/fcgiwrap.nix
```diff
@@ -3,70 +3,128 @@
...
       };
 
-      user = mkOption {
+      socket.user = mkOption {
         type = types.nullOr types.str;
         default = null;
-        description = "User permissions for the socket.";
+        description = ''
+          User to be set as owner of the UNIX socket.
+          Defaults to the process running user.
```
sternenseemann, line 57: This is not true (also in the case of true). If unset, it defaults to the empty string in the systemd configuration which results in the socket being owned by `root`.

LeSuisse:
>     * This should _not_ be backported to stable if it requires manual user interaction:
>       
>       * Lots of people run stable with [`system.autoUpgrade`](https://search.nixos.org/options?channel=24.05&show=system.autoUpgrade.enable&from=0&size=50&sort=relevance&type=packages&query=autoUpgrade) to get 0-manual-involvement security updates (great value!). They rely on servers staying online, and getting automatic fixes for critical security bugs, such as the recent [OpenSSH Remote Code Execution vulnerability](https://discourse.nixos.org/t/security-advisory-openssh-cve-2024-6387-regresshion-update-your-servers-asap/48220/21).
>       * Thus **stable changes should never break `autoUpgrade`**.
>         Introducing e.g. a `throw` error during evaluation (e.g. telling the user that they should change their NixOS options) would stop automatic updates, preventing systems from automatically getting much more important fixes, like the OpenSSH fix.
>       * Making a change stable that just turns the services in question _off_ would be wrong. Lots of servers are set up such that Local Privilege Escalation does not matter / is a very small threat. The services staying online is important. Stable should backport straightforward fixes; when re-archtitecturing is needed to fix wrong security architecture (as happens here), that should be in `unstable` and be part of the next release.
> 

It is unrelated to this change but note this already happen from time to time when packages get tagged with `knownVulnerabilities` due to the lack of a better option.

Assuming the threat model of everyone by concluding that a LPE to root does not matter / is a very small threat is also probably not something that should be done at a distro level. We do not have access to the context of all possible deployments that could allow us to say that with certitude.

I'm also a bit puzzled by the fact the described scenario assumes people will not notice a failure in the auto-upgrade mechanism but they will notice an advisory.

nh2:
> Assuming the threat model of everyone by concluding that a LPE to is a very small threat [..] not something that should be done at a distro level

@LeSuisse I didn't say "everybody", I said "lots of servers are set up such that LPE is a very small threat". Most servers I've encountered in industry in the last 10 years were not "multi UNIX user systems" where multiple humans SSH in. So LPE is _usually_ a danger of risk escalation from one already-compromised service to root. But of course true multi-user systems still exist, e.g. universities often have them.

My point is that a distro should not assume that users prefer their services being turned off over continuing running with a bug that may or may not be a real risk for them. The users should decide that.

(Of course the distro should automatically deploy any straightforward fixes; here we're talking only about the situation where you have to pick between turning the service off or keeping it running bugged.)

I also want to point out that turning the service off explicitly is the only thing NixOS could do for a backwards-incompatible change. Because just inserting `throw` would make evaluation fail, but that keeps existing services running (and thus brings no security fix but turns off other updates).

> I'm also a bit puzzled by the fact the described scenario assumes people will not notice a failure in the auto-upgrade mechanism but they will notice an advisory.

Can I configure [`system.autoUpgrade`](https://search.nixos.org/options?channel=24.05&from=0&size=50&sort=relevance&type=packages&query=autoUpgrade) easily to notify me about failures?

For the advisories, all I need to do is to click the bell icon on https://discourse.nixos.org/c/announcements/security/56, very easy and convenient:

image

Issue #57577, opened 2019-03-13
Request to add options for configuring the IO scheduler for NixOS


CMCDragonkai:
Right now there's no option for NixOS users to configure the IO schedulers for the disk block devices.

To do this right now, one has to write an activation script that iterates over each block device and sets the IO scheduler. 

Something like this:

```
    system.activationScripts.diskScheduler = ''
      for path in "''${disks[@]}"; do
        echo noop >/sys/block/$(
          ${pkgs.coreutils}/bin/basename $(
            ${pkgs.coreutils}/bin/readlink -f $path
          )
        )/queue/scheduler
      done
    '';
```

It would be really nice and robust, if NixOS had dedicated option to do this. It would have to be configurable on each block device. So it would not correspond to the `fileSystems` option. Instead it would be more closer to the `swapDevices` option. Perhaps we should have a `blockDevices` option in general that organises all the options relating to connected block devices (which would have to be separate from udev like devices).

tadfisher:
The proper way to do this is to use udev.

For example, this is in my configuration:

```nix
  services.udev.extraRules = ''
    ACTION=="add|change", KERNEL=="[sv]d[a-z]", ATTR{queue/rotational}=="0", ATTR{queue/scheduler}="bfq"
  '';
```

worldofpeace:
Maybe we should have option, just because it's such a common thing, that generates a udev rule like the one mentioned?

CMCDragonkai:
We need to be allowed to set the IO scheduler for different devices differently. I suggest 2 kinds of settings, one that allows us to set it to be the same for all (any block device), another that sets individual ones. They can be combined so that you can set the scheduler for all, and then individual override it for specific devices.

fadenb:
I think this still would be a nice functionality.

CMCDragonkai:
For Linux 5.0, this should now be set to use `none` rather than `noop`. The `none` is the equivalent of `noop`, as there's no more `noop`.

This is quite important for using ZFS to prevent using other IO schedulers with ZFS.

Atemu:
Still important to me.

gshpychka:
This should be closed - `hardware.block.scheduler` exists

PR #318600, opened 2024-06-09
nixos/iosched: init module

illdefined:
## Description of changes

This module introduces the following options to set I/O schedulers for block devices and generates appropriate udev rules:

- `hardware.block.defaultScheduler`: Default scheduler for all block devices.
- `hardware.block.defaultSchedulerRotational`: Default scheduler for rotational drives, such as hard drives.
- `hardware.block.scheduler`: Schedulers for specific block devices.

This solves #57577. My personal motivation is that I often run workloads that do not perform well with the default scheduler and I would like to have a way to override it without writing individual udev rules for simple cases.

`hardware.block.scheduler` is an attribute set assigning schedulers by device name pattern. Using an attrset limits its use to non‐overlapping patterns. Overlapping patterns could be supported by instead using a list of attrsets, but I believe that users requiring these are served better by producing their own custom udev rules.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [x] Tested, as applicable:
- [ ] ~Tested basic functionality of all binary files (usually in `./result/bin/`)~

---

### DIFF
--- a/nixos/modules/hardware/iosched.nix
+++ b/nixos/modules/hardware/iosched.nix
@@ -1,0 +1,150 @@
+{
+  config,
+  lib,
+  pkgs,
+  ...
+}:
+
+let
+  cfg = config.hardware.block;
+  escape = lib.strings.escape [ ''"'' ];
+  udevValue = types.addCheck types.nonEmptyStr (x: builtins.match "[^\n\r]*" x != null) // {
+    name = "udevValue";
+    description = "udev rule value";
+    descriptionClass = "noun";
+  };
+
+  inherit (lib)
+    mkIf
+    mkOption
+    types
+
+    concatLines
+    concatStringsSep
+    mapAttrsToList
+    optional
+    ;
+in
+{
+  options.hardware.block = {
+    defaultScheduler = mkOption {
+      type = types.nullOr udevValue;
+      default = null;
+      description = ''
+        Default block I/O scheduler.
+
+        Unless `null`, the value is assigned through a udev rule matching all
+        block devices.
+      '';
+      example = "kyber";
+    };
+
+    defaultSchedulerRotational = mkOption {
+      type = types.nullOr udevValue;
+      default = null;
+      description = ''
+        Default block I/O scheduler for rotational drives (e.g. hard disks).
+
+        Unless `null`, the value is assigned through a udev rule matching all
+        rotational block devices.
+
+        This option takes precedence over
+        {option}`config.hardware.block.defaultScheduler`.
+      '';
+      example = "bfq";
+    };
+
+    scheduler = mkOption {
+      type = types.attrsOf udevValue;
+      default = { };
+      description = ''
+        Assign block I/O scheduler by device name pattern.
+
+        Names are matched using the {manpage}`udev(7)` pattern syntax:
+
+        `*`
+        :  Matches zero or more characters.
+
+        `?`
+        :  Matches any single character.
+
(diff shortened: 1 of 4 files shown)


Atemu (approved):
This is something I think NixOS needs.

The diff LGTM and is very high quality I might add.

Could you go into detail what testing you have done on this? A NixOS test would also be amazing but not required.

One thing I was a little unsure of was that we build bfq as a module and don't load it by default but the kernel appears to just load the module when bfq is asked for.

illdefined:
> This is something I think NixOS needs.
> 
> The diff LGTM and is very high quality I might add.

Thank you!

> Could you go into detail what testing you have done on this? A NixOS test would also be amazing but not required.

I have been using this module as part of my Nix files with the following configuration since mid of last year without any issues:

```
hardware.block = {
  defaultScheduler = "kyber";
  defaultSchedulerRotational = "bfq";
  scheduler = {
    "mmcblk*" = "bfq";
  };
};
```

Testing this in a virtualised environment could be tricky, but I’ll think of something.

I suppose that providing a basic configuration and verifying that the generated udev rules are correct would be a good start, but it might be nice to check if the settings are actually applied to block devices.

illdefined:
I modified the module to reject values containing a line break, because these are not permitted within udev rules.

I also added a test to check a reasonable module configuration results in a syntactically and semantically correct udev rule set as verified by `udevadm verify --no-style` and that these rules are applied correctly to various block devices.

nixos/modules/hardware/iosched.nix
```diff
@@ -0,0 +1,150 @@
...
+}:
+
+let
+  cfg = config.hardware.block;
+  escape = lib.strings.escape [ ''"'' ];
+  udevValue = types.addCheck types.nonEmptyStr (x: builtins.match "[^\n\r]*" x != null) // {
+    name = "udevValue";
+    description = "udev rule value";
+    descriptionClass = "noun";
+  };
```
Atemu, line 15: 👌

This could perhaps be made available as a type or something as this should be something that is generally useful in other modules/user config.

Atemu (approved):
I didn't give it very thorough testing because the code looks very good to me and the NixOS test gives me confidence that it does work as expected.

The basic functionality works and there isn't anything obviously wrong with it. Should any issues crop up, we can fix them later.

Amazing work, thank you :)

Atemu:
Ah one minor thing I missed: This should have a release note!

I consider this module to be useful to many people, so they should be made aware that it exists now.

illdefined:
> Ah, one thing I noticed in actual deployment is that this affects loop devices which probably shouldn't have an iosched? I'm actually not quite sure in that regard. Perhaps it can make sense to have a nested queue in some cases?
> 
> It's easy enough to disable using the provided options but perhaps the module should handle that case somehow; either through default config or through documentation.

I’ll prepare a PR.

nixos/modules/hardware/iosched.nix
illdefined: Do you think it might be useful to have a generic framework for representing udev rules in Nix?

Atemu: That sounds like a good idea to me, though not necessarily a very high priority.

PR #318604, opened 2024-06-09
nixos/systemd-networkd: add PREF64 related options

jmbaur:
## Description of changes

Tested to work in the legacy nested `*Config` format as well as the newer flattened format.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/nixos/modules/system/boot/networkd.nix
+++ b/nixos/modules/system/boot/networkd.nix
@@ -952,6 +952,7 @@ let
           "UseGateway"
           "UseRoutePrefix"
           "Token"
+          "UsePREF64"
         ])
         (assertValueOneOf "UseDNS" boolValues)
         (assertValueOneOf "UseDomains" (boolValues ++ ["route"]))
@@ -962,6 +963,7 @@ let
         (assertValueOneOf "UseMTU" boolValues)
         (assertValueOneOf "UseGateway" boolValues)
         (assertValueOneOf "UseRoutePrefix" boolValues)
+        (assertValueOneOf "UsePREF64" boolValues)
       ];
 
       sectionDHCPServer = checkUnitConfig "DHCPServer" [
@@ -1031,6 +1033,14 @@ let
         (assertValueOneOf "RouterPreference" ["high" "medium" "low" "normal" "default"])
         (assertValueOneOf "EmitDNS" boolValues)
         (assertValueOneOf "EmitDomains" boolValues)
+      ];
+
+      sectionIPv6PREF64Prefix = checkUnitConfigWithLegacyKey "ipv6PREF64PrefixConfig" "IPv6PREF64Prefix" [
+        (assertOnlyFields [
+          "Prefix"
+          "LifetimeSec"
+        ])
+        (assertInt "LifetimeSec")
       ];
 
       sectionIPv6Prefix = checkUnitConfigWithLegacyKey "ipv6PrefixConfig" "IPv6Prefix" [
@@ -2009,6 +2019,16 @@ let
       description = ''
         Each attribute in this set specifies an option in the
         `[IPv6SendRA]` section of the unit.  See
+        {manpage}`systemd.network(5)` for details.
+      '';
+    };
+
+    ipv6PREF64Prefixes = mkOption {
+      default = [];
+      example = [ { Prefix = "64:ff9b::/96"; } ];
+      type = types.listOf (mkSubsectionType "ipv6PREF64PrefixConfig" check.network.sectionIPv6PREF64Prefix);
+      description = ''
+        A list of IPv6PREF64Prefix sections to be added to the unit. See
         {manpage}`systemd.network(5)` for details.
       '';
     };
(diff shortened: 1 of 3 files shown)


flokli:
Is there a chance we can actually add a test for this?

jmbaur:
> Is there a chance we can actually add a test for this?

Sure thing! A functional test or a test that the config is what we expect?

nixos/modules/system/boot/networkd.nix
```diff
@@ -2013,6 +2023,16 @@ let
       '';
     };
 
+    ipv6PREF64Prefixes = mkOption {
+      default = [];
+      example = [ { Prefix = "2001:db8::/96"; } ];
```
mweinelt, line 2028:
```suggestion
      example = [ { Prefix = "64:ff9b::/96"; } ];
```

https://www.rfc-editor.org/rfc/rfc6052

jmbaur: I don't mind either way, but we have a few other places in nixpkgs where NAT64-related things are being done, and we use the rfc3849 documentation prefix (albeit with varying prefix sizes) for those. I guess it's a matter of how the examples are interpreted, as an example of how a valid config looks, or as an example of a config that would be used in the real world. For consistency, I think keeping it as is is preferred.

nixos/modules/system/boot/networkd.nix
```diff
@@ -2013,6 +2023,16 @@ let
       '';
     };
 
+    ipv6PREF64Prefixes = mkOption {
+      default = [];
+      example = [ { Prefix = "64:ffb9::/96"; } ];
```
mweinelt, line 2028: ff9b, not ffb9

jmbaur: Thanks!

jmbaur:
@flokli So we do have a clatd test! I added a commit that augments that test to ensure the networkd options work as expected.

jmbaur:
@ofborg test clatd

nixos/tests/clatd.nix
```diff
@@ -102,6 +103,37 @@ import ./make-test-python.nix ({ pkgs, lib, ... }:
...
+        networkConfig.IPv6SendRA = true;
+        ipv6Prefixes = [ { Prefix = "2001:db8::/64"; } ];
+        ipv6PREF64Prefixes = [ { Prefix = "64:ff9b::/96"; } ];
+        ipv6SendRAConfig = {
+          EmitDNS = true;
+          DNS = "_link_local";
+        };
+      };
+
+      services.resolved.extraConfig = ''
```
flokli, line 116: So resolved does not have any support for this yet? Is there an upstream issue we can link to?

I assume once that's supported, we would not need to do any custom DNS setup here?

jmbaur: This config is just set as to not conflict with the coredns instance running. clatd will fetch the PLAT prefix via DNS64, which is served up by coredns in the configuration I made.

jmbaur: Ah I see what you're saying, does resolved have support for serving up DNS64 synthesized records. AFAIK no it does not, and I don't see any issue upstream.

flokli: Yes, exactly. Would you feel like opening one?

- This feels like a logical successor after the pref64 stuff, so it would make sense to signal interest in it
- It would allow others looking at the code and module options to understand why we went with the coredns route 
- it makes it easy to know when we can drop coredns from here again

jmbaur: I'm actually not super sure that adding a DNS64 component to systemd-resolved would be a good idea. If a client running systemd-networkd is already able to obtain the PREF64 prefix from the router advertisements, what good is it for systemd-resolved to serve up modified DNS responses to applications when the machine should be smart enough to setup a clat interface so the unmodified DNS A RRs can be used directly. I do get that we are kinda in a weird place right now where systemd-networkd exposes this information about the PLAT prefix existing, but doesn't (yet) take any responsibility in setting up a clat interface, and at the same time no software that I'm aware yet reads this exposed information, so software such as clatd is left to do "discovery" of the presence of a NAT64 machine by virtue of it receiving DNS64 responses. I think the logical move forward is for systemd-networkd to add proper 464xlat support and/or for more software to start reading from the info exposed by networkd.

In regards to using coredns for this test, I'm happy to use another DNS server if you think that would be best, I am just personally familiar with the coredns config and with the config being quite terse for this use case.

jmbaur: Relevant systemd issue: https://github.com/systemd/systemd/issues/23674

flokli: Ok yeah, fair points. I guess we can generally follow this as a tracking issue. Thanks for the digging!

flokli (approved):
LGTM, but I'd like to give @hax404 a chance to also review the test additions.

mweinelt:
I think it is reasonable to backport this change, as it introduces the declarative interface for functionality present in systemd 255.

hax404 (approved):
Very nice work!

flokli:
@mweinelt please do the backport manually.

PR #318609, opened 2024-06-09
godot_4: rename package to fit godot3

liperium:
## Description of changes

This is a breaking change for those using godot_4, the package will now be named in the same format as godot3, which can be usefull if we want to add mono support.

## Things done

I searched for all **godot_4** references in nixpkgs, and replaced them with **godot4**.

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/by-name/go/godot_4-export-templates/package.nix
+++ b/pkgs/by-name/go/godot_4-export-templates/package.nix
@@ -3,11 +3,11 @@
 # Export templates version should be kept in sync with Godot version.
 # https://docs.godotengine.org/en/stable/tutorials/export/exporting_projects.html#export-templates
 
-{ fetchzip, godot_4, ... }:
+{ fetchzip, godot4, ... }:
 
 fetchzip {
   pname = "export_templates";
   extension = "zip";
-  url = "https://github.com/godotengine/godot/releases/download/${godot_4.version}/Godot_v${godot_4.version}_export_templates.tpz";
+  url = "https://github.com/godotengine/godot/releases/download/${godot4.version}/Godot_v${godot4.version}_export_templates.tpz";
   hash = "sha256-eomGLH9lbZhl7VtHTWjJ5mxVt0Yg8LfnAnpqoCksPgs=";
 }
--- a/pkgs/top-level/all-packages.nix
+++ b/pkgs/top-level/all-packages.nix
@@ -8437,7 +8437,7 @@ with pkgs;
 
   gocryptfs = callPackage ../tools/filesystems/gocryptfs { };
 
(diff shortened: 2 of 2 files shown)


wegank:
I agree with this change, but you'll probably want to rename `godot_4-export-templates` to `godot4-export-templates`, too.

superherointj:
From what I understand `_` separates the package name from the version. If you are removing the `_` this could mean that `4` is part of the package name. And not version. So I'm against this change.

superherointj:
Next time you decide to change a package name, at least notify the package maintainers:

https://github.com/NixOS/nixpkgs/blob/master/pkgs/development/tools/godot/4/default.nix#L158

CC @shiryel

liperium:
@superherointj Thanks for the input, just looking forward to have it all standardized. Will leave the PR up just for discussion purposes.

superherointj:
> I agree with this change

@wegank What are your thoughts on `_` being used as separator between package name and version?

A (collective) decision has to be made on which format to use as standard. If someone knows best, let me know.

eclairevoyant:
https://github.com/NixOS/nixpkgs/blob/master/pkgs/README.md#package-naming uses underscores in all examples, and the only counterexample I've seen so far is `electron` (which probably should use underscores too...). Also, scripts like `nixpkgs-update` rely on the convention of underscores to prevent updating past the specified version in the attrname.

In any case, I don't see a reason to break that convention here.

superherointj:
> I don't see a reason to break that convention here.

Agreed. I see the need of properly documenting the convention (to avoid friction).

wegank:
My original goal was just to make the naming of `godot3` and `godot_4` consistent, so I'm also okay with renaming all `godot3` to `godot_3` in another PR.

eclairevoyant:
We also don't need to make a fully breaking change, as we can use `aliases.nix` to ease the transition

Rotaerk:
I wasn't aware of any kind of convention when I named them godot3.  I did see other packages named without an underscore after the version number though.  Ultimately it doesn't really matter to me.  I'd approve a change to rename `godot3` to `godot_3` if that's the right thing to do.

liperium:
I'll close the PR, as my changes provided do not do what my original intention was ( renaming godot_4 )
If anyone wants to do the PR for renaming godot3, go forward as my knowledge to do so is really limited as I've learned here.

PR #299622, opened 2024-03-28
cmake: add __structuredAttrs support to setup hooks

ShamrockLee:
### Summary

Provide structural attributes support to the setup hooks of CMake and lint them with ShelllCheck.

Fix #289037.

~### New interface~

~Provide interface to structurally specify CMake `-D` flags, i.e. `cmakeDFlagAttrs` and `cmakeDFlagEvalAttrs`. Not only can it be specified by phases/hooks run before `cmakeConfigurePhase`, but also as Nix Language attributes when `__structuralAttrs = true`.~

Update: The interface change is hold back here, and will be discussed in a subsequent PR.

### ~Questions~

- ~Naming.~
    - ~Any suggestions to the new names `cmakeDFlagAttrs` and `cmakeDFlagEvalAttrs`?~
- ~Boolean and non-string type management.~
    - ~Should we allow boolean and other non-string type when `cmakeDFlagAttrs` is used inside the Nix Language expression? That would be quite convenient in Nix expression level, but require an extra layer of conversion. This PR currently doesn't provide such support.~

## Things done

- Built on platform(s)
  - [X] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
- For non-Linux: Is sandboxing enabled in `nix.conf`? (See [Nix manual](https://nixos.org/manual/nix/stable/command-ref/conf-file.html))
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [X] Tested, as applicable:
  - [NixOS test(s)](https://nixos.org/manual/nixos/unstable/index.html#sec-nixos-tests) (look inside [nixos/tests](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests))
  - and/or [package tests](https://nixos.org/manual/nixpkgs/unstable/#sec-package-tests)
  - or, for functions and "core" functionality, tests in [lib/tests](https://github.com/NixOS/nixpkgs/blob/master/lib/tests) or [pkgs/test](https://github.com/NixOS/nixpkgs/blob/master/pkgs/test)
  - made sure NixOS tests are [linked](https://nixos.org/manual/nixpkgs/unstable/#ssec-nixos-tests-linking) to the relevant packages
[...]

PR #297590, opened 2024-03-21
python3Packages.torch: allow lazy loading libnvrtc

SomeoneSerge:
## Description of changes

Fixes https://github.com/NixOS/nixpkgs/issues/296179, refactors `autoAddDriverRunapth` and `autoAddCudaCompatHook`

CC @yannham 

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
- For non-Linux: Is sandboxing enabled in `nix.conf`? (See [Nix manual](https://nixos.org/manual/nix/stable/command-ref/conf-file.html))
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:
  - [NixOS test(s)](https://nixos.org/manual/nixos/unstable/index.html#sec-nixos-tests) (look inside [nixos/tests](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests))
  - and/or [package tests](https://nixos.org/manual/nixpkgs/unstable/#sec-package-tests)
  - or, for functions and "core" functionality, tests in [lib/tests](https://github.com/NixOS/nixpkgs/blob/master/lib/tests) or [pkgs/test](https://github.com/NixOS/nixpkgs/blob/master/pkgs/test)
  - made sure NixOS tests are [linked](https://nixos.org/manual/nixpkgs/unstable/#ssec-nixos-tests-linking) to the relevant packages
- [ ] Tested compilation of all packages that depend on this change using `nix-shell -p nixpkgs-review --run "nixpkgs-review rev HEAD"`. Note: all changes have to be committed, also see [nixpkgs-review usage](https://github.com/Mic92/nixpkgs-review#usage)
- [ ] Tested basic functionality of all binary files (usually in `./result/bin/`)
- [24.05 Release Notes](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2405.section.md) (or backporting [23.05](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2305.section.md) and [23.11](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2311.section.md) Release notes)
  - [ ] (Package updates) Added a release notes entry if the change is major or breaking
  - [ ] (Module updates) Added a release notes entry if the change is significant
[...]

PR #318614, opened 2024-06-09
treewide: support structuredAttrs in setup hooks

wolfgangwalther:
## Description of changes

I tried to build a package with `__structuredAttrs` which uses meson, realized that meson setup hooks don't support structuredAttrs, yet, and then... this happened.

The first 5 commits make the helpers in `stdenv` related to structuredAttrs a bit better re-usable. The remaining commits are using the "new" `concatTo` and `prependToVar` / `appendToVar` to fix many setup hooks.

The main goal here is to support structuredAttrs in those setup hooks *without* having any "if structuredAttrs then ..." conditionals except for a few in `stdenv`. I think this works pretty well.

There's a lot of changes in here, but many of them should be very easy to review. Most of them were tested by building a package using the related hooks: Once with `structuredAttrs` on master, which would fail, then with and without `structuredAttrs` on this branch - both would pass.

In any case, I pushed all those commits mostly to show how those helpers in `stdenv` would be used across the different hooks. I am happy to split up the PR into smaller pieces later and would welcome feedback especially on the first 5 commits.

The problem in 99% of those hooks is that `$xxxFlags` is used without `[@]`, which will always just return the first array item when using `structuredAttrs`. However a simple change to `$xxxFlags[@]` isn't going to fly, if whitespace should be treated properly.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/build-support/setup-hooks/autoreconf.sh
+++ b/pkgs/build-support/setup-hooks/autoreconf.sh
@@ -2,6 +2,11 @@ preConfigurePhases="${preConfigurePhases:-} autoreconfPhase"
 
 autoreconfPhase() {
     runHook preAutoreconf
-    autoreconf ${autoreconfFlags:---install --force --verbose}
+
+    local flagsArray=()
+    : "${autoreconfFlags:=--install --force --verbose}"
+    concatTo flagsArray autoreconfFlags
+
+    autoreconf "${flagsArray[@]}"
     runHook postAutoreconf
 }
--- a/pkgs/by-name/cm/cmake/setup-hook.sh
+++ b/pkgs/by-name/cm/cmake/setup-hook.sh
@@ -38,7 +38,7 @@ cmakeConfigurePhase() {
     fi
 
     if [ -z "${dontAddPrefix-}" ]; then
-        cmakeFlags="-DCMAKE_INSTALL_PREFIX=$prefix $cmakeFlags"
+        prependToVar cmakeFlags "-DCMAKE_INSTALL_PREFIX=$prefix"
     fi
 
     # We should set the proper `CMAKE_SYSTEM_NAME`.
@@ -47,21 +47,21 @@ cmakeConfigurePhase() {
     # Unfortunately cmake seems to expect absolute paths for ar, ranlib, and
     # strip. Otherwise they are taken to be relative to the source root of the
     # package being built.
-    cmakeFlags="-DCMAKE_CXX_COMPILER=$CXX $cmakeFlags"
-    cmakeFlags="-DCMAKE_C_COMPILER=$CC $cmakeFlags"
-    cmakeFlags="-DCMAKE_AR=$(command -v $AR) $cmakeFlags"
-    cmakeFlags="-DCMAKE_RANLIB=$(command -v $RANLIB) $cmakeFlags"
-    cmakeFlags="-DCMAKE_STRIP=$(command -v $STRIP) $cmakeFlags"
+    prependToVar cmakeFlags "-DCMAKE_CXX_COMPILER=$CXX"
+    prependToVar cmakeFlags "-DCMAKE_C_COMPILER=$CC"
+    prependToVar cmakeFlags "-DCMAKE_AR=$(command -v $AR)"
+    prependToVar cmakeFlags "-DCMAKE_RANLIB=$(command -v $RANLIB)"
+    prependToVar cmakeFlags "-DCMAKE_STRIP=$(command -v $STRIP)"
 
     # on macOS we want to prefer Unix-style headers to Frameworks
     # because we usually do not package the framework
-    cmakeFlags="-DCMAKE_FIND_FRAMEWORK=LAST $cmakeFlags"
+    prependToVar cmakeFlags "-DCMAKE_FIND_FRAMEWORK=LAST"
 
     # we never want to use the global macOS SDK
-    cmakeFlags="-DCMAKE_OSX_SYSROOT= $cmakeFlags"
+    prependToVar cmakeFlags "-DCMAKE_OSX_SYSROOT="
 
     # correctly detect our clang compiler
-    cmakeFlags="-DCMAKE_POLICY_DEFAULT_CMP0025=NEW $cmakeFlags"
+    prependToVar cmakeFlags "-DCMAKE_POLICY_DEFAULT_CMP0025=NEW"
 
     # This installs shared libraries with a fully-specified install
     # name. By default, cmake installs shared libraries with just the
@@ -70,7 +70,7 @@ cmakeConfigurePhase() {
     # libraries are in a system path or in the same directory as the
     # executable. This flag makes the shared library accessible from its
     # nix/store directory.
-    cmakeFlags="-DCMAKE_INSTALL_NAME_DIR=${!outputLib}/lib $cmakeFlags"
+    prependToVar cmakeFlags "-DCMAKE_INSTALL_NAME_DIR=${!outputLib}/lib"
 
     # The docdir flag needs to include PROJECT_NAME as per GNU guidelines,
     # try to extract it from CMakeLists.txt.
@@ -93,39 +93,42 @@ cmakeConfigurePhase() {
     # This ensures correct paths with multiple output derivations
     # It requires the project to use variables from GNUInstallDirs module
     # https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html
-    cmakeFlags="-DCMAKE_INSTALL_BINDIR=${!outputBin}/bin $cmakeFlags"
-    cmakeFlags="-DCMAKE_INSTALL_SBINDIR=${!outputBin}/sbin $cmakeFlags"
-    cmakeFlags="-DCMAKE_INSTALL_INCLUDEDIR=${!outputInclude}/include $cmakeFlags"
-    cmakeFlags="-DCMAKE_INSTALL_OLDINCLUDEDIR=${!outputInclude}/include $cmakeFlags"
-    cmakeFlags="-DCMAKE_INSTALL_MANDIR=${!outputMan}/share/man $cmakeFlags"
-    cmakeFlags="-DCMAKE_INSTALL_INFODIR=${!outputInfo}/share/info $cmakeFlags"
-    cmakeFlags="-DCMAKE_INSTALL_DOCDIR=${!outputDoc}/share/doc/${shareDocName} $cmakeFlags"
-    cmakeFlags="-DCMAKE_INSTALL_LIBDIR=${!outputLib}/lib $cmakeFlags"
-    cmakeFlags="-DCMAKE_INSTALL_LIBEXECDIR=${!outputLib}/libexec $cmakeFlags"
-    cmakeFlags="-DCMAKE_INSTALL_LOCALEDIR=${!outputLib}/share/locale $cmakeFlags"
+    prependToVar cmakeFlags "-DCMAKE_INSTALL_BINDIR=${!outputBin}/bin"
+    prependToVar cmakeFlags "-DCMAKE_INSTALL_SBINDIR=${!outputBin}/sbin"
+    prependToVar cmakeFlags "-DCMAKE_INSTALL_INCLUDEDIR=${!outputInclude}/include"
+    prependToVar cmakeFlags "-DCMAKE_INSTALL_OLDINCLUDEDIR=${!outputInclude}/include"
+    prependToVar cmakeFlags "-DCMAKE_INSTALL_MANDIR=${!outputMan}/share/man"
+    prependToVar cmakeFlags "-DCMAKE_INSTALL_INFODIR=${!outputInfo}/share/info"
+    prependToVar cmakeFlags "-DCMAKE_INSTALL_DOCDIR=${!outputDoc}/share/doc/${shareDocName}"
+    prependToVar cmakeFlags "-DCMAKE_INSTALL_LIBDIR=${!outputLib}/lib"
+    prependToVar cmakeFlags "-DCMAKE_INSTALL_LIBEXECDIR=${!outputLib}/libexec"
+    prependToVar cmakeFlags "-DCMAKE_INSTALL_LOCALEDIR=${!outputLib}/share/locale"
 
     # Don’t build tests when doCheck = false
     if [ -z "${doCheck-}" ]; then
-        cmakeFlags="-DBUILD_TESTING=OFF $cmakeFlags"
+        prependToVar cmakeFlags "-DBUILD_TESTING=OFF"
     fi
 
     # Always build Release, to ensure optimisation flags
-    cmakeFlags="-DCMAKE_BUILD_TYPE=${cmakeBuildType:-Release} $cmakeFlags"
+    prependToVar cmakeFlags "-DCMAKE_BUILD_TYPE=${cmakeBuildType:-Release}"
 
     # Disable user package registry to avoid potential side effects
     # and unecessary attempts to access non-existent home folder
     # https://cmake.org/cmake/help/latest/manual/cmake-packages.7.html#disabling-the-package-registry
-    cmakeFlags="-DCMAKE_EXPORT_NO_PACKAGE_REGISTRY=ON $cmakeFlags"
-    cmakeFlags="-DCMAKE_FIND_USE_PACKAGE_REGISTRY=OFF $cmakeFlags"
-    cmakeFlags="-DCMAKE_FIND_USE_SYSTEM_PACKAGE_REGISTRY=OFF $cmakeFlags"
+    prependToVar cmakeFlags "-DCMAKE_EXPORT_NO_PACKAGE_REGISTRY=ON"
+    prependToVar cmakeFlags "-DCMAKE_FIND_USE_PACKAGE_REGISTRY=OFF"
+    prependToVar cmakeFlags "-DCMAKE_FIND_USE_SYSTEM_PACKAGE_REGISTRY=OFF"
 
     if [ "${buildPhase-}" = ninjaBuildPhase ]; then
-        cmakeFlags="-GNinja $cmakeFlags"
+        prependToVar cmakeFlags "-GNinja"
     fi
 
-    echo "cmake flags: $cmakeFlags ${cmakeFlagsArray[@]}"
+    local flagsArray=()
+    concatTo flagsArray cmakeFlags cmakeFlagsArray
 
-    cmake "$cmakeDir" $cmakeFlags "${cmakeFlagsArray[@]}"
+    echoCmd 'cmake flags' "${flagsArray[@]}"
+
+    cmake "$cmakeDir" "${flagsArray[@]}"
 
     if ! [[ -v enableParallelBuilding ]]; then
         enableParallelBuilding=1
--- a/pkgs/by-name/ni/ninja/setup-hook.sh
+++ b/pkgs/by-name/ni/ninja/setup-hook.sh
@@ -1,3 +1,5 @@
+# shellcheck shell=bash
+
 ninjaBuildPhase() {
     runHook preBuild
 
@@ -9,7 +11,7 @@ ninjaBuildPhase() {
     fi
 
     local flagsArray=(
-        -j$buildCores
+        "-j$buildCores"
     )
     concatTo flagsArray ninjaFlags ninjaFlagsArray
 
@@ -24,7 +26,7 @@ ninjaCheckPhase() {
 
     if [ -z "${checkTarget:-}" ]; then
         if ninja -t query test >/dev/null 2>&1; then
-            checkTarget=test
+            checkTarget="test"
         fi
     fi
 
@@ -38,7 +40,7 @@ ninjaCheckPhase() {
         fi
 
         local flagsArray=(
-            -j$buildCores
+            "-j$buildCores"
         )
         concatTo flagsArray ninjaFlags ninjaFlagsArray checkTarget
 
@@ -61,9 +63,9 @@ ninjaInstallPhase() {
 
     # shellcheck disable=SC2086
     local flagsArray=(
-        -j$buildCores
+        "-j$buildCores"
     )
-    : ${installTargets:=install}
+    : "${installTargets:=install}"
     concatTo flagsArray ninjaFlags ninjaFlagsArray installTargets
 
     echoCmd 'install flags' "${flagsArray[@]}"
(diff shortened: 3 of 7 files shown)


emilazy:
Maybe you could keep the Meson, Ninja, CMake, and Autotools changes in this PR to handle the bulk of things and just split the long tail off?

ShamrockLee:
Regarding the CMake configuration setup hook, the original implementation prepends the default flags to the `cmkeFlags` shell variable through string concatenation, inherently prone to incorrect argument separation due to spaces, tabs, or newlines inside the definition value. The revised implementation in this PR inherits such a design pattern and its inherent weakness.

We should collect command-line arguments using Bash arrays whenever possible to fix such design flaws.

In the implementation proposed by #299622, I collect flags into the `cmakeFlagsArray` Bash array. This not only fixes the space-separation issue but also integrates all flags into one global shell variable (instead of both `cmakeFlags` and `cmakeFlagsArray`), opening the possibility to extract the flag-collection functionality into an independent Bash function to be utilized by other workflows in the future.

ShamrockLee:
> > Regarding the CMake configuration setup hook, the original implementation prepends the default flags to the `cmkeFlags` shell variable through string concatenation, inherently prone to incorrect argument separation due to spaces, tabs, or newlines inside the definition value. The revised implementation in this PR inherits such a design pattern and its inherent weakness.
> 
> This is only the case for `structuredAttrs = false`, though. Once you set `structuredAttrs = true` and use `cmakeFlags = [ ... ]` in nix, you will have an array in bash, too. And `prependToVar` already does the right thing in this case.
> 
> > We should collect command-line arguments using Bash arrays whenever possible to fix such design flaws.
> 
> Yes, but changing the `structuredAttrs = false` case is out of scope for me here - there's just way too much to fix for that everywhere.
> 
> The long term goal would be to enable structuredAttrs by default at some point - and then all the whitespace issues should already have been taken care of that way.

I agree.



> ```shell
> cmakeDefaultFlags = (
>   "-DCMAKE_CXX_COMPILER=$CXX"
>   "-DCMAKE_C_COMPILER=$CC"
>   ...
> )
> ...
> concatTo flagsArray cmakeDefaultFlags cmakeFlags cmakeFlagsArray
>     
> ```

How brilliant!

AndersonTorres:
More or less related, can we detach those hooks as we do with Waf?

pkgs/stdenv/generic/setup.sh
```diff
@@ -334,36 +334,44 @@ appendToVar() {
...
+    for name in "$@"; do
+        if type=$(declare -p "$name" 2> /dev/null); then
             local -n nameref="$name"
-            case "$name" in
-                *Array)
-                    # shellcheck disable=SC2206
-                    flagsArray+=( ${nameref+"${nameref[@]}"} ) ;;
+            case "${type#* }" in
+                -A*|-a*)
+                    targetref+=( "${nameref[@]}" ) ;;
```
SomeoneSerge, line 348: Maybe need to document and motivate what happens to associative arrays. If `nameref` is an associative array, this only adds the values? It's seemingly expected that `targetref` is not associative?

pkgs/stdenv/generic/setup.sh
```diff
@@ -334,36 +334,44 @@ appendToVar() {
...
+separateBy() {
+    local sep="$1"
+    local name="$2"
+    local type oldifs
+    if type=$(declare -p "$name" 2> /dev/null); then
+        local -n nameref="$name"
+        case "${type#* }" in
+            -A*|-a*)
+                oldifs="$IFS"; IFS="$sep"
+                echo -n "${namref[*]}"
```
SomeoneSerge, line 369: nameref

pkgs/stdenv/generic/setup.sh
```diff
@@ -334,36 +334,44 @@ appendToVar() {
...
-        done
-    fi
+        fi
+    done
+}
 
+# Separates a list of things by $1 separator.
+# For arrays each item is separated, for strings whitespace is replaced.
+# To be used for partial support of structuredAttrs.
+separateBy() {
```
SomeoneSerge, line 360: I'm not sure I understand what's happening here.
I only see one usage example, with `mesonInstallTags`.
Looking at the nixpkgs manual, the hook expects the user to specify a list of strings, and promises to concatenate the strings, joined by commas.

- The way I read the comment it does the opposite, i.e. it splits, not concats
- I'm not sure that either concatenation or the splitting are what `IFS=, echo "${nameref[*]}"` does, at least it's not what I see when I try it in repl?
- If this were to do the splitting, I don't see why you'd ever want to destructure an array in a hook?
- Also not obvious from the description that the result goes into stdout

pkgs/stdenv/generic/setup.sh
SomeoneSerge, line 1: AFAIU the main proposal here is to introduce a polymorphic primitive that can split input strings into structured arrays and keep structured inputs intact. The chosen primitive, `concatTo`, is more general and accommodates combinations of correlated  variables like `cmakeFlags` and `cmakeFlagsArray`, but, with introduction of an auxiliary variable, can effectively be used as "cast and merge".

AFAIU the way we'd use it is we'd replace conditionals such as https://github.com/NixOS/nixpkgs/blob/e20e12a9c76a453974c3e1f9434cfe14569f3f01/pkgs/build-support/setup-hooks/auto-patchelf.sh#L59 https://github.com/NixOS/nixpkgs/blob/e20e12a9c76a453974c3e1f9434cfe14569f3f01/pkgs/build-support/setup-hooks/auto-patchelf.sh#L64 ...with the simpler `concatTo runtimeDependenciesArray runtimeDependencies` (et cetera).

I recall looking for a primitive just like this recently (https://github.com/NixOS/nixpkgs/pull/297590/commits/a725317c923e2df6308f499eeb5ecbf9d35d9a25#diff-729a9eb1be81b1eb110498b0fb876255a16b50c39812fa7232a50ca53968d47eR70-R89), which to me is a validation for the approach taken in this PR

AndersonTorres:
> No idea what you mean, can you elaborate?

E.g. we declare `setupHook = ./setup-hook.sh` in cmake.
Any and every change in `setup-hook.sh` - even fixing a comment - needlessly rebuilds cmake itself.

In Waf we have `waf` and `waf.hook`. This way a rebuild of waf is avoided.

AndersonTorres:
Let's put this idea in the worklog.

pkgs/stdenv/generic/setup.sh
SomeoneSerge: `concatStringsSep` would be familiar from nix, `join` might be familiar to python users

SomeoneSerge: I'm still concerned about the behaviour with the structured attrs on

pkgs/stdenv/generic/setup.sh
```diff
@@ -334,36 +334,49 @@ appendToVar() {
...
+    for name in "$@"; do
+        if type=$(declare -p "$name" 2> /dev/null); then
             local -n nameref="$name"
-            case "$name" in
-                *Array)
-                    # shellcheck disable=SC2206
-                    flagsArray+=( ${nameref+"${nameref[@]}"} ) ;;
+            case "${type#* }" in
+                # Deals with associative arrays by only taking the values,
+                # because that's the most sensible thing to do. Letting it
```
SomeoneSerge, line 348: Well that's the thing: from a glance, I'd think the sensible thing to do is to merge the associative arrays, given how generic the name is. I see how that's completely out of scope for the use-case, which tempts me to just prohibit the `-A` branch (make it fail) until we've figured out where it's useful. Otherwise someone will depend on the "discard the keys" behaviour and we'll be stuck with it

pkgs/stdenv/generic/setup.sh
```diff
@@ -334,36 +334,49 @@ appendToVar() {
...
+                    targetref+=( ${nameref-} ) ;;
             esac
-        done
-    fi
+        fi
+    done
+}
 
+# Place separator $1 between each element of a list and return the string.
+# For arrays each item is separated, for strings whitespace is replaced.
```
SomeoneSerge, line 364: "each item is separated" - I really fail to read this one; I think/hope this just means that the array is joined on the separator, original items kept intact, like so:

```console
$ a=("foo bar" "biba boba")
$ intersperse ; a
foo bar;biba boba
```

pkgs/by-name/cm/cmake/setup-hook.sh
SomeoneSerge, line 1: > In Waf we have waf and waf.hook. This way a rebuild of waf is avoided.

Yeah I guess not urgent for this PR, but this could help with tinkering with the local overrides. E.g. if you're iterating on the hook you get different latency if it's `cudaPackages.saxpy.override { cmake = myCustomCmake; }` and `cudaPackages.saxpy.override { cmakeHook = myCustomCmakeHook; }`

But also then we probably want to replace `nativeBuildInputs = [ cmake ]` with `nativeBuildInputs = [ cmakeHook ]`, `cmakeHook` propagating `cmake`, otherwise it's just annoying to write both

AndersonTorres: > But also then we probably want to replace nativeBuildInputs = [ cmake ] with nativeBuildInputs = [ cmakeHook ], cmakeHook propagating cmake, otherwise it's just annoying to write both

Yes, this is the exact idea of `wafHook`.

pkgs/stdenv/generic/setup.sh
SomeoneSerge: > silently be caught by the subshell anyway.

Even with `set -e`? Exits for me in the interactive shell

pkgs/stdenv/generic/setup.sh
SomeoneSerge: I think we should existing phrasing as much as we can, and then clarify the behaviour for the lists:

- https://nix.dev/manual/nix/2.23/language/builtins.html#builtins-concatStringsSep
- https://docs.python.org/3/library/stdtypes.html#str.join

E.g.

> `concatStringsSep separator array`
>
> Concatenate an array of strings with a separator between each element.
> The second input may be a string in which case it is split on whitespaces:
>
> ```console
> $ flags="a b c"
> $ concatStringsSep ";" flags
> a;b;c
> ```
>
> ```console
> $ flags=("lorem ipsum" "dolor" "sit amet")
> $ concatStringsSep ";" flags
> lorem ipsum;dolor;sit amet
> ```

emilazy: The actual corresponding Haskell function is `intercalate`, though I can’t exactly recommend that name.

pkgs/stdenv/generic/setup.sh
```diff
@@ -334,36 +334,57 @@ appendToVar() {
...
+    done
+}
 
+# Concatenate a list of strings ($2) with a separator ($1) between each element.
+# The list can be an indexed array of strings or a single string. A single string
+# is split on spaces and then concatenated with the separator.
+#
+# $ flags="a b c"
+# $ concatStringsSep ";" flags
+# a;b;c
```
SomeoneSerge, line 366: Totally picking nits: I didn't think about it when writing the comment, but it's more didactic to use the same sentence as input: `"lorem ipsum dolor sit amet"` vs `("lorem ipsum" "dolor" "sit amet")`

SomeoneSerge:
I think I'm happy with the implementation, now we just need to confirm this still works.
We might want to add, sooner rather than later but maybe not yet, a manual section on "Writing __structuredAttrs-compatible hooks". E.g. `cmakeFlags+="..."` is fairly widespread and many may not be aware of `prependToVar` (I wasn't)

emilazy:
Hopefully we can turn it on by default soon enough that people have to learn! Though good documentation is, of course, a good thing.

pkgs/by-name/cm/cmake/setup-hook.sh
AndersonTorres: https://github.com/NixOS/nixpkgs/issues/331858

AndersonTorres:
> Given that `prependToVar` [is already documented](https://nixos.org/manual/nixpkgs/stable/#fun-prependToVar), but still not widely used, I wonder how much more documentation in this area will give us.

In this cases we should just use it, and especially find instances where this code could be used.

philiptaron:
I'm going to read through this PR. Requested a review from @tie too since his bash knowledge on #331560 was quite high.

pkgs/stdenv/generic/setup.sh
```diff
@@ -334,36 +334,57 @@ appendToVar() {
...
+        local -n nameref="$name"
+        case "${type#* }" in
+            -A*)
+                echo "concatStringsSep(): ERROR: trying to use concatStringsSep on an associative array." >&2
+                return 1 ;;
+            -a*)
+                local IFS="$sep"
+                echo -n "${nameref[*]}" ;;
+            *)
+                echo -n "${nameref// /${sep}}" ;;
```
SomeoneSerge, line 385: Has the always been `" "` or did this previously split using `IFS`?

SomeoneSerge (approved):

pkgs/stdenv/generic/setup.sh
```diff
@@ -313,16 +313,16 @@ appendToVar() {
...
-        if [[ "$type" =~ "declare -A" ]]; then
-            echo "appendToVar(): ERROR: trying to use appendToVar on an associative array, use variable+=([\"X\"]=\"Y\") instead." >&2
-            return 1
-        elif [[ "$type" =~ "declare -a" ]]; then
-            useArray=true
-        else
-            useArray=false
-        fi
+    if type=$(declare -p "$1" 2> /dev/null); then
+        case "${type#* }" in
```
tie, line 317: Since Bash 4.4 we can use `${var@a}` to check variable attributes (e.g. `[[ ${var@a} == *a* ]]` to check that it’s an array, or equivalent pattern in `case` statement), and use the good ol’ `[[ -z ${var+x} ]]` to check that variable is defined. See https://www.gnu.org/software/bash/manual/html_node/Shell-Parameter-Expansion.html

Also applies to other cases where we check variable attributes.

Note that currently stdenv requires at least Bash ≥4 but I don’t think we require a particular minor version. Perhaps it’s time to explicitly require at least 4.4 (released in 2016) (use `"${BASH_VERSINFO[1]}"` to check that since `BASH_VERSINFO` is actually an array, see https://www.gnu.org/software/bash/manual/html_node/Bash-Variables.html#index-BASH_005fVERSINFO). See also https://mywiki.wooledge.org/BashFAQ/061

pkgs/stdenv/generic/setup.sh
tie, line 385: I think we should be using `printf -v "$varName" …` (assigning output to the variable) and avoid returning the value via stdout to discourage the use of `$(…)` expressions with these functions (so *concatStringsSep****To*** instead of *concatStringsSep*, similar to *concatTo*).

See https://gist.github.com/tie/7e81aaa4b357c486abc197bdbe00f27c#file-tricks-bash-L7-L10 for rationale and examples.

E.g.
```console
$ bash -e -u -o pipefail -O inherit_errexit -c 'echo "$(false)oops"; echo oh no'
oops
oh no
```

philiptaron (approved):
Tested with `nix-build --expr 'with import ./. { }; let p1 = import pkgs/test/release/default.nix { }; p2 = import lib/tests/release.nix { }; in lib.filter lib.isDerivation (builtins.attrValues (p1 // p2 // tests.stdenv // tests.stdenv-inputs))'`

pkgs/stdenv/generic/setup.sh
tie: This seems to be only used in Meson’s setup hook that used to split on `" "`. Probably makes sense to use `IFS` here instead.

That said, `${sep}` should be at least quoted (see `patsub_replacement` examples in https://www.gnu.org/software/bash/manual/html_node/Shell-Parameter-Expansion.html).

tie (commented):
Some of the code can be simplified if we require Bash ≥4.4 (2016) instead of ≥4 (2009) for stdenv (see comments).

pkgs/by-name/ni/ninja/setup-hook.sh
```diff
@@ -62,25 +63,25 @@ ninjaInstallPhase() {
...
+    concatTo flagsArray ninjaFlags ninjaFlagsArray installTargets
 
     echoCmd 'install flags' "${flagsArray[@]}"
     TERM=dumb ninja "${flagsArray[@]}"
 
     runHook postInstall
 }
 
-if [ -z "${dontUseNinjaBuild-}" -a -z "${buildPhase-}" ]; then
+if [ -z "${dontUseNinjaBuild-}" ] && [ -z "${buildPhase-}" ]; then
```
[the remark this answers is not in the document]
tie: Also `[[ … && … ]]` instead of `[[ … ]] && [[ … ]]`, though both should be identical at runtime as far as I’m aware.

pkgs/test/stdenv/default.nix
```diff
@@ -96,6 +96,62 @@ let
...
+
+          string="lorem ipsum dolor sit amet"
+          stringWithSep="$(concatStringsSep ";" string)"
+          [[ "$stringWithSep" == "lorem;ipsum;dolor;sit;amet" ]] || (echo "'\$stringWithSep' was not 'lorem;ipsum;dolor;sit;amet'" && false)
+
+          array=("lorem ipsum" "dolor" "sit amet")
+          arrayWithSep="$(concatStringsSep ";" array)"
+          [[ "$arrayWithSep" == "lorem ipsum;dolor;sit amet" ]] || (echo "'\$arrayWithSep' was not 'lorem ipsum;dolor;sit amet'" && false)
+
+          touch $out
```
tie, line 152: Please add a test for `&` separator since Bash has a special treatment for this character for replacement strings in `${var//pattern/replacement}` substitutions that this function currently uses.

SomeoneSerge:
> Some of the code can be simplified if we require Bash ≥4.4 (2016) instead of ≥4 (2009) for stdenv (see comments).

I think this decision can be left for a separate PR

> I note that most places use the [[ syntax.

Is this blocking?

I assume that the following three were blocking:


- [ ] ["`${sep}` should be quoted (see patsub_replacement...)"](https://github.com/NixOS/nixpkgs/pull/318614#discussion_r1706183135)
- [ ] ["Please add a test for & separator"](https://github.com/NixOS/nixpkgs/pull/318614#discussion_r1706191919)
- [ ] ["I think we should be using printf -v "$varName" …"](https://github.com/NixOS/nixpkgs/pull/318614#discussion_r1706168114)

philiptaron:
> > I note that most places use the [[ syntax.
> Is this blocking?
No. I do Accept with Comments often - the code is fine as submitted for review, and here's a way in my subjective opinion to enhance, either now or in a future PR.

pkgs/by-name/ni/ninja/setup-hook.sh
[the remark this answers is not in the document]
AndersonTorres: Any difference?

SomeoneSerge:
The `&` works fine as a separator, tested with https://gist.github.com/SomeoneSerge/80f63f35e60fa7b5e037829b001233a3#file-test-stdenv-nix

```
❯ nix-build test-stdenv.nix -A test-concat-to -A test-concat-strings-sep
/nix/store/783gjjz2qbp4y4wrmch4a2hih1ljjih0-test-concat-to
/nix/store/abcx2fas7p5i2v5zd1rsgldblwjhqkfi-test-concat-strings-sep
```


@wolfgangwalther @tie we need a decision on `printf -v`/`concatStringsSepTo`

pkgs/build-support/setup-hooks/autoreconf.sh
SomeoneSerge, line 1: >      This test is written with an array as input, but IIUC the concern is about the string case, which uses the replacement syntax.

Ouch, yes, you're right

philiptaron:
I did a full build and test of the stdenv and everything checks out. I think the printf suggestion can be a separate PR.

I raise the motion that we merge this PR.

tie (approved):

SomeoneSerge:
We'll have to look out to ensure nobody else starts to depend on the stdout-based `concatStringsSep` in the meantime, but I think that's manageable:)

emilazy:
Really great to see this land! Looking forward to the follow‐up for the other hooks. Let’s all do our best to get structured attributes on by default as soon as we can.

trofi:
Bisect says the bfd97a691f165c2a160d5c86a7dbfd89fcbdb33c `stdenv: make _accumFlagsArray independent of structuredAttrs` changed quoting rules for some variables. For example `sudo` is now broken in `staging` as:

```
$ nix build --no-link -f. -L sudo
...
sudo> ./configure: interpreter directive changed from "#! /bin/sh" to "/nix/store/k5i08diz6c0bvklbzdp35rpprrnml9bh-bash-5.2p32/bin/sh"                               
sudo> configure flags: --disable-static --prefix=/nix/store/8kqimiqs90p0vlvm43ia2v46wc040gbw-sudo-1.9.15p5 --with-env-editor --with-editor=/run/current-system/sw/bin
/nano --with-rundir=/run/sudo --with-vardir=/var/db/sudo --with-logpath=/var/log/sudo.log --with-iologdir=/var/log/sudo-io --with-sendmail=/run/wrappers/bin/sendmail
 --enable-tmpfiles.d=no password for %p:                                                                                                                             
sudo> configure: WARNING: you should use --build, --host, --target                                                                                                   
sudo> configure: WARNING: you should use --build, --host, --target                                                                                                   
sudo> configure: WARNING: you should use --build, --host, --target                                                                                                   
sudo> configure: WARNING: invalid host type: %p:        
...
sudo> checking for password-ar... ar
sudo> checking for password-ranlib... ranlib
sudo> checking build system type... Invalid configuration 'password': machine 'password-unknown' not recognized
sudo> configure: error: /nix/store/k5i08diz6c0bvklbzdp35rpprrnml9bh-bash-5.2p32/bin/bash ./scripts/config.sub password failed
```

This happenens because `sudo` has this snippet:

```nix
  configureFlagsArray = [
    "--with-passprompt=[sudo] password for %p: " # intentional trailing space
  ];
```

It used to be passed as a single string, but now it gets expanded to an argument list.

emilazy:
I guess we just need to restore the special handling of `*Array`? I’ll try to take a look in the next day or so if nobody beats me to it.

SomeoneSerge:
> It used to be passed as a single string, but now it gets expanded to an argument list.

Most peculiar, I'm surprised this used to work because this does not seem to be a `__structuredAttrs` derivation yet it sets `*Array` as a nix variable...

emilazy:
I guess just toggling on `__structuredAttrs` would be nice, but this is going to break out‐of‐tree packages. We probably need to fix it.

SomeoneSerge:
> I guess just toggling on __structuredAttrs would be nice, but this is going to break out‐of‐tree packages. We probably need to fix it.

👍🏻 Two separate matters: make sudo structuredAttrs, add a warning/error out-of-tree users when `[[ name = *"Array" ]]`, possibly recovering the old behavior

philiptaron:
With #334973 merged, I think we're safe for staging-next. Anyone disagree?

PR #318630, opened 2024-06-09
nixos/ups: run upsmon as non-root user "nutmon" by default 

### LINKED ISSUE #318582: power.ups.upsmon.settings doesn't allow RUN_AS_USER to be set
### Describe the bug
The UPS module allows additional settings for the upsmon.conf file to be specified. `RUN_AS_USER` is a setting defined with a default of "root", and if anything else is specified an error is thrown at build time.

### Steps To Reproduce
1. Add the following configuration to a NixOS configuration
  - `power.ups.upsmon.settings.RUN_AS_USER = "otheruser";` 
2. Run `sudo nixos-rebuild switch`

### Expected behavior
The configuration value of `RUN_AS_USER otheruser` in /etc/nut/upsmon.conf.

### Additional context
It looks like the default `RUN_AS_USER` value is missing `mkdefault`. https://github.com/NixOS/nixpkgs/blob/9b5328b7f761a7bbdc0e332ac4cf076a3eedb89b/nixos/modules/services/monitoring/ups.nix#L331

### Notify maintainers

@Majiir 

### Metadata

```console
 - system: `"x86_64-linux"`
 - host os: `Linux 6.1.79, NixOS, 23.11 (Tapir), 23.11.7503.a2e1d0414259`
 - multi-user?: `yes`
 - sandbox: `yes`
 - channels(root): `"nixos-24.05"`
```

---

yodaldevoid:
## Description of changes

Fixes #318582

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
- [ ] Tested, as applicable:

---

### DIFF
--- a/nixos/modules/services/monitoring/ups.nix
+++ b/nixos/modules/services/monitoring/ups.nix
@@ -74,9 +74,9 @@ let
   };
 
   installSecrets =
-    source: target: secrets:
+    source: target: owner: secrets:
     pkgs.writeShellScript "installSecrets.sh" ''
-      install -m0600 -D ${source} "${target}"
+      install -m0600 -o${owner} -D ${source} "${target}"
       ${lib.concatLines (
         lib.forEach secrets (name: ''
           ${pkgs.replace-secret}/bin/replace-secret \
@@ -327,6 +327,23 @@ let
         description = "Whether to enable `upsmon`.";
       };
 
+      user = lib.mkOption {
+        type = lib.types.str;
+        default = "nutmon";
+        description = ''
+          User to run `upsmon` as. `upsmon.conf` will have its owner set to this
+          user. If not specified, a default user will be created.
+        '';
+      };
+      group = lib.mkOption {
+        type = lib.types.str;
+        default = "nutmon";
+        description = ''
+          Group for the default `nutmon` user. If the default user is created
+          and this is not specified, a default group will be created.
+        '';
+      };
+
       monitor = lib.mkOption {
         type = with lib.types; attrsOf (submodule monitorOptions);
         default = { };
@@ -344,7 +361,6 @@ let
             MONITOR = <generated from config.power.ups.upsmon.monitor>
             NOTIFYCMD = "''${pkgs.nut}/bin/upssched";
             POWERDOWNFLAG = "/run/killpower";
-            RUN_AS_USER = "root";
             SHUTDOWNCMD = "''${pkgs.systemd}/bin/shutdown now";
           }
         '';
@@ -382,7 +398,6 @@ let
         );
         NOTIFYCMD = lib.mkDefault "${pkgs.nut}/bin/upssched";
         POWERDOWNFLAG = lib.mkDefault "/run/killpower";
-        RUN_AS_USER = "root"; # TODO: replace 'root' by another username.
         SHUTDOWNCMD = lib.mkDefault "${pkgs.systemd}/bin/shutdown now";
       };
     };
@@ -581,7 +596,7 @@ in
     systemd.services.upsmon =
       let
         secrets = lib.mapAttrsToList (name: monitor: "upsmon_password_${name}") cfg.upsmon.monitor;
-        createUpsmonConf = installSecrets upsmonConf "/run/nut/upsmon.conf" secrets;
+        createUpsmonConf = installSecrets upsmonConf "/run/nut/upsmon.conf" cfg.upsmon.user secrets;
       in
       {
         enable = cfg.upsmon.enable;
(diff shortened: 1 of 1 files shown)


bjornfor:
@yodaldevoid: Do you have a working config with non-root user?

yodaldevoid:
> @yodaldevoid: Do you have a working config with non-root user?

Yes, you can use `lib.mkForce` to force the setting. e.g.
```
power.ups.upsmon.settings.RUN_AS_USER = lib.mkForce config.users.users.nutmonitor.name;
```

bjornfor:
> > @yodaldevoid: Do you have a working config with non-root user?
> 
> Yes, on stable you can use `lib.mkForce` to force the setting. e.g.
> 
> ```
> power.ups.upsmon.settings.RUN_AS_USER = lib.mkForce config.users.users.nutmonitor.name;
> ```

So there's no permission errors resulting from that? I assumed it was running as root for a reason (permissions). If it's that easy to run as non-root, why don't we create that "nutmonitor" user in NixOS by default and have everyone use that? What's the use case for users to be able to run NUT as arbitrary user?

yodaldevoid:
> So there's no permission errors resulting from that?
Nope, no issues. It even has access to a root-only password file at startup.

> I assumed it was running as root for a reason (permissions). If it's that easy to run as non-root, why don't we create that "nutmonitor" user in NixOS by default and have everyone use that? What's the use case for users to be able to run NUT as arbitrary user?
I can only make conjectures at this point, but I figure it was the easier thing at the time. If a user was generated and used, I know I would want a way to configure that (at least to turn it off to replace with my custom configured user). At that point, with how easy it is to generate users, why no just have the user do it themself.

bjornfor:
> If a user was generated and used, I know I would want a way to configure that (at least to turn it off to replace with my custom configured user). At that point, with how easy it is to generate users, why no just have the user do it themself.

That doesn't really explain a use case except for "because we can". What exactly do you want to customize? User name? Additional groups? (That can be done by extending/overriding a base user provided by NixOS.)

After reading https://networkupstools.org/docs/man/upsmon.conf.html (the RUN_AS_USER part), I think this patch is going in the wrong direction.

The way I think it should work is that NixOS generates a user (e.g. "nutmon" as suggested by upstream) and ensures the config file is readable for that user, so that reloading works. (This is actually how it works today, except that the generated user is "root".) Having to use lib.mkForce to change the user from under NixOS is a good thing; it implies a bit of "I know what I'm doing and accept the risks". Because this patch makes it easier to break the ownership of the config file.

bjornfor (dismissed):
I think we should fix the TODO in NixOS (for all people) instead of making it a bit easier to fix the issue out-of-tree only for expert users, see https://github.com/NixOS/nixpkgs/pull/318630#issuecomment-2292869018. (Expert users can just `lib.mkForce` the setting anyway.)

bjornfor:
I realized that we can turn this around; we can make RUN_AS_USER easily configurable, but then the permissions on the config files must automatically follow. I'm not a fan of making them easily diverge.

yodaldevoid:
I've finally got around to making this a more proper solution. I've followed the pattern I've seen other packages use; a default user is created if the user does not specify a different one. The same goes for this user's group.

bjornfor (commented):
Looks good. Didn't test. Commit message needs update ("run as non-root user 'nutmon'", or something?).

yodaldevoid:
I've updated the commit message with rationale for this change.

bjornfor:
Does shutdown work with this non-root?

bjornfor:
Thanks!

bjornfor (approved):

nixos/modules/services/monitoring/ups.nix
```diff
@@ -382,7 +399,6 @@ let
         );
         NOTIFYCMD = lib.mkDefault "${pkgs.nut}/bin/upssched";
         POWERDOWNFLAG = lib.mkDefault "/run/killpower";
-        RUN_AS_USER = "root"; # TODO: replace 'root' by another username.
```
bjornfor, line 385: In the previous hunk, RUN_AS_USER is set to nutmon. Why isn't this set here too? Feels like doc and impl. disagrees.

yodaldevoid: Ah, good catch. I missed that one when switching to the `-u` option.

bjornfor:
@yodaldevoid: I had plans to test this before merge, but it keep getting pushed back. So I think what we do now is that you say to me whether you think I should merge :-)

yodaldevoid:
@bjornfor I've been running this patch rebased on release-24.11 for a while on my personal machine with no issue, so I'd say merge it.

bjornfor:
@yodaldevoid: Thanks!

PR #318647, opened 2024-06-09
nexusmods-app: 0.4.1 -> 0.5.2

### LINKED ISSUE #317852: Update request: nexusmods-app 0.4.1 → 0.5.2
- Package name: nexusmods-app
- Latest released version: 0.5.2

- Current version on the unstable channel: N/A
- Current version on the stable/release channel: N/A

Make sure to follow [upstream package maintainers documentation](https://github.com/Nexus-Mods/NexusMods.App/blob/main/docs/Contributing.md#for-package-maintainers).

**Notify maintainers**

-----

---

l0b0:
## Description of changes

[Release notes](https://github.com/Nexus-Mods/NexusMods.App/releases/tag/v0.5.2)

[Commits](https://github.com/Nexus-Mods/NexusMods.App/compare/v0.4.1...v0.5.2)

- Don't include upstream's 7zz binary
- Tell the app it is a distro package
- Document more of the build configuration
- Format using nixfmt-rfc-style

Closes #317852.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [x] Tested, as applicable:

---


pkgs/by-name/ne/nexusmods-app/package.nix
```diff
@@ -51,9 +51,12 @@ buildDotnetModule rec {
     ln --force --symbolic "${lib.getExe _7zzWithOptionalUnfreeRarSupport}" src/ArchiveManagement/NexusMods.FileExtractor/runtimes/linux-x64/native/7zz
   '';
 
+  NEXUSMODS_APP_USE_SYSTEM_EXTRACTOR = 1; # Use nixpkgs 7zz executable
+
   makeWrapperArgs = [
     "--prefix PATH : ${lib.makeBinPath [desktop-file-utils]}"
     "--set APPIMAGE $out/bin/${meta.mainProgram}" # Make associating with nxm links work on Linux
+    "--set INSTALLATION_METHOD_PACKAGE_MANAGER 1" # Disable integrated updater
```
MattSturgeon, line 59: Technically, there is no integrated updater. Only an update _checker/notifier_.

This env doesn't disable anything, rather it changes the user-facing instructions to something along the lines of _"update available. Use your system package manager to install it."_ (I haven't looked at the actual wording, this is based off of the docs & and discord conversations).

```suggestion
    "--set INSTALLATION_METHOD_PACKAGE_MANAGER 1" # Tell the app it's a distro package
```

pkgs/by-name/ne/nexusmods-app/package.nix
```diff
@@ -51,9 +51,12 @@ buildDotnetModule rec {
     ln --force --symbolic "${lib.getExe _7zzWithOptionalUnfreeRarSupport}" src/ArchiveManagement/NexusMods.FileExtractor/runtimes/linux-x64/native/7zz
   '';
 
+  NEXUSMODS_APP_USE_SYSTEM_EXTRACTOR = 1; # Use nixpkgs 7zz executable
```
MattSturgeon, line 54: Does passing arbitrary attrs to `buildDotnetModule` put them in the env or should this line be moved elsewhere?

pkgs/by-name/ne/nexusmods-app/package.nix
```diff
@@ -51,9 +51,12 @@ buildDotnetModule rec {
...
+
   makeWrapperArgs = [
     "--prefix PATH : ${lib.makeBinPath [desktop-file-utils]}"
     "--set APPIMAGE $out/bin/${meta.mainProgram}" # Make associating with nxm links work on Linux
+    "--set INSTALLATION_METHOD_PACKAGE_MANAGER 1" # Tell the app it's a distro package
```
MattSturgeon, line 59: [Upstream docs](https://github.com/Nexus-Mods/NexusMods.App/blob/main/docs/Contributing.md#for-package-maintainers) say to set the documented vars "when publishing".

Have you checked if they mean at build time or at runtime?

(I assume wrapper args will affect the app at runtime)

MattSturgeon: @erri120 would it be possible to clarify your publishing docs, to specify if these variables apply when building or when running the app?

pkgs/by-name/ne/nexusmods-app/package.nix
```diff
@@ -51,9 +51,12 @@ buildDotnetModule rec {
     ln --force --symbolic "${lib.getExe _7zzWithOptionalUnfreeRarSupport}" src/ArchiveManagement/NexusMods.FileExtractor/runtimes/linux-x64/native/7zz
```
MattSturgeon, line 51: Is this still needed? If so does it still need `--force`?

MattSturgeon: And can it use `lib.getExe`?

MattSturgeon: I think having the exe on the PATH should be enough, so I guess just list the package in `propagatedBuildInputs`?

pkgs/by-name/ne/nexusmods-app/package.nix
l0b0: According to [this](https://stackoverflow.com/a/27719330/96588), yes.

pkgs/by-name/ne/nexusmods-app/package.nix
l0b0: I would assume this means at runtime, since otherwise the variable would have to be set in some file which is sourced at runtime, or compiled into the program.

pkgs/by-name/ne/nexusmods-app/package.nix
l0b0: Good catch! I don't think it's even needed at build time, so I've added it to `runtimeDeps`.

pkgs/by-name/ne/nexusmods-app/package.nix
MattSturgeon: > or compiled into the program.

I think that's the case. Looking at https://github.com/Nexus-Mods/NexusMods.App/commit/95782f541987b7b6f5fe5bbf09b6c5b57412a944, these are build props, not than env vars.

pkgs/by-name/ne/nexusmods-app/package.nix
```diff
@@ -44,23 +39,23 @@ buildDotnetModule rec {
...
     "--set APPIMAGE $out/bin/${meta.mainProgram}" # Make associating with nxm links work on Linux
+    "--set INSTALLATION_METHOD_PACKAGE_MANAGER 1" # Tell the app it's a distro package
   ];
 
   runtimeDeps = [
     fontconfig
     libICE
     libSM
     libX11
+    (lib.getExe (_7zz.override { inherit enableUnfree; }))
```
MattSturgeon, line 58: Correct me if I'm wrong, but I think you just pass the package?

```suggestion
    (_7zz.override { inherit enableUnfree; })
```

pkgs/by-name/ne/nexusmods-app/package.nix
l0b0: [You're right](https://github.com/Nexus-Mods/NexusMods.App/commit/95782f541987b7b6f5fe5bbf09b6c5b57412a944#diff-4fb37a10b38749d0e7315711dbe7a65f3d06e337055721656a8df6bd559ebfedR12). Now to figure out how to set build properties in `buildDotnetModule`. I guess that's in [`dotnetBuildFlags`](https://github.com/NixOS/nixpkgs/blob/cd7ec2419a9a98e9119d64f5b0cbaa069e53c298/doc/languages-frameworks/dotnet.section.md#packaging-a-dotnet-application-packaging-a-dotnet-application).

MattSturgeon: Although `INSTALLATION_METHOD_PACKAGE_MANAGER` is not in 0.5.1, that'll need the upcoming release.

- `NEXUSMODS_APP_USE_SYSTEM_EXTRACTOR` [exists in 0.5.1](https://github.com/Nexus-Mods/NexusMods.App/blob/v0.5.1/Directory.Build.props#L17C17-L17C51)
- `INSTALLATION_METHOD_PACKAGE_MANAGER` was [named `INSTALLATION_METHOD_FLATPAK`](https://github.com/Nexus-Mods/NexusMods.App/blob/v0.5.1/Directory.Build.props#L27C17-L27C44) at the time

MattSturgeon: > [You're right](https://github.com/Nexus-Mods/NexusMods.App/commit/95782f541987b7b6f5fe5bbf09b6c5b57412a944#diff-4fb37a10b38749d0e7315711dbe7a65f3d06e337055721656a8df6bd559ebfedR12). Now to figure out how to set build properties in `buildDotnetModule`. I guess that's in [`dotnetBuildFlags`](https://github.com/NixOS/nixpkgs/blob/cd7ec2419a9a98e9119d64f5b0cbaa069e53c298/doc/languages-frameworks/dotnet.section.md#packaging-a-dotnet-application-packaging-a-dotnet-application).

https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet-build

Looks like you want `-p` or `--property` in `dotnetBuildFlags`

l0b0: WTH? That is not described in `nix-shell --packages dotnet-sdk_8 --run 'dotnet build --help'`. Ah well.

l0b0: I've gone with the relevant property for now; we can update it when upgrading.

pkgs/by-name/ne/nexusmods-app/package.nix
```diff
@@ -44,12 +39,13 @@ buildDotnetModule rec {
...
+      --replace-fail '</PropertyGroup>' '<ErrorOnDuplicatePublishOutputFiles>false</ErrorOnDuplicatePublishOutputFiles></PropertyGroup>'
   '';
 
-  postPatch = ''
-    ln --force --symbolic "${lib.getExe _7zzWithOptionalUnfreeRarSupport}" src/ArchiveManagement/NexusMods.FileExtractor/runtimes/linux-x64/native/7zz
-  '';
+  dotnetBuildFlags = [
+    "--property:INSTALLATION_METHOD_FLATPAK=1"
+    "--property:NEXUSMODS_APP_USE_SYSTEM_EXTRACTOR=1"
+  ];
```
MattSturgeon, line 48: You're welcome to ignore this suggestion if you think it's overkill

```suggestion
  dotnetBuildFlags = [
    # Tell the app it is a distro package, e.g. affects wording in update prompts.
    # TODO: renamed to INSTALLATION_METHOD_PACKAGE_MANAGER after 0.5.1
    "--property:INSTALLATION_METHOD_FLATPAK=1"

    # Don't include upstream's 7zz binary; we want the nixpkgs version.
    "--property:NEXUSMODS_APP_USE_SYSTEM_EXTRACTOR=1"
  ];
```

MattSturgeon:
I cant build (`error : No space left on device [/build/source/NexusMods.App.sln]`), but that's probably because I have `/` mounted as tmpfs...

MattSturgeon (approved):
Looks good.

While I'm not experienced in dotnet, I am excited about the nexusmods-app. Therefore I'm tempted to add myself as a maintainer.

Should I open a PR to do that, or is normal procedure to wait until actually contributing something to the package?

l0b0:
> Looks good.

Thank you for the help!
 
> While I'm not experienced in dotnet, I am excited about the nexusmods-app. Therefore I'm tempted to add myself as a maintainer.

That would be great! I was about to ask :grin: 

> Should I open a PR to do that, or is normal procedure to wait until actually contributing something to the package?

You have contributed to this PR, but it's probably easiest to [add yourself](https://github.com/NixOS/nixpkgs/blob/master/CONTRIBUTING.md#commit-conventions) in a separate PR.

qubitnano:
This builds but I get a crash when running:

```
[nix-shell:~/.cache/nixpkgs-review/pr-318647-3/results/nexusmods-app/bin]$ ./NexusMods.App 
Unhandled exception. System.IO.FileNotFoundException: Could not load file or assembly 'Microsoft.Extensions.DependencyInjection, Version=8.0.0.0, Culture=neutral, PublicKeyToken=adb9793829ddae60'. The system cannot find the file specified.

File name: 'Microsoft.Extensions.DependencyInjection, Version=8.0.0.0, Culture=neutral, PublicKeyToken=adb9793829ddae60'
   at Microsoft.Extensions.Hosting.HostBuilder..ctor()
   at NexusMods.App.Program.BuildSettingsHost() in /build/source/src/NexusMods.App/Program.cs:line 171
   at NexusMods.App.Program.Main(String[] args) in /build/source/src/NexusMods.App/Program.cs:line 39
Aborted (core dumped)
```

pkgs/by-name/ne/nexusmods-app/package.nix
```diff
@@ -44,44 +38,53 @@ buildDotnetModule rec {
...
+    # Tell the app it is a distro package; affects wording in update prompts
+    "--property:INSTALLATION_METHOD_FLATPAK=1"
+
+    # Don't include upstream's 7zz binary; we use the nixpkgs version
+    "--property:NEXUSMODS_APP_USE_SYSTEM_EXTRACTOR=1"
+  ];
 
   makeWrapperArgs = [
-    "--prefix PATH : ${lib.makeBinPath [desktop-file-utils]}"
+    "--prefix PATH : ${lib.makeBinPath [ desktop-file-utils ]}"
```
MattSturgeon, line 53: Would it be enough to just list `desktop-file-utils` in `runtimeDeps`, instead of explicitly adding them to the PATH here?

Runtime dpes all end up on the PATH anyway, right?

l0b0: > Would it be enough to just list `desktop-file-utils` in `runtimeDeps`, instead of explicitly adding them to the PATH here?
> 
> Runtime dpes all end up on the PATH anyway, right?

I think so? I'll give it a try.

l0b0:
> This builds but I get a crash when running:

Same, and I have no idea what that error means. @erri120?

pkgs/by-name/ne/nexusmods-app/package.nix
erri120: > when publishing

`dotnet build` is a generic build command, while `dotnet publish` is a command made specifically for publishing executables. We also use https://github.com/kuiperzone/PupNet-Deploy to create AppImages.

You should consider using `dotnet publish` instead of `dotnet build` if possible, as it allows you to do stuff like AOT compiling, single-file publishing, and stripping unnecessary parts.

erri120:
> > This builds but I get a crash when running:
> 
> Same, and I have no idea what that error means. @erri120?

If you refer to https://github.com/NixOS/nixpkgs/pull/318647#issuecomment-2157147560 then that means the assembly `Microsoft.Extensions.DependencyInjection.dll` is either missing or the assembly loader couldn't find it. This can happen with funky build outputs and broken runtime paths. I'd suggest checking the output directory and see if `NexusMods.App.deps.json` exists and is valid.

pkgs/by-name/ne/nexusmods-app/package.nix
```diff
@@ -44,44 +38,53 @@ buildDotnetModule rec {
...
-        "Category!=Disabled"
-        "FlakeyTest!=True"
-        "RequiresNetworking!=True"
-        "FullyQualifiedName!=NexusMods.UI.Tests.ImageCacheTests.Test_LoadAndCache_RemoteImage"
-        "FullyQualifiedName!=NexusMods.UI.Tests.ImageCacheTests.Test_LoadAndCache_ImageStoredFile"
-      ] ++ lib.optionals (! enableUnfree) [
-        "FullyQualifiedName!=NexusMods.Games.FOMOD.Tests.FomodXmlInstallerTests.InstallsFilesSimple_UsingRar"
-       ])))
+      (lib.strings.concatStrings (
+        lib.strings.intersperse "&" (
```
MattSturgeon, line 75: I `concatStringsSep "&"` does the same thing in one step.

Might make the diff smaller if it means the indentation doesn't change?

pkgs/by-name/ne/nexusmods-app/package.nix
```diff
@@ -44,44 +38,53 @@ buildDotnetModule rec {
...
     fontconfig
     libICE
     libSM
     libX11
   ];
 
-  executables = [
-    nexusmods-app.meta.mainProgram
-  ];
+  executables = [ nexusmods-app.meta.mainProgram ];
```
MattSturgeon, line 66:
```suggestion
  executables = [ meta.mainProgram ];
```

Since we're using a recursive attrset anyway

pkgs/by-name/ne/nexusmods-app/package.nix
MattSturgeon: Did this not work in the end?

MattSturgeon:
> This builds but I get a crash when running:
> 
> ```
> [nix-shell:~/.cache/nixpkgs-review/pr-318647-3/results/nexusmods-app/bin]$ ./NexusMods.App 
> Unhandled exception. System.IO.FileNotFoundException: Could not load file or assembly 'Microsoft.Extensions.DependencyInjection, Version=8.0.0.0, Culture=neutral, PublicKeyToken=adb9793829ddae60'. The system cannot find the file specified.
> 
> File name: 'Microsoft.Extensions.DependencyInjection, Version=8.0.0.0, Culture=neutral, PublicKeyToken=adb9793829ddae60'
>    at Microsoft.Extensions.Hosting.HostBuilder..ctor()
>    at NexusMods.App.Program.BuildSettingsHost() in /build/source/src/NexusMods.App/Program.cs:line 171
>    at NexusMods.App.Program.Main(String[] args) in /build/source/src/NexusMods.App/Program.cs:line 39
> Aborted (core dumped)
> ```

I get the same error with `nix run .#nexusmods-app`


> If you refer to [#318647 (comment)](https://github.com/NixOS/nixpkgs/pull/318647#issuecomment-2157147560) then that means the assembly `Microsoft.Extensions.DependencyInjection.dll` is either missing or the assembly loader couldn't find it.

Is that DLL part of `dotnet-runtime-8.0.6` or are we missing a dependency?

If we're not missing anything, it should be on the `LD_LIBRARY_PATH` somewhere... The wrapper script (installed at `bin/NexusMods.App`) looks like this in my build:

<details><summary>Details</summary>
<p>

```shell
#! /nix/store/306znyj77fv49kwnkpxmb0j2znqpa8bj-bash-5.2p26/bin/bash -e
LD_LIBRARY_PATH=${LD_LIBRARY_PATH:+':'$LD_LIBRARY_PATH':'}
if [[ $LD_LIBRARY_PATH != *':''/nix/store/azbphnp68h4fkm0fsbird88bn76ggjsh-fontconfig-2.15.0-lib/lib'':'* ]]; then
    LD_LIBRARY_PATH=$LD_LIBRARY_PATH'/nix/store/azbphnp68h4fkm0fsbird88bn76ggjsh-fontconfig-2.15.0-lib/lib'
fi
LD_LIBRARY_PATH=${LD_LIBRARY_PATH#':'}
LD_LIBRARY_PATH=${LD_LIBRARY_PATH%':'}
export LD_LIBRARY_PATH
LD_LIBRARY_PATH=${LD_LIBRARY_PATH:+':'$LD_LIBRARY_PATH':'}
if [[ $LD_LIBRARY_PATH != *':''/nix/store/alryr2fk5z2n19s72q1yf0r71ych8jml-libICE-1.1.1/lib'':'* ]]; then
    LD_LIBRARY_PATH=$LD_LIBRARY_PATH'/nix/store/alryr2fk5z2n19s72q1yf0r71ych8jml-libICE-1.1.1/lib'
fi
LD_LIBRARY_PATH=${LD_LIBRARY_PATH#':'}
LD_LIBRARY_PATH=${LD_LIBRARY_PATH%':'}
export LD_LIBRARY_PATH
LD_LIBRARY_PATH=${LD_LIBRARY_PATH:+':'$LD_LIBRARY_PATH':'}
if [[ $LD_LIBRARY_PATH != *':''/nix/store/h8j225cl14p2yz55xa76rasfins4540w-libSM-1.2.4/lib'':'* ]]; then
    LD_LIBRARY_PATH=$LD_LIBRARY_PATH'/nix/store/h8j225cl14p2yz55xa76rasfins4540w-libSM-1.2.4/lib'
fi
LD_LIBRARY_PATH=${LD_LIBRARY_PATH#':'}
LD_LIBRARY_PATH=${LD_LIBRARY_PATH%':'}
export LD_LIBRARY_PATH
LD_LIBRARY_PATH=${LD_LIBRARY_PATH:+':'$LD_LIBRARY_PATH':'}
if [[ $LD_LIBRARY_PATH != *':''/nix/store/x9fw7rbdb34gq0f8q750kw344lbv9nk1-libX11-1.8.9/lib'':'* ]]; then
    LD_LIBRARY_PATH=$LD_LIBRARY_PATH'/nix/store/x9fw7rbdb34gq0f8q750kw344lbv9nk1-libX11-1.8.9/lib'
fi
LD_LIBRARY_PATH=${LD_LIBRARY_PATH#':'}
LD_LIBRARY_PATH=${LD_LIBRARY_PATH%':'}
export LD_LIBRARY_PATH
export DOTNET_ROOT='/nix/store/wzpmhmjxv1rvh8z8zlliga2yql1y5ax8-dotnet-runtime-8.0.6'
PATH=${PATH:+':'$PATH':'}
PATH=${PATH/':''/nix/store/wzpmhmjxv1rvh8z8zlliga2yql1y5ax8-dotnet-runtime-8.0.6/bin'':'/':'}
PATH='/nix/store/wzpmhmjxv1rvh8z8zlliga2yql1y5ax8-dotnet-runtime-8.0.6/bin'$PATH
PATH=${PATH#':'}
PATH=${PATH%':'}
export PATH
PATH=${PATH:+':'$PATH':'}
PATH=${PATH/':''/nix/store/rmk31w7i6ijps82v2zdy6jp2hpqz65x4-desktop-file-utils-0.27/bin'':'/':'}
PATH='/nix/store/rmk31w7i6ijps82v2zdy6jp2hpqz65x4-desktop-file-utils-0.27/bin'$PATH
PATH=${PATH#':'}
PATH=${PATH%':'}
export PATH
export APPIMAGE='/nix/store/b711v6aacyapswz16dn1bq0pq168dkg9-nexusmods-app-0.5.1/bin/NexusMods.App'
LD_LIBRARY_PATH=${LD_LIBRARY_PATH:+':'$LD_LIBRARY_PATH':'}
LD_LIBRARY_PATH=${LD_LIBRARY_PATH/':''/nix/store/lvq4hib247ayxib31hp7gryx444m1l3a-icu4c-73.2/lib'':'/':'}
LD_LIBRARY_PATH='/nix/store/lvq4hib247ayxib31hp7gryx444m1l3a-icu4c-73.2/lib'$LD_LIBRARY_PATH
LD_LIBRARY_PATH=${LD_LIBRARY_PATH#':'}
LD_LIBRARY_PATH=${LD_LIBRARY_PATH%':'}
export LD_LIBRARY_PATH
exec "/nix/store/b711v6aacyapswz16dn1bq0pq168dkg9-nexusmods-app-0.5.1/lib/nexusmods-app/NexusMods.App"  "$@"
```

</p>
</details> 

> This can happen with funky build outputs and broken runtime paths. I'd suggest checking the output directory and see if `NexusMods.App.deps.json` exists and is valid.

I don't think we propagate any data files (other than the `.desktop` files). I'm not familiar with dotnet, but I think nixpkgs' dotnet toolchain is setup to not need manifests like that at runtime - everything should get baked into the wrapper script.

erri120:
> Is that DLL part of dotnet-runtime-8.0.6 or are we missing a dependency?

`Microsoft.Extensions.DependencyInjection` is just a normal NuGet package published by Microsoft.

erri120:
I'm guessing you're running into a dependency issue. We don't directly reference `Microsoft.Extensions.DependencyInjection`. Instead we reference both `Microsoft.Extensions.DependencyInjection.Abstraction` and `Microsoft.Extensions.Hosting`. As such, `Microsoft.Extensions.DependencyInjection` is a transient dependency, and with enough version fuckery, this might be the problem.

l0b0:
> I'm guessing you're running into a dependency issue. We don't directly reference `Microsoft.Extensions.DependencyInjection`. Instead we reference both `Microsoft.Extensions.DependencyInjection.Abstraction` and `Microsoft.Extensions.Hosting`. As such, `Microsoft.Extensions.DependencyInjection` is a transient dependency, and with enough version fuckery, this might be the problem.

It does look like [`Microsoft.Extensions.DependencyInjection` is included in the dependency list](https://github.com/l0b0/nixpkgs/blob/nexusmods-app-update/pkgs/by-name/ne/nexusmods-app/deps.nix#L226), though.

MattSturgeon:
> It does look like [`Microsoft.Extensions.DependencyInjection` is included in the dependency list](https://github.com/l0b0/nixpkgs/blob/nexusmods-app-update/pkgs/by-name/ne/nexusmods-app/deps.nix?rgh-link-date=2024-06-12T20%3A55%3A58Z#L226), though.

A transient dependency is a dependency _of another_ dependency, so I guess that's why it's missing?

l0b0:
[0.5.2 is available](https://github.com/Nexus-Mods/NexusMods.App/releases/tag/v0.5.2), but upgrade is blocked by #320204 not being in nixos-unstable yet.

MattSturgeon:
> upgrade is blocked by #320204 [not being in nixos-unstable](https://nixpk.gs/pr-tracker.html?pr=320204) yet.

Why is that a blocker, considering this PR is targeting master (not unstable)?

l0b0:
> > upgrade is blocked by #320204 [not being in nixos-unstable](https://nixpk.gs/pr-tracker.html?pr=320204) yet.
> 
> Why is that a blocker, considering this PR is targeting master (not unstable)?

It's a blocker for my local system, not for this branch. But I was able to work around it with [this patch](https://github.com/NixOS/nixpkgs/issues/315574#issuecomment-2137120500), so I'm just building the latest version to see if I can upload a new one.

erri120:
I've update the [docs](https://nexus-mods.github.io/NexusMods.App/developers/Contributing/#for-package-maintainers) with more details on how to publish. You can try using the same arguments that we're using.

MattSturgeon:
Looking at nixpkgs manual https://nixos.org/manual/nixpkgs/stable/#packaging-a-dotnet-application

I can't see any flags around using `dotnet publish` instead of `dotnet build`. I have a feeling that wouldn't work well with nix's architecture.

I did spot `selfContainedBuild` though, wondering if that'd make any difference?

l0b0:
> I did spot `selfContainedBuild` though, wondering if that'd make any difference 🤔

That's in https://github.com/Nexus-Mods/NexusMods.App/blob/main/src/NexusMods.App/app.pupnet.conf as well, so I'll add it.

erri120:
> I can't see any flags around using dotnet publish instead of dotnet build. I have a feeling that wouldn't work well with nix's architecture?

`dotnet publish` is the "indented" way to prepare an application for deployment. I don't know the specifics. Some differences between `dotnet publish` and `dotnet build` include defaulting to `Release` instead of `Debug` and some additional arguments.

We use `--self-contained true` which will include the runtime in the output. This means that anyone installing the application doesn't need to also install the runtime. It does require setting a runtime identifier like `linux-x64` (or `--use-current-runtime`) so that the build system knows what runtime to put in the output.

`-p:PublishReadyToRun=true` and  `-p:PublishSingleFile=true` are also pretty great, the former enables ReadyToRun which is a variation of AOT for better startup performance, while the latter will bundle everything into one file (plus dbg symbols).

l0b0:
> > I can't see any flags around using dotnet publish instead of dotnet build. I have a feeling that wouldn't work well with nix's architecture?
> 
> `dotnet publish` is the "indented" way to prepare an application for deployment. I don't know the specifics. Some differences between `dotnet publish` and `dotnet build` include defaulting to `Release` instead of `Debug` and some additional arguments.
> 
> We use `--self-contained true` which will include the runtime in the output. This means that anyone installing the application doesn't need to also install the runtime. It does require setting a runtime identifier like `linux-x64` (or `--use-current-runtime`) so that the build system knows what runtime to put in the output.
> 
> `-p:PublishReadyToRun=true` and `-p:PublishSingleFile=true` are also pretty great, the former enables ReadyToRun which is a variation of AOT for better startup performance, while the latter will bundle everything into one file (plus dbg symbols).

Thanks! I've set `selfContainedBuild = true` and all the property flags.

erri120:
> -p:Version=${APP_VERSION} probably isn't necessary, I'd guess.

You do want to set the version somewhere. I don't know if you magic nix script already does that, but the version should be passed into the build.

l0b0:
> > -p:Version=${APP_VERSION} probably isn't necessary, I'd guess.
> 
> You do want to set the version somewhere. I don't know if you magic nix script already does that, but the version should be passed into the build.

Oh, maybe that's causing the NETSDK1134 error?

MattSturgeon:
> > I can't see any flags around using dotnet publish instead of dotnet build. I have a feeling that wouldn't work well with nix's architecture?
> 
> `dotnet publish` is the "indented" way to prepare an application for deployment. I don't know the specifics. Some differences between `dotnet publish` and `dotnet build` include defaulting to `Release` instead of `Debug` and some additional arguments.
> 
> We use `--self-contained true` which will include the runtime in the output. This means that anyone installing the application doesn't need to also install the runtime. It does require setting a runtime identifier like `linux-x64` (or `--use-current-runtime`) so that the build system knows what runtime to put in the output.
> 
> `-p:PublishReadyToRun=true` and `-p:PublishSingleFile=true` are also pretty great, the former enables ReadyToRun which is a variation of AOT for better startup performance, while the latter will bundle everything into one file (plus dbg symbols).

While I can see how it would be useful for appimages and maybe other distros, a lot of that stuff probably isn't what we want on nix, because we want to use the dotnet runtime that already exists in the nix store, rather than bundling it again in the package. Maybe some @NixOS/dotnet gurus can point us in the right direction?

pkgs/by-name/ne/nexusmods-app/package.nix
```diff
@@ -44,44 +38,59 @@ buildDotnetModule rec {
...
   '';
 
-  postPatch = ''
-    ln --force --symbolic "${lib.getExe _7zzWithOptionalUnfreeRarSupport}" src/ArchiveManagement/NexusMods.FileExtractor/runtimes/linux-x64/native/7zz
-  '';
+  selfContainedBuild = true;
+
+  dotnetBuildFlags = [
+    # Tell the app it is a distro package; affects wording in update prompts
+    "--property:INSTALLATION_METHOD_PACKAGE_MANAGER=1"
```
GGG-KILLER, line 48: This looks like an env var, shouldn't it be set in the wrapper instead of here?

pkgs/by-name/ne/nexusmods-app/package.nix
```diff
@@ -44,44 +38,59 @@ buildDotnetModule rec {
...
+
+    # Don't include upstream's 7zz binary; we use the nixpkgs version
+    "--property:NEXUSMODS_APP_USE_SYSTEM_EXTRACTOR=1"
```
GGG-KILLER, line 51: Same as above, looks like an env var.

pkgs/by-name/ne/nexusmods-app/package.nix
```diff
@@ -44,44 +38,59 @@ buildDotnetModule rec {
...
+
+    # From https://github.com/Nexus-Mods/NexusMods.App/blob/v0.5.2/src/NexusMods.App/app.pupnet.conf#L39
+    "--property:Version=${version}"
+    "--property:TieredCompilation=true"
+    "--property:PublishReadyToRun=true"
```
GGG-KILLER, line 56: This makes it so that the code is pre-compiled to the target runtime, given that we don't add the framework's RID unless it's a self-contained build, I don't recommend setting this.

pkgs/by-name/ne/nexusmods-app/package.nix
```diff
@@ -44,44 +38,59 @@ buildDotnetModule rec {
...
+    "--property:PublishSingleFile=true"
```
GGG-KILLER, line 57: This makes it so that the end program is only a single file, but considering we're in Nix land and not just passing around a link to download the program, it is unnecessary.

GGG-KILLER (changes requested):
A few things that need changing

GGG-KILLER:
Also about the self-contained part, I'd prefer it is not set because then we'd be unnecessarily packing the .NET Runtime along with this app thus duplicating it in the nix store unnecessarily.

pkgs/by-name/ne/nexusmods-app/package.nix
MattSturgeon: It is a build prop

pkgs/by-name/ne/nexusmods-app/package.nix
MattSturgeon: It is a build prop

MattSturgeon:
@GGG-KILLER thanks for your input! Do you have any thoughts on:

```
[nix-shell:~/.cache/nixpkgs-review/pr-318647-3/results/nexusmods-app/bin]$ ./NexusMods.App 
Unhandled exception. System.IO.FileNotFoundException: Could not load file or assembly 'Microsoft.Extensions.DependencyInjection, Version=8.0.0.0, Culture=neutral, PublicKeyToken=adb9793829ddae60'. The system cannot find the file specified.

File name: 'Microsoft.Extensions.DependencyInjection, Version=8.0.0.0, Culture=neutral, PublicKeyToken=adb9793829ddae60'
   at Microsoft.Extensions.Hosting.HostBuilder..ctor()
   at NexusMods.App.Program.BuildSettingsHost() in /build/source/src/NexusMods.App/Program.cs:line 171
   at NexusMods.App.Program.Main(String[] args) in /build/source/src/NexusMods.App/Program.cs:line 39
Aborted (core dumped)
```
?

GGG-KILLER:
> @GGG-KILLER thanks for your input! Do you have any thoughts on the runtime error we're all getting:
> 
> ```
> [nix-shell:~/.cache/nixpkgs-review/pr-318647-3/results/nexusmods-app/bin]$ ./NexusMods.App 
> Unhandled exception. System.IO.FileNotFoundException: Could not load file or assembly 'Microsoft.Extensions.DependencyInjection, Version=8.0.0.0, Culture=neutral, PublicKeyToken=adb9793829ddae60'. The system cannot find the file specified.
> 
> File name: 'Microsoft.Extensions.DependencyInjection, Version=8.0.0.0, Culture=neutral, PublicKeyToken=adb9793829ddae60'
>    at Microsoft.Extensions.Hosting.HostBuilder..ctor()
>    at NexusMods.App.Program.BuildSettingsHost() in /build/source/src/NexusMods.App/Program.cs:line 171
>    at NexusMods.App.Program.Main(String[] args) in /build/source/src/NexusMods.App/Program.cs:line 39
> Aborted (core dumped)
> ```
> 
> ?

@MattSturgeon this seems to be because it is not finding its dependencies, maybe because of the single file or self-contained deployment, I'd try disabling those and see if it works.

I'll run some tests locally and see if I can get it to build

pkgs/by-name/ne/nexusmods-app/package.nix
GGG-KILLER: Oh my bad, didn't check previous conversations. Really confusing from the Nexus App folks.

pkgs/by-name/ne/nexusmods-app/package.nix
GGG-KILLER: Oh my bad, didn't check previous conversations. Really confusing from the Nexus App folks.

GGG-KILLER:
@l0b0 and @MattSturgeon I've gotten it to work by making the following changes, if you could review it to see if it is and works as expected:
```patch
diff --git a/pkgs/by-name/ne/nexusmods-app/package.nix b/pkgs/by-name/ne/nexusmods-app/package.nix
index 8ccc2ad3de5e..571970860a1f 100644
--- a/pkgs/by-name/ne/nexusmods-app/package.nix
+++ b/pkgs/by-name/ne/nexusmods-app/package.nix
@@ -27,7 +27,7 @@ buildDotnetModule rec {
     hash = "sha256-NQUHvMGFeexy/EYvEwZrXUHAAuBqA3fyGm7FYjXTq6s=";
   };
 
-  projectFile = "NexusMods.App.sln";
+  projectFile = "src/NexusMods.App/NexusMods.App.csproj";
 
   nativeBuildInputs = [ copyDesktopItems ];
 
@@ -41,8 +41,6 @@ buildDotnetModule rec {
       --replace-fail '</PropertyGroup>' '<ErrorOnDuplicatePublishOutputFiles>false</ErrorOnDuplicatePublishOutputFiles></PropertyGroup>'
   '';
 
-  selfContainedBuild = true;
-
   dotnetBuildFlags = [
     # Tell the app it is a distro package; affects wording in update prompts
     "--property:INSTALLATION_METHOD_PACKAGE_MANAGER=1"
@@ -53,8 +51,6 @@ buildDotnetModule rec {
     # From https://github.com/Nexus-Mods/NexusMods.App/blob/v0.5.2/src/NexusMods.App/app.pupnet.conf#L39
     "--property:Version=${version}"
     "--property:TieredCompilation=true"
-    "--property:PublishReadyToRun=true"
-    "--property:PublishSingleFile=true"
   ];
 
   makeWrapperArgs = [
```

MattSturgeon:
> @l0b0 and @MattSturgeon I've gotten it to work by making the following changes

Thanks, that's brilliant! Sounds like `src/NexusMods.App/NexusMods.App.csproj` made the difference then, could you explain why that works better than using the sln?

The other lines you removed were only added as failed attempts to solve the runtime issues.

> if you could review it to see if it is and works as expected

I'll take a look as soon as I can get to a suitable machine!

GGG-KILLER:
> Thanks, that's brilliant! Sounds like `src/NexusMods.App/NexusMods.App.csproj` made the difference then, could you explain why that works better than using the sln?

buildDotnetModule is mostly made to work with `.csproj` files from my experience, I tried removing the flags to see if they were the issue but once I realized it still wasn't working, I though that the issue was the `ErrorOnDuplicatePublishOutputFiles` that was being set in `preConfigure`.
It looked more like it was hiding an error rather than actually fixing something.

MattSturgeon:
> > Thanks, that's brilliant! Sounds like `src/NexusMods.App/NexusMods.App.csproj` made the difference then, could you explain why that works better than using the sln?
> 
> buildDotnetModule is mostly made to work with `.csproj` files from my experience, I tried removing the flags to see if they were the issue but once I realized it still wasn't working, I though that the issue was the `ErrorOnDuplicatePublishOutputFiles` that was being set in `preConfigure` as it looked more like it was hiding an error rather than actually fixing something, and in this case it was, given that multiple projects had the Dependency Injection package, it was probably just skipping it.

Thanks for explaining. Re-reading the manual again and this paragraph now makes a little more sense:

> `projectFile` is used for specifying the dotnet project file, relative to the source root. These have `.sln` (entire solution) or `.csproj` (single project) file extensions. This can be a list of multiple projects as well. When omitted, will attempt to find and build the solution (`.sln`). If running into problems, make sure to set it to a file (or a list of files) with the `.csproj` extension - building applications as entire solutions is not fully supported by the .NET CLI.

With your patch everything builds ok, but I cannot get the app to actually run. This time it exits silently with no output or exit codes.

The build also seems much faster, and succeeds on my tmpfs laptop (which normally runs out of memory while building).

My hunch is that we need more than just the `src/NexusMods.App/NexusMods.App.csproj` project? The top-level sln lists all of these projects:

<details><summary>All project files</summary>
<p>

```nix
  projectFile = [
    "benchmarks/NexusMods.Benchmarks/NexusMods.Benchmarks.csproj"
    "src/Abstractions/NexusMods.Abstractions.Activities/NexusMods.Abstractions.Activities.csproj"
    "src/Abstractions/NexusMods.Abstractions.Cli/NexusMods.Abstractions.Cli.csproj"
    "src/Abstractions/NexusMods.Abstractions.DiskState/NexusMods.Abstractions.DiskState.csproj"
    "src/Abstractions/NexusMods.Abstractions.FileExtractor/NexusMods.Abstractions.FileExtractor.csproj"
    "src/Abstractions/NexusMods.Abstractions.FileStore/NexusMods.Abstractions.FileStore.csproj"
    "src/Abstractions/NexusMods.Abstractions.GameLocators/NexusMods.Abstractions.GameLocators.csproj"
    "src/Abstractions/NexusMods.Abstractions.Games.Diagnostics/NexusMods.Abstractions.Games.Diagnostics.csproj"
    "src/Abstractions/NexusMods.Abstractions.Games/NexusMods.Abstractions.Games.csproj"
    "src/Abstractions/NexusMods.Abstractions.GuidedInstallers/NexusMods.Abstractions.GuidedInstallers.csproj"
    "src/Abstractions/NexusMods.Abstractions.HttpDownloader/NexusMods.Abstractions.HttpDownloader.csproj"
    "src/Abstractions/NexusMods.Abstractions.IO/NexusMods.Abstractions.IO.csproj"
    "src/Abstractions/NexusMods.Abstractions.Installers/NexusMods.Abstractions.Installers.csproj"
    "src/Abstractions/NexusMods.Abstractions.Loadouts.Synchronizers/NexusMods.Abstractions.Loadouts.Synchronizers.csproj"
    "src/Abstractions/NexusMods.Abstractions.Loadouts/NexusMods.Abstractions.Loadouts.csproj"
    "src/Abstractions/NexusMods.Abstractions.Messaging/NexusMods.Abstractions.Messaging.csproj"
    "src/Abstractions/NexusMods.Abstractions.MnemonicDB.Attributes/NexusMods.Abstractions.MnemonicDB.Attributes.csproj"
    "src/Abstractions/NexusMods.Abstractions.NexusWebApi/NexusMods.Abstractions.NexusWebApi.csproj"
    "src/Abstractions/NexusMods.Abstractions.Serialization/NexusMods.Abstractions.Serialization.csproj"
    "src/Abstractions/NexusMods.Abstractions.Settings/NexusMods.Abstractions.Settings.csproj"
    "src/Abstractions/NexusMods.Abstractions.Telemetry/NexusMods.Abstractions.Telemetry.csproj"
    "src/Abstractions/NexusMods.Abstractions.Triggers/NexusMods.Abstractions.Triggers.csproj"
    "src/ArchiveManagement/NexusMods.FileExtractor/NexusMods.FileExtractor.csproj"
    "src/Examples/Examples.csproj"
    "src/Extensions/NexusMods.Extensions.BCL/NexusMods.Extensions.BCL.csproj"
    "src/Extensions/NexusMods.Extensions.DependencyInjection/NexusMods.Extensions.DependencyInjection.csproj"
    "src/Extensions/NexusMods.Extensions.DynamicData/NexusMods.Extensions.DynamicData.csproj"
    "src/Extensions/NexusMods.Extensions.Hashing/NexusMods.Extensions.Hashing.csproj"
    "src/Games/NexusMods.Games.AdvancedInstaller.UI/NexusMods.Games.AdvancedInstaller.UI.csproj"
    "src/Games/NexusMods.Games.AdvancedInstaller/NexusMods.Games.AdvancedInstaller.csproj"
    "src/Games/NexusMods.Games.BethesdaGameStudios/NexusMods.Games.BethesdaGameStudios.csproj"
    "src/Games/NexusMods.Games.BladeAndSorcery/NexusMods.Games.BladeAndSorcery.csproj"
    "src/Games/NexusMods.Games.DarkestDungeon/NexusMods.Games.DarkestDungeon.csproj"
    "src/Games/NexusMods.Games.FOMOD.UI/NexusMods.Games.FOMOD.UI.csproj"
    "src/Games/NexusMods.Games.FOMOD/NexusMods.Games.FOMOD.csproj"
    "src/Games/NexusMods.Games.Generic/NexusMods.Games.Generic.csproj"
    "src/Games/NexusMods.Games.MountAndBlade2Bannerlord/NexusMods.Games.MountAndBlade2Bannerlord.csproj"
    "src/Games/NexusMods.Games.RedEngine/NexusMods.Games.RedEngine.csproj"
    "src/Games/NexusMods.Games.Reshade/NexusMods.Games.Reshade.csproj"
    "src/Games/NexusMods.Games.Sifu/NexusMods.Games.Sifu.csproj"
    "src/Games/NexusMods.Games.StardewValley.SMAPI/NexusMods.Games.StardewValley.SMAPI.csproj"
    "src/Games/NexusMods.Games.StardewValley/NexusMods.Games.StardewValley.csproj"
    "src/Games/NexusMods.Games.TestHarness/NexusMods.Games.TestHarness.csproj"
    "src/Networking/NexusMods.Networking.Downloaders/NexusMods.Networking.Downloaders.csproj"
    "src/Networking/NexusMods.Networking.HttpDownloader/NexusMods.Networking.HttpDownloader.csproj"
    "src/Networking/NexusMods.Networking.NexusWebApi/NexusMods.Networking.NexusWebApi.csproj"
    "src/NexusMods.Activities/NexusMods.Activities.csproj"
    "src/NexusMods.App.BuildInfo/NexusMods.App.BuildInfo.csproj"
    "src/NexusMods.App.Cli/NexusMods.App.Cli.csproj"
    "src/NexusMods.App.Generators.Diagnostics/NexusMods.App.Generators.Diagnostics.Sample/NexusMods.App.Generators.Diagnostics.Sample.csproj"
    "src/NexusMods.App.Generators.Diagnostics/NexusMods.App.Generators.Diagnostics/NexusMods.App.Generators.Diagnostics.csproj"
    "src/NexusMods.App.UI/NexusMods.App.UI.csproj"
    "src/NexusMods.App/NexusMods.App.csproj"
    "src/NexusMods.CrossPlatform/NexusMods.CrossPlatform.csproj"
    "src/NexusMods.DataModel/NexusMods.DataModel.csproj"
    "src/NexusMods.Icons/NexusMods.Icons.csproj"
    "src/NexusMods.ProxyConsole.Abstractions/NexusMods.ProxyConsole.Abstractions.csproj"
    "src/NexusMods.ProxyConsole/NexusMods.ProxyConsole.csproj"
    "src/NexusMods.Settings/NexusMods.Settings.csproj"
    "src/NexusMods.SingleProcess/NexusMods.SingleProcess.csproj"
    "src/NexusMods.StandardGameLocators/NexusMods.StandardGameLocators.csproj"
    "src/NexusMods.Telemetry/NexusMods.Telemetry.csproj"
    "src/Themes/NexusMods.Themes.NexusFluentDark/NexusMods.Themes.NexusFluentDark.csproj"
    "tests/Abstractions/NexusMods.Abstractions.Games.Tests/NexusMods.Abstractions.Games.Tests.csproj"
    "tests/Abstractions/NexusMods.Abstractions.GuidedInstallers.Tests/NexusMods.Abstractions.GuidedInstallers.Tests.csproj"
    "tests/Abstractions/NexusMods.Abstractions.Installers.Tests/NexusMods.Abstractions.Installers.Tests.csproj"
    "tests/Abstractions/NexusMods.Abstractions.Serialization.Tests/NexusMods.Abstractions.Serialization.Tests.csproj"
    "tests/Extensions/NexusMods.BCL.Extensions.Tests/NexusMods.BCL.Extensions.Tests.csproj"
    "tests/Games/NexusMods.Games.AdvancedInstaller.Tests/NexusMods.Games.AdvancedInstaller.Tests.csproj"
    "tests/Games/NexusMods.Games.AdvancedInstaller.UI.Tests/NexusMods.Games.AdvancedInstaller.UI.Tests.csproj"
    "tests/Games/NexusMods.Games.BethesdaGameStudios.Tests/NexusMods.Games.BethesdaGameStudios.Tests.csproj"
    "tests/Games/NexusMods.Games.BladeAndSorcery.Tests/NexusMods.Games.BladeAndSorcery.Tests.csproj"
    "tests/Games/NexusMods.Games.DarkestDungeon.Tests/NexusMods.Games.DarkestDungeon.Tests.csproj"
    "tests/Games/NexusMods.Games.FOMOD.Tests/NexusMods.Games.FOMOD.Tests.csproj"
    "tests/Games/NexusMods.Games.Generic.Tests/NexusMods.Games.Generic.Tests.csproj"
    "tests/Games/NexusMods.Games.MountAndBlade2Bannerlord.Tests/NexusMods.Games.MountAndBlade2Bannerlord.Tests.csproj"
    "tests/Games/NexusMods.Games.RedEngine.Tests/NexusMods.Games.RedEngine.Tests.csproj"
    "tests/Games/NexusMods.Games.Sifu.Tests/NexusMods.Games.Sifu.Tests.csproj"
    "tests/Games/NexusMods.Games.StardewValley.Tests/NexusMods.Games.StardewValley.Tests.csproj"
    "tests/Games/NexusMods.Games.TestFramework/NexusMods.Games.TestFramework.csproj"
    "tests/Networking/NexusMods.Networking.Downloaders.Tests/NexusMods.Networking.Downloaders.Tests.csproj"
    "tests/Networking/NexusMods.Networking.HttpDownloader.Tests/NexusMods.Networking.HttpDownloader.Tests.csproj"
    "tests/Networking/NexusMods.Networking.NexusWebApi.Tests/NexusMods.Networking.NexusWebApi.Tests.csproj"
    "tests/NexusMods.App.Generators.Diagnostics.Tests/NexusMods.App.Generators.Diagnostics.Tests.csproj"
    "tests/NexusMods.CLI.Tests/NexusMods.CLI.Tests.csproj"
    "tests/NexusMods.CrossPlatform.Tests/NexusMods.CrossPlatform.Tests.csproj"
    "tests/NexusMods.DataModel.Tests/NexusMods.DataModel.Tests.csproj"
    "tests/NexusMods.FileExtractor.Tests/NexusMods.FileExtractor.Tests.csproj"
    "tests/NexusMods.ProxyConsole.Tests/NexusMods.ProxyConsole.Tests.csproj"
    "tests/NexusMods.Settings.Tests/NexusMods.Settings.Tests.csproj"
    "tests/NexusMods.StandardGameLocators.TestHelpers/NexusMods.StandardGameLocators.TestHelpers.csproj"
    "tests/NexusMods.StandardGameLocators.Tests/NexusMods.StandardGameLocators.Tests.csproj"
    "tests/NexusMods.UI.Tests/NexusMods.UI.Tests.csproj"
  ];
```

</p>
</details>


Building with all project files (as above) results in the following compile error:

(Oddly, I got to the check phase once, but re-built with `doCheck = false` because some test was either hung or taking forever, this error only happened on that second build):

```
error: builder for '/nix/store/bkbapbn9s0m2dqyb47l3hl9nn46jjw84-nexusmods-app-0.5.2.drv' failed with exit code 1;
       last 10 log lines:
       > /build/source/tests/NexusMods.UI.Tests/WorkspaceSystem/GridUtilsTests/GetResizersTests.cs(74,22): warning CS0219: The variable 'max' is assigned but its value is never used [/build/source/tests/NexusMods.UI.Tests/NexusMods.UI.Tests.csproj]
       >     125 Warning(s)
       >     0 Error(s)
       >
       > Time Elapsed 00:00:11.84
       > Finished dotnetBuildHook
       > buildPhase completed in 6 minutes 23 seconds
       > Running phase: installPhase
       > Executing dotnetInstallHook
       > /nix/store/588kjk7hlqjg6ixw6nnsz9qd4qkqa7wj-dotnet-sdk-8.0.301/sdk/8.0.301/Sdks/Microsoft.NET.Sdk/targets/Microsoft.NET.ConflictResolution.targets(112,5): error NETSDK1152: Found multiple publish output files with the same relative path: /build/source/tests/NexusMods.DataModel.Tests/xunit.runner.json, /build/source/tests/NexusMods.StandardGameLocators.TestHelpers/xunit.runner.json. [/build/source/benchmarks/NexusMods.Benchmarks/NexusMods.Benchmarks.csproj]
       For full logs, run 'nix log /nix/store/bkbapbn9s0m2dqyb47l3hl9nn46jjw84-nexusmods-app-0.5.2.drv'.
```

[Full log](https://github.com/user-attachments/files/15920864/bkbapbn9s0m2dqyb47l3hl9nn46jjw84-nexusmods-app-0.5.2.log) (it's big)

Also, some tests are either hanging or taking a _very_ long time, so I'm building with `doCheck = false`.

I had figured the various `ProjectReference` elements in https://github.com/Nexus-Mods/NexusMods.App/blob/main/src/NexusMods.App/NexusMods.App.csproj would mean other projects were included automatically?

I hope we don't have to explicitly list all project files; keeping that list in sync would make updating the package a nightmare!

pkgs/by-name/ne/nexusmods-app/package.nix
```diff
@@ -44,44 +38,59 @@ buildDotnetModule rec {
...
+  ];
 
   makeWrapperArgs = [
-    "--prefix PATH : ${lib.makeBinPath [desktop-file-utils]}"
+    "--prefix PATH : ${lib.makeBinPath [ desktop-file-utils ]}"
```
MattSturgeon, line 61:
```suggestion
```

pkgs/by-name/ne/nexusmods-app/package.nix
```diff
@@ -44,44 +38,59 @@ buildDotnetModule rec {
...
     "--set APPIMAGE $out/bin/${meta.mainProgram}" # Make associating with nxm links work on Linux
   ];
 
+  propagatedBuildInputs = [ (_7zz.override { inherit enableUnfree; }) ];
+
   runtimeDeps = [
```
MattSturgeon, line 67:
```suggestion
  runtimeDeps = [
    desktop-file-utils
```

MattSturgeon (changes requested):
I'm still able to run `nix run .#nexusmods-app -- associate-nxm` with this patch:

<details><summary>Details</summary>
<p>

```diff
diff --git a/pkgs/by-name/ne/nexusmods-app/package.nix b/pkgs/by-name/ne/nexusmods-app/package.nix
index e5a688a33b4a..5162cc9d396a 100644
--- a/pkgs/by-name/ne/nexusmods-app/package.nix
+++ b/pkgs/by-name/ne/nexusmods-app/package.nix
@@ -49,13 +49,13 @@ buildDotnetModule rec {
   ];
 
   makeWrapperArgs = [
-    "--prefix PATH : ${lib.makeBinPath [ desktop-file-utils ]}"
     "--set APPIMAGE $out/bin/${meta.mainProgram}" # Make associating with nxm links work on Linux
   ];
 
   propagatedBuildInputs = [ (_7zz.override { inherit enableUnfree; }) ];
 
   runtimeDeps = [
+    desktop-file-utils
     fontconfig
     libICE
     libSM
```

</p>
</details>

pkgs/by-name/ne/nexusmods-app/package.nix
```diff
@@ -1,41 +1,35 @@
...
   src = fetchFromGitHub {
     owner = "Nexus-Mods";
     repo = "NexusMods.App";
     rev = "v${version}";
     fetchSubmodules = true;
-    hash = "sha256-FzQphMhiC1g+6qmk/R1v4rq2ldy35NcaWm0RR1UlwLA=";
+    hash = "sha256-NQUHvMGFeexy/EYvEwZrXUHAAuBqA3fyGm7FYjXTq6s=";
   };
 
   projectFile = "NexusMods.App.sln";
```
MattSturgeon, line 30:
```suggestion
  projectFile = "src/NexusMods.App/NexusMods.App.csproj";
```

pkgs/by-name/ne/nexusmods-app/package.nix
```diff
@@ -44,44 +38,55 @@ buildDotnetModule rec {
 
   preConfigure = ''
     substituteInPlace Directory.Build.props \
-      --replace '</PropertyGroup>' '<ErrorOnDuplicatePublishOutputFiles>false</ErrorOnDuplicatePublishOutputFiles></PropertyGroup>'
+      --replace-fail '</PropertyGroup>' '<ErrorOnDuplicatePublishOutputFiles>false</ErrorOnDuplicatePublishOutputFiles></PropertyGroup>'
   '';
```
MattSturgeon, line 43: Remove `preConfigure`

MattSturgeon (changes requested):
As-per https://github.com/NixOS/nixpkgs/pull/318647#issuecomment-2181189269, this fixes the build.

l0b0:
How's this? I tried applying all the changes from the discussion above.

MattSturgeon:
Also, I can't find any .desktop files in `result`, even if I add 

> How's this? I tried applying all the changes from the discussion above.

Diff looks good. How's it run on your end?

I still have two issues outstanding:

#### Desktop files don't seem to be installed

I don't mean the nxm mimetype association done at runtime, rather the ones templated at build time.

I assume I should be able to find these in `result/share` after running `nix build .#nexusmods-app`?

I think we need to do something in the install phase to copy "data" files over to `xdg_data_home`? How do other packages handle their desktop files, man pages, etc?

I think this was already an issue with the 0.4.1 package.

#### I can't figure out how to launch the UI

`nix run .#nexusmods-app -- --help` works fine, and some other commands like `nix run .#nexusmods-app -- associate-nxm`.

However `nix run .#nexusmods-app` and `nix run .#nexusmods-app -- as-main-ui` both do nothing.

I suspect I am missing something or hitting an upstream bug?

MattSturgeon:
> Also:
> 
> ```
> ❯ nix run .#nexusmods-app -- associate-nxm
> gio: file:///home/victor/my%20projects/nixpkgs/$out/bin/NexusMods.App: Error when getting information for file “/home/victor/my projects/nixpkgs/$out/bin/NexusMods.App”: No such file or directory
> ```

Huh, `associate-nxm` worked fine for me. I wonder if something from my system env is affecting this? :thinking:

MattSturgeon:
> `./result/bin/NexusMods.App` terminates immediately with exit code 0 on my machine

> #### I can't figure out how to launch the UI
> 
> `nix run .#nexusmods-app -- --help` works fine, and some other commands like `nix run .#nexusmods-app -- associate-nxm`.
> 
> However `nix run .#nexusmods-app` and `nix run .#nexusmods-app -- as-main-ui` both do nothing.
> 
> I suspect I am missing something or hitting an upstream bug?

I wonder if https://github.com/Nexus-Mods/NexusMods.App/pull/1345 or some other upstream issue/pr is related?

GGG-KILLER:
> #### Desktop files don't seem to be installed
> 
> I don't mean the nxm mimetype association done at runtime, rather the ones templated at build time.
> 
> I assume I should be able to find these in `result/share` after running `nix build .#nexusmods-app`?

`buildDotnetModule` doesn't do anything in special with respect to those, it'd have to be something in the nexus app itself that generates the `.desktop` files for other tools to find it.

> #### I can't figure out how to launch the UI
> 
> `nix run .#nexusmods-app -- --help` works fine, and some other commands like `nix run .#nexusmods-app -- associate-nxm`.
> 
> However `nix run .#nexusmods-app` and `nix run .#nexusmods-app -- as-main-ui` both do nothing.
> 
> I suspect I am missing something or hitting an upstream bug?
It could be that our build command is still different from the one that is run on upstream, if they have verbose build logs it'd be good to check the actual command being ran and compare to ours

GGG-KILLER:
I think I've figured out a few issues, I'm testing some things and will update if I find anything.

One of the issues is that `INSTALLATION_METHOD_PACKAGE_MANAGER` and `NEXUSMODS_APP_USE_SYSTEM_EXTRACTOR` are being set incorrectly, they should be set using `--property:DefineConstants=INSTALLATION_METHOD_PACKAGE_MANAGER,NEXUSMODS_APP_USE_SYSTEM_EXTRACTOR`.
I'm gonna try to make the build process more similar to the ones in the GHA logs.

GGG-KILLER:
Also @MattSturgeon, the `.desktop` file is generated by their packing tool (PupNet), so we'd have to emulate the .desktop file creation on our side to have it.

MattSturgeon:
> I think I've figured out a few issues, I'm testing some things and will update if I find anything.

Thanks for continuing to look into this. Both myself and @l0b0 have limited dotnet experience, so this is very helpful!

> `NEXUSMODS_APP_USE_SYSTEM_EXTRACTOR` seems to be an env var??? I'll leave it on `--property` for now and see if it works.

Based on how it's declared (and documented) in [this file](https://github.com/Nexus-Mods/NexusMods.App/blob/v0.5.2/Directory.Build.props#L25-L28) I'm pretty sure it _is_ a build prop not an env var.

> Also, the `.desktop` file is generated by their packing tool (PupNet), so we'd have to emulate the .desktop file creation on our side to have it.

Good to know. Based on https://github.com/Nexus-Mods/NexusMods.App/issues/1669, https://github.com/Nexus-Mods/NexusMods.App/issues/1668 and similar issues, upstream seem keen to streamline the packaging process. I wonder if they'd be willing to move desktop files into the "main" build process?

GGG-KILLER:
Ok, I've managed to make some progress, although I don't know how this happened exactly 😅 
image

MattSturgeon:
> although I don't know how this happened exactly 😅

Same error as https://github.com/NixOS/nixpkgs/pull/318647#issuecomment-2181736897 but in a GUI dialog

GGG-KILLER:
For now the changes I've made are the following:

```patch
diff --git a/pkgs/by-name/ne/nexusmods-app/package.nix b/pkgs/by-name/ne/nexusmods-app/package.nix
index 5162cc9d396a..ddf9cb63920f 100644
--- a/pkgs/by-name/ne/nexusmods-app/package.nix
+++ b/pkgs/by-name/ne/nexusmods-app/package.nix
@@ -38,7 +38,7 @@ buildDotnetModule rec {
 
   dotnetBuildFlags = [
     # Tell the app it is a distro package; affects wording in update prompts
-    "--property:INSTALLATION_METHOD_PACKAGE_MANAGER=1"
+    "--property:DefineConstants=INSTALLATION_METHOD_PACKAGE_MANAGER"
 
     # Don't include upstream's 7zz binary; we use the nixpkgs version
     "--property:NEXUSMODS_APP_USE_SYSTEM_EXTRACTOR=1"
@@ -49,7 +49,7 @@ buildDotnetModule rec {
   ];
 
   makeWrapperArgs = [
-    "--set APPIMAGE $out/bin/${meta.mainProgram}" # Make associating with nxm links work on Linux
+    "--set APPIMAGE ${placeholder "out"}/bin/${meta.mainProgram}" # Make associating with nxm links work on Linux
   ];
 
   propagatedBuildInputs = [ (_7zz.override { inherit enableUnfree; }) ];
```

still trying to get the UI to work but unable to.

erri120:
> Good to know. Based on Nexus-Mods/NexusMods.App#1669, Nexus-Mods/NexusMods.App#1668 and similar issues, upstream seem keen to streamline the packaging process. I wonder if they'd be willing to move desktop files into the "main" build process?

What would be the easiest for you? I was thinking about having a `nexusmods-app.desktop` and `nexusmods-app-nxm.desktop` file inside the repo itself, containing everything except the `Path` value. Example:

```
[Desktop Entry]
Name=NexusMods.App NXM Handler
Terminal=false
Type=Application
Exec=NexusMods.App protocol-invoke --url %u
NoDisplay=true
MimeType=x-scheme-handler/nxm
```

`Exec` and either be a full path or just the file name. We can't use a static full path, and if you just have a file name, then the executable has to be inside `$PATH`, as detailed by the [specs](https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html#exec-variables).

MattSturgeon:
> What would be the easiest for you? I was thinking about having a `nexusmods-app.desktop` and `nexusmods-app-nxm.desktop` file inside the repo itself, containing everything except the `Path` value.

I think that'd probably work fine

IDK about other distros, but for our purposes it probably isn't a big deal to just have templates we can modify with sed. We'd probably do something like this:

```shell
# Nix's string interpolation will replace ${lib.getExe nexusmods-app} with something like
# /nix/store/zcpv3243hwcyg9q8dv2h84sbivfxnq42-nexusmods-app-0.5.2/bin/NexusMods.App
sed -e 's/^Exec=[^ ]\+/Exec=${lib.getExe nexusmods-app}/'
```

> `Exec` could either be a full path or just the file name. We can't use a static full path, and if you just have a file name, then the executable has to be inside `$PATH`, as detailed by the [specs](https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html#exec-variables).

I think some distros like to specify an "install prefix" that build scripts should remove from any absolute paths, though it's been a while since I messed with other distro's packaging systems.

I think practically speaking, the binary would likely be on the PATH whenever a .desktop file is being used to run it, however I think we'd still rather inject the absolute nix-store path into the files to avoid any weird edge-cases.

Incidentally, the `associate-nxm` command seems to be doing what we'd expect (at least on systems where we don't get the gio error). It correctly uses the absolute path, meaning it'd still work if someone ran the app ephemerally (without installing persistently to PATH etc), at least until the build gets gc'd from the nix store.

I haven't fully thought through whether we'd prefer to use this runtime mimetype association or to install that statically at build time :thinking:

pkgs/by-name/ne/nexusmods-app/package.nix
```diff
@@ -1,87 +1,87 @@
...
     owner = "Nexus-Mods";
     repo = "NexusMods.App";
     rev = "v${version}";
     fetchSubmodules = true;
-    hash = "sha256-FzQphMhiC1g+6qmk/R1v4rq2ldy35NcaWm0RR1UlwLA=";
+    hash = "sha256-NQUHvMGFeexy/EYvEwZrXUHAAuBqA3fyGm7FYjXTq6s=";
   };
 
-  projectFile = "NexusMods.App.sln";
+  projectFile = "src/NexusMods.App/NexusMods.App.csproj";
```
corngood, line 30: I noticed that making this change stops all the `dotnet test` tests from running.  Perhaps this change can be removed once https://github.com/NixOS/nixpkgs/pull/327651 is in?

PR #266705, opened 2023-11-10
lib: add fromHexString

lucasew:
## Description of changes
Add a function that convert an hexadecimal string to it's int representation

Simplification of https://github.com/NixOS/nixpkgs/pull/258834

How to test:
- ` nix-instantiate --eval lib/tests/misc.nix`

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
- For non-Linux: Is sandboxing enabled in `nix.conf`? (See [Nix manual](https://nixos.org/manual/nix/stable/command-ref/conf-file.html))
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:
  - [NixOS test(s)](https://nixos.org/manual/nixos/unstable/index.html#sec-nixos-tests) (look inside [nixos/tests](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests))
  - and/or [package tests](https://nixos.org/manual/nixpkgs/unstable/#sec-package-tests)
  - or, for functions and "core" functionality, tests in [lib/tests](https://github.com/NixOS/nixpkgs/blob/master/lib/tests) or [pkgs/test](https://github.com/NixOS/nixpkgs/blob/master/pkgs/test)
  - made sure NixOS tests are [linked](https://nixos.org/manual/nixpkgs/unstable/#ssec-nixos-tests-linking) to the relevant packages
- [ ] Tested compilation of all packages that depend on this change using `nix-shell -p nixpkgs-review --run "nixpkgs-review rev HEAD"`. Note: all changes have to be committed, also see [nixpkgs-review usage](https://github.com/Mic92/nixpkgs-review#usage)
- [ ] Tested basic functionality of all binary files (usually in `./result/bin/`)
- [23.11 Release Notes](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2311.section.md) (or backporting [23.05 Release notes](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2305.section.md))
  - [ ] (Package updates) Added a release notes entry if the change is major or breaking
  - [ ] (Module updates) Added a release notes entry if the change is significant
  - [ ] (Module addition) Added a release notes entry if adding a new NixOS module
[...]

PR #299409, opened 2024-03-27
Add network library

djacu:
## Description of changes

See https://github.com/NixOS/nixpkgs/pull/258250 for discussion about why this would be beneficial to nixpkgs.

This PR was created at the request of @infinisil to be smaller and include a design document.
The design document is fairly terse. Any comments or feedback are welcome to help improve it.

Considerations for this PR and differences from the previous PR.

- Going forward, the number of functions exposed will be significantly less. Rather than exporting many functions, a few functions (currently one) that create `IPv4Address` attrsets will be exposed.
- The attrset will have (lazy) attributes that accomplish the same functions as in the previous PR.
- There are now tests and a design document similar to the fileset and path libraries.
- Currently only IPv4 is implement to keep the PR small. IPv6 will be implemented in the future.
- The number of attributes in the `IPv4Address` attrset is limited to keep the PR small. More attributes will be implemented in the future.

## Things done

- [x] Tested, as applicable:
  - [NixOS test(s)](https://nixos.org/manual/nixos/unstable/index.html#sec-nixos-tests) (look inside [nixos/tests](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests))
  - and/or [package tests](https://nixos.org/manual/nixpkgs/unstable/#sec-package-tests)
  - or, for functions and "core" functionality, tests in [lib/tests](https://github.com/NixOS/nixpkgs/blob/master/lib/tests) or [pkgs/test](https://github.com/NixOS/nixpkgs/blob/master/pkgs/test)
- [x] Fits [CONTRIBUTING.md](https://github.com/NixOS/nixpkgs/blob/master/CONTRIBUTING.md).

---

Add a :+1: [reaction] to [pull requests you find important].

[reaction]: https://github.blog/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/
[pull requests you find important]: https://github.com/NixOS/nixpkgs/pulls?q=is%3Aopen+sort%3Areactions-%2B1-desc

PR #318712, opened 2024-06-10
lib.network: add ipv6 parser

woojiq:
### Description of changes

* Add a library function to parse and validate an IPv6 address from a
string. It can parse the first two versions of an IPv6 address according
to https://datatracker.ietf.org/doc/html/rfc4291#section-2.2. The third
form "x:x:x:x:x:x.d.d.d.d" is not yet implemented.
* Add shell script network.sh to test IPv6 parser functionality.
* Copy fromHexString function from the https://github.com/NixOS/nixpkgs/pull/266705

Mentor: @Janik-Haag 

### Things done
- [x] Tested, as applicable:
* [NixOS test(s)](https://nixos.org/manual/nixos/unstable/index.html#sec-nixos-tests) (look inside [nixos/tests](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests))
* and/or [package tests](https://nixos.org/manual/nixpkgs/unstable/#sec-package-tests)
* or, for functions and "core" functionality, tests in [lib/tests](https://github.com/NixOS/nixpkgs/blob/master/lib/tests) or [pkgs/test](https://github.com/NixOS/nixpkgs/blob/master/pkgs/test)
- [x] Formatted using nixfmt-rfc-style

Add a 👍 [reaction](https://github.blog/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to [pull requests you find important](https://github.com/NixOS/nixpkgs/pulls?q=is%3Aopen+sort%3Areactions-%2B1-desc).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,7 @@
 .idea/
 .nixos-test-history
 .vscode/
+.helix/
 outputs/
 result-*
 result
--- a/lib/network/default.nix
+++ b/lib/network/default.nix
@@ -1,0 +1,49 @@
+{ lib }:
+let
+  inherit (import ./internal.nix { inherit lib; }) _ipv6;
+in
+{
+  ipv6 = {
+    /**
+      Creates an `IPv6Address` object from an IPv6 address as a string. If
+      the prefix length is omitted, it defaults to 64. The parser is limited
+      to the first two versions of IPv6 addresses addressed in RFC 4291.
+      The form "x:x:x:x:x:x:d.d.d.d" is not yet implemented. Addresses are
+      NOT compressed, so they are not always the same as the canonical text
+      representation of IPv6 addresses defined in RFC 5952.
+
+      # Type
+
+      ```
+      fromString :: String -> IPv6Address
+      ```
+
+      # Examples
+
+      ```nix
+      fromString "2001:DB8::ffff/32"
+      => {
+        address = "2001:db8:0:0:0:0:0:ffff";
+        prefixLength = 32;
+      }
+      ```
+
+      # Arguments
+
+      - [addr] An IPv6 address with optional prefix length.
+    */
+    fromString =
+      addr:
+      let
+        splittedAddr = _ipv6.split addr;
+
+        addrInternal = splittedAddr.address;
+        prefixLength = splittedAddr.prefixLength;
+
+        address = _ipv6.toStringFromExpandedIp addrInternal;
+      in
+      {
+        inherit address prefixLength;
+      };
+  };
+}
--- a/lib/network/internal.nix
+++ b/lib/network/internal.nix
@@ -1,0 +1,209 @@
+{
+  lib ? import ../.,
+}:
+let
+  inherit (builtins)
+    map
+    match
+    genList
+    length
+    concatMap
+    head
+    toString
+    ;
+
+  inherit (lib) lists strings trivial;
+
+  inherit (lib.lists) last;
+
+  /*
+    IPv6 addresses are 128-bit identifiers. The preferred form is 'x:x:x:x:x:x:x:x',
+    where the 'x's are one to four hexadecimal digits of the eight 16-bit pieces of
+    the address. See RFC 4291.
+  */
+  ipv6Bits = 128;
+  ipv6Pieces = 8; # 'x:x:x:x:x:x:x:x'
+  ipv6PieceBits = 16; # One piece in range from 0 to 0xffff.
+  ipv6PieceMaxValue = 65535; # 2^16 - 1
+in
+let
+  /**
+    Expand an IPv6 address by removing the "::" compression and padding them
+    with the necessary number of zeros. Converts an address from the string to
+    the list of strings which then can be parsed using `_parseExpanded`.
+    Throws an error when the address is malformed.
+
+    # Type: String -> [ String ]
+
+    # Example:
+
+    ```nix
+    expandIpv6 "2001:DB8::ffff"
+    => ["2001" "DB8" "0" "0" "0" "0" "0" "ffff"]
+    ```
+  */
+  expandIpv6 =
+    addr:
+    if match "^[0-9A-Fa-f:]+$" addr == null then
+      throw "${addr} contains malformed characters for IPv6 address"
+    else
+      let
+        pieces = strings.splitString ":" addr;
+        piecesNoEmpty = lists.remove "" pieces;
+        piecesNoEmptyLen = length piecesNoEmpty;
+        zeros = genList (_: "0") (ipv6Pieces - piecesNoEmptyLen);
+        hasPrefix = strings.hasPrefix "::" addr;
+        hasSuffix = strings.hasSuffix "::" addr;
+        hasInfix = strings.hasInfix "::" addr;
+      in
+      if addr == "::" then
+        zeros
+      else if
+        let
+          emptyCount = length pieces - piecesNoEmptyLen;
+          emptyExpected =
+            # splitString produces two empty pieces when "::" in the beginning
+            # or in the end, and only one when in the middle of an address.
+            if hasPrefix || hasSuffix then
+              2
+            else if hasInfix then
+              1
+            else
+              0;
+        in
+        emptyCount != emptyExpected
+        || (hasInfix && piecesNoEmptyLen >= ipv6Pieces) # "::" compresses at least one group of zeros.
+        || (!hasInfix && piecesNoEmptyLen != ipv6Pieces)
+      then
+        throw "${addr} is not a valid IPv6 address"
+      # Create a list of 8 elements, filling some of them with zeros depending
+      # on where the "::" was found.
+      else if hasPrefix then
+        zeros ++ piecesNoEmpty
+      else if hasSuffix then
+        piecesNoEmpty ++ zeros
+      else if hasInfix then
+        concatMap (piece: if piece == "" then zeros else [ piece ]) pieces
+      else
+        pieces;
+
+  /**
+    Parses an expanded IPv6 address (see `expandIpv6`), converting each part
+    from a string to an u16 integer. Returns an internal representation of IPv6
+    address (list of integers) that can be easily processed by other helper
+    functions.
+    Throws an error some element is not an u16 integer.
+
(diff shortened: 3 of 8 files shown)


lib/network/default.nix
```diff
@@ -0,0 +1,107 @@
+{ lib }:
+{
+  ipv6 = {
+    /**
+      Normalize an IPv6 address by removing the "::" compressions and padding
+      them with the necessary number of zeros to create the correct address. Converts
+      an address from the string to the list of strings which then can be parsed using
+      `_parseNormalized`. Throws an error when the address is malformed.
```
mweinelt, line 8: Please talk about expanding and compressing, as that is the official lingo used in the RFC 4291. And make sure to reference RFC as the canonical source for these representations.

> Expand an IPv6 Address according to RFC 4291, by doing this and that

lib/network/default.nix
```diff
@@ -0,0 +1,107 @@
...
+
+      # Type: String -> [ String ]
+    */
+    _normalizeIpv6 =
```
Janik-Haag, line 12:
```suggestion
    _normalizeIp =
```
This is already defined in the ipv6 scope, so I think it's preferable to have:
```nix
lib.network.ipv6._normalizeIp
```
over
```nix
lib.network.ipv6._normalizeIpv6
```

lib/network/default.nix
```diff
@@ -0,0 +1,107 @@
+{ lib }:
+{
+  ipv6 = {
```
Janik-Haag, line 3: for ergonomics it would be nicer to just use v6 and v4 but I don't have hard feelings here.

lib/network/default.nix
```diff
@@ -0,0 +1,107 @@
...
+        let
+          doubleColon = "::";
+          piecesMax = 8;
+          pieces = lib.strings.splitString ":" addr;
+          piecesNoEmpty = lib.lists.remove "" pieces;
+          piecesNoEmptyLen = builtins.length piecesNoEmpty;
+          emptyCnt = builtins.length pieces - builtins.length piecesNoEmpty;
+          hasPrefix = lib.strings.hasPrefix doubleColon addr;
+          hasSuffix = lib.strings.hasSuffix doubleColon addr;
+          hasInfix = lib.strings.hasInfix doubleColon addr;
```
Janik-Haag, line 26: I think we should keep it simple and just write "::" over introducing a extra binding, because it makes the code easier to read since you don't have to look up how doubleColon.

```suggestion
          hasPrefix = lib.strings.hasPrefix "::" addr;
          hasSuffix = lib.strings.hasSuffix "::"  addr;
          hasInfix = lib.strings.hasInfix "::" addr;
```

lib/network/default.nix
```diff
@@ -0,0 +1,107 @@
...
+      if builtins.match "^[0-9A-Fa-f:]+$" addr == null then
+        throw "${addr} contains malformed characters for IPv6 address"
+      else
+        let
+          doubleColon = "::";
+          piecesMax = 8;
+          pieces = lib.strings.splitString ":" addr;
+          piecesNoEmpty = lib.lists.remove "" pieces;
+          piecesNoEmptyLen = builtins.length piecesNoEmpty;
+          emptyCnt = builtins.length pieces - builtins.length piecesNoEmpty;
```
Janik-Haag, line 23: emptyCnt is only used once in line 38, so I think we should move it directly into the local scope. I also think that we should call this emptyCount since everything else is written out and it's only internal anyways.

lib/network/default.nix
```diff
@@ -0,0 +1,107 @@
...
+            0 <= parsed && parsed <= 65535 # 2^16 - 1
+          then
+            parsed
+          else
+            throw "0x${hex} is not valid u16 integer";
+      in
+      assert lib.asserts.assertMsg (
+        builtins.length addr == 8
+      ) "_parseNormalized expected to recieve normalized IPv6 address";
+      builtins.map (el: u16FromHexStr el) addr;
```
Janik-Haag, line 79:
```suggestion
      builtins.map (piece: u16FromHexStr piece) addr;
```

We should keep the same language everywhere to avoid confusion.

Janik-Haag:
To address your questions:

> Do I need to somehow hide the internal functions more than just underscoring them?

Yes currently they would all get exposed, you have to put them into a let binding to make them truly internal. Take a look at https://github.com/NixOS/nixpkgs/blob/master/lib/fileset/default.nix#L103 as an example.

> Are the doc comments in the correct format (already existing files have different style)?

Some of them are some of them aren't, please take a look at the [example](https://github.com/nix-community/nixdoc?tab=readme-ov-file#example)

> Do I need to pull the ipv4 parser implementation to merge them together? https://github.com/NixOS/nixpkgs/pull/299409

no, we should keep them separate and preferably go for small prs, e.g. once the parser is fine merge this pr and then do another v6 one for cidr related functions.

lib/network/default.nix
woojiq: Personally, I think the "ip" prefix gives more clarity.
`network.v6.func` looks weird.

lib/tests/network.sh
```diff
@@ -0,0 +1,84 @@
...
+expectSuccess 'ipv6.fromString "::"'                                      '[0,0,0,0,0,0,0,0]'
+expectSuccess 'ipv6.fromString "::0000"'                                  '[0,0,0,0,0,0,0,0]'
+expectSuccess 'ipv6.fromString "::1"'                                     '[0,0,0,0,0,0,0,1]'
+expectSuccess 'ipv6.fromString "::ffff"'                                  '[0,0,0,0,0,0,0,65535]'
+expectSuccess 'ipv6.fromString "::000f"'                                  '[0,0,0,0,0,0,0,15]'
+expectSuccess 'ipv6.fromString "::1:1:1:1:1:1:1"'                         '[0,1,1,1,1,1,1,1]'
+expectSuccess 'ipv6.fromString "1::"'                                     '[1,0,0,0,0,0,0,0]'
+expectSuccess 'ipv6.fromString "1:1:1:1:1:1:1::"'                         '[1,1,1,1,1,1,1,0]'
+expectSuccess 'ipv6.fromString "1:1:1:1::1:1:1"'                          '[1,1,1,1,0,1,1,1]'
+expectSuccess 'ipv6.fromString "1::1"'                                    '[1,0,0,0,0,0,0,1]'
```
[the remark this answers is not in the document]
woojiq: I think we can use reserved ip in examples but for testing purpose reserved ip might not be enough.

woojiq:
@Janik-Haag @mweinelt I refactored the code and moved internal functions to internal.nix. Can you review again please?

lib/network/default.nix
mweinelt: I agree, proper lingo everywhere is what we should aim for.

Janik-Haag (changes requested):
I think this is getting into shape. There are some smaller points left to do, and you should add a lot more comments explaining different parts, I think. This part of lib currently is very undocumented internally compared to others like lib.fileset or lib.generators.

lib/network/internal.nix
```diff
@@ -0,0 +1,143 @@
...
+
+  inherit (lib.lists) last;
+
+  ipv6Bits = 128;
+  ipv6Pieces = 8;
+  ipv6PieceBits = 16;
+in
+let
+  /**
+    Expand an IPv6 address by removing the "::" compression and padding them
```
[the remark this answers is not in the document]
woojiq: RFC 4291:

> In order to make writing addresses containing zero bits easier, a special syntax is available to compress the zeros.

> or in compressed form:

lib/network/internal.nix
```diff
@@ -0,0 +1,143 @@
...
+      let
+        splitted = strings.splitString "/" addr;
+        splittedLength = length splitted;
+      in
+      if splittedLength == 1 then # [ ip ]
+        {
+          address = parseIpv6FromString apiName addr;
+          prefixLength = 64; # /64 is a standard size IPv6 subnet as defined by the IETF.
+        }
+      else if splittedLength == 2 then # [ ip subnet ]
```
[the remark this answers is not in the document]
woojiq: I agree, 128 sounds more logical if prefix length is omitted.

> I think we should be stricter with the user input and don't silently just set some prefix but instead fail saying that it's not a cidr address.

I added one `fromString` function and user can also bot default and cidr notation. I don't see a reason to create two separate API functions like `fromString` and `fromCidrString`, because this can be easily observed from the passed string - if it contains a `/`, then we try to parse it as cidr notation.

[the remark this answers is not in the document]
woojiq: Ok, when I started working on ipv6 types, I must have understood why we need separate functions for default and cidr notation. In progress.

lib/network/default.nix
```diff
@@ -0,0 +1,49 @@
...
+let
+  inherit (import ./internal.nix { inherit lib; }) _ipv6;
+in
+{
+  ipv6 = {
+    /**
+      Creates an `IPv6Address` object from an IPv6 address as a string. If
+      the prefix length is omitted, it defaults to 64. The parser is limited
+      to the first two versions of IPv6 addresses addressed in RFC 4291.
+      The form "x:x:x:x:x:x:d.d.d.d" is not yet implemented. Addresses are
```
Janik-Haag, line 11: >       The form "x:x:x:x:x:x:d.d.d.d" is not yet implemented.

We probably want to be more explicit and say The form... won't be implemented.

Janik-Haag (commented):
@infinisil as a lib code owner, do you want to give this a review before we merge it?

Janik-Haag (approved):

.gitignore
```diff
@@ -7,6 +7,7 @@
 .idea/
 .nixos-test-history
 .vscode/
+.helix/
```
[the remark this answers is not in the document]
drupol: I would add that this shouldn't be part of this PR and added into the user global's `.gitignore` file.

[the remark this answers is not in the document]
Janik-Haag: I think it's reasonable to have this, esp. since we also have it for other editors like .idea, .vscode, etc... but yes I agree it would have been more transparent if we moved it into a separate pr.

PR #290104, opened 2024-02-20
nixos/seafile: add persistent user, configurable storage path, and gc service

melvyn2:
## Description of changes
Disable `DynamicUser` for seafile in favor of a persistent configurable service user and group. This allows moving the storage directory out of the systemd-managed `StateDirectory` (previously, external paths could be bind-mounted or symlinked into the state directory, but UID reassignments meant having to manually and slowly `chown` the entire storage). 

This also allows access by other services, such as by a garbage-collection timer also added here.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
- For non-Linux: Is sandboxing enabled in `nix.conf`? (See [Nix manual](https://nixos.org/manual/nix/stable/command-ref/conf-file.html))
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [x] Tested, as applicable:
  - [NixOS test(s)](https://nixos.org/manual/nixos/unstable/index.html#sec-nixos-tests) (look inside [nixos/tests](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests))
  - and/or [package tests](https://nixos.org/manual/nixpkgs/unstable/#sec-package-tests)
  - or, for functions and "core" functionality, tests in [lib/tests](https://github.com/NixOS/nixpkgs/blob/master/lib/tests) or [pkgs/test](https://github.com/NixOS/nixpkgs/blob/master/pkgs/test)
  - made sure NixOS tests are [linked](https://nixos.org/manual/nixpkgs/unstable/#ssec-nixos-tests-linking) to the relevant packages
- [x] Tested compilation of all packages that depend on this change using `nix-shell -p nixpkgs-review --run "nixpkgs-review rev HEAD"`. Note: all changes have to be committed, also see [nixpkgs-review usage](https://github.com/Mic92/nixpkgs-review#usage)
- [x] Tested basic functionality of all binary files (usually in `./result/bin/`)
- [24.05 Release Notes](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2405.section.md) (or backporting [23.05](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2305.section.md) and [23.11](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2311.section.md) Release notes)
[...]

PR #313628, opened 2024-05-22
python3Packages: format with nixfmt

mweinelt:
## Description of changes

People have started reformatting individual packages with nixfmt, so a treewide change should get rid of that review churn.

Also, we're at an ideal point to sneak in a reformat, when we consider backports from master to release.

If this causes rebuilds, we'll need to identify those and postpone.

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
- For non-Linux: Is sandboxing enabled in `nix.conf`? (See [Nix manual](https://nixos.org/manual/nix/stable/command-ref/conf-file.html))
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:
  - [NixOS test(s)](https://nixos.org/manual/nixos/unstable/index.html#sec-nixos-tests) (look inside [nixos/tests](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests))
  - and/or [package tests](https://github.com/NixOS/nixpkgs/blob/master/pkgs/README.md#package-tests)
  - or, for functions and "core" functionality, tests in [lib/tests](https://github.com/NixOS/nixpkgs/blob/master/lib/tests) or [pkgs/test](https://github.com/NixOS/nixpkgs/blob/master/pkgs/test)
  - made sure NixOS tests are [linked](https://nixos.org/manual/nixpkgs/unstable/#ssec-nixos-tests-linking) to the relevant packages
- [ ] Tested compilation of all packages that depend on this change using `nix-shell -p nixpkgs-review --run "nixpkgs-review rev HEAD"`. Note: all changes have to be committed, also see [nixpkgs-review usage](https://github.com/Mic92/nixpkgs-review#usage)
- [ ] Tested basic functionality of all binary files (usually in `./result/bin/`)
- [24.05 Release Notes](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2405.section.md) (or backporting [23.05](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2305.section.md) and [23.11](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2311.section.md) Release notes)
  - [ ] (Package updates) Added a release notes entry if the change is major or breaking
  - [ ] (Module updates) Added a release notes entry if the change is significant
[...]

PR #318727, opened 2024-06-10
Seafile: 10.0.1 -> 11.0.12

### LINKED ISSUE #258719: Update request: seahub 9.0.10 → 11.0.0
- Package name: seahub
- Latest released version: 11.0.0

- Current version on the unstable channel: 9.0.10
- Current version on the stable/release channel: 9.0.10

The package is broken on nixos-unstable because Django 4.2 removed the `providing_args` argument for `django.dispatch.Signal`. <https://docs.djangoproject.com/en/4.2/internals/deprecation/#deprecation-removed-in-4-0>.

The package failed in runtime instead of build time because the package sets `dontBuild = true`.

The issue was fixed in upstream <https://github.com/haiwen/seahub/pull/5502>. And the fix is already released in `v11.0.0-pro`.

Log:

```log
Oct 03 04:19:16 mtl0 systemd[1]: Starting Seafile Server Web Frontend...
Oct 03 04:19:16 mtl0 systemd[1]: Started Seafile Server Web Frontend.
Oct 03 04:19:16 mtl0 gunicorn[8928]: Traceback (most recent call last):
Oct 03 04:19:16 mtl0 gunicorn[8928]:   File "/nix/store/6d2hnh4p82r0w3zhslxnr019k7fl8rnx-python3.10-gunicorn-21.2.0/bin/.gunicorn-wrapped", line 9, in <module>
Oct 03 04:19:16 mtl0 gunicorn[8928]:     sys.exit(run())
Oct 03 04:19:16 mtl0 gunicorn[8928]:   File "/nix/store/6d2hnh4p82r0w3zhslxnr019k7fl8rnx-python3.10-gunicorn-21.2.0/lib/python3.10/site-packages/gunicorn/app/wsgiapp.py", line 67, in run
Oct 03 04:19:16 mtl0 gunicorn[8928]:     WSGIApplication("%(prog)s [OPTIONS] [APP_MODULE]").run()
Oct 03 04:19:16 mtl0 gunicorn[8928]:   File "/nix/store/6d2hnh4p82r0w3zhslxnr019k7fl8rnx-python3.10-gunicorn-21.2.0/lib/python3.10/site-packages/gunicorn/app/base.py", line 236, in run
Oct 03 04:19:16 mtl0 gunicorn[8928]:     super().run()
Oct 03 04:19:16 mtl0 gunicorn[8928]:   File "/nix/store/6d2hnh4p82r0w3zhslxnr019k7fl8rnx-python3.10-gunicorn-21.2.0/lib/python3.10/site-packages/gunicorn/app/base.py", line 72, in run
Oct 03 04:19:16 mtl0 gunicorn[8928]:     Arbiter(self).run()
Oct 03 04:19:16 mtl0 gunicorn[8928]:   File "/nix/store/6d2hnh4p82r0w3zhslxnr019k7fl8rnx-python3.10-gunicorn-21.2.0/lib/python3.10/site-packages/gunicorn/arbiter.py", line 58, in __init__
Oct 03 04:19:16 mtl0 gunicorn[8928]:     self.setup(app)
Oct 03 04:19:16 mtl0 gunicorn[8928]:   File "/nix/store/6d2hnh4p82r0w3zhslxnr019k7fl8rnx-python3.10-gunicorn-21.2.0/lib/python3.10/site-packages/gunicorn/arbiter.py", line 118, in setup
Oct 03 04:19:16 mtl0 gunicorn[8928]:     self.app.wsgi()
Oct 03 04:19:16 mtl0 gunicorn[8928]:   File "/nix/store/6d2hnh4p82r0w3zhslxnr019k7fl8rnx-python3.10-gunicorn-21.2.0/lib/python3.10/site-packages/gunicorn/app/base.py", line 67, in wsgi
Oct 03 04:19:16 mtl0 gunicorn[8928]:     self.callable = self.load()
Oct 03 04:19:16 mtl0 gunicorn[8928]:   File "/nix/store/6d2hnh4p82r0w3zhslxnr019k7fl8rnx-python3.10-gunicorn-21.2.0/lib/python3.10/site-packages/gunicorn/app/wsgiapp.py", line 58, in load
Oct 03 04:19:16 mtl0 gunicorn[8928]:     return self.load_wsgiapp()
Oct 03 04:19:16 mtl0 gunicorn[8928]:   File "/nix/store/6d2hnh4p82r0w3zhslxnr019k7fl8rnx-python3.10-gunicorn-21.2.0/lib/python3.10/site-packages/gunicorn/app/wsgiapp.py", line 48, in load_wsgiapp
Oct 03 04:19:16 mtl0 gunicorn[8928]:     return util.import_app(self.app_uri)
Oct 03 04:19:16 mtl0 gunicorn[8928]:   File "/nix/store/6d2hnh4p82r0w3zhslxnr019k7fl8rnx-python3.10-gunicorn-21.2.0/lib/python3.10/site-packages/gunicorn/util.py", line 371, in import_app
Oct 03 04:19:16 mtl0 gunicorn[8928]:     mod = importlib.import_module(module)
Oct 03 04:19:16 mtl0 gunicorn[8928]:   File "/nix/store/pzf6dnxg8gf04xazzjdwarm7s03cbrgz-python3-3.10.12/lib/python3.10/importlib/__init__.py", line 126, in import_module
Oct 03 04:19:16 mtl0 gunicorn[8928]:     return _bootstrap._gcd_import(name[level:], package, level)
Oct 03 04:19:16 mtl0 gunicorn[8928]:   File "<frozen importlib._bootstrap>", line 1050, in _gcd_import
Oct 03 04:19:16 mtl0 gunicorn[8928]:   File "<frozen importlib._bootstrap>", line 1027, in _find_and_load
Oct 03 04:19:16 mtl0 gunicorn[8928]:   File "<frozen importlib._bootstrap>", line 992, in _find_and_load_unlocked
Oct 03 04:19:16 mtl0 gunicorn[8928]:   File "<frozen importlib._bootstrap>", line 241, in _call_with_frames_removed
Oct 03 04:19:16 mtl0 gunicorn[8928]:   File "<frozen importlib._bootstrap>", line 1050, in _gcd_import
Oct 03 04:19:16 mtl0 gunicorn[8928]:   File "<frozen importlib._bootstrap>", line 1027, in _find_and_load
Oct 03 04:19:16 mtl0 gunicorn[8928]:   File "<frozen importlib._bootstrap>", line 1006, in _find_and_load_unlocked
Oct 03 04:19:16 mtl0 gunicorn[8928]:   File "<frozen importlib._bootstrap>", line 688, in _load_unlocked
Oct 03 04:19:16 mtl0 gunicorn[8928]:   File "<frozen importlib._bootstrap_external>", line 883, in exec_module
Oct 03 04:19:16 mtl0 gunicorn[8928]:   File "<frozen importlib._bootstrap>", line 241, in _call_with_frames_removed
Oct 03 04:19:16 mtl0 gunicorn[8928]:   File "/nix/store/08dwi9sf17j33dnm1gawl655nfw9rli5-seahub-9.0.10/seahub/__init__.py", line 2, in <module>
Oct 03 04:19:16 mtl0 gunicorn[8928]:     from .signals import repo_created, repo_deleted, clean_up_repo_trash, \
Oct 03 04:19:16 mtl0 gunicorn[8928]:   File "/nix/store/08dwi9sf17j33dnm1gawl655nfw9rli5-seahub-9.0.10/seahub/signals.py", line 5, in <module>
Oct 03 04:19:16 mtl0 gunicorn[8928]:     repo_created = django.dispatch.Signal(providing_args=["org_id", "creator", "repo_id", "repo_name", "library_template"])
Oct 03 04:19:16 mtl0 gunicorn[8928]: TypeError: Signal.__init__() got an unexpected keyword argument 'providing_args'
Oct 03 04:19:16 mtl0 systemd[1]: seahub.service: Main process exited, code=exited, status=1/FAILURE
Oct 03 04:19:16 mtl0 systemd[1]: seahub.service: Failed with result 'exit-code'.
```

**Notify maintainers**

@greizgh @schmittlauch

-----

melvyn2:
## Description of changes
Updates seafile-server and seahub to 11.0.9. This requires adding the python package djangosaml2. ~~Finally, cleans up the seafile module a little to ensure the proper, matching packages are used.~~

This required adding a migration script as the sqlite backend was deprecated in favor of MySQL as of version 11. To make this change easier, the persistent user setup of #290104 was merged in.

Finally, a change to the CSRF setup means some users might need to manually whitelist their domains, like so:
```nix
  services.seafile = {
    ...

   seahubExtraConf = ''
     CSRF_TRUSTED_ORIGINS = ["https://example.com"]
   '';
  };
```

This closes #258719

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [x] Tested, as applicable:

---

### DIFF
--- a/nixos/doc/manual/release-notes/rl-2411.section.md
+++ b/nixos/doc/manual/release-notes/rl-2411.section.md
@@ -529,6 +529,16 @@
 
 - The kubelet configuration file can now be amended with arbitrary additional content using the `services.kubernetes.kubelet.extraConfig` option.
 
+- The `services.seafile` module was updated to major version 11.
+  - As part of this upgrade, the database backend will be migrated to MySQL.
+    This process should be automatic, but in case of a botched migration,
+    old sqlite files are not removed and can be used to manually migrate the database.
+  - Additionally, the updated CSRF protection may prevent some users from logging in.
+    Specific origin addresses can be whitelisted using the `services.seafile.seahubExtraConf` option
+    (e.g. `services.seafile.seahubExtraConf = ''CSRF_TRUSTED_ORIGINS = ["https://example.com"]'';`).
+    Note that first solution of the [official FAQ answer](https://cloud.seatable.io/dtable/external-links/7b976c85f504491cbe8e/?tid=0000&vid=0000&row-id=BQhH-2HSQs68Nq2EW91DBA)
+    is not allowed by the `services.nginx` module's config-checker.
+
 - To facilitate dependency injection, the `imgui` package now builds a static archive using vcpkg' CMake rules.
   The derivation now installs "impl" headers selectively instead of by a wildcard.
   Use `imgui.src` if you just want to access the unpacked sources.
--- a/nixos/modules/services/networking/seafile.nix
+++ b/nixos/modules/services/networking/seafile.nix
@@ -29,8 +29,8 @@ let
 
   seafRoot = "/var/lib/seafile"; # hardcode it due to dynamicuser
   ccnetDir = "${seafRoot}/ccnet";
-  dataDir = "${seafRoot}/data";
   seahubDir = "${seafRoot}/seahub";
+  defaultUser = "seafile";
 
 in
 {
@@ -118,11 +118,39 @@ in
       type = types.str;
       description = ''
         Seafile Seahub Admin Account initial password.
-        Should be change via Seahub web front-end.
+        Should be changed via Seahub web front-end.
       '';
     };
 
     seafilePackage = mkPackageOption pkgs "seafile-server" { };
+
+    user = mkOption {
+      type = types.str;
+      default = defaultUser;
+      description = "User account under which seafile runs.";
+    };
+
+    group = mkOption {
+      type = types.str;
+      default = defaultUser;
+      description = "Group under which seafile runs.";
+    };
+
+    dataDir = mkOption {
+      type = types.nullOr types.path;
+      default = null;
+      defaultText = "${seafRoot}/data";
+      description = "Path in which to store user data";
+    };
+
+    gcDates = mkOption {
+      type = types.listOf types.str;
+      default = [ ];
+      example = [ "Sun 03:00:00" ];
+      description = ''
+        When to run garbage collection on stored data blocks.
+      '';
+    };
 
     seahubExtraConf = mkOption {
       default = "";
@@ -143,6 +171,17 @@ in
     environment.etc."seafile/seafile.conf".source = seafileConf;
     environment.etc."seafile/seahub_settings.py".source = seahubSettings;
 
+    users.users = lib.optionalAttrs (cfg.user == defaultUser) {
+      "${defaultUser}" = {
+        group = cfg.group;
+        isSystemUser = true;
+      };
+    };
+
+    users.groups = lib.optionalAttrs (cfg.group == defaultUser) {
+      "${defaultUser}" = {};
+    };
+
     systemd.targets.seafile = {
       wantedBy = [ "multi-user.target" ];
       description = "Seafile components";
@@ -154,6 +193,8 @@ in
           ProtectHome = true;
           PrivateUsers = true;
           PrivateDevices = true;
+          PrivateTmp = true;
+          ProtectSystem = "strict";
           ProtectClock = true;
           ProtectHostname = true;
           ProtectProc = "invisible";
@@ -162,30 +203,35 @@ in
           ProtectKernelLogs = true;
           ProtectControlGroups = true;
           RestrictNamespaces = true;
+          RemoveIPC = true;
           LockPersonality = true;
           RestrictRealtime = true;
           RestrictSUIDSGID = true;
+          NoNewPrivileges = true;
           MemoryDenyWriteExecute = true;
           SystemCallArchitectures = "native";
           RestrictAddressFamilies = [ "AF_UNIX" "AF_INET" ];
         };
+        dataDir = if (!builtins.isNull cfg.dataDir) then cfg.dataDir else "${seafRoot}/data";
+        optionalDataDir = lib.lists.optional (!builtins.isNull cfg.dataDir) cfg.dataDir;
       in
       {
         seaf-server = {
           description = "Seafile server";
           partOf = [ "seafile.target" ];
           after = [ "network.target" ];
+          unitConfig.RequiresMountsFor = optionalDataDir;
           wantedBy = [ "seafile.target" ];
           restartTriggers = [ ccnetConf seafileConf ];
           path = [ pkgs.sqlite ];
           serviceConfig = securityOptions // {
-            User = "seafile";
-            Group = "seafile";
-            DynamicUser = true;
+            User = cfg.user;
+            Group = cfg.group;
             StateDirectory = "seafile";
             RuntimeDirectory = "seafile";
             LogsDirectory = "seafile";
             ConfigurationDirectory = "seafile";
+            ReadWritePaths = optionalDataDir;
             ExecStart = ''
               ${cfg.seafilePackage}/bin/seaf-server \
               --foreground \
@@ -237,6 +283,7 @@ in
           partOf = [ "seafile.target" ];
           after = [ "network.target" "seaf-server.service" ];
           requires = [ "seaf-server.service" ];
+          unitConfig.RequiresMountsFor = optionalDataDir;
           restartTriggers = [ seahubSettings ];
           environment = {
             PYTHONPATH = "${pkgs.seahub.pythonPath}:${pkgs.seahub}/thirdpart:${pkgs.seahub}";
@@ -248,13 +295,13 @@ in
             SEAHUB_LOG_DIR = "/var/log/seafile";
           };
           serviceConfig = securityOptions // {
-            User = "seafile";
-            Group = "seafile";
-            DynamicUser = true;
+            User = cfg.user;
+            Group = cfg.group;
             RuntimeDirectory = "seahub";
             StateDirectory = "seafile";
             LogsDirectory = "seafile";
             ConfigurationDirectory = "seafile";
+            ReadWritePaths = optionalDataDir;
             ExecStart = ''
               ${pkgs.seahub.python.pkgs.gunicorn}/bin/gunicorn seahub.wsgi:application \
               --name seahub \
@@ -293,6 +340,52 @@ in
             fi
           '';
         };
+
+        seaf-gc = {
+          description = "Seafile storage garbage collection";
+          conflicts = [ "seaf-server.service" "seahub.service" ];
+          after = [ "seaf-server.service" "seahub.service" ];
+          unitConfig.RequiresMountsFor = optionalDataDir;
+          onSuccess = [ "seaf-server.service" "seahub.service" ];
+          onFailure = [ "seaf-server.service" "seahub.service" ];
+          startAt = cfg.gcDates;
+          serviceConfig = securityOptions // {
+            User = cfg.user;
+            Group = cfg.group;
+            StateDirectory = "seafile";
+            RuntimeDirectory = "seafile";
+            LogsDirectory = "seafile";
+            ConfigurationDirectory = "seafile";
+            ReadWritePaths = optionalDataDir;
+            Type = "oneshot";
+          };
+          script = ''
+            if [ ! -f "${seafRoot}/server-setup" ]; then
+                echo "Server not setup yet, GC not needed" >&2
+                exit
+            fi
+
+            # checking for pending upgrades
+            installedMajor=$(cat "${seafRoot}/server-setup" | cut -d"-" -f1 | cut -d"." -f1)
+            installedMinor=$(cat "${seafRoot}/server-setup" | cut -d"-" -f1 | cut -d"." -f2)
+            pkgMajor=$(echo "${cfg.seafilePackage.version}" | cut -d"." -f1)
+            pkgMinor=$(echo "${cfg.seafilePackage.version}" | cut -d"." -f2)
+
+            if [[ $installedMajor != $pkgMajor || $installedMinor != $pkgMinor ]]; then
+                echo "Server not upgraded yet" >&2
+                exit
+            fi
+
+            # Clean up user-deleted blocks and libraries
+            ${cfg.seafilePackage}/bin/seafserv-gc \
+              -F /etc/seafile \
+              -c ${ccnetDir} \
+              -d ${dataDir} \
+              --rm-fs
+          '';
+        };
       };
   };
+
+  meta.maintainers = with maintainers; [ greizgh schmittlauch ];
 }
--- a/nixos/modules/services/networking/seafile.nix
+++ b/nixos/modules/services/networking/seafile.nix
@@ -1,19 +1,42 @@
-{ config, lib, pkgs, ... }:
-with lib;
+{
+  config,
+  lib,
+  pkgs,
+  ...
+}:
 let
   cfg = config.services.seafile;
   settingsFormat = pkgs.formats.ini { };
 
-  ccnetConf = settingsFormat.generate "ccnet.conf" cfg.ccnetSettings;
+  ccnetConf = settingsFormat.generate "ccnet.conf" (
+    lib.attrsets.recursiveUpdate {
+      Database = {
+        ENGINE = "mysql";
+        UNIX_SOCKET = "/var/run/mysqld/mysqld.sock";
+        DB = "ccnet_db";
+        CONNECTION_CHARSET = "utf8";
+      };
+    } cfg.ccnetSettings
+  );
 
-  seafileConf = settingsFormat.generate "seafile.conf" cfg.seafileSettings;
+  seafileConf = settingsFormat.generate "seafile.conf" (
+    lib.attrsets.recursiveUpdate {
+      database = {
+        type = "mysql";
+        unix_socket = "/var/run/mysqld/mysqld.sock";
+        db_name = "seafile_db";
+        connection_charset = "utf8";
+      };
+    } cfg.seafileSettings
+  );
 
   seahubSettings = pkgs.writeText "seahub_settings.py" ''
     FILE_SERVER_ROOT = '${cfg.ccnetSettings.General.SERVICE_URL}/seafhttp'
     DATABASES = {
         'default': {
-            'ENGINE': 'django.db.backends.sqlite3',
-            'NAME': '${seahubDir}/seahub.db',
+            'ENGINE': 'django.db.backends.mysql',
+            'NAME' : 'seahub_db',
+            'HOST' : '/var/run/mysqld/mysqld.sock',
         }
     }
     MEDIA_ROOT = '${seahubDir}/media/'
@@ -21,13 +44,15 @@ let
 
     SERVICE_URL = '${cfg.ccnetSettings.General.SERVICE_URL}'
 
+    CSRF_TRUSTED_ORIGINS = ["${cfg.ccnetSettings.General.SERVICE_URL}"]
+
     with open('${seafRoot}/.seahubSecret') as f:
         SECRET_KEY = f.readline().rstrip()
 
     ${cfg.seahubExtraConf}
   '';
 
-  seafRoot = "/var/lib/seafile"; # hardcode it due to dynamicuser
+  seafRoot = "/var/lib/seafile";
   ccnetDir = "${seafRoot}/ccnet";
   seahubDir = "${seafRoot}/seahub";
   defaultUser = "seafile";
@@ -37,7 +62,7 @@ in
 
   ###### Interface
 
-  options.services.seafile = {
+  options.services.seafile = with lib; {
     enable = mkEnableOption "Seafile server";
 
     ccnetSettings = mkOption {
@@ -47,7 +72,7 @@ in
         options = {
           General = {
             SERVICE_URL = mkOption {
-              type = types.str;
+              type = types.singleLineStr;
               example = "https://www.example.com";
               description = ''
                 Seahub public URL.
@@ -78,11 +103,17 @@ in
               '';
             };
             host = mkOption {
-              type = types.str;
-              default = "127.0.0.1";
-              example = "0.0.0.0";
+              type = types.singleLineStr;
+              default = "ipv4:127.0.0.1";
+              example = "unix:/run/seafile/server.sock";
               description = ''
-                The binding address used by seafile fileserver.
+                The bind address used by seafile fileserver.
+
+                The addr can be defined as one of the following:
+                - ipv6:<ipv6addr> for binding to an IPv6 address.
+                - unix:<named pipe> for binding to a unix named socket
+                - ipv4:<ipv4addr> for binding to an ipv4 address
+                Otherwise the addr is assumed to be ipv4.
               '';
             };
           };
@@ -93,6 +124,19 @@ in
         Configuration for seafile-server, see
         <https://manual.seafile.com/config/seafile-conf/>
         for supported values.
+      '';
+    };
+
+    seahubAddress = mkOption {
+      type = types.singleLineStr;
+      default = "unix:/run/seahub/gunicorn.sock";
+      example = "[::1]:8083";
+      description = ''
+        Which address to bind the seahub server to, of the form:
+        - HOST
+        - HOST:PORT
+        - unix:PATH.
+        IPv6 HOSTs must be wrapped in brackets.
       '';
     };
 
@@ -107,7 +151,7 @@ in
 
     adminEmail = mkOption {
       example = "john@example.com";
-      type = types.str;
+      type = types.singleLineStr;
       description = ''
         Seafile Seahub Admin Account Email.
       '';
@@ -115,45 +159,79 @@ in
 
     initialAdminPassword = mkOption {
       example = "someStrongPass";
-      type = types.str;
+      type = types.singleLineStr;
       description = ''
         Seafile Seahub Admin Account initial password.
         Should be changed via Seahub web front-end.
       '';
     };
 
-    seafilePackage = mkPackageOption pkgs "seafile-server" { };
+    seahubPackage = mkPackageOption pkgs "seahub" { };
 
     user = mkOption {
-      type = types.str;
+      type = types.singleLineStr;
       default = defaultUser;
       description = "User account under which seafile runs.";
     };
 
     group = mkOption {
-      type = types.str;
+      type = types.singleLineStr;
       default = defaultUser;
       description = "Group under which seafile runs.";
     };
 
     dataDir = mkOption {
-      type = types.nullOr types.path;
-      default = null;
-      defaultText = "${seafRoot}/data";
+      type = types.path;
+      default = "${seafRoot}/data";
       description = "Path in which to store user data";
     };
 
-    gcDates = mkOption {
-      type = types.listOf types.str;
-      default = [ ];
-      example = [ "Sun 03:00:00" ];
-      description = ''
-        When to run garbage collection on stored data blocks.
-      '';
+    gc = {
+      enable = mkEnableOption "automatic garbage collection on stored data blocks";
+
+      dates = mkOption {
+        type = types.listOf types.singleLineStr;
+        default = [ "Sun 03:00:00" ];
+        description = ''
+          When to run garbage collection on stored data blocks.
+          The time format is described in {manpage}`systemd.time(7)`.
+        '';
+      };
+
+      randomizedDelaySec = mkOption {
+        default = "0";
+        type = types.singleLineStr;
+        example = "45min";
+        description = ''
+          Add a randomized delay before each garbage collection.
+          The delay will be chosen between zero and this value.
+          This value must be a time span in the format specified by
+          {manpage}`systemd.time(7)`
+        '';
+      };
+
+      persistent = mkOption {
+        default = true;
+        type = types.bool;
+        example = false;
+        description = ''
+          Takes a boolean argument. If true, the time when the service
+          unit was last triggered is stored on disk. When the timer is
+          activated, the service unit is triggered immediately if it
+          would have been triggered at least once during the time when
+          the timer was inactive. Such triggering is nonetheless
+          subject to the delay imposed by RandomizedDelaySec=. This is
+          useful to catch up on missed runs of the service when the
+          system was powered down.
+        '';
+      };
     };
 
     seahubExtraConf = mkOption {
       default = "";
+      example = ''
+        CSRF_TRUSTED_ORIGINS = ["https://example.com"]
+      '';
       type = types.lines;
       description = ''
         Extra config to append to `seahub_settings.py` file.
@@ -165,7 +243,26 @@ in
 
   ###### Implementation
 
-  config = mkIf cfg.enable {
+  config = lib.mkIf cfg.enable {
+    services.mysql = {
+      enable = true;
+      package = lib.mkDefault pkgs.mariadb;
+      ensureDatabases = [
+        "ccnet_db"
+        "seafile_db"
+        "seahub_db"
+      ];
+      ensureUsers = [
+        {
+          name = cfg.user;
+          ensurePermissions = {
+            "ccnet_db.*" = "ALL PRIVILEGES";
+            "seafile_db.*" = "ALL PRIVILEGES";
+            "seahub_db.*" = "ALL PRIVILEGES";
+          };
+        }
+      ];
+    };
 
     environment.etc."seafile/ccnet.conf".source = ccnetConf;
     environment.etc."seafile/seafile.conf".source = seafileConf;
@@ -178,9 +275,7 @@ in
       };
     };
 
-    users.groups = lib.optionalAttrs (cfg.group == defaultUser) {
-      "${defaultUser}" = {};
-    };
+    users.groups = lib.optionalAttrs (cfg.group == defaultUser) { "${defaultUser}" = { }; };
 
     systemd.targets.seafile = {
       wantedBy = [ "multi-user.target" ];
@@ -189,7 +284,7 @@ in
 
     systemd.services =
       let
-        securityOptions = {
+        serviceOptions = {
           ProtectHome = true;
           PrivateUsers = true;
           PrivateDevices = true;
@@ -210,34 +305,42 @@ in
           NoNewPrivileges = true;
           MemoryDenyWriteExecute = true;
           SystemCallArchitectures = "native";
-          RestrictAddressFamilies = [ "AF_UNIX" "AF_INET" ];
+          RestrictAddressFamilies = [
+            "AF_UNIX"
+            "AF_INET"
+          ];
+
+          User = cfg.user;
+          Group = cfg.group;
+          StateDirectory = "seafile";
+          RuntimeDirectory = "seafile";
+          LogsDirectory = "seafile";
+          ConfigurationDirectory = "seafile";
+          ReadWritePaths = lib.lists.optional (cfg.dataDir != "${seafRoot}/data") cfg.dataDir;
         };
-        dataDir = if (!builtins.isNull cfg.dataDir) then cfg.dataDir else "${seafRoot}/data";
-        optionalDataDir = lib.lists.optional (!builtins.isNull cfg.dataDir) cfg.dataDir;
       in
       {
         seaf-server = {
           description = "Seafile server";
           partOf = [ "seafile.target" ];
-          after = [ "network.target" ];
-          unitConfig.RequiresMountsFor = optionalDataDir;
+          unitConfig.RequiresMountsFor = lib.lists.optional (cfg.dataDir != "${seafRoot}/data") cfg.dataDir;
+          requires = [ "mysql.service" ];
+          after = [
+            "network.target"
+            "mysql.service"
+          ];
           wantedBy = [ "seafile.target" ];
-          restartTriggers = [ ccnetConf seafileConf ];
-          path = [ pkgs.sqlite ];
-          serviceConfig = securityOptions // {
-            User = cfg.user;
-            Group = cfg.group;
-            StateDirectory = "seafile";
-            RuntimeDirectory = "seafile";
-            LogsDirectory = "seafile";
-            ConfigurationDirectory = "seafile";
-            ReadWritePaths = optionalDataDir;
+          restartTriggers = [
+            ccnetConf
+            seafileConf
+          ];
+          serviceConfig = serviceOptions // {
             ExecStart = ''
-              ${cfg.seafilePackage}/bin/seaf-server \
+              ${lib.getExe cfg.seahubPackage.seafile-server} \
               --foreground \
               -F /etc/seafile \
               -c ${ccnetDir} \
-              -d ${dataDir} \
+              -d ${cfg.dataDir} \
               -l /var/log/seafile/server.log \
               -P /run/seafile/server.pid \
               -p /run/seafile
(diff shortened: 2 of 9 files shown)


melvyn2:
Tests for seafile are failing on master but worked on nixos-24.05; I'll take a look when I can.

Also, the update to 11.0 requires changing a reverse proxy setting that nixos doesn't allow. From https://cloud.seatable.io/dtable/external-links/7b976c85f504491cbe8e/?tid=0000&vid=0000&row-id=BQhH-2HSQs68Nq2EW91DBA, the new django CSRF protection requires the proxy server to pass the Host header untouched, but this is blocked by https://github.com/yandex/gixy/blob/master/docs/en/plugins/hostspoofing.md. This means that either nginx config validation has to be entirely disabled, or CSRF protection has to be disabled.
Neither are ideal solutions; maybe the validation should be a little more configurable, so that skip flags can be passed to gixy.

melvyn2:
Migration from 10.0 to 11.0 seem to be working properly now, and tests are passing both on 24.05 and unstable. The reverse proxy issue remains, but this is technically not caused by this module, so I'll undraft. A release note line may be necessary.

greizgh:
Thank you @melvyn2 for looking into this. I've been put off by the deprecation of sqlite (see [upstream changelog](https://manual.seafile.com/upgrade/upgrade_notes_for_11.0.x/#dropped-sqlite-database-support)). It seems to work, but I wouldn't assume this will be the case for long. Are you aware of a decision change?
I see you setup seahub DB from the sql scripts, I've relied on django migrations because it felt weird to have both mechanisms in place. But I guess this is closer to what upstream does.
Regarding CSRF protection, disabling it is not an option (at least to me), but the `CSRF_TRUSTED_ORIGINS` approach in conf is quite common (and recommended by upstream). Maybe we should expose this through the module (alternatively, this can already be set through `extraConfig`)?

melvyn2:
> I've been put off by the deprecation of sqlite (see [upstream changelog](https://manual.seafile.com/upgrade/upgrade_notes_for_11.0.x/#dropped-sqlite-database-support)). It seems to work, but I wouldn't assume this will be the case for long. Are you aware of a decision change?

No, I just missed that in the changelog somehow :facepalm:
I'll re-draft this and see if I can get mysql working in a decent amount of time.

>  I see you setup seahub DB from the sql scripts, I've relied on django migrations because it felt weird to have both mechanisms in place. But I guess this is closer to what upstream does.

I actually undid this, I originally thought the SQL script was required because during my first attempted upgrade the migration silently failed. The django migrations do work normally and are simpler to use so I've reverted to only using them.

>  Regarding CSRF protection, disabling it is not an option (at least to me), but the CSRF_TRUSTED_ORIGINS approach in conf is quite common (and recommended by upstream). Maybe we should expose this through the module (alternatively, this can already be set through extraConfig)?

By "disabling" CSRF I meant simply adding the domain to CSRF_TRUSTED_ORIGINS, because I saw this as "bypassing" the protection for that specific origin - it was just bad phrasing from me. Exposing the setting in the module seems like a good choice considering how often this will be required for users to set.

melvyn2:
The MariaDB-based seafile module is passing tests, and migrations from sqlite/10.0 are _almost_ working. Last error is https://manual.seafile.com/deploy/migrate_from_sqlite_to_mysql/#encountered-errno-150-foreign-key-constraint-is-incorrectly-formed,
which seems to require manually reordering the output of the sqlite dump...
At worst I can solve this with a nasty find-and-replace but I'm looking for a better solution in the mean time.

melvyn2:
Of course as soon as I post that, I find https://mariadb.com/kb/en/server-system-variables/#foreign_key_checks ...
Will try it next time

melvyn2:
MariaDB migration is now working in a test environment, and using the ccnet service URL as a default URL for the CSRF whitelist seems like a decent solution for the CSRF issue.
I'll try the migration on my own setup now and see how it goes!

melvyn2:
Works on my end, after these little changes!

pkgs/development/python-modules/djangosaml2/default.nix
```diff
@@ -0,0 +1,38 @@
+{ lib, buildPythonPackage, fetchFromGitHub, defusedxml, django, pysaml2
+, pythonOlder, }:
+
+buildPythonPackage rec {
+  pname = "djangosaml2";
+  version = "1.9.3";
+  format = "pyproject";
```
getchoo, line 7:
```suggestion
  pyproject = true;
```

pkgs/development/python-modules/djangosaml2/default.nix
```diff
@@ -0,0 +1,38 @@
...
+  disabled = pythonOlder "3.8";
+
+  src = fetchFromGitHub {
+    owner = "IdentityPython";
+    repo = "djangosaml2";
+    rev = "refs/tags/v${version}";
+    hash = "sha256-rbmEJuG2mgozpCFOXZUJFxv8v52IRQeaeAKfeUDACeU=";
+  };
+
+  propagatedBuildInputs = [ django defusedxml pysaml2 ];
```
getchoo, line 18:
```suggestion
  dependencies = [ django defusedxml pysaml2 ];
```

pkgs/development/python-modules/djangosaml2/default.nix
```diff
@@ -0,0 +1,38 @@
...
+
+  checkPhase = ''
+    runHook preCheck
+
+    python tests/run_tests.py
+
+    runHook postCheck
+  '';
+
+  meta = with lib; {
```
getchoo, line 30:
```suggestion
  meta = {
```
https://github.com/NixOS/nixpkgs/issues/208242

pkgs/development/python-modules/djangosaml2/default.nix
```diff
@@ -0,0 +1,38 @@
...
+    description = "Django SAML2 Service Provider based on pySAML2";
+    homepage = "https://github.com/IdentityPython/djangosaml2";
+    changelog =
+      "https://github.com/IdentityPython/djangosaml2/blob/v${version}/CHANGES";
+    license = licenses.asl20;
```
getchoo, line 35:
```suggestion
    license = lib.licenses.asl20;
```

pkgs/development/python-modules/djangosaml2/default.nix
```diff
@@ -0,0 +1,38 @@
...
+    maintainers = with maintainers; [ ];
```
getchoo, line 36:
```suggestion
    maintainers = with lib.maintainers; [ ];
```

nixos/modules/services/networking/seafile.nix
```diff
@@ -4,33 +4,50 @@ let
...
 
-  ccnetConf = settingsFormat.generate "ccnet.conf" cfg.ccnetSettings;
+  ccnetConf = settingsFormat.generate "ccnet.conf" (cfg.ccnetSettings // {
+    Database = {
+      ENGINE = "mysql";
+      UNIX_SOCKET = "/var/run/mysqld/mysqld.sock";
+      DB = "ccnet_db";
+      CONNECTION_CHARSET = "utf8";
+    };
+  });
```
getchoo, line 14: wouldn't this be better as a default for the `ccnetSettings` submodule? having this basically forced internally could be pretty annoying if one wanted to run this with a remote database for example. maybe these options could be made explicit to better handle service settings for cases like that as well

nixos/modules/services/networking/seafile.nix
```diff
@@ -4,33 +4,50 @@ let
...
 
-  seafileConf = settingsFormat.generate "seafile.conf" cfg.seafileSettings;
+  seafileConf = settingsFormat.generate "seafile.conf" (cfg.seafileSettings // {
```
getchoo, line 16: ditto

nixos/modules/services/networking/seafile.nix
```diff
@@ -162,32 +245,38 @@ in
...
+        dataDir = if (!builtins.isNull cfg.dataDir) then cfg.dataDir else "${seafRoot}/data";
+        optionalDataDir = lib.lists.optional (!builtins.isNull cfg.dataDir) cfg.dataDir;
       in
       {
         seaf-server = {
           description = "Seafile server";
           partOf = [ "seafile.target" ];
-          after = [ "network.target" ];
+          unitConfig.RequiresMountsFor = optionalDataDir;
+          requires = [ "mysql.service" ];
```
getchoo, line 266: you could use a new option as mentioned above here to only conditionally require this if it's running locally

nixos/modules/services/networking/seafile.nix
```diff
@@ -162,32 +245,38 @@ in
...
+            User = cfg.user;
+            Group = cfg.group;
             StateDirectory = "seafile";
             RuntimeDirectory = "seafile";
             LogsDirectory = "seafile";
             ConfigurationDirectory = "seafile";
+            ReadWritePaths = optionalDataDir;
             ExecStart = ''
-              ${cfg.seafilePackage}/bin/seaf-server \
+              ${seafile-server}/bin/seaf-server \
```
getchoo, line 279: would `lib.getExe` work here instead?

nixos/modules/services/networking/seafile.nix
```diff
@@ -200,32 +289,58 @@ in
...
-                mkdir -p ${ccnetDir}/{GroupMgr,misc,OrgMgr,PeerMgr}
-                sqlite3 ${ccnetDir}/GroupMgr/groupmgr.db ".read ${cfg.seafilePackage}/share/seafile/sql/sqlite/groupmgr.sql"
-                sqlite3 ${ccnetDir}/misc/config.db ".read ${cfg.seafilePackage}/share/seafile/sql/sqlite/config.sql"
-                sqlite3 ${ccnetDir}/OrgMgr/orgmgr.db ".read ${cfg.seafilePackage}/share/seafile/sql/sqlite/org.sql"
-                sqlite3 ${ccnetDir}/PeerMgr/usermgr.db ".read ${cfg.seafilePackage}/share/seafile/sql/sqlite/user.sql"
-                sqlite3 ${dataDir}/seafile.db ".read ${cfg.seafilePackage}/share/seafile/sql/sqlite/seafile.sql"
-                echo "${cfg.seafilePackage.version}-sqlite" > "${seafRoot}"/server-setup
+                # Load schema on first install
+                ${pkgs.mysql-client}/bin/mysql --database=ccnet_db < ${seafile-server}/share/seafile/sql/mysql/ccnet.sql
+                ${pkgs.mysql-client}/bin/mysql --database=seafile_db < ${seafile-server}/share/seafile/sql/mysql/seafile.sql
```
getchoo, line 294:
```suggestion
                ${lib.getExe pkgs.mysql-client} --database=ccnet_db < ${seafile-server}/share/seafile/sql/mysql/ccnet.sql
                ${lib.getExe pkgs.mysql-client} --database=seafile_db < ${seafile-server}/share/seafile/sql/mysql/seafile.sql
```

nixos/modules/services/networking/seafile.nix
```diff
@@ -200,32 +289,58 @@ in
...
+                # Upgrade from 10.0 to 11.0: migrate to mysql
+                echo Migrating from version 10 to 11
+
+                # From https://github.com/haiwen/seahub/blob/e12f941bfef7191795d8c72a7d339c01062964b2/scripts/sqlite2mysql.sh
+
+                echo Migrating ccnet database to MySQL
+                ${pkgs.sqlite}/bin/sqlite3 ${ccnetDir}/PeerMgr/usermgr.db ".dump" | \
+                  ${python}/bin/python3 ${seahub}/scripts/sqlite2mysql.py > ${ccnetDir}/ccnet.sql
+                ${pkgs.sqlite}/bin/sqlite3 ${ccnetDir}/GroupMgr/groupmgr.db ".dump" | \
+                  ${python}/bin/python3 ${seahub}/scripts/sqlite2mysql.py >> ${ccnetDir}/ccnet.sql
```
getchoo, line 316:
```suggestion
                ${lib.getExe pkgs.sqlite} ${ccnetDir}/PeerMgr/usermgr.db ".dump" | \
                  ${lib.getExe python} ${seahub}/scripts/sqlite2mysql.py > ${ccnetDir}/ccnet.sql
                ${lib.getExe pkgs.sqlite} ${ccnetDir}/GroupMgr/groupmgr.db ".dump" | \
                  ${lib.getExe python} ${seahub}/scripts/sqlite2mysql.py >> ${ccnetDir}/ccnet.sql
```

nixos/modules/services/networking/seafile.nix
```diff
@@ -200,32 +289,58 @@ in
...
+                sed 's/ctime INTEGER/ctime BIGINT/g' -i ${ccnetDir}/ccnet.sql
+                sed 's/email TEXT, role TEXT/email VARCHAR(255), role TEXT/g' -i ${ccnetDir}/ccnet.sql
+                ${pkgs.mysql-client}/bin/mysql --database=ccnet_db < ${ccnetDir}/ccnet.sql
```
getchoo, line 319:
```suggestion
                ${lib.getExe pkgs.mysql-client} --database=ccnet_db < ${ccnetDir}/ccnet.sql
```

nixos/modules/services/networking/seafile.nix
```diff
@@ -200,32 +289,58 @@ in
...
+
+                echo Migrating seafile database to MySQL
+                ${pkgs.sqlite}/bin/sqlite3 ${dataDir}/seafile.db ".dump" | \
+                  ${python}/bin/python3 ${seahub}/scripts/sqlite2mysql.py > ${dataDir}/seafile.sql
```
getchoo, line 323:
```suggestion
                ${lib.getExe pkgs.sqlite}/ ${dataDir}/seafile.db ".dump" | \
                  ${lib.getExe python} ${seahub}/scripts/sqlite2mysql.py > ${dataDir}/seafile.sql
```

nixos/modules/services/networking/seafile.nix
```diff
@@ -200,32 +289,58 @@ in
...
+                sed 's/owner_id TEXT/owner_id VARCHAR(255)/g' -i ${dataDir}/seafile.sql
+                sed 's/user_name TEXT/user_name VARCHAR(255)/g' -i ${dataDir}/seafile.sql
+                ${pkgs.mysql-client}/bin/mysql --database=seafile_db < ${dataDir}/seafile.sql
```
getchoo, line 326:
```suggestion
                ${lib.getExe pkgs.mysql-client} --database=seafile_db < ${dataDir}/seafile.sql
```

nixos/modules/services/networking/seafile.nix
```diff
@@ -200,32 +289,58 @@ in
...
+
+                echo Migrating seahub database to MySQL
+                echo 'SET FOREIGN_KEY_CHECKS=0;' > ${seahubDir}/seahub.sql
+                ${pkgs.sqlite}/bin/sqlite3 ${seahubDir}/seahub.db ".dump" | \
+                  ${python}/bin/python3 ${seahub}/scripts/sqlite2mysql.py >> ${seahubDir}/seahub.sql
```
getchoo, line 331:
```suggestion
                ${lib.getExe pkgs.sqlite} ${seahubDir}/seahub.db ".dump" | \
                  ${lib.getExe python} ${seahub}/scripts/sqlite2mysql.py >> ${seahubDir}/seahub.sql
```

nixos/modules/services/networking/seafile.nix
```diff
@@ -200,32 +289,58 @@ in
...
+                sed 's/`permission` , `reporter` text NOT NULL/`permission` longtext NOT NULL/g' -i ${seahubDir}/seahub.sql
+                sed 's/varchar(256) NOT NULL UNIQUE/varchar(255) NOT NULL UNIQUE/g' -i ${seahubDir}/seahub.sql
+                sed 's/,    UNIQUE (`user_email`, `contact_email`)//g' -i ${seahubDir}/seahub.sql
+                sed '/INSERT INTO `base_dirfileslastmodifiedinfo`/d' -i ${seahubDir}/seahub.sql
+                sed '/INSERT INTO `notifications_usernotification`/d' -i ${seahubDir}/seahub.sql
+                sed 's/DEFERRABLE INITIALLY DEFERRED//g' -i ${seahubDir}/seahub.sql
+                ${pkgs.mysql-client}/bin/mysql --database=seahub_db < ${seahubDir}/seahub.sql
```
getchoo, line 338:
```suggestion
                ${lib.getExe pkgs.mysql-client} --database=seahub_db < ${seahubDir}/seahub.sql
```

nixos/modules/services/networking/seafile.nix
```diff
@@ -248,15 +364,15 @@ in
...
+            User = cfg.user;
+            Group = cfg.group;
             RuntimeDirectory = "seahub";
             StateDirectory = "seafile";
             LogsDirectory = "seafile";
             ConfigurationDirectory = "seafile";
+            ReadWritePaths = optionalDataDir;
             ExecStart = ''
-              ${pkgs.seahub.python.pkgs.gunicorn}/bin/gunicorn seahub.wsgi:application \
+              ${python.pkgs.gunicorn}/bin/gunicorn seahub.wsgi:application \
```
getchoo, line 375:
```suggestion
              ${lib.getExe python.pkgs.gunicorn} seahub.wsgi:application \
```

nixos/modules/services/networking/seafile.nix
```diff
@@ -269,30 +385,76 @@ in
           preStart = ''
             mkdir -p ${seahubDir}/media
             # Link all media except avatars
-            for m in `find ${pkgs.seahub}/media/ -maxdepth 1 -not -name "avatars"`; do
+            for m in `find ${seahub}/media/ -maxdepth 1 -not -name "avatars"`; do
               ln -sf $m ${seahubDir}/media/
             done
             if [ ! -e "${seafRoot}/.seahubSecret" ]; then
-                ${pkgs.seahub.python}/bin/python ${pkgs.seahub}/tools/secret_key_generator.py > ${seafRoot}/.seahubSecret
+                ${python}/bin/python ${seahub}/tools/secret_key_generator.py > ${seafRoot}/.seahubSecret
```
getchoo, line 392:
```suggestion
                ${lib.getExe python} ${seahub}/tools/secret_key_generator.py > ${seafRoot}/.seahubSecret
```

nixos/modules/services/networking/seafile.nix
```diff
@@ -269,30 +385,76 @@ in
...
-                install -D -t ${seahubDir}/media/avatars/groups ${pkgs.seahub}/media/avatars/groups/default.png
+                install -D -t ${seahubDir}/media/avatars/ ${seahub}/media/avatars/default.png
+                install -D -t ${seahubDir}/media/avatars/groups ${seahub}/media/avatars/groups/default.png
                 # init database
-                ${pkgs.seahub}/manage.py migrate
+                ${seahub}/manage.py migrate
                 # create admin account
-                ${pkgs.expect}/bin/expect -c 'spawn ${pkgs.seahub}/manage.py createsuperuser --email=${cfg.adminEmail}; expect "Password: "; send "${cfg.initialAdminPassword}\r"; expect "Password (again): "; send "${cfg.initialAdminPassword}\r"; expect "Superuser created successfully."'
-                echo "${pkgs.seahub.version}-sqlite" > "${seafRoot}/seahub-setup"
+                ${pkgs.expect}/bin/expect -c 'spawn ${seahub}/manage.py createsuperuser --email=${cfg.adminEmail}; expect "Password: "; send "${cfg.initialAdminPassword}\r"; expect "Password (again): "; send "${cfg.initialAdminPassword}\r"; expect "Superuser created successfully."'
```
getchoo, line 402:
```suggestion
                ${lib.getExe pkgs.expect} -c 'spawn ${seahub}/manage.py createsuperuser --email=${cfg.adminEmail}; expect "Password: "; send "${cfg.initialAdminPassword}\r"; expect "Password (again): "; send "${cfg.initialAdminPassword}\r"; expect "Superuser created successfully."'
```

nixos/modules/services/networking/seafile.nix
```diff
@@ -269,30 +385,76 @@ in
...
-                ${pkgs.seahub}/manage.py migrate
-                echo "${pkgs.seahub.version}-sqlite" > "${seafRoot}/seahub-setup"
+                # run django migrations
+                ${seahub}/manage.py migrate
+                echo "${seahub.version}-mysql" > "${seafRoot}/seahub-setup"
             fi
           '';
         };
+
+        seaf-gc = {
```
getchoo, line 413: it'd be nice if this was closer to the [`nix.gc` module](https://github.com/NixOS/nixpkgs/blob/9fb66c698910d30a9e41b53b6ba79a55c9da1a9c/nixos/modules/services/misc/nix-gc.nix) with it's timer

pkgs/applications/networking/seahub/default.nix
```diff
@@ -4,34 +4,29 @@
 , python3
 , makeWrapper
 , nixosTests
+, seafile-server-override ? null
```
getchoo, line 7:
```suggestion
, seafile-server ? python3.pkgs.seaserv
```
this makes a bit more sense imo

pkgs/applications/networking/seahub/default.nix
```diff
@@ -4,34 +4,29 @@
...
 }:
 let
-  python = python3.override {
-    packageOverrides = self: super: {
-      django = super.django_3;
+  python = python3.override (lib.optionalAttrs (!builtins.isNull seafile-server-override) {
+    packageOverrides = final: prev: {
+      seaserv = prev.toPythonModule (seafile-server-override.override {
+        python3 = prev.python;
+      });
```
getchoo, line 14:
```suggestion
```

this is a bit complex to account for overrides here to me

pkgs/applications/networking/seahub/default.nix
```diff
@@ -59,14 +55,21 @@ python.pkgs.buildPythonApplication rec {
...
+    pyopenssl
+    python-ldap
     qrcode
     pysearpc
-    seaserv
     gunicorn
     markdown
     bleach
+
+    seaserv
```
getchoo, line 66:
```suggestion
    (seaserv.override { inherit python3; })
```
something like this should achieve the same as above

pkgs/applications/networking/seahub/default.nix
```diff
@@ -79,6 +82,7 @@ python.pkgs.buildPythonApplication rec {
     tests = {
       inherit (nixosTests) seafile;
     };
+    seafile-server = python.pkgs.seaserv;
```
getchoo, line 85:
```suggestion
    inherit seafile-server;
```

pkgs/development/python-modules/djangosaml2/default.nix
getchoo, line 18:
```suggestion
  dependencies = [ django defusedxml pysaml2 ];
```

pkgs/development/python-modules/djangosaml2/default.nix
getchoo, line 7:
```suggestion
  pyproject = true;
```

pkgs/development/python-modules/djangosaml2/default.nix
getchoo, line 30:
```suggestion
  meta = {
```
https://github.com/NixOS/nixpkgs/issues/208242

pkgs/development/python-modules/djangosaml2/default.nix
getchoo, line 35:
```suggestion
    license = lib.licenses.asl20;
```

pkgs/development/python-modules/djangosaml2/default.nix
getchoo, line 36:
```suggestion
    maintainers = with lib.maintainers; [ ];
```

pkgs/development/python-modules/djangosaml2/default.nix
```diff
@@ -0,0 +1,38 @@
...
+  format = "pyproject";
+
+  disabled = pythonOlder "3.8";
+
+  src = fetchFromGitHub {
+    owner = "IdentityPython";
+    repo = "djangosaml2";
+    rev = "refs/tags/v${version}";
+    hash = "sha256-rbmEJuG2mgozpCFOXZUJFxv8v52IRQeaeAKfeUDACeU=";
+  };
```
getchoo, line 16:
```suggestion
  };
  
  build-system = [ setuptools ];
```

this should be explicitly listed

pkgs/development/python-modules/djangosaml2/default.nix
getchoo, line 1: you should probably use nixfmt here. see https://github.com/NixOS/nixpkgs/pull/313628

getchoo (changes requested):

nixos/modules/services/networking/seafile.nix
SuperSandro2000: It should be done in a way, that the settings are merged properly and not overwritten if one value is changed.

nixos/modules/services/networking/seafile.nix
```diff
@@ -137,23 +185,58 @@ in
...
       wantedBy = [ "multi-user.target" ];
       description = "Seafile components";
     };
 
     systemd.services =
       let
+        seahub = cfg.seahubPackage;
+        seafile-server = seahub.seafile-server;
+        python = seahub.python;
+
```
SuperSandro2000, line 233:
```suggestion
```
please just write those out, we don't need so many aliases

nixos/modules/services/networking/seafile.nix
SuperSandro2000: well, we don't want them to run at the same time ideally.

pkgs/applications/networking/seahub/default.nix
```diff
@@ -59,14 +55,21 @@ python.pkgs.buildPythonApplication rec {
...
-    seaserv
     gunicorn
     markdown
     bleach
+
+    seaserv
   ];
 
+  postPatch = ''
+    sed -i -e "s/SEAFILE_VERSION = '6.3.3'/SEAFILE_VERSION = '${version}'/" seahub/settings.py
```
SuperSandro2000, line 70: Please use substituteInPlace --replace-fail to notice when this no longer matches.

pkgs/development/python-modules/djangosaml2/default.nix
```diff
@@ -0,0 +1,38 @@
+{ lib, buildPythonPackage, fetchFromGitHub, defusedxml, django, pysaml2
+, pythonOlder, }:
```
SuperSandro2000, line 2:
```suggestion
, pythonOlder }:
```

pkgs/development/python-modules/djangosaml2/default.nix
```diff
@@ -0,0 +1,38 @@
...
+    python tests/run_tests.py
+
+    runHook postCheck
+  '';
+
+  meta = with lib; {
+    description = "Django SAML2 Service Provider based on pySAML2";
+    homepage = "https://github.com/IdentityPython/djangosaml2";
+    changelog =
+      "https://github.com/IdentityPython/djangosaml2/blob/v${version}/CHANGES";
```
SuperSandro2000, line 34:
```suggestion
    changelog = "https://github.com/IdentityPython/djangosaml2/blob/v${version}/CHANGES";
```

pkgs/development/python-modules/djangosaml2/default.nix
SuperSandro2000: we should set a maintainer

getchoo: @melvyn2 if you don't want to maintain this, you can add me. seems like a simple package

nixos/modules/services/networking/seafile.nix
melvyn2: I think this would be better as a second PR, as this one is getting pretty big already. Handling different database settings in the 2 formats and then also in the migration script would take non-insignificant effort/time, I think.

nixos/modules/services/networking/seafile.nix
melvyn2: see above

nixos/modules/services/networking/seafile.nix
melvyn2: mysql-client doesn't have `meta.mainProgram` set, but otherwise I switched to using `getExe`.

nixos/modules/services/networking/seafile.nix
melvyn2: I reformatted the gc options into its own attrset, and added the options that nix.gc presented.

pkgs/applications/networking/seahub/default.nix
melvyn2: This was a workaround for `makePythonPath` ignoring the module if it wasn't part of the python package set. I'll leave this open because the current workaround isn't great either.

pkgs/development/python-modules/djangosaml2/default.nix
melvyn2: applied nixfmt

pkgs/development/python-modules/djangosaml2/default.nix
melvyn2: applied nixfmt

pkgs/applications/networking/seahub/default.nix
melvyn2: Nevermind, I just forgot to apply `toPythonModule`.

getchoo (changes requested):

nixos/modules/services/networking/seafile.nix
```diff
@@ -162,30 +203,35 @@ in
...
           LockPersonality = true;
           RestrictRealtime = true;
           RestrictSUIDSGID = true;
+          NoNewPrivileges = true;
           MemoryDenyWriteExecute = true;
           SystemCallArchitectures = "native";
           RestrictAddressFamilies = [ "AF_UNIX" "AF_INET" ];
         };
+        dataDir = if (!builtins.isNull cfg.dataDir) then cfg.dataDir else "${seafRoot}/data";
+        optionalDataDir = lib.lists.optional (!builtins.isNull cfg.dataDir) cfg.dataDir;
```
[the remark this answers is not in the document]
getchoo: this is old. oops!

melvyn2:
@greizgh @schmittlauch sorry to bother but would you mind reviewing this?

greizgh (commented):
Thank you for your work. My main concern is the coupling with mysql: see my inline comment. Otherwise LGTM. It might be worth mentioning in the release notes that there are no migration path other than 10->11 (OTOH it's not like nextcloud where multiple versions are packaged)?

melvyn2:
Nearly done with remote database support, but I don't have any easy solutions for providing database password for seafile-server and ccnet. Seahub loads its settings from a python script so we can read the password from a `passwordFile`, but seafile-server reads it directly from the configuration file in `/etc/seafile`.
Might need an activation script to make it work.

pkgs/applications/networking/seahub/default.nix
```diff
@@ -1,46 +1,31 @@
-{ lib
-, fetchFromGitHub
-, fetchpatch
-, python3
-, makeWrapper
-, nixosTests
+{
+  lib,
+  fetchFromGitHub,
+  fetchpatch,
```
nixf-tidy-review, line 4: definition `fetchpatch` is not used

pkgs/servers/seafile-server/default.nix
```diff
@@ -1,48 +1,60 @@
...
+    inherit
+      stdenv
+      lib
+      fetchFromGitHub
+      cmake
+      libevent
+      ;
   };
 in
 stdenv.mkDerivation rec {
```
nixf-tidy-review, line 39: attrset is not necessary to be `rec`ursive

nixf-tidy-review (approved):

nixf-tidy-review (approved):

pkgs/servers/seafile-server/default.nix
```diff
@@ -1,48 +1,60 @@
...
+  glib,
+  libevent,
+  libsearpc,
+  openssl,
+  fuse,
+  libarchive,
+  libjwt,
+  curl,
+  which,
+  vala,
```
schmittlauch, line 20: A few of the function inputs appear to be unused, e.g libarchive, vala, oniguruma.

nixos/modules/services/networking/seafile.nix
```diff
@@ -138,22 +245,52 @@ in
   ###### Implementation
 
   config = mkIf cfg.enable {
+    services.mysql = {
+      enable = true;
```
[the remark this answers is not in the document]
greizgh: Hmm, that's a way to see it. I don't have a strong opinion, but this should be mentioned in the release notes (users will now have a mysql DB to backup).

nixos/modules/services/networking/seafile.nix
```diff
@@ -162,137 +474,255 @@ in
...
-            elif [[ $installedMajor == 8 && $installedMinor == 0 && $pkgMajor == 9 && $pkgMinor == 0 ]]; then
-                # Upgrade from 8.0 to 9.0
-                sqlite3 ${dataDir}/seafile.db ".read ${pkgs.seahub}/scripts/upgrade/sql/9.0.0/sqlite3/seafile.sql"
-                echo "${cfg.seafilePackage.version}-sqlite" > "${seafRoot}"/server-setup
-            elif [[ $installedMajor == 9 && $installedMinor == 0 && $pkgMajor == 10 && $pkgMinor == 0 ]]; then
-                # Upgrade from 9.0 to 10.0
-                sqlite3 ${dataDir}/seafile.db ".read ${pkgs.seahub}/scripts/upgrade/sql/10.0.0/sqlite3/seafile.sql"
-                echo "${cfg.seafilePackage.version}-sqlite" > "${seafRoot}"/server-setup
+            elif [[ $installedMajor == 10 && $installedMinor == 0 && $pkgMajor == 11 && $pkgMinor == 0 ]]; then
+                # Upgrade from 10.0 to 11.0: migrate to mysql
```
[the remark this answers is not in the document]
greizgh: That was the case before, I'm not sure the strategy will change (except we "reset" the supported migrations).

pkgs/servers/seafile-server/default.nix
melvyn2: `libarchive`, `vala`, and `oniguruma` are all needed dependencies, the build will not complete without them.
`openssl` and `curl` seem unused though.

melvyn2:
I haven't been able to work on this for a decent period of time, but I think with this rebase and modifications (all passing tests and running on my server), it is ready to be merged.

I've removed the remote-mysql additions because they are incomplete and will take non-negligible effort to get into a working state. This was "extra" functionality anyways that should really be in a separate PR; the old sqlite database only supported running on the same host anyways.
I've attached the last state of the WIP commit of this additions in case someone else feels an urge to add this functionality.
[mysql_configurable.patch](https://github.com/user-attachments/files/17171718/mysql_configurable.patch)

melvyn2:
Seems like the failed PR is an OfBorg/GitHub API issue and not something to do with this PR...

nixos/doc/manual/release-notes/rl-2411.section.md
```diff
@@ -529,6 +529,16 @@
...
 
+- The `services.seafile` module was updated to major version 11.
+  - As part of this upgrade, the database backend will be migrated
+    to MySQL. This process should be automatic, but in case of a botched migration, old sqlite files are not removed and can be used
+    to manually migrate the database.
+  - Additionally, the updated CSRF protection may prevent some users from logging in.
+    Specific origin addresses can be whitelisted using the `services.seafile.seahubExtraConf` option
+    (e.g. `services.seafile.seahubExtraConf = ''CSRF_TRUSTED_ORIGINS = ["https://example.com"]'';`).
+    Note that first solution of the [official FAQ answer](https://cloud.seatable.io/dtable/external-links/7b976c85f504491cbe8e/?tid=0000&vid=0000&row-id=BQhH-2HSQs68Nq2EW91DBA)
+    is not allowed by the `services.nginx` module's config-checker.
```
SuperSandro2000, line 540:
```suggestion
- The `services.seafile` module was updated to major version 11.
  - As part of this upgrade, the database backend will be migrated to MySQL.
    This process should be automatic, but in case of a botched migration,
    old sqlite files are not removed and can be used to manually migrate the database.
  - Additionally, the updated CSRF protection may prevent some users from logging in.
    Specific origin addresses can be whitelisted using the `services.seafile.seahubExtraConf` option
    (e.g. `services.seafile.seahubExtraConf = ''CSRF_TRUSTED_ORIGINS = ["https://example.com"]'';`).
    Note that first solution of the [official FAQ answer](https://cloud.seatable.io/dtable/external-links/7b976c85f504491cbe8e/?tid=0000&vid=0000&row-id=BQhH-2HSQs68Nq2EW91DBA)
    is not allowed by the `services.nginx` module's config-checker.
```

nixos/modules/services/networking/seafile.nix
```diff
@@ -1,36 +1,64 @@
...
 with lib;
 let
   cfg = config.services.seafile;
   settingsFormat = pkgs.formats.ini { };
 
-  ccnetConf = settingsFormat.generate "ccnet.conf" cfg.ccnetSettings;
+  ccnetConf = settingsFormat.generate "ccnet.conf" (
+    cfg.ccnetSettings
+    // {
+      Database = {
```
SuperSandro2000, line 15: We should probably swap this to allow end users to overwrite parts of the config. Also I think we should use lib.recursiveUpdate just to be safe.

nixos/modules/services/networking/seafile.nix
```diff
@@ -1,36 +1,64 @@
...
+        UNIX_SOCKET = "/var/run/mysqld/mysqld.sock";
+        DB = "ccnet_db";
+        CONNECTION_CHARSET = "utf8";
+      };
+    }
+  );
 
-  seafileConf = settingsFormat.generate "seafile.conf" cfg.seafileSettings;
+  seafileConf = settingsFormat.generate "seafile.conf" (
+    cfg.seafileSettings
```
SuperSandro2000, line 25: same here

nixos/modules/services/networking/seafile.nix
```diff
@@ -78,11 +106,17 @@ in
...
+              type = types.singleLineStr;
+              default = "ipv4:127.0.0.1";
+              example = "unix:/run/seafile/server.sock";
               description = ''
                 The binding address used by seafile fileserver.
+
+                The addr can be defined as one of the following:
+                ipv6:<ipv6addr> for binding to an IPv6 address.
+                unix:<named pipe> for binding to a unix named socket
+                ipv4:<ipv4addr> for binding to an ipv4 address
```
SuperSandro2000, line 118:
```suggestion
                - ipv6:<ipv6addr> for binding to an IPv6 address.
                - unix:<named pipe> for binding to a unix named socket
                - ipv4:<ipv4addr> for binding to an ipv4 address
```

nixos/modules/services/networking/seafile.nix
```diff
@@ -96,6 +130,17 @@ in
       '';
     };
 
+    seahubAddress = lib.mkOption {
+      type = types.singleLineStr;
+      default = "unix:/run/seahub/gunicorn.sock";
+      example = "[::1]:8083";
+      description = ''
+        Which address to bind the seahub server to, of the form:
+        HOST, HOST:PORT, unix:PATH.
```
SuperSandro2000, line 139:
```suggestion
        - HOST
        - HOST:PORT
        - unix:PATH.
```

nixos/modules/services/networking/seafile.nix
```diff
@@ -107,25 +152,87 @@ in
...
+    };
+
+    dataDir = mkOption {
+      type = types.path;
+      default = "${seafRoot}/data";
+      description = "Path in which to store user data";
+    };
+
+    gc = {
+      enable = mkEnableOption "automatic garbage collection on stored data blocks.";
```
SuperSandro2000, line 191:
```suggestion
      enable = mkEnableOption "automatic garbage collection on stored data blocks";
```

nixos/modules/services/networking/seafile.nix
```diff
@@ -162,137 +299,250 @@ in
...
-                sqlite3 ${ccnetDir}/GroupMgr/groupmgr.db ".read ${cfg.seafilePackage}/share/seafile/sql/sqlite/groupmgr.sql"
-                sqlite3 ${ccnetDir}/misc/config.db ".read ${cfg.seafilePackage}/share/seafile/sql/sqlite/config.sql"
-                sqlite3 ${ccnetDir}/OrgMgr/orgmgr.db ".read ${cfg.seafilePackage}/share/seafile/sql/sqlite/org.sql"
-                sqlite3 ${ccnetDir}/PeerMgr/usermgr.db ".read ${cfg.seafilePackage}/share/seafile/sql/sqlite/user.sql"
-                sqlite3 ${dataDir}/seafile.db ".read ${cfg.seafilePackage}/share/seafile/sql/sqlite/seafile.sql"
-                echo "${cfg.seafilePackage.version}-sqlite" > "${seafRoot}"/server-setup
+                mkdir -p ${cfg.dataDir}/library-template
+                # Load schema on first install
+                ${pkgs.mysql-client}/bin/mysql --database=ccnet_db < ${cfg.seahubPackage.seafile-server}/share/seafile/sql/mysql/ccnet.sql
+                ${pkgs.mysql-client}/bin/mysql --database=seafile_db < ${cfg.seahubPackage.seafile-server}/share/seafile/sql/mysql/seafile.sql
```
SuperSandro2000, line 355: mysql-client is an alias to mariadb.client and we should avoid using aliases. Normally ofborg should fail on those.

nixos/modules/services/networking/seafile.nix
```diff
@@ -162,137 +299,250 @@ in
...
-                echo "Unsupported upgrade" >&2
+                echo "Unsupported upgrade: $installedMajor.$installedMinor to $pkgMajor.$pkgMinor" >&2
                 exit 1
             fi
           '';
+
+          postStart = ''
+            # Fix unix socket permissions
+            host="${cfg.seafileSettings.fileserver.host}"
+            if [[ "''${host:0:5}" =~ ^unix:.* ]]; then
```
SuperSandro2000, line 412: Can we do that condition in nix?

nixos/doc/manual/release-notes/rl-2411.section.md
melvyn2: Done, it's strange that I left it formatted that way...

nixos/modules/services/networking/seafile.nix
melvyn2: I don't know why I didn't do it that way first, it's much cleaner now. I also made the nixos test use this codepath now.

nixos/modules/services/networking/seafile.nix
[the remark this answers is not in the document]
melvyn2: As I mentioned below, I think this feature belongs in a separate PR, else this one might never land.

greizgh (approved):
The migration went fine. Let's merge this since seafile is broken on master.

h7x4:
I'll merge this as is since its better than having it broken on master, but I don't think enabling mysql on the local host is ideal. Would be great if you could open an issue to track progress so others can find it and see that it's not intended as a long term solution. That being said, thank you for the great effort spanning several months!

pkgs/development/python-modules/djangosaml2/default.nix
```diff
@@ -0,0 +1,54 @@
...
+
+  # Falsely complains that 'defusedxml>=0.4.1 not satisfied by version 0.8.0rc2'
+  pythonRelaxDeps = [ "defusedxml" ];
+
+  pythonImportsCheck = [ "djangosaml2" ];
+
+  checkPhase = ''
+    runHook preCheck
+
+    python tests/run_tests.py
```
dotlambda, line 42:
```suggestion
  ${python.interpreter} tests/run_tests.py
```

Issue #149519, opened 2021-12-08
ddclient.service no longer accepts configFile


bernadinm:
### Describe the bug

It looks like a regression was introduced with this change here that no longer allows the `configFile` path to be considered.

### Steps To Reproduce
1. In your configuration.nix file, simply add the following below:
```
services.ddclient = { enable = true; configFile = "/home/user/configs/ddclient/ddclient.conf"; }; # dynamicdns
```
This apparently is introduced from PR #148417.

### Expected behavior
The ddclient.service systems unit should pick up the file specified on the `configFile` but it now defaults to `/run/ddclient/ddclient.conf`

### Screenshots

```console
× ddclient.service - Dynamic DNS Client
     Loaded: loaded (/etc/systemd/system/ddclient.service; enabled; vendor preset: enabled)
     Active: failed (Result: exit-code) since Tue 2021-12-07 20:41:47 PST; 29ms ago
TriggeredBy: ● ddclient.timer
    Process: 748995 ExecStartPre=/nix/store/<store>-ddclient-prestart (code=exited, status=0/SUCCESS)
    Process: 748997 ExecStart=/nix/store/<store>-perl5.34.0-ddclient-3.9.1/bin/ddclient -file /run/ddclient/ddclient.conf (code=exited, status=1/FAILURE)
   Main PID: 748997 (code=exited, status=1/FAILURE)
         IP: 0B in, 0B out
        CPU: 73ms
```

### Additional context
I simply did the following command to have this fail:
```
nixos-rebuild switch --upgrade
```

### Notify maintainers

@felschr @symphorien @bjornfor

### Metadata
Please run `nix-shell -p nix-info --run "nix-info -m"` and paste the result.

```console
this path will be fetched (0.06 MiB download, 0.30 MiB unpacked):
  /nix/store/4i653qmy6rs7r3m6zh60j0nwnnc5gi6k-bash-interactive-5.1-p12-dev
copying path '/nix/store/4i653qmy6rs7r3m6zh60j0nwnnc5gi6k-bash-interactive-5.1-p12-dev' from 'https://cache.nixos.org'...
 - system: `"x86_64-linux"`
 - host os: `Linux 5.10.79, NixOS, 22.05 (Quokka)`
 - multi-user?: `yes`
 - sandbox: `yes`
 - version: `nix-env (Nix) 2.4`
 - channels(miguel): `"nixos-21.11pre331460.931ab058daa"`
[...]

PR #318869, opened 2024-06-10
nixos/ddclient: create RunTime directory during install

andresgongora:
**ddclient** (`services.ddclient`) failed to find its configuration file. As reported in #149519 , this file is expected to be `/run/ddclient/ddclient.conf`. From that same issue report, a fix emerged to install the file in place using the service's `ExecStartPre`. However, this later command fails on my machine because the run time directory `/run/ddclient` does not exist.

Running the preinstall script manually leads to:
```
$ /nix/store/m7g1lz2xgjhv3x0wvgdqgi971zpskabd-ddclient-prestart
install: cannot create regular file '/run/ddclient/ddclient.conf': No such file or directory
```

I'm unsure if this change makes the situation worse by hiding an underlying problem away. I would have expected the run time directory to be created in some other manner or fashion, especially since `ddclient` is being used by more users. This is my very first PR to NixOS, kindly let me know if I should better covert it into issue, instead.

## Description of changes

Append `-D` to `install` to create leading path to files. This solves issues on instances where the configuration file can not be emplaced under `/run/ddclient/ddclient.conf` because the directory `/run/ddclient/` does not exist.

## Things done

- Built on platform(s)
  - [X] x86_64-linux
- notes: service activated without restarting/rebooting after install.

### DIFF
--- a/nixos/modules/services/networking/ddclient.nix
+++ b/nixos/modules/services/networking/ddclient.nix
@@ -30,9 +30,9 @@ let
   configFile = if (cfg.configFile != null) then cfg.configFile else configFile';
 
   preStart = ''
-    install --mode=600 --owner=$USER ${configFile} /run/${RuntimeDirectory}/ddclient.conf
+    install -D --mode=600 --owner=$USER ${configFile} /run/${RuntimeDirectory}/ddclient.conf
     ${lib.optionalString (cfg.configFile == null) (if (cfg.protocol == "nsupdate") then ''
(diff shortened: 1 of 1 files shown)


andresgongora:
My bad, this change is **not** needed.
`ddclient` runs as a service, and when it does, it creates the directory `/run/ddclient/`, allowing it to run as intended.
It can be triggered with `sudo systemctl start ddclient.service`.

PR #318872, opened 2024-06-10
trivial: make pkgs.writeText always fixed output

hraban:
## Description of changes

Makes all `pkgs.writeText` derivations fixed-output. This can prevent a cascade of rebuilds across nixpkgs versions because these small files tend to be found at the leaves of dependency trees.

To be honest this seems like such low hanging fruit that I'm surely missing some important reason not to do this...? I guess this PR is more of a "why don't we do this?" question.  (I googled quite a bit but couldn't find any answers to this specific question--just material around the concept of FODs in general)

Obviously this can be extended to other trivial builders but this is the simplest one.

@SuperSandro2000 I'm sure you know more about this?

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [x] aarch64-linux
  - [x] x86_64-darwin
  - [x] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/build-support/trivial-builders/default.nix
+++ b/pkgs/build-support/trivial-builders/default.nix
@@ -127,7 +127,13 @@ rec {
 
   # See doc/build-helpers/trivial-build-helpers.chapter.md
   # or https://nixos.org/manual/nixpkgs/unstable/#trivial-builder-text-writing
-  writeText = name: text: writeTextFile { inherit name text; };
+  writeText = name: text:
+    # TODO: To fully deprecate, replace the assertion with `lib.isString` and remove the warning
+    assert lib.assertMsg (lib.strings.isConvertibleWithToString text) ''
+      pkgs.writeText ${lib.strings.escapeNixString name}: The second argument should be a string, but it's a ${builtins.typeOf text} instead.'';
+    lib.warnIf (! lib.isString text) ''
+      pkgs.writeText ${lib.strings.escapeNixString name}: The second argument should be a string, but it's a ${builtins.typeOf text} instead, which is deprecated. Use `toString` to convert the value to a string first.''
+    writeTextFile { inherit name text; };
 
   # See doc/build-helpers/trivial-build-helpers.chapter.md
   # or https://nixos.org/manual/nixpkgs/unstable/#trivial-builder-text-writing
--- a/pkgs/build-support/trivial-builders/default.nix
+++ b/pkgs/build-support/trivial-builders/default.nix
@@ -133,7 +133,14 @@ rec {
       pkgs.writeText ${lib.strings.escapeNixString name}: The second argument should be a string, but it's a ${builtins.typeOf text} instead.'';
     lib.warnIf (! lib.isString text) ''
       pkgs.writeText ${lib.strings.escapeNixString name}: The second argument should be a string, but it's a ${builtins.typeOf text} instead, which is deprecated. Use `toString` to convert the value to a string first.''
-    writeTextFile { inherit name text; };
+    writeTextFile {
+      inherit name text;
+      derivationArgs = rec {
+        outputHashAlgo = "sha256";
+        outputHashMode = "flat";
+        outputHash = builtins.hashString outputHashAlgo (builtins.toString text);
+      };
+    };
 
   # See doc/build-helpers/trivial-build-helpers.chapter.md
   # or https://nixos.org/manual/nixpkgs/unstable/#trivial-builder-text-writing
--- a/pkgs/build-support/trivial-builders/test/concat-test.nix
+++ b/pkgs/build-support/trivial-builders/test/concat-test.nix
@@ -2,7 +2,7 @@
 let
   stri = writeText "pathToTest";
   txt1 = stri "abc";
-  txt2 = stri hello;
+  txt2 = stri (builtins.toString hello);
   res = concatText "textToTest" [ txt1 txt2 ];
 in
 runCommand "test-concatPaths" { } ''


hraban:
@ofborg build hello

pkgs/build-support/trivial-builders/default.nix
```diff
@@ -127,7 +127,14 @@ rec {
 
   # See doc/build-helpers/trivial-build-helpers.chapter.md
   # or https://nixos.org/manual/nixpkgs/unstable/#trivial-builder-text-writing
-  writeText = name: text: writeTextFile { inherit name text; };
+  writeText = name: text: writeTextFile {
+    inherit name text;
+    derivationArgs = rec {
+      outputHashAlgo = "sha256";
+      outputHashMode = "flat";
+      outputHash = builtins.hashString outputHashAlgo (builtins.toString text);
```
cdepillabout, line 135: This is an interesting idea.  I don't think I've seen this type of thing tried before.

I couldn't come up with a reason not to do this, except maybe:

- Does having `writeText` not be fixed output cause a lot of rebuilds in practice?  It feels like it should be relatively trivial (at least compared to everything else Hydra has to do), but I don't really have a good intuition for this.

- Would this change cause any significant amount of increased evaluation time?  (As long as the hashes for all calls to `writeText` aren't calculated at evaluation time, it seems like evaluation time likely wouldn't significantly increase?)

midchildan: I have two issues:

* Aren't all inputs to `writeText` already pure? Do rebuilds really happen without FODs?
* If FODs are necessary to avoid rebuilds, it might be better to create a separate FOD variant. `writeText` is often used in NixOS and Home Manager modules. Converting all uses of `writeText` to FOD would require hashes to be specified everywhere throughout NixOS and Home Manager configuration.

SuperSandro2000: You don't need to specify any hashes since we can hash everything without causing IFD or side effects. 

Name shouldn't change the hash of the FOD, can it? 

But tbh, this should be a ca-derivativion 😂

cdepillabout: > Aren't all inputs to writeText already pure? Do rebuilds really happen without FODs?

`writeText` just calls `writeTextFile` under the hood, which uses `runCommand` (which uses `stdenv.mkDerivation`, right?).  So calls to `writeText` will rebuild whenever something in `runCommand` / `stdenv.mkDerivation` changes, right?

> Converting all uses of writeText to FOD would require hashes to be specified everywhere throughout NixOS and Home Manager configuration.

Like Sandro said, it is not necessary to manually specify hashes in calls to `writeText`, since the input text is available in Nix, and then the `builtins.hashString` function is used to automatically calculate the hash.

> Name shouldn't change the hash of the FOD, can it?

I don't the the hash changes if the name changes.  However, I believe that the derivation will get rebuilt if the name changes, even if the hash doesn't change.  (I talk about this a little bit in https://functor.tokyo/blog/2023-07-24-fodonuts).

hraban: It should definitely go into staging thanks--updated.

@cdepillabout is correct about current (non-FOD) writeText causing plenty rebuilds. Example:

```
$ nix-build --no-link -E '(import <nixpkgs> {}).writeText "foo" "foobar"'
/nix/store/plm84lpdixks9wfh5a4pp2k5sgzsmfxd-foo
$ nix-build --no-link -E '(import <nixpkgs> { overlays = [(self: super: { gnutar = super.gnutar.overrideAttrs { name = "foobar"; };}) ];}).writeText "foo" "foobar"'
/nix/store/wwsbs9xckbrmj7zvwxlqcj3s4393gp45-foo
```

Yes the name change will trigger a rebuild (because the $out will change) but that's ok--this patch is more any of the *upstream* dependencies of writeText itself causing a rebuild. In the case of writeText this is pointless because we know the exact contents of the output in advance. On larger build farms this can cause an avalanche of rebuilds for no particular reason.

> If FODs are necessary to avoid rebuilds, it might be better to create a separate FOD variant. writeText is often used in NixOS and Home Manager modules. Converting all uses of writeText to FOD would require hashes to be specified everywhere throughout NixOS and Home Manager configuration.

@SuperSandro2000 already answered this, but this patch is a drop-in replacement for writeText, itself computing the hash from its own input. The text is known at runtime so you can just calculate the output hash during eval.

> But tbh, this should be a ca-derivativion 😂
I also thought about this, and am on the fence. I guess the trade-off is: doing FOD like this will increase all eval time, but is strictly more powerful for build farms who never even have to try rebuilding. Doing CA keeps eval time lower but does require that initial rebuild of only the text file itself every time something changes, just to realize that nothing has changed in the output.

Maybe something to measure?

hraban: Finally:

> Does having writeText not be fixed output cause a lot of rebuilds in practice? It feels like it should be relatively trivial (at least compared to everything else Hydra has to do), but I don't really have a good intuition for this.

I don't have any hard numbers but I manage a decent sized scope ("https://github.com/hraban/cl-nix-lite") with a few hundred packages which get rebuilt nightly based on nixpkgs-unstable: take this for the hearsay it is, but I feel like writeText is particularly painful because it blocks an entire "worker" on a menial non-CPU intensive task for a while. This means even with something like `-j 30`, I spend a few minutes at the start of each build at ~5% CPU just waiting for all these carbon copy files to be recreated before it can move on to actual building.

Obviously this is just based off of eye-balling build outputs, nothing scientific. Is there a way to just "try" a full hydra rebuild and compare numbers?

midchildan: Cool, I didn't realize FODs could be used this way.

> Obviously this is just based off of eye-balling build outputs, nothing scientific. Is there a way to just "try" a full hydra rebuild and compare numbers?

I tried something like this:

```console
❯ nix-build --dry-run -E 'let pkgs = import ./. { overlays = [ (final: prev: { gnutar = prev.gnutar.overrideAttrs { name = "foobar"; }; }) ]; }; in pkgs.mpv'
these 1554 derivations will be built
...
these 758 paths will be fetched (2908.90 MiB download, 9389.42 MiB unpacked):
...
```

But got the same numbers before and after the change. Maybe there are better examples other than `pkgs.mpv`.

infinisil, line 135:
```suggestion
      outputHash = builtins.hashString outputHashAlgo text;
```

infinisil (commented):
That's very neat! I think the reason this hasn't been done before is because it's a very non-conventional approach, and only works in very specific circumstances. E.g. it wouldn't work for directories, and it doesn't work when you need any post-processing on the output.

I can't see any reason this would break anything, the only problem could be that it makes eval too slow. Indeed, ofborg is reporting a [74% increase in eval time](https://github.com/NixOS/nixpkgs/pull/318872/checks?check_run_id=26143053438)! This could be a fluke though, so let's rerun

@ofborg eval

pkgs/build-support/trivial-builders/default.nix
hraban: This was my original approach but it breaks a test which passes a derivation as the text, and relies on implicit toString functionality somewhere to convert it to its string representation. builtins.hashString is less forgiving and requires the input to already be a string. I'll try and find it again and add a comment instead.

hraban:
@ofborg eval

pkgs/build-support/trivial-builders/default.nix
infinisil: Oh I see, yeah then we can't do that, but in the future we should consider deprecating non-string arguments to `writeText`, because this behavior can easily lead to bugs.

hraban: The test is in https://github.com/nixos/nixpkgs/blob/f78bb370168b06e51af27ada73618b12525a415c/pkgs/build-support/trivial-builders/test/concat-test.nix#L5 -- should I just add builtins.toString there, instead?

infinisil: Since `writeText` can be used by code outside Nixpkgs, I'd prefer to still support non-strings for now, but adding `toString` there is good in any case.

If you want to go for deprecating non-string arguments:

```nix
# TODO: To fully deprecate, replace the assertion with `lib.isString` and remove the warning
assert assertMsg (lib.strings.isConvertibleWithToString text) ''
  pkgs.writeText ${lib.strings.escapeNixString name}: The second argument should be a string, but it's a ${builtins.typeOf text} instead.'';
lib.warnIf (! lib.isString text) ''
  pkgs.writeText ${lib.strings.escapeNixString name}: The second argument should be a string, but it's a ${builtins.typeOf text} instead, which is deprecated. Use `toString` to convert the value to a string first.''
writeTextFile {
  inherit name name;
  # ...
    outputHash = builtins.hashString outputHashAlgo (builtins.toString text);
}
```

hraban: Done. I tried moving it to `writeTextFile` but that caused an infinite recursion.. 🤷‍♀️ so I've left it here.

hraban:
Any other thoughts on how / whether to proceed with this?

infinisil (approved):
Looking good to me! Let's merge this, but be ready to revert if there's a breakage from this :smiley:

alyssais:
I'm afraid I've found some breakage. :(

Since this change, I am no longer able to use maintainers/scripts/update.nix:

```
% nix-shell maintainers/scripts/update.nix --argstr path libinput
this derivation will be built:
  /nix/store/5clkl4i907cym7ryr9gw9b7ajg6san2j-packages.json.drv
building '/nix/store/5clkl4i907cym7ryr9gw9b7ajg6san2j-packages.json.drv'...
error: fixed output derivation 'packages.json' is not allowed to refer to other store paths.
       You may need to use the 'unsafeDiscardReferences' derivation attribute, see the manual for more details.
```

alyssais:
In general I don't think this is correct.  FODs are not allowed to have references, but `writeText` is often used to create a file with references.  In cases where references are not required, there's already `builtins.toFile`, which does the same thing but is a FOD.

So I think this needs to be reverted, but there's still something we could do to reduce the rebuilds as mentioned in the motivation: switch to using builtins.toFile everywhere we don't need references, and document clearly that builtins.toFile should be preferred where possible.  Currently it seems writeText is way overused in Nixpkgs.

Thanks a lot for trying this though!

infinisil:
Ahh too bad, thanks for the report @alyssais! I opened https://github.com/NixOS/nixpkgs/pull/323094 for the revert (only the last commit)

eclairevoyant:
> So I think this needs to be reverted, but there's still something we could do to reduce the rebuilds as mentioned in the motivation: switch to using builtins.toFile everywhere we don't need references, and document clearly that builtins.toFile should be preferred where possible.

Good catch; just noticed even [the manual section aboux `toFile`](https://nix.dev/manual/nix/2.22/language/builtins#builtins-toFile) mentions this:

>  It is also not possible to reference the result of a derivation. If you are using Nixpkgs, the `writeTextFile` function is able to do that.

trofi:
It also breaks in-tree uses, like in `xdg-utils`:

```
$ nix build --no-link -f. xdg-utils
error: fixed output derivation 'xdg-utils-prologue' is not allowed to refer to other store paths.
       You may need to use the 'unsafeDiscardReferences' derivation attribute, see the manual for more details.
```

roberth:
This would have been caught if ofborg had a way to find its use in `pkgs/build-support/trivial-builders/test/sample.nix`. That's not for a test that's dedicated for `writeFile` but rather some other "trivial" builders (I hate that term) that happen to use it to generate test data. So that might not be feasible, because it can't possibly run everything remotely related to any change, so it'd be best to have some dedicated tests for `writeFile` and make ofborg find those based on commit messages or changed file paths.

hraban:
No kidding; `builtins.toFile`! There's the answer. Thanks, and sorry for the goose chase :D

PR #318892, opened 2024-06-10
nixosTests.taskchampion-sync-server: init

doronbehar:
###### Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
- [ ] Tested, as applicable:

### DIFF
--- a/nixos/modules/services/misc/taskchampion-sync-server.nix
+++ b/nixos/modules/services/misc/taskchampion-sync-server.nix
@@ -1,0 +1,85 @@
+{
+  config,
+  pkgs,
+  lib,
+  ...
+}:
+let
+  inherit (lib) types;
+  cfg = config.services.taskchampion-sync-server;
+in
+{
+  options.services.taskchampion-sync-server = {
+    enable = lib.mkEnableOption "TaskChampion Sync Server for Taskwarrior 3";
+    package = lib.mkPackageOption pkgs "taskchampion-sync-server" { };
+    user = lib.mkOption {
+      description = "Unix User to run the server under";
+      type = types.str;
+      default = "taskchampion";
+    };
+    group = lib.mkOption {
+      description = "Unix Group to run the server under";
+      type = types.str;
+      default = "taskchampion";
+    };
+    port = lib.mkOption {
+      description = "Port on which to serve";
+      type = types.port;
+      default = 10222;
+    };
+    openFirewall = lib.mkEnableOption "Open firewall port for taskchampion-sync-server";
+    dataDir = lib.mkOption {
+      description = "Directory in which to store data";
+      type = types.path;
+      default = "/var/lib/taskchampion-sync-server";
+    };
+    snapshot = {
+      versions = lib.mkOption {
+        description = "Target number of versions between snapshots";
+        type = types.ints.positive;
+        default = 100;
+      };
+      days = lib.mkOption {
+        description = "Target number of days between snapshots";
+        type = types.ints.positive;
+        default = 14;
+      };
+    };
+  };
+
+  config = lib.mkIf cfg.enable {
+    users.users.${cfg.user} = {
+      isSystemUser = true;
+      inherit (cfg) group;
+    };
+    users.groups.${cfg.group} = { };
+    networking.firewall.allowedTCPPorts = lib.mkIf (cfg.openFirewall) [ cfg.port ];
+    systemd.tmpfiles.settings = {
+      "10-taskchampion-sync-server" = {
+        "${cfg.dataDir}" = {
+          d = {
+            inherit (cfg) group user;
+            mode = "0750";
+          };
+        };
+      };
+    };
+
+    systemd.services.taskchampion-sync-server = {
+      wantedBy = [ "multi-user.target" ];
+      after = [ "network.target" ];
+      serviceConfig = {
+        User = cfg.user;
+        Group = cfg.group;
+        DynamicUser = false;
+        ExecStart = ''
+          ${lib.getExe cfg.package} \
+            --port ${builtins.toString cfg.port} \
+            --data-dir ${cfg.dataDir} \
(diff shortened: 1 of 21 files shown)


herberteuler:
I did not end up using TaskWarrier 3 (because I prefer to manage all personal data with git), and I do not use VIM as my primary editor, so my comments are only based on scanning the change set but not on the actual experience of trying it.

nixos/modules/services/misc/taskchampion-sync-server.nix
```diff
@@ -0,0 +1,85 @@
...
+            mode = "0750";
+          };
+        };
+      };
+    };
+
+    systemd.services.taskchampion-sync-server = {
+      wantedBy = [ "multi-user.target" ];
+      after = [ "network.target" ];
+      serviceConfig = {
```
herberteuler, line 71: Is `Restart = "always"` needed here to ensure high availability?

pkgs/applications/editors/vim/plugins/overrides.nix
```diff
@@ -44,7 +44,7 @@
 , statix
 , stylish-haskell
 , tabnine
-, taskwarrior
+, taskwarrior3
```
herberteuler, line 47: Several, if not many, people seem to prefer to use Git to manage task data. For example, see [this discussion](https://www.reddit.com/r/taskwarrior/comments/1dffv6b/should_i_use_task_warrior_v3_or_v2/). The current option is sticking to TaskWarrior 2. Perhaps instead of replacing `taskwarrior` with `taskwarrior3`, keeping both of them is a better option?

nixos/modules/services/misc/taskchampion-sync-server.nix
doronbehar: > Is `Restart = "always"` needed here to ensure high availability?

In my experience this service never crashed or something like that after my Laptop's hibernation / sleeps, so I don't see a reason to make such an effort.

pkgs/applications/editors/vim/plugins/overrides.nix
doronbehar: > Several, if not many, people seem to prefer to use Git to manage task data. For example, see [this discussion](https://www.reddit.com/r/taskwarrior/comments/1dffv6b/should_i_use_task_warrior_v3_or_v2/). The current option for them is to stick with TaskWarrior 2. Therefore, perhaps instead of replacing `taskwarrior` with `taskwarrior3`, keeping both of them is better here?

The vim plugin that is evaluated there eventually is based on the `ftplugin` and `ftdetect` and `syntax` files of Taskwarrior - simply meant to give you syntax highlighting for the `.taskrc` file. These files were barely changed between 2.6 -> 3.0, so I don't see a reason to put them both.

nixos/tests/taskserver.nix
```diff
@@ -81,7 +81,7 @@ in {
     };
 
     client1 = { pkgs, ... }: {
-      environment.systemPackages = [ pkgs.taskwarrior pkgs.gnutls ];
+      environment.systemPackages = [ pkgs.taskwarrior2 pkgs.gnutls ];
```
herberteuler, line 84: Perhaps this `taskserver.nix` should be renamed to `taskserver2.nix` to highlight its purpose of testing version 2 of TaskWarrior?

nixos/tests/taskchampion-sync-server.nix
```diff
@@ -0,0 +1,48 @@
...
+        client.succeed("touch ~/.taskrc")
+        client.succeed("echo sync.server.origin=http://server:${port} >> ~/.taskrc")
+        client.succeed("echo sync.server.client_id=${uuid} >> ~/.taskrc")
+        client.succeed("echo sync.encryption_secret=${password} >> ~/.taskrc")
+        client.succeed("task add hello world")
+        client.succeed("task sync")
+
+        # Useful for debugging
+        client.copy_from_vm("/root/.task", "client")
+        server.copy_from_vm("${cfg.dataDir}", "server")
```
herberteuler, line 45: Do we need to compare the copied data and somehow check for their equivalence?

herberteuler (approved):
I left a few comments, but none of them were critical.

pkgs/applications/editors/vim/plugins/overrides.nix
doronbehar: > The change is not trivial either:
> 
> ```
> $ diff /nix/store/m99g4v6xj29yb8mgqkc7v90lcrbk26di-taskwarrior-3.0.2/share/vim-plugins/task/syntax/taskrc.vim /nix/store/p32chw38kv19i3lnx85fsxcalbjdd9yf-taskwarrior-2.6.2/share/vim-plugins/task/syntax/taskrc.vim | wc -l
> 58
> ```
> 
> The change may become bigger in the future, too.

Hmm that is convincing. I added the attributes `vimPlugins.taskwarrior{2,3}` and made `vimPlugins.taskwarrior` a throwing alias.

pkgs/applications/editors/vim/plugins/deprecated.json
```diff
@@ -7,6 +7,10 @@
         "date": "2021-12-21",
         "new": "cmp-tmux"
     },
+    "taskwarrior": {
+        "date": "2024-08-13",
+        "new": "taskwarrior3 or taskwarrior3"
```
herberteuler, line 12: Is this supposed to be "taskwarrior2 or taskwarrior3"?

herberteuler (changes requested):
There seems to be a typo.

pkgs/applications/editors/vim/plugins/deprecated.json
doronbehar: > Is this supposed to be "taskwarrior2 or taskwarrior3"?

Yes it is. Thanks for catching that!

PR #318935, opened 2024-06-11
nixd: fix cross compiling

danielbarter:
## Description of changes

Fixing cross compiling for nixd.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [x] aarch64-linux (cross)
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/development/tools/language-servers/nixd/default.nix
+++ b/pkgs/development/tools/language-servers/nixd/default.nix
@@ -2,6 +2,7 @@
   lib,
   stdenv,
   fetchFromGitHub,
+  cmake,
   boost182,
   gtest,
   llvmPackages,
@@ -35,6 +36,8 @@ let
     ];
 
     mesonBuildType = "release";
+
+    strictDeps = true;
 
     doCheck = true;
 
@@ -128,6 +131,8 @@ in
         gtest
         boost182
       ];
+
+      nativeBuildInputs = common.nativeBuildInputs ++ [ cmake ];
 
       env.CXXFLAGS = "-include ${nix.dev}/include/nix/config.h";
 


pkgs/development/tools/language-servers/nixd/default.nix
```diff
@@ -32,6 +32,7 @@ let
       meson
       ninja
       pkg-config
+      llvmPackages.llvm
```
inclyc, line 35: I don't know why llvm is required for all packages. Maybe just adding this item for libnixt would be sufficient?

inclyc (changes requested):

pkgs/development/tools/language-servers/nixd/default.nix
danielbarter: Actually, it isn't needed for `libnixt`, it is just needed for `nixd`. You OK with the current version, where I just override the native build inputs for `nixd`?

pkgs/development/tools/language-servers/nixd/default.nix
```diff
@@ -129,6 +129,13 @@ in
         boost182
       ];
 
+      nativeBuildInputs = [
+        meson
+        ninja
+        pkg-config
+        llvmPackages.llvm
```
inclyc, line 136: I don't see the rationale? What's the error behind this?

danielbarter: currently, `nix build .#pkgsCross.aarch64-multiplatform.nixd` fails to build because `llvmPackages.llvm` is a build time dependency of `nixd`. To fix, it needs to be added to `nativeBuildInputs`.

inclyc: > fails to build because

Sorry for my confusing words, I know you may submitting patches in order to improve nixd cross compilation, but how about pasting some logs here? For example, detailed error information about "missing" llvm native build inputs.

pkgs/development/tools/language-servers/nixd/default.nix
```diff
@@ -129,6 +129,13 @@ in
         boost182
       ];
 
+      nativeBuildInputs = [
```
inclyc, line 132:
```suggestion
      nativeBuildInputs = common.nativeBuildInputs ++ [ llvmPackages.llvm ];
```

Maybe better just declaring what is need for cross compilation.

Aleksanaa:
Could you also add a `strictDeps = true` to common?

pkgs/development/tools/language-servers/nixd/default.nix
danielbarter: Sorry for being confusing also! Here is the log: https://gist.github.com/danielbarter/862362de655b0ca7d130cd742684caa2

nothing really interesting going on, just the missing runtime dep

inclyc: > meson.build:34:7: ERROR: Dependency lookup for llvm with method 'cmake' failed: CMake binary for machine host machine not found. Giving up.

They how about adding "cmake" to `nativeBuildInputs`? Use cmake modules to find llvm then.

danielbarter: adding `llvmPackages.llvm` to `nativeBuildInputs` is the correct solution here.

inclyc: > Why would we add cmake as a build time dependency to just tell meson how to find another build time dependency.

As per my understanding, meson have several ways to find llvm. The first choice is "llvm-config" binary tool, and the second one is the LLVM cmake module. LLVM itself is built from cmake thus it is make sense to add "cmake" to find it. "llvm-config" binary tool is suitable for finding llvm configured with suitable installing prefix and may not suitable for cross-compiling.

> another build time dependency.

I suppose if you have `cmake` in `nativeBuildInputs`, meson will not complain missing llvm as build time dep. LLVM supporting libraries are required as a runtime library, it is *indeed* runtime deps, not a build time dep.

danielbarter: OK fine.

`nixd` does not cross compile. You guys fix it.

Aleksanaa:
Eh, accidental touch on mobile. Why did you close it?

danielbarter:
Based on what @inclyc was saying, it seems to be a bit more subtle than I realized. Will leave it up to you guys if you want to add cmake as a build time dependency

inclyc:
> Based on what @inclyc was saying, it seems to be a bit more subtle than I realized. Will leave it up to you guys if you want to add cmake as a build time dependency

Could you please test if adding `cmake` works for your environment? If there are something still failing maybe we should add llvm then.

danielbarter:
@inclyc: adding cmake does fix it. Apologies for before, didn't want to spend to much time fixing it.

inclyc (approved):
Thanks! @danielbarter

inclyc:
@NixOS/nixpkgs-merge-bot merge

inclyc (approved):

PR #319033, opened 2024-06-11
bpftrace: add glibc.dev to default includes

arianvp:
## Description of changes

This makes the `$out/bin/tcpconnect.bt` and friends work again.

Unrelated; But we  still get a weird WARNING about kernel headers not found which is a bit strange to me as

https://github.com/bpftrace/bpftrace/pull/3156 mentions this WARNING only should be printed if kernel headers are neccesery but we don't need them as we use BTF

```
$ sudo ./result/bin/tcpconnect.bt
WARNING: Could not find kernel headers in  or . To specify a particular path to kernel headers, set the env variables BPFTRACE_KERNEL_SOURCE and, optionally, BPFTRACE_KERNEL_BUILD if the kernel was built in a different directory than its source. To create kernel headers run 'modprobe kheaders', which will create a tar file at /sys/kernel/kheaders.tar.xz Attaching 2 probes...
Tracing tcp connections. Hit Ctrl-C to end.
TIME     PID      COMM             SADDR                                   SPORT  DADDR                                   DPORT

```

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/by-name/bp/bpftrace/package.nix
+++ b/pkgs/by-name/bp/bpftrace/package.nix
@@ -2,7 +2,9 @@
 , llvmPackages, elfutils, bcc
 , libbpf, libbfd, libopcodes
 , cereal, asciidoctor
+, glibc
 , cmake, pkg-config, flex, bison
+, makeWrapper
 , util-linux
 , nixosTests
 }:
@@ -30,6 +32,7 @@ stdenv.mkDerivation rec {
     cmake pkg-config flex bison
     llvmPackages.llvm.dev
     util-linux
+    makeWrapper
   ];
 
   # tests aren't built, due to gtest shenanigans. see:
@@ -48,6 +51,7 @@ stdenv.mkDerivation rec {
   # Pull BPF scripts into $PATH (next to their bcc program equivalents), but do
   # not move them to keep `${pkgs.bpftrace}/share/bpftrace/tools/...` working.
   postInstall = ''
+    wrapProgram $out/bin/bpftrace --add-flags "-I ${glibc.dev}/include"
     ln -sr $out/share/bpftrace/tools/*.bt $out/bin/
     # do not use /usr/bin/env for shipped tools
     # If someone can get patchShebangs to work here please fix.


pkgs/by-name/bp/bpftrace/package.nix
```diff
@@ -48,6 +51,7 @@ stdenv.mkDerivation rec {
   # Pull BPF scripts into $PATH (next to their bcc program equivalents), but do
   # not move them to keep `${pkgs.bpftrace}/share/bpftrace/tools/...` working.
   postInstall = ''
+    wrapProgram $out/bin/bpftrace --add-flags "-I ${glibc.dev}/include"
```
arianvp, line 54: Not a 100% sure if this is the right thing to do in the presence of cross-compilation. I assume there is some way to get the correct include path from `stdenv` somehow and set it up. but IDK how.

martinetd:
Thanks!
I want to check things a bit more (especially around that warning) so please give me a few days.

Let's also add a test that at least one of these shipped .bt script works (execsnoop is probably easy enough to check output?), we really ought to have noticed that they stopped working with the latest release.
I'll do that if you don't have time/want to dig into this.

martinetd:
Ok so had a quick look, there are two distinct problems:
- that warning is printed because bpftrace cannot guess the kernel location (src/utils.cpp `get_kernel_dirs()`) - in the BTF case that directory shouldn't actually be needed, but I guess something got mixed up here.
Either way, since we stopped building bpftrace as a kernel package we no longer have access to the kernel sources (they're not available in /run/booted-system), so this part only worked because of the kheaders extraction; since we do have kheader we could just extract it (being a bit more careful than the version in bpftrace that got a CVE... For example not extracting it in a world-writable directory!), but since it's not needed I'll have a second look to try to only print the warning if it's really needed.
- in the BTF case we don't need the kernel headers, but we apparently need the glibc headers (for things like AF_INET) as you've added here -- looking at the bt files actually most don't include anything with BTF, you were unlucky looking at tcp one :/
Anyway, bpftrace tries to get it from `clang -Wp,-v -x c -fsyntax-only /dev/null` in src/clang_parser.cpp `system_include_paths()` but this is calling an unqualified clang-<major> which is unlikely to be in PATH.

Running that on my system gives:
```
 /nix/store/nz8z2b1q8mypm5m4mg9wv33rxwvazl73-clang-wrapper-16.0.6/resource-root/include (which is /nix/store/8g97fdfhjqyc38jpjz32p1nxyh3b0jx0-clang-16.0.6-lib/lib/clang/16/include)
 /nix/store/gzxqm8dyfirbysqjhh78ivam62ll0m87-glibc-2.39-5-dev/include
```
so it's a bit more than the glibc one, but it shouldn't be too hard to just patch that function to return these paths from build time (and since it's a native input that should work for cross? bpftrace derivation already pulls in clang-lib and glibc so that's probably already a problem if that doesn't work)

Just fixing that will fix the tcp*.bt examples, similarly to what your patch does.
I think it's cleaner than wrapping bpftrace even if it's a bit more intrusive, if only because I think these paths aren't always added (in the non-BTF case we don't want them afaiu, because we use kernel headers instead)


Ran out of time for tonight, but I'll do that unless you want to get your hands dirty :)

Recap short term:
1. patch `ClangParser::system_include_paths()` to return glibc/clang-lib paths directly
2. patch `get_kernel_dirs()`

Longer term (to be sent upstream):
1. add a CMake option to override system includes and bypass `system_include_paths()` if set, and use that
2. work out how to have `get_kernel_dirs()` only called if really required (non-btf) with upstream, so we won't need to patch it anymore

martinetd:
I've done the work in https://github.com/NixOS/nixpkgs/pull/319836 so closing this (marked you as co-author of the glibc headers patch)

Please check!

For kernel headers I've considered redoing the extraction ourselves (the problem is just that "it's hard to get right", not that it's insecure in itself), but I think that'll require more discussions so I just hid the warning for now; will open an issue to discuss this upstream tomorrow.

PR #319047, opened 2024-06-11
mesa: split out the Darwin build into a separate expression

K900:
The way we build Mesa for Linux (and potentially others) is very, very different from the way we build Mesa on Darwin. In short, on Linux we generally want to build as much of it as possible, while on Darwin we want to build just enough of it to get GLX for legacy X applications.

So, split out the Darwin build into its own expression entirely, and give it just enough inputs and just enough feature flags to actually do the one thing we want it to do, and nothing else.

Then use the newly gained flexibility to remove most of the platform conditionals from the Linux expression.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [x] aarch64-linux
  - [x] x86_64-darwin
  - [x] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/development/libraries/mesa/default.nix
+++ b/pkgs/development/libraries/mesa/default.nix
@@ -92,6 +92,7 @@
     "intel-nullhw"
   ]
 , mesa
+, makeSetupHook
 }:
 
 # When updating this package, please verify at least these build (assuming x86_64-linux):
@@ -394,5 +395,10 @@ in stdenv.mkDerivation {
       '';
       disallowedRequisites = [ llvmPackages.llvm mesa.drivers ];
     };
+
+    llvmpipeHook = makeSetupHook {
+      name = "llvmpipe-hook";
+      substitutions.drivers = mesa.drivers;
+    } ./llvmpipe-hook.sh;
   };
 }
--- a/pkgs/development/libraries/mesa/default.nix
+++ b/pkgs/development/libraries/mesa/default.nix
@@ -1,6 +1,4 @@
 { lib
-, OpenGL
-, Xplugin
 , bison
 , buildPackages
 , directx-headers
@@ -42,19 +40,9 @@
 , xcbutilkeysyms
 , xorg
 , zstd
-, withValgrind ?
-  lib.meta.availableOn stdenv.hostPlatform valgrind-light
-  && !valgrind-light.meta.broken
-, withLibunwind ? lib.meta.availableOn stdenv.hostPlatform libunwind
-, enableGalliumNine ? stdenv.isLinux
-, enableOSMesa ? stdenv.isLinux
-, enableOpenCL ? stdenv.isLinux && stdenv.isx86_64
-, enableTeflon ? stdenv.isLinux && stdenv.isAarch64  # currently only supports aarch64 SoCs, may change in the future
 , enablePatentEncumberedCodecs ? true
 
-, galliumDrivers ?
-  if stdenv.isLinux
-  then [
+, galliumDrivers ? [
     "d3d12" # WSL emulated GPU (aka Dozen)
     "iris" # new Intel (Broadwell+)
     "kmsro" # special "render only" driver for GPUs without a display controller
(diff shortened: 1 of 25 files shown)


pkgs/top-level/all-packages.nix
```diff
@@ -23285,10 +23285,12 @@ with pkgs;
   # Default libGLU
   libGLU = mesa_glu;
 
-  mesa = darwin.apple_sdk_11_0.callPackage ../development/libraries/mesa {
-    inherit (darwin.apple_sdk_11_0.frameworks) OpenGL;
```
alyssais, line 23289: OpenGL no longer required?

K900: I don't think it ever was, actually - it's only ever dlopened from a fixed path.

K900:
Re-added split outputs to the Darwin expression to make libGL stubs build again.

pkgs/development/libraries/mesa/default.nix
```diff
@@ -204,7 +173,7 @@ self = stdenv.mkDerivation {
     "out" "dev" "drivers"
   ] ++ lib.optionals enableOSMesa [
     "osmesa"
-  ] ++ lib.optionals stdenv.isLinux [
+  ] ++ [
     "driversdev"
```
philiptaron, line 177: I'd move this up to the unilateral section so it looks neater, even thought it causes a rebuild.

pkgs/development/libraries/mesa/default.nix
```diff
@@ -100,28 +93,19 @@
...
 
 # When updating this package, please verify at least these build (assuming x86_64-linux):
 # nix build .#mesa .#pkgsi686Linux.mesa .#pkgsCross.aarch64-multiplatform.mesa .#pkgsMusl.mesa
 
 let
-  version = "24.1.1";
-  hash = "sha256-ADiCbG9+iNkLTOb3GRkvpYyn3t9O3KoRdM972SDvieo=";
-
-  # Release calendar: https://www.mesa3d.org/release-calendar.html
-  # Release frequency: https://www.mesa3d.org/releasing.html#schedule
```
philiptaron, line 120: These want to move over to common.nix

philiptaron (approved):
Read through the Nix code carefully and built on x86_64-darwin x86_64-linux. Couple of in-no-way-blocking comments.

[ofBorg is unhappy](https://gist.github.com/GrahamcOfBorg/ba39f905102286c86bfa9cb02119ca46), though, due to `pkgs/development/libraries/libva/default.nix` on Darwin. (It's marked as unsupported, but it doesn't eval with this PR.)

pkgs/top-level/all-packages.nix
```diff
@@ -23285,10 +23285,12 @@ with pkgs;
...
-    inherit (darwin.apple_sdk_11_0.libs) Xplugin;
-  };
+  mesa = if stdenv.isDarwin
+    then darwin.apple_sdk_11_0.callPackage ../development/libraries/mesa/darwin.nix {
```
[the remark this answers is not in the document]
K900: This is the newest one we have.

pkgs/development/libraries/mesa/default.nix
K900: They're really not relevant information.

pkgs/development/libraries/mesa/default.nix
K900: I'll do more cleanups in a follow-up PR to staging.

philiptaron (approved):

vcunat:
@ofBorg eval

vcunat:
ofBorg is still broken.  An assertion on the right place would do it, I think.  Or maybe there's some better way.

K900:
Retargeted to next staging, current staging is unblocked by horrible hack.

philiptaron:
On `x86_64-darwin`, I went through each of these derivations and tried to build them directly. The only one that failed to build also failed to build on master: `nix-build -A python3Packages.pyopencl`.

I did not try to build on x86_64-linux yet.

emilazy (approved):
I hacked up glxgears to build on macOS and got a blazing 12445.432 FPS on XQuartz with this PR applied. LGTM

Ninlives:
This PR breaks mesa build on riscv, as the valgrind is currently not supported on riscv, and the option for removing it from the buildInputs is deleted in this PR.

PR #319053, opened 2024-06-11
python311Packages.rapidocr-onnxruntime: init at 1.3.22

### LINKED ISSUE #317065: Package request: Python3Packages.rapidocr_onnxruntime
**Project description**

Awesome OCR multiple programing languages toolkits based on ONNXRuntime, OpenVION and PaddlePaddle.

**Metadata**

* homepage URL: https://github.com/RapidAI/RapidOCR
* source URL: https://github.com/RapidAI/RapidOCR/tree/main/python/rapidocr_onnxruntime
* license: asl20
* platforms: linux, darwin

---

pluiedev:
## Description of changes

Fixes #317065

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/development/python-modules/rapidocr-onnxruntime/default.nix
+++ b/pkgs/development/python-modules/rapidocr-onnxruntime/default.nix
@@ -1,0 +1,129 @@
+{
+  lib,
+  stdenv,
+  buildPythonPackage,
+  fetchFromGitHub,
+
+  fetchzip,
+  substitute,
+  pythonRelaxDepsHook,
+  pytestCheckHook,
+
+  setuptools,
+  pyclipper,
+  opencv4,
+  numpy,
+  six,
+  shapely,
+  pyyaml,
+  pillow,
+  onnxruntime,
+}:
+let
+  version = "1.3.22";
+
+  src = fetchFromGitHub {
+    owner = "RapidAI";
+    repo = "RapidOCR";
+    rev = "v${version}";
+    hash = "sha256-8h4DFhnI9imr+bYQZdlrl8UKUdpwnGK+SGxLXSMmcag=";
+  };
+
+  models = fetchzip {
+    url = "https://github.com/RapidAI/RapidOCR/releases/download/v1.1.0/required_for_whl_v1.3.0.zip";
+    hash = "sha256-j/0nzyvu/HfNTt5EZ+2Phe5dkyPOdQw/OZTz0yS63aA=";
+    stripRoot = false;
+  } + "/required_for_whl_v1.3.0/resources/models";
+in
+buildPythonPackage {
+  pname = "rapidocr-onnxruntime";
+  inherit version src;
+  pyproject = true;
+
+  sourceRoot = "${src.name}/python";
+
+  # HACK:
+  # Upstream uses a very unconventional structure to organize the packages, and we have to coax the
+  # existing infrastructure to work with it.
+  # See https://github.com/RapidAI/RapidOCR/blob/02829ef986bc2a5c4f33e9c45c9267bcf2d07a1d/.github/workflows/gen_whl_to_pypi_rapidocr_ort.yml#L80-L92
+  # for the "intended" way of building this package.
+
+  # The setup.py supplied by upstream tries to determine the current version by
+  # fetching the latest version of the package from PyPI, and then bumping the version number.
+  # This is not allowed in the Nix build environment as we do not have internet access,
+  # hence we patch that out and get the version from the build environment directly.
+  patches = [
+    (substitute {
+      src = ./setup-py-override-version-checking.patch;
+      substitutions = [
+        "--subst-var-by"
+        "version"
+        version
+      ];
+    })
+  ];
+
+  postPatch = ''
+    mv setup_onnxruntime.py setup.py
+    mkdir -p rapidocr_onnxruntime/models
+
+    ln -s ${models}/* rapidocr_onnxruntime/models
+
+    # Magic patch from upstream - what does this even do??
+    echo "from .rapidocr_onnxruntime.main import RapidOCR, VisRes" > __init__.py
+  '';
+
+  nativeBuildInputs = [ pythonRelaxDepsHook ];
+
+  # Upstream expects the source files to be under rapidocr_onnxruntime/rapidocr_onnxruntime
+  # instead of rapidocr_onnxruntime for the wheel to build correctly.
+  preBuild = ''
+    mkdir rapidocr_onnxruntime_t
+    mv rapidocr_onnxruntime rapidocr_onnxruntime_t
+    mv rapidocr_onnxruntime_t rapidocr_onnxruntime
+  '';
+
+  # Revert the above hack
+  postBuild = ''
+    mv rapidocr_onnxruntime rapidocr_onnxruntime_t
+    mv rapidocr_onnxruntime_t/* .
+  '';
+
+  build-system = [ setuptools ];
+
+  dependencies = [
+    pyclipper
+    opencv4
+    numpy
+    six
+    shapely
+    pyyaml
+    pillow
+    onnxruntime
+  ];
+
+  # Remove because we have adopted the `opencv4` as an attribute name.
+  pythonRemoveDeps = [ "opencv-python" ];
+
+  pythonImportsCheck = [ "rapidocr_onnxruntime" ];
+
+  nativeCheckInputs = [ pytestCheckHook ];
+
+  # These are tests for different backends.
+  disabledTestPaths = [
+    "tests/test_vino.py"
+    "tests/test_paddle.py"
+  ];
+
+  meta = {
+    # This seems to be related to https://github.com/microsoft/onnxruntime/issues/10038
+    # Also some related issue: https://github.com/NixOS/nixpkgs/pull/319053#issuecomment-2167713362
+    broken = (stdenv.isLinux && stdenv.isAarch64);
+    changelog = "https://github.com/RapidAI/RapidOCR/releases/tag/v${version}";
+    description = "Cross platform OCR Library based on OnnxRuntime";
+    homepage = "https://github.com/RapidAI/RapidOCR";
+    license = with lib.licenses; [ asl20 ];
+    maintainers = with lib.maintainers; [ pluiedev ];
+    mainProgram = "rapidocr_onnxruntime";
+  };
+}
--- a/pkgs/development/python-modules/rapidocr-onnxruntime/setup-py-override-version-checking.patch
+++ b/pkgs/development/python-modules/rapidocr-onnxruntime/setup-py-override-version-checking.patch
@@ -1,0 +1,30 @@
+diff --git a/setup_onnxruntime.py b/setup_onnxruntime.py
+index cd698d1..c6e8a33 100644
+--- a/setup_onnxruntime.py
++++ b/setup_onnxruntime.py
+@@ -6,7 +6,6 @@ from pathlib import Path
+ from typing import List, Union
+ 
+ import setuptools
+-from get_pypi_latest_version import GetPyPiLatestVersion
+ 
+ 
+ def read_txt(txt_path: Union[Path, str]) -> List[str]:
+@@ -26,16 +25,7 @@ def get_readme():
+ 
+ MODULE_NAME = "rapidocr_onnxruntime"
+ 
+-obtainer = GetPyPiLatestVersion()
+-latest_version = obtainer(MODULE_NAME)
+-VERSION_NUM = obtainer.version_add_one(latest_version)
+-
+-if len(sys.argv) > 2:
+-    match_str = " ".join(sys.argv[2:])
+-    matched_versions = obtainer.extract_version(match_str)
+-    if matched_versions:
+-        VERSION_NUM = matched_versions
+-sys.argv = sys.argv[:2]
++VERSION_NUM = "@version@"
+ 
+ project_urls = {
+     "Documentation": "https://rapidai.github.io/RapidOCRDocs/docs/install_usage/rapidocr/usage/",
--- a/pkgs/top-level/python-packages.nix
+++ b/pkgs/top-level/python-packages.nix
@@ -13071,6 +13071,8 @@ self: super: with self; {
 
   rapidgzip = callPackage ../development/python-modules/rapidgzip { inherit (pkgs) nasm; };
 
+  rapidocr-onnxruntime = callPackage ../development/python-modules/rapidocr-onnxruntime { };
+
   rapt-ble = callPackage ../development/python-modules/rapt-ble { };
 
   rarfile = callPackage ../development/python-modules/rarfile {


drupol (commented):
I think it should be moved under `pkgs/development/python-modules`. This is not an application, this is a module.

natsukium (dismissed):
Please read the document first, especially the contributing guidelines section.
https://github.com/NixOS/nixpkgs/blob/master/doc/languages-frameworks/python.section.md#contributing-guidelines-contributing-guidelines
And if necessary.
https://github.com/NixOS/nixpkgs/blob/master/doc/languages-frameworks/python.section.md#building-packages-and-applications-building-packages-and-applications

pluiedev:
> Please read the document first, especially the contributing guidelines section. [`master`/doc/languages-frameworks/python.section.md#contributing-guidelines-contributing-guidelines](https://github.com/NixOS/nixpkgs/blob/master/doc/languages-frameworks/python.section.md?rgh-link-date=2024-06-12T01%3A04%3A07Z#contributing-guidelines-contributing-guidelines) And if necessary. [`master`/doc/languages-frameworks/python.section.md#building-packages-and-applications-building-packages-and-applications](https://github.com/NixOS/nixpkgs/blob/master/doc/languages-frameworks/python.section.md?rgh-link-date=2024-06-12T01%3A04%3A07Z#building-packages-and-applications-building-packages-and-applications)

I'm honestly not sure what I'm missing here — if you have any specific concerns, please just spell them out instead of making people guess

pkgs/development/python-modules/rapidocr-onnxruntime/default.nix
```diff
@@ -0,0 +1,48 @@
...
+  };
+in
+python3.pkgs.buildPythonPackage {
+  pname = "rapidocr-onnxruntime";
+  inherit version src;
+  pyproject = true;
+
+  build-system = with python3.pkgs; [ setuptools ];
+
+  dependencies = with python3.pkgs; [
```
drupol, line 23: Inherit those dependencies from top level parameters, not from `python3.pkgs`

pluiedev: I guess this is more a personal preference thing, but sure

pkgs/development/python-modules/rapidocr-onnxruntime/default.nix
```diff
@@ -0,0 +1,48 @@
...
+  version = "1.3.22";
+
+  src = fetchFromGitHub {
+    owner = "RapidAI";
+    repo = "RapidOCR";
+    rev = "v${version}";
+    hash = "sha256-8h4DFhnI9imr+bYQZdlrl8UKUdpwnGK+SGxLXSMmcag=";
+  };
+in
+python3.pkgs.buildPythonPackage {
```
drupol, line 16: I also think it needs a `pyImportCheck = [ "rapidocr_onnxruntime" ];`

natsukium:
> I'm honestly not sure what I'm missing here — if you have any specific concerns, please just spell them out instead of making people guess. Telling people to RTFM is not particularly productive IMO

Are you going to contribute without even reading the contribution guide?
I'm not asking you to read all the documents. Just a checklist.

pkgs/development/python-modules/rapidocr-onnxruntime/default.nix
```diff
@@ -0,0 +1,60 @@
...
+    pyyaml
+    pillow
+    onnxruntime
+  ];
+
+  sourceRoot = "${src.name}/python";
+  patches = [ ./setup-py-patch.patch ];
+
+  postPatch = ''
+    substituteInPlace requirements_ort.txt --replace-fail "opencv_python" "opencv"
```
natsukium, line 48:
```suggestion
```
Please use `pythonRelaxDepsHook` instead.
> The version restrictions of dependencies can be relaxed by [pythonRelaxDepsHook](https://github.com/NixOS/nixpkgs/blob/master/doc/languages-frameworks/python.section.md#using-pythonrelaxdepshook).

https://github.com/NixOS/nixpkgs/blob/master/doc/languages-frameworks/python.section.md#contributing-contributing

pkgs/development/python-modules/rapidocr-onnxruntime/default.nix
```diff
@@ -0,0 +1,60 @@
...
+    mv setup_onnxruntime.py setup.py
+  '';
+
+  pyImportCheck = [ "rapidocr_onnxruntime" ];
```
natsukium, line 52:
```suggestion
  pythonImportsCheck = [ "rapidocr_onnxruntime" ];
```
> pythonImportsCheck is set. This is still a good smoke test even if pytestCheckHook is set.

https://github.com/NixOS/nixpkgs/blob/master/doc/languages-frameworks/python.section.md#contributing-contributing

pkgs/development/python-modules/rapidocr-onnxruntime/default.nix
```diff
@@ -0,0 +1,60 @@
...
+
```
natsukium, line 53:
```suggestion
  nativeCheckInputs = [ pytestCheckHook ];
```
> Make sure the tests are enabled using for example [pytestCheckHook](https://github.com/NixOS/nixpkgs/blob/master/doc/languages-frameworks/python.section.md#using-pytestcheckhook) and, in the case of libraries, are passing for all interpreters. If certain tests fail they can be disabled individually. Try to avoid disabling the tests altogether. In any case, when you disable tests, leave a comment explaining why.

https://github.com/NixOS/nixpkgs/blob/master/doc/languages-frameworks/python.section.md#contributing-contributing

pkgs/development/python-modules/rapidocr-onnxruntime/default.nix
```diff
@@ -0,0 +1,60 @@
...
+    shapely
+    pyyaml
+    pillow
+    onnxruntime
+  ];
+
+  sourceRoot = "${src.name}/python";
+  patches = [ ./setup-py-patch.patch ];
+
+  postPatch = ''
```
natsukium, line 47: Everything related to patches should be near the `src`.
It doesn't matter where it is as a processor, but the cognitive load should be low since humans do the maintenance.

natsukium:
The commit message and PR title should be `python311Packages.rapidocr-onnxruntime`.
> Commit names of Python libraries should reflect that they are Python libraries, so write for example python311Packages.numpy: 1.11 -> 1.12. It is highly recommended to specify the current default version to enable automatic build by ofborg. Note that pythonPackages is an alias for python27Packages.

https://github.com/NixOS/nixpkgs/blob/master/doc/languages-frameworks/python.section.md#contributing-contributing

pkgs/development/python-modules/rapidocr-onnxruntime/default.nix
drupol: Oh my bad. Sorry for the misleading information.

pluiedev:
> > I'm honestly not sure what I'm missing here — if you have any specific concerns, please just spell them out instead of making people guess. Telling people to RTFM is not particularly productive IMO
> 
> Are you going to contribute without even reading the contribution guide? I'm not asking you to read all the documents. Just a checklist.

Look, I have packaged Python packages before, and while I'm not quite up to speed on what's the expected standard of today, that's not a reason to just throw a big document at me and say "fix it." Please, just state problems directly.

pkgs/development/python-modules/rapidocr-onnxruntime/default.nix
pluiedev: This assumes the package requirements to be at `requirements.txt`, right? That doesn't work since the upstream file is at `requirements-ort.txt`.

pkgs/development/python-modules/rapidocr-onnxruntime/default.nix
```diff
@@ -0,0 +1,70 @@
...
+    hash = "sha256-8h4DFhnI9imr+bYQZdlrl8UKUdpwnGK+SGxLXSMmcag=";
+  };
+in
+buildPythonPackage {
+  pname = "rapidocr-onnxruntime";
+  inherit version src;
+  pyproject = true;
+
+  sourceRoot = "${src.name}/python";
+  patches = [ ./setup-py-patch.patch ];
```
drupol, line 35: I think the patch will require a comment explaining why it is needed.

pluiedev: That's true — it's just that the script by default pulls the current version from PyPI and bumps it, which is obviously not allowed in the Nix sandbox

pluiedev: Might as well rename the patch to something more descriptive I think — the current name is very vague

pluiedev:
Ugh, looks like the tests use some models that are not bundled with the Git repo — how are you even supposed to get them, then?

drupol:
> ModuleNotFoundError: No module named 'rapidocr_onnxruntime'

@natsukium Do you have an idea on how we could fix this?

pluiedev:
> > ModuleNotFoundError: No module named 'rapidocr_onnxruntime'
> 
> @natsukium Do you have an idea on how we could fix this?

I suspect that the wheel isn't actually built correctly unless the models are bundled with it — I tried adding some of the models and the build process actually added those to the wheel, so that might be the reason why

pluiedev:
Well, adding the models didn't fix the import check but at least now all tests work :tada:

drupol:
Stupid question, what's `wheel` in this context?

pluiedev:
> Stupid question, what's `wheel` in this context?

AFAICT a wheel is just the built binary Python package, in this case `rapidocr_onnxruntime-1.3.22-py3-none-any.whl`

pkgs/development/python-modules/rapidocr-onnxruntime/default.nix
```diff
@@ -0,0 +1,95 @@
...
+  version = "1.3.22";
+
+  src = fetchFromGitHub {
+    owner = "RapidAI";
+    repo = "RapidOCR";
+    rev = "v${version}";
+    hash = "sha256-8h4DFhnI9imr+bYQZdlrl8UKUdpwnGK+SGxLXSMmcag=";
+  };
+
+  models = fetchgit {
```
drupol, line 31: I'm wondering if this is a good idea...

pkgs/development/python-modules/rapidocr-onnxruntime/default.nix
```diff
@@ -0,0 +1,95 @@
...
+    pyyaml
+    pillow
+    onnxruntime
+  ];
+
+  # Don't use vendored OpenCV
+  pythonRemoveDeps = [ "opencv-python" ];
+
+  # FIXME: Why isn't this working???
+  #pythonImportsCheck = [ "rapidocr_onnxruntime" ];
```
drupol, line 73: I guess this will have to be fixed before merging. What's the error log?

pluiedev: Same as before: ModuleNotFoundError: No module named 'rapidocr_onnxruntime'.
Though, I think I found a reason for that — somehow the `__init__,py` and `main.py` are just completely missing from the built wheel. Given this project uses a very unconventional layout, I wonder if we're missing something again...

natsukium: This is a blocker.
```shell
$ nix-shell -p 'python311.withPackages (ps: with ps; [rapidocr-onnxruntime])' -I nixpkgs=. --run 'python -c "import rapidocr_onnxruntime"'

  File "<string>", line 1, in <module>
ModuleNotFoundError: No module named 'rapidocr_onnxruntime'
```

pluiedev: Well, of course it's a blocker, as it doesn't even work :sweat_smile:

pkgs/development/python-modules/rapidocr-onnxruntime/default.nix
natsukium: If it is used for testing only, it is not necessary.

pkgs/development/python-modules/rapidocr-onnxruntime/default.nix
natsukium: Regarding `__init__.py`, we need to add the following command to `postPatch`.
```
echo "from .rapidocr_onnxruntime.main import RapidOCR, VisRes" > rapidocr_onnxruntime/__init__.py
```
https://github.com/RapidAI/RapidOCR/blob/02829ef986bc2a5c4f33e9c45c9267bcf2d07a1d/.github/workflows/gen_whl_to_pypi_rapidocr_ort.yml#L86-L87

pluiedev: That doesn't quite work with the current setup — tests fail:
```
       > ______________________ ERROR collecting tests/test_ort.py ______________________
       > ImportError while importing test module '/build/source/python/tests/test_ort.py'.
       > Hint: make sure your test modules/packages have valid Python names.
       > Traceback:
       > /nix/store/7hnr99nxrd2aw6lghybqdmkckq60j6l9-python3-3.11.9/lib/python3.11/importlib/__init__.py:126: in import_module
       >     return _bootstrap._gcd_import(name[level:], package, level)
       > tests/test_ort.py:16: in <module>
       >     from rapidocr_onnxruntime import LoadImageError, RapidOCR
       > rapidocr_onnxruntime/__init__.py:1: in <module>
       >     from .rapidocr_onnxruntime.main import RapidOCR, VisRes
       > E   ModuleNotFoundError: No module named 'rapidocr_onnxruntime.rapidocr_onnxruntime'
```

Or with `from .rapidocr_onnxruntime.main import RapidOCR, VisRes`:
```
       > ______________________ ERROR collecting tests/test_ort.py ______________________
       > ImportError while importing test module '/build/source/python/tests/test_ort.py'.
       > Hint: make sure your test modules/packages have valid Python names.
       > Traceback:
       > /nix/store/7hnr99nxrd2aw6lghybqdmkckq60j6l9-python3-3.11.9/lib/python3.11/importlib/__init__.py:126: in import_module
       >     return _bootstrap._gcd_import(name[level:], package, level)
       > tests/test_ort.py:16: in <module>
       >     from rapidocr_onnxruntime import LoadImageError, RapidOCR
       > E   ImportError: cannot import name 'LoadImageError' from 'rapidocr_onnxruntime' (/build/source/python/rapidocr_onnxruntime/__init__.py)
```

natsukium: I suspect the `setuptools` namespace is causing the problem.
https://github.com/RapidAI/RapidOCR/blob/02829ef986bc2a5c4f33e9c45c9267bcf2d07a1d/python/setup_onnxruntime.py#L59-L60

https://setuptools.pypa.io/en/latest/userguide/package_discovery.html

drupol: > Regarding `__init__.py`, we need to add the following command to `postPatch`.
> 
> ```
> echo "from .rapidocr_onnxruntime.main import RapidOCR, VisRes" > rapidocr_onnxruntime/__init__.py
> ```
> 
> https://github.com/RapidAI/RapidOCR/blob/02829ef986bc2a5c4f33e9c45c9267bcf2d07a1d/.github/workflows/gen_whl_to_pypi_rapidocr_ort.yml#L86-L87

Why god why are they doing that ?!

pkgs/development/python-modules/rapidocr-onnxruntime/default.nix
```diff
@@ -0,0 +1,107 @@
...
+  # hence we patch that out and get the version from the build environment directly.
+  patches = [ ./setup-py-override-version-checking.patch ];
+
+  nativeBuildInputs = [
+    pythonRelaxDepsHook
+    pytestCheckHook
+  ];
+
+  build-system = [ setuptools ];
+
```
natsukium, line 70:
```suggestion
  preBuild = ''
    mkdir rapidocr_onnxruntime_t
    mv rapidocr_onnxruntime rapidocr_onnxruntime_t
    mv rapidocr_onnxruntime_t rapidocr_onnxruntime
  '';
```
I don't know what they want to do.
`pythonImportsCheck` now passes, but it still seems to be missing some files during the test.

https://github.com/RapidAI/RapidOCR/blob/02829ef986bc2a5c4f33e9c45c9267bcf2d07a1d/.github/workflows/gen_whl_to_pypi_rapidocr_ort.yml#L83-L85

pkgs/development/python-modules/rapidocr-onnxruntime/default.nix
```diff
@@ -0,0 +1,107 @@
...
+  # The setup.py supplied by upstream tries to determine the current version by
+  # fetching the latest version of the package from PyPI, and then bumping the version number.
+  # This is not allowed in the Nix build environment as we do not have internet access,
+  # hence we patch that out and get the version from the build environment directly.
+  patches = [ ./setup-py-override-version-checking.patch ];
+
+  nativeBuildInputs = [
+    pythonRelaxDepsHook
+    pytestCheckHook
```
natsukium, line 66: `pytestCheckHook` should go to `nativeCheckInputs`.

pkgs/development/python-modules/rapidocr-onnxruntime/default.nix
natsukium: before
```shell
$ cat /tmp/nix-build-python3.11-rapidocr-onnxruntime-1.3.22.drv-2/source/python/rapidocr_onnxruntime/__init__.py 
from .rapidocr_onnxruntime.main import RapidOCR, VisRes
```

after
```shell
$ cat /tmp/nix-build-python3.11-rapidocr-onnxruntime-1.3.22.drv-3/source/python/rapidocr_onnxruntime/rapidocr_onnxruntime/__init__.py 
# -*- encoding: utf-8 -*-
# @Author: SWHL
# @Contact: liekkaskono@163.com
from .main import RapidOCR
from .utils import LoadImageError, VisRes
```

pkgs/development/python-modules/rapidocr-onnxruntime/default.nix
```diff
@@ -0,0 +1,107 @@
...
+    pyyaml
+    pillow
+    onnxruntime
+  ];
+
+  # Don't use vendored OpenCV
+  pythonRemoveDeps = [ "opencv-python" ];
+
+  #pythonImportsCheck = [ "rapidocr_onnxruntime" ];
+
```
natsukium, line 86:
```suggestion
  preCheck = ''
    mv rapidocr_onnxruntime rapidocr_onnxruntime_t
    mv rapidocr_onnxruntime_t/* .
  '';
```
It's a dirty hack.
We can create a directory structure `rapidocr_onnxruntime/rapidocr_onnxruntime` before the build and restore it after the build.

natsukium:
In short, we need three things.
1. create `__init__.py` in `postPatch` phase [ref](https://github.com/NixOS/nixpkgs/pull/319053#discussion_r1637848253)
2. move `rapidocr_onnxruntime` to `rapidocr_onnxruntime/rapidocr_onnxruntime` in `preBuild` phase [ref](https://github.com/NixOS/nixpkgs/pull/319053#discussion_r1637927181)
3. revert the operation 2 in `preCheck` phase (maybe `postBuild` phase is ok) [ref](https://github.com/NixOS/nixpkgs/pull/319053#discussion_r1637955272)

pkgs/development/python-modules/rapidocr-onnxruntime/default.nix
pluiedev: > > Regarding `__init__.py`, we need to add the following command to `postPatch`.
> > ```
> > echo "from .rapidocr_onnxruntime.main import RapidOCR, VisRes" > rapidocr_onnxruntime/__init__.py
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > [RapidAI/RapidOCR@`02829ef`/.github/workflows/gen_whl_to_pypi_rapidocr_ort.yml#L86-L87](https://github.com/RapidAI/RapidOCR/blob/02829ef986bc2a5c4f33e9c45c9267bcf2d07a1d/.github/workflows/gen_whl_to_pypi_rapidocr_ort.yml#L86-L87)
> 
> Why god why are they doing that ?!

Beats me... This package is just built in an unorthodox way in every sense of the word.

natsukium:
In general, machine learning engineers are only interested in their own environment.
I know this well because I am also a machine learning engineer.

pluiedev:
Right, makes sense... should we mark this ready for review then?

natsukium:
Yes, please.

pkgs/development/python-modules/rapidocr-onnxruntime/default.nix
```diff
@@ -0,0 +1,134 @@
...
+  # The setup.py supplied by upstream tries to determine the current version by
+  # fetching the latest version of the package from PyPI, and then bumping the version number.
+  # This is not allowed in the Nix build environment as we do not have internet access,
+  # hence we patch that out and get the version from the build environment directly.
+  patches = [ ./setup-py-override-version-checking.patch ];
+
+  nativeBuildInputs = [ pythonRelaxDepsHook ];
+
+  nativeCheckInputs = [ pytestCheckHook ];
```
natsukium, line 70: Could you move this between `pythonImportsCheck` and `disabledTestPaths`?

pkgs/development/python-modules/rapidocr-onnxruntime/default.nix
```diff
@@ -0,0 +1,134 @@
...
+    numpy
+    six
+    shapely
+    pyyaml
+    pillow
+    onnxruntime
+  ];
+
+  # Don't use vendored OpenCV
+  pythonRemoveDeps = [ "opencv-python" ];
```
natsukium, line 86: `opencv-python` is not a vendor package, it is a standard package called opencv-python in pypi.
We need `pythonRemoveDeps` because we have adopted the `opencv4` as an attribute name.https://pypi.org/project/opencv-python/

pkgs/development/python-modules/rapidocr-onnxruntime/default.nix
```diff
@@ -0,0 +1,134 @@
...
+  postPatch = ''
+    mv setup_onnxruntime.py setup.py
+
+    mkdir -p rapidocr_onnxruntime/models
+    # Flatten the structure so that all models are placed in the same directory
+    find ${models} -name "*.onnx" -exec ln -st rapidocr_onnxruntime/models {} +
+
+    # Magic patch from upstream - what does this even do??
+    echo "from .rapidocr_onnxruntime.main import RapidOCR, VisRes" > __init__.py
+  '';
```
natsukium, line 112: Could you move `postPatch` after `patches`?

pkgs/development/python-modules/rapidocr-onnxruntime/default.nix
```diff
@@ -0,0 +1,134 @@
...
+    mkdir rapidocr_onnxruntime_t
+    mv rapidocr_onnxruntime rapidocr_onnxruntime_t
+    mv rapidocr_onnxruntime_t rapidocr_onnxruntime
+  '';
+
+  # Revert the above hack
+  postBuild = ''
+    mv rapidocr_onnxruntime rapidocr_onnxruntime_t
+    mv rapidocr_onnxruntime_t/* .
+  '';
```
natsukium, line 126: These are related to the build phase, so it is preferred to be near `nativeBuildInputs` or `build-system`.

pkgs/development/python-modules/rapidocr-onnxruntime/default.nix
```diff
@@ -0,0 +1,134 @@
...
+
+  meta = {
+    description = "Cross platform OCR Library based on OnnxRuntime";
+    homepage = "https://github.com/RapidAI/RapidOCR";
```
natsukium, line 130:
```suggestion
    homepage = "https://github.com/RapidAI/RapidOCR";
    changelog = "https://github.com/RapidAI/RapidOCR/releases/tag/v${version}";
```
Adding meta.changelog is useful for future updates.

drupol:
@pluiedev Hope you don't mind me pushing a commit, it is mainly for helping you. I didn't rebased it myself so it's easier to pickup.

pkgs/development/python-modules/rapidocr-onnxruntime/default.nix
drupol: Fixed.

pkgs/development/python-modules/rapidocr-onnxruntime/default.nix
drupol: Fixed.

pkgs/development/python-modules/rapidocr-onnxruntime/default.nix
drupol: Fixed.

pkgs/development/python-modules/rapidocr-onnxruntime/default.nix
drupol: Fixed.

pkgs/development/python-modules/rapidocr-onnxruntime/default.nix
drupol: Fixed.

natsukium:
on aarch64-linux🤔
something is wrong with the import check.
https://github.com/NixOS/nixpkgs/pull/319053/checks?check_run_id=26206549989

drupol:
```
       > Error in cpuinfo: failed to parse the list of possible processors in /sys/devices/system/cpu/possible
       > Error in cpuinfo: failed to parse the list of present processors in /sys/devices/system/cpu/present
       > Error in cpuinfo: failed to parse both lists of possible and present processors
       > terminate called after throwing an instance of 'onnxruntime::OnnxRuntimeException'
```

pluiedev:
Can repro by running `nix build .#legacyPackages.aarch64-linux.python3Packages.rapidocr-onnxruntime`. This looks like perhaps an OnnxRuntime issue...?

pluiedev:
With further digging I realized what's going wrong here — on aarch64, ONNX Runtime uses pytorch's cpuinfo to look for CPU core configuration, by looking at info provided by `/sys/devices/system/cpu/{present,possible}`. However, that information is not exposed in the Nix builder, possibly to disallow builders from relying on the build machine's specific hardware configuration that may result in software that doesn't work for other systems. See pytorch/cpuinfo#143, microsoft/onnxruntime#10038.

In my mind there's two ways going forward — either mark aarch64 builds as broken, or patch ONNX Runtime to build without using cpuinfo by disabling the CMake flag [`onnxruntime_ENABLE_CPUINFO`](https://github.com/microsoft/onnxruntime/blob/983fd8393aa30b0c788275ea6e513a614274e1ba/cmake/CMakeLists.txt#L229).

drupol:
IMHO, we should disable [onnxruntime_ENABLE_CPUINFO](https://github.com/microsoft/onnxruntime/blob/983fd8393aa30b0c788275ea6e513a614274e1ba/cmake/CMakeLists.txt#L229) on `aarch64`. Anyone has another opinion?

pluiedev:
Agreed — working slowly is better than not working at all! :P

pluiedev:
I did some more work to package the OpenVINO and PaddlePaddle versions of this package too — wasn't that hard

pkgs/development/python-modules/rapidocr/default.nix
```diff
@@ -0,0 +1,158 @@
+backendId:
```
drupol, line 1: I don't think this is allowed in `nixpkgs`.

pluiedev: Perhaps, or just don't supply a default to make it fail if it's not called via callPackage with the right arg

drupol: @natsukium What do you recommend here?

natsukium: We should have one file per package.
The python library is often broken, so we set attributes like `disabled` and `meta.broken`, and I don't want to put another condition, backend. `paddlepaddle` is an example.

Monorepo packages, for example `opentelemetry`, could share the same version, src, etc.
It may be more duplicated than the current programmatic one, and it may be uglier, depending on your preference, but for future maintenance, this is the way to go.
https://github.com/NixOS/nixpkgs/blob/nixos-unstable/pkgs/development/python-modules/opentelemetry-api/default.nix
https://github.com/NixOS/nixpkgs/blob/nixos-unstable/pkgs/development/python-modules/opentelemetry-sdk/default.nix

pluiedev: Honestly the approach showed above isn't very DRY and requires a lot of duplicated code, or sourcing code from one place only to change it later. Since 95% of the derivation is exactly the same across all backends, I find that approach adds a needless burden on maintainers who have to make changes to all three runtimes once upstream changes something.

drupol: How about merging `rapidocr-onnxruntime` for the moment which seems to be in good state, and decide what to do after that? I'm badly looking for that package in `open-webui` actually.

pluiedev: Sure!

drupol:
Should we patch `onnxruntime` for `aarch64-linux` ?

drupol:
@natsukium @pluiedev Do you think it's ready for merge or... ?

drupol:
@ofborg build python3Packages.rapidocr-onnxruntime

drupol:
OK Planning to merge this in 1 or 2 hours.

PR #319087, opened 2024-06-11
doc/passthru: improve sanctioned use of installCheckPhase

pbsds:
## Description of changes

I only found https://github.com/NixOS/nixpkgs/pull/315616 after it was merged. Here are my two cents.

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/doc/stdenv/passthru.chapter.md
+++ b/doc/stdenv/passthru.chapter.md
@@ -83,7 +83,7 @@ Most programs distributed by Nixpkgs support such a `--version` flag, and succes
 
 ## Checking builds with `installCheckPhase`
 
-When building `git`, a rudimentary test for successful compilation would be running `git --version`:
+When building `git`, a rudimentary test for successful compilation would be running `git --version` with a clean environment:
 
 ```nix
 stdenv.mkDerivation (finalAttrs: {
@@ -94,19 +94,21 @@ stdenv.mkDerivation (finalAttrs: {
   installCheckPhase = ''
     runHook preInstallCheck
     echo checking if 'git --version' mentions ${finalAttrs.version}
-    $out/bin/git --version | grep ${finalAttrs.version}
+    env --ignore-environment $out/bin/git --version | grep --silent "${finalAttrs.version}"
     runHook postInstallCheck
   '';
   # ...
 })
 ```
+
+Testing with a clean environment is encouraged to ensure the `bin/` entrypoints are properly wrapped and self-contained, and do not depend on variables like PYTHONPATH and PERL5LIB being configured.
 :::
 
 However, tests that are non-trivial will better fit into `passthru.tests` because they:
 
-- Access the package as consumers would, independently from the environment in which it was built
+- Access the package outputs as any consumer would, with an environment as setup up with either `stdenv` or `environment.systemPackages`, and without temporary build artifacts and state produced in `configurePhase` and `buildPhase`.
 - Can be run and debugged without rebuilding the package, which is useful if that takes a long time
-- Don't add overhad to each build, as opposed to `installCheckPhase`
+- Don't add overhead to each build, as opposed to `installCheckPhase`
 
 It is also possible to use `passthru.tests` to test the version with [`testVersion`](#tester-testVersion).
 


doc/stdenv/passthru.chapter.md
```diff
@@ -94,7 +94,7 @@ stdenv.mkDerivation (finalAttrs: {
   installCheckPhase = ''
     runHook preInstallCheck
     echo checking if 'git --version' mentions ${finalAttrs.version}
-    $out/bin/git --version | grep ${finalAttrs.version}
+    env --ignore-environment $out/bin/git --version | grep --silent "${finalAttrs.version}"
```
abathur, line 97: Is this change solving for a specific failure mode in the example, or is it just demonstrating one way to guard against state inside the build environment in order to make a general point about the fact that tests run during the build may be sensitive to temporary state that won't exist after the build completes?

If it's the former, I think it might be better to just lay the cards down here and show the problem we'd have without this workaround (because it will provide a better illustration of the underlying risk of these tests depending on the state inside the sandbox) and also show the fix?

If this just demonstrates how to guard against state without an acute need, I think it's still worth explaining why a "clean" environment is important (and perhaps noting that this only guards against environment variables which are only one kind of state present in  the sandbox).

doc/stdenv/passthru.chapter.md
```diff
@@ -104,9 +104,9 @@ stdenv.mkDerivation (finalAttrs: {
 
 However, tests that are non-trivial will better fit into `passthru.tests` because they:
 
-- Access the package as consumers would, independently from the environment in which it was built
+- Access the package as consumers would, e.g. the environment setup up the way `mkDerivation` or `environment.systemPackages` would.
```
raboof, line 107: I'm not sure `mkDerivation` makes sense to me here

raboof (commented):
Not sure about the 'as consumers would' description, otherwise nice improvements!

doc/stdenv/passthru.chapter.md
pbsds: `s/mkDerivation/stdenv/`?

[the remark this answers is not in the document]
pbsds: Added a explanation to the _note_

doc/stdenv/passthru.chapter.md
pbsds: This is specifically to avoid variables like `PYTHONPATH`, to guard against build artifacts one could `pushd "$(mkdtemp -d)"`

doc/stdenv/passthru.chapter.md
```diff
@@ -94,19 +94,21 @@ stdenv.mkDerivation (finalAttrs: {
...
 })
 ```
+
+Testing with a clean environment is encouraged to ensure the `bin/` entrypoints are properly wrapped and self-contained, and do not depend on variables like PYTHONPATH and PERL5LIB being configured.
 :::
 
 However, tests that are non-trivial will better fit into `passthru.tests` because they:
 
-- Access the package as consumers would, independently from the environment in which it was built
+- Access the package outputs as any consumer would, with an environment as setup up with either `stdenv` or `environment.systemPackages`, and without temporary build artifacts and state produced in `configurePhase` and `buildPhase`.
```
abathur, line 109:
```suggestion
- Access the package outputs as any consumer would, with an environment set up with either `stdenv` or `environment.systemPackages`, and without temporary build artifacts and state produced in `configurePhase` and `buildPhase`.
```

Just fixes. Generally like this better, though I'm still personally a little unsure about the "with an environment set up with either `stdenv` or `environment.systemPackages`" bit: 
- Is it clarifying something specific beyond what "as any consumer would" already implies?
- I'm not sure I grok how the `environment.systemPackages` bit fits--is there any testing context in which a passthru test would run with all of the user's systemPackages in scope? (That seems undesirable and a sure source of flakiness, unless maybe it's gesturing at grafting in NixOS VM tests? If the latter, it's worth clarifying.)
- It reads like a complete list, but I imagine it's leaving out cases?

pbsds: The reason I mention those two means of consuming a package is because of the various hooks that run through the packages and ensure their propagated inputs are also added  and set up in various ways. It should read as non-exhausive.

The 

how about

```suggestion
- Access the package outputs as any consumer would, for example like the environments set up when using `nix-shell`, `stdenv`, or `environment.systemPackages`, and without the temporary build artifacts and state produced in `configurePhase` and `buildPhase`.
```

doronbehar:
Maybe we should create a `versionInstallCheckHook` that will take care of all the details that you try to explain here, and we can instruct users to simply read the shell code of the hook if they want to know the details of how we ignore the environment etc.

pbsds:
closing in favor of https://github.com/NixOS/nixpkgs/pull/320266

PR #319137, opened 2024-06-11
onnxruntime: 1.16.3 -> 1.18.0

jkachmar:
## Description of changes

version bump; release notes for versions between 1.16.3 & 1.18.0:
- https://github.com/microsoft/onnxruntime/releases/tag/v1.17.0
- https://github.com/microsoft/onnxruntime/releases/tag/v1.17.1
- https://github.com/microsoft/onnxruntime/releases/tag/v1.17.3
- https://github.com/microsoft/onnxruntime/releases/tag/v1.18.0

i haven't tested any of the CUDA stuff.

maintainers: @puffnfresh @ck3d @cbourjau

possibly interested parties from https://github.com/NixOS/nixpkgs/pull/309015: @wexder @madsmtm

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---


pkgs/development/libraries/onnxruntime/default.nix
```diff
@@ -6,6 +6,9 @@
 , abseil-cpp
 , cmake
 , eigen
+, flatbuffers
```
jkachmar, line 9: i'm surprised `flatbuffers` works given that it should be a clear version mismatch

pkgs/development/libraries/onnxruntime/default.nix
```diff
@@ -6,6 +6,9 @@
...
+, gbenchmark
+, glibcLocales
```
jkachmar, line 11: need to support `#include <locale>` in `onnxruntime/core/providers/cpu/text/string_normalizer.cc` (& probably elsewhere).

without this a bunch of `ContribOpTest`s fail (`StringNormalizer*` & `WordConvEmbedding*`)

pkgs/development/libraries/onnxruntime/default.nix
```diff
@@ -59,30 +62,36 @@ let
...
+    pname = "clog";
+    version = "3c8b153";
+    src = "${pytorch_cpuinfo}/deps/clog";
+
+    nativeBuildInputs = [ cmake gbenchmark gtest ];
+    cmakeFlags = [
+      "-DUSE_SYSTEM_GOOGLEBENCHMARK=ON"
+      "-DUSE_SYSTEM_GOOGLETEST=ON"
+      "-DUSE_SYSTEM_LIBS=ON"
+    ];
```
jkachmar, line 79: probably cleaner to factor out `pytorch_cpuinfo` rev & share that between the source & the `clog` version string.

also even though `pytorch_clog` doesn't seem to use `gbenchmark` it'll fail without `-DUSE_SYSTEM_GOOGLEBENCHMARK=ON` so we have to leave it there

pkgs/development/libraries/onnxruntime/default.nix
```diff
@@ -171,7 +182,8 @@ effectiveStdenv.mkDerivation rec {
     "-DFETCHCONTENT_QUIET=OFF"
     "-DFETCHCONTENT_SOURCE_DIR_ABSEIL_CPP=${abseil-cpp.src}"
     "-DFETCHCONTENT_SOURCE_DIR_DATE=${howard-hinnant-date}"
-    "-DFETCHCONTENT_SOURCE_DIR_FLATBUFFERS=${flatbuffers}"
+    "-DFETCHCONTENT_SOURCE_DIR_FLATBUFFERS=${flatbuffers.src}"
+    "-DFETCHCONTENT_SOURCE_DIR_GOOGLETEST=${gtest.src}"
```
jkachmar, line 186: i feel like `flatbuffers` should fail given issues like https://github.com/microsoft/onnxruntime/issues/14981; maybe the local build i tested out was flawed and it's actually not gonna work in CI

pkgs/development/libraries/onnxruntime/nvcc-gsl.patch
```diff
@@ -1,8 +1,8 @@
 diff --git a/cmake/external/onnxruntime_external_deps.cmake b/cmake/external/onnxruntime_external_deps.cmake
-index 9effd1a2db..faff5e8de7 100644
+index 775576a..ccea13c 100644
```
jkachmar, line 2: i think this patch might be unnecessary, given the comment in `onnxruntime/default.nix` on the line where this patch is included

pkgs/development/libraries/onnxruntime/default.nix
```diff
@@ -59,30 +62,36 @@ let
...
   };
 
   onnx = fetchFromGitHub {
     owner = "onnx";
     repo = "onnx";
-    rev = "refs/tags/v1.14.1";
-    hash = "sha256-ZVSdk6LeAiZpQrrzLxphMbc1b3rNUMpcxcXPP8s/5tE=";
+    rev = "refs/tags/v1.16.1";
+    hash = "sha256-I1wwfn91hdH3jORIKny0Xc73qW2P04MjkVCgcaNnQUE=";
   };
```
jkachmar, line 87: why are we vendoring `onnx`? it looks like there's an open MR to bump to 1.16 which would be compatible here https://github.com/NixOS/nixpkgs/pull/299313

jkachmar: @acairncross, since you're listed as the `onnx` maintainer

pkgs/development/libraries/onnxruntime/0001-eigen-allow-dependency-injection.patch
```diff
@@ -1,45 +1,20 @@
...
+ 
 -    FetchContent_Populate(eigen)
 -    set(eigen_INCLUDE_DIRS  "${eigen_SOURCE_DIR}")
 +    FetchContent_MakeAvailable(eigen)
 +    add_library(eigen ALIAS Eigen3::Eigen)
-+
 +    # Onnxruntime doesn't always use `eigen` as a target in
 +    # `target_link_libraries`, sometimes it just uses
 +    # `target_include_directories`:
 +    get_target_property(eigen_INCLUDE_DIRS Eigen3::Eigen INTERFACE_INCLUDE_DIRECTORIES)
```
jonringer, line 19: I really need to upstream a `-DPREFER_SYSTEM_LIBS` type of switch..... tired of having to carry around patches which disable network behaviors.

pkgs/development/libraries/onnxruntime/default.nix
jonringer: IIRC, onnxruntime is loosely correlated to onnx (the python package) so the don't have to be in lockstep. Although, I would probably prefer that we try to move to a world where we don't vendor a local copy.

jkachmar: so i can open a PR updating `onnx` to 1.16.1 & then reference the source here; is there a preference for how to structure that?

i guess `python3Packages` is already an argument, so it could just be `python3Packages.onnx_1_16.src` & some comment or assertion indicating that we want compatibility checks in a similar fashion to how other packages with source dependencies like this are structured?

jonringer: There's a tradeoff, as now if you bump onnx, you have to be concerned about onnxruntime. And it's a matter of if that should be the case.

jkachmar:
okay looks like there are some failures with `clog` so i'm going to have to actually test this out more on nixpkgs-24.05:
```
make[2]: *** [CMakeFiles/clog-test.dir/build.make:76: CMakeFiles/clog-test.dir/test/clog.cc.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:111: CMakeFiles/clog-test.dir/all] Error 2
make: *** [Makefile:146: all] Error 2
error: builder for '/nix/store/2kh962g19v3yglcadnjy2i71yaabivzz-clog-3c8b153.drv' failed with exit code 2;
       last 10 log lines:
       > /nix/store/8l47dmnifglj4lrwxsjf08jf1qzmbh1m-gtest-1.14.0-dev/include/gtest/gtest.h: At global scope:
       > /nix/store/8l47dmnifglj4lrwxsjf08jf1qzmbh1m-gtest-1.14.0-dev/include/gtest/gtest.h:302:30: error: 'std::enable_if_t' has not been declared
       >   302 |   template <typename T, std::enable_if_t<std::is_convertible<T, int64_t>::value,
       >       |                              ^~~~~~~~~~~
       > /nix/store/8l47dmnifglj4lrwxsjf08jf1qzmbh1m-gtest-1.14.0-dev/include/gtest/gtest.h:302:41: error: expected '>' before '<' token
       >   302 |   template <typename T, std::enable_if_t<std::is_convertible<T, int64_t>::value,
       >       |                                         ^
       > make[2]: *** [CMakeFiles/clog-test.dir/build.make:76: CMakeFiles/clog-test.dir/test/clog.cc.o] Error 1
       > make[1]: *** [CMakeFiles/Makefile2:111: CMakeFiles/clog-test.dir/all] Error 2
       > make: *** [Makefile:146: all] Error 2
       For full logs, run 'nix log /nix/store/2kh962g19v3yglcadnjy2i71yaabivzz-clog-3c8b153.drv'.
```

jonringer:
maybe a CXX_STANDARD issue?

jkachmar:
ah yeah might be; `clog` tests set CXX 11. i disabled that & it looks like there's another error with this:

```
CMake Error: install(EXPORT "abslTargets" ...) includes target "test_allocator" which requires target "gmock" that is not in any export set.
```

...which i think is resolved by applying [this patch](https://github.com/microsoft/onnxruntime/commit/b522df0ae477e59f60acbe6c92c8a64eda96cace) from upstream which isn't in 1.18.0...?

i guess i didn't run into this at work b/c we're on older versions of `gtest` & `abseil-cpp` (from 23.11); rebuilding `onnxruntime` & i'll push changes if it passes i guess.

pkgs/development/libraries/onnxruntime/default.nix
vcunat: Well, see https://github.com/NixOS/nixpkgs/pull/330416

PR #317755, opened 2024-06-06
 BuildRustCrate: proc macros must be built for build's platform 

thehabbos007:
## Description of changes

When cross compiling proc macros, the proc macro needs to be built for the build platform's architecture. 

Without this change cross compiling from Darwin to Linux would simply fail because it tries to link to a library with a file extension that doesn't exist on the builder's platform.

This is something we've had to patch locally in our private repo. We use crate2nix and have a special zig based cross compilation setup we wish to upstream :)

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [x] x86_64-darwin
  - [ ] aarch64-darwin
- For non-Linux: Is sandboxing enabled in `nix.conf`? (See [Nix manual](https://nixos.org/manual/nix/stable/command-ref/conf-file.html))
  - [x] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:
  - [NixOS test(s)](https://nixos.org/manual/nixos/unstable/index.html#sec-nixos-tests) (look inside [nixos/tests](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests))
  - and/or [package tests](https://github.com/NixOS/nixpkgs/blob/master/pkgs/README.md#package-tests)
  - or, for functions and "core" functionality, tests in [lib/tests](https://github.com/NixOS/nixpkgs/blob/master/lib/tests) or [pkgs/test](https://github.com/NixOS/nixpkgs/blob/master/pkgs/test)
  - made sure NixOS tests are [linked](https://nixos.org/manual/nixpkgs/unstable/#ssec-nixos-tests-linking) to the relevant packages
- [ ] Tested compilation of all packages that depend on this change using `nix-shell -p nixpkgs-review --run "nixpkgs-review rev HEAD"`. Note: all changes have to be committed, also see [nixpkgs-review usage](https://github.com/Mic92/nixpkgs-review#usage)
- [ ] Tested basic functionality of all binary files (usually in `./result/bin/`)
- [24.11 Release Notes](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2411.section.md) (or backporting [23.11](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2311.section.md) and [24.05](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2405.section.md) Release notes)
[...]

PR #217428, opened 2023-02-21
Add pkgsCross.emscripten, replacing hand-crafted emscriptenPackages

AtnNn:
###### Description of changes

- Added a pre-built Emscripten `SYSTEM` cache to the `emscripten` package
- Added `emscriptenFull` which additionally contains Emscripten `PORTS`
- Added `emscriptenNoCache` which requires setting `EM_CACHE` to use
- Add wrapped `emscriptenCC` and `emscriptenBintools` 
- Added Emscripten to `lib/systems` and `pkgsCross`
- Moved the `xmlmirror` package from inside `top-level/emscripten-packages.nix` to `tools/text/xml`
- Removed custom `emscriptenPackages.json_c` and `emscriptenPackages.zlib`
- Removed `buildEmscriptenPackage`
- Added some `pkgsCross.emscripten` packages to `release-cross.nix`

Possibly fixes #28929 #141246 #136396. Improves the fix for #139943.

Tested with

```
$ nix-build --show-trace -E 'with import ./pkgs/top-level/release-cross.nix {}; builtins.mapAttrs (k: v: v."x86_64-linux") emscripten'
$ node result-3/bin/hello
Hello, world!
```

###### Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
- [ ] For non-Linux: Is `sandbox = true` set in `nix.conf`? (See [Nix manual](https://nixos.org/manual/nix/stable/command-ref/conf-file.html))
- [ ] Tested, as applicable:
  - [NixOS test(s)](https://nixos.org/manual/nixos/unstable/index.html#sec-nixos-tests) (look inside [nixos/tests](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests))
  - and/or [package tests](https://nixos.org/manual/nixpkgs/unstable/#sec-package-tests)
  - or, for functions and "core" functionality, tests in [lib/tests](https://github.com/NixOS/nixpkgs/blob/master/lib/tests) or [pkgs/test](https://github.com/NixOS/nixpkgs/blob/master/pkgs/test)
  - made sure NixOS tests are [linked](https://nixos.org/manual/nixpkgs/unstable/#ssec-nixos-tests-linking) to the relevant packages
[...]

PR #319153, opened 2024-06-11
buildRustCrate: add support for compiling to wasm32-unknown-unknown

Kranzes:
## Description of changes

- **lib.systems.examples: add wasm32-unknown-none**
This system was added to use the nixpkgs cross compilation logic when
compiling to wasm32-unknown-unknown in rust.

- **buildRustCrate: support cross compiling to wasm32-unknown-unknown**
With this change you will finally be able to use
buildRustCrate/crate2nix to build your wasm32-unknown-unknown
rust projects.

Simply import nixpkgs like so:
```
lib = <nixpkgs/lib>;
pkgsForWasm32 = import <nixpkgs> {
  crossSystem = lib.examples.wasm32-unknown-none;
}
```
or use pkgsCross directly with
```
pkgsCross.wasm32-unknown-none.callPackage ./. { };
```

- **tests.buildRustCrate: add crateLibOutputsWasm32**
Added a cross compilation test for wasm32-unknown-unknown. This also
required using crate.metadata instead of using the regex to get rid of
the hash in the library filename. It also required adding a mkCrate
argument to assertOutputs so we can override the buildRustCrate used.

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/lib/systems/examples.nix
+++ b/lib/systems/examples.nix
@@ -356,6 +356,12 @@ rec {
     useLLVM = true;
   };
 
+  wasm32-unknown-none = {
+    config = "wasm32-unknown-none";
+    rust.rustcTarget = "wasm32-unknown-unknown";
+    useLLVM = true;
+  };
+
   # Ghcjs
   ghcjs = {
     # This triple is special to GHC/Cabal/GHCJS and not recognized by autotools
--- a/pkgs/build-support/rust/build-rust-crate/default.nix
+++ b/pkgs/build-support/rust/build-rust-crate/default.nix
@@ -16,6 +16,16 @@
 }:
 
 let
+  # Returns a true if the builder's rustc was built with support for the target.
+  targetAlreadyIncluded = lib.elem stdenv.hostPlatform.rust.rustcTarget
+    (lib.splitString "," (lib.removePrefix "--target=" (
+      lib.elemAt (lib.filter (f: lib.hasPrefix "--target=" f) pkgsBuildBuild.rustc.unwrapped.configureFlags) 0)
+    ));
+
+  # If the build's rustc was built with support for the target then reuse it. (Avoids uneeded compilation for targets like `wasm32-unknown-unknown`)
+  rustc' = if targetAlreadyIncluded then pkgsBuildBuild.rustc else rustc;
+  cargo' = if targetAlreadyIncluded then pkgsBuildBuild.cargo else cargo;
+
   # Create rustc arguments to link against the given list of dependencies
   # and renames.
   #
@@ -75,6 +85,7 @@ let
 
   buildCrate = import ./build-crate.nix {
     inherit lib stdenv mkRustcDepArgs mkRustcFeatureArgs needUnstableCLI;
+    rustc = rustc';
   };
 
   installCrate = import ./install-crate.nix { inherit stdenv; };
@@ -272,7 +283,8 @@ crate_: lib.makeOverridable
       name = "rust_${crate.crateName}-${crate.version}${lib.optionalString buildTests_ "-test"}";
       version = crate.version;
       depsBuildBuild = [ pkgsBuildBuild.stdenv.cc ];
-      nativeBuildInputs = [ rust stdenv.cc cargo jq ]
+      nativeBuildInputs = [ rustc' cargo' jq ]
+        ++ lib.optionals stdenv.hasCC [ stdenv.cc ]
         ++ lib.optionals stdenv.buildPlatform.isDarwin [ libiconv ]
         ++ (crate.nativeBuildInputs or [ ]) ++ nativeBuildInputs_;
       buildInputs = lib.optionals stdenv.isDarwin [ libiconv ] ++ (crate.buildInputs or [ ]) ++ buildInputs_;
@@ -378,7 +390,7 @@ crate_: lib.makeOverridable
     )
   )
 {
-  rust = rustc;
+  rust = rustc';
   release = crate_.release or true;
   verbose = crate_.verbose or true;
   extraRustcOpts = [ ];
--- a/lib/systems/default.nix
+++ b/lib/systems/default.nix
@@ -84,20 +84,21 @@ let
       useLLVM = final.isFreeBSD;
 
       libc =
-        /**/ if final.isDarwin              then "libSystem"
-        else if final.isMinGW               then "msvcrt"
-        else if final.isWasi                then "wasilibc"
-        else if final.isRedox               then "relibc"
-        else if final.isMusl                then "musl"
-        else if final.isUClibc              then "uclibc"
-        else if final.isAndroid             then "bionic"
-        else if final.isLinux /* default */ then "glibc"
-        else if final.isFreeBSD             then "fblibc"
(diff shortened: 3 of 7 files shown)


alyssais:
> * Without this change evaluation may fail when cross compiling to a system
>       that does not have:
>       `stdenv.hostPlatform.extensions.staticLibrary`

What systems are these?  It looks to me from lib/systems/default.nix that every system should have that set.

thehabbos007:
Just chiming in to say I've also been looking at cross compilation improvements. It's nice to see there is more people working on this/needing this. We have applied the following changes locally for cross compilation with crate2nix

https://github.com/NixOS/nixpkgs/pull/317755
https://github.com/NixOS/nixpkgs/pull/319782

pkgs/build-support/rust/build-rust-crate/default.nix
```diff
@@ -16,6 +17,16 @@
 }:
 
 let
+  # Returns a true if the builder's rustc was built with support for the target.
+  targetAlreadyIncluded = lib.elem stdenv.hostPlatform.rust.rustcTarget
+    (lib.splitString "," (lib.removePrefix "--target=" (
```
Kranzes, line 22: Is there a better way to write this?

alyssais:
> > What systems are these? It looks to me from lib/systems/default.nix that every system should have that set.
> 
> https://github.com/nix-community/crate2nix/blob/cf034861fdc4e091fc7c5f01d6c022dc46686cf1/sample_projects/empty_cross/default.nix#L11

```
nix-repl> (import <nixpkgs> { crossSystem.config = "riscv32-unknown-none-elf"; }).stdenv.hostPlatform.extensions.staticLibrary
".a"
```

Kranzes:
it seems it does have staticLibrary, but it doesn't match isStatic. I'll look into this,

```
nix-repl> (import <nixpkgs> { crossSystem.config = "riscv32-unknown-none-elf"; }).stdenv.hostPlatform.extensions.staticLibrary
".a"
nix-repl> (import <nixpkgs> { crossSystem.config = "riscv32-unknown-none-elf"; }).stdenv.hostPlatform.extensions.library
error:
       … while evaluating the attribute 'stdenv.hostPlatform.extensions.library'
         at /nix/store/dydg48djlykksz8cxq0xjplyxpa9pvf4-source/lib/systems/default.nix:132:9:
          131|           else                         ".a";
          132|         library =
             |         ^
          133|           /**/ if final.isStatic then final.extensions.staticLibrary

       … while evaluating the attribute 'extensions.sharedLibrary'
         at /nix/store/dydg48djlykksz8cxq0xjplyxpa9pvf4-source/lib/systems/default.nix:123:7:
          122|       else null;
          123|       extensions = optionalAttrs final.hasSharedLibraries {
             |       ^
          124|         sharedLibrary =

       error: attribute 'sharedLibrary' missing
       at /nix/store/dydg48djlykksz8cxq0xjplyxpa9pvf4-source/lib/systems/default.nix:134:39:
          133|           /**/ if final.isStatic then final.extensions.staticLibrary
          134|           else                        final.extensions.sharedLibrary;
             |                                       ^
          135|         executable =
```

Kranzes:
I am not too familiar with embedded architectures and their linking so I think I will drop that commit and let someone else handle it, it isn't related to the wasm32 thing anyway. With this PR the support for wasm32 doesn't make it possible to statically link to C libraries so for now it is dynamic (as It uses stdenvNoCC and the build's rustc). In the future I will look into supporting normal stdenv too.

lib/systems/examples.nix
```diff
@@ -356,6 +356,12 @@ rec {
     useLLVM = true;
   };
 
+  wasm32-unknown-none = {
+    config = "wasm32-unknown-none-unknown";
+    rust.rustcTarget = "wasm32-unknown-unknown";
+    useLLVM = true;
+  };
```
alyssais, line 363: Is this useful for anything besides Rust?

Kranzes: Maybe? Idk of enough languages that can compile to WASI-less wasm.

alyssais: Where did the triple come from then?

Anyway, if it's Rust-only, I'm wondering whether pkgsCross is the right mechanism.  We can already handle (buildRustPackage) packages containing wasm32-unknown-unknown builds (and we can't use pkgsCross for those, because they contain a mix of wasm and native code).  There's no standard library to build, nor (if it's only Rust) are there likely to be packaged libraries to link against.  So why not just let buildRustCrate take a target parameter (if it doesn't already)?  That would also be useful for other custom targets, as well as the -none ones.

@Ericson2314 wdyt?

Kranzes: I think the way buildRustCrate and crate2nix work is slightly more involved than that, it's not just about passing `--target` to rustc, there's a whole logic of matching dependencies depending on the target's family, arch etc. Crate2nix directly uses the nixpkgs cross compilation logic. Also what you are commenting on is just about having an example of it. I made the logic pretty generic so it will apply to other wasm targets that rust support (haven't tested tho!).

flokli: I can certainly see a future where we might want to statically link together some rust crates with some static C code (maybe [via Emscripten](https://developer.mozilla.org/en-US/docs/WebAssembly/existing_C_to_Wasm), and using the nixpkgs cross ecosystem should be the right vehicle for it, no?

alyssais: Well that's the thing — Emscripten is likely to be a different cross target (https://github.com/NixOS/nixpkgs/pull/217428), so that wouldn't really be at all aligned with how pkgsCross expects the world to work.  It's not really designed for building for more than one target at the same time, so where we can do that in another way (e.g. by making one Rust toolchain able to build both native and Rust code, for packages like polkadot or lldap) that lets us avoid those limitations of pkgsCross.

Kranzes: I don't think Emscripten is particularly related tbh. As I previously mentioned a better example is ghcjs, this target is more like ghcjs, it targets the web, it doesn't have a libc, it's compiler specific (rustc) (maybe TinyGo too?), except ghcjs is much more obtrusive as there a couple of explicit conditionals referencing it in the stdenv logic. The changes in this PR are meant to be somewhat generic and only touch buildRustCrate for now.

Kranzes:
@alyssais I made some changes to the lib.systems.examples parsing logic so it accepts the system as a triple rather than a quadruple, this was already the case for WASI, just had to modify a conditional to also work for WASI-less WASM. I also added a test for buildRustCrate, let me know if there's anything else that's blocking this PR from being merged.

flokli:
Let's get this in, it's useful. If we ever decide to do things differently that PR is not gonna make it harder.

bouk:
Running into a problem setting the rust toolchain since this version: https://github.com/NixOS/nixpkgs/issues/326112

PR #307173, opened 2024-04-27
maintainers: remove delroth

delroth:
## Description of changes

It's been fun, until the lack of proper, decisive moderation finally became a problem. Maybe if the right people had been banned a month ago I'd be sending a different PR today...

---

Add a :+1: [reaction] to [pull requests you find important].

[reaction]: https://github.blog/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/
[pull requests you find important]: https://github.com/NixOS/nixpkgs/pulls?q=is%3Aopen+sort%3Areactions-%2B1-desc

PR #319160, opened 2024-06-11
pahole: 1.26 -> 1.27

martinetd:
## Description of changes

(Note I don't have enough cpu/disk to build a linux kernel with this, so letting ofborg do the work... Always a bit of a bet on staging. Maybe someday...)

From the release e-mail:

```
        The v1.27 release of pahole and its friends is out, supporting
parallel reproducible builds and encoding kernel kfuncs in BTF, allowing
tools such as bpftrace to enumerate the available kfuncs and obtain its
function signatures and return types.

BTF encoder:

- Inject kfunc decl tags into BTF from the BTF IDs ELF section in the Linux
  kernel vmlinux file.
  
  This allows tools such as bpftools and pfunct to enumerate the available kfuncs
  and to gets its function signature, the type of its return and of its
  arguments. See the example in the BTF loader changes description, below.

- Support parallel reproducible builds, where it doesn't matter how many
  threads are used, the end BTF encoding result is the same.

- Sanitize unsupported DWARF int type with greater-than-16 byte, as BTF doesn't
  support it.

BTF loader:

- Initial support for BTF_KIND_DECL_TAG:

  $ pfunct --prototypes -F btf vmlinux.btf.decl_tag,decl_tag_kfuncs | grep ^bpf_kfunc | head
  bpf_kfunc void cubictcp_init(struct sock * sk);
  bpf_kfunc void cubictcp_cwnd_event(struct sock * sk, enum tcp_ca_event event);
  bpf_kfunc void cubictcp_cong_avoid(struct sock * sk, u32 ack, u32 acked);
  bpf_kfunc u32 cubictcp_recalc_ssthresh(struct sock * sk);
  bpf_kfunc void cubictcp_state(struct sock * sk, u8 new_state);
  bpf_kfunc void cubictcp_acked(struct sock * sk, const struct ack_sample  * sample);
  bpf_kfunc int bpf_iter_css_new(struct bpf_iter_css * it, struct cgroup_subsys_state * start, unsigned
int flags);
  bpf_kfunc struct cgroup_subsys_state * bpf_iter_css_next(struct bpf_iter_css * it);
  bpf_kfunc void bpf_iter_css_destroy(struct bpf_iter_css * it);
  bpf_kfunc s64 bpf_map_sum_elem_count(const struct bpf_map  * map);
  $ pfunct --prototypes -F btf vmlinux.btf.decl_tag,decl_tag_kfuncs | grep ^bpf_kfunc | wc -l
  116
  $

pretty printing:

- Fix hole discovery with inheritance in C++.
```

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/by-name/pa/pahole/package.nix
+++ b/pkgs/by-name/pa/pahole/package.nix
@@ -26,6 +26,7 @@ stdenv.mkDerivation rec {
     musl-obstack
   ];
 
+  # https://github.com/acmel/dwarves/pull/51
   patches = [ ./threading-reproducibility.patch ];
 
   # Put libraries in "lib" subdirectory, not top level of $out
--- a/pkgs/by-name/pa/pahole/package.nix
+++ b/pkgs/by-name/pa/pahole/package.nix
@@ -13,10 +13,10 @@
 
 stdenv.mkDerivation rec {
   pname = "pahole";
-  version = "1.26";
+  version = "1.27";
   src = fetchzip {
     url = "https://git.kernel.org/pub/scm/devel/pahole/pahole.git/snapshot/pahole-${version}.tar.gz";
-    hash = "sha256-Lf9Z4vHRFplMrUf4VhJ7EDPn+S4RaS1Emm0wyEcG2HU=";
+    hash = "sha256-BwA17lc2yegmOzLfoIu8OmG/PVdc+4sOGzB8Jc4ZjGM=";
   };
 
   nativeBuildInputs = [ cmake pkg-config ];
--- a/pkgs/by-name/pa/pahole/package.nix
+++ b/pkgs/by-name/pa/pahole/package.nix
@@ -9,6 +9,7 @@
 , argp-standalone
 , musl-obstack
 , nixosTests
+, fetchpatch
 }:
 
 stdenv.mkDerivation rec {
@@ -26,8 +27,16 @@ stdenv.mkDerivation rec {
     musl-obstack
   ];
 
-  # https://github.com/acmel/dwarves/pull/51
-  patches = [ ./threading-reproducibility.patch ];
+  patches = [
+    # https://github.com/acmel/dwarves/pull/51 / https://lkml.kernel.org/r/20240626032253.3406460-1-asmadeus@codewreck.org
+    ./threading-reproducibility.patch
+    # https://github.com/acmel/dwarves/issues/53
+    (fetchpatch {
+      name = "fix-clang-btf-generation-bug.patch";
+      url = "https://github.com/acmel/dwarves/commit/6a2b27c0f512619b0e7a769a18a0fb05bb3789a5.patch";
+      hash = "sha256-Le1BAew/a/QKkYNLgSQxEvZ9mEEglUw8URwz1kiheeE=";
+    })
+  ];
 
   # Put libraries in "lib" subdirectory, not top level of $out
   cmakeFlags = [ "-D__LIB=lib" "-DLIBBPF_EMBEDDED=OFF" ];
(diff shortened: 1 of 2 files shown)


martinetd:
ofborg tests passed, yay.
I'm surprise it didn't add Ccs, so here: @bosu

BTW @delroth -- they added `--btf_features=+reproducible_build` to build the BTF output in a consistent manner even with multiple threads; I'll send a patch to enable it by default if SOURCE_DATE_EPOCH is set at which point we'll be able to drop the threading-reproducibility.patch

martinetd:
> BTW @delroth -- they added --btf_features=+reproducible_build to build the BTF output in a consistent manner even with multiple threads; I'll send a patch to enable it by default if SOURCE_DATE_EPOCH is set at which point we'll be able to drop the threading-reproducibility.patch

That wasn't so bad so updated patch. The output is identical to -j1 output from what I could see.

martinetd:
Note this apparently introduces a regression in linux builds with clang with LTO enabled:
- https://lkml.kernel.org/r/20240613214019.GA1423015@thelio-3990X
- https://github.com/acmel/dwarves/issues/53

I don't think we actively do that, but on the other hand there's no strict hurry to merge this either, so let's put this on hold until that is fixed

martinetd:
> Note this apparently introduces a regression in linux builds with clang with LTO enabled:

I was expecting a new release with it but it's been fixed without such ceremony, so I picked the patch up.

Also didn't get any response to the reproducibility patch so resent it by e-mail, let's see how that one goes.. This PR was only waiting for the clang fix, so marked as ready again.

vcunat:
I believe there's no use in pinging Delroth anymore: https://github.com/NixOS/nixpkgs/pull/307173

martinetd:
Doh, thanks for the heads up.
And thanks for merging as well.

FWIW that patch was more or less refused upstream after a few exchanges with upstream, between pahole folks telling me that this belongs in kbuild, and kbuild folks saying this deserves a new .config option... let's just keep that harmless patch around for now.

PedroHLC:
> I don't think we actively do that, but on the other hand there's no strict hurry to merge this either, so let's put this on hold until that is fixed

Maybe I should upstream a kernel built with Clang and LTO (would help to assert our kernel modules build against this scenario), [mine](https://www.nyx.chaotic.cx/#t-t1-packages) broke after this bump :cry:

PedroHLC:
> I don't think we actively do that, but on the other hand there's no strict hurry to merge this either, so let's put this on hold until that is fixed

Maybe I should upstream a kernel built with Clang and LTO (would help to assert our kernel modules build against this scenario), [mine](https://www.nyx.chaotic.cx/#t-t1-packages) broke after this bump :cry: 

**EDIT** Solution is a second patch by Arnaldo: https://lore.kernel.org/all/ZnGZ71a4E29kPrvS@x1/

martinetd:
@PedroHLC doh, sorry... I wish he'd just tagged a fix release instead of posting patches like this :/

especially since on https://github.com/acmel/dwarves/issues/53 someone commented the other patch was enough..

I've just opened https://github.com/NixOS/nixpkgs/pull/331589 - if you could confirm it works manually that'd be great.
Long term I agree a test would be welcome, the bpf nixos test builds the regular kernel as a dependency but we could use a minimal clang kernel build test. Heck, I can't even build a kernel on my machines because we have way too many modules enabled, I need to figure out how to build a minimal virt kernel so I can run these tests properly... but at least ofborg would build it for now!

PR #319220, opened 2024-06-12
Fix failures with pkgs/top-level/release-attrpaths-superset.nix

### LINKED ISSUE #319147: nixpkgs-lib test `pkgs/top-level/release-attrpaths-superset.nix` infinite recurses on eval-system=aarch64-darwin
### Describe the bug

We found (because it broke our build in Lix) that the following nixpkgs lib test is broken. It is run through some derivation in the test suite somewhere; finding that is an exercise to the reader, but it is included underneath `lib/tests/release.nix` somewhere.

Run the following on any platform; it has been observed with this command on aarch64-darwin and x86_64-linux:

```
nix-instantiate --eval --system aarch64-darwin --strict --json pkgs/top-level/release-attrpaths-superset.nix -A names
```

Then you will get an infinite output deepening like the following:

```
trace: warning: tryEval failed at: freshBootstrapTools.test-pkgs.__splicedPackages.__splicedPackages.__splicedPackages.__splicedPackages.AAAAAASomeThingsFailToEvaluate
trace: warning: tryEval failed at: freshBootstrapTools.test-pkgs.__splicedPackages.__splicedPackages.__splicedPackages.__splicedPackages.__splicedPackages.AAAAAASomeThingsFailToEvaluate
```

Eventually it will segfault or otherwise exceed the recursion limit.

We checked that this reproduces on Lix `main`, Nix 2.18, and Nix 2.22, so it is not a Nix implementation bug for any reasonable value of "nix implementation".

### Steps To Reproduce
1. Run the above command from a checkout of nixpkgs.
2. ???
3. Stack overflow!

We don't think this is a regression from https://github.com/NixOS/nixpkgs/pull/317772, but instead, that broken stuff was exposed as broken by actually running it on other platforms than x86_64-linux.

### Expected behavior

Evaluate as normal as is the case for other values of `system`.

### Notify maintainers

Not really a maintainers list as I don't see one, but people who touched the file or the infra recently:

@infinisil 
@RossComputerGuy 
@Gabriella439 
@Ericson2314 

### Metadata

Note that this is *not* a Lix bug.

 - system: `"x86_64-linux"`
 - host os: `Linux 6.6.23-1-lts, Arch Linux, noversion, rolling`
 - multi-user?: `yes`
 - sandbox: `yes`
 - version: `nix-env (Lix, like Nix) 2.90.0-lixpre20240611-8a3d063`
 - channels(jade): `""`
 - nixpkgs: `/nix/var/nix/profiles/per-user/root/channels/nixpkgs`

---

RossComputerGuy:
## Description of changes

Fixes #319147

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [x] aarch64-linux
  - [x] x86_64-darwin
  - [x] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/test/stdenv/default.nix
+++ b/pkgs/test/stdenv/default.nix
@@ -99,6 +99,9 @@ let
 in
 
 {
+  # Disable on Darwin due to assumptions with __bootPackages
+  __attrsFailEvaluation = stdenv.isDarwin;
+
   # tests for hooks in `stdenv.defaultNativeBuildInputs`
   hooks = lib.recurseIntoAttrs (import ./hooks.nix { stdenv = bootStdenv; pkgs = earlyPkgs; inherit lib; });
 
--- a/pkgs/build-support/kernel/make-initrd.nix
+++ b/pkgs/build-support/kernel/make-initrd.nix
@@ -56,7 +56,7 @@ in
 , prepend ? []
 
 # Whether to wrap the initramfs in a u-boot image.
-, makeUInitrd ? stdenvNoCC.hostPlatform.linux-kernel.target == "uImage"
+, makeUInitrd ? stdenvNoCC.hostPlatform.linux-kernel.target or "dummy" == "uImage"
 
 # If generating a u-boot image, the architecture to use. The default
 # guess may not align with u-boot's nomenclature correctly, so it can
@@ -75,10 +75,8 @@ let
   toValidStoreName = x: with builtins;
     lib.concatStringsSep "-" (filter (x: !(isList x)) (split "[^a-zA-Z0-9_=.?-]+" x));
 
-in stdenvNoCC.mkDerivation rec {
+in stdenvNoCC.mkDerivation (rec {
   inherit name makeUInitrd extension uInitrdArch prepend;
-
-  ${if makeUInitrd then "uInitrdCompression" else null} = uInitrdCompression;
 
   builder = ./make-initrd.sh;
 
@@ -110,4 +108,6 @@ in stdenvNoCC.mkDerivation rec {
       contents
       (lib.range 0 (lib.length contents - 1));
   pathsFromGraph = ./paths-from-graph.pl;
-}
+} // lib.optionalAttrs makeUInitrd {
+  uInitrdCompression = uInitrdCompression;
+})
--- a/pkgs/top-level/release-attrpaths-superset.nix
+++ b/pkgs/top-level/release-attrpaths-superset.nix
@@ -52,6 +52,7 @@ let
     pkgsMusl = true;
     pkgsStatic = true;
     pkgsCross = true;
+    pkgsx86_64Darwin = true;
     pkgsi686Linux = true;
     pkgsLinux = true;
     pkgsExtraHardening = true;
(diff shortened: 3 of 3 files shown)


tomberek:
Adding optionalAttrs and protection from non-existent attribute.

tomberek (approved):
Looks better. Requesting @lheckemann as well, if available.

philiptaron (approved):

pkgs/test/stdenv/default.nix
```diff
@@ -99,6 +99,8 @@ let
 in
 
 {
+  __attrsFailEvaluation = stdenv.isDarwin;
```
lf-, line 102: I'm not the most pleased this entire suite is turned off on Darwin. It's ok if necessary but it's a little unfortunate.

lf- (commented):
mostly lgtm but I'm not super stoked we're turning off an entire suite of which a lot looks applicable

pkgs/test/stdenv/default.nix
RossComputerGuy: Yeah, it had a funky error without it.

RossComputerGuy: This shouldn't be an issue, idk why it's a problem. I kinda didn't feel like diving into figuring out why this was happening so the easiest solution was to just disable on Darwin. Linux targets work so this is a Darwin only issue. If this isn't a blocker and we can't find a solution before this PR is merged, a follow up PR to properly fix it should be fine.
```
error:
       … while calling the 'map' builtin

         at /home/ross/ExpidusOS/nixpkgs/pkgs/top-level/release-attrpaths-superset.nix:193:5:

          192|   names =
          193|     map (path: (lib.concatStringsSep "." path)) paths;
             |     ^
          194|

         at /home/ross/ExpidusOS/nixpkgs/pkgs/top-level/release-attrpaths-superset.nix:162:9:

          161|       result =
          162|         if tried.success
             |         ^
          163|         then tried.value

       error: attribute '__bootPackages' missing

       at /home/ross/ExpidusOS/nixpkgs/pkgs/test/stdenv/default.nix:13:17:

           12|   earlyPkgs = stdenv.__bootPackages.stdenv.__bootPackages;
           13|   earlierPkgs = stdenv.__bootPackages.stdenv.__bootPackages.stdenv.__bootPackages.stdenv.__bootPackages.stdenv.__bootPackages;
             |                 ^
           14|   # use a early stdenv so when hacking on stdenv this test can be run quickly
```

lf-: there's got to be a way to fix this better because `__bootPackages` is certainly not the only attr that is broken in the entirety of nixpkgs, right?? but can you add a FIXME above that disabling, if you want to defer it?

lf-: also cc @reckenrode as a darwin maintainer noting that this was deferred.

RossComputerGuy: > but can you add a FIXME above that disabling, if you do want to defer it; with that change this is lgtm without figuring this out.

Will do.

RossComputerGuy: Added the note.

lf- (approved):

PR #319325, opened 2024-06-12
neovim: take into account plugin dependencies

### LINKED ISSUE #318925: lua dependencies not seen for vim plugins built with buildNeovimPlugin
### Describe the bug

Neovim does't see the dependencies of some plugins, for instance luasnip and jsregexp and reported here https://github.com/NixOS/nixpkgs/pull/311386 and  https://github.com/NixOS/nixpkgs/issues/306367 (there are other occurrences that we can link to this ticket).

### Steps To Reproduce
add luasnip to neovim plugins and notice that checkhealth warns about missing jsregexp.

### Expected behavior
No warning

### Additional context
I solved the problem locally some time ago by creating a lua env with the closure of all vim plugins but it's not upstreamable IMO, not efficient nor clean.
I started writing a test to see how it worked on master and saw jsregexp was loaded into the packpath as a lua plugin. That's because of https://github.com/NixOS/nixpkgs/pull/311386 and is a red herring.
My current plan is:
1. to customize  vimUtils.packDir to aggregate the different propagated-build-inputs from the various plugins
2. and call `addToLuaPath` (a variant I have that loads propagatedBuildInputs as well) on the packpath dir.
3. wrap neovim with LUA_PATH and LUA_CPATH generated by this call.

NB: https://github.com/NixOS/nixpkgs/pull/311386 made me realize that packDir will pull transitive deps in the packdir regardless of if they are vim plugins. I think we should only include vim plugins in the closure to avoid mistakes but I will let that for the future as it could break stuff.

### Notify maintainers
cc @teto

### Metadata

```console
output here
```

---

teto:
- **neovim.tests: test lua transitive deps are available**
- **neovimUtils.packDir: init: extend vimUtils.packDir**
- **neovim: wrap LUA_PATH and LUA_CPATH**
- **lua: take into propagated-build-inputs when building LUA_PATH**

## Description of changes

Attempt to fix https://github.com/NixOS/nixpkgs/issues/318925, e.g., luasnip
depends on jsregexp but it's currently not possible to load jsregexp when
luasnip is added to neovim.

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/applications/editors/neovim/utils.nix
+++ b/pkgs/applications/editors/neovim/utils.nix
@@ -229,12 +229,36 @@ let
         ln -s ${grammar}/parser $out/parser/${name}.so
       '');
 
+  /*
+    Fork of vimUtils.packDir that additionnally generates a propagated-build-inputs-file that
+    can be used by the lua hooks to generate a proper LUA_PATH
+
+    Generates a packpath folder as expected by vim
+       Example:
+       packDir ( {myVimPackage = { start = [ vimPlugins.vim-fugitive ]; opt = []; }; })
+       => "/nix/store/xxxxx-pack-dir"
+  */
+  packDir = packages:
+  let
+    rawPackDir = vimUtils.packDir packages;
+
+  in
+    rawPackDir.override ({
+    postBuild = ''
+      mkdir $out/nix-support
+      for i in $(find -L $out -name propagated-build-inputs ); do
+        cat "$i" >> $out/nix-support/propagated-build-inputs
+      done
+      '';});
+
+
 in
 {
   inherit makeNeovimConfig;
   inherit generateProviderRc;
   inherit legacyWrapper;
   inherit grammarToPlugin;
+  inherit packDir;
 
   inherit buildNeovimPlugin;
   buildNeovimPluginFrom2Nix = lib.warn "buildNeovimPluginFrom2Nix was renamed to buildNeovimPlugin" buildNeovimPlugin;
--- a/pkgs/applications/editors/neovim/wrapper.nix
+++ b/pkgs/applications/editors/neovim/wrapper.nix
@@ -24,6 +24,8 @@ let
     , withNodeJs ? false
     , withPerl ? false
     , rubyEnv ? null
+
+    # wether to create symlinks in $out/bin/vi(m) -> $out/bin/nvim
     , vimAlias ? false
     , viAlias ? false
 
@@ -45,6 +47,8 @@ let
   stdenv.mkDerivation (finalAttrs:
   let
 
+    finalPackdir = neovimUtils.packDir packpathDirs;
+
     rcContent = ''
       ${luaRcContent}
     '' + lib.optionalString (!isNull neovimRcContent) ''
@@ -58,9 +62,10 @@ let
           [
             "--add-flags" ''--cmd "lua ${providerLuaRc}"''
             # (lib.intersperse "|" hostProviderViml)
-          ] ++ lib.optionals (packpathDirs.myNeovimPackages.start != [] || packpathDirs.myNeovimPackages.opt != []) [
-            "--add-flags" ''--cmd "set packpath^=${vimUtils.packDir packpathDirs}"''
-            "--add-flags" ''--cmd "set rtp^=${vimUtils.packDir packpathDirs}"''
+          ]
+          ++ lib.optionals (packpathDirs.myNeovimPackages.start != [] || packpathDirs.myNeovimPackages.opt != []) [
+            "--add-flags" ''--cmd "set packpath^=${finalPackdir}"''
+            "--add-flags" ''--cmd "set rtp^=${finalPackdir}"''
           ]
           ;
 
--- a/pkgs/applications/editors/neovim/wrapper.nix
+++ b/pkgs/applications/editors/neovim/wrapper.nix
@@ -13,6 +13,9 @@
 neovim-unwrapped:
 
 let
+  # inherit interpreter from neovim
+  lua = neovim-unwrapped.lua;
+
   wrapper = {
       extraName ? ""
     # should contain all args but the binary. Can be either a string or list
@@ -61,7 +64,6 @@ let
       # vim accepts a limited number of commands so we join them all
           [
             "--add-flags" ''--cmd "lua ${providerLuaRc}"''
-            # (lib.intersperse "|" hostProviderViml)
           ]
           ++ lib.optionals (packpathDirs.myNeovimPackages.start != [] || packpathDirs.myNeovimPackages.opt != []) [
             "--add-flags" ''--cmd "set packpath^=${finalPackdir}"''
@@ -165,7 +167,17 @@ let
       + ''
         rm $out/bin/nvim
         touch $out/rplugin.vim
-        makeWrapper ${lib.escapeShellArgs finalMakeWrapperArgs} ${wrapperArgsStr}
+
+        echo "Looking for lua dependencies..."
+        source ${lua}/nix-support/utils.sh
+
+        _addToLuaPath "${finalPackdir}"
+
+        echo "LUA_PATH towards the end of packdir: $LUA_PATH"
+
+        makeWrapper ${lib.escapeShellArgs finalMakeWrapperArgs} ${wrapperArgsStr} \
+            --prefix LUA_PATH ';' "$LUA_PATH" \
+            --prefix LUA_CPATH ';' "$LUA_CPATH"
       '';
 
     buildPhase = ''
--- a/pkgs/development/interpreters/lua-5/utils.sh
+++ b/pkgs/development/interpreters/lua-5/utils.sh
@@ -1,4 +1,8 @@
-#!/bin/sh
+#!/bin/bash
+
+declare -gA luaPathsSeen=()
+
+# shellcheck disable=SC2164,SC2041
 nix_print() {
   if [ ${NIX_DEBUG:-0} -ge $1 ]; then
     echo "$2"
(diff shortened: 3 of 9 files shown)


BirdeeHub:
Isn't this what the extraLuaPackages option is for? It seems redundant because makeNeovimConfig function already sets LUA_PATH and LUA_CPATH based on that option.

BirdeeHub:
So, I made an innacurate comment earlier but, it is actually possible to add jsregexp manually to neovim via the extraLuaPackages option of the makeNeovimConfig function? However, from what I gather, this is for doing it automatically when plugins declare propagated build inputs? If so thats kinda cool

teto:
> However, from what I gather, this is for doing it automatically when plugins declare propagated build inputs so that things like jsregexp will be added automatically? If so thats cool and good for user experience.

That's correct.

> This issue aside, maybe this option should be exposed in the module options for programs.neovim? 

Rather the opposite. I would have to check the history but I am pretty sure `extraLuaPackages` was introduced as a workaround for the issue this PR fixes. I plan changes to the neovim wrapper for 24.11 and I will ask you and nixvim authors for feedback but let's leave that for another issue.

@GaetanLepage @mrcjkb Could you test please (cherry-pick on top of unstable) ? I reverted much of my local changes and looks like it's working here aka the neovim wrapper adds entries in LUA_PATH / LUA_CPATH with the wrapper.

I tested with (there is a test for it in `nix-build -A neovim.tests` as well. One  of the test is broken but it's already the case on master). I used:
```
let
  pluginsMap = pkgs.callPackage ../../hm/profiles/neovim/plugins.nix {};
  
  # a list of [ { plugin = DRV; config (optional) =  string } ] 
# for instance [ { plugin = vimPlugins.luasnip' } ]
  allPlugins = 
       pluginsMap.basePlugins
    ++ pluginsMap.luaPlugins
    ++ pluginsMap.colorschemePlugins
    ++ pluginsMap.filetypePlugins
  ;

  myNvimConfig = pkgs.neovimUtils.makeNeovimConfig {
    plugins = allPlugins;
  };
  myNvim = pkgs.wrapNeovimUnstable pkgs.neovim-unwrapped myNvimConfig;

in {
  environment.systemPackages = [
    (builtins.trace "${myNvim}" myNvim)
  ];
```
After rebuilding, `less PRINTED_NVIM_PATH/bin/nvim` shows jsregexp added to LUA_PATH.
Loading the executable worked too: `/nix/store/m7qjpvyax1wb4zj6ccn4a051f5d4z0w3-neovim-0.10.0/bin/nvim --clean +'lua require"jsregexp"'`.

BirdeeHub:
I dont know enough around the lua packaging to know if you missed something but its clear enough that I can at least read it. I checked out your fork, added your wrapper changes to nixCats and it builds so, all good for me. I do have to copy in the changes to wrapper.nix though into nixCats when you push this and it hits live because i needed to do some hackery to install multiple versions of neovim XD

pkgs/applications/editors/neovim/wrapper.nix
```diff
@@ -13,6 +13,9 @@
 neovim-unwrapped:
 
 let
+  # inherit interpreter from neovim
+  lua = neovim-unwrapped.lua;
```
BirdeeHub, line 17: could we get this from callPackage above if it doesnt need to match neovim's version and only the utils are needed? It would reduce error

pkgs/applications/editors/neovim/wrapper.nix
```diff
@@ -160,7 +167,17 @@ let
       + ''
         rm $out/bin/nvim
         touch $out/rplugin.vim
-        makeWrapper ${lib.escapeShellArgs finalMakeWrapperArgs} ${wrapperArgsStr}
+
```
BirdeeHub, line 170: should we check for existence of the path first?

pkgs/applications/editors/neovim/utils.nix
```diff
@@ -229,12 +229,36 @@ let
...
+  packDir = packages:
+  let
+    rawPackDir = vimUtils.packDir packages;
+
+  in
+    rawPackDir.override ({
+    postBuild = ''
+      mkdir $out/nix-support
+      for i in $(find -L $out -name propagated-build-inputs ); do
+        cat "$i" >> $out/nix-support/propagated-build-inputs
```
mrcjkb, line 250: question: if multiple packages have the same dependency, would this result in duplicate entries? 
If so, could this be problematic?

pkgs/applications/editors/neovim/wrapper.nix
```diff
@@ -24,6 +27,8 @@ let
     , withNodeJs ? false
     , withPerl ? false
     , rubyEnv ? null
+
+    # wether to create symlinks in $out/bin/vi(m) -> $out/bin/nvim
```
mrcjkb, line 31: quibble: typo

```suggestion
    # whether to create symlinks in $out/bin/vi(m) -> $out/bin/nvim
```

mrcjkb:
> Could you test please (cherry-pick on top of unstable) ? 

I'm sick right now, but hopefully I can test this in a few days 😄

pkgs/applications/editors/neovim/wrapper.nix
teto: it does need to match the neovim interpreter. the default LUA_PATH differs between lua versions. And when loading shared libraries, ABI differs as well

pkgs/applications/editors/neovim/wrapper.nix
teto: this wraps the neovim executable: if it didn't exist the build wouldn't work. (i.e., it exists)

pkgs/applications/editors/neovim/utils.nix
teto: this would result in duplicate entries. I think lua will load the first plugin on the LUA_PATH. 
Until we have lua lockfiles, if you have several different lua dependencies, I guess you are actively screwing up since in nixpkgs we only keep one version of each lua plugin.

teto:
> Edit: actually, I do have 1 nitpick. In the wrapper.nix file we should maybe guard the search with an if check like the following in cae someone is using an old neovim version but the new wrapper or something due to using the module from unstable but the neovim from stable or something

Do you have an exemple in mind where there would be *breakage* ?  the API doesn't change soi I have trouble envisioning one,the hook should do nothing if file/folder doesn't work. Anyway nixos-unstable doesn't need to work with 24.05 code.

Wish the best recovery to mister rocks.nvim ;)

BirdeeHub:
> Do you have an exemple in mind where there would be *breakage* ?  the API doesn't change soi I have trouble envisioning one,the hook should do nothing if file/folder doesn't work. Anyway nixos-unstable doesn't need to work with 24.05 code.


So, if they built their own nvim in a flake, and updated their main config and not the flake, because the Lua you get the utils from is the one passed in via neovim-unwrapped they could hit an issue. However if you got luajit from callPackage instead of neovim-unwrapped it would avoid this issue, as would adding the check, so only one of the changes I suggested are needed to avoid that case.

pkgs/applications/editors/neovim/wrapper.nix
BirdeeHub: Ah. Ok.

pkgs/applications/editors/neovim/wrapper.nix
BirdeeHub: Ah. I suppose then just the which is relevant

mrcjkb:
> this would happen for everyone who had pinned their version of the nightly overlay and then update their nixpkgs for everything else, causing their wrapNeovim function to update but be pulling the old version of lua that the flake pulled, which would cause the build to fail

This isn't a problem, because nixpkgs doesn't need to be backwards compatible with older versions of itself.

BirdeeHub:
Carry on then. Ill be adding the check to nixCats though just in case because I was able to break it when testing the changes in that manner. But that doesnt need to match what is done in nixpkgs.

teto:
@BirdeeHub thanks for the feedback.

`nix-build -A lua5_1.tests luajit.tests knot-resolver` pass. I have driven this for a few days without issues. Let's unlock the power of the lua ecosystem and finally enable rocks.nvim and neorg/orgmode in nixpkgs.

I will merge and do further testing with the previouysly mentioned plugins. Having it in staging lets us some time to correct course should we find issues.

BirdeeHub:
whoops... ammending the commit before the one that mentioned this one was a mistake apparently sorry for the spam

pkgs/development/interpreters/lua-5/utils.sh
```diff
@@ -49,6 +93,43 @@ addToLuaPath() {
...
+#   # variables.
+  # declare -gA luaPathsSeen=()
+#   # shellcheck disable=SC2034
+  program_PATH=
+  luaPathsSeen["@lua@"]=1
+#   addToSearchPath program_PATH @lua@/bin
+  for path in $luaPath; do
+    _addToLuaPath "$path"
+  done
+}
```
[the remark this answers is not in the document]
teto: feeel free to remove the commented code. It's a merge where I forgot to remove dead code.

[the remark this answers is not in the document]
BirdeeHub: program_PATH=
  luaPathsSeen["@lua@"]=1
  
  ^ is this not being replaced when the script is made available and ran?

I didnt write the code mentioned but I THOUGHT I knew what that did?
  
  I thought nix gets it from lua.luaPathSearchPaths and puts it in there with substituteAll?
  
  Does it not? Maybe someone forgot and now its been bugged this whole time?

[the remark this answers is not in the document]
teto: IMO if it went unnoticed dont worry about it and change what you need. 
There are some tests in `nix-build -A lua.tests neovim.tests` but probably some of them are already broken on master so dont waste time on this.

PR #319348, opened 2024-06-12
 misskey: init at 2024.5.0 & nixos/misskey: init module

feathecutie:
## Description of changes

This adds the Misskey blogging platform by providing a package and accompanying NixOS module.

Source code: https://github.com/misskey-dev/misskey
Homepage: https://misskey-hub.net/en/

This PR is largely inspired by https://github.com/NixOS/nixpkgs/pull/296697.

### TODO / Nice to have:

- [x] Add the ability to pass secrets via files that are read at runtime.
  - Currently only for `settings.{db.pass,redis.pass,meilisearch.apiKey}`. 
- [x] Correctly type all config options in the NixOS module, as there aren't that many.
  - Created typed options with defaults for the most important options, while leaving the rest freeform. 
- [x] Split out the config options that affect PostgreSQL/Redis/Nginx/Caddy instead of reading them from the untyped `settings` field.
  - Not split out from settings, but all relevant fields are properly typed now.
- [x] Assert the `settings` to be in a state where Misskey doesn't crash at runtime, e.g. when certain options are `null`.
  - All relevant fields have proper types and defaults and can't be null.
- [x] Automatically configure SSL for Nginx (I never used Nginx before).
  - [ ] Potentially detect if SSL is even necessary by checking if `settings.url` starts with `https://` or `http://`. This seems quite error prone, so a seperate toggle for SSL in nginx might make sense. Luckily, Caddy handles this automatically based on the URL type.
  - [ ] Alternative: Remove the option for auto-configuring a Nginx reverse proxy and let the user set it up completely themselves. This seems less ergonomic, but would leave the specifics of setting up SSL (among other things) to the user.

Feedback is welcome for any of the remaining (or even completed) TODOs.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [x] Tested, as applicable:

---

### DIFF
--- a/nixos/modules/services/web-apps/misskey.nix
+++ b/nixos/modules/services/web-apps/misskey.nix
@@ -1,0 +1,418 @@
+{
+  config,
+  pkgs,
+  lib,
+  ...
+}:
+
+let
+  cfg = config.services.misskey;
+  settingsFormat = pkgs.formats.yaml { };
+  redisType = lib.types.submodule {
+    freeformType = lib.types.attrsOf settingsFormat.type;
+    options = {
+      host = lib.mkOption {
+        type = lib.types.str;
+        default = "localhost";
+        description = "The Redis host.";
+      };
+      port = lib.mkOption {
+        type = lib.types.port;
+        default = 6379;
+        description = "The Redis port.";
+      };
+    };
+  };
+  settings = lib.mkOption {
+    description = ''
+      Configuration for Misskey, see
+      [`example.yml`](https://github.com/misskey-dev/misskey/blob/develop/.config/example.yml)
+      for all supported options.
+    '';
+    type = lib.types.submodule {
+      freeformType = lib.types.attrsOf settingsFormat.type;
+      options = {
+        url = lib.mkOption {
+          type = lib.types.str;
+          example = "https://example.tld/";
+          description = ''
+            The final user-facing URL. Do not change after running Misskey for the first time.
+
+            This needs to match up with the configured reverse proxy and is automatically configured when using `services.misskey.reverseProxy`.
+          '';
+        };
+        port = lib.mkOption {
+          type = lib.types.port;
+          default = 3000;
+          description = "The port your Misskey server should listen on.";
+        };
+        socket = lib.mkOption {
+          type = lib.types.nullOr lib.types.path;
+          default = null;
+          example = "/path/to/misskey.sock";
+          description = "The UNIX socket your Misskey server should listen on.";
+        };
+        chmodSocket = lib.mkOption {
+          type = lib.types.nullOr lib.types.str;
+          default = null;
+          example = "777";
+          description = "The file access mode of the UNIX socket.";
+        };
+        db = lib.mkOption {
+          description = "Database settings.";
+          type = lib.types.submodule {
+            options = {
+              host = lib.mkOption {
+                type = lib.types.str;
+                default = "/var/run/postgresql";
+                example = "localhost";
+                description = "The PostgreSQL host.";
+              };
+              port = lib.mkOption {
+                type = lib.types.port;
+                default = 5432;
+                description = "The PostgreSQL port.";
+              };
+              db = lib.mkOption {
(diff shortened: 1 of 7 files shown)


nixos/modules/services/web-apps/misskey.nix
```diff
@@ -0,0 +1,361 @@
...
+      }) // (lib.optionalAttrs (cfg.database.passwordFile != null) {
+        pass = "@DATABASE_PASSWORD@";
+      });
+      redis = (lib.optionalAttrs cfg.redis.createLocally {
+        host = "localhost";
+      }) // (lib.optionalAttrs (cfg.redis.passwordFile != null) {
+        pass = "@REDIS_PASSWORD@";
+      });
+      meilisearch =
+        if (cfg.meilisearch.createLocally || (cfg.meilisearch.keyFile != null)) then
```
feathecutie, line 259: This `if` expression breaks the ability to configure meilisearch via `settings.meilisearch`, resulting in `The option `services.misskey.settings.meilisearch` is defined both null and not null`.

The reason why I can't simply use `lib.optionalAttrs` here is that `settings.meilisearch = null` and `settings.meilisearch = { }` are parsed differently by Misskey. `settings.meilisearch` being `null` means Misskey won't try to use meilisearch, while `settings.meilisearch` being `{ }` results in an error.

I'll need to look into how to do this differently.

nixos/modules/services/web-apps/misskey.nix
```diff
@@ -0,0 +1,361 @@
...
+          extraConfig = ''
+            reverse_proxy localhost:${toString cfg.settings.port}
+          '';
+        }
+      ];
+    };
+
+    services.nginx.virtualHosts = lib.mkIf (cfg.reverseProxy.enable && cfg.reverseProxy.webserver ? nginx) {
+      ${cfg.settings.url} = lib.mkMerge [
+        cfg.reverseProxy.webserver.nginx
```
feathecutie, line 345: Setting `reverseProxy.webserver.nginx.enableACME = true` results in the error
```
A definition for option `systemd.services."acme-https://example.com".after."[definition 1-entry 5]"' is not of type `string matching the pattern [a-zA-Z0-9@%:_.\-]+[.](service|socket|device|mount|automount|swap|target|path|timer|scope|slice)
```

I believe this is the case because that the name of the Nginx virtualHost should not include the protocol (`http://` or `https://`). This could be fixed by either removing these prefixes from `cfg.settings.url` when using it as the virtualHost name, or by exposing seperate options for setting the protocol and the domain name itself.

The first option seems very error prone, but the second option does not integrate into the existing configuration system of simply parsing configuration options from `settings`.

I'll test a few different options to try to resolve this and see what feels and works best.

pkgs/by-name/mi/misskey/package.nix
```diff
@@ -0,0 +1,124 @@
...
+
+  nativeBuildInputs = [
+    nodejs
+    pnpm.configHook
+    makeWrapper
+    python3
+  ];
+
+  patches = [
+    ./pnpm_version.patch
```
JulienMalka, line 37: Could you add a comment here explaining the reason for the patch and if there is a point in time / a condition when it should be dropped ?

JulienMalka (approved):
Code looks good to me appart from minor comment

JulienMalka:
Before merging please squash your changes so that there is one that adds yourself as maintainer, one that adds the package, one that adds the module

pkgs/by-name/mi/misskey/package.nix
feathecutie: ..appearently that patch is not necessary after all, I thought I got build failures without it in the beginning but the package seems to build fine without it, so I removed it again.

fricklerhandwerk:
Thanks a lot @feathecutie!

PR #319401, opened 2024-06-12
munge: Fix installation paths

jtojnar:
## Description of changes

Removing `src/etc/Makefile` is overly intrusive, as the file is responsible for installing pkg-config file.  The issue when it would try to write to the specified out-of-store paths can be easily resolved by overriding them to ones within `$out` during installation.

Also set `sysconfdir` path properly to allow configuring it with `environment.etc`. There should be no problem with missing files, as nothing was installed to `$out/etc` previously either, other than an empty `munge` directory.

`runstatedir` defaults to `$(localstatedir)/run` but that has long been deprecated in favour of `/run`, so let’s fix that as well.

`pkgconfigdir` and `sysconfigdir` (not to be confused with the standard `sysconfdir`) rely on FHS for proper detection and `systemdunitdir` tries to run `systemd --version`, let’s just hardcode them.  `sysconfigdir` is mentioned in the newly installed `munge.service` and a file within is loaded as `EnvironmentFile` when it exists so let’s override the path for installation so the file can serve as a template in `$out`. The other two are installation-only so we can directly set them to `$out` subdirectories.

Also clean up the expression.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/tools/security/munge/default.nix
+++ b/pkgs/tools/security/munge/default.nix
@@ -1,42 +1,63 @@
-{ lib, stdenv, fetchFromGitHub, autoreconfHook, libgcrypt, zlib, bzip2 }:
+{
+  lib,
+  stdenv,
+  fetchFromGitHub,
+  autoreconfHook,
+  libgcrypt,
+  zlib,
+  bzip2,
+}:
 
-stdenv.mkDerivation rec {
+stdenv.mkDerivation (finalAttrs: {
   pname = "munge";
   version = "0.5.16";
 
   src = fetchFromGitHub {
     owner = "dun";
     repo = "munge";
-    rev = "${pname}-${version}";
+    rev = "munge-${finalAttrs.version}";
     sha256 = "sha256-fv42RMUAP8Os33/iHXr70i5Pt2JWZK71DN5vFI3q7Ak=";
   };
 
-  strictDeps = true;
   nativeBuildInputs = [
     autoreconfHook
     libgcrypt # provides libgcrypt.m4
   ];
-  buildInputs = [ libgcrypt zlib bzip2 ];
+
+  buildInputs = [
+    libgcrypt
+    zlib
+    bzip2
+  ];
+
+  strictDeps = true;
+
+  configureFlags = [
+    "--localstatedir=/var"
+
+    # Cross-compilation hacks
+    "--with-libgcrypt-prefix=${lib.getDev libgcrypt}"
+    # workaround for cross compilation: https://github.com/dun/munge/issues/103
+    "ac_cv_file__dev_spx=no"
+    "x_ac_cv_check_fifo_recvfd=no"
+  ];
 
   preAutoreconf = ''
     # Remove the install-data stuff, since it tries to write to /var
     substituteInPlace src/Makefile.am --replace "etc \\" "\\"
   '';
 
-  configureFlags = [
-    "--localstatedir=/var"
-    "--with-libgcrypt-prefix=${libgcrypt.dev}"
-    # workaround for cross compilation: https://github.com/dun/munge/issues/103
-    "ac_cv_file__dev_spx=no"
-    "x_ac_cv_check_fifo_recvfd=no"
-  ];
-
   meta = with lib; {
     description = ''
       An authentication service for creating and validating credentials
     '';
-    license = licenses.lgpl3;
+    license = [
+      # MUNGE
+      licenses.gpl3Plus
+      # libmunge
+      licenses.lgpl3Plus
+    ];
     platforms = platforms.unix;
     maintainers = [ maintainers.rickynils ];
   };
-}
+})
--- a/pkgs/tools/security/munge/default.nix
+++ b/pkgs/tools/security/munge/default.nix
@@ -33,7 +33,15 @@ stdenv.mkDerivation (finalAttrs: {
   strictDeps = true;
 
   configureFlags = [
+    # Load data from proper global paths
     "--localstatedir=/var"
+    "--sysconfdir=/etc"
+    "--runstatedir=/run"
+    "--with-sysconfigdir=/etc/default"
+
+    # Install data to proper directories
+    "--with-pkgconfigdir=${placeholder "out"}/lib/pkgconfig"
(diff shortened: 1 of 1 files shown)


veprbl:
Result of `nixpkgs-review pr 319401` run on x86_64-darwin [1](https://github.com/Mic92/nixpkgs-review)
<details>
  <ul>
    <li>munge</li>
  </ul>
</details>

markuskowa:
@GrahamcOfBorg test slurm

pkgs/tools/security/munge/default.nix
```diff
@@ -1,42 +1,73 @@
...
+    "--with-systemdunitdir=${placeholder "out"}/lib/systemd/system"
+
+    # Cross-compilation fixes
     "--with-libgcrypt-prefix=${libgcrypt.dev}"
     # workaround for cross compilation: https://github.com/dun/munge/issues/103
     "ac_cv_file__dev_spx=no"
     "x_ac_cv_check_fifo_recvfd=no"
   ];
 
+  installFlags = [
```
markuskowa, line 53: I am not sure what the purpose of `localstatedir`, `runstatedir` and `sysconfdir` is here. The installer just creates empty directories for these in `$out`. If it is just to avoid the installer accessing paths like /etc, /var, ... can we just direct these to to a dummy location so we do not have these empty directories in `$out`?

pkgs/tools/security/munge/default.nix
```diff
@@ -1,42 +1,73 @@
...
+    "--sysconfdir=/etc"
+    "--runstatedir=/run"
+    "--with-sysconfigdir=/etc/default"
+
+    # Install data to proper directories
+    "--with-pkgconfigdir=${placeholder "out"}/lib/pkgconfig"
+    "--with-systemdunitdir=${placeholder "out"}/lib/systemd/system"
+
+    # Cross-compilation fixes
     "--with-libgcrypt-prefix=${libgcrypt.dev}"
```
markuskowa, line 47:
```suggestion
    "--with-libgcrypt-prefix=${lib.getDev libgcrypt}"
```

jtojnar: I am not even sure if this is still necessary. \<rant\>Cross people keep introducing junk like this without any mention how to reproduce the failure or attempts to resolve this upstream. I am half inclined to just revert <https://github.com/NixOS/nixpkgs/commit/faacc88c93086982e2c95708176863f415e03810> @NickCao\</rant\>

pkgs/tools/security/munge/default.nix
markuskowa: Good point. I would be for deleting `$out/[var,etc/munge]`

pkgs/tools/security/munge/default.nix
SomeoneSerge: +1 to `getDev`, but I'm not sure I understand what issue the comment refers to (even though I suspect I've seen it before)

markuskowa (approved):

pkgs/tools/security/munge/default.nix
jtojnar: Oops, I previously pushed `echo rmdir`. Fixed now and double checked that the directories are removed.

pkgs/tools/security/munge/default.nix
jtojnar: Okay, done.

By the way, I managed to reproduce a cross failure with `pkgsCross.aarch64-multiplatform.munge` when I revert https://github.com/NixOS/nixpkgs/commit/faacc88c93086982e2c95708176863f415e03810.

The `libgcrypt.dev` in `nativeBuildInputs` is needed for `libgcrypt.m4` to be picked up for `ACLOCAL_PATH` (see also https://github.com/NixOS/nixpkgs/pull/311773#discussion_r1612536323). And `--with-libgcrypt-prefix` is probably needed because the m4 file/`libgcrypt-config` cannot handle multiple outputs.

It appears possible to force it to use `pkg-config` but that is no less hacky:

<details>
<summary>Patch</summary>


```patch
diff --git a/pkgs/tools/security/munge/default.nix b/pkgs/tools/security/munge/default.nix
index f21a9e17add3..1a7ada38fa95 100644
--- a/pkgs/tools/security/munge/default.nix
+++ b/pkgs/tools/security/munge/default.nix
@@ -1,6 +1,7 @@
 {
   lib,
   stdenv,
+  pkg-config,
   fetchFromGitHub,
   autoreconfHook,
   libgcrypt,
@@ -21,7 +22,7 @@ stdenv.mkDerivation (finalAttrs: {
 
   nativeBuildInputs = [
     autoreconfHook
-    libgcrypt # provides libgcrypt.m4
+    pkg-config
   ];
 
   buildInputs = [
@@ -44,7 +45,7 @@ stdenv.mkDerivation (finalAttrs: {
     "--with-systemdunitdir=${placeholder "out"}/lib/systemd/system"
 
     # Cross-compilation hacks
-    "--with-libgcrypt-prefix=${lib.getDev libgcrypt}"
+    "GPGRT_CONFIG=${stdenv.cc.targetPrefix}pkg-config"
     # workaround for cross compilation: https://github.com/dun/munge/issues/103
     "ac_cv_file__dev_spx=no"
     "x_ac_cv_check_fifo_recvfd=no"
@@ -57,6 +58,11 @@ stdenv.mkDerivation (finalAttrs: {
     "sysconfigdir=${placeholder "out"}/etc/default"
   ];
 
+  preAutoreconf = ''
+    # For some reason, automake setup hook does not pick it up from buildInputs.
+    addAclocals "${lib.getDev libgcrypt}"
+  '';
+
   postInstall = ''
     # rmdir will notify us if anything new is installed to the directories.
     rmdir "$out"/{var{/{lib,log}{/munge,},},etc/munge}
```

</details>

Ideally, all of this would be resolved upstream but `m4` is the opposite of exciting to work with.

PR #319432, opened 2024-06-13
python: automate pythonRelaxDepsHook

superherointj:
python: automate pythonRelaxDepsHook

As a test, removed `pythonRelaxDepsHook` from:
* libretranslate
* aio-geojson-generic-client
* gimme-aws-creds
* robomachine
* django-two-factor-auth

Seems to work fine.

Pending work unplanned for this PR (because this has to be accepted first):

- Remove `pythonRelaxDepsHook` globally.
- Update documentation.

CC @thiagokokada @mweinelt @fabaff @marsam @tjni @jonringer 

Considering `stable`, a back-port is important because any package not including `pythonRelaxDepsHook` will break. I've tested back-port to `release-24.05` and merges cleanly. But then, if this causes a mass rebuild. What should be done?

### DIFF
--- a/pkgs/development/interpreters/python/mk-python-derivation.nix
+++ b/pkgs/development/interpreters/python/mk-python-derivation.nix
@@ -17,6 +17,7 @@
 , pythonImportsCheckHook
 , pythonNamespacesHook
 , pythonOutputDistHook
+, pythonRelaxDepsHook
 , pythonRemoveBinBytecodeHook
 , pythonRemoveTestsDirHook
 , pythonRuntimeDepsCheckHook
@@ -213,7 +214,11 @@ let
 
   isBootstrapPackage = isBootstrapInstallPackage || isBootstrapPackage' (attrs.pname or null);
 
+  isNonEmptyList = v: builtins.isList v && builtins.length v > 0;
+
   isSetuptoolsDependency = isSetuptoolsDependency' (attrs.pname or null);
+
+  isTrue = v: builtins.isBool v && v;
 
   passthru =
     attrs.passthru or { }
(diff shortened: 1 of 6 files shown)


thiagokokada (approved):

superherointj:
@ofborg build python3Packages.libretranslate python3Packages.aio-geojson-generic-client python3Packages.gimme-aws-creds python3Packages.robomachine python3Packages.django-two-factor-auth

pkgs/development/interpreters/python/mk-python-derivation.nix
```diff
@@ -290,6 +295,11 @@ let
     ] ++ optionals (python.pythonAtLeast "3.3") [
       # Optionally enforce PEP420 for python3
       pythonNamespacesHook
+    ] ++ lib.optionals (
+      isTrue (attrs.pythonRelaxDeps or false) || isNonEmptyList (attrs.pythonRelaxDeps or []) ||
+      isTrue (attrs.pythonRemoveDeps or false) || isNonEmptyList (attrs.pythonRemoveDeps or [])
```
natsukium, line 300: Is this expression equivalent to this one?

```nix
attrs?pythonRelaxDeps || attrs?pythonRemoveDeps
```

superherointj: I believe the `?` operator only checks for the existence of an attribute: [builtins.hasAttr](https://nix.dev/manual/nix/2.22/language/builtins#builtins-hasAttr)
Goal was to test if the value is a boolean (and true) or a list with items in it.
It shouldn't add the hook for invalid values (string/int/null/etc).

natsukium: We have no type to prevent them (string/int/null/etc), but such values should not be set in the first place.
Since it is not a heavy hook, adding it by mistake is not so bad, IMO.

superherointj:
Closed in favor of #319573

PR #319435, opened 2024-06-13
nixos/xss-lock: warn and exit if DISPLAY unset

presto8:
## Description of changes

As discussed in #310880, the xss-lock service will run without a display manager provided that the necessary environment variables are imported into systemd.

This changeset removes nixos warning at build time and instead adds a pre-condition to the systemd service: if startx is enabled, and DISPLAY is not set, the service will display a message with a tip to how to fix it and exit.

## Things done

- Built on platform(s)
  - [X] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

### DIFF
--- a/nixos/modules/programs/xss-lock.nix
+++ b/nixos/modules/programs/xss-lock.nix
@@ -31,6 +31,13 @@ in
       description = "XSS Lock Daemon";
       wantedBy = [ "graphical-session.target" ];
       partOf = [ "graphical-session.target" ];
+      preStart = ''
+        if [[ -z "$DISPLAY" ]]; then
+          echo "can't run xss-lock without DISPLAY set"
+          echo "if you are using X without a display manager, please see: https://nixos.wiki/wiki/Using_X_without_a_Display_Manager"
+          exit 1
+        fi
+      '';
       serviceConfig.ExecStart =
         builtins.concatStringsSep " " ([
             "${pkgs.xss-lock}/bin/xss-lock" "--session \${XDG_SESSION_ID}"
@@ -40,10 +47,5 @@ in
         ]);
       serviceConfig.Restart = "always";
     };
-
-    warnings = lib.mkIf (config.services.xserver.displayManager.startx.enable) [
-      "xss-lock service only works if a displayManager is set; it doesn't work when services.xserver.displayManager.startx.enable = true"
-    ];
-
   };
 }


presto8:
I tested this with a dummy variable, but I do not have a system without a display manager to test it on. I would appreciate if somebody with such a system could test it out by disabling the import of DISPLAY into systemd and then seeing if this service fails. Thank you!

nixos/modules/programs/xss-lock.nix
```diff
@@ -31,6 +31,13 @@ in
       description = "XSS Lock Daemon";
       wantedBy = [ "graphical-session.target" ];
       partOf = [ "graphical-session.target" ];
+      serviceConfig.ExecStartPre = lib.mkIf (config.services.xserver.displayManager.startx.enable) pkgs.writers.writeBash "xss-lock-prereqs" ''
```
rnhmjoj, line 34: You're missing parenthesis around writeBash. Anyway, it's better to do:
```suggestion
      preStart = lib.mkIf config.services.xserver.displayManager.startx.enable)  ''
```

rnhmjoj:
Besides the eval error, it works as you would expect:
image

NixOS test if you want to try:
<details>

```nix
import ./make-test-python.nix ({ pkgs, ...} : {
  name = "simple";
  meta = with pkgs.lib.maintainers; {
    maintainers = [ ];
  };

  nodes.machine = { config, lib, ... }:
  let
    # X session scripts
    session = with builtins;
      head (filter (x: x.name == "bspwm")
            config.services.xserver.windowManager.session);

    xinitrc = pkgs.writers.writeDashBin "xinitrc" ''
      # load resources
      xrdb -load "$XDG_CONFIG_HOME/x/resources"

      # start user services
      #systemctl --user import-environment DISPLAY SSH_AUTH_SOCK XDG_SESSION_ID
      systemctl --user start nixos-fake-graphical-session.target

      # start programs
      dbus-daemon --session --address="unix:path=/run/user/$(id -u)/bus" &
      xterm &
      ${session.start}
      wait $waitPID

      # stop programs and services
      systemctl --user stop nixos-fake-graphical-session.target
      kill 0
    '';

    custom-startx = pkgs.writers.writeDashBin "startx" ''
      mkdir -p "$XDG_CACHE_HOME/x"
      export XINITRC=${xinitrc}/bin/xinitrc
      export XSERVERRC=/etc/X11/xinit/xserverrc
      exec ${pkgs.xorg.xinit}/bin/xinit "$@" -- $XSERVERRC >/dev/null 2>&1
    '';

  in {
    #imports = [ ../modules/profiles/minimal.nix ];
    environment.systemPackages = lib.singleton (lib.hiPrio custom-startx);

    services.xserver =
      { enable = true;
        windowManager.bspwm.enable = true;
        displayManager.startx.enable = true;
        displayManager.xserverArgs = lib.mkAfter [ "vt$XDG_VTNR" ];
      };

    security.polkit.enable = true;

    programs.xss-lock =
      { enable = true;
        lockerCommand = lib.getExe pkgs.i3lock;
        extraOptions = [ "--transfer-sleep-lock" ];
      };
    };

  testScript =
    ''
      start_all()
      machine.wait_for_unit("multi-user.target")
      machine.shutdown()
    '';
})
```

</details>

rnhmjoj:
I wonder if we could add this check to every service in the `graphical-session.target` when using startx.

nixos/modules/programs/xss-lock.nix
presto8: Thanks, I didn't know about preStart. It runs it as bash? I'll give it a try.

presto8: Ok, I see. preStart is more or less the same as what I had, except more concise :)

presto8:
It seems like a good idea to add it to other services! I myself use xss-lock, redshift and greenclip, but in theory any service that relies on DISPLAY would fail. I'd be comfortable doing PRs for the ones I use, and any others except that maybe I should figure out how to re-set-up my startx test environment so I can test it!

turboMaCk:
Do we want to abstract it somehow to be reused across these? I think that would also have some implication on if this should come in separate PRs or one PR that changes this across many modules.

rnhmjoj:
Ideally it should be done startx.nix. Maybe by filtering the services in `systemd.services` that have `PartOf=graphical-session.target`. This should catch most of the them, unless the upstream unit file is being installed.

presto8:
That is a good idea. It generally works but it may pull in services that don't rely on DISPLAY, like the one I wrote below (disable-middle-paste). Also basic.target which is easily filtered out.

But how do we detect that DISPLAY was not exported to systemd from startx?

Another option is to have a "startx-display-check" service that would be created and added whenever startx is enabled. It would fail and display the message. Then xss-lock and all the other services would be left as-is. The message in the startx-display-check can list all of the dependencies of graphical-session that may fail.

```
❯ systemctl --user list-dependencies graphical-session.target
graphical-session.target
○ ├─disable-middle-paste.service
● ├─greenclip.service
● ├─redshift.service
● ├─xss-lock.service
● └─basic.target
●   ├─[...lots of services...]
```

turboMaCk:
In my view even listing all the services to check might be acceptable. In a way I would maybe prefer it. Sure the risk will always be we don't cover everything but covering at least something is better than covering nothing. While printing irrelevant warning in case we apply check to something irrelevant is more harmful then helpful. So in a way being helpful in 80% of cases while not harming anything seems like a good principle to me.

rnhmjoj:
> But how do we detect that DISPLAY was not exported to systemd from startx?

With the existing a preStart check.

> Another option is to have a "startx-display-check" service that would be created and added whenever startx is enabled. It would fail and display the message.

Yes, I also thought about this. It looks cleaner, but the problem is the error message may not so evident.
I think the failure would only be visible from `systemctl --user status`, right?

presto8:
> > But how do we detect that DISPLAY was not exported to systemd from startx?
> 
> With the existing preStart check.

Sorry, I don't follow: There isn't a startx service is there? Which service would the preStart check go in?

> > Another option is to have a "startx-display-check" service that would be created and added whenever startx is enabled. It would fail and display the message.
> 
> Yes, I also thought about this. It looks cleaner, but the problem is the error message may not so evident. I think the failure would only be visible from `systemctl --user status`, right?

As far as I know, it would only be available in systemctl --user status. Agree it is less obvious there.

If the check does live in the startx startup flow somewhere, it should have a variable to ignore/override the check so the user can get X running again once the message has been delivered (even though the affected services would still be unavailable until the DISPLAY issue is fixed).

PR #319516, opened 2024-06-13
wstunnel: 9.6.2 -> 9.7.0

r-vdp:
## Description of changes

This also removes the top-level attribute for the old Haskell-based implementation in favour of the new Rust implementation by the same author.
The Haskell-based version is still accessible from `haskellPackages`.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/nixos/modules/services/networking/wstunnel.nix
+++ b/nixos/modules/services/networking/wstunnel.nix
@@ -277,7 +277,7 @@ let
         environment.RUST_LOG = serverCfg.loggingLevel;
 
         serviceConfig = {
-          Type = "simple";
+          Type = "exec";
           EnvironmentFile =
             lib.optional (serverCfg.environmentFile != null) serverCfg.environmentFile;
           DynamicUser = true;
@@ -334,7 +334,7 @@ let
       environment.RUST_LOG = clientCfg.loggingLevel;
 
       serviceConfig = {
-        Type = "simple";
+        Type = "exec";
         EnvironmentFile =
           lib.optional (clientCfg.environmentFile != null) clientCfg.environmentFile;
         DynamicUser = true;
--- a/nixos/modules/services/networking/wstunnel.nix
+++ b/nixos/modules/services/networking/wstunnel.nix
@@ -1,83 +1,94 @@
-{ config, lib, options, pkgs, utils, ... }:
-with lib;
+{ config
+, lib
+, pkgs
+, ...
+}:
+
 let
   cfg = config.services.wstunnel;
-  attrsToArgs = attrs: utils.escapeSystemdExecArgs (
-    mapAttrsToList
-    (name: value: if value == true then "--${name}" else "--${name}=${value}")
-    attrs
-  );
 
-  hostPortToString = { host, port }: "${host}:${builtins.toString port}";
+  hostPortToString = { host, port }: "${host}:${toString port}";
 
   hostPortSubmodule = {
     options = {
-      host = mkOption {
+      host = lib.mkOption {
         description = "The hostname.";
-        type = types.str;
+        type = lib.types.str;
       };
-      port = mkOption {
+      port = lib.mkOption {
         description = "The port.";
-        type = types.port;
+        type = lib.types.port;
       };
     };
   };
 
   commonOptions = {
-    enable = mkOption {
-      description = "Whether to enable this `wstunnel` instance.";
-      type = types.bool;
+    enable = lib.mkEnableOption "this `wstunnel` instance." // {
       default = true;
     };
 
-    package = mkPackageOption pkgs "wstunnel" {};
+    package = lib.mkPackageOption pkgs "wstunnel" { };
 
-    autoStart = mkOption {
-      description = "Whether this tunnel server should be started automatically.";
-      type = types.bool;
-      default = true;
-    };
+    autoStart =
+      lib.mkEnableOption "starting this wstunnel instance automatically." // {
+        default = true;
+      };
 
-    extraArgs = mkOption {
-      description = "Extra command line arguments to pass to `wstunnel`. Attributes of the form `argName = true;` will be translated to `--argName`, and `argName = \"value\"` to `--argName=value`.";
-      type = with types; attrsOf (either str bool);
-      default = {};
+    extraArgs = lib.mkOption {
+      description = ''
+        Extra command line arguments to pass to `wstunnel`.
+        Attributes of the form `argName = true;` will be translated to `--argName`,
+        and `argName = \"value\"` to `--argName value`.
+      '';
+      type = with lib.types; attrsOf (either str bool);
+      default = { };
       example = {
         "someNewOption" = true;
         "someNewOptionWithValue" = "someValue";
       };
     };
 
-    loggingLevel = mkOption {
+    loggingLevel = lib.mkOption {
       description = ''
         Passed to --log-lvl
 
         Control the log verbosity. i.e: TRACE, DEBUG, INFO, WARN, ERROR, OFF
         For more details, checkout [EnvFilter](https://docs.rs/tracing-subscriber/latest/tracing_subscriber/filter/struct.EnvFilter.html#example-syntax)
       '';
-      type = types.nullOr types.str;
+      type = lib.types.nullOr lib.types.str;
       example = "INFO";
       default = null;
     };
 
-    environmentFile = mkOption {
-      description = "Environment file to be passed to the systemd service. Useful for passing secrets to the service to prevent them from being world-readable in the Nix store. Note however that the secrets are passed to `wstunnel` through the command line, which makes them locally readable for all users of the system at runtime.";
-      type = types.nullOr types.path;
+    environmentFile = lib.mkOption {
+      description = ''
+        Environment file to be passed to the systemd service.
+        Useful for passing secrets to the service to prevent them from being
+        world-readable in the Nix store.
+        Note however that the secrets are passed to `wstunnel` through
+        the command line, which makes them locally readable for all users of
+        the system at runtime.
+      '';
+      type = lib.types.nullOr lib.types.path;
       default = null;
       example = "/var/lib/secrets/wstunnelSecrets";
     };
   };
 
-  serverSubmodule = { config, ...}: {
+  serverSubmodule = { config, ... }: {
     options = commonOptions // {
-      listen = mkOption {
-        description = "Address and port to listen on. Setting the port to a value below 1024 will also give the process the required `CAP_NET_BIND_SERVICE` capability.";
-        type = types.submodule hostPortSubmodule;
+      listen = lib.mkOption {
+        description = ''
+          Address and port to listen on.
+          Setting the port to a value below 1024 will also give the process
+          the required `CAP_NET_BIND_SERVICE` capability.
+        '';
+        type = lib.types.submodule hostPortSubmodule;
         default = {
           host = "0.0.0.0";
           port = if config.enableHTTPS then 443 else 80;
         };
-        defaultText = literalExpression ''
+        defaultText = lib.literalExpression ''
           {
             host = "0.0.0.0";
             port = if enableHTTPS then 443 else 80;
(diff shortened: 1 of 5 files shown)


Aleksanaa:
I opened a PR https://github.com/NixOS/nixpkgs/pull/289652 but if you want to proceed just go on.

The most problematic part is the nixos module, as the new rewrite uses different command line args.

r-vdp:
Oh, I should've searched for that first!

Yeah, as noted there, if the old package doesn't build anymore, we can probably just hijack the name.

I will have a look at the module, to bring it up to date with the rust version. A NixOS test would've been really nice for this, maybe we can work on adding that too.

h7x4:
I think #317840 might conflict with this PR. It was just merged

r-vdp:
That's fine, I'll rebase on top of it

nixos/modules/services/networking/wstunnel.nix
```diff
@@ -1,123 +1,159 @@
...
+    loggingLevel = lib.mkOption {
       description = ''
         Passed to --log-lvl
 
         Control the log verbosity. i.e: TRACE, DEBUG, INFO, WARN, ERROR, OFF
         For more details, checkout [EnvFilter](https://docs.rs/tracing-subscriber/latest/tracing_subscriber/filter/struct.EnvFilter.html#example-syntax)
       '';
-      type = types.nullOr types.str;
+      type = lib.types.nullOr (lib.types.enum [
+        "TRACE"
```
NeverBehave, line 65: Per description, it seems enum does not cover all cases of EnvFilter, such as `warn,tokio::net=info`

r-vdp: Hmm, good point, I can revert that part

r-vdp: Done!

NeverBehave (approved):
LGTM if the changes has tested locally. Only checked diff.

nixos/modules/services/networking/wstunnel.nix
```diff
@@ -228,44 +275,55 @@ let
...
+      in
+      ''
+        ${lib.getExe package} \
+          server \
+          ${lib.concatMapStringsSep " " (restriction: ''--restrict-to ${lib.escapeShellArg (hostPortToString restriction)}'') restrictTo} \
+          ${lib.optionalString (resolvedTlsCertificate != null) "--tls-certificate ${lib.escapeShellArg resolvedTlsCertificate}"} \
+          ${lib.optionalString (resolvedTlsKey != null) "--tls-private-key ${lib.escapeShellArg resolvedTlsKey}"} \
+          ${attrsToArgs extraArgs} \
+          ${lib.escapeShellArg "${if enableHTTPS then "wss" else "ws"}://${hostPortToString listen}"}
+      '';
```
h7x4, line 326:
```suggestion
      script = with serverCfg; ''
        ${lib.getExe package} server ${lib.cli.toGNUCommandLineShell { } (
          lib.recursiveUpdate
          {
            restrict-to = map hostPortToString restrictTo;
            tls-certificate = if useACMEHost != null
                              then "${certConfig.directory}/fullchain.pem"
                              else tlsCertificate;
            tls-private-key = if useACMEHost != null
                              then "${certConfig.directory}/key.pem"
                              else tlsKey;
          }
          extraAttrs
        )} ${lib.escapeShellArg "${if enableHTTPS then "wss" else "ws"}://${hostPortToString listen}"}
      '';
```

nixos/modules/services/networking/wstunnel.nix
```diff
@@ -228,44 +275,55 @@ let
       description = "wstunnel server - ${name}";
       requires = [ "network.target" "network-online.target" ];
       after = [ "network.target" "network-online.target" ];
-      wantedBy = optional serverCfg.autoStart "multi-user.target";
+      wantedBy = lib.optional serverCfg.autoStart "multi-user.target";
+
+      serviceConfig =
+        let
+          certConfig = config.security.acme.certs."${serverCfg.useACMEHost}";
```
h7x4, line 282:
```suggestion
          certConfig = config.security.acme.certs.${serverCfg.useACMEHost};
```

nixos/modules/services/networking/wstunnel.nix
```diff
@@ -126,95 +155,113 @@ let
...
-      soMark = mkOption {
-        description = "Mark network packets with the SO_MARK sockoption with the specified value. Setting this option will also enable the required `CAP_NET_ADMIN` capability for the systemd service.";
-        type = types.nullOr types.int;
+      soMark = lib.mkOption {
+        description = ''
+          Mark network packets with the SO_MARK sockoption with the specified value.
+          Setting this option will also enable the required `CAP_NET_ADMIN` capability
+          for the systemd service.
+        '';
+        type = lib.types.nullOr lib.types.int;
```
h7x4, line 209:
```suggestion
        type = lib.types.nullOr lib.types.ints.unsigned;
```

nixos/modules/services/networking/wstunnel.nix
```diff
@@ -126,95 +155,113 @@ let
...
           "unix:///tmp/wstunnel.sock:g.com:443"
         ];
       };
 
-      remoteToLocal = mkOption {
+      remoteToLocal = lib.mkOption {
         description = "Listen on remote and forwards traffic from local. Only tcp is supported";
-        type = types.listOf (types.str);
-        default = [];
+        type = lib.types.listOf (lib.types.str);
```
h7x4, line 176:
```suggestion
        type = lib.types.listOf lib.types.str;
```

nixos/modules/services/networking/wstunnel.nix
```diff
@@ -1,123 +1,152 @@
...
       };
 
-      restrictTo = mkOption {
-        description = "Accepted traffic will be forwarded only to this service. Set to `null` to allow forwarding to arbitrary addresses.";
-        type = types.listOf (types.submodule hostPortSubmodule);
-        default = [];
+      restrictTo = lib.mkOption {
+        description = ''
+          Accepted traffic will be forwarded only to this service.
+          Set to `null` to allow forwarding to arbitrary addresses.
```
h7x4, line 108: I know you have not edited this description, but could we have a look into what the original author was trying to say here? AFAIK, `null` is not an accepted value. Is arbitrary traffic accepted by default?

nixos/modules/services/networking/wstunnel.nix
```diff
@@ -1,123 +1,152 @@
...
-      description = "Extra command line arguments to pass to `wstunnel`. Attributes of the form `argName = true;` will be translated to `--argName`, and `argName = \"value\"` to `--argName=value`.";
-      type = with types; attrsOf (either str bool);
-      default = {};
+    extraArgs = lib.mkOption {
+      description = ''
+        Extra command line arguments to pass to `wstunnel`.
+        Attributes of the form `argName = true;` will be translated to `--argName`,
+        and `argName = \"value\"` to `--argName value`.
+      '';
+      type = with lib.types; attrsOf (either str bool);
```
h7x4, line 49: If you decide to go for the `lib.cli.toGNUCommandLineShell` thing, this can be expanded to allow for more types (I'm working on making this a `formats.cli` thing, but it's not finished yet).

```suggestion
      type = with lib.types; let
        atomType = nullOr (oneOf [ str int float bool package ]);
      in attrsOf (either atomType (listOf atomType));
```

nixos/modules/services/networking/wstunnel.nix
```diff
@@ -1,123 +1,152 @@
...
-  attrsToArgs = attrs: utils.escapeSystemdExecArgs (
-    mapAttrsToList
-    (name: value: if value == true then "--${name}" else "--${name}=${value}")
-    attrs
+
+  attrsToArgs = attrs: lib.escapeShellArgs (
+    lib.mapAttrsToList
+      (name: value: if value == true then "--${name}" else "--${name}=${value}")
+      attrs
   );
```
h7x4, line 14: If you decide to go for the `lib.cli.toGNUCommandLineShell` thing, this becomes obsolete.

```suggestion
```

nixos/modules/services/networking/wstunnel.nix
```diff
@@ -1,123 +1,152 @@
...
-    package = mkPackageOption pkgs "wstunnel" {};
+    package = lib.mkPackageOption pkgs "wstunnel" { };
 
-    autoStart = mkOption {
-      description = "Whether this tunnel server should be started automatically.";
-      type = types.bool;
-      default = true;
-    };
+    autoStart =
+      lib.mkEnableOption "Whether this tunnel server should be started automatically." // {
```
h7x4, line 39: This docstring becomes funny because of the inner working of `mkEnableOption`, [it wraps the given string in `"Whether to enable ${str}."`](https://github.com/NixOS/nixpkgs/blob/a2a0316319cc0625ab75baba5084aacb272ba6d0/lib/options.nix#L98-L105)

nixos/modules/services/networking/wstunnel.nix
```diff
@@ -356,28 +422,43 @@ in {
...
+            Options services.wstunnel.servers."${name}".useACMEHost and services.wstunnel.servers."${name}".{tlsCertificate, tlsKey} are mutually exclusive.
+          '';
+        })
+        cfg.servers) ++
+
+      (lib.mapAttrsToList
+        (name: serverCfg: {
+          assertion =
+            !((serverCfg.tlsCertificate != null || serverCfg.tlsKey != null) &&
+            !(serverCfg.tlsCertificate != null && serverCfg.tlsKey != null));
```
h7x4, line 446: This can be reduced to xnor

```suggestion
            (serverCfg.tlsCertificate == null && serverCfg.tlsKey == null)
            || (serverCfg.tlsCertificate != null && serverCfg.tlsKey != null)
```

nixos/modules/services/networking/wstunnel.nix
```diff
@@ -356,28 +422,43 @@ in {
...
+    systemd.services =
+      (lib.mapAttrs' generateServerUnit (lib.filterAttrs (n: v: v.enable) cfg.servers)) //
+      (lib.mapAttrs' generateClientUnit (lib.filterAttrs (n: v: v.enable) cfg.clients));
+
+    assertions =
+      (lib.mapAttrsToList
+        (name: serverCfg: {
+          assertion =
+            !(serverCfg.useACMEHost != null &&
+            (serverCfg.tlsCertificate != null || serverCfg.tlsKey != null));
```
h7x4, line 435: The fact that a certificate needs a key is an implementation detail from the perspective of the intent behind this assertion. I'm also a bit uncertain if this assertion should be here whenever `enableHttps = false`, it might make sense to prefix the assertion with `serverCfg.enableHttps -> <pred>`

```suggestion
            !(serverCfg.useACMEHost != null && serverCfg.tlsCertificate != null);
```

h7x4 (commented):
Here's a few suggestions, although some of them are comments about things you have not authored. Please do consider them though :)

pkgs/by-name/ws/wstunnel/package.nix
```diff
@@ -3,30 +3,46 @@
...
+    runHook preInstallCheck
+    actual="$($out/bin/wstunnel --version)"
+    expected="${pname} ${version}"
+    echo "Check that 'wstunnel --version' returns: $expected"
+    if [[ "$actual" != "$expected" ]]; then
+      echo "'wstunnel --version' returned: $actual"
+      exit 1
+    fi
+    runHook postInstallCheck
+  '';
```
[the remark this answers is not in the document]
r-vdp: Right, good point.

nixos/modules/services/networking/wstunnel.nix
r-vdp: The reference to null is a left-over from before we turned this option into a list, I removed that part.

Without any restrictions specified on the server, a client can indeed ask for any forwarding. Setting this option on the server, limits what a client can ask for.

nixos/modules/services/networking/wstunnel.nix
r-vdp: I will keep this one for later, since it's not directly useful currently.

r-vdp:
> Here's a few suggestions, although some of them are comments about things you have not authored. Please do consider them though :)

Thanks! I added most of them to this PR!

nixos/modules/services/networking/wstunnel.nix
```diff
@@ -126,95 +148,113 @@ let
...
+      tlsSNI = lib.mkOption {
         description = "Use this as the SNI while connecting via TLS. Useful for circumventing hostname-based firewalls.";
-        type = types.nullOr types.str;
+        type = lib.types.nullOr lib.types.str;
         default = null;
       };
 
-      tlsVerifyCertificate = mkOption {
+      tlsVerifyCertificate = lib.mkOption {
         description = "Whether to verify the TLS certificate of the server. It might be useful to set this to `false` when working with the `tlsSNI` option.";
```
h7x4, line 224: It would be very nice if we could add a note here about how this might be useful to disable when using tls with self signed certificates. Considering a lot of people might use self signed certs for this, I'd even consider disabling it by default.

h7x4:
Do you mind if I push a nixos test to this PR?

h7x4:
Would you mind adding a point to the release notes? Something like this maybe?

```
- `wstunnel` has had a major version upgrade that entailed rewriting the program in rust.
  The module was updated to accomodate for breaking changes without breaking the module API,
  but be aware that arguments listed in `services.wstunnel.{clients,servers}.<name>.extraArgs`
  might have been removed or modified upstream.
```

r-vdp:
> Would you mind adding a point to the release notes?

Yeah, I reworded it slightly since the previous PR did break the module API. I think this was hard to avoid given the upstream changes.

r-vdp:
I also updated the test to use the new, `runTest`-based way of defining tests, as recommended in the manual, and I changed it to use the existing fake certificates that we already have for other tests.

pkgs/by-name/ws/wstunnel/package.nix
```diff
@@ -1,32 +1,44 @@
...
 , rustPlatform
-, lib
+, testers
+, wstunnel
+, nixosTests
 }:
 
-rustPlatform.buildRustPackage rec {
+let
   pname = "wstunnel";
```
Aleksanaa, line 10: This is only used once

r-vdp: Right, I inlined it.

h7x4 (approved):

nixos/modules/services/networking/wstunnel.nix
r-vdp: Let's do this when we actually need it, I don't think we have a use for the other types right now.

JohnRTitor:
Result of `nixpkgs-review pr 319516` run on x86_64-linux [1](https://github.com/Mic92/nixpkgs-review)
<details>
  <ul>
    <li>nixos-install-tools</li>
  </ul>
</details>
<details>
  <ul>
    <li>wstunnel</li>
  </ul>
</details>

r-vdp:
Are we good to merge this?

h7x4:
I think so

PR #240780, opened 2023-06-30
nixos/doc/gnome: Document how python nautilus extensions work

doronbehar:
###### Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
- [ ] For non-Linux: Is `sandbox = true` set in `nix.conf`? (See [Nix manual](https://nixos.org/manual/nix/stable/command-ref/conf-file.html))
- [ ] Tested, as applicable:
  - [NixOS test(s)](https://nixos.org/manual/nixos/unstable/index.html#sec-nixos-tests) (look inside [nixos/tests](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests))
  - and/or [package tests](https://nixos.org/manual/nixpkgs/unstable/#sec-package-tests)
  - or, for functions and "core" functionality, tests in [lib/tests](https://github.com/NixOS/nixpkgs/blob/master/lib/tests) or [pkgs/test](https://github.com/NixOS/nixpkgs/blob/master/pkgs/test)
  - made sure NixOS tests are [linked](https://nixos.org/manual/nixpkgs/unstable/#ssec-nixos-tests-linking) to the relevant packages
- [ ] Tested compilation of all packages that depend on this change using `nix-shell -p nixpkgs-review --run "nixpkgs-review rev HEAD"`. Note: all changes have to be committed, also see [nixpkgs-review usage](https://github.com/Mic92/nixpkgs-review#usage)
- [ ] Tested basic functionality of all binary files (usually in `./result/bin/`)
- [23.11 Release Notes](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2311.section.md) (or backporting [23.05 Release notes](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2305.section.md))
  - [ ] (Package updates) Added a release notes entry if the change is major or breaking
  - [ ] (Module updates) Added a release notes entry if the change is significant
  - [ ] (Module addition) Added a release notes entry if adding a new NixOS module
- [ ] Fits [CONTRIBUTING.md](https://github.com/NixOS/nixpkgs/blob/master/CONTRIBUTING.md).

Issue #33277, opened 2018-01-01
gedit gsettings schema not reachable


tiramiseb:
## Issue description

```
$ gsettings list-schemas | grep -i gedit
$ 
```

I have also notices that `/nix/store/k5zrf5gz97938mymfsawgfncyyv1caly-gedit-3.22.1/share` is not in the `XDG_DATA_DIRS` environment variable, it may be the cause of all this.

I started a discussion about a similar topic on the mailing list, however please note this is not the same. I am now running NixOS unstable. On 17.09, the gsettings schemas were not overriden correctly. Here, the schema for gedit is "not there". The other schema I had problem with (Tilix) is now okay. This issue looks specific to gedit.

### Steps to reproduce

Install gedit. Try the aforementioned command.

## Technical details

 - system: `"x86_64-linux"`
 - host os: `Linux 4.9.73, NixOS, 18.03.git.0bbf671 (Impala)`
 - multi-user?: `yes`
 - sandbox: `no`
 - version: `nix-env (Nix) 1.11.16`
 - nixpkgs: `/nix/var/nix/profiles/per-user/root/channels/nixos/nixpkgs`

jtojnar:
As I mentioned in the mailing list:

> The problem is, gsettings (or dconf-editor) expect the schemas in some concrete locations. That obviously fails with Nix, where the schemas are spread throughout the nix store. You can provide the location manually
>
>     env XDG_DATA_DIRS=/nix/store/k5zrf5gz97938mymfsawgfncyyv1caly-gedit-3.22.1/share/gsettings-schemas/gedit-3.22.1
>     gsettings list-schemas
>
> but it is not very convenient. Alternately, you could use a script like this:
>
> ```bash
> #!/usr/bin/env nix-shell
> #!nix-shell -i bash -p glib.dev
> schemas=""
> for p in $NIX_PROFILES; do
>     if [[ -d "$p" ]]; then
>         for d in $(nix-store --query --references "$p"); do
>             schemas_dir=$(echo "$d"/share/gsettings-schemas/*/glib-2.0/schemas)
>             if [[ -d "$schemas_dir" ]]; then
>                 schemas="$schemas${schemas:+:}$schemas_dir"
>             fi
>         done
>     fi
> done
> env XDG_DATA_DIRS=$schemas gsettings "$@"
> ```

[...]

PR #319535, opened 2024-06-13
nautilus: Add NixOS module and pass extensions thru wrapper

jtojnar:
## Description of changes

No more environment variables pointing to global locations. This will allow easy creation of self-contained Nautilus installations with extensions (e.g. for nix-shell or Home manager).

The environment variable will be set by a wrapper configured via `nautilusExtensions` argument of the derivation, that can be changed using `.override` mechanism. This is also what to Thunar does. I decided to not create a separate entrypoint like `gimp-with-plugins` since we Nautilus extension are a more open set and we can just not wrap when the building with the default empty list of extensions.

`nautilus` and is now expected to be managed through `programs.nautilus.enable` NixOS option. Additionally, binary Nautilus extensions will no longer be picked up from `environment.systemPackages`, they need to be installed through `programs.nautilus.extensions`. (This does not apply to `nautilus-python` scripts, which should be installed through as before.)

Also includes some cleanups.

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/nixos/doc/manual/release-notes/rl-2411.section.md
+++ b/nixos/doc/manual/release-notes/rl-2411.section.md
@@ -49,7 +49,7 @@
 
 - `buildbot` was updated to 4.0, the AngularJS frontend has been replaced by a React frontend, see the [upstream release notes](https://docs.buildbot.net/current/manual/upgrading/4.0-upgrade.html).
 
-- `gnome.nautilus` was renamed to `nautilus` and is now expected to be managed through [](#opt-programs.nautilus.enable) NixOS option.
+- `gnome.nautilus` was renamed to `nautilus` and is now expected to be managed through [](#opt-programs.nautilus.enable) NixOS option. Additionally, binary Nautilus extensions will no longer be picked up from [](#opt-environment.systemPackages), they need to be installed through [](#opt-programs.nautilus.extensions). (This does not apply to `nautilus-python` scripts, which should be installed through as before.)
 
 - `nginx` package no longer includes `gd` and `geoip` dependencies. For enabling it, override `nginx` package with the optionals `withImageFilter` and `withGeoIP`.
 
--- a/nixos/doc/manual/release-notes/rl-2411.section.md
+++ b/nixos/doc/manual/release-notes/rl-2411.section.md
@@ -49,6 +49,8 @@
 
 - `buildbot` was updated to 4.0, the AngularJS frontend has been replaced by a React frontend, see the [upstream release notes](https://docs.buildbot.net/current/manual/upgrading/4.0-upgrade.html).
 
+- `gnome.nautilus` was renamed to `nautilus` and is now expected to be managed through [](#opt-programs.nautilus.enable) NixOS option.
+
 - `nginx` package no longer includes `gd` and `geoip` dependencies. For enabling it, override `nginx` package with the optionals `withImageFilter` and `withGeoIP`.
 
 - `openssh` and `openssh_hpn` are now compiled without Kerberos 5 / GSSAPI support in an effort to reduce the attack surface of the components for the majority of users. Users needing this support can
--- a/pkgs/by-name/na/nautilus/package.nix
+++ b/pkgs/by-name/na/nautilus/package.nix
@@ -10,6 +10,8 @@
   gettext,
   desktop-file-utils,
   wrapGAppsHook4,
+  symlinkJoin,
+  makeWrapper,
   gtk4,
   libadwaita,
   libportal-gtk4,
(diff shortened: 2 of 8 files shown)


pkgs/by-name/na/nautilus/wrapper.nix
```diff
@@ -0,0 +1,38 @@
...
+        --replace-fail "${nautilus}" "$out"
+    done
+  '';
+
+  meta = nautilus.meta // {
+    description =
+      nautilus.meta.description
+      +
+        lib.optionalString (nautilusExtensions != [])
+          " (with extensions: ${lib.concatStringsSep ", " (map (x: x.name) nautilusExtensions)})";
```
doronbehar, line 36: Maybe use `x.pname`?

pkgs/by-name/na/nautilus/wrapper.nix
```diff
@@ -0,0 +1,38 @@
...
+symlinkJoin {
+  name = "nautilus-with-extensions-${nautilus.version}";
+
+  paths = [ nautilus ] ++ nautilusExtensions;
+
+  nativeBuildInputs = [ makeWrapper ];
+
+  postBuild = ''
+    wrapProgram "$out/bin/nautilus" \
+      --set "NAUTILUS_4_EXTENSION_DIR" "$out/lib/nautilusx-3"
```
doronbehar, line 18: This seems a bit like a version mismatch..

pkgs/by-name/na/nautilus/package.nix
```diff
@@ -35,105 +37,121 @@
...
+else
+  import ./wrapper.nix {
+    inherit
+      makeWrapper
+      symlinkJoin
+      nautilusExtensions
+      lib
+      ;
+    nautilus = unwrapped;
+  }
```
doronbehar, line 157: Maybe we can put this logic in the service, and expose in `pkgs` both `nautilus-with-extensions` and `nautilus`? This way also the diff will be smaller.

pkgs/by-name/na/nautilus/wrapper.nix
jtojnar: Good catch

```suggestion
      --set "NAUTILUS_4_EXTENSION_DIR" "$out/lib/nautilus/extensions-4"
```

pkgs/by-name/na/nautilus/wrapper.nix
jtojnar:
```suggestion
          " (with extensions: ${lib.concatStringsSep ", " (map (x: x.pname) nautilusExtensions)})";
```

pkgs/by-name/na/nautilus/package.nix
jtojnar: I can imagine having something like `makeExtensionWrapper { varName = "NAUTILUS_4_EXTENSION_DIR"; extensionPath = "lib/…"; package = unwrapped; plugins = nautilusExtensions; }` that would scan the whole derivation but there are several design questions that I did not want to dig into now. We can introduce such helper function later.

I did not want to have two different attributes since that is confusing for some users (based on questions I have seen), see also the commit message.

nixos/doc/manual/release-notes/rl-2411.section.md
```diff
@@ -18,7 +18,7 @@
 
 ## Backward Incompatibilities {#sec-release-24.11-incompatibilities}
 
-- `gnome.nautilus` was renamed to `nautilus` and is now expected to be managed through [](#opt-programs.nautilus.enable) NixOS option.
+- `gnome.nautilus` was renamed to `nautilus` and is now expected to be managed through [](#opt-programs.nautilus.enable) NixOS option. Additionally, binary Nautilus extensions will no longer be picked up from [](#opt-environment.systemPackages), they need to be installed through [](#opt-programs.nautilus.extensions). (This does not apply to `nautilus-python` scripts, which should be installed through as before.)
```
doronbehar, line 21: BTW what does "install as before" mean?

pkgs/by-name/na/nautilus/package.nix
doronbehar: > I can imagine having something like `makeExtensionWrapper { varName = "NAUTILUS_4_EXTENSION_DIR"; extensionPath = "lib/…"; package = unwrapped; plugins = nautilusExtensions; }` that would scan the whole derivation but there are several design questions that I did not want to dig into now. We can introduce such helper function later.
> 
> I did not want to have two different attributes since that is confusing for some users (based on questions I have seen), see also the commit message.

Fair enough. Especially when the attribute eventually evaluates to the unwrapped derivation if there are no extensions enabled.

nixos/doc/manual/release-notes/rl-2411.section.md
jtojnar: Just that no change is necessary.

doronbehar: > Just that no change is necessary.

If there were no instructions before how to install python extensions for nautilus, it is a bit inconsistent to write this...

jtojnar: In the absence of specific documentation, standard methods of installation (e.g. `environment.systemPackages` or user profile) and [upstream instructions](https://gitlab.gnome.org/GNOME/nautilus-python/#running-extensions) apply.

And the backwards incompatibilities section applies to people that already had the software installed anyway.

nixos/doc/manual/release-notes/rl-2411.section.md
```diff
@@ -18,6 +18,8 @@
 
 ## Backward Incompatibilities {#sec-release-24.11-incompatibilities}
 
+- `gnome.nautilus` was renamed to `nautilus` and is now expected to be managed through [](#opt-programs.nautilus.enable) NixOS option. Additionally, binary Nautilus extensions will no longer be picked up from [](#opt-environment.systemPackages), they need to be installed through [](#opt-programs.nautilus.extensions). (This does not apply to `nautilus-python` scripts, which should be installed through as before.)
```
jtojnar, line 21: Need to decide what to do about packages `file-roller` like or `gnome-terminal`. People might be no be aware that the Nautilus features come from those packages and be confused when they disappear.

Though those two have NixOS modules so perhaps we could set `programs.nautilus.extensions` unconditionally. It feels a bit iffy but that is the nature of one package extending another. It also would not be possible to disable just the Nautilus extension but that is not different from status quo.

jtojnar: Or maybe we could adjust the patch to load extensions both from the directory specified in the environment variable and from `/run/current-system/sw/…`. That would avoid the BC break.

pkgs/by-name/na/nautilus/wrapper.nix
```diff
@@ -0,0 +1,38 @@
...
+  makeWrapper,
+  symlinkJoin,
+  nautilus,
+  nautilusExtensions,
+}:
+
+symlinkJoin {
+  name = "nautilus-with-extensions-${nautilus.version}";
+
+  paths = [ nautilus ] ++ nautilusExtensions;
```
jtojnar, line 12: We probably ought to check that all paths contain extension directory. Unfortunately, will have to be done at build time, if we do not want to go the way of adding `passthru.nautilusExtension` to every extension package and check that.

nixos/doc/manual/release-notes/rl-2411.section.md
doronbehar: I'm a bit confused, on the one hand, there is the idea of `symlinkJoin` which puts many extensions together, and on the other hand there is already `NAUTILUS_4_EXTENSION_DIR` which works already. Why do we need `symlinkJoin` to begin with?

nixos/doc/manual/release-notes/rl-2411.section.md
doronbehar: It is as if you are saying that all of the documentation that I added in https://github.com/NixOS/nixpkgs/pull/240780 is trivially inferred from reading the upstream instructions, and or simply using `environment.systemPackages`.. Did I understand you correctly?

jtojnar: Except for the `NAUTILUS_4_EXTENSIONS_DIR` and the GSettings behaviour, yes.

- The environment variable, I do not want to expose.
- The GSettings behaviour is not extension specific, and the whole section needs rewrite anyway.
  - We need to mention dconf NixOS module
  - Ideally remove GSettings overrides and/or implement GSettings module based on the dconf one
  - Mention that we do not install GSettings globally as you mention (I have to keep linking people to https://github.com/NixOS/nixpkgs/issues/33277)

nixos/doc/manual/release-notes/rl-2411.section.md
jtojnar: `NAUTILUS_4_EXTENSION_DIR` requires a single path. Previously, the `symlinkJoin` was just done by `environment.systemPackages`.

doronbehar: > `NAUTILUS_4_EXTENSION_DIR` requires a single path.

Still, you can revert the `symlinkJoin` change, and use the `programs.nautilus.extensions` to simply put packages' paths in `$NAUTILUS_4_EXTENSION_DIR`.

jtojnar: Each package will be in a different store path so we cannot put them all into the variable that expects a single path.

doronbehar: > Each package will be in a different store path so we cannot put them all into the variable that expects a single path.

Yes but what still, in the module level, you can symlinkJoin the extensions to a single path and use that as the value for `$NAUTILUS_4_EXTENSION_DIR`.

jtojnar: Well, one of the points is that we want to avoid relying on NixOS module, so that Nautilus with extensions can be used e.g. in nix-shell or h-m.

doronbehar:
@jtojnar I think I finally fully understood your intentions in your last 2 comments<sup>[1,2]</sup>, so I'll reply to both of them here, as my reply is general about this PR and not specific to the details commented upon in the linked review threads.

So basically it seems to me we have a general issue about some functionality in Nixpkgs not available when decoupled from the NixOS modules system - one example is the [mentioned `gsettings` issue](https://github.com/NixOS/nixpkgs/issues/33277) and also the environment variable `NAUTILUS_4_EXTENSION_DIR` which is currently loaded with files only by the NixOS module option `services.gnome.core-utilities.enable`.

Indeed this is a bit annoying, and it would have been wonderful if one could launch nautilus even on a plasma desktop with:

```sh
nix run --impure \
  --expr 'with (builtins.getFlake "nixpkgs").legacyPackages.${builtins.currentSystem};
    nautilus.withExtensions [
      nautilus-python
      kdeconnect
      wezterm
      nautilus-open-any-terminal
    ]
  '
```

However I'm not sure we'll ever be able to reach such a utopia in the near future. I also doubt whether the developers of Gnome and nautilus will ever support such usage on an FHS-compliant system without Gnome fully and properly installed.

Due to that, I strongly believe it won't be bad at all if all of the nautilus extensions system will only be supported by the `gnome` or `programs.nautilus` NixOS module - hence I don't see any value in the effort to create a `nautilus-with-extensions` derivation.

Not only that, this PR isn't worth anything IMO if the extensions' gsettings are not handled by the `programs.nautilus.extensions` module or whatever module you may edit in this PR, as this is the most non trivial thing I wanted to document in https://github.com/NixOS/nixpkgs/pull/240780 .

Your knowledge @jtojnar on the inner details of the gnome ecosystem is impressive and deeply valuable to Nixpkgs, but it is frustrating how your aspiration to reach a perfect interface with it, that will cover all possible use-cases, eventually smears out more modest contributions such as documentation efforts like https://github.com/NixOS/nixpkgs/pull/240780 , and hence blocks other users from at least getting to know too, just a little bit the inner details of the `services.gnome` module options.

------

1. https://github.com/NixOS/nixpkgs/pull/319535#discussion_r1639804595
2. https://github.com/NixOS/nixpkgs/pull/319535#discussion_r1639468000

pkgs/by-name/na/nautilus/wrapper.nix
```diff
@@ -0,0 +1,38 @@
...
+
+  nativeBuildInputs = [ makeWrapper ];
+
+  postBuild = ''
+    wrapProgram "$out/bin/nautilus" \
+      --set "NAUTILUS_4_EXTENSION_DIR" "$out/lib/nautilus/extensions-4"
```
jtojnar, line 18: We might want to also wrap `XDG_DATA_DIRS` so that the wrapper can install `nautilus-python` scripts. Not sure if we should re-use the `nautilusExtensions` argument or create a new one, though.

JohnRTitor:
I'd note that there is also `nautilus-open-any-terminal` [NixOS module](https://github.com/NixOS/nixpkgs/blob/nixos-unstable/nixos/modules/programs/nautilus-open-any-terminal.nix), which configures a lot of what this module trying to do.

repparw:
Tested `programs.nautilus.enable` from this branch on niri — wrapper + dbus wiring works great.

Outside GNOME `/share/thumbnailers` is empty, so PDF/image previews stay broken. `gnome.nix` only pulls `glycin/gst` when GNOME is enabled, and PDF needs `papers`/`evince`.

Would you consider a follow-up `programs.nautilus.thumbnailers.enable` (default true) pulling `glycin-thumbnailer` + `gst-thumbnailers` + `papers`? Keeps this PR minimal, makes standalone niri usable.

*Assisted-by: Muse Spark 1.2 via OpenCode*

PR #319539, opened 2024-06-13
obsidian: add commandLineArgs

CrackTC:
## Description of changes
add `commandLineArgs` so that we can override some electron flags more easily 

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/applications/misc/obsidian/default.nix
+++ b/pkgs/applications/misc/obsidian/default.nix
@@ -8,6 +8,7 @@
 , writeScript
 , undmg
 , unzip
+, commandLineArgs ? ""
 }:
 let
   inherit (stdenv.hostPlatform) system;
@@ -52,7 +53,8 @@ let
       mkdir -p $out/bin
       makeWrapper ${electron}/bin/electron $out/bin/obsidian \
         --add-flags $out/share/obsidian/app.asar \
-        --add-flags "\''${NIXOS_OZONE_WL:+\''${WAYLAND_DISPLAY:+--ozone-platform=wayland}}"
+        --add-flags "\''${NIXOS_OZONE_WL:+\''${WAYLAND_DISPLAY:+--ozone-platform=wayland}}" \
+        --add-flags ${lib.escapeShellArg commandLineArgs}
       install -m 444 -D resources/app.asar $out/share/obsidian/app.asar
       install -m 444 -D resources/obsidian.asar $out/share/obsidian/obsidian.asar
       install -m 444 -D "${desktopItem}/share/applications/"* \


w-lfchen (approved):
other applications [seem to be doing this as well](https://github.com/search?q=repo%3ANixOS%2Fnixpkgs%20commandLineArgs&type=code), therefore i don't see any issues with adding this.
i'm unsure how useful this will be but i might be missing something here; i'd love to hear somebody else's opinion on it

kashw2 (approved):

zaninime:
If the flags are electron-related, why not adding them to electron directly? An overlay that adds them once to electron would then solve the problem for all apps using it. @CrackTC can you provide an example of what you're trying to achieve?

CrackTC:
> If the flags are electron-related, why not adding them to electron directly?

Mainly because the flags passed to electron might vary from application to application. In my use case, some apps work on wayland by passing flags like `--enable-features=UseOzonePlatform` and `--ozone-platform=wayland`. Some apps may need `--disable-gpu` to work properly, while some apps won't work on wayland at all. I'm not sure if this is the best way to do it, but imo adding an argument to specific apps is more straightforward.

zaninime:
I'm not following too much how electron is behaving on different platforms nowadays. I find it strange that the same electron binary needs different GPU-configurations based on the (JS) app it is running. Also while this is a straightforward change to make, reverting it in the future will be a breaking one for whoever relies on it.

With that aside, did you realize the line above your change already sets `--ozone-platform=wayland` under some conditions? I would find it confusing if I'd be making an override such as:

```nix
electron.override {
  commandLineArgs = ["a" "b" "c"];
}
```

and suddently my `ps aux | grep electron` shows that also `--ozone-platform=wayland` is set. For Nix beginners this might be quite unexpected behaviour.

CrackTC:
> I find it odd that the same electron binary needs different GPU configurations depending on the (JS) application it is running.

I noticed that some of them have a bundled electron binary, some of them use the one in nixpkgs. And you're right, obsidian is the latter, for which we can simply add flags to the shared electron binary. But wayland support is just one of the things I'm trying to achieve. There are other things I have come across that vary between application, such as scale factor and proxy related flags.

> while this is an easy change to make, undoing it in the future will be a break for anyone relying on it.

Sorry I hadn't thought of this. I'm not quite sure in which case we will undo it in the future, if electron changes certain flags we can just update the value of `commandLineArgs`. And there are applications that already use this argument, so I think it's safe to assume that they will continue to use it.

> For Nix newbies, this may be quite unexpected behaviour.

The nixos wiki mentions that we can add ozone flags to electron apps globally via `environment.sessionVariables.NIXOS_OZONE_WL = "1";`, so if someone has added this variable, they should know what they're doing. Also, in the case some apps are broken because of the presence of `NIXOS_OZONE_WL` (probably those bundled with outdated electron), we can easily opt out with the help of `commandLineArgs`.

JohnRTitor (approved):
LGTM, runs fine

JohnRTitor:
Result of `nixpkgs-review pr 319539` run on x86_64-linux [1](https://github.com/Mic92/nixpkgs-review)
<details>
  <ul>
    <li>obsidian</li>
  </ul>
</details>

PR #319634, opened 2024-06-13
ananicy-cpp: support wrapped applications

diniamo:
## Description of changes

Ananicy-cpp now matches wrapped applications with this patch.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/by-name/an/ananicy-cpp/match-wrappers.patch
+++ b/pkgs/by-name/an/ananicy-cpp/match-wrappers.patch
@@ -1,0 +1,16 @@
+diff --git a/src/worker.cpp b/src/worker.cpp
+index 0cf8955..b9dc70f 100644
+--- a/src/worker.cpp
++++ b/src/worker.cpp
+@@ -29,7 +29,10 @@ void Worker::work(const std::stop_token &stop_token) {
+   while (!stop_token.stop_requested()) {
+     while ((proc = process_queue->poll(500ms)).has_value()) {
+       const auto &p = proc.value();
+-      const auto &rule = rules->get_rule(p.name);
++      auto name = p.name;
++      if (name.starts_with('.') && name.ends_with("-wrapped"))
++        name = name.substr(1, name.find_last_of('-') - 1);
++      const auto &rule = rules->get_rule(name);
+ 
+       processed_count++;
+ 
--- a/pkgs/by-name/an/ananicy-cpp/package.nix
+++ b/pkgs/by-name/an/ananicy-cpp/package.nix
@@ -33,6 +33,7 @@ clangStdenv.mkDerivation rec {
       url = "https://gitlab.com/ananicy-cpp/ananicy-cpp/-/commit/6ea2dccceec39b6c4913f617dad81d859aa20f24.patch";
       hash = "sha256-C+7x/VpVwewXEPwibi7GxGfjuhDkhcjTyGbZHlYL2Bs=";
     })
+    ./match-wrappers.patch
   ];
 
   strictDeps = true;
--- a/pkgs/by-name/an/ananicy-cpp/package.nix
+++ b/pkgs/by-name/an/ananicy-cpp/package.nix
@@ -80,6 +80,7 @@ clangStdenv.mkDerivation rec {
     maintainers = with lib.maintainers; [
       artturin
       johnrtitor
+      diniamo
     ];
     mainProgram = "ananicy-cpp";
   };


diniamo:
I can't find how to request reviews so: @JohnRTitor @Artturin

pkgs/by-name/an/ananicy-cpp/match-wrappers.patch
```diff
@@ -0,0 +1,24 @@
...
+ 
+ #include <chrono>
+@@ -29,7 +30,10 @@ void Worker::work(const std::stop_token &stop_token) {
+   while (!stop_token.stop_requested()) {
+     while ((proc = process_queue->poll(500ms)).has_value()) {
+       const auto &p = proc.value();
+-      const auto &rule = rules->get_rule(p.name);
++      auto name = p.name;
++      if (utils::regex::is_str_matching_regex(name, "^\\..+-wrap(p(ed?)?)?$"))
++        name = name.substr(1, name.find_last_of('-') - 1);
```
Artturin, line 20: So it matches -wrapped and then removes the -wrapped right?

What's up with the `(p(ed?)?)`?

pkgs/by-name/an/ananicy-cpp/package.nix
```diff
@@ -33,7 +34,8 @@ clangStdenv.mkDerivation rec {
       url = "https://gitlab.com/ananicy-cpp/ananicy-cpp/-/commit/6ea2dccceec39b6c4913f617dad81d859aa20f24.patch";
       hash = "sha256-C+7x/VpVwewXEPwibi7GxGfjuhDkhcjTyGbZHlYL2Bs=";
     })
-  ];
+  ]
+  ++ lib.optional wrapperSupport ./match-wrappers.patch;
```
Artturin, line 38: Let's just always add it

pkgs/by-name/an/ananicy-cpp/match-wrappers.patch
JohnRTitor: AFAIK, sometimes process names don't fully have "-wrapped" suffix in their name, due to size limitations.

pkgs/by-name/an/ananicy-cpp/package.nix
JohnRTitor: Agreed, is this on upstream patch, by the way? If not, I think it should go to upstream directly.

pkgs/by-name/an/ananicy-cpp/match-wrappers.patch
diniamo: It matches
```
.x-wrap
.x-wrapp
.x-wrappe
.x-wrapped
```
and removes everything after the last `-` (the `-` included).

pkgs/by-name/an/ananicy-cpp/package.nix
diniamo: No, I wrote it.

Artturin: > Agreed, is this an upstream patch, by the way? If not, I think it should go to upstream directly.

It's nix specific so idk if upstream would accept it, though we can try.

JohnRTitor: Not all process names end like that, as you can see. 

https://github.com/CachyOS/ananicy-rules/blob/e3638ebb2320202ede9facfc9657cc8af2098a2d/00-default/Services/gvfs.rules#L10

I had to manually add these rules to match `.x-` `.x-w`, `.x-wra` and so on.

All wrapped app's processes have `.` in the beginning. 
I think the rules should run on any process `x` and `.x+.`

pkgs/by-name/an/ananicy-cpp/match-wrappers.patch
JohnRTitor: Not all process names end like that, as you can see. 

https://github.com/CachyOS/ananicy-rules/blob/e3638ebb2320202ede9facfc9657cc8af2098a2d/00-default/Services/gvfs.rules

image


I had to manually add these rules to match `.x-` `.x-w`, `.x-wra` and so on.

All wrapped app's processes have `.` in the beginning. 
I think the rules should run on any process `x` and `\.x.*` (regex)

diniamo: @JohnRTitor are you sure that screenshots includes the full process names?

diniamo: Looks like they do, in which case I'll try to come up with a better way of stripping the extra characters. I'm not sure how yet, so ideas are welcome.

diniamo: Never mind, it looks like `ps` is doing something funky. If you check the [nixpkgs makeWrapper implementation](https://github.com/NixOS/nixpkgs/blob/683aa7c4e385509ca651d49eeb35e58c7a1baad6/pkgs/build-support/setup-hooks/make-wrapper.sh#L219), you can see that wrapers always start with a `.` and end with `-wrapped`. You can also confirm this by check the store paths for programs, eg.: `ls -a $(dirname $(realpath $(which firefox)))`. And fortunately the process names in ananicy-cpp are also correct, why I confirmed by logging:

tmp ElXKWq8NkR

(The right side is `ps`, the left side is a log line from ananicy-cpp)

diniamo:
@JohnRTitor @Artturin could we finish this up?

VeilSilence:
Is there anything need to do to merge this?

JohnRTitor:
> @JohnRTitor @Artturin could we finish this up?

Looks good, let me do a quick check before I approve.

pkgs/by-name/an/ananicy-cpp/match-wrappers.patch
```diff
@@ -0,0 +1,16 @@
...
++++ b/src/worker.cpp
+@@ -29,7 +29,10 @@ void Worker::work(const std::stop_token &stop_token) {
+   while (!stop_token.stop_requested()) {
+     while ((proc = process_queue->poll(500ms)).has_value()) {
+       const auto &p = proc.value();
+-      const auto &rule = rules->get_rule(p.name);
++      auto name = p.name;
++      if (name.starts_with('.') && name.ends_with("-wrapped"))
++        name = name.substr(1, name.find_last_of('-') - 1);
++      const auto &rule = rules->get_rule(name);
```
Artturin, line 13: Does the `.` have to be removed?

JohnRTitor:
By the way, is there a reason we are not using `ananicy-cpp` as the default in `services.ananicy`?

JohnRTitor (approved):
LGTM based on my testing, approved

pkgs/by-name/an/ananicy-cpp/match-wrappers.patch
diniamo: Yes, it does. Both the `.` and the `-wrapped` are appended by the wrapper script, so we want the part between them for matching.

Artturin: I meant is the . being removed, I guess the (1 is doing that? Not familiar with cpp

diniamo:
@JohnRTitor can you make a pr to remove all the now reduant rules in the cachyos rules repo?

JohnRTitor:
I think those can stay, for those using just `ananicy` (non cpp) and who are not on the version with the patch. 

Future rules to the repo, won't need to account for wrapped processes.

Pandapip1:
FYI, having not seen the discussion on this PR, I opened https://github.com/CachyOS/ananicy-rules/pull/174 and it was merged

diniamo:
I'm still fine with that, and at this point, John is probably as well. This patch is in the latest stable release.

PR #319711, opened 2024-06-14
python3Packages.torch: general cleanups and fixups

### LINKED ISSUE #294111: Pytorch: pytorch uses unsupported versions of cuDNN
Pytorch only supports very specific versions of `cudatoolkit` and `cuDNN` (cf: [currently-supported-cuda-and-cudnn-libraries](https://github.com/pytorch/builder/blob/main/CUDA_UPGRADE_GUIDE.MD#a-currently-supported-cuda-and-cudnn-libraries)).

If you build `cudatoolkit 11.8` on an `x86_64`, you will use `cuDNN 8.9.7.29`. Pytorch only supports `cuDNN 8.7.0.24` for `cudatoolkit 11.8`.

Nixpkgs uses the [nvidia support matrix](https://docs.nvidia.com/deeplearning/cudnn/archives/index.html) to build the latest supported `cuDNN` for a given version of `cudatoolkit` (the matrix is encoded in [cudnn/releases.nix](https://github.com/NixOS/nixpkgs/blob/8421b4c5339ce0950b5904f9660998e29fd32544/pkgs/development/cuda-modules/cudnn/releases.nix)). This does not match what is expected by pytorch. We've experienced pytorch crashes that were resolved by pegging cuDNN to the version supported by pytorch.

ConnorBaker:
## Description of changes

- Moved things which had some sort of selection logic to the caller, which now has a recursive attribute set, allowing the selection logic to change in response to overrides.
- Removed constraint that Magma and MPI use the same version of CUDA as Torch, as they shouldn't interfere with each other and are loaded in different processes.
- Fixes a warning thrown when `allowAliases` is true caused by a `with cudaPackages;` shadowing the copy of `autoAddDriverRunpath` provided by `callPackage` (cc @mweinelt).
- Construction of a list of CUDA build inputs (creatively named `cudaBuildInputs`).
- Cleaned up and moved shell variables to `env` so they are provided as environment variables to the builder, since PyTorch's `setup.py` expects those (also ensures we catch variables which aren't set to strings).
- Fixed deprecation warnings caused by usage of `substituteInPlace`'s `--replace` argument.
- Updated the list of supported CUDA capabilities to include Thor (that's Hopper with additional features, `9.0a`).
- Closes #294111.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [x] Tested, as applicable:

CC @NixOS/cuda-maintainers @NixOS/rocm-maintainers 

---


ConnorBaker:
I'm able to build and run [`nix-cuda-test`](https://github.com/connorbaker/nix-cuda-test) on my 4090:

```console
$ nix run -L --override-input nixpkgs "github:NixOS/nixpkgs/f8b21116e19a7787194630d38ea3c39d7ffc9625" .#nix-cuda-test
...
Seed set to 42
Using bfloat16 Automatic Mixed Precision (AMP)
GPU available: True (cuda), used: True
TPU available: False, using: 0 TPU cores
IPU available: False, using: 0 IPUs
HPU available: False, using: 0 HPUs
Files already downloaded and verified
Files already downloaded and verified
LOCAL_RANK: 0 - CUDA_VISIBLE_DEVICES: [0]

  | Name      | Type             | Params
-----------------------------------------------
0 | criterion | CrossEntropyLoss | 0     
1 | model     | ViT              | 86.3 M
-----------------------------------------------
86.3 M    Trainable params
0         Non-trainable params
86.3 M    Total params
345.317   Total estimated model params size (MB)
Epoch 0:  58%|████████████████████████████████████████████████████████████████████████████▏                                                       | 451/781 [00:52<00:38,  8.64it/s, v_num=7, train_loss=2.310]
```

pkgs/development/python-modules/torch/default.nix
```diff
@@ -117,8 +189,9 @@ let
         "8.7"
         "8.9"
         "9.0"
+        "9.0a"
       ];
-      ptx = lists.map (x: "${x}+PTX") real;
+      ptx = map (x: "${x}+PTX") real;
```
SomeoneSerge, line 194: Since we're reviewing the whole expression, this probably should be conditional on `cudaEnableForwardCompat`

pkgs/development/python-modules/torch/default.nix
```diff
@@ -7,17 +7,9 @@
...
   cudaPackages,
+  cudnn ? null,
   autoAddDriverRunpath,
-  effectiveMagma ?
-    if cudaSupport then
-      magma-cuda-static
-    else if rocmSupport then
-      magma-hip
-    else
-      magma,
```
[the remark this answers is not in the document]
ConnorBaker: I wanted to limit the override surface of the package and make it easier to override:

- by having a single argument for a package, it's unambiguous which will be used; the alternative is to know to override either `effectiveMagma` or whichever argument `effectiveMagma` selects
- moving the selection logic to the caller gives us more room to express it (since we don't need to put it all inside an expression to the right of a `?`)
- `callPackage` pulls packages from the scope it was constructed with in a strict fashion, so missing packages cause evaluation errors (even if they're unused)
- overrides of `torch` (e.g., `torchWithCuda` or `torchWithoutRocm`) are in `python-packages.nix`, so it made sense to me that the selection logic should belong in the same file (since we use package selection to ensure packages are built with proper versions of their dependencies)

I had thought about trying to do the same in the function argument attribute set in the expression itself, but even if `pkgs` was in scope for me to do the selection argument, I'd still need to use `effective*` values in the attribute set for each of the packages with selection logic. I say that because I'm not aware of a way to tell callPackages not to provide a value for some value: within the expression, I'd have no way of knowing whether arguments like `cudaPackages` were supplied by `callPackage` from `pkgs`, or if they're an override, and therefore no way of knowing whether to set `effectiveCudaPackages` to the value of `cudaPackages` or ignore it and use the package selection logic to get what should be the default value.

I'm also hesitant to use versioned package names; `torch` can build with CUDA 11.8 or 12.1, but it feels odd to require both exist by adding them as arguments to the attribute set fed to `callPackage`. I don't like the idea of picking just one of them either, as `torch.override { cudaPackages_11_8 = cudaPackages_12_1; }` feels bad to me.

What are your thoughts?

pkgs/development/python-modules/torch/default.nix
ConnorBaker: Good catch!

ConnorBaker: Resolved in https://github.com/NixOS/nixpkgs/pull/319711/commits/269ed33e57ec379e822890ad1f9e879337045443

ConnorBaker:
Because PyTorch is still using FindCUDA, they need to vendor fixes like https://github.com/pytorch/pytorch/commit/7d10b06e1a7ff87eb38d39e01eb717bd1001714e

Gotta find out how to get theirs working with our CMake 🙃

pkgs/development/python-modules/torch/default.nix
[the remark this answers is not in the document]
SomeoneSerge: - Eval errors are prevented by supplying default values
- The point of using `effective*` names is that we know for sure these don't exist in the package sets, so they're either explicitly supplied by the user or computed by us in the default value expression. (This is in response e.g. to "callPackages not to provide a value for some value")
- IIRC the call-site override approach can break with overlays (maybe this is what you refer to with the torch's magma selection logic)

But yes, it's annoying that `callPackage` actually serves multiple different purposes here, and also that it doesn't allow distinguishing between parent scopes at different depths... (to concur with "I'm also hesitant to use versioned package names; torch can build with CUDA 11.8 or 12.1")

pkgs/development/python-modules/torch/default.nix
```diff
@@ -80,24 +73,104 @@
...
+    cudaAtLeast
+    cudaFlags
+    cudaMajorMinorVersion
+    cudaOlder
+    libcublas
+    libcufft
+    libcurand
+    libcusolver
+    libcusparse
+    nccl
```
SomeoneSerge, line 127: Honestly, I think in this case `with;` and `cudaPackages.${name}` are more reasonable

pkgs/development/python-modules/torch/default.nix
```diff
@@ -80,24 +73,104 @@
...
 
-  rocmPackages = rocmPackages_5;
+  cudaBuildInputs =
+    let
+      # NOTE: Reminder that due to the way cudaPackages are created, lib.getOutput and friends don't work.
+      # As such, we need to create our own functions to get the dev and lib outputs.
+      getDevAndLib = pkg: [
+        pkg.dev
+        pkg.lib
+      ];
```
SomeoneSerge, line 137: Usually `dev` propagates `lib`

PR #319825, opened 2024-06-14
nekoray: macos support

teidesu:
## Description of changes

added support for macos in nekoray package

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [x] aarch64-darwin
  - [x] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/by-name/ne/nekoray/package.nix
+++ b/pkgs/by-name/ne/nekoray/package.nix
@@ -1,6 +1,7 @@
 {
   lib,
   stdenv,
+  overrideSDK,
   fetchFromGitHub,
   libsForQt5,
   cmake,
@@ -69,11 +70,13 @@ let
 
   installGeodata = lib.concatStringsSep "\n" (
     lib.mapAttrsToList (filename: file: ''
-      install -Dm644 ${file} "$out/share/nekoray/${filename}"
+      install -Dm644 ${file} "$install_dir/${filename}"
     '') geodata
   );
+
+  stdenv' = if stdenv.isDarwin then overrideSDK stdenv "11.0" else stdenv;
 in
-stdenv.mkDerivation (finalAttrs: {
+stdenv'.mkDerivation (finalAttrs: {
   pname = "nekoray";
   version = "3.26";
 
@@ -90,17 +93,19 @@ stdenv.mkDerivation (finalAttrs: {
     libsForQt5.wrapQtAppsHook
     cmake
     ninja
+  ] ++ lib.optionals stdenv.isLinux [
     copyDesktopItems
   ];
 
   buildInputs = [
     libsForQt5.qtbase
     libsForQt5.qttools
-    libsForQt5.qtwayland
-    libsForQt5.qtx11extras
     protobuf
     yaml-cpp
     zxing-cpp
+  ] ++ lib.optionals stdenv.isLinux [
+    libsForQt5.qtwayland
+    libsForQt5.qtx11extras
   ];
 
   # NKR_PACKAGE makes sure the app uses the user's config directory to store it's non-static content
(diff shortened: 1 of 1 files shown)


teidesu:
x86_64-darwin build seems to be borked? would be cool if someone tested, i'd then mark it as broken (i don't have an intel mac to test locally nor fix)

TomaSajt:
> x86_64-darwin build seems to be borked? would be cool if someone tested, i'd then mark it as broken (i don't have an intel mac to test locally nor fix)

We probably need force it to use the 11.0 apple SDK, (which is the default on aarch64) (x86_64 has version 10.3 by default)

pkgs/by-name/ne/nekoray/package.nix
```diff
@@ -110,22 +111,30 @@ stdenv.mkDerivation (finalAttrs: {
...
+
+      install -Dm755 nekoray "$out/share/nekoray/nekoray"
+      mkdir -p "$out/bin"
+      ln -s "$out/share/nekoray/nekoray" "$out/bin"
+      install -Dm644 "$src/res/public/nekoray.png" "$out/share/icons/hicolor/256x256/apps/nekoray.png"
+    '' else ''
+      install_dir="$out/Applications/nekoray.app/Contents/MacOS"
+
+      mkdir -p $out/Applications
+      cp -r nekoray.app $out/Applications
```
TomaSajt, line 125: Maybe we could also symlink it back into $out/bin, since `nix run` will try to run the executable from there.

*(If the symlinking breaks it, try to use `makeWrapper` instead)*

teidesu: regarding this - i'm not entirely sure this is a good idea
most gui applications don't need to be in path, and if i symlink it to `bin/` it will inevitably end up there :c
it's somewhat of a necessary evil on linux, but on macos we have `/Applications`

to support `nix run`-ning maybe it's a better idea to do something like this?
```diff
@@ -121,7 +124,7 @@ stdenv.mkDerivation (finalAttrs: {
     '' else ''
       install_dir="$out/Applications/nekoray.app/Contents/MacOS"
 
-      mkdir -p $out/Applications
+      mkdir -p $out/Applications $out/bin
       cp -r nekoray.app $out/Applications
     ''}
 
@@ -164,7 +167,7 @@ stdenv.mkDerivation (finalAttrs: {
     description = "Qt based cross-platform GUI proxy configuration manager";
     homepage = "https://github.com/MatsuriDayo/nekoray";
     license = lib.licenses.gpl3Plus;
-    mainProgram = "nekoray";
+    mainProgram = if stdenv.isDarwin then "../Applications/nekoray.app/Contents/MacOS/nekoray" else "nekoray";
     maintainers = with lib.maintainers; [ tomasajt ];
   };
 })
```

though it definitely looks a bit hacky 🥴

teidesu:
> This is a bit hack-y

image

other packages do this too so i think it's fine :D

pkgs/by-name/ne/nekoray/package.nix
```diff
@@ -110,22 +111,30 @@ stdenv.mkDerivation (finalAttrs: {
...
 
     ${installGeodata}
 
-    install -Dm644 "$src/res/public/nekoray.png" "$out/share/icons/hicolor/256x256/apps/nekoray.png"
-
     runHook postInstall
   '';
 
-  desktopItems = [
+  desktopItems = lib.optionals stdenv.isLinux [
```
Aleksanaa, line 137: It's `copyDesktopItems` in `nativeBuildInputs` which should be conditional, not this.

pkgs/by-name/ne/nekoray/package.nix
```diff
@@ -157,6 +166,5 @@ stdenv.mkDerivation (finalAttrs: {
     license = lib.licenses.gpl3Plus;
     mainProgram = "nekoray";
     maintainers = with lib.maintainers; [ tomasajt ];
-    platforms = lib.platforms.linux;
```
TomaSajt, line 160: I think instead of being implicit about the platforms, we should do lib.platforms.linux ++ lib.platforms.darwin

pkgs/by-name/ne/nekoray/package.nix
```diff
@@ -69,11 +70,13 @@ let
 
   installGeodata = lib.concatStringsSep "\n" (
     lib.mapAttrsToList (filename: file: ''
-      install -Dm644 ${file} "$out/share/nekoray/${filename}"
+      install -Dm644 ${file} "$install_dir/${filename}"
     '') geodata
   );
+
+  stdenv' = if stdenv.isDarwin then overrideSDK stdenv "11.0" else stdenv;
```
FliegendeWurst, line 77: This should no longer be needed, I think.

PR #319860, opened 2024-06-14
maintainers/scripts/update: support skipping prompts

superherointj:
maintainers/scripts/update.py: support skipping prompts

Useful for not having to press ENTER for continuing execution.

To test:

> nix-shell maintainers/scripts/update.nix --argstr package bicep --argstr skip-prompt true

Found a new solution that works:

> yes '' | nix-shell maintainers/scripts/update.nix --argstr package bicep

This clever solution might not be obvious to everyone. The argument is obvious.

### DIFF
--- a/maintainers/scripts/update.nix
+++ b/maintainers/scripts/update.nix
@@ -13,6 +13,7 @@
 , include-overlays ? false
 , keep-going ? null
 , commit ? null
+, skip-prompt ? null
 }:
 
 let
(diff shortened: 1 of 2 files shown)


ulrikstrid (approved):
This is not an area of the repo I'm super familiar with but the change looks reasonable

AndersonTorres (approved):

PR #313408, opened 2024-05-21
doc: autogenerate python interpreter table

alejandrosame:
This serves as a practical example on generating documentation by inspection of the evaluated nixpkgs tree.

PR #319875, opened 2024-06-14
Minor Nixpkgs doc improvements

infinisil:
## Description of changes
- Insert the python interpreter table at build-time instead of eval-time (originally from https://github.com/NixOS/nixpkgs/pull/313408)
- Remove an unnecessary dir that doesn't have any files in it anymore for years
- Prevent unnecessary rebuilds (e.g. when only Nix files change), see https://github.com/NixOS/nixpkgs/issues/301014

## Things done
- [x] Built the manual successfully using `nix-build doc`
- [x] Checked that changing the Nix files doesn't cause a rebuild

---

### DIFF
--- a/doc/default.nix
+++ b/doc/default.nix
@@ -2,6 +2,7 @@
 let
   inherit (pkgs) lib;
   inherit (lib) hasPrefix removePrefix;
+  fs = lib.fileset;
 
   common = import ./common.nix;
 
@@ -99,7 +100,19 @@ in pkgs.stdenv.mkDerivation {
     nixos-render-docs
   ];
 
-  src = ./.;
+  src = fs.toSource {
+    root = ./.;
+    fileset = fs.unions [
+      (fs.fileFilter (file:
+        file.hasExt "md"
+        || file.hasExt "md.in"
+      ) ./.)
+      ./style.css
+      ./anchor-use.js
+      ./anchor.min.js
+      ./manpage-urls.json
+    ];
+  };
 
   postPatch = ''
     ln -s ${optionsDoc.optionsJSON}/share/doc/nixos/options.json ./config-options.json
--- a/doc/default.nix
+++ b/doc/default.nix
@@ -105,14 +105,12 @@ in pkgs.stdenv.mkDerivation {
     ln -s ${optionsDoc.optionsJSON}/share/doc/nixos/options.json ./config-options.json
   '';
 
-  buildPhase = let
-    pythonInterpreterTable = pkgs.callPackage ./doc-support/python-interpreter-table.nix {};
-    pythonSection = with lib.strings; replaceStrings
-      [ "@python-interpreter-table@" ]
-      [ pythonInterpreterTable ]
-      (readFile ./languages-frameworks/python.section.md);
-  in ''
-    cp ${builtins.toFile "python.section.md" pythonSection} ./languages-frameworks/python.section.md
+  pythonInterpreterTable = pkgs.callPackage ./doc-support/python-interpreter-table.nix {};
+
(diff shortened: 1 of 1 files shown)


doc/default.nix
```diff
@@ -2,6 +2,7 @@
 let
   inherit (pkgs) lib;
   inherit (lib) hasPrefix removePrefix;
+  fs = lib.fileset;
```
philiptaron, line 5: @infinisil, I've seen this `fs = lib.fileset` idiom a couple times in your patches now -- interested in why you're doing it. `fs.unions`, `fs.fileFilter` don't look like better names to me than `fileset.unions` or `fileset.fileFilter`. I'm not strongly opposed or have something else to suggest, just that the choice doesn't yet make me go "heck yes."

djacu (approved):
<3 filesets

doc/default.nix
```diff
@@ -99,20 +100,30 @@ in pkgs.stdenv.mkDerivation {
...
+    fileset = fs.unions [
+      (fs.fileFilter (file:
+        file.hasExt "md"
+        || file.hasExt "md.in"
+      ) ./.)
+      ./style.css
+      ./anchor-use.js
+      ./anchor.min.js
+      ./manpage-urls.json
+    ];
```
philiptaron, line 114: I was interested in the difference, so I built the derivation on master and on this PR:

```
$ $ diff -u4 -r /nix/store/w483x0vv260dikx0wzl49kf6zkd5i673-doc /nix/store/j1nq1anc8hc6gai6h7ib7rv9jjcs9aq5-source
Only in /nix/store/w483x0vv260dikx0wzl49kf6zkd5i673-doc: common.nix
Only in /nix/store/w483x0vv260dikx0wzl49kf6zkd5i673-doc: default.nix
Only in /nix/store/w483x0vv260dikx0wzl49kf6zkd5i673-doc: doc-support
Only in /nix/store/w483x0vv260dikx0wzl49kf6zkd5i673-doc/functions: library
Only in /nix/store/w483x0vv260dikx0wzl49kf6zkd5i673-doc: old
Only in /nix/store/w483x0vv260dikx0wzl49kf6zkd5i673-doc: shell.nix
Only in /nix/store/w483x0vv260dikx0wzl49kf6zkd5i673-doc: tests
```

This does make me think that this would be less error-prone to _exclude_ the following file types:

1. `.nix`
2. `.py`
3. `.txt` (but also: consider removing `doc/old/cross.txt`.)

doc/default.nix
infinisil: Yeah that's fair, I think it's just smaller and I'm kind of used to it by now. It's like a qualified import (`use foo::bar as fb` in Rust, or `import qualified Foo.Bar as FB` in Haskell)

doc/default.nix
infinisil: I think it's cleaner to explicitly list what's needed, so that it's directly associated with the build instructions further down. Also this way we don't need to worry about new file types accidentally influencing the build.

doc/default.nix
infinisil: I'd say let's keep it this way though, it's not a problem (unlike the `with` pattern!)

doc/default.nix
philiptaron: Right, my thought is that it's expected in docs to have css, js, and json. Sort of the converse, where there's unexpected breakage due to needing to allowlist one sort of file (css, js) but not another (md).

doc/default.nix
philiptaron: Sure thing. Side note: I'm a filesystems engineer by trade, and so fs _always_ means file system to me ;-)

PR #320106, opened 2024-06-15
llvmPackages_12.compiler-rt: fix symlink conflict on darwin

414owen:
## Description of changes

When building llvm's compiler-rt-libc on an M2 Aarch64 Darwin, I consistently ran into a cmake error, caused by a duplicate symlink creation attempt:

```
CMake Error: failed to create symbolic link '/tmp/nix-build-compiler-rt-libc-12.0.1.drv-0/compiler-rt-12.0.1.src/build/lib/builtins/outline_atomic_helpers.dir/outline_atomic_cas4_3.S': File exists
```

A patch that seems to make CMake create these symlinks only once, is [here](https://github.com/llvm/llvm-project/commit/4dec62f4d4a0a496a8067e283bf66897fbf6e598).

The patch I've added ports these changes to the 12.x release.

This will cause an absolute tonne of rebuilds.

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [x] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/development/compilers/llvm/12/compiler-rt/fix-duplicate-atomics-symlink.patch
+++ b/pkgs/development/compilers/llvm/12/compiler-rt/fix-duplicate-atomics-symlink.patch
@@ -1,0 +1,19 @@
+diff --git a/compiler-rt/lib/builtins/CMakeLists.txt b/compiler-rt/lib/builtins/CMakeLists.txt
+index 73b6bead8424..73acce29b1a4 100644
+--- a/compiler-rt/lib/builtins/CMakeLists.txt
++++ b/compiler-rt/lib/builtins/CMakeLists.txt
+@@ -525,10 +525,12 @@ foreach(pat cas swp ldadd ldclr ldeor ldset)
+   foreach(size 1 2 4 8 16)
+     foreach(model 1 2 3 4)
+       if(pat STREQUAL "cas" OR NOT size STREQUAL "16")
++        set(source_asm "${CMAKE_CURRENT_SOURCE_DIR}/aarch64/lse.S")
+         set(helper_asm "${OA_HELPERS_DIR}/outline_atomic_${pat}${size}_${model}.S")
+         add_custom_command(
+-          OUTPUT ${helper_asm}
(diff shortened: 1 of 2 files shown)


RossComputerGuy (changes requested):
1. Needs to have a properly formatted title such as `llvmPackages_12.compiler-rt: fix symlink conflict on darwin`
2. The patch should be added conditionally using `optional`
3. Could you share more details of the issue? I see it's only for LLVM 12. Does this affect later versions?

414owen:
@RossComputerGuy I'm able to build `llvmPackages_13.compiler-rt-libc` just fine. So this seems to only affect llvm 12. I've checked out the 13.x branch of llvm-project, and it seems to have a slightly different stanza in its CMakeLists.

As to why cmake refuses to overwrite symlinks with new ones on MacOS, but not linux, I couldn't tell you. Maybe this case isn't even hit on linux, for some reason.

I don't really know why that stanza is executed twice. Maybe it's a simple as that cmake file is imported by multiple other cmake files?

I can confirm that this patch fixes the build, and I know enough CMake to know that you should always specify the dependencies of a custom target.

414owen:
Actually the reason I didn't have an `optional` clause initially, is that I can't say that this case is *never* hit on linux. Hydra doesn't seem to have any problems on darwin *or* linux (although I haven't dug through the build history).

414owen:
@RossComputerGuy Hit merge conflicts. Rebased.

414owen:
@RossComputerGuy I implemented the changes you requested.

mrcjkb:
> This will cause an absolute tonne of rebuilds.

In that case, you should probably redirect this PR to the `staging` branch.

mrcjkb:
Tried applying the patch on `master` (https://github.com/NixOS/nixpkgs/commit/f79630208453fd557a8c32b62a7e2c7e9fb34636) and it can no longer be applied.
Also, the build now fails with a new error: https://github.com/NixOS/nixpkgs/issues/359591.

paparodeo:
this should already be fixed on staging and staging-24.11 branches.

https://github.com/NixOS/nixpkgs/pull/356814
https://github.com/NixOS/nixpkgs/pull/348568

PR #320122, opened 2024-06-15
pcsx2: darwin support, add maintainer

matteo-pacini:
## Description of changes

- Split the package in `linux.nix` and `darwin.nix`
- `darwin.nix` fetches the pre-compiled binaries from GitHub
    - Tried to compile this but fails on `aarch64-darwin`, requires a `x86_64-darwin` machine
    -  Cannot be compiled also as Xcode is needed to build Metal shaders
    - The pre-compiled app runs fine under Rosetta, the version is in sync with the Linux one
- The three files have been formatted with `nixfmt-rfc-style`.
- Added myself as maintainer for Darwin

Screenshot 2024-06-15 at 20 09 40

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [x] x86_64-darwin
  - [X] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/by-name/pc/pcsx2/darwin.nix
+++ b/pkgs/by-name/pc/pcsx2/darwin.nix
@@ -1,0 +1,32 @@
+{
+  stdenvNoCC,
+  fetchurl,
+  pname,
+  version,
+  meta,
+  makeWrapper
+}:
+stdenvNoCC.mkDerivation (finalAttrs: {
+  inherit pname version meta;
+
+  src = fetchurl {
+    url = "https://github.com/PCSX2/pcsx2/releases/download/v${version}/pcsx2-v${version}-macos-Qt.tar.xz";
+    hash = "sha256-QdYV63lrAwYSDhUOy4nB8qL5LfZkrg/EYHtY2smtZuk=";
+  };
+
+  nativeBuildInputs = [ makeWrapper ];
+
+  dontPatch = true;
+  dontConfigure = true;
+  dontBuild = true;
+
+  sourceRoot = ".";
+
+  installPhase = ''
+    runHook preInstall
+    mkdir -p $out/{bin,Applications}
+    cp -r "PCSX2-v${finalAttrs.version}.app" $out/Applications/PCSX2.app
+    makeWrapper $out/Applications/PCSX2.app/Contents/MacOS/PCSX2 $out/bin/pcsx2-qt
+    runHook postInstall
+  '';
+})
--- a/pkgs/by-name/pc/pcsx2/package.nix
+++ b/pkgs/by-name/pc/pcsx2/package.nix
@@ -1,139 +1,11 @@
-{ cmake
-, fetchFromGitHub
-, lib
-, llvmPackages_17
-, callPackage
-, cubeb
-, curl
-, extra-cmake-modules
-, fetchpatch
-, ffmpeg
-, libaio
-, libbacktrace
-, libpcap
-, libwebp
-, libXrandr
-, lz4
-, makeWrapper
-, pkg-config
-, qt6
-, SDL2
-, soundtouch
-, strip-nondeterminism
-, vulkan-headers
-, vulkan-loader
-, wayland
-, zip
-, zstd
+{
+  stdenv,
+  lib,
+  callPackage,
 }:
-
 let
-  shaderc-patched = callPackage ./shaderc-patched.nix { };
-  # The pre-zipped files in releases don't have a versioned link, we need to zip them ourselves
-  pcsx2_patches = fetchFromGitHub {
-    owner = "PCSX2";
-    repo = "pcsx2_patches";
-    rev = "b3a788e16ea12efac006cbbe1ece45b6b9b34326";
-    sha256 = "sha256-Uvpz2Gpj533Sr6wLruubZxssoXefQDey8GHIDKWhW3s=";
-  };
-  inherit (qt6)
-    qtbase
-    qtsvg
-    qttools
-    qtwayland
-    wrapQtAppsHook
-  ;
-in
-llvmPackages_17.stdenv.mkDerivation (finalAttrs: {
   pname = "pcsx2";
   version = "1.7.5779";
-
-  src = fetchFromGitHub {
-    owner = "PCSX2";
-    repo = "pcsx2";
-    fetchSubmodules = true;
-    rev = "v${finalAttrs.version}";
-    sha256 = "sha256-WiwnP5yoBy8bRLUPuCZ7z4nhIzrY8P29KS5ZjErM/A4=";
-  };
-
-  patches = [
-    ./define-rev.patch
-    # Backport patches to fix random crashes on startup
-    (fetchpatch {
-      url = "https://github.com/PCSX2/pcsx2/commit/e47bcf8d80df9a93201eefbaf169ec1a0673a833.patch";
-      sha256 = "sha256-7CL1Kpu+/JgtKIenn9rQKAs3A+oJ40W5XHlqSg77Q7Y=";
-    })
-    (fetchpatch {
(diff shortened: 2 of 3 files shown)


afh:
Result of `nixpkgs-review pr 320122` run on aarch64-darwin [1](https://github.com/Mic92/nixpkgs-review)
<details>
  <ul>
    <li>pcsx2</li>
  </ul>
</details>

pkgs/by-name/pc/pcsx2/darwin.nix
afh, line 1: ❓ What are your thoughts around adding support to build this package on `x86_64-darwin` instead of installing the pre-built binary from upstream?

pkgs/by-name/pc/pcsx2/package.nix
```diff
@@ -144,9 +16,23 @@ llvmPackages_17.stdenv.mkDerivation (finalAttrs: {
...
+      hrdinka
+      govanify
+      matteopacini
+    ];
     mainProgram = "pcsx2-qt";
-    platforms = [ "x86_64-linux" ];
+    platforms = [ "x86_64-linux" ] ++ platforms.darwin;
+    sourceProvenance =
+      lib.optional stdenv.isDarwin sourceTypes.binaryNativeCode
+      ++ lib.optional stdenv.isLinux sourceTypes.fromSource;
```
afh, line 32: Nice! 👍 

My first thought was to overwrite add `sourceProvenance` in the platforms specific package files, e.g. for `darwin.nix`:
```
meta = meta // {
  sourceProvenance = lib.sourceTypes.binaryNativeCode;
};
```

Not sure if there is any benefit to that, but thought I'd share…

afh (approved):
Nicely done, @matteo-pacini, please find below some hopefully helpful comments for potential future improvement…

afh:
The app requires Rosetta to run, how could this dependency be declared in nixpkgs in general to differentiate between packages that run natively on Apple Silicon and packages needing Rosetta?

pkgs/by-name/pc/pcsx2/darwin.nix
matteo-pacini: @afh Imho,  It would make this package harder to maintain - I would make it compile from source only if it's supported on both `aarch64-darwin` and `x86_64-darwin`.

I also expect `x86_64-darwin` to slowly lose traction and disappear, so any effort towards that architecture could potentially become a waste of maintainers' time.

This is just my personal opinion, and up to debate: I expect plenty of new Nix adopter to be on `aarch64-darwin`, rather than `x86_64-darwin`.

pkgs/by-name/pc/pcsx2/package.nix
matteo-pacini: I've seen plenty of different approaches for this, not sure which is the best / preferred way of doing it.

Happy to change it if you think it can be improved.

matteo-pacini:
@afh That's a good point - my gut tells me this should only be declared as a `x86_64-darwin`, and then `aarch64-darwin` can download this anyway by using Nix config / extra platforms.

@NixOS/darwin-maintainers thoughts on this?

SuperSamus:
FYI, upstream is working on natively supporting ARM.
`https://github.com/PCSX2/pcsx2/pull/11364`

emilazy:
> That's a good point - my gut tells me this should only be declared as a `x86_64-darwin` derivation, and then `aarch64-darwin` can download this anyway by using Nix config / extra platforms.

I agree. There’s nothing about this app that is `aarch64-darwin` (currently); it just so happens that all `aarch64-darwin` machines support software that lets them run `x86_64-darwin` software.

pkgs/by-name/pc/pcsx2/darwin.nix
emilazy: It might be worth reconsidering this in light of the upcoming Apple Silicon support?

pkgs/by-name/pc/pcsx2/package.nix
afh: I don't think it would be an improvement per se, just a different way of doing it. Until there is a good reason to change it or a better way to do it, I'd leave it as is 🙂

pkgs/by-name/pc/pcsx2/package.nix
```diff
@@ -144,9 +16,23 @@ llvmPackages_17.stdenv.mkDerivation (finalAttrs: {
...
+      lgpl3Plus
+    ];
+    maintainers = with maintainers; [
+      hrdinka
+      govanify
+      matteopacini
+    ];
     mainProgram = "pcsx2-qt";
-    platforms = [ "x86_64-linux" ];
+    platforms = [ "x86_64-linux" ] ++ platforms.darwin;
```
afh, line 29: Given the latest comments as this PR stands now I kindly request the following change:
```suggestion
    platforms = [ "x86_64-linux" "x86_64-darwin" ];
```

pkgs/by-name/pc/pcsx2/darwin.nix
afh: That makes sense, @matteo-pacini, I'm alright with this being `x86_64-darwin` until Apple Silicon supports lands upstream and this can be built from source.

matteo-pacini:
@SuperSamus @afh @emilazy Thanks for the feedback, I'm going to change the derivation and make it natively compile for `x86_64-darwin`, so it's ready for when `aarch64-darwin` support will be officially released.

afh (changes requested):
Thanks everyone for chiming in on this, very helpful 🙏

@matteo-pacini please find below updated comments 🙂

matteo-pacini:
@emilazy hit a roadblock, got most of the project to compile, but it fails when trying to compile Metal shaders, as that would require the host's `xcrun` and `Xcode`.

What's the best way forward here?
1) Using tools on the host would be tricky, i.e. it will depend on Xcode - lot of impurity.
2) I could patch the app so the shaders are not compiled, just copied over, but then I would have to change the renderer pipeline, so those are compiled at runtime instead. Not impossible but this would be highly time consuming. Also we don't know for how long this will work and might require constant maintenance over time.
3) Not a big fan of this idea: fetch the binary dist from Github regardless, and copy over the pre-compiled `*.metallib` files to the final app bundle
4) We could just stick to the binary distribution as a temporary solution, and think how to bring Metal shader compiling into the packaging pipeline, so other derivations can benefit from it too, and revisit this at a later stage.

Pinging also @NixOS/darwin-maintainers, any input will be greatly appreciated.

matteo-pacini:
@emilazy @afh reverted this PR to its previous status.

As stated in https://github.com/NixOS/nixpkgs/pull/320122#issuecomment-2185396923, some thinking is required to bring Metal shaders into the pipeline.
I don't thinks there's an immediate / obvious solution to this, so I suggest to go with option (4) from the previous comment, rather than having no Darwin support at all.

I'll bring this topic up on Matrix when I got some time 👍🏻

afh (approved):
I agree with you, @matteo-pacini.

Would you be interested in setting up a draft PR for the native darwin build that also tracks the development of Apple silicon in upstream? Might be helpful to get folks together and advise on best paths forward in regards to xcrun and Xcode…

pkgs/by-name/pc/pcsx2/package.nix
toonn: Don't really have an opinion on where to put the attribute, whatever works is fine.

However, packages that are pretty much just fetching a binary usually set `hydraPlatforms = []` because binaries don't really add value to the cache. In this case that would have to be `hydraPlatforms = lib.platforms.linux` or a different list as long as it excludes the Darwins.

matteo-pacini: @toonn done, and thanks for the feedback!

afh (approved):
🙂👍

toonn (approved):
LGTM

PR #320146, opened 2024-06-15
simplemoji: init at 0.2.0

SergioRibera:
## Description of changes
Application for look your emojis

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/by-name/si/simplemoji/package.nix
+++ b/pkgs/by-name/si/simplemoji/package.nix
@@ -1,0 +1,60 @@
+{
+  autoPatchelfHook,
+  darwin,
+  fetchFromGitHub,
+  lib,
+  libxkbcommon,
+  fontconfig,
+  pkg-config,
+  rustPlatform,
+  stdenv,
+  wayland,
+  xorg,
+}:
+
+rustPlatform.buildRustPackage rec {
+  pname = "simplemoji";
+  version = "0.2.0";
+
+  src = fetchFromGitHub {
+    owner = "SergioRibera";
+    repo = "simplemoji";
+    rev = "v${version}";
+    hash = "sha256-MU31gRikMqXelWjwjMPZtfbDpybYlPTFr1u9w2843Kc=";
+  };
+
+  cargoHash = "sha256-NvNNc+V7wloPxM532uNrXoou4vnKLhFK29L2/17eB98=";
+
+  nativeBuildInputs = [
+    pkg-config
+  ] ++ lib.optionals stdenv.hostPlatform.isLinux [ autoPatchelfHook ];
+
+  buildInputs = lib.optionals stdenv.hostPlatform.isLinux [
+    stdenv.cc.cc.libgcc
+    fontconfig.dev
+    libxkbcommon.dev
+    wayland
+    xorg.libxcb
+    xorg.libX11
+  ];
+
+  doCheck = false;
+
+  runtimeDependencies = lib.optionals stdenv.hostPlatform.isLinux [
+    wayland
+    libxkbcommon
+    xorg.libX11
+    xorg.libXcursor
+    xorg.libXrandr
+    xorg.libXi
+  ];
+
+  meta = {
+    description = "Fast Application for look your amazing emojis write in Rust";
+    homepage = "https://github.com/SergioRibera/simplemoji";
+    license = lib.licenses.mit;
+    maintainers = with lib.maintainers; [ sergioribera ];
+    mainProgram = "simplemoji";
+    platforms = lib.platforms.linux ++ lib.platforms.darwin;
+  };
+}


pkgs/by-name/si/simplemoji/package.nix
```diff
@@ -0,0 +1,64 @@
...
+  noto-fonts-color-emoji,
+}:
+
+rustPlatform.buildRustPackage rec {
+  pname = "simplemoji";
+  version = "0.2.0";
+
+  src = fetchFromGitHub {
+    owner = "SergioRibera";
+    repo = pname;
```
AndersonTorres, line 21:
```suggestion
    repo = "simplemoji";
```

pkgs/by-name/si/simplemoji/package.nix
```diff
@@ -0,0 +1,64 @@
...
+
+  cargoHash = "sha256-NvNNc+V7wloPxM532uNrXoou4vnKLhFK29L2/17eB98=";
+
+  nativeBuildInputs = [
+    autoPatchelfHook
+    pkg-config
+  ];
+
+  buildInputs = [
+    stdenv.cc.cc.libgcc or null # libgcc_s.so
```
AndersonTorres, line 34: Why the `or null`?

pkgs/by-name/si/simplemoji/package.nix
```diff
@@ -0,0 +1,64 @@
...
+  ] ++ lib.optionals stdenv.hostPlatform.isLinux [
+    fontconfig.dev
+    libxkbcommon.dev
+    wayland
+    xorg.libxcb
+    xorg.libXrandr
+    wayland # libwayland-client.so
+    xorg.libX11 # libX11.so
+    xorg.libXcursor # libXcursor.so
+    xorg.libXi # libXi.so
```
AndersonTorres, line 44:
```suggestion
    wayland
    xorg.libX1
    xorg.libXcursor
    xorg.libXi
```

pkgs/by-name/si/simplemoji/package.nix
```diff
@@ -0,0 +1,64 @@
...
+  ];
+
+  runtimeDependencies = [
+    noto-fonts-color-emoji
+  ] ++ lib.optionals stdenv.hostPlatform.isLinux [
+    wayland # libwayland-client.so
+    libxkbcommon # libxkbcommon.so
+    xorg.libX11 # libX11.so
+    xorg.libXcursor # libXcursor.so
+    xorg.libXi # libXi.so
```
AndersonTorres, line 54:
```suggestion
    wayland
    libxkbcommon
    xorg.libX1
    xorg.libXcursor
    xorg.libXi
```

pkgs/by-name/si/simplemoji/package.nix
```diff
@@ -0,0 +1,64 @@
...
+  ];
+
+  meta = with lib; {
+    description = "Fast Application for look your amazing emojis write in Rust";
+    homepage = "https://github.com/SergioRibera/simplemoji";
+    license = licenses.mit;
+    maintainers = with maintainers; [ sergioribera ];
+    mainProgram = "simplemoji";
+  };
```
AndersonTorres, line 63:
```suggestion
  meta = {
    description = "Fast Application for look your amazing emojis write in Rust";
    homepage = "https://github.com/SergioRibera/simplemoji";
    license = lib.licenses.mit;
    maintainers = with lib.maintainers; [ sergioribera ];
    mainProgram = "simplemoji";
  };
```

AndersonTorres (changes requested):

pkgs/by-name/si/simplemoji/package.nix
SergioRibera: My mistake, I honestly missed it

pkgs/by-name/si/simplemoji/package.nix
RoGreat: Looks similar to what is done [here](https://github.com/NixOS/nixpkgs/pull/319461). I don't know if all of those runtime dependencies are necessary, so you might want to check with a tool like `strace`.

SergioRibera: Yep, I was reviewing some PRs to see how to put together the file and from there I took it, I left the comments because I found them useful.

Originally I took the necessary dependencies from the original package [here](https://github.com/SergioRibera/Simplemoji/blob/main/nix/default.nix)

pkgs/by-name/si/simplemoji/package.nix
SergioRibera: Real question, why not use the `pname` recursively?

SergioRibera:
Sorry for the several pushes, I was testing and it seems that some xorg libraries were no longer needed, I will have to update the source of the repo to take it into account.

> [!NOTE]
> If there is a possibility to test the macOS target, please let me know, for now I did the tests in x86_64-linux.

pkgs/by-name/si/simplemoji/package.nix
h7x4: `pname` and the repository name are not inherently connected. While they happen to be the same in this case, one is not a function of the other.

AndersonTorres: https://github.com/NixOS/nixpkgs/issues/277994

SergioRibera: @h7x4 @AndersonTorres Amazing, thanks a lot for the explanation, here I fixed it

RoGreat (approved):
For the Darwin version I had some success by making sure the `autoPatchelfHook` was used only for Linux. That would also include the `runtimeDependencies` array.

AndersonTorres (approved):
I have some doubts about using `xorg.`,  since the Xorg libraries can be called directly.

Besides, LGTM

pkgs/by-name/si/simplemoji/package.nix
```diff
@@ -0,0 +1,64 @@
...
+    pkg-config
+  ];
+
+  buildInputs = lib.optionals stdenv.isDarwin [
+      darwin.apple_sdk.frameworks.AppKit
+      darwin.apple_sdk.frameworks.CoreFoundation
+      darwin.apple_sdk.frameworks.CoreGraphics
+      darwin.apple_sdk.frameworks.Foundation
+      darwin.apple_sdk.frameworks.OpenGL
+    ] ++ lib.optionals stdenv.hostPlatform.isLinux [
```
Aleksanaa, line 40:
```suggestion
  ] ++ lib.optionals stdenv.hostPlatform.isLinux [
```

pkgs/by-name/si/simplemoji/package.nix
```diff
@@ -0,0 +1,64 @@
...
+    libxkbcommon
+    xorg.libX11
+  ];
+
+  meta = {
+    description = "Fast Application for look your amazing emojis write in Rust";
+    homepage = "https://github.com/SergioRibera/simplemoji";
+    license = lib.licenses.mit;
+    maintainers = with lib.maintainers; [ sergioribera ];
+    mainProgram = "simplemoji";
```
Aleksanaa, line 62:
```suggestion
    mainProgram = "simplemoji";
    platforms = with lib.platforms; linux ++ darwin;
```

pkgs/by-name/si/simplemoji/package.nix
```diff
@@ -0,0 +1,64 @@
...
+    stdenv.cc.cc.libgcc
+    fontconfig.dev
+    libxkbcommon.dev
+    wayland
+    xorg.libxcb
+    xorg.libX11
+  ];
+
+  runtimeDependencies = [
+    noto-fonts-color-emoji
```
Aleksanaa, line 50: We are `patchelf` it into the package?? If you need it, you probably want to put it somewhere in the package, or offer it by environment variables set in a wrapper.

Aleksanaa: I mean, the fonts package is just fonts, it's not a library. It makes no sense to put it in there. The `autoPatchelfHook` needs to be put in `nativeBuildInputs` of course

SergioRibera: @Aleksanaa I'm moved `autoPatchelfHook` as runtime dependency, thanks for the suggestion

SergioRibera: > I mean, the fonts package is just fonts, it's not a library. It makes no sense to put it in there.

I'm afraid not, the application doesn't embed the source in the final binary, it loads it at runtime so the sources are as a runtime dependency, maybe a good idea is to add a configuration parameter so they can change the source, but I don't know how to do that.

> The `autoPatchelfHook` needs to be put in `nativeBuildInputs` of course

My mistake, I'll fix it in a couple of hours please

Aleksanaa: Some things you can't judge by name. The role of runtimeDependencies here is consumed by autoPatchelfHook.

Aleksanaa: There's no "runtime dependencies" like traditional package managers. If you want your result to depend on another output path, you have to associate it with that output path (Which usually means you need to include it explicitly).

SergioRibera: @Aleksanaa 
> Some things you can't judge by name. The role of runtimeDependencies here is consumed by autoPatchelfHook.

Okay, I got it all wrong.

> There's no "runtime dependencies" like traditional package managers.

So, it seems to me that in a new version I will have to give support to some parameter to assign custom fonts, for the moment how can I do so that whenever you use the `simplemoji` package also install the `noto-fonts-color-emoji`?

Aleksanaa: If you are not playing tricks on searching fonts (well if you definitely want to bind the font, you should), it uses fontconfig. There should be some fontconfig environment variables, and you can wrap your binary with it. But it seems that other emoji pickers don't specially handle this, so they actually don't care about whether they can be displayed or not.

Or maybe you need to use a set of SVG to display emojis instead of directly displaying fonts.

SergioRibera: @Aleksanaa That would require a change in the way the application works, I am willing to do it, but I would like to know if there is any option I can do so that this PR can be merged and in a next version make the change.

Would removing the `noto-fonts-color-emoji` dependency be enough?

pkgs/by-name/si/simplemoji/package.nix
```diff
@@ -0,0 +1,55 @@
...
+
+  runtimeDependencies = lib.optionals stdenv.isLinux [ wayland libxkbcommon xorg.libX11 ];
+
+  meta = {
+    description = "Fast Application for look your amazing emojis write in Rust";
+    homepage = "https://github.com/SergioRibera/simplemoji";
+    license = lib.licenses.mit;
+    maintainers = with lib.maintainers; [ sergioribera ];
+    mainProgram = "simplemoji";
+    platforms = with lib.platforms; linux ++ darwin;
```
Aleksanaa, line 53:
```suggestion
    platforms = lib.platforms.linux ++ lib.platforms.darwin;
```

Not needed every time, but this time the `darwin` variable escaped `with`.

FliegendeWurst:
@ofborg build simplemoji

pkgs/by-name/si/simplemoji/package.nix
```diff
@@ -0,0 +1,61 @@
...
+  src = fetchFromGitHub {
+    owner = "SergioRibera";
+    repo = "simplemoji";
+    rev = "v${version}";
+    hash = "sha256-MU31gRikMqXelWjwjMPZtfbDpybYlPTFr1u9w2843Kc=";
+  };
+
+  cargoHash = "sha256-NvNNc+V7wloPxM532uNrXoou4vnKLhFK29L2/17eB98=";
+
+  nativeBuildInputs = [ pkg-config ] ++ lib.optionals stdenv.isLinux [ autoPatchelfHook ];
```
FliegendeWurst, line 28:
```suggestion
  nativeBuildInputs = [ pkg-config ] ++ lib.optionals stdenv.hostPlatform.isLinux [ autoPatchelfHook ];
```

pkgs/by-name/si/simplemoji/package.nix
```diff
@@ -0,0 +1,61 @@
...
+
+  buildInputs =
+    lib.optionals stdenv.isDarwin [
```
FliegendeWurst, line 31:
```suggestion
    lib.optionals stdenv.hostPlatform.isDarwin [
```

pkgs/by-name/si/simplemoji/package.nix
```diff
@@ -0,0 +1,61 @@
...
+      darwin.apple_sdk.frameworks.AppKit
+      darwin.apple_sdk.frameworks.CoreFoundation
+      darwin.apple_sdk.frameworks.CoreGraphics
+      darwin.apple_sdk.frameworks.Foundation
+      darwin.apple_sdk.frameworks.OpenGL
+    ]
+    ++ lib.optionals stdenv.isLinux [
```
FliegendeWurst, line 38:
```suggestion
    ++ lib.optionals stdenv.hostPlatform.isLinux [
```

pkgs/by-name/si/simplemoji/package.nix
```diff
@@ -0,0 +1,61 @@
...
+      stdenv.cc.cc.libgcc
+      fontconfig.dev
+      libxkbcommon.dev
+      wayland
+      xorg.libxcb
+      xorg.libX11
+    ];
+
+  runtimeDependencies = lib.optionals stdenv.isLinux [
```
FliegendeWurst, line 47:
```suggestion
  runtimeDependencies = lib.optionals stdenv.hostPlatform.isLinux [
```

FliegendeWurst (dismissed):
Fails to run with
```
thread 'main' panicked at /build/simplemoji-0.2.0-vendor.tar.gz/iced_winit/src/application.rs:119:10:
Create event loop: NotSupported(NotSupportedError)
```
I'm not sure the runtimeDependencies are processed correctly.

SergioRibera:
@FliegendeWurst Hey there, thanks a lot for the review, the error seems a bit strange to me since it happens when the platform is not supported, anyway I made the suggested changes, I hope it works correctly with that

pkgs/by-name/si/simplemoji/package.nix
```diff
@@ -0,0 +1,62 @@
...
+      libxkbcommon.dev
+      wayland
+      xorg.libxcb
+      xorg.libX11
+    ];
+
+  runtimeDependencies = lib.optionals stdenv.hostPlatform.isLinux [
+    wayland
+    libxkbcommon
+    xorg.libX11
```
FliegendeWurst, line 51: Also add libXcursor, libXrandr, libXi

SergioRibera:
Hey @FliegendeWurst, all bugs and changes are fixed, let me know any other changes that are needed

pkgs/by-name/si/simplemoji/package.nix
```diff
@@ -0,0 +1,67 @@
...
+
+  buildInputs =
+    lib.optionals stdenv.hostPlatform.isDarwin [
+      darwin.apple_sdk.frameworks.Accessibility
+      darwin.apple_sdk.frameworks.AppKit
+      darwin.apple_sdk.frameworks.CoreFoundation
+      darwin.apple_sdk.frameworks.CoreGraphics
+      darwin.apple_sdk.frameworks.Foundation
+      darwin.apple_sdk.frameworks.OpenGL
+    ]
```
Aleksanaa, line 40:
```suggestion
```

Not needed anymore

SergioRibera: Oooh, I didn't know it was no longer necessary, thank you!

SergioRibera:
@Aleksanaa @FliegendeWurst All changes made, thanks for your suggestions

FliegendeWurst:
Program launches successfully but there is no search bar.

Also, the combination of "copy to clipboard" + "close" will lead to an empty clipboard on many Linux systems using X. So by default the program is almost impossible to use, since it will close when losing focus.

SergioRibera:
@FliegendeWurst The search bar is activated with a flag `-s` and the copy is because of the `alboard` dependency, that's why it has the `-c` command to pass a custom copy command to it.

You may want to try running `simplemoji -s -x -c "xclip"`

What I am thinking is that the package may offer a very Nix way to configure these options and wrap the package.

SergioRibera:
Closing because I'll open other with the new version

PR #276558, opened 2023-12-24
nixos: opengl -> impure-drivers.opengl

SomeoneSerge:
## Description of changes

WIP. Following https://github.com/NixOS/nixpkgs/pull/275241 and https://github.com/NixOS/nixpkgs/pull/269475

The motivation is that it's extremely confusing for the new users when instructions for enabling something like CUDA involve setting `hardware.opengl.enable` and `services.xserver.videoDrivers`. The module's new name is meant to suggest that it's there to ensure integration with the `addDriverRunpath`, and it hopefully explains why/when the module has to be enabled (when we're forced to link drivers "impurely").

I think it'd also make sense to introduce an option like `hardware.impure-drivers.cuda.enable`, which would guide the user through the process of setting up the x11 or datacenter drivers (e.g. through assertions). I'd keep that for a separate PR

Thoughts? @jonringer @Kiskae @NixOS/cuda-maintainers @NixOS/rocm-maintainers

## Things done

- [x] Tested, as applicable:
  - [NixOS test(s)](https://nixos.org/manual/nixos/unstable/index.html#sec-nixos-tests) (look inside [nixos/tests](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests))
  - and/or [package tests](https://nixos.org/manual/nixpkgs/unstable/#sec-package-tests)
  - or, for functions and "core" functionality, tests in [lib/tests](https://github.com/NixOS/nixpkgs/blob/master/lib/tests) or [pkgs/test](https://github.com/NixOS/nixpkgs/blob/master/pkgs/test)
  - made sure NixOS tests are [linked](https://nixos.org/manual/nixpkgs/unstable/#ssec-nixos-tests-linking) to the relevant packages
  - [x] `nixosTests.quake3`
  - [x] `nixosTests.tinywl`
  - [ ] `nixosTests.openarena`
  - [ ] `nixosTests.seatd`
  - [ ] `nixosTests.cagebreak`
- [24.05 Release Notes](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2405.section.md) (or backporting [23.05](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2305.section.md) and [23.11](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2311.section.md) Release notes)
  - [ ] (Package updates) Added a release notes entry if the change is major or breaking
[...]

PR #158079, opened 2022-02-04
WIP nixos/opengl: move to hardware.drivers

jonringer:
###### Motivation for this change
Implementation part of https://github.com/NixOS/rfcs/pull/121

closes: #141803

###### Things needing to be done:

- [x] `services.xserver.videoDrivers` -> `hardware.gpu.drivers`
- [x] `hardware.opengl` -> `hardware.drivers`
- [x] Update nixos manual
- [x] Update other nixos modules
- [ ] Update hardcoded references to `/run/opengl-driver` (postponed til after 22.11)
- [x] Rename `addOpenGLRunpath` -> `addHardwareRunpath` https://github.com/NixOS/nixpkgs/pull/196174
- [ ] Add release notes

Issue #141803, opened 2021-10-15
Deprecate usage of OpenGL utilties and NixOS options.


jonringer:
~~The `/run/opengl-driver/` directory contains more than just opengl libraries now. It also includes vulkan implementations, machine learning libraries, hardware-accelerated encoding and decoding, opencl implementations, and many others.~~ Directory renaming is now out of scope

This directory is also being used as the way to get access to the host machines gpu drivers. I think it's another source of confusion for many users to have to add `addOpenGLRunpath` so that they can get a vulkan game / cuda workload / opencl workload to work.

I think we should deprecate this prior to the 21.11 release, and fully move it to  something like `/run/gpu-driver/` or something more representative of the desired affect.

~~I would also like if someone specifies a value for `services.xserver.videoDrivers`, that they would also get the gpu libraries mounted automatically instead of needing to do a separate `hardware.opengl-drivers.enable = true;`.~~ This was done in https://github.com/NixOS/nixpkgs/commit/02cef04c8187 

cc @ryantm 

EDIT: roadmap (2.5 years later... lol):

- addOpenGLRunpath -> addDriverRunpath
  - Init addDriverRunpath: #269475
  - Deprecate ~~addOpenGLRunpath~~ `cudaPackages.addDriverRunpath`: https://github.com/NixOS/nixpkgs/pull/281576

- autoAddOpenGLRunpathHook -> autoAddDriverRunpath
  - Init autoAddDriverRunpath: #297212
  - Deprecate cudaPackages.autoAdd{OpenGLRunpathHook,DriverRunpath}:
    -  #300277
    -  #300278
    -  #300287
    -  #300288
    -  #300290
    -  #300292
    -  #300293
    -  #300294
    -  #300295
    -  #300296
    -  #300297
    -  #300298
    -  #300299
    -  #300300
    -  #300301
    -  #300302
    - #300321
- NixOS Module changes: TODO @atemu?

jonringer:
For the transition period ( NixOS 21.11), I think keeping the old `/run/opengl-driver` would help alleviate a lot of the pain, would just like to get the ball rolling away from using `opengl` explicitly

jonringer:
I propose two changes:

- move `/run/opengl-driver/` -> `/run/gpu-driver/`
  - make `/run/opengl-driver/` a symlink to `/run/gpu-driver/` for backwards compatibility of existing derivations
- `services.xserver.videoDrivers` -> `hardware.videoDrivers`
[...]

PR #320228, opened 2024-06-16
treewide: big opengl cleanups

K900:
## Description of changes

This is long overdue.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---


mweinelt:
Manual build probably requires updating `nixos/doc/manual/configuration/gpu-accel.chapter.md`.

K900:
Yep, forgot to commit all the renames. Should be there now.

nixos/modules/hardware/graphics.nix
```diff
@@ -0,0 +1,126 @@
...
+  driversEnv32 = pkgs.buildEnv {
+    name = "graphics-drivers-32bit";
+    paths = [ cfg.package32 ] ++ cfg.extraPackages32;
+  };
+in
+{
+  imports = [
+    (lib.mkRenamedOptionModule [ "services" "xserver" "vaapiDrivers" ] [ "hardware" "opengl" "extraPackages" ])
+    (lib.mkRemovedOptionModule [ "hardware" "opengl" "s3tcSupport" ] "S3TC support is now always enabled in Mesa.")
+    (lib.mkRemovedOptionModule [ "hardware" "opengl" "driSupport"] "The setting can be removed.")
```
SuperSandro2000, line 19: We should probably mention, that this was a no-op in the first place

Tungsten842:
I am not sure that "graphics" is a good name in this case.
Consider that hardware.opengl.extraPackages is currently also used by: [opencl](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/configuration/gpu-accel.chapter.md#intel-sec-gpu-accel-opencl-intel) and level-zero.
Putting compute only APIs in a "graphics" namespace causes unnecessary confusion.
I believe that the previous naming, hardware.drivers made more sense: https://github.com/NixOS/nixpkgs/pull/158079

K900:
`hardware.drivers` is even worse, because it gives you absolutely no idea what "drivers" there are or why you should care. `graphics` is suboptimal, but `accelerators` is way too non-obvious.

Atemu:
The idea would be to have `hardware.drivers.vulkan`, `hardware.drivers.opengl` etc. which would make it quite obvious which drivers there are. Even non-nerdy people use the term "drivers" to refer to this software component. If they typed "vulkan drivers nvidia" into search.nixos.org, I believe they'd find the correct option.

I also take issue with `graphics` though, especially as many signs point to more task-specific accelerator hardware in the future that has nothing to do with graphical processing.   
I'd find enabling drivers for a compression accelerator via `hardware.graphics` equally confusing as configuring Vulkan via `hardware.opengl`. Let's not repeat that mistake.

I agree that `accelerator` is much to technical of a term but I simply know of no better term that is neutral with respect to the task being carried out.

Additionally, uninformed users should never need to be touching these. A typical desktop user for instance should only have to enable a desktop session and/or Steam and those would in turn enable the accelerator APIs typically required for those use-cases.  
Anyone who does need to touch these manually ought to learn some terminology. I think that's a reasonable ask.

I won't block on this as this is already an improvement over the status quo and I've clearly not found the priority to produce code for my preferred options API but please consider my design choices and why I made them.

K900:
Really the problem with separating the options like what you're proposing is that Mesa is all of those at the same time, and Mesa is what most people will be running. Technically, we could separate OpenGL, OpenCL, Vulkan, VAAPI etc lookup paths, since they're symlinks anyway, and just link the right manifests into place, but that feels like a lot of work for questionable gain?

Atemu:
> Really the problem with separating the options like what you're proposing is that Mesa is all of those at the same time

Not quite and that's an issue. It's a geat implementation of most of the most critical and useful acceleration APIs but by far not all of them. I love mesa but I don't think it's a good idea to equate it with hardware acceleration as a whole. There is a world outside of Mesa and it does not support every accel API under the sky and that's an issue if you center your NixOS options around everything being mesa. Think CUDA, ROCm, oneAPI etc. or even just alternative implementations of the APIs mesa supports for which there are valid use-cases.  
This is also again quite graphics-centric thinking as there are far more hardware accelerators than just GPUs but require userspace drivers in much the same way.

> Mesa is what most people will be running

That is true but enforcing that through the design of the NixOS options is the wrong way of achieving that IMO. Ensuring the user has the most sensible drivers installed should be handled through use-case-driven sane defaults.

> Technically, we could separate OpenGL, OpenCL, Vulkan, VAAPI etc lookup paths, since they're symlinks anyway, and just link the right manifests into place, but that feels like a lot of work for questionable gain?

You don't have to do but designing it like that leaves you the option to should that become relevant. In my initial implementation I still threw them into `/run/opengl-driver` but you could totally separate it in the implementation later you wanted to.

Speaking of which, I had a draft with some progress somewhere with some more discussion but I'm having a really hard time finding it. I'll link it here when I find it.

Atemu:
Found it: https://github.com/NixOS/nixpkgs/issues/141803#issuecomment-2010803483

K900:
I do really like your draft. I'll see if I can muster the spoons to actually try implementing it, because it does not seem _too_ difficult now...

jonringer:
I'll defer to @atemu on this matter. I removed most of the references of opengl from packaging concerns, module re-design I like their direction, and looks like the related information was already brought up.

K900:
OK, let's just send this, and I'll try a more ambitious cleanup over the weekend.

JohnRTitor:
Can we backport this? Needed for https://github.com/NixOS/nixos-hardware/pull/997

K900:
Definitiely not.

themaxhero:
> `hardware.drivers` is even worse, because it gives you absolutely no idea what "drivers" there are or why you should care. `graphics` is suboptimal, but `accelerators` is way too non-obvious.

What about hardware.gpu?

K900:
It's not really just GPUs, but anyway, I'll have a PR with a more granular approach to this eventually.

JohnRTitor:
> Definitely not.

Is there a reason not to? Due to nixos-hardware not being versioned, about time we get bug reports from stable users.

K900:
It's an API break. We don't do those on stable releases. nixos-hardware should just use the old options for now, and remove the removed ones.

SomeoneSerge:
Oh, not sure how I missed this. This is great and indeed long overdue, thank you @K900!

Regarding the naming, I think `hardware.graphics` isn't much different from `hardware.opengl` and about as misleading. As you say it's not "just gpus", and it's not "just graphics". I still like `hardware.drivers` (personal preference: [`hardware.impure-drivers`](https://github.com/NixOS/nixpkgs/pull/276558)) as the most precise so far. I am not concerned with how much or little information this name conveys to an uninformed reader on the first encounter. NixOS is in the process of establishing essentially a new concept here: separating pure dependencies/services from those that must be kept impure. As long as the new name (whatever it be) makes sense, is reasonably documented and pointed at with `mkRenamedOptionModule`, it's perfectly reasonable to introduce a new term once in a while.

jhvst:
Why was `setLdLibraryPath` removed?

Atemu:
AFAICT because the last driver relying on this terrible hack was removed (AMDGPU-PRO).

It should receive a `mkRenamedOptionModule` though.

What's your use-case?

jhvst:
I develop Vulkan applications and many of those libraries and programs have hard-coded load paths. The `setLdLibraryPath` helped to set aside some of those problems.

K900:
That should generally not be necessary if you link nixpkgs' libglvnd/vulkan-loader.

jhvst:
Thanks, that answers my question.

PR #307211, opened 2024-04-27
llvmPackages_{12,13,14,15,16,17,18,git}: Simplify argument passthrough

pwaller:
## Description of changes

This patch is not intended to introduce any functional change. drvPaths
should remain unchanged; if they do change, it's a bug (note: this revealed
a bug in the derivation's source paths for LLVM 13 which I've preserved
to avoid a rebuild).

Prior to this patch, a set of packages gets passed through from the
llvmPackages top level function to the individual packages via
callPackages, which is a newScope constructed with some specific
arguments.

As it stands this makes it harder to override dependencies of LLVM; for
example take ncurses. If you want to override it, it is an argument to
libllvm, however, if you override libllvm you then have to write a lot
of code to have correctly overridden clang.

Instead, I propose to make sure that all the dependencies of all
llvmPackages are listed as an inputs to the top level llvmPackages,
and then the resulting newScope will contain all of them. This in
turn will make `llvmPackages.override` work as expected for any
input to each of the llvm packages.

We'll achieve this by simplifying the code a bit and ensuring that all
arguments to llvmPackage get forwarded to all packages (via `{}@args`).

This represents a chance to simplify things a bit so I propose doing it
in two steps:

1. This patch: Simplify argument pass through.
2. (Later): Ensure all arguments to each llvm package are listed in the
   top level `llvm/X/default.nix`.

Once 2 lands, this means that `(llvmPackages.override { ncurses =
myncurses; }).clang` would consist of a clang whose libllvm had the
ncurses overridden. This is not the case prior to this patch.

~For v1 draft of this patch I have only fixed llvmPackages 18 to keep the
first pass review simple. If others agree this is a good way forward, I
will fix up all the other llvmPackages revision.~

(No longer draft after getting some preliminary private approval from a maintainer)

Signed-off-by: Peter Waller <p@pwaller.net>

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
- For non-Linux: Is sandboxing enabled in `nix.conf`? (See [Nix manual](https://nixos.org/manual/nix/stable/command-ref/conf-file.html))
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:
[...]

PR #320261, opened 2024-06-16
llvmPackages_{12,13,14,15,16,17,18,git}: Allow overriding dependencies 

pwaller:
... consistently.

This should require 0 package rebuilds.

## Description of changes

Further to https://github.com/NixOS/nixpkgs/pull/307211, allow overriding arguments through llvmPackages.override.

This makes it possible to override any dependency of LLVM or
clang by overriding it on `llvmPackages.override { <dependency> = ...; }`.

This is useful in development or customization where sometimes it is
desirable to turn features on or off.

Without this patch the only way to for example override ncurses was to
do `overriddenLLVM = llvmPackages.llvm.override { ncurses }`, but then
you would have to thread `overriddenLLVM` as dependencies into clang and
other packages, which results in quite a difficult expression to write
correctly in cross compilation scenarios.

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/development/compilers/llvm/12/default.nix
+++ b/pkgs/development/compilers/llvm/12/default.nix
@@ -1,7 +1,7 @@
-{ lowPrio, newScope, pkgs, lib, stdenv, cmake
+{ lowPrio, newScope, pkgs, lib, stdenv
 , preLibcCrossHeaders
-, substitute, substituteAll, fetchFromGitHub, fetchpatch
-, libxml2, python3, isl, fetchurl, overrideCC, wrapCCWith, wrapBintoolsWith
+, substitute, substituteAll, fetchFromGitHub, fetchpatch, fetchurl
+, overrideCC, wrapCCWith, wrapBintoolsWith
 , buildLlvmTools # tools, but from the previous stage, for cross
 , targetLlvmLibraries # libraries, but from the next stage, for cross
 , targetLlvm
@@ -17,6 +17,10 @@
     then null
     else pkgs.bintools
 , darwin
+# Allows passthrough to packages via newScope. This makes it possible to
+# do `(llvmPackages.override { <someLlvmDependency> = bar; }).clang` and get
+# an llvmPackages whose packages are overridden in an internally consistent way.
+, ...
 }@args:
 
 let
--- a/pkgs/development/compilers/llvm/13/default.nix
+++ b/pkgs/development/compilers/llvm/13/default.nix
@@ -1,7 +1,7 @@
-{ lowPrio, newScope, pkgs, lib, stdenv, cmake
+{ lowPrio, newScope, pkgs, lib, stdenv
 , preLibcCrossHeaders
-, fetchpatch
-, libxml2, python3, isl, fetchFromGitHub, substitute, substituteAll, overrideCC, wrapCCWith, wrapBintoolsWith
+, substitute, substituteAll, fetchFromGitHub, fetchpatch
+, overrideCC, wrapCCWith, wrapBintoolsWith
 , buildLlvmTools # tools, but from the previous stage, for cross
 , targetLlvmLibraries # libraries, but from the next stage, for cross
 , targetLlvm
@@ -40,7 +40,10 @@
 # to you to make sure that the LLVM repo given matches the release configuration
 # specified.
 , monorepoSrc ? null
-
+# Allows passthrough to packages via newScope. This makes it possible to
+# do `(llvmPackages.override { <someLlvmDependency> = bar; }).clang` and get
+# an llvmPackages whose packages are overridden in an internally consistent way.
+, ...
 }@args:
 
 assert
--- a/pkgs/development/compilers/llvm/14/default.nix
+++ b/pkgs/development/compilers/llvm/14/default.nix
@@ -1,6 +1,7 @@
-{ lowPrio, newScope, pkgs, lib, stdenv, cmake
+{ lowPrio, newScope, pkgs, lib, stdenv
 , preLibcCrossHeaders
-, libxml2, python3, fetchFromGitHub, substitute, substituteAll, fetchpatch, overrideCC, wrapCCWith, wrapBintoolsWith
+, substitute, substituteAll, fetchFromGitHub, fetchpatch
+, overrideCC, wrapCCWith, wrapBintoolsWith
 , buildLlvmTools # tools, but from the previous stage, for cross
 , targetLlvmLibraries # libraries, but from the next stage, for cross
 , targetLlvm
@@ -39,6 +40,10 @@
 # to you to make sure that the LLVM repo given matches the release configuration
 # specified.
 , monorepoSrc ? null
+# Allows passthrough to packages via newScope. This makes it possible to
+# do `(llvmPackages.override { <someLlvmDependency> = bar; }).clang` and get
+# an llvmPackages whose packages are overridden in an internally consistent way.
+, ...
 }@args:
 
 assert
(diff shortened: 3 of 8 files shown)


RossComputerGuy (approved):
This looks great.

pwaller:
@Artturin can this get the same backport treatment as #307211 please? :)

Thanks for the speedy review @RossComputerGuy!

paparodeo:
can this be dune with a named attribute set argument?
```nix
{ ninja, foo, bar, ...} @args:
[...]
libclang callPackage ../common/clang (args // { });
```
? i'm not sure but would be nice to not have to manually keep the union of all llvmPackges_xx.xxx args in llvm/default.nix

pwaller:
> can this be accomplished with a named attribute set argument?

Are you suggesting adding an `overrides` arg, and then doing `llvmPackages.override { overrides.ncurses = ...; }`? I suppose this would work, and allow overriding all the packages at once without needing to name all the args to all of the packages in the top level. If this was not what you were suggesting please write something more complete out so I can understand better.

I'm not sure how I feel about this. I would rather use the existing override mechanism than introduce yet-another-override pattern.

Another side effect of the PR as it stands is that it results in all of the dependencies being listed in the top level.

I note that the only harm of missing one out from the top level is that it cannot be overridden via `llvmPackags.override`, if someone discovers a need for it they can add it when the time comes.

paparodeo:
> > can this be accomplished with a named attribute set argument?
> 
> Are you suggesting adding an `overrides` arg, and then doing `llvmPackages.override { overrides.ncurses = ...; }`?

no.

https://nix.dev/tutorials/nix-language.html#functions

using a named attribute set argument should allow you to pass down the top level args to all the llvm submodules, i think.

in the top level llvm you'd use ellipses followed by `@args`. where `args` is the attribute set of all arguments including those not named which are contained in the  `...`.

then when calling the subpackage the attribute set to callPackage is setup using `(args // { patches = [ the patches ], etc = something; })`.

and the subpackage may need to be modified to take additional arguments so in libcxx say, you have to add an ellipses to the end of the argument set.

pwaller:
> using a named attribute set argument should allow you to pass down the top level args to all the llvm submodules, i think.

This is already happening and I arranged it in https://github.com/NixOS/nixpkgs/pull/307211 exactly to allow this.

Unfortunately, if the argument is not named in the parameter set, you cannot `.override` it, you get an error:

```
error: function 'anonymous lambda' called with unexpected argument 'foo'
```

See for example `nix eval -I nixpkgs=flake:nixpkgs --impure --expr '(import <nixpkgs> {}).llvmPackages.override { foo = 1; }'`.

Therefore, I currently conclude it's necessary to specify the parameters which we want to be able to override as inputs to llvmPackages as this PR does.

paparodeo:
> > using a named attribute set argument should allow you to pass down the top level args to all the llvm submodules, i think.
> 
> This is already happening and I arranged it in #307211 exactly to allow this.

thanks. i didn't see the `@args` in the diff and my tree was not updated with the change.

> 
> Unfortunately, if the argument is not named in the parameter set, you cannot `.override` it, you get an error:
> 
> ```
> error: function 'anonymous lambda' called with unexpected argument 'foo'
> ```
> 
> See for example `nix eval -I nixpkgs=flake:nixpkgs --impure --expr '(import <nixpkgs> {}).llvmPackages.override { foo = 1; }'`.

so it doesn't seem like there are any `...` in the llvm packge set. for example, if i do.

```diff
diff --git a/pkgs/development/compilers/llvm/12/default.nix b/pkgs/development/compilers/llvm/12/default.nix
index d3b823215c52..0c2267a68205 100644
--- a/pkgs/development/compilers/llvm/12/default.nix
+++ b/pkgs/development/compilers/llvm/12/default.nix
@@ -17,6 +17,7 @@
     then null
     else pkgs.bintools
 , darwin
+, ...
 }@args:
 
 let
```

i can run
```
nix-build -E 'with import ./. {}; llvmPackages_12.override { foo = 1; }'
```
just fine.

pwaller:
OK. Are you happy if I add  `, ...` to this patch? That way it would be possible to override other things in the future should they be added, and the top level documents the arguments to all llvmPackages as they are today, at least.

paparodeo:
> OK. Are you happy if I add `, ...` to this patch? That way it would be possible to override other things in the future should they be added, and the top level documents the arguments to all llvmPackages as they are today, at least.

this makes sense to me as we don't have to manually adjust the top level arguments when the subpackages change. i'm not sure what the possible downsides are.

pwaller:
Thanks @paparodeo, I've added it as well.

pkgs/development/compilers/llvm/12/default.nix
```diff
@@ -17,6 +19,7 @@
     then null
     else pkgs.bintools
 , darwin
+, ... # Allows passthrough to packages via newScope.
```
paparodeo, line 22: > OK. Are you happy if I add , ... to this patch? That way it would be possible to override other things in the future should they be added, and the top level documents the arguments to all llvmPackages as they are today, at least.

(i didn't read the last part of this comment correctly when responding to it).

thanks for adding the `...`.

> the top level documents the arguments to all llvmPackages as they are today, at least.

so i am not sure about having the union of all llvmPackages subpackage arguments in the lop level given they can go stale and it is somewhat non-obvious what all the extra arguments are for given they aren't used in `llvm/[:digit:]/default.nix`. my preference would be to only have the args used in `llvm/[:digit:]/default.nix` and let the `...` take care of the rest.

paparodeo: tho -- maybe i'm missing something as it seems that before your change there are a bunch of extra args that aren't used in the top level `default.nix`, for example `libxml2`

pwaller: I have modified the PR and updated the title and commit message accordingly. I have also gone to the trouble of removing anything which was not referenced in the default.nix.

pwaller:
Please note I won't have a chance to test this for a moment; I'm relying partially on the ofBorg package count rebuild result which should be zero and a good test. I'll also do a bit of testing over the coming days.

paparodeo (approved):
looks good. 0 builds for nixpkgs-review

Result of `nixpkgs-review pr 320261` run on x86_64-linux [1](https://github.com/Mic92/nixpkgs-review)

tried a few overrides and the correct things looked like they were being rebuilt.

```
$  nix-build -E 'with import ./. {}; llvmPackages_18.override { lua5_3 = lua5_3.overrideAttrs { foo =1; };}'
these 4 derivations will be built:
  /nix/store/z7vxw6cgh4gzhk5w44x90jnmml3ih0n4-utils.sh.drv
  /nix/store/784hhifl221ny531a8kzpbj3hmw7msdy-lua-5.3.6.drv
  /nix/store/n8sr8zdj4g3gxlmjqg85zhknb1s203fp-lldb-18.1.7.drv
  /nix/store/r7926xvs7abds7m2mjwrbyvy26g9nk6z-lldb-manpages-18.1.7.drv
```

```
$ nix-build -E 'with import ./. {}; llvmPackages_18.override { lit = lit.overrideAttrs { foo =1; };}'
these 4 derivations will be built:
  /nix/store/9zg056a72jkwrhvg2zz8yjpfp5751i0c-python3.11-lit-17.0.6.drv
  /nix/store/58qg67s9a06ci66r77sjsn8174i4s9zb-lldb-18.1.7.drv
  /nix/store/aizdk1y7ksv4pd93dwhd0kp43h4fk4d0-lldb-manpages-18.1.7.drv
  /nix/store/gp963g3pbn4a1z9fgxhaam2fiafnl6sm-openmp-18.1.7.drv
```

philiptaron (approved):
Code changes look good and consistent.

PR #257527, opened 2023-09-27
python3Packages.pythonImportsCheckHook: restrict to propagated build inputs

pbsds:
## Description of changes

Too often do i pull a python package, only for it to error on import due to `pkg_resources` from `setuptools` missing. This PR stops the problem during build, by ensuring that `pythonImportsCheck` only knows about the propagated build inputs. I also add a `pythonImportsExtrasCheck` escape hatch for packages which require the old behavior.

My computers have been running hot for a few weeks now trying to test this. Perhaps it is time I either merge it into staging/python-develop or ask for some hydra assistance.

The first commit detail the major change, and is *currently with `FIXME`s and debugging code in them*. The following commits fix all the packages that break.

A motivational photo, showcasing the diff of between the old and new `PATH` and `PYTHONPATH` entries in `pythonImportCheck`: 
image

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
- [ ] For non-Linux: Is `sandbox = true` set in `nix.conf`? (See [Nix manual](https://nixos.org/manual/nix/stable/command-ref/conf-file.html))
- [ ] Tested, as applicable:
  - [NixOS test(s)](https://nixos.org/manual/nixos/unstable/index.html#sec-nixos-tests) (look inside [nixos/tests](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests))
  - and/or [package tests](https://nixos.org/manual/nixpkgs/unstable/#sec-package-tests)
  - or, for functions and "core" functionality, tests in [lib/tests](https://github.com/NixOS/nixpkgs/blob/master/lib/tests) or [pkgs/test](https://github.com/NixOS/nixpkgs/blob/master/pkgs/test)
  - made sure NixOS tests are [linked](https://nixos.org/manual/nixpkgs/unstable/#ssec-nixos-tests-linking) to the relevant packages
- [ ] Tested compilation of all packages that depend on this change using `nix-shell -p nixpkgs-review --run "nixpkgs-review rev HEAD"`. Note: all changes have to be committed, also see [nixpkgs-review usage](https://github.com/Mic92/nixpkgs-review#usage)
- [ ] Tested basic functionality of all binary files (usually in `./result/bin/`)
[...]

PR #320266, opened 2024-06-16
versionCheckHook: init

doronbehar:
###### Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
- [ ] Tested, as applicable:

### DIFF
--- a/doc/build-helpers/testers.chapter.md
+++ b/doc/build-helpers/testers.chapter.md
@@ -120,9 +120,10 @@ It has two modes:
 
 Checks that the output from running a command contains the specified version string in it as a whole word.
 
-Although simplistic, this test assures that the main program can run.
-While there's no substitute for a real test case, it does catch dynamic linking errors and such.
-It also provides some protection against accidentally building the wrong version, for example when using an "old" hash in a fixed-output derivation.
+NOTE: In most cases, [`versionCheckHook`](#versioncheckhook) should be preferred, but this function is provided and documented here anyway. The motivation for adding either tests would be:
+
+- Catch dynamic linking errors and such and missing environment variables that should be added by wrapping.
+- Probable protection against accidentally building the wrong version, for example when using an "old" hash in a fixed-output derivation.
 
 By default, the command to be run will be inferred from the given `package` attribute:
 it will check `meta.mainProgram` first, and fall back to `pname` or `name`.
--- a/doc/hooks/versionCheckHook.section.md
+++ b/doc/hooks/versionCheckHook.section.md
@@ -1,0 +1,35 @@
+# versionCheckHook {#versioncheckhook}
+
+This hook adds a `versionCheckPhase` to the [`preInstallCheckHooks`](#ssec-installCheck-phase) that runs the main program of the derivation with a `--help` or `--version` argument, and checks that the `${version}` string is found in that output. You use it like this:
+
+```nix
+{
+  lib,
+  stdenv,
+  versionCheckHook,
+  # ...
+}:
+
+stdenv.mkDerivation (finalAttrs: {
+  # ...
+
+  nativeInstallCheckInputs = [
+    versionCheckHook
+  ];
+  doInstallCheck = true;
+
+  # ...
+})
+```
+
+Note that for [`buildPythonPackage`](#buildpythonpackage-function) and [`buildPythonApplication`](#buildpythonapplication-function), `doInstallCheck` is enabled by default.
+
+It does so in a clean environment (using `env --ignore-environment`), and it checks for the `${version}` string in both the `stdout` and the `stderr` of the command. It will report to you in the build log the output it received and it will fail the build if it failed to find `${version}`.
+
+The variables that this phase control are:
+
+- `dontVersionCheck`: Disable adding this hook to the [`preDistPhases`](#var-stdenv-preDist). Useful if you do want to load the bash functions of the hook, but run them differently.
+- `versionCheckProgram`: The full path to the program that should print the `${version}` string. Defaults roughly to `${placeholder "out"}/bin/${pname}`. Using `$out` in the value of this variable won't work, as environment variables from this variable are not expanded by the hook. Hence using `placeholder` is unavoidable.
+- `versionCheckProgramArg`: The argument that needs to be passed to `versionCheckProgram`. If undefined the hook tries first `--help` and then `--version`. Examples: `version`, `-V`, `-v`.
+- `preVersionCheck`: A hook to run before the check is done.
+- `postVersionCheck`: A hook to run after the check is done.
--- a/pkgs/by-name/ve/versionCheckHook/hook.sh
+++ b/pkgs/by-name/ve/versionCheckHook/hook.sh
@@ -1,0 +1,60 @@
+_handleCmdOutput(){
+    local versionOutput
+    versionOutput="$(env --chdir=/ --argv0="$(basename "$1")" --ignore-environment "$@" 2>&1 || true)"
+    if [[ "$versionOutput" =~ "$version" ]]; then
+        echoPrefix="Successfully managed to"
+    else
+        echoPrefix="Did not"
+    fi
+    # The return value of this function is this variable:
+    echo "$echoPrefix"
+    # And in anycase we want these to be printed in the build log, useful for
+    # debugging, so we print these to stderr.
+    echo "$echoPrefix" find version "$version" in the output of the command \
+        "$@" >&2
+    echo "$versionOutput" >&2
+}
+versionCheckHook(){
+    runHook preVersionCheck
+    echo Executing versionCheckPhase
+
+    local cmdProgram cmdArg echoPrefix
+    if [[ -z "${versionCheckProgram-}" ]]; then
+        if [[ -z "${pname-}" ]]; then
+            echo "both \$pname and \$versionCheckProgram are empty, so" \
+                "we don't know which program to run the versionCheckPhase" \
+                "upon" >&2
+            exit 2
+        else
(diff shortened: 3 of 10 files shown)


pkgs/development/python-modules/ftfy/default.nix
```diff
@@ -30,7 +31,10 @@ buildPythonPackage rec {
 
   propagatedBuildInputs = [ wcwidth ];
 
-  nativeCheckInputs = [ pytestCheckHook ];
+  nativeCheckInputs = [
+    versionInstallHook
```
pbsds, line 35: This should go in `nativeInstallCheckInputs`, which requires `doInstallCheck = true`

pkgs/by-name/ve/versionInstallHook/hook.sh
```diff
@@ -0,0 +1,56 @@
+_handleCmdOutput(){
+    local versionOutput cmdProgram chdir
+    cmdProgram=./"$(basename "$1")"
+    chdir="$(dirname "$1")"
```
pbsds, line 4: I'd prefer if a different workdir is used, since this will pass on programs which only works when in CWD, a pretty serious failure mode.

pbsds:
related effort i need to revive at some point 
* https://github.com/NixOS/nixpkgs/pull/257527

pkgs/development/python-modules/ftfy/default.nix
doronbehar: > This should go in `nativeInstallCheckInputs`, which requires `doInstallCheck = true`

Yea you are sort of correct, but thinking about it the correct thing to do would be to rename the hook to `versionCheckHook`, because it is being run as part of the `distPhase`, and hence it has nothing to do with the `installCheckPhase`.

pkgs/by-name/ve/versionInstallHook/hook.sh
doronbehar: > I'd prefer if a different workdir is used, since this will pass on programs which only works when in CWD, a pretty serious failure mode. Root might be a decent alternative

Yes so this was meant to handle a certain different edge case which I now take care of differently. I used `--chdir=/` which is a good idea.

pkgs/development/python-modules/ftfy/default.nix
doronbehar: > > This should go in `nativeInstallCheckInputs`, which requires `doInstallCheck = true`
> 
> Yea you are sort of correct, but thinking about it the correct thing to do would be to rename the hook to `versionCheckHook`, because it is being run as part of the `distPhase`, and hence it has nothing to do with the `installCheckPhase`.

Hmm on a 2nd thought, using the `distPhase` for this hook is a bit wierd, and it is not enabled by default in `stdenv`, and enabling `doDist = true;` will make the build try to run `make dist` which is also inappropriate if the goal is to run this test. I switched to `preInstallCheckHooks` eventually.

doronbehar:
I don't understand the [ofborg error](https://gist.github.com/GrahamcOfBorg/c254d4f502b20899b3e01d6f0c70bf1a)... hopefully it wasn't related and will go away.

I pushed a significant documentation rewrite - both for the hook itself, and for other tests related sections in the docs. And I added the hook to 2 non Python packages. Anyway this PR should be fully ready now.

doc/hooks/versionCheckHook.section.md
```diff
@@ -0,0 +1,35 @@
...
+```
+
+Note that for [`buildPythonPackage`](#buildpythonpackage-function) and [`buildPythonApplication`](#buildpythonapplication-function), `doInstallCheck` is enabled by default.
+
+It does so in a clean environment (using `env --ignore-environment`), and it checks for the `${version}` string in both the `stdout` and the `stderr` of the command. It will report to you in the build log the output it received and it will fail the build if it failed to find `${version}`.
+
+The variables that this phase control are:
+
+- `dontVersionCheck`: Disable adding this hook to the [`preDistPhases`](#var-stdenv-preDist). Useful if you do want to load the bash functions of the hook, but run them differently.
+- `versionCheckProgram`: The full path to the program that should print the `${version}` string. Defaults roughly to `$out/bin/$pname`.
```
pbsds, line 32: "full path" is problematic, or at least very confusing.

I would assume setting `versionCheckProgram = "$out/bin/foobar";` will not expand `$out`, and i'm not sure if `${placeholder "out"}bin/foobar` will work since, the placeholder may be know ahead of the build or after the build is finished. A canonical example of how to set this correctly is much appreciated.

doronbehar: > "full path" is problematic, or at least very confusing.
> 
> I would assume setting `versionCheckProgram = "$out/bin/foobar";` will not expand `$out`, 

Correct.

> and i'm not sure if `${placeholder "out"}bin/foobar` will work, since whether the placeholder is known ahead of the build or only after the build is finished may vary.

Really? I thought that once all the inputs to a derivation are known to Nix, it knows the output path of it...

> A canonical example of how to set this correctly is much appreciated.

Using `placeholder` worked for me, so I added an example and mentioned the fact that `$out` is not expanded and `placeholder` will probably always be needed. I also checked this interface like this:

```patch
diff --git a/pkgs/by-name/he/hello/package.nix b/pkgs/by-name/he/hello/package.nix
index 7b8db4c7c3c2..2a1540d8918b 100644
--- a/pkgs/by-name/he/hello/package.nix
+++ b/pkgs/by-name/he/hello/package.nix
@@ -23,6 +23,8 @@ stdenv.mkDerivation (finalAttrs: {
   nativeInstallCheckInputs = [
     versionCheckHook
   ];
+  versionCheckProgram = "${placeholder "out"}/bin/hello";
+  versionCheckProgramArg = "--version";
 
   # Give hello some install checks for testing purpose.
   postInstallCheck = ''
```

And it worked according to the build log.

pkgs/by-name/ve/versionCheckHook/hook.sh
```diff
@@ -0,0 +1,60 @@
...
+    if [[ "$versionOutput" =~ "$version" ]]; then
+        echoPrefix="Successfully"
+    else
+        echoPrefix="Did not"
+    fi
+    # The return value of this function is this variable:
+    echo "$echoPrefix"
+    # And in anycase we want these to be printed in the build log, useful for
+    # debugging, so we print these to stderr.
+    echo "$echoPrefix" found version "$version" in the  output of the command \
```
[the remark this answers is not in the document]
pbsds: Also, the output `Did not found` should read `Did not find`.

pbsds (commented):
Logic look good to me. I checked whether `installCheckPhase` is disabled for cross builds and that do seem to be the case. 

Some minor nits on my end, and I think an entry into the release notes would do good. :)

doronbehar:
Thanks for the rephrasing comments! I fixed the issues and will merge this in a few days.

doronbehar:
Oh as for the release notes, should I add it to `nixos/doc/manual/release-notes/rl-2411.section.md`? I'm not sure because it is not really about NixOS, but Nixpkgs...

doronbehar:
As far as I understand there are no release notes for Nixpkgs features, so I'll merge this.

doc/hooks/versionCheckHook.section.md
roberth: As someone who stumbled into this check being applied, it was confusing to me that `versionCheckProgram` had to be a full path. The placeholder does work.

doc/build-helpers/testers.chapter.md
```diff
@@ -120,9 +120,10 @@ It has two modes:
 
 Checks that the output from running a command contains the specified version string in it as a whole word.
 
-Although simplistic, this test assures that the main program can run.
-While there's no substitute for a real test case, it does catch dynamic linking errors and such.
-It also provides some protection against accidentally building the wrong version, for example when using an "old" hash in a fixed-output derivation.
+NOTE: In most cases, [`versionCheckHook`](#versioncheckhook) should be preferred, but this function is provided and documented here anyway. The motivation for adding either tests would be:
```
roberth, line 123: For content addressed derivations, a passthru test would be preferable because it tests the executable after rewriting instead of before, and rewriting may break the executable.

doronbehar: > For content addressed derivations, a passthru test would be preferable because it tests the executable after rewriting instead of before, and rewriting may break the executable.

Hmm interesting.. Perhaps we can check inside the hook whether `__contentAddressed == true` and output a message or something like that? We barely have enough content addressed derivations at the moment in Nixpkgs so I don't know how would we test the need for this check...

doc/hooks/versionCheckHook.section.md
```diff
@@ -0,0 +1,35 @@
...
+})
+```
+
+Note that for [`buildPythonPackage`](#buildpythonpackage-function) and [`buildPythonApplication`](#buildpythonapplication-function), `doInstallCheck` is enabled by default.
+
+It does so in a clean environment (using `env --ignore-environment`), and it checks for the `${version}` string in both the `stdout` and the `stderr` of the command. It will report to you in the build log the output it received and it will fail the build if it failed to find `${version}`.
+
+The variables that this phase control are:
+
+- `dontVersionCheck`: Disable adding this hook to the [`preDistPhases`](#var-stdenv-preDist). Useful if you do want to load the bash functions of the hook, but run them differently.
```
jian-lin, line 31: I created https://github.com/NixOS/nixpkgs/pull/347826 to fix typos.

PR #320288, opened 2024-06-16
 services/hardware/undervolt: Fix several issues

Pandapip1:
## Description of changes

This fixes several issues:

1. Inconsistent indentation
2. Adds an assertion that will fail if the service would fail to start due to no undervolt settings being applied.
3. Fixes incorrect temperature handling.

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/nixos/modules/services/hardware/undervolt.nix
+++ b/nixos/modules/services/hardware/undervolt.nix
@@ -8,12 +8,21 @@ let
     if (limit == null && window == null) then null
     else assert asserts.assertMsg (limit != null && window != null) "Both power limit and window must be set";
       "${toString limit} ${toString window}";
+  hasCliArgs = (
+    # One of the following must be set to enable the service
+    (cfg.coreOffset != null) ||
+    (cfg.gpuOffset != null) ||
+    (cfg.uncoreOffset != null) ||
+    (cfg.analogioOffset != null) ||
+    (cfg.temp != null) ||
+    (cfg.tempBat != null) ||
+    (cfg.turbo != null) ||
+    # Both limit and window must be set if either are set. This is enforced by the assertion in mkPLimit
+    (cfg.p1.limit != null) ||
+    (cfg.p2.limit != null)
+  );
   cliArgs = lib.cli.toGNUCommandLine {} {
-    inherit (cfg)
-      verbose
-      temp
-      turbo
-      ;
+    inherit (cfg) verbose turbo;
     # `core` and `cache` are both intentionally set to `cfg.coreOffset` as according to the undervolt docs:
     #
     #     Core or Cache offsets have no effect. It is not possible to set different offsets for
@@ -25,19 +34,23 @@ let
     uncore = cfg.uncoreOffset;
     analogio = cfg.analogioOffset;
 
+    temp = cfg.temp;
     temp-bat = cfg.tempBat;
-    temp-ac = cfg.tempAc;
 
     power-limit-long = mkPLimit cfg.p1.limit cfg.p1.window;
     power-limit-short = mkPLimit cfg.p2.limit cfg.p2.window;
   };
 in
 {
+  imports = [
+    (lib.mkRemovedOptionModule [ "options" "services" "undervolt" "tempAc" ] "Use options.services.undervolt.temp instead")
+  ];
+
   options.services.undervolt = {
     enable = mkEnableOption ''
-       Undervolting service for Intel CPUs.
+      Undervolting service for Intel CPUs.
 
-       Warning: This service is not endorsed by Intel and may permanently damage your hardware. Use at your own risk
+      Warning: Use at your own risk! This service is not endorsed by Intel and may permanently damage your hardware
     '';
 
     verbose = mkOption {
@@ -86,15 +99,7 @@ in
       type = types.nullOr types.int;
       default = null;
       description = ''
-        The temperature target in Celsius degrees.
-      '';
-    };
-
-    tempAc = mkOption {
-      type = types.nullOr types.int;
-      default = null;
-      description = ''
-        The temperature target on AC power in Celsius degrees.
+        The temperature target on AC power in degrees Celsius.
       '';
     };
 
@@ -102,7 +107,9 @@ in
       type = types.nullOr types.int;
       default = null;
       description = ''
-        The temperature target on battery power in Celsius degrees.
+        The temperature target on battery power in degrees Celsius.
+
+        If unset, defaults to the temperature target on AC power.
       '';
     };
 
@@ -161,6 +168,23 @@ in
   };
 
   config = mkIf cfg.enable {
+    assertions = [ {
+      assertion = hasCliArgs;
+      message = ''
+        At least one of the following undervolt parameters must be set to enable service.undervolt
+
+        - services.undervolt.coreOffset
+        - services.undervolt.gpuOffset
+        - services.undervolt.uncoreOffset
+        - services.undervolt.analogioOffset
+        - services.undervolt.temp
+        - services.undervolt.tempBat
+        - services.undervolt.turbo
+        - services.undervolt.p1.limit and services.undervolt.p1.window
+        - services.undervolt.p2.limit and services.undervolt.p2.window
+      '';
+    } ];
+
     hardware.cpu.x86.msr.enable = true;
 
     environment.systemPackages = [ cfg.package ];


nixos/modules/services/hardware/undervolt.nix
```diff
@@ -165,7 +177,7 @@ in
 
     environment.systemPackages = [ cfg.package ];
 
-    systemd.services.undervolt = {
+    systemd.services.undervolt = mkIf hasCliArgs {
```
eclairevoyant, line 180:
```suggestion
    systemd.services.undervolt = mkIf (cliArgs != [ ]) {
```

nixos/modules/services/hardware/undervolt.nix
```diff
@@ -179,7 +191,7 @@ in
       };
     };
 
-    systemd.timers.undervolt = mkIf cfg.useTimer {
+    systemd.timers.undervolt = mkIf (cfg.useTimer && hasCliArgs) {
```
eclairevoyant, line 194:
```suggestion
    systemd.timers.undervolt = mkIf (cfg.useTimer && (cliArgs != [ ])) {
```

eclairevoyant (commented):
wouldn't this be simpler? (obviating the need for `hasCliArgs`)

jian-lin (commented):
I think it is better to use [assetions](https://nixos.org/manual/nixos/unstable/#sec-assertions-assetions) than `mkIf`.

nixos/modules/services/hardware/undervolt.nix
jian-lin: I am afraid this will not work.  If users only set `services.undervolt.verbose = true` but no other settings, then `cliArgs` will be `[ "--verbose" ]`

eclairevoyant: eh fair enough, I still feel it's redundant to list out every single flag though. Maybe exclude the flags that don't allow the service to start?

Pandapip1: It's redundant, but this is one of the times where a tradeoff has to be made between correct functionality and pretty code. This code, while not pretty, is self-evident. Having a list of flags be *excluded* will result in harder-to-read code that looks marginally prettier.

Pandapip1:
> I think it is better to use [assetions](https://nixos.org/manual/nixos/unstable/#sec-assertions-assetions) than mkIf.

Could you provide a rationale? I'm not opposed to it, but my rationale for using `mkIf` is twofold:

1. It simplifies configuration for multiple machines
2. The expected result of enabling the undervolt service with no arguments is nothing, not a compile-time error

nixos/modules/services/hardware/undervolt.nix
eclairevoyant: It's not about "pretty" but "error-prone".

jian-lin:
> > I think it is better to use [assetions](https://nixos.org/manual/nixos/unstable/#sec-assertions-assetions) than mkIf.
> 
> Could you provide a rationale? 

The semantics of `services.foo.enable = true` is to **run** the **service**.  For example, you can tell that from the description of `services.undervolt.enable`: `Whether to enable Undervolting service for Intel CPUs`.  The *service* means some systemd services together with some related system configurations.

So when a user sets `services.foo.enable = true`, he expects some services to run.  If, for some reason, the services fail to run, the NixOS module ideally should give the user clear signals, such as eval-time errors which is "compile-time error" in your words.  And assertions are a perfect fit for this.

> I'm not opposed to it, but my rationale for using `mkIf` is twofold:
> 
> 1. It simplifies configuration for multiple machines

I do not follow. Could you give a real-live example?

> 2. The expected result of enabling the undervolt service with no arguments is nothing, not a compile-time error

As I said above, an eval-time error is the expected result in this case.  Hiding the fact that the service is not run even if `services.undervolt.enable = true` is a very bad behavior and is against the convention. 

What you want sounds like a new option `services.undervolt.runIfPossible`, which I do not think is a good idea.  Just do not enable `services.undervolt.enable` if you do not want to run the service.

nixos/modules/services/hardware/undervolt.nix
Pandapip1: How is what I have error-prone?

eclairevoyant: You don't think maintaining a list of every single flag by hand is error prone?

Pandapip1:
> We should give an error (when possible) regarding invalid configurations.

I disagree that the configuration would be invalid. That being said, there seems to be a consensus that it *is* somehow invalid. While I don't completely understand the rationale for it, for the time being I'll switch to assertions.

nixos/modules/services/hardware/undervolt.nix
jian-lin: > Maintaining a list of every single flag by hand is error prone

This is true.  However, I guess we have to do this for undervolt.  Both of your proposals are essentially doing so.  The difference is maintaining every "important" flags or maintaining every "unimportant" flags like `verbose`.  I do not have a strong opinion on this.

Pandapip1: > Maintaining a list of every single flag by hand is error prone.

No it's not?

nixos/modules/services/hardware/undervolt.nix
```diff
@@ -161,11 +173,16 @@ in
...
+      assertion = !hasCliArgs;
+      message = "At least one undervolt parameter must be set to enable service.undervolt";
+    } ];
+
     hardware.cpu.x86.msr.enable = true;
 
     environment.systemPackages = [ cfg.package ];
 
-    systemd.services.undervolt = {
+    systemd.services.undervolt = mkIf hasCliArgs {
```
jian-lin, line 185: nit: after an assertion is added, there is not need to check `mkIf hasCliArgs` again.

nixos/modules/services/hardware/undervolt.nix
```diff
@@ -161,6 +173,11 @@ in
   };
 
   config = mkIf cfg.enable {
+    assertions = [ {
+      assertion = !hasCliArgs;
```
jian-lin, line 177:
```suggestion
      assertion = hasCliArgs;
```

nixos/modules/services/hardware/undervolt.nix
```diff
@@ -8,6 +8,18 @@ let
...
+    (cfg.gpuOffset != null) ||
+    (cfg.uncoreOffset != null) ||
+    (cfg.analogioOffset != null) ||
+    (cfg.tempAc != null) ||
+    (cfg.tempBat != null) ||
+    (cfg.p1.limit != null) ||
+    (cfg.p1.window != null) ||
+    (cfg.p2.limit != null) ||
+    (cfg.p2.window != null)
+  );
```
jian-lin, line 22: Why some flags such as `temp` and `turbo` are not listed here?

nixos/modules/services/hardware/undervolt.nix
eclairevoyant: > However, I guess we have to do this for undervolt. Both of your proposals are essentially doing so. The difference is maintaining all "important" flags or maintaining all "unimportant" flags like `verbose`.

You're correct. Here's my reasoning - if we maintain an inclusion list manually, and we miss adding a flag, then the result is that the module is unusable (i.e. this blocks rebuilds) unless users are copypasting and replacing the module entirely.

If we maintain an _exclusion_ list, and the exclusion list is missing an entry, the failure mode is that the service doesn't start - which can be fixed immediately by reading upstream docs and creating a GH issue/PR to add the exclusion entry.

NB I am making the assumption that it's more likely to forget to add a new flag to either type of list, than to incorrectly add it.

nixos/modules/services/hardware/undervolt.nix
```diff
@@ -161,6 +173,11 @@ in
   };
 
   config = mkIf cfg.enable {
+    assertions = [ {
+      assertion = !hasCliArgs;
+      message = "At least one undervolt parameter must be set to enable service.undervolt";
```
eclairevoyant, line 178: This message is insufficient; we should specify which flags are needed for this config to be considered valid.

nixos/modules/services/hardware/undervolt.nix
Pandapip1: Because one of the args that are listed here need to be present for the undervolt utility to actually work. `temp` or `turbo` are insufficient.

Pandapip1: Actually, it turns out that just setting turbo is sufficient for the utility to work. Additionally, there were some issues with the way that temperature was handled, which are now fixed.

nixos/modules/services/hardware/undervolt.nix
```diff
@@ -25,19 +35,23 @@ let
...
+    (lib.mkRemovedOptionModule [ "options" "services" "undervolt" "tempAc" ] "Use options.services.undervolt.temp instead")
+  ];
+
   options.services.undervolt = {
     enable = mkEnableOption ''
-       Undervolting service for Intel CPUs.
+      Undervolting service for Intel CPUs.
 
-       Warning: This service is not endorsed by Intel and may permanently damage your hardware. Use at your own risk
+      Warning: This service is not endorsed by Intel and may permanently damage your hardware. Use at your own risk!
```
eclairevoyant, line 54: nit: period is appended here when using `mkEnableOption`, so this should be reworded to not have `!.` at the end, or use `mkOption` itself.

nixos/modules/services/hardware/undervolt.nix
```diff
@@ -8,12 +8,22 @@ let
...
+  hasCliArgs = (
+    # One of the following must be set to enable the service
+    (cfg.coreOffset != null) ||
+    (cfg.gpuOffset != null) ||
+    (cfg.uncoreOffset != null) ||
+    (cfg.analogioOffset != null) ||
+    (cfg.temp != null) ||
+    (cfg.tempBat != null) ||
+    (cfg.turbo != null) ||
+    (cfg.p1.limit != null) ||
```
eclairevoyant, line 20: in light of the assertion below:
```suggestion
    (cfg.p1.limit != null) &&
```

nixos/modules/services/hardware/undervolt.nix
```diff
@@ -8,12 +8,22 @@ let
...
+    (cfg.p1.window != null) ||
+    (cfg.p2.limit != null) ||
```
eclairevoyant, line 22:
```suggestion
    (cfg.p2.limit != null) &&
```

nixos/modules/services/hardware/undervolt.nix
jian-lin: > The service will fail to start if no undervolt settings are applied. This disables the service to not pollute the journal with failures.

This is the original version of PR description and probably is the author's motivation.

I am a bit confused about how this service pollutes the journal with failures.  There is no restart logic defined for this service so it only produces **one** failure log after `nixos-rebuild test (or switch)`.

I am not a fan of maintaining a list in assertion.  IMHO, in this case, a runtime failure signal from `nixos-rebuild` is better than an eval time error (assertion).

Pandapip1: > I am a bit confused about how this service pollutes the journal with failures. There is no restart logic defined for this service so it only produces one failure log after nixos-rebuild test (or switch).

I reboot frequently.

jian-lin: > > I am a bit confused about how this service pollutes the journal with failures. There is no restart logic defined for this service so it only produces one failure log after nixos-rebuild test (or switch).
> 
> I reboot frequently.

I do not follow.  Why don't you fix your config for `services.undervolt` when `nixos-rebuild` gives you an error?

Pandapip1: `nixos-rebuild boot` doesn't give an error though? It evals and builds the system just fine.

Pandapip1:
Abandoned; if someone would like to fork off of it they should feel welcome to do so.

PR #320332, opened 2024-06-16
vscode-extensions: document meta.maintainers usage

superherointj:
vscode-extensions: document meta.maintainers usage

CC @katexochen @Pandapip1

### DIFF
--- a/pkgs/applications/editors/vscode/extensions/README.md
+++ b/pkgs/applications/editors/vscode/extensions/README.md
@@ -20,6 +20,11 @@
   - `description` should mention it is a Visual Studio Code extension.
   - `downloadPage` is the VSCode marketplace URL.
   - `homepage` is the source-code URL.
+  - `maintainers`:
+    - optionally consider adding yourself as a maintainer to be notified of updates, breakages and help with upkeep.
+    - recommended format is:
+      - a `non-nested with`, such as: `with lib.maintainers; [ your-username ];`.
+      - maintainers are listed in alphabetical order.
   - verify `license` in upstream.
 
 * On commit messages:


pkgs/applications/editors/vscode/extensions/README.md
```diff
@@ -20,6 +20,11 @@
   - `description` should mention it is a Visual Studio Code extension.
   - `downloadPage` is the VSCode marketplace URL.
   - `homepage` is the source-code URL.
+  - `maintainers`:
+    - optionally consider adding yourself as maintainer to be notified of updates, breakages and help with upkeep.
+    - recommended format is a `non-nested with`, such as: `with lib.maintainers; [ your-username ];`.
+    - use a discrete git commit for adding yourself as a maintainer, such that, you can directly `git revert` the commit without altering the package.
+    - use as commit message title: `vscode-extensions.publisher.extension_name: add myusername as maintainer`)
```
katexochen, line 27: Should those two points also apply when init a new extension?

superherointj: > Should those two points also apply when init a new extension?

Seems ideal since `git revert` won't work. And package may be tainted by transient maintainers.

katexochen: Just seems to be uncommon, as it isn't a requirement for other packages in nixpkgs to do that in a separate commit.

superherointj: > Just seems to be uncommon, as it isn't a requirement for other packages in nixpkgs to do that in a separate commit.

What do you want to be done? Drop those lines?

katexochen (approved):
lgtm, thanks!

Pandapip1 (approved):

PR #320554, opened 2024-06-17
deltachat-desktop: 1.44.1 -> 1.46.0 (fix build and outdated electron)

Mic92:
## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [x] Tested, as applicable:

---


pkgs/applications/networking/instant-messengers/deltachat-desktop/default.nix
```diff
@@ -76,10 +76,6 @@ buildNpmPackage rec {
     VERSION_INFO_GIT_REF = src.rev;
   };
 
-  preBuild = ''
-    rm -r node_modules/deltachat-node/node/prebuilds
```
dotlambda, line 80: Aren't there other prebuilds we should delete now? I think we have to use deltachat-rpc-server.

pkgs/applications/networking/instant-messengers/deltachat-cursed/default.nix
```diff
@@ -25,12 +25,12 @@ python3.pkgs.buildPythonApplication rec {
 
   propagatedBuildInputs = with python3.pkgs; [
```
dotlambda, line 26:
```suggestion
  dependencies = with python3.pkgs; [
```

pkgs/applications/networking/instant-messengers/deltachat-cursed/default.nix
```diff
@@ -7,15 +7,15 @@
...
 
   src = fetchFromGitHub {
     owner = "adbenitez";
     repo = "deltachat-cursed";
     rev = "v${version}";
-    hash = "sha256-1QNhNPa6ZKn0lGQXs/cmfdSFHscwlYwFC/2DpnMoHvY=";
+    hash = "sha256-z4JKe5soR4FdIn8hugxtnxQr/9V8m8a7QRzE1liIexc=";
   };
 
   nativeBuildInputs = with python3.pythonOnBuildForHost.pkgs; [
```
dotlambda, line 21:
```suggestion
  build-system = with python3.pythonOnBuildForHost.pkgs; [
```

dotlambda (changes requested):

pkgs/applications/networking/instant-messengers/deltachat-desktop/default.nix
Mic92: I couldn't find any.

SuperSandro2000 (commented):
otherwise LGTM

pkgs/development/libraries/libdeltachat/default.nix
```diff
@@ -31,17 +31,17 @@ let
...
     owner = "deltachat";
     repo = "deltachat-core-rust";
     rev = "v${version}";
-    hash = "sha256-/ZWpPpxnOCLGswrfbEPvfUn1LpdBQeR5LecRAB0PEhI=";
+    hash = "sha256-BSbvgKiI89B+nxp5McBKTJAwgePt27C1QvSQLhTL7pQ=";
   };
 
   patches = [
-    ./no-static-lib.patch
+    ./0001-no-stabic-lib.patch
```
emilylange, line 44: No need to prefix the patch with `0001-` here.
```suggestion
    ./no-stabic-lib.patch
```

But more importantly, you forgot to delete the outdated patch at `./no-static-lib.patch`.
If you had kept the filename as is, this would likely not have happened.

https://github.com/Mic92/nixpkgs/blob/b839e92bea2d19bea84720ef360ff4a2b9919fc4/pkgs/development/libraries/libdeltachat/no-static-lib.patch

pkgs/applications/networking/instant-messengers/deltachat-desktop/default.nix
```diff
@@ -1,7 +1,7 @@
 { lib
 , buildNpmPackage
 , copyDesktopItems
-, electron_28
+, electron_30
```
emilylange, line 4: Nit:
```suggestion
, electron
```

and override the version in `all-packages.nix`.

dotlambda: Why?

emilylange: https://github.com/NixOS/nixpkgs/blob/aa14af3e9f30695ea6bf62fb334c5044e477a735/pkgs/by-name/README.md#changing-implicit-attribute-defaults but applies to packages that aren't `by-name` as well.

pkgs/development/libraries/libdeltachat/default.nix
Mic92: I dropped the old name, but can we keep the prefix? This is compatible with `git format-patch`, which I used to generate this patch.

pkgs/applications/networking/instant-messengers/deltachat-desktop/default.nix
Mic92: Done. Also I find it less readable and I don't expect many people wanting to patch their electron in comparison.

pkgs/development/libraries/libdeltachat/default.nix
emilylange: I don't think we have policy for how to name patches.
Feel free to keep the prefix.

What made me comment was the fact that the old patch hasn't been deleted and that could have been prevented.

But you fixed that now, thanks. Fine with me.

PR #320564, opened 2024-06-17
nixos/cockpit: add listenAddress and listenPort

lcliu-gh:
## Description of changes

Add the `listenAddress` and `listenPort` options, this replaces
the old `port` option. Currently the cockpit module only allows
for configuring the port of the systemd socket. It may be desirable
for some to also change the address the socket listens on, e.g.
only listen on localhost. The default will still be to
listen on all interfaces.

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/nixos/modules/services/monitoring/cockpit.nix
+++ b/nixos/modules/services/monitoring/cockpit.nix
@@ -2,9 +2,12 @@
 
 let
   cfg = config.services.cockpit;
-  inherit (lib) types mkEnableOption mkOption mkIf literalMD mkPackageOption;
+  inherit (lib) types mkEnableOption mkOption mkIf literalMD mkPackageOption mkRenamedOptionModule concatStringsSep;
   settingsFormat = pkgs.formats.ini {};
 in {
+  imports = [
+    (mkRenamedOptionModule [ "services" "cockpit" "port" ] [ "services" "cockpit" "listenPort" ])
+  ];
   options = {
     services.cockpit = {
       enable = mkEnableOption "Cockpit";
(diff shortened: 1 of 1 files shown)


llkaze:
@lucasew

lucasew:
- Why not an option to set ListenStream directly? That could be set to `0.0.0.0:${cfg.port}`.

BTW for this to get merged you will need to add a NixOS test for that. It would be mostly a copypaste from the one already there. The way it is it will not work properly if changing the port because cockpit does a origin check and you need to add in the settings which origins are allowed.

llkaze:
Thank you for the feedback.

> * Why not an option to set ListenStream directly? That could be set to `0.0.0.0:${cfg.port}`.

Mostly because I looked around at other modules and the convention seemed to be to have listenAddress and port/ listenPort be separate options.

Also I'm not quite sure what you mean by setting it to `0.0.0.0:${cfg.port}`, would you have an option called listenStream where you set the entire string (like listenStream = "127.0.0.1:9090") and then still keep the port option so you would have to set it twice? Or do you want to only have a listenStream option and then extrapolate port to a variable to be used elsewhere e.g in allowedTCPPorts? IMO keeping them separate seems the easiest/ most straightforward while still allowing the user more fine-grained control.

I saw the test beforehand and I wasn't aware this would change things, I'll have a look again tomorrow but yeah it doesn't seem like I would need to change much about it. Thanks again, I'll update this soon.

lucasew (approved):

PR #320572, opened 2024-06-17
tests.haskell: prevent unnecessary rebuilds

philiptaron:
## Description of changes

Prevent unnecessary rebuilds when Nix code is reformatted.

- #301014 

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
- [x] Tested, as applicable:
  - [x] `nix-build -A tests.haskell.setBuildTarget`
  - [x] `nix-build -A tests.haskell.cabalSdist.localFromCabalSdist`

### DIFF
--- a/pkgs/test/haskell/cabalSdist/local/generated.nix
+++ b/pkgs/test/haskell/cabalSdist/local/generated.nix
@@ -3,7 +3,14 @@
 mkDerivation {
   pname = "local";
   version = "0.1.0.0";
-  src = ./.;
+  src = lib.fileset.toSource {
+    root = ./.;
+    fileset = lib.fileset.unions [
+      ./app
+      ./CHANGELOG.md
+      ./local.cabal
+    ];
+  };
   isLibrary = false;
   isExecutable = true;
   executableHaskellDepends = [ base ];
--- a/pkgs/test/haskell/setBuildTarget/default.nix
+++ b/pkgs/test/haskell/setBuildTarget/default.nix
@@ -7,7 +7,15 @@ let
       mkDerivation {
         pname = "haskell-setBuildTarget";
         version = "0.1.0.0";
-        src = ./.;
+        src = lib.fileset.toSource {
+          root = ./.;
+          fileset = lib.fileset.unions [
+            ./haskell-setBuildTarget.cabal
+            ./Bar.hs
+            ./Foo.hs
+            ./Setup.hs
+          ];
+        };
         isLibrary = false;
         isExecutable = true;
         executableHaskellDepends = [ base ];


cdepillabout (approved):
This seems reasonable.  Thanks

maralorn:
Hugh, I didn’t know this was a thing. Cool.

sternenseemann:
This change breaks the resulting tests, namely the assumption tests. Revert at #321246.

Serves to illustrate that a) all changes should be tested properly even if it appears that making them couldn't possibly break anything (in this case it is a little surprising, but ofborg correctly reports the other tests as changed!) and b) clearly, grepping for `src = ./.` is not enough, as the `./local` path would also contain `generated.nix`.

maralorn:
Sorry. but serious question: What's the correct workflow to catch this. I remember verifying that ofBorg built the tests correctly. Clearly that was not enough. My general stance was, that we will see build errors once they reach hydra, (which is a luxury we would have if we merge into haskell-updates, I should insist on that more)...

philiptaron:
Yeah, I'd love to learn. I _thought_ I did my due diligence (see the tests listed).

I see after investigation into the breadcrumbs you left in #321246 that the two tests are:

* `tests.haskell.cabalSdist.hercules-ci-cnix-store`
* `tests.haskell.cabalSdist.localHasNoDirectReference`

roberth:
> 1. What is `tests.haskell.cabalSdist.assumptionLocalHasDirectReference` actually testing?

A non-essential property of the way Haskell derivations are put together. If the assumption is broken, that suggests that the test may need to be updated, specifically because the more relevant test `localHasNoDirectReference` may succeed even if `buildFromCabalSdist` becomes a no-op by mistake.



> 1. The `builtins.unsafeDiscardOutputDependency` appears to be load bearing there, and there's no documentation of that function that I can find. What's going on there?!

It turns off the implicit behavior that Nix will build the outputs of a `.drv` if you refer to it like that. It's a performance optimization, and not really load bearing. It just makes it so that you're not substituting many unnecessary gigabytes.
Can't find the right phrasing right now, so an issue it is
- https://github.com/NixOS/nix/issues/10965



> 2\. Why adding the `default.nix` file back in caused the test to pass.

The test just treats `${./local}` as a constant. In any test, if you change a single constant literal to a different value, you can expect it to fail. Take for example `assert 1 == 1;`; change one of the integer literals, and then it fails.
This is so obvious that I guess you might feel offended. Of course it's not obvious in this case, because the literals are in different files, and they don't even look the same, so how could you know.
-  `./.` in `generated.nix`
- `./local` in the test

Both refer to the same thing. I just didn't communicate my intent.

roberth:
I've opened https://github.com/NixOS/nixpkgs/pull/322556, in case it helps.

philiptaron:
That explains it. Thanks, Robert.

PR #320607, opened 2024-06-17
nixos/redis: allow to move the state directory

JulesdeCube:
## Description of changes

The NixOS Redis module use the systemd `StateDirectory` directive to create the directory where redis will store it's data. this limit the placement of directory under `/var/lib`.

This PR:
- introduce the `dir` redis server option to allow to move the redis folder outside `/var/lib`.
- introduce a new oneshot systemd service to create the redis state directory when it's override by `dir` option.
- introduce the `dbFilename` redis server option to allow to modifed the previously hardcoded value.
- modify the redis test to validate the new `dir` redis server option.

## Things done

- Built on platform(s)
  - [X] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [X] `sandbox = true`
- [X] Tested, as applicable:

---

### DIFF
--- a/nixos/modules/services/databases/redis.nix
+++ b/nixos/modules/services/databases/redis.nix
@@ -10,15 +10,18 @@ let
     else if value == false then "no"
     else generators.mkValueStringDefault { } value;
 
-  redisConfig = settings: pkgs.writeText "redis.conf" (generators.toKeyValue {
-    listsAsDuplicateKeys = true;
-    mkKeyValue = generators.mkKeyValueDefault { inherit mkValueString; } " ";
-  } settings);
+  redisConfig = settings: pkgs.writeText "redis.conf" (generators.toKeyValue
+    {
+      listsAsDuplicateKeys = true;
+      mkKeyValue = generators.mkKeyValueDefault { inherit mkValueString; } " ";
+    }
+    settings);
 
-  redisName = name: "redis" + optionalString (name != "") ("-"+name);
+  redisName = name: "redis" + optionalString (name != "") ("-" + name);
   enabledServers = filterAttrs (name: conf: conf.enable) config.services.redis.servers;
 
-in {
+in
+{
   imports = [
     (mkRemovedOptionModule [ "services" "redis" "user" ] "The redis module now is hardcoded to the redis user.")
     (mkRemovedOptionModule [ "services" "redis" "dbpath" ] "The redis module now uses /var/lib/redis as data directory.")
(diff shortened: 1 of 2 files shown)


bjornfor:
> this limit the placement of directory under /var/lib.

What about using a bind mount on top of /var/lib/STATE_DIRECTORY? Doesn't that work and is simpler (works for all services without NixOS module changes)?

JulesdeCube:
> > this limit the placement of directory under /var/lib.
> 
> What about using a bind mount on top of /var/lib/STATE_DIRECTORY? Doesn't that work and is simpler (works for all services without NixOS module changes)?

You can´t use a bind mount because the `/var/lib/STATE_DIRECTORY` directory is created at run time by systemd `StateDirectory` directive therefore the bind mount will fail as the target and propably source directory (example `/data/redis`) didn´t exist. furthermore with a bind mount mounting you can´t specify permission and ownership.

bjornfor:
> You can´t use a bind mount because the `/var/lib/STATE_DIRECTORY` directory is created at run time by systemd `StateDirectory` directive therefore the bind mount will fail as the target and propably source directory (example `/data/redis`) didn´t exist. furthermore with a bind mount mounting you can´t specify permission and ownership.

Did you test it?

The `fileSystems` module creates both source and destination directories, and permissions can be configured with systemd-tmpfiles on the source directory:

```nix
  fileSystems = {
    "/var/lib/foo" = {
      device = "/data/foo";
      options = [ "bind" ];
    };
  };
  systemd.tmpfiles.rules = [
    # Type Path Mode User Group Age Argument
    "d /data/foo 0770 user1 group1 - -"
  ];
```

I didn't test this with systemd, but it [should work](https://github.com/systemd/systemd/issues/25097).

JulesdeCube:
excuse me for the delay,

I try your proposition this weekend and it work (no need for `systemd.tmpfiles.rules`).

It have try before with only the `mount.8` command that fail when the folder don´t exist.

`fileSystems` generate the `/etc/fstab` file that is used to by [`systemd-fstab-generator`](https://www.freedesktop.org/software/systemd/man/latest/systemd-fstab-generator.html) to generate the `systemd.mount` units.

According to the `systemd.mount` it create the mounting point if it didn´t exist (see [`Where=`](https://www.freedesktop.org/software/systemd/man/latest/systemd.mount.html#Where=) )

After that the redis service set the right permission at statup.

So excuse me for this useless PR and thank for the time you take to answer me.

PR #320836, opened 2024-06-18
nixos/bluemap: fix defaults issue with `services.bluemap.host`

philiptaron:
## Description of changes

The default for the `services.bluemap.host` value depends on `config.networking.domain`, which is typed as `types.nullOr types.str` in [nixos/modules/tasks/network-interfaces.nix](https://github.com/NixOS/nixpkgs/blob/master/nixos/modules/tasks/network-interfaces.nix#L609)

As a result, the default for `services.bluemap.host` either has to be `types.nullOr types.str`, or we need to drop the default.

Based on PR feedback, this commit drops the default and requires configuration through the `services.bluemap.host` option.

While this is a breaking change, since the module is [a month old](https://github.com/NixOS/nixpkgs/pull/312518), there should be very few users so far.

I discovered this while testing out https://github.com/NixOS/nixpkgs/pull/313497.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

### DIFF
--- a/nixos/modules/services/web-servers/bluemap.nix
+++ b/nixos/modules/services/web-servers/bluemap.nix
@@ -71,9 +71,7 @@ in {
 
     host = mkOption {
       type = lib.types.str;
-      default = "bluemap.${config.networking.domain}";
-      defaultText = lib.literalExpression "bluemap.\${config.networking.domain}";
-      description = "Domain to configure nginx for";
+      description = "Domain on which nginx will serve the bluemap webapp";
     };
 
     onCalendar = mkOption {


dali99:
Woops! Thanks for noticing!

A problem with this approach is that you now have to read the module source to find out that it's hosted on `bluemap.${domain}`. You could fix this by describing it in `defaultText` maybe?

A bigger issue is maybe that you can't now refer to `config.services.bluemap.host` to get the domain since this would be `null` - even though the nginx host is listening on `bluemap.${domain}`

Maybe we could just remove the default entirely?

philiptaron:
> Maybe we could just remove the default entirely?

I definitely started with that, but it would be a breaking change for users that were relying on it. If you're OK with that, I'll gladly update the PR.

pbsds:
perhaps retarget this to 24.05 and drop the default on unstable?

dali99:
I think there's no good solution here. If someone has this for example:

```nix
services.nginx.virtualHosts."${config.services.bluemap.host}" = {
  enableACME = true;
  forceSSL = true;
};
```

their nginx config will break at runtime completely silently in a probably hard to debug way.

I dont think either solution has a large impact so none of this is really important. The module is less than a month old, and has probably very very few users. But I'm erring towards just cleanly breaking at build time and ripping the band-aid off

philiptaron:
Since it's a month old, that makes my concern about existing users pretty minor. I'll simplify and remove the `bluemap.${domain}` default.

SuperSandro2000 (approved):
LGTM

nixos/modules/services/web-servers/bluemap.nix
```diff
@@ -70,9 +70,8 @@ in {
     };
 
     host = mkOption {
-      type = lib.types.str;
-      default = "bluemap.${config.networking.domain}";
-      defaultText = lib.literalExpression "bluemap.\${config.networking.domain}";
+      type = lib.types.nullOr lib.types.str;
+      default = null;
```
dali99, line 74:
```suggestion
      type = lib.types.str;
```

I'd be fine with just keeping it undefined and letting the module system complain if it's needed

nixos/modules/services/web-servers/bluemap.nix
```diff
@@ -267,6 +266,12 @@ in {
             You can achieve this through setting `services.bluemap.eula = true`
           '';
         }
+        { assertion = cfg.enableNginx -> cfg.host != null;
+          message = ''
+            You have enabled nginx to serve the bluemap webapp, but have not assigned a host
+            to bluemap through setting `services.bluemap.host = "hostname"`
```
dali99, line 272: Since `enableNginx` is `true` by default, maybe it'd better to say something like

```suggestion
            Nginx is enabled to serve the bluemap webapp, but you have not assigned a host
            to use. You can do this by setting `services.bluemap.host = "hostname"`.
```

Or something similar that doesn't imply the user did something

nixos/modules/services/web-servers/bluemap.nix
```diff
@@ -267,6 +266,12 @@ in {
...
+          '';
+        }
```
dali99, line 274: Alternatively if we just keep it undefined as I suggest in the option declaration we can just remove it entirely and keep things simple

```suggestion
```

D4ndellion (commented):
These are some completely non-blocking nitpicks!

nixos/modules/services/web-servers/bluemap.nix
SuperSandro2000: Yeah, good point. If null is not valid anyway, we should not allow in the type.

nixos/modules/services/web-servers/bluemap.nix
philiptaron: I liked this wording.

nixos/modules/services/web-servers/bluemap.nix
philiptaron: I dropped the default and kept it typed as str.

philiptaron:
Hold off on merging: I haven't actually built a VM with these settings set.

nixos/modules/services/web-servers/bluemap.nix
```diff
@@ -267,6 +265,12 @@ in {
             You can achieve this through setting `services.bluemap.eula = true`
           '';
         }
+        { assertion = cfg.enableNginx -> cfg.host != null;
+          message = ''
+            Nginx is enabled to serve the bluemap webapp, but you have not assigned a host
+            to use. You can do this by setting `services.bluemap.host = "hostname"`.
+          '';
+        }
```
dali99, line 273: I think unfortunately since the option now doesn't have a default it won't be null.

Instead it will  just not be present in the configuration. That means you can't have an assertion on this being null. You have to just trust the module system to complain that it is undefined (with a slightly worse error message), rather than having this manual assertion

```suggestion
```

D4ndellion (changes requested):

nixos/modules/services/web-servers/bluemap.nix
philiptaron: Right you are. I tested this out, and the error that results when your suggestion is applied is this:

```
error:
       … while calling the 'head' builtin
         at /nix/store/9gsbvj93aqfffkslk9fsrw3af78s1kz7-source/lib/attrsets.nix:1575:11:
         1574|         || pred here (elemAt values 1) (head values) then
         1575|           head values
             |           ^
         1576|         else

       … while evaluating the attribute 'value'
         at /nix/store/9gsbvj93aqfffkslk9fsrw3af78s1kz7-source/lib/modules.nix:809:9:
          808|     in warnDeprecation opt //
          809|       { value = builtins.addErrorContext "while evaluating the option `${showOption loc}':" value;
             |         ^
          810|         inherit (res.defsFinal') highestPrio;

       … while evaluating the option `system.build.toplevel':

       … while evaluating definitions from `/nix/store/9gsbvj93aqfffkslk9fsrw3af78s1kz7-source/nixos/modules/system/activation/top-level.nix':

       … while evaluating the option `assertions':

       … while evaluating definitions from `/nix/store/9gsbvj93aqfffkslk9fsrw3af78s1kz7-source/nixos/modules/system/boot/systemd.nix':

       … while evaluating the option `systemd.services':

       … while evaluating definitions from `/nix/store/9gsbvj93aqfffkslk9fsrw3af78s1kz7-source/nixos/modules/security/acme':

       … while evaluating the option `security.acme.certs':

       … while evaluating definitions from `/nix/store/9gsbvj93aqfffkslk9fsrw3af78s1kz7-source/nixos/modules/services/web-servers/nginx/default.nix':

       … while evaluating the option `services.nginx.virtualHosts':

       … while evaluating definitions from `/nix/store/9gsbvj93aqfffkslk9fsrw3af78s1kz7-source/nixos/modules/services/web-servers/bluemap.nix':

       … while evaluating the option `services.bluemap.host':

       error: The option `services.bluemap.host' is used but not defined.
```

That's definitely good enough.

nixos/modules/services/web-servers/bluemap.nix
philiptaron: I ended up removing the assertion.

philiptaron:
Now it's a super-simple change. Thanks for the feedback; I learned a lot about how defaults (or lack thereof) work in the module system. It's nice that a variable which is used but not defined blows up with a good error.

PR #320852, opened 2024-06-18
Clean up cross bootstrapping

Ericson2314:
**Contains #320840, will recompute diff once that one is merged.**

This will just be one commit then.

## Description of changes

For a long time, we've had `crossLibcStdenv`, `*Cross` libc attributes, and `*bsdCross` pre-libc package sets. This was always bad because having "cross" things is "not declarative": the naming doesn't reflect what packages *need* but rather how we *provide* something. This is ugly, and creates needless friction between cross and native building.

Now, almost all of these `*Cross` attributes are gone: just these are kept:

- Glibc's and Musl's are kept, because those packages are widely used and I didn't want to risk changing the native builds of those at this time.

- generic `libcCross`, `theadsCross`, and friends, because these relate to the convolulted GCC bootstrap which still needs to be redone.

The BSD and obscure Linux or freestnanding libcs have conversely all been made to use a new `stdenvNoLibc`, which is like the old `crossLibcStdenv` except:

1. It usable for native and cross alike

2. It named according to what it *is* ("a standard environment without libc but with a C compiler"), rather than some non-compositional jargon ("the stdenv used for building libc when cross compiling", yuck).

I should have done this change long ago, but I was stymied because of "infinite recursions". The problem was that in too many cases we are overriding `stdenv` to *remove* things we don't need, and this risks cyles since those more minimal stdenvs are used to build things in the more maximal stdenvs.

The solution is to pass `stage.nix` `stdenvNoCC`, so we can override to *build up* rather than *tear down*. For now, the full `stdenv` is also passed, so I don't need to change the native bootstraps, but I can see this changing as we make things more uniform and clean those up.

Finally, the BSDs also had to be cleaned up, since they have a few pre-libc dependencies, demanding a systematic approach. I realized what @rhelmot did in 6120256 (specify what packages just need `stdenvNoLibc`) is definitely the right approach for this, and adjusted NetBSD and OpenBSD to likewise use it.
## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/os-specific/bsd/openbsd/pkgs/mkDerivation.nix
+++ b/pkgs/os-specific/bsd/openbsd/pkgs/mkDerivation.nix
@@ -2,6 +2,7 @@
   lib,
   stdenv,
   stdenvNoCC,
+  stdenvNoLibc,
   runCommand,
   rsync,
   source,
@@ -14,7 +15,13 @@
 lib.makeOverridable (
   attrs:
   let
-    stdenv' = if attrs.noCC or false then stdenvNoCC else stdenv;
+    stdenv' =
+      if attrs.noCC or false then
+        stdenvNoCC
+      else if attrs.noLibc or false then
+        stdenvNoLibc
+      else
+        stdenv;
   in
   stdenv'.mkDerivation (
     rec {
@@ -42,9 +49,6 @@ lib.makeOverridable (
       ];
 
       HOST_SH = stdenv'.shell;
-
-      # Since STRIP below is the flag
-      STRIPBIN = "${stdenv.cc.bintools.targetPrefix}strip";
 
       makeFlags = [
         "STRIP=-s" # flag to install, not command
@@ -81,6 +85,9 @@ lib.makeOverridable (
     // lib.optionalAttrs stdenv'.hasCC {
       # TODO should CC wrapper set this?
       CPP = "${stdenv'.cc.targetPrefix}cpp";
+
+      # Since STRIP below is the flag
+      STRIPBIN = "${stdenv'.cc.bintools.targetPrefix}strip";
     }
     // lib.optionalAttrs (attrs.headersOnly or false) {
       installPhase = "includesPhase";
--- a/pkgs/stdenv/cross/default.nix
+++ b/pkgs/stdenv/cross/default.nix
@@ -41,25 +41,43 @@ in lib.init bootStages ++ [
       if crossSystem.isStatic
       then buildPackages.stdenvAdapters.makeStatic
       else lib.id;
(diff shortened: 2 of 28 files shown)


pkgs/os-specific/bsd/openbsd/pkgs/mkDerivation.nix
```diff
@@ -81,6 +85,9 @@ lib.makeOverridable (
     // lib.optionalAttrs stdenv'.hasCC {
       # TODO should CC wrapper set this?
       CPP = "${stdenv'.cc.targetPrefix}cpp";
+
+      # Since STRIP below is the flag
```
rhelmot, line 89: see above

pkgs/stdenv/cross/default.nix
```diff
@@ -41,25 +41,43 @@ in lib.init bootStages ++ [
...
+      extraBuildInputs = [ ]; # Old ones run on wrong platform
+      allowedRequisites = null;
+
+      cc = null;
+      hasCC = false;
+
+      extraNativeBuildInputs = old.extraNativeBuildInputs
+        ++ lib.optionals
+             (hostPlatform.isLinux && !buildPlatform.isLinux)
+             [ buildPackages.patchelf ]
```
rhelmot, line 61: What's up with this? I know it's just being moved, but who's asking for patchelf on literally everything?

rhelmot (commented):
This is so cool!

pkgs/stdenv/cross/default.nix
Ericson2314: No idea!

Ericson2314:
In the interest of not waiting a long time for ofborg again, I will move those comments in a second PR.

pkgs/stdenv/cross/default.nix
rhelmot: I suppose [glass houses](https://github.com/rhelmot/nixpkgs/blob/a7cf88713e88a75074eaaf65c21692a2fc5542c1/pkgs/stdenv/freebsd/default.nix#L315) and so forth

vcunat:
@Ericson2314: how did you miss the ofBorg label that it rebuilds a stdenv?

vcunat:
```
51f1ecaa59a3 Clean up cross bootstrapping
```
This one rebuilds both *-darwin stdenvs.  Let me revert the PR; you can redo it later.

Ericson2314:
I am sorry @vcunat. I do not know how I missed that; there is no excuse.

PR #321076, opened 2024-06-19
yazi: clean up wrapper, add options and format 

eljamm:
## Description of changes

~~Add exiftool, which is needed for showing EXIF data when interactively opening images with `O`~~

- Remove redundant `withPackage` options
- Add `optionalDeps` for packages that provide additional features
- Add `extraPackages` for user-defined packages
- Format with nixfmt

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/by-name/ya/yazi-unwrapped/package.nix
+++ b/pkgs/by-name/ya/yazi-unwrapped/package.nix
@@ -44,7 +44,7 @@ rustPlatform.buildRustPackage rec {
     description = "Blazing fast terminal file manager written in Rust, based on async I/O";
     homepage = "https://github.com/sxyazi/yazi";
     license = lib.licenses.mit;
-    maintainers = with lib.maintainers; [ xyenon matthiasbeyer linsui ];
+    maintainers = with lib.maintainers; [ xyenon matthiasbeyer linsui eljamm ];
     mainProgram = "yazi";
   };
 }
--- a/pkgs/by-name/ya/yazi/package.nix
+++ b/pkgs/by-name/ya/yazi/package.nix
@@ -1,89 +1,93 @@
-{ lib
-, runCommand
-, makeWrapper
-, yazi-unwrapped
+{
+  lib,
+  formats,
+  runCommand,
+  makeWrapper,
 
-, withRuntimeDeps ? true
-, withFile ? true
-, file
-, withJq ? true
-, jq
-, withPoppler ? true
-, poppler_utils
-, withUnar ? true
-, unar
-, withFfmpegthumbnailer ? true
-, ffmpegthumbnailer
-, withFd ? true
-, fd
-, withRipgrep ? true
-, ripgrep
-, withFzf ? true
-, fzf
-, withZoxide ? true
-, zoxide
-, settings ? { }
-, formats
-, plugins ? { }
-, flavors ? { }
-, initLua ? null
+  extraPackages ? [ ],
+  optionalDeps ? [
+    jq
+    poppler_utils
+    unar
+    ffmpegthumbnailer
+    fd
+    ripgrep
+    fzf
+    zoxide
+  ],
+
+  # deps
+  file,
+  yazi-unwrapped,
+
+  # optional deps
+  jq,
+  poppler_utils,
+  unar,
+  ffmpegthumbnailer,
+  fd,
+  ripgrep,
+  fzf,
+  zoxide,
+
+  settings ? { },
+  plugins ? { },
+  flavors ? { },
+  initLua ? null,
 }:
 
 let
-  runtimePaths = with lib;
-    [ ]
-    ++ optional withFile file
-    ++ optional withJq jq
-    ++ optional withPoppler poppler_utils
-    ++ optional withUnar unar
-    ++ optional withFfmpegthumbnailer ffmpegthumbnailer
-    ++ optional withFd fd
-    ++ optional withRipgrep ripgrep
-    ++ optional withFzf fzf
-    ++ optional withZoxide zoxide;
+  runtimePaths = [ file ] ++ optionalDeps ++ extraPackages;
 
   settingsFormat = formats.toml { };
 
-  files = [ "yazi" "theme" "keymap" ];
+  files = [
+    "yazi"
+    "theme"
+    "keymap"
+  ];
 
-  configHome = if (settings == { } && initLua == null && plugins == { } && flavors == { }) then null else
-  runCommand "YAZI_CONFIG_HOME" { } ''
-    mkdir -p $out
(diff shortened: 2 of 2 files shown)


eljamm:
Result of `nixpkgs-review pr 321076` run on x86_64-linux [1](https://github.com/Mic92/nixpkgs-review)
<details>
  <ul>
    <li>yazi</li>
  </ul>
</details>

XYenon:
I don't think we should add exiftool to the wrapper. Unlike other tools which are used to provide yazi functions in rust or lua codes, the only usage of exiftool is in the [default opener config](https://github.com/sxyazi/yazi/blob/1f562031f433df53a065e5315bd10ad13e43bd2f/yazi-config/preset/yazi.toml#L44). And I don't think we should add tools only in config to the wrapper, because config can and usually be changed.

matthiasbeyer (approved):

eljamm:
@XYenon As a new user, I expect things to work out of the box without me having to configure anything. Also, looking at the history of that config file, `exiftool` has always been there, so I don't think it's likely to change in the future.

That being said, if you still don't want to add it, how about adding an `extraPackages` option, instead?

That way, users will easily be able to simply add the tools they need to the wrapper.

linsui:
What's the benefit to add packages to the wrapper instead of globally?

eljamm:
@linsui Well, the main benefit is making sure those packages are installed/deleted with the program and that it functions as it should out of the box.

For example, say that I want to try yazi but `unar` has not been added to the wrapper. This means I also need to install `unar` in order to preview archive files. As such, I need to remember to remove `unar` when I no longer want to use yazi, else I'd have a package I don't need in my PATH, which also won't be cleaned up by the GC.

Therefore, by adding the tools we need to the wrapper, we're effectively coupling them to the main program.

XYenon:
mpv, mediainfo, and even vscode are in the default openers, I don't think exiftool is special. Add a `extraPackages` is fine for me.

eljamm:
Alright, should I open a new PR or do I force push here?

XYenon (approved):

eljamm:
Tested the change with:

```console
$ nix-build -E 'with import ./. {}; (yazi.override { extraPackages = [ exiftool ]; })'
```

And everything works correctly.

linsui:
Maybe we should unify those options. E.g.

```
extraPackages ? [ file jq ... ]
```

eljamm:
That won't work because those packages are essential for running the program, while the `extraPackages` option is meant for users to add their own **extra** packages.

From the example above, having `extraPackages = [ exiftool ];` will replace just `[ file jq ... ]`.

That said, we already have all the tools grouped together in `runtimePaths`, both the default and user-defined ones.

linsui:
None of those packages except file is needed to run yazi.

> From the example above, having extraPackages = [ exiftool ]; will just replace [ file jq ... ].

This is what I want. What should a user do if they only need file?

```nix
yazi.override {
  withJq = false;
  withPoppler = false;
  withUnar = false;
  withFfmpegthumbnailer = false;
  withFd = false;
  withRipgrep = false;
  withFzf = false;
  withZoxide = false;
}
```

or

```nix
yazi.override { withRuntimeDeps = false; }
```

then install file globally. What a bad UX.

XYenon:
> What should a user do if they only need file?

I think few users really want this. Most users want a fully functional yazi.

linsui:
> Most users want a fully functional yazi.

They can get a full-featured yazi by default. When they want to override yazi they can keep what they want. The benefit is that we have less options.

eljamm:
Technically, we could control all optional deps using another option `enableOptional`, for example, which can disable them all at once, but this won't work if you also want to enable specific ones.

In that case, you'd need to pass them with `extraPackages`:

```nix
yazi.override {
  enableOptional = false;
  extraPackages = [ jq fd rg ];
}
```

The UX of `"all on, few off"` is better to me as it's more realistic to disable what you don't need, for example:

```nix
yazi.override {
  # I don't like PDF and Archive previews
  withPoppler = false;
  withUnar = false;
}
```

However, if you'd like to discuss this further, I believe we can do so in a new PR.

XYenon:
> They can get a full-featured yazi by default. When they want to override yazi they can keep what they want. The benefit is that we have less options.

But if a user want to remove, for example, unar, maybe because a build failure, they have to add all others to `extraPackages` manually.

teto:
I find @linsui 's suggestion simpler and more tractable from an implementation point of view. Having all those `withX` is a recipe for errors. Having a simple list makes for clearer code, easier overrriding and so on.

On the question of adding exif by default, I dont really mind but the more you add, the harder it becomes to maintain (you get tickets because the dependency failed while it is not very impactful) and the closure increases. I would like to use yazi on servers as well where exif files is not required but space is limited so I would favor not including it

eljamm:
Thanks @teto, you make some valid points there. I went ahead and implemented @linsui 's idea in https://github.com/eljamm/nixpkgs/commit/2fb8165db07cf58000847c59e94b70877e0d9cf4 and I gotta say that it does look cleaner and more readable.

I'm personally okay with this as it doesn't change much for me. How does everyone else feel about this?

XYenon (approved):

linsui:
How about putting file in optionalDeps and removing withRuntimeDeps?

eljamm:
According to the [yazi docs](https://yazi-rs.github.io/docs/installation/), file is not optional.

linsui:
Without file yazi can still run and browse files though previewer doesn't work. I feel that it doesn't worth adding withRuntimeDeps only for that. withRuntimeDeps becomes useless anyway.

eljamm:
I don't mind removing `withRuntimeDeps` if its usefulness is limited, but I prefer to keep `file` as a non-optional dependency since that's what upstream considers it.

eljamm:
Does it make sense to add `extraPackages` or `optionalDeps` to the yazi module or should the package just be overwritten?

Also, are these changes considered a breaking API change and thus can't be backported to the stable branch?

sxyazi:
> Without file yazi can still run and browse files though previewer doesn't work

It can also cause files to be unable to open - I think being able to open files is a baseline for a file manager, so it's worth mentioning here.

linsui:
> I don't mind removing withRuntimeDeps if its usefulness is limited, but I prefer to keep file as a non-optional dependency since that's what upstream considers it.

Sounds good to me.

XYenon (approved):

linsui (approved):

PR #305036, opened 2024-04-18
treewide: remove licenses.gpl2

Lurkki14:
## Description of changes

Removed `licenses.gpl2` in favor of `gpl2Only` and `gpl2Plus`, also in some cases the license was incorrectly GPL2.

### Method used

First, I dumped all attributes that had `meta.license = gpl2;`  like so:
```
nix-eval-jobs --meta --force-recurse pkgs/top-level/release.nix 2> /dev/null \
	| jq ' select(.meta.license.shortName=="gpl2") | .attrPath | del(.[-1]) | join(".")' \
    -r --unbuffered 2> /dev/null | tee gpl2attrs
```

I wrote a script to go through them afterwards:

```
while true; do
	attr=$(head -n 1 gpl2attrs)
	if [ -z $attr ]; then
		break
	fi

    echo "Checking sources for $attr"
    NIX_PATH=nixpkgs=. \
        nix-build  --keep-failed --expr "let pkgs = import ./. {}; in  pkgs.srcOnly pkgs.$attr" && rg "any later version" result
    checkPid=$!
    echo "nix-build for $attr done"

    attrFile=$(NIX_PATH=nixpkgs=. nix eval --impure --expr "with import <nixpkgs> {}; pkgs.$attr.meta.position" --raw \
        | sed 's/\:.*$//')

    # Check if file has only one match for 'gpl2'
    matchCount=$(rg -wc "gpl2" $attrFile)
    echo $attr is defined in $attrFile
    if [[ $matchCount == "1" ]]; then
        echo "Press 'o' for gpl2Only, 'p' for gpl2Plus; or just Enter to do nothing (edit manually):"
        read -r choice
        if [ $choice == "o" ]; then
            sed -i 's/gpl2/gpl2Only/g' $attrFile
        fi

        if [ $choice == "p" ]; then
            sed -i 's/gpl2/gpl2Plus/g' $attrFile
        fi
    else
        echo "Found $matchCount matches, edit manually, and Enter when done:"
        read -r vuota
    fi

    sed -i "/$attr/d" gpl2attrs
done
```

This got around 90% of the cases, afterwards I just grepped the repo for "gpl2" and went through them more manually.

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
[...]

PR #321125, opened 2024-06-19
brother-dcpl2550dw: init at 4.0.0-1

p4p4j0hn:
## Description of changes

This adds the printer driver for the Brother DCPL2550DW.
It combines the cups-wrapper and the driver in one file which allows
sharing common variables.

## Things done

Currently using these packages on my laptop.

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/by-name/br/brother-dcpl2550dw/package.nix
+++ b/pkgs/by-name/br/brother-dcpl2550dw/package.nix
@@ -1,0 +1,124 @@
+{
+  lib,
+  stdenv,
+  fetchurl,
+  dpkg,
+  autoPatchelfHook,
+  makeWrapper,
+  perl,
+  gnused,
+  ghostscript,
+  file,
+  coreutils,
+  gnugrep,
+  which,
+}:
+
+let
+
+  arches = [
+    "x86_64"
+    "i686"
+    "armv7l"
+  ];
+
+  runtimeDeps = [
+    ghostscript
+    file
+    gnused
+    gnugrep
+    coreutils
+    which
+  ];
+
+in
+stdenv.mkDerivation (finalAttrs: {
+  pname = "brother-dcpl2550dw";
+  version = "4.0.0-1";
+
+  src = fetchurl {
+    url = "https://download.brother.com/welcome/dlf103577/dcpl2550dwpdrv-${finalAttrs.version}.i386.deb";
+    hash = "sha256-gGW8KDXlmYSMxvI29pfXd/lusg4rd4HaXiGkbBnUuQY=";
+  };
+
+  nativeBuildInputs = [
+    dpkg
+    makeWrapper
+    autoPatchelfHook
+  ];
+  buildInputs = [ perl ];
+
+  patches = [
+    # The brother lpdwrapper uses a temporary file to convey the printer settings.
+    # The original settings file will be copied with "400" permissions and the "brprintconflsr3"
+    # binary cannot alter the temporary file later on. This fixes the permissions so the can be modified.
+    # Since this is all in briefly in the temporary directory of systemd-cups and not accessible by others,
+    # it shouldn't be a security concern.
+    ./fix-perm.patch
+  ];
+
+  installPhase =
+    ''
+      runHook preInstall
+      mkdir -p $out
+      cp -ar opt $out/opt
+      # delete unnecessary files for the current architecture
+    ''
+    + lib.concatMapStrings (arch: ''
+      echo Deleting files for ${arch}
+      rm -r "$out/opt/brother/Printers/DCPL2550DW/lpd/${arch}"
+    '') (builtins.filter (arch: arch != stdenv.hostPlatform.linuxArch) arches)
+    + ''
+      # bundled scripts don't understand the arch subdirectories for some reason
+      ln -s \
+        "$out/opt/brother/Printers/DCPL2550DW/lpd/${stdenv.hostPlatform.linuxArch}/"* \
+        "$out/opt/brother/Printers/DCPL2550DW/lpd/"
+
+      # Fix global references and replace auto discovery mechanism with hardcoded values
+      substituteInPlace $out/opt/brother/Printers/DCPL2550DW/lpd/lpdfilter \
+        --replace-fail "my \$BR_PRT_PATH =" "my \$BR_PRT_PATH = \"$out/opt/brother/Printers/DCPL2550DW\"; #" \
+        --replace-fail "PRINTER =~" "PRINTER = \"DCPL2550DW\"; #"
+      substituteInPlace $out/opt/brother/Printers/DCPL2550DW/cupswrapper/lpdwrapper \
+        --replace-fail "my \$basedir = C" "my \$basedir = \"$out/opt/brother/Printers/DCPL2550DW\" ; #" \
+        --replace-fail "PRINTER =~" "PRINTER = \"DCPL2550DW\"; #"
+
+      # Make sure all executables have the necessary runtime dependencies available
+      find "$out" -executable -and -type f | while read file; do
+        wrapProgram "$file" --prefix PATH : "${lib.makeBinPath runtimeDeps}"
+      done
+
+      # Symlink filter and ppd into a location where CUPS will discover it
+      mkdir -p $out/lib/cups/filter
+      mkdir -p $out/share/cups/model
+      mkdir -p $out/etc/opt/brother/Printers/DCPL2550DW/inf
+
+      ln -s $out/opt/brother/Printers/DCPL2550DW/inf/brDCPL2550DWrc \
+        $out/etc/opt/brother/Printers/DCPL2550DW/inf/brDCPL2550DWrc
+      ln -s \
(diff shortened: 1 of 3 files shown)


p4p4j0hn:
Result of `nixpkgs-review pr 321125` run on x86_64-linux [1](https://github.com/Mic92/nixpkgs-review)
<details>
  <ul>
    <li>dcpl2550dw-cupswrapper</li>
    <li>dcpl2550dw-drv</li>
  </ul>
</details>

Yarny0:
#### Review per https://github.com/NixOS/nixpkgs/blob/b59dfac3c65d7ccc12426527553c9250cb71e90a/pkgs/README.md#new-packages

Thanks for your contribution to nixpkgs!

##### Reviewed points

- [x] package path fits guidelines: *seems ok if moving to `by-name` is not an option (see below)*
- [x] package name fits guidelines
- [x] package version fits guidelines
- [x] package builds on *x86-64*
- [ ] ~~executables tested on ARCHITECTURE~~ *(not tested)*
- [x] `meta.description` is set and fits guidelines
- [ ] `meta.license` fits upstream license: *no*
- [x] `meta.platforms` is set
- [ ] `meta.mainProgram` is set, if applicable.
- [x] build time only dependencies are declared in `nativeBuildInputs`
- [x] source is fetched using the appropriate function
- [x] the list of `phases` is not overridden
- [x] when a phase (like `installPhase`) is overridden it starts with `runHook preInstall` and ends with `runHook postInstall`.
- [x] patches have a comment describing either the upstream URL or a reason why the patch wasn't upstreamed
- [x] patches that are remotely available are fetched rather than vendored

##### Possible improvements

*   I suggest to resolve most of the constant values (like `model` and `MODEL`) as they probably never change within this derivation.  The few that might change (like `version`) can be defined inside the `mkDerviation` attribute set and then be referenced by a `finalAttrs` recursion (you can find examples for the `finalAttrs` pattern in other derivations in nixpkgs).

*   In the `driver` output, there is a broken symlink in `etc`.  That's probably a typo where this symlink is created.

*   The `gpl2` license is deprecated (some info here: https://github.com/NixOS/nixpkgs/pull/305036 ).  The `paperconfigml2` and `lpdwrapper` files in the source archive state `...either version 2 of the License, or (at your option) any later version`.  So it would be `gpl2Plus`.

*   The `installPhase` of the `cupswrapper` should be indented.

*   There are some message in the build log that might need attention:

    *   Can the error `dpkg-deb: error: --extract needs a target directory.` from the build log be resolved?  Maybe the `unpackCmd` isn't necessary at all, as the `dpkg` setuphook takes care of unpacking the archive.
    *   `'--replace' is deprecated, use --replace-{fail,warn,quiet}.` for `substituteInPlace`.

*    There is currently a movement to move away from `sha256="..."` hashes to `hash="sha256-..."` SRI hashes: https://github.com/NixOS/nixpkgs/issues/325892 .  It might be more future-proof to use an SRI hash.

##### Comments

*    Are you sure that both derivations are independent of each other?  The file `opt/brother/Printers/DCPL2550DW/cupswrapper/.lpdwrapper-wrapped` in the `cupswrapper` references `brprintconflsr3` and `lpdfilter` and apparently attempts to execute these files as binaries.  The `driver` drivation has executable files with these names, which makes me think that `cupswrapper` needs `driver`.
    In that case, the `lpdwrapper` should be patched so it can find those files, which also makes `driver` a runtime dependency of `cupswrapper`.  And with this constellation, and given that both derivations already share a lot of build code, it might be better to create one derivation.  This would shorten the build recipe a lot.  Then you should also move it to `by-name`.  If there is a realistic scenario where one would need the `driver`, but not the `cupswrapper`, this derivation could also be structured as a multi-output derivation.

*    I don't know how to use this package (I also don't have the hardware, but even then I probably wouldn't now).  I suggest to add some infos, either as comment in the build recipe file, or in `meta.longDescription`.

pkgs/by-name/br/brother-dcpl2550dw/package.nix
```diff
@@ -0,0 +1,104 @@
...
+      "$out/opt/brother/Printers/DCPL2550DW/lpd/${stdenv.hostPlatform.linuxArch}/"* \
+      "$out/opt/brother/Printers/DCPL2550DW/lpd/"
+
+    # Fix global references and replace auto discovery mechanism with hardcoded values
+    substituteInPlace $out/opt/brother/Printers/DCPL2550DW/lpd/lpdfilter \
+      --replace-quiet "my \$BR_PRT_PATH =" "my \$BR_PRT_PATH = \"$out/opt/brother/Printers/DCPL2550DW\"; #" \
+      --replace-quiet "PRINTER =~" "PRINTER = \"DCPL2550DW\"; #"
+    substituteInPlace $out/opt/brother/Printers/DCPL2550DW/cupswrapper/lpdwrapper \
+      --replace-quiet "my \$basedir = C" "my \$basedir = \"$out/opt/brother/Printers/DCPL2550DW\" ; #" \
+      --replace-quiet "PRINTER =~" "PRINTER = \"DCPL2550DW\"; #"
```
SuperSandro2000, line 74:
```suggestion
      --replace-fail "my \$BR_PRT_PATH =" "my \$BR_PRT_PATH = \"$out/opt/brother/Printers/DCPL2550DW\"; #" \
      --replace-fail "PRINTER =~" "PRINTER = \"DCPL2550DW\"; #"
    substituteInPlace $out/opt/brother/Printers/DCPL2550DW/cupswrapper/lpdwrapper \
      --replace-fail "my \$basedir = C" "my \$basedir = \"$out/opt/brother/Printers/DCPL2550DW\" ; #" \
      --replace-fail "PRINTER =~" "PRINTER = \"DCPL2550DW\"; #"
```
we want to notice if this no longer matches for whatever reason

pkgs/by-name/br/brother-dcpl2550dw/package.nix
```diff
@@ -0,0 +1,104 @@
...
+  nativeBuildInputs = [ dpkg makeWrapper autoPatchelfHook ];
+  buildInputs = [ perl ];
+
+  src = fetchurl {
+    url = "https://download.brother.com/welcome/dlf103577/dcpl2550dwpdrv-${version}.i386.deb";
+    sha256 = "01mrshcnr911bvd82xrb1sr6xybpsybzcdpjqs6896g56llbqrc0";
+  };
+
+  # unpackCmd = "dpkg-deb -x $src .";
+
```
SuperSandro2000, line 42:
```suggestion
```

pkgs/by-name/br/brother-dcpl2550dw/package.nix
```diff
@@ -0,0 +1,104 @@
...
+    which
+  ];
+in
+
+stdenv.mkDerivation rec {
+  pname = "brother-dcpl2550dw";
+  version = "4.0.0-1";
+
+  nativeBuildInputs = [ dpkg makeWrapper autoPatchelfHook ];
+  buildInputs = [ perl ];
```
SuperSandro2000, line 34: We normally place those below src

pkgs/by-name/br/brother-dcpl2550dw/package.nix
```diff
@@ -0,0 +1,104 @@
...
+, file
+, coreutils
+, gnugrep
+, which
+}:
+
+let
+  arches = [ "x86_64" "i686" "armv7l" ];
+
+  runtimeDeps = [
```
SuperSandro2000, line 19: We could just inline this

p4p4j0hn: Is there an example you can point me to? Thanks

Yarny0: I'm confident that means to just use the values of these variables in those places where use current use the variables.  This means the `let .. in` block  is no longer necessary.

Yarny0:
This looks really good now.  I tested it (not with real hardware, but by printing into a file), and the ppd file and the filter produce data without error.

pkgs/by-name/br/brother-dcpl2550dw/package.nix
p4p4j0hn: So do I move the variable declaration into the derivation? (without the let ... in part)

FliegendeWurst (approved):

pkgs/by-name/br/brother-dcpl2550dw/package.nix
```diff
@@ -0,0 +1,124 @@
...
+      ...
+        services.printing.enable = true;
+        services.printing.drivers = [
+          pkgs.brother-dcpl2550dw
+        ];
+      ...
+      ```
+    '';
+    license = licenses.unfree;
+    platforms = builtins.map (arch: "${arch}-linux") arches;
```
SuperSandro2000, line 121:
```suggestion
    platforms = map (arch: "${arch}-linux") arches;
```

Issue #213177, opened 2023-01-28
cloudflare-warp isn't working


ewwhash:
### Describe the bug
cloudflare-warp client is not working, neither wgcf is. It works in docker vm, so it's clearly a problem with managing DNS/nixos-fw. 
filestat /etc/resolv.conf is not changed at all, but on other distros it does. 
CloudflareWarp adapter isn't exist, and warp-svc says that network is unreachable. However, the RUST_BACKTRACE=debug hasn't enough information to properly identify the problem
ip route is still points to default gateway
iptables -vnL has 0.0.0.0/0 in every chain, but in other distros it creates firewall rules 

### Steps To Reproduce

1. /etc/nixos/configuration.nix
networking.nameservers = [ "1.1.1.1" "1.1.0.1" ];
services.resolved.enable = false;
networking.networkmanager.dns = "none";
environment.systemPackages = with pkgs; [cloudflare-warp nft]
2. sudo warp-svc && warp-cli register && warp-cli connect
3. warp-svc points that network is unreachable 

### Expected behavior
warp-svc should change the DNS and ip route, CloudflareWarp tun adapter should exist,

### Notify maintainers
@WolfangAukang

### Metadata
 - system: `"x86_64-linux"`
 - host os: `Linux 5.15.88, NixOS, 23.05 (Stoat), 23.05pre444846.d7705c01ef0`
 - multi-user?: `yes`
 - sandbox: `yes`
 - version: `nix-env (Nix) 2.12.0`
 - channels(chebuya): `"nixos"`
 - channels(root): `"nixos"`
 - nixpkgs: `/nix/var/nix/profiles/per-user/root/channels/nixos`

drishal:
exactly the same issue here, used to work before with just using 
```nix
 environment.systemPackages = with pkgs; [cloudflare-warp];
 systemd.packages = with pkgs; [ cloudflare-warp ];
  ```
  
  Now the systemd service no longer works

WolfangAukang:
@drishal Can you test it with [this module](https://github.com/NixOS/nixpkgs/pull/168092#issuecomment-1432710743)? I have it working just fine, or at least I am getting `warp=on` on my device.
@BrightYC To be honest, I never used these options you indicate together with Cloudflare Warp, so I have no idea on how to proceed here. If you have any recommendations, I can gladly help.

jeslinmx:
Any chance you could upload your `journalctl -u warp-svc`?

drishal:
[...]

PR #168092, opened 2022-04-10
nixos/cloudflare-warp: init

WolfangAukang:
###### Description of changes

Creating module based on #148395

###### Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
- [ ] For non-Linux: Is `sandbox = true` set in `nix.conf`? (See [Nix manual](https://nixos.org/manual/nix/stable/command-ref/conf-file.html))
- [x] Tested, as applicable:
  - [NixOS test(s)](https://nixos.org/manual/nixos/unstable/index.html#sec-nixos-tests) (look inside [nixos/tests](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests))
  - and/or [package tests](https://nixos.org/manual/nixpkgs/unstable/#sec-package-tests)
  - or, for functions and "core" functionality, tests in [lib/tests](https://github.com/NixOS/nixpkgs/blob/master/lib/tests) or [pkgs/test](https://github.com/NixOS/nixpkgs/blob/master/pkgs/test)
  - made sure NixOS tests are [linked](https://nixos.org/manual/nixpkgs/unstable/#ssec-nixos-tests-linking) to the relevant packages
- [ ] Tested compilation of all packages that depend on this change using `nix-shell -p nixpkgs-review --run "nixpkgs-review rev HEAD"`. Note: all changes have to be committed, also see [nixpkgs-review usage](https://github.com/Mic92/nixpkgs-review#usage)
- [x] Tested basic functionality of all binary files (usually in `./result/bin/`)
- [22.05 Release Notes (or backporting 21.11 Release notes)](https://github.com/NixOS/nixpkgs/blob/master/CONTRIBUTING.md#generating-2205-release-notes)
  - [ ] (Package updates) Added a release notes entry if the change is major or breaking
  - [ ] (Module updates) Added a release notes entry if the change is significant
  - [ ] (Module addition) Added a release notes entry if adding a new NixOS module
  - [ ] (Release notes changes) Ran `nixos/doc/manual/md-to-db.sh` to update generated release notes
- [x] Fits [CONTRIBUTING.md](https://github.com/NixOS/nixpkgs/blob/master/CONTRIBUTING.md).

PR #321142, opened 2024-06-20
nixos/cloudflare-warp: init module

treyfortmuller:
## Description of changes

Cloudflare warp is a VPN service which uses Cloudflare's network built on WireGuard. Warp has free plans and there's a thing called WARP+ (which I haven't used personally) which is a paid subscription. The derivations for the warp tooling are already in `nixpkgs` as `pkgs.cloudflare-warp`. 

Here's the homepage and docs/downloads: 
* https://developers.cloudflare.com/cloudflare-one/connections/connect-devices/warp/download-warp/
* https://pkg.cloudflareclient.com/

This is my first nixpkgs PR so looking for feedback! Especially around tests for a service like this, a basic `nixosTest` which enables the service and checks that its up would be easy to do, but I imagine we wouldn't want tests actually attempting to register as VPN user.

Some background:
* looks like there was an attempt to get a module in for this service before: https://github.com/NixOS/nixpkgs/pull/168092 this fizzled and was closed after some back and forth on the PR
* an open issue with a request for this module: https://github.com/NixOS/nixpkgs/issues/213177

I've been using this module for a few weeks without issue at work so I'm confident in the basic functionality. I attempted some `DynamicUser` configuration as a hardening step but the service reads/writes to `/etc/resolv.conf` so I've only been able to make this service work as intended while running as root.

I tested this manually by building a VM of my workstation off of my fork of nixpkgs and enabling `services.cloudflare-warp`. `warp-cli` is added to `systemPackages` and I used that to verify registration flows:

```
[nixosvmtest@kearsarge:~]$ warp-cli registration show
NOTICE:

Cloudflare only collects limited DNS query and traffic data (excluding payload)
that is sent to our network when you have the app enabled on your device. We
will not sell, rent, share, or otherwise disclose your personal information to
anyone, except as otherwise described in this Policy, without first providing
you with notice and the opportunity to consent. All information is handled in
accordance with our Privacy Policy.

More information is available at:
- https://www.cloudflare.com/application/terms/
- https://www.cloudflare.com/application/privacypolicy/

Accept Terms of Service and Privacy Policy? [y/N] y

Account type: Free
Device ID: 2914ecbd-ef6f-4db8-8397-d50453606e6a
Public key: beea6273fa27f78eb68ea092fc6aac776b1d5ea3aa06cf859587635dd442a62e
Account ID: 72a780f901b7416ca06bb2894e9086c5
License: X4a65S8U-60we45No-t1f89B0x

[nixosvmtest@kearsarge:~]$ warp-cli status
Status update: Connected
Success

[nixosvmtest@kearsarge:~]$ systemctl status cloudflare-warp.service
● cloudflare-warp.service - Cloudflare Zero Trust Client Daemon
     Loaded: loaded (/etc/systemd/system/cloudflare-warp.service; enabled; preset: enabled)
     Active: active (running) since Thu 2024-06-20 01:25:59 BST; 1min 56s ago
   Main PID: 692 (.warp-svc-wrapp)
         IP: 91.3K in, 27.1K out
         IO: 48.0K read, 140.0K written
      Tasks: 18 (limit: 2338)
     Memory: 61.9M (peak: 65.9M)
        CPU: 311ms
     CGroup: /system.slice/cloudflare-warp.service
             └─692 /nix/store/x6f7y11maam9gx6lx35anfx7knhg64ap-cloudflare-warp-2024.4.133/bin/warp-svc
```

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [x] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/nixos/modules/services/networking/cloudflare-warp.nix
+++ b/nixos/modules/services/networking/cloudflare-warp.nix
@@ -1,0 +1,91 @@
+{ config, lib, pkgs, ... }:
+let
+  cfg = config.services.cloudflare-warp;
+in
+{
+  options.services.cloudflare-warp = {
+    enable = lib.mkEnableOption "Cloudflare Zero Trust client daemon";
+
+    package = lib.mkPackageOption pkgs "cloudflare-warp" { };
+
+    rootDir = lib.mkOption {
+      type = lib.types.str;
+      default = "/var/lib/cloudflare-warp";
+      description = ''
+        Working directory for the warp-svc daemon.
+      '';
+    };
+
+    udpPort = lib.mkOption {
+      type = lib.types.port;
+      default = 2408;
+      description = ''
+        The UDP port to open in the firewall. Warp uses port 2408 by default, but fallback ports can be used
+        if that conflicts with another service. See the [firewall documentation](https://developers.cloudflare.com/cloudflare-one/connections/connect-devices/warp/deployment/firewall#warp-udp-ports)
+        for the pre-configured available fallback ports.
+      '';
+    };
+
+    openFirewall = lib.mkEnableOption "opening UDP ports in the firewall" // {
+      default = true;
+    };
+  };
+
+  config = lib.mkIf cfg.enable {
+    environment.systemPackages = [ cfg.package ];
+
+    networking.firewall = lib.mkIf cfg.openFirewall {
+      allowedUDPPorts = [ cfg.udpPort ];
+    };
+
+    systemd.tmpfiles.rules = [
+      "d ${cfg.rootDir}    - root root"
+      "z ${cfg.rootDir}    - root root"
+    ];
+
+    systemd.services.cloudflare-warp = {
+      enable = true;
+      description = "Cloudflare Zero Trust Client Daemon";
+
+      # lsof is used by the service to determine which UDP port to bind to
+      # in the case that it detects collisions.
+      path = [ pkgs.lsof ];
+      requires = [ "network.target" ];
+      wantedBy = [ "multi-user.target" ];
+
+      serviceConfig =
+        let
+          caps = [
+            "CAP_NET_ADMIN"
+            "CAP_NET_BIND_SERVICE"
+            "CAP_SYS_PTRACE"
+          ];
+        in
+        {
+          Type = "simple";
+          ExecStart = "${cfg.package}/bin/warp-svc";
+          ReadWritePaths = [ "${cfg.rootDir}" "/etc/resolv.conf" ];
+          CapabilityBoundingSet = caps;
+          AmbientCapabilities = caps;
+          Restart = "always";
+          RestartSec = 5;
+          Environment = [ "RUST_BACKTRACE=full" ];
+          WorkingDirectory = cfg.rootDir;
+
+          # See the systemd.exec docs for the canonicalized paths, the service
+          # makes use of them for logging, and account state info tracking.
+          # https://www.freedesktop.org/software/systemd/man/latest/systemd.exec.html#RuntimeDirectory=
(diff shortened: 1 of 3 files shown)


nixos/modules/services/networking/cloudflare-warp.nix
```diff
@@ -0,0 +1,96 @@
...
+      '';
+    };
+
+    openFirewall = mkOption {
+      type = types.bool;
+      default = true;
+      description = ''
+        Open the `udpPort` port in the firewall.
+      '';
+    };
```
treyfortmuller, line 36: Probably against best-practice to open ports on the firewall by default. Another option might be to leave this default-false, and then add an eval-time warning against none of the fallback ports being allowed in the firewall config.

From the [docs](https://developers.cloudflare.com/cloudflare-one/connections/connect-devices/warp/deployment/firewall#warp-udp-ports):

> WARP utilizes UDP for all of its communications. By default, the UDP port required for WARP is UDP 2408. WARP can fallback to UDP 500, UDP 1701, or UDP 4500.

MaxwellDupre:
Will have a better chance of being merged if commits squashed down to two.

treyfortmuller:
> Will have a better chance of being merged if commits squashed down to two.

Done! Lmk if these aren't the two commits you had in mind, I think this is right based on the contrib guidelines.

nixos/modules/services/networking/cloudflare-warp.nix
```diff
@@ -0,0 +1,96 @@
...
+      type = types.port;
+      default = 2408;
+      description = ''
+        The UDP port to open in the firewall. Warp uses port 2408 by default, but fallback ports can be used
+        if that conflicts with another service. See the [firewall documentation](https://developers.cloudflare.com/cloudflare-one/connections/connect-devices/warp/deployment/firewall#warp-udp-ports)
+        for the pre-configured available fallback ports.
+      '';
+    };
+
+    openFirewall = mkOption {
```
JohnRTitor, line 30: This could use `mkEnableOption` I believe. Additionally, it is discouraged to use `with lib;` in the whole module, but rather required functions/components should be called explicitly instead. like `lib.mkOption`.

JohnRTitor (dismissed):

nixos/modules/services/networking/cloudflare-warp.nix
```diff
@@ -0,0 +1,96 @@
...
+in
+{
+  options.services.cloudflare-warp = {
+    enable = mkEnableOption "Cloudflare Zero Trust client daemon";
+
+    package = mkPackageOption pkgs "cloudflare-warp" { };
+
+    rootDir = mkOption {
+      type = types.str;
+      default = "/var/cloudflare-warp";
```
liff, line 14: I think `/var/lib/cloudflare-warp` would be more conventional.

liff (commented):
> Added a release notes entry if adding a new NixOS module

Given that this is a new NixOS module, should come with release notes entry.

nixos/modules/services/networking/cloudflare-warp.nix
JohnRTitor: Sometimes you need to keep things enabled by default so it works out of the box. For example [KDE connect also does it](https://github.com/NixOS/nixpkgs/blob/nixos-unstable/nixos/modules/programs/kdeconnect.nix).

JohnRTitor (approved):
LGTM

treyfortmuller:
> I'll take this over, if you don't mind.

Thanks for the assist John!

PR #321150, opened 2024-06-20
libpsl: make python both a build and runtime dependency

rhelmot:
## Description of changes

Previously, python3 was only provided as a buildInput. However, libpsl.bin distributes python scripts. To make them use the right python, provide python as a nativeBuildInput and specify which python to use for the build with a configure flag.

This rebuilds curl.dev. I'm not sure how far-reaching the consequences of that are, but it at least includes rebuilding rustc, so I'm targeting staging.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [x] x86_64-freebsd
  - [ ] `sandbox = relaxed`
  - [x] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/development/libraries/libpsl/default.nix
+++ b/pkgs/development/libraries/libpsl/default.nix
@@ -10,6 +10,7 @@
 , libxslt
 , pkg-config
 , python3
+, buildPackages
 , valgrind
 , publicsuffix-list
 }:
@@ -33,7 +34,6 @@ stdenv.mkDerivation rec {
     gtk-doc
     lzip
     pkg-config
-    python3
     libxslt
   ];
 
@@ -41,6 +41,7 @@ stdenv.mkDerivation rec {
     libidn2
     libunistring
     libxslt
+    python3
   ];
 
   propagatedBuildInputs = [
@@ -61,6 +62,7 @@ stdenv.mkDerivation rec {
     "--with-psl-distfile=${publicsuffix-list}/share/publicsuffix/public_suffix_list.dat"
     "--with-psl-file=${publicsuffix-list}/share/publicsuffix/public_suffix_list.dat"
     "--with-psl-testfile=${publicsuffix-list}/share/publicsuffix/test_psl.txt"
+    "PYTHON=${lib.getExe buildPackages.python3}"
   ];
 
   enableParallelBuilding = true;


pkgs/development/libraries/libpsl/default.nix
```diff
@@ -61,6 +62,7 @@ stdenv.mkDerivation rec {
     "--with-psl-distfile=${publicsuffix-list}/share/publicsuffix/public_suffix_list.dat"
     "--with-psl-file=${publicsuffix-list}/share/publicsuffix/public_suffix_list.dat"
     "--with-psl-testfile=${publicsuffix-list}/share/publicsuffix/test_psl.txt"
+    "PYTHON=${lib.getDev buildPackages.python3}/bin/python3"
```
kirillrdy, line 65: Why getDev,  dev output usually doesn't have bin

rhelmot: Oh dear, now THAT is a typo. I have no idea how this worked in my testing...

pkgs/development/libraries/libpsl/default.nix
```diff
@@ -61,6 +62,7 @@ stdenv.mkDerivation rec {
     "--with-psl-distfile=${publicsuffix-list}/share/publicsuffix/public_suffix_list.dat"
     "--with-psl-file=${publicsuffix-list}/share/publicsuffix/public_suffix_list.dat"
     "--with-psl-testfile=${publicsuffix-list}/share/publicsuffix/test_psl.txt"
+    "PYTHON=${lib.getBin buildPackages.python3}/bin/python3"
```
kirillrdy, line 65:
```suggestion
    "PYTHON=${lib.getExe buildPackages.python3}"
```

rhelmot: Fixed

Ericson2314:
Description seems backwards? Also do we still need the one runnable at build time for cross?

rhelmot:
You're right, the description is backwards. I'll fix it in a moment.

I recall attempting to switch python to just being a buildInput, but it failed, and I was in fact testing linux -> freebsd cross while working on this.

Ericson2314 (approved):

vcunat:
@Ericson2314, @alyssais: this has consequences for netbsd cross.  `nix` now fails with
```
error: Failed to evaluate python3-3.11.9: «unsupported»: is not available on the requested hostPlatform:
```

I just noticed by accident that jobs fell of Hydra.  Easy eval test e.g.:
```bash
nix eval -f pkgs/top-level/release-cross.nix x86_64-netbsd.nix.x86_64-linux.outPath
```

alyssais:
Oh, thanks for noticing!

alyssais:
Python can't (or at least refuses to) cross compile for NetBSD, so maybe the thing to do where we can't cross compile Python is just to not install the Python scripts?  (Or leave them with unpatched shebangs, in case there's system Python?)

jcaesar:
This also broke `pkgsStatic.curl`. I assume that's used somewhere, given it [is actively maintained](https://github.com/NixOS/nixpkgs/pull/287515)?

(I understand it's not possible to make something a run-time dependency of just one output, but not a build-time dependency.)

emilazy:
Maybe the binaries should be a separate package altogether, as they’re probably relatively marginal?

PR #321236, opened 2024-06-20
logseq: fix build, move to by-name

superherointj:
logseq: fix build, move to by-name

CC @Aman9das @bclaud

Proper fix depends on upstream:
* https://github.com/logseq/logseq/issues/11378

### DIFF
--- a/pkgs/by-name/lo/logseq/package.nix
+++ b/pkgs/by-name/lo/logseq/package.nix
@@ -3,8 +3,10 @@
 , fetchurl
 , appimageTools
 , makeWrapper
-# graphs will not sync without matching upstream's major electron version
-, electron_27
+# Notice: graphs will not sync without matching upstream's major electron version
+#         the specific electron version is set at top-level file to preserve override interface.
+#         whenever updating this package also sync electron version at top-level file.
+, electron
 , autoPatchelfHook
 , git
 , nix-update-script
@@ -59,7 +61,7 @@ in {
 
   postFixup = ''
     # set the env "LOCAL_GIT_DIRECTORY" for dugite so that we can use the git in nixpkgs
-    makeWrapper ${electron_27}/bin/electron $out/bin/${pname} \
+    makeWrapper ${electron}/bin/electron $out/bin/${pname} \
       --set "LOCAL_GIT_DIRECTORY" ${git} \
       --add-flags $out/share/${pname}/resources/app \
       --add-flags "\''${NIXOS_OZONE_WL:+\''${WAYLAND_DISPLAY:+--ozone-platform-hint=auto --enable-features=WaylandWindowDecorations}}"
--- a/pkgs/top-level/all-packages.nix
+++ b/pkgs/top-level/all-packages.nix
@@ -3060,8 +3060,6 @@ with pkgs;
 
   lilo = callPackage ../tools/misc/lilo { };
 
-  logseq = callPackage ../applications/misc/logseq { };
-
   natls = callPackage ../tools/misc/natls { };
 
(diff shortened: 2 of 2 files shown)


pkgs/by-name/lo/logseq/package.nix
```diff
@@ -4,7 +4,7 @@
 , appimageTools
 , makeWrapper
 # graphs will not sync without matching upstream's major electron version
```
wegank, line 6: I guess from the comment, `electron` should be `electron_30` here?

superherointj: Electron 28?

https://github.com/logseq/logseq/blob/master/package.json#L116

wegank: Oh, yes then.

onny (approved):

bclaud:
Logseq is building and working. But electron 28 reached EOL.

For more information on this issue: https://github.com/logseq/logseq/issues/11378

Thanks @superherointj :1st_place_medal:

superherointj:
> But electron 28 reached EOL.

This has to be fixed at upstream. The fix here is the process of syncing version to upstream. Once upstream has it upgraded, a new PR can be made.

Also, electron 28 has to be deprecated here. Another issue.

superherointj:
> But electron 28 reached EOL.

This has to be fixed at upstream.

wegank:
Oh, that's too bad, we're not really fixing the build as both Electrons are EOL...

Electron 30 does seem preferable in this case, I think.

superherointj:
> Oh, that's too bad, we're not really fixing the build as both Electrons are EOL...
> Electron 30 does seem preferable in this case, I think.

But then, we're not fixing the process. And there's the issue of application not working properly still.
The fix has to come from upstream. We don't have a good fix here.
I'd keep format right and seek a fix at upstream.
Also, whoever wants to still use this package can add the security exception.

I'd favor keeping package right.

@wegank What do we do?

wegank (approved):

bclaud (approved):

RobFisher:
I appreciate the work you're doing, @superherointj . Could this explain also https://github.com/logseq/logseq/issues/10851 :
```
Error: The module '/nix/store/qiv0cgr5l5cl5j8x0wxl267519b2zw3v-logseq-0.10.9/share/logseq/resources/app/node_modules/better-sqlite3/build/Release/better_sqlite3.node'
was compiled against a different Node.js version using
NODE_MODULE_VERSION 118. This version of Node.js requires
NODE_MODULE_VERSION 119. Please try re-compiling or re-installing
the module (for instance, using `npm rebuild` or `npm install`).
```
?

In any case, my current workaround is:
```nix
  nixpkgs.config.permittedInsecurePackages = [
    "electron-28.3.3"
  ];
```

superherointj:
> Could this explain also [logseq/logseq#10851](https://github.com/logseq/logseq/issues/10851) :

> Error: The module '/nix/store/qiv0cgr5l5cl5j8x0wxl267519b2zw3v-logseq-0.10.9/share/logseq/resources/app/node_modules/better-sqlite3/build/Release/better_sqlite3.node'
> was compiled against a different Node.js version using
> NODE_MODULE_VERSION 118. This version of Node.js requires
> NODE_MODULE_VERSION 119.

I don't use `logseq`. So I haven't encountered this problem myself. But you can test to pin `nodejs` to 18:

At nixpkgs, change:

image

Then test it. If it works, feel free to propose a PR pinning nodejs to 18. And explain the reasoning at `nodejs` parameter at `logseq` package file.

PR #321281, opened 2024-06-20
xsimd: refactor

superherointj:
xsimd: refactor

- Move to by-name
- Add strictDeps
- Add update script
- Apply nixfmt
- Add meta.changelog
- Remove nested with
- Order attributes by phases order
- Order elements of attrs by alphabetical order

CC @tobim @doronbehar


doronbehar:
Please reopen against `staging`.

doronbehar:
You should still fix your mistake of merging https://github.com/NixOS/nixpkgs/pull/321253 by opening against `staging` a PR with the same changes as there. Feel free to also refactor if you want to.

superherointj:
> You should still fix your mistake of merging #321253 by opening against `staging` a PR with the same changes as there.

I have asked original author to re-target staging. I think, this sorts the situation then. (And keep credits right.)

PR #321282, opened 2024-06-20
Revert "xsimd: fix cross compilation"

doronbehar:
Reverts NixOS/nixpkgs#321253


doronbehar:
I merged before CI was green, to avoid triggering too many rebuilds on Hydra...

PR #321288, opened 2024-06-20
xsimd: fix cross compilation

superherointj:
xsimd: fix cross compilation, enable strictDeps
xsimd: move package to by-name
xsimd: add meta.changelog

Fixes #321253


doronbehar:
I'm down to setting `strictDeps = true;` BTW, if a good comment is written near it.

superherointj:
> I'm down to setting `strictDeps = true;` BTW, if a good comment is written near it.

Sorry. I don't know what to comment on `strictDeps`.
I usually think of `strictDeps` as something that should be ideally true.
If you have a suggestion, let me know.

doronbehar:
Maybe a comment such as:

```
# Setting strictDeps helps raise the chance that xsimd will be able to be cross compiled as well
```

But format it to 80 characters please :)

uninsane:
the first two commits (fixing cross compilation, then setting strictDeps = true) should be squashed into a single commit. otherwise, builds fine and looks good to me.

superherointj:
> the first two commits (fixing cross compilation, then setting strictDeps = true) should be squashed into a single commit.

The commits are separate to preserve credit.

uninsane:
> > the first two commits (fixing cross compilation, then setting strictDeps = true) should be squashed into a single commit.
> 
> The commits are separate to preserve credit.

i'd rather a clean commit history than to be credited, to be honest. you can squash the commits with you as commit author, and add to the commit message body `Co-Authored-By: Colin <colin@uninsane.org>`. or just leave yourself as commit author & don't mention me at all.

uninsane (approved):

tobim (approved):

doronbehar:
I like how the commits are ordered now - makes it easier to track changes IMO.

PR #321297, opened 2024-06-20
nixos/hardware/amdgpu: change default opencl runtime

Mic92:
Opencl is used for example in ffmpeg for video filters. Users might not be aware that they are missing drivers, so that seems like a reasonable default.

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/nixos/modules/services/hardware/amdgpu.nix
+++ b/nixos/modules/services/hardware/amdgpu.nix
@@ -10,11 +10,27 @@ in {
       series cards. Note: this removes support for analog video outputs,
       which is only available in the `radeon` driver
     '';
-    initrd.enable = lib.mkEnableOption ''
-      loading `amdgpu` kernelModule in stage 1.
-      Can fix lower resolution in boot screen during initramfs phase
-    '';
-    opencl.enable = lib.mkEnableOption ''OpenCL support using ROCM runtime library'';
+    enable = lib.mkEnableOption "AMD GPU";
+    initrd.enable = lib.mkOption {
+      description = ''
+        Enable loading `amdgpu` kernelModule in stage 1.
+        Can fix lower resolution in boot screen during initramfs phase
+      '';
+      type = lib.types.bool;
+      default = cfg.enable;
+      defaultText = "hardware.amdgpu.enable";
+    };
+    opencl.enable = lib.mkOption {
+      description = "Enable OpenCL support using Rusticl runtime library.";
+      type = lib.types.bool;
+      default = cfg.enable;
+      defaultText = "hardware.amdgpu.enable";
+    };
+    opencl.runtime = lib.mkOption {
+      description = "Enable OpenCL support using Rusticl runtime library.";
+      type = lib.types.enum [ "rusticl" "rocm" ];
+      default = "rusticl";
+    };
     # cfg.amdvlk option is defined in ./amdvlk.nix module
   };
 
@@ -30,7 +46,9 @@ in {
 
     hardware.opengl = lib.mkIf cfg.opencl.enable {
       enable = lib.mkDefault true;
-      extraPackages = [
+      extraPackages = if cfg.runtime == "rusticl" then [
+        pkgs.mesa.opencl
+      ] else [
         pkgs.rocmPackages.clr
         pkgs.rocmPackages.clr.icd
       ];


K900:
We should already have OpenCL via Rusticl, though I'm not sure how the performance compares. But we should definitely not subject people to the trainwreck that is ROCm by default.

Mic92:
> We should already have OpenCL via Rusticl, though I'm not sure how the performance compares. But we should definitely not subject people to the trainwreck that is ROCm by default.

Ok. I have not much insights on the state of amd gpus, but I am open to use a different opencl implementation by default.

K900:
Rusticl comes for free with Mesa.

Mic92:
Is it vendor neutral?

K900:
-ish. It has an allowlist of Gallium drivers it works on, but mostly due to bugs.

Mic92:
Otherwise, we could also change the defaults in nixos-hardware as well and not use rocm at all.

Mic92:
Is rusticl enabled by default, or can we enable it here?

K900:
It is enabled by default on _some_ hardware, I'll have to check the Mesa sources to know what exactly.

K900:
OK so, we currently don't actually get rusticl by default, because we don't have `mesa.opencl` in the default driver package set, but we should really add that instead.

nixos/modules/services/hardware/amdgpu.nix
```diff
@@ -30,10 +41,7 @@ in {
 
     hardware.opengl = lib.mkIf cfg.opencl.enable {
       enable = lib.mkDefault true;
-      extraPackages = [
-        pkgs.rocmPackages.clr
-        pkgs.rocmPackages.clr.icd
-      ];
+      extraPackages = [ pkgs.mesa.opencl ];
```
Mic92, line 44: Do we have any independent benchmarks for this claim? https://www.phoronix.com/news/Rusticl-RadeonSI-Near

K900: We should keep ROCm here as an option, I think, and add a global Mesa OpenCL option (that's on my list of yaks for this weekend probably). Also, which claim?

Mic92: That rusticl is faster than opencl. Or more generally speaking faster.

K900: I can do some quick tests to confirm faster, it's definitely better because ROCm is
- really shitty code overall (source: I've read it)
- comically large
- only officially supports RDNA3

JohnRTitor:
Haven't used `rusticl`, but as someone who used radv (Mesa's implementation of AMD's Vulkan) at some point, it is still experimental and does not support most games/softwares yet. Amdvlk is preferred, and I believe ROCM OpenCL would have far better support as well.

nixos/modules/services/hardware/amdgpu.nix
Mic92: It's an option again.

K900:
The WHAT. RADV is absolutely not experimental, and even AMD recommends you use it. It's also shipping on literally every Steam Deck.

Mic92:
> Haven't used `rusticl`, but as someone who used radv (Mesa's implementation of AMD's Vulkan) at some point, it is still experimental and does not support most games/softwares yet. Amdvlk is preferred, and I believe ROCM OpenCL would have far better support as well.

This goes far beyond my comfort zone. Do you mind taking this pull request over? I can close this.

Mic92:
> Haven't used `rusticl`, but as someone who used radv (Mesa's implementation of AMD's Vulkan) at some point, it is still experimental and does not support most games/softwares yet. Amdvlk is preferred, and I believe ROCM OpenCL would have far better support as well.

Is vulkan not orthogonal to opencl anyway? I don't quite see the conflict here.

nixos/modules/services/hardware/amdgpu.nix
```diff
@@ -10,11 +10,27 @@ in {
...
+    initrd.enable = lib.mkOption {
+      description = ''
+        Enable loading `amdgpu` kernelModule in stage 1.
+        Can fix lower resolution in boot screen during initramfs phase
+      '';
+      type = lib.types.bool;
+      default = cfg.enable;
+      defaultText = "hardware.amdgpu.enable";
+    };
+    opencl.enable = lib.mkOption {
```
Mic92, line 23: If we allow `runtime` to be null, than we don't need an enable option here.

JohnRTitor:
> The WHAT. RADV is absolutely not experimental, and even AMD recommends you use it. It's also shipping on literally every Steam Deck.

Ah, my knowledge is maybe outdated ish, ~but I did saw some benchmarks online where AMDVLK performed better long ago.~

> This goes far beyond my comfort zone. Do you mind taking this pull request over? I can close this. I don't own any AMD GPU at all.

I agree with the option of selecting a backend. Though I'll let K900 review this further and defer to their judgement.

nixos/modules/services/hardware/amdgpu.nix
```diff
@@ -10,11 +10,27 @@ in {
       series cards. Note: this removes support for analog video outputs,
       which is only available in the `radeon` driver
     '';
-    initrd.enable = lib.mkEnableOption ''
-      loading `amdgpu` kernelModule in stage 1.
-      Can fix lower resolution in boot screen during initramfs phase
-    '';
-    opencl.enable = lib.mkEnableOption ''OpenCL support using ROCM runtime library'';
+    enable = lib.mkEnableOption "AMD GPU";
```
Atemu, line 13: I don't see the point in this enable option. It's not at all obvious that it enables initrd and opencl. I think it'd be better to simply let users pick and choose what they like.

The existence of an enable option also implies that it's necessary to enable to have AMDGPU support which is not at all true; it merely enables some optional AMDGPU features.

PR #321302, opened 2024-06-20
`lib`: Inherit exported builtins from the respective sub-library

infinisil:
## Description of changes

`lib.*` exports a lot of builtins that are also available in sub-libraries, currently in two ways:
1. `lib.attrValues` -> `lib.attrsets.attrValues` -> `builtins.attrValues`
2. `lib.attrNames` -> `builtins.attrNames` and `lib.attrsets.attrNames` -> `builtins.attrNames`

This PR changes all that use 2. to use the approach from 1. instead, because:
- It's more consistent internally
- In case we ever change anything about the exported symbols, it won't lead to inconsistencies where `lib.attrNames` wouldn't be the same as `lib.attrsets.attrNames` anymore

Ping @roberth

## Things done

Tested that all `lib.*` symbols exist:

```
❯ nix repl -f .
Nix 2.22.1
Type :? for help.
Loading installable ''...
Added 21758 variables.
nix-repl> lib.debug.traceSeqN 1 lib null
...
null
```

---

### DIFF
--- a/lib/default.nix
+++ b/lib/default.nix
@@ -64,9 +64,9 @@ let
     # linux kernel configuration
     kernel = callLibs ./kernel.nix;
 
-    inherit (builtins) addErrorContext
-      isPath
-      trace;
+    # TODO: For consistency, all builtins should also be available from a sub-library;
+    # these are the only ones that are currently not
+    inherit (builtins) addErrorContext isPath trace;
     inherit (self.trivial) id const pipe concat or and xor bitAnd bitOr bitXor
       bitNot boolToString mergeAttrs flip mapNullable inNixShell isFloat min max
       importJSON importTOML warn warnIf warnIfNot throwIf throwIfNot checkListOfEnum
--- a/lib/default.nix
+++ b/lib/default.nix
@@ -65,7 +65,6 @@ let
     kernel = callLibs ./kernel.nix;
 
     inherit (builtins) addErrorContext
-      genericClosure
       isPath
       readFile replaceStrings
       trace;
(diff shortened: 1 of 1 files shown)


roberth:
Needs a rebase since #321302.
I agree with the change.

infinisil:
Rebased :)

lib/default.nix
```diff
@@ -64,37 +64,38 @@ let
...
     kernel = callLibs ./kernel.nix;
 
-    inherit (builtins) add addErrorContext attrNames concatLists
-      deepSeq elem elemAt filter genericClosure genList getAttr
-      hasAttr head isAttrs isBool isInt isList isPath isString length
-      lessThan listToAttrs pathExists readFile replaceStrings seq
-      stringLength sub substring tail trace;
+    # TODO: For consistency, all builtins should also be available from a sub-library;
+    # these are the only ones that are currently not
+    inherit (builtins) addErrorContext isPath trace;
```
roberth, line 69: Thought: 
`addErrorContext`, `trace`, `seq`, `warnIf` share a similarity that they return identity functions in terms of the returned value, adding side effects. Perhaps these should be in a sub-library together?
- `lib.nonfunctional` - they're not just functions and they tend to be about non-functional requirements
- `lib.identities` - meh, only half of the story
- `lib.sideEffects` or `lib.effectful` - maybe? technically `derivation` or `toFile` might also qualify though, if we don't equate the creation of content addressed immutable things with the construction of Nix language values, which in practice I think we all do.


Couldn't think of a great name. Maybe it's just a fairly diverse set of functions anyway.

roberth (approved):
Thanks!
The individual commits made this very easy to review.

PR #321305, opened 2024-06-20
nixos/nix: convert buildMachines to an attrset to allow overriding

r-vdp:
## Description of changes

I used `types.either` because submodules are not allowed in `types.coercedTo`. The other alternative is to use `apply`, but I think its use is discouraged.

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/applications/audio/amarok/default.nix
+++ b/pkgs/applications/audio/amarok/default.nix
@@ -8,11 +8,11 @@
 
 stdenv.mkDerivation (finalAttrs: {
   pname = "amarok";
-  version = "3.0.0";
+  version = "3.0.1";
 
   src = fetchurl {
     url = "mirror://kde/stable/amarok/${finalAttrs.version}/amarok-${finalAttrs.version}.tar.xz";
-    sha256 = "sha256-FKh2eDBfrXagodrKVVpndf+mQuXrvMzs2R9JcJOZLBw=";
+    sha256 = "sha256-S+zyrp3cvoYRA+ZCFDaireIvU4yxw2ANTmZ/cb6ljys=";
   };
 
   outputs = [ "out" "doc" ];


flokli:
Why is there a need for ordering?

The nix manual suggests `speedFactor` should work as a preference system, so I'm a bit confused about where the order does matter...

r-vdp:
I don't think there is a need for ordering, we don't have any explicit ordering when using an attrset, do we? Or maybe I'm missing something?

I want to be able to override though, that's why I'd like this to be an attrset instead of a list.

flokli:
> I don't think there is a need for ordering, we don't have any explicit ordering when using an attrset, do we? Or maybe I'm missing something?
> 
> I want to be able to override though, that's why I'd like this to be an attrset instead of a list.

Sorry, I initially misread.

Though I'm not convinced. The list of machines ultimately is a list for Nix(cpp), so exposing it as an attrset just for the sake of being able to set various aspects of individual builders from different places feels like a workaround for a lack of features in the module system.

I think the cleaner way to solve that usecase would be to write a module downstream, allowing you to have that attrset as config in your module, and then ultimately configuring the Nixpkgs one.

r-vdp:
> > I don't think there is a need for ordering, we don't have any explicit ordering when using an attrset, do we? Or maybe I'm missing something?
> > 
> > I want to be able to override though, that's why I'd like this to be an attrset instead of a list.
> 
> Sorry, I initially misread.
> 
> Though I'm not convinced. The list of machines ultimately is a list for Nix(cpp), so exposing it as an attrset just for the sake of being able to set various aspects of individual builders from different places feels like a workaround for a lack of features in the module system.
> 
> I think the cleaner way to solve that usecase would be to write a module downstream, allowing you to have that attrset as config in your module, and then ultimately configuring the Nixpkgs one.

Yeah, it's just sad to have to copy all the options into my own config and keep them in sync.

flokli:
> Yeah, it's just sad to have to copy all the options into my own config and keep them in sync.

I agree, but it's a bit of an obscure usecase, and requiring everyone to update their builder config just for this is a hard sell.

PR #321469, opened 2024-06-21
gnomeExtensions.systemd-manager: package manually

### LINKED ISSUE #279638: gnomeExtensions.systemd-manager outdated and crashes in Gnome 45.2 Wayland
### Describe the bug
After reboot, while systemd units are still loading, if user clicks on Systemd-Manger extension the entire DE will crash and Gnome will force a log out of that user.

Moreover, Systemd-Manager is at v14 when v15 is out from the developer.

### Steps To Reproduce
1. Restart Computer
2. After DE loads and user is logged in
3. Click on Systemd-Manager (while the systemd units [custom added systemd units by the user into the extension] are still loading for example "bitcoind")
4. DE crashes and forces log out

### Expected behavior
No crashing of DE and you should be able to see in Systemd-Manager the units that have not started without it crashing

### Notify maintainers

@piegamesde 

### Metadata

``` 
- system: `"x86_64-linux"`
 - host os: `Linux 6.6.10, NixOS, 24.05 (Uakari), 24.05.20240106.46ae021`
 - multi-user?: `yes`
 - sandbox: `yes`
 
```

---

linsui:
## Description of changes

Closes https://github.com/NixOS/nixpkgs/issues/279638
https://github.com/hardpixel/systemd-manager/issues/19

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/desktops/gnome/extensions/systemd-manager/default.nix
+++ b/pkgs/desktops/gnome/extensions/systemd-manager/default.nix
@@ -10,7 +10,11 @@
   systemd ? config.systemd.package,
 }:
 
-assert lib.elem allowPolkitPolicy [ "none" "pkexec" "systemctl" ];
+assert lib.elem allowPolkitPolicy [
+  "none"
+  "pkexec"
+  "systemctl"
+];
 
 stdenvNoCC.mkDerivation rec {
   pname = "gnome-shell-extension-systemd-manager";
@@ -27,22 +31,25 @@ stdenvNoCC.mkDerivation rec {
 
   nativeBuildInputs = [ glib ];
 
-  postInstall = ''
-    rm systemd-manager@hardpixel.eu/schemas/gschemas.compiled
-    glib-compile-schemas systemd-manager@hardpixel.eu/schemas
+  postInstall =
+    ''
+      rm systemd-manager@hardpixel.eu/schemas/gschemas.compiled
+      glib-compile-schemas systemd-manager@hardpixel.eu/schemas
 
-    mkdir -p $out/share/gnome-shell/extensions
-    mv systemd-manager@hardpixel.eu $out/share/gnome-shell/extensions
-  '' + lib.optionalString (allowPolkitPolicy == "pkexec") ''
-    local bn=org.freedesktop.policykit.pkexec.systemctl.policy
-    mkdir -p $out/share/polkit-1/actions
-    substitute systemd-policies/$bn $out/share/polkit-1/actions/$bn \
-      --replace-fail /usr/bin/systemctl ${lib.getBin systemd}/bin/systemctl
-  '' + lib.optionalString (allowPolkitPolicy == "systemctl") ''
-    install -Dm0644 \
-      systemd-policies/10-service_status.rules \
-      $out/share/polkit-1/rules.d/10-gnome-extension-systemd-manager.rules
-  '';
+      mkdir -p $out/share/gnome-shell/extensions
+      mv systemd-manager@hardpixel.eu $out/share/gnome-shell/extensions
+    ''
+    + lib.optionalString (allowPolkitPolicy == "pkexec") ''
+      local bn=org.freedesktop.policykit.pkexec.systemctl.policy
+      mkdir -p $out/share/polkit-1/actions
+      substitute systemd-policies/$bn $out/share/polkit-1/actions/$bn \
+        --replace-fail /usr/bin/systemctl ${lib.getBin systemd}/bin/systemctl
+    ''
+    + lib.optionalString (allowPolkitPolicy == "systemctl") ''
+      install -Dm0644 \
+        systemd-policies/10-service_status.rules \
+        $out/share/polkit-1/rules.d/10-gnome-extension-systemd-manager.rules
+    '';
 
   passthru = {
     extensionUuid = "systemd-manager@hardpixel.eu";
@@ -53,6 +60,9 @@ stdenvNoCC.mkDerivation rec {
     description = "GNOME Shell extension to manage systemd services";
     homepage = "https://github.com/hardpixel/systemd-manager";
     license = licenses.gpl3Only;
-    maintainers = with maintainers; [ linsui ];
+    maintainers = with maintainers; [
+      linsui
+      doronbehar
+    ];
   };
 }
--- a/pkgs/desktops/gnome/extensions/systemd-manager/default.nix
+++ b/pkgs/desktops/gnome/extensions/systemd-manager/default.nix
@@ -1,0 +1,42 @@
+{
+  lib,
+  stdenvNoCC,
+  fetchFromGitHub,
+  glib,
+}:
+
+stdenvNoCC.mkDerivation rec {
+  pname = "gnome-shell-extension-systemd-manager";
+  version = "16";
+
+  # Upstream doesn't post new versions in extensions.gnome.org anymore, see also:
+  # https://github.com/hardpixel/systemd-manager/issues/19
+  src = fetchFromGitHub {
+    owner = "hardpixel";
+    repo = "systemd-manager";
+    rev = "v${version}";
+    hash = "sha256-JecSIRj582jJWdrCQYBWFRkIhosxRhD3BxSAy8/0nVw=";
+  };
+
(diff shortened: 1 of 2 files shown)


pkgs/desktops/gnome/extensions/extensions.json
```diff
@@ -3479,15 +3479,6 @@
...
   }}
-, {"uuid": "systemd-manager@hardpixel.eu", "name": "Systemd Manager", "pname": "systemd-manager", "description": "Toggle systemd services on/off from a popup menu in the top gnome panel. Can be used to start services like apache2, mysql, postgres. It uses `pkexec' to run `sytemctl'. If you want to start services without entering a password you have to polkit policy file. An example policy file can be found in the github repository.", "link": "https://extensions.gnome.org/extension/4174/systemd-manager/", "shell_version_map": {
```
jtojnar, line 3482: This file is auto-generated and should not be edited?

pkgs/desktops/gnome/extensions/extensions.json
jtojnar, line 1: Please mention that as per https://extensions.gnome.org/extension/4174/systemd-manager/ is no longer uploaded to e.g.o in the commit message.

pkgs/desktops/gnome/extensions/systemd-manager/default.nix
```diff
@@ -0,0 +1,36 @@
...
+
+  src = fetchFromGitHub {
+    owner = "hardpixel";
+    repo = "systemd-manager";
+    rev = "v${version}";
+    hash = "sha256-JecSIRj582jJWdrCQYBWFRkIhosxRhD3BxSAy8/0nVw=";
+  };
+
+  installPhase = ''
+    mkdir -p $out/share/gnome-shell/extensions/
```
jtojnar, line 20: When overriding `installPhase`, you should also provide `runHook preInstall` and `runHook postInstall`. Though, in this case, it is better to rely on the default `installPhase`. See <https://github.com/jtojnar/nixpkgs-hammering/blob/master/explanations/missing-phase-hooks.md> for more info.

pkgs/desktops/gnome/extensions/systemd-manager/default.nix
```diff
@@ -0,0 +1,36 @@
...
+  stdenvNoCC,
+  fetchFromGitHub,
+  nix-update-script,
+}:
+
+stdenvNoCC.mkDerivation rec {
+  pname = "gnome-shell-extension-systemd-manager";
+  version = "16";
+
+  src = fetchFromGitHub {
```
jtojnar, line 12: [gschemas.compiled](https://github.com/hardpixel/systemd-manager/blob/master/systemd-manager%40hardpixel.eu/schemas/gschemas.compiled) is platform-specific so you need to remove it and recompile it using `glib-compile-schemas`.

pkgs/desktops/gnome/extensions/systemd-manager/default.nix
doronbehar, line 20: `runHook`...

pkgs/desktops/gnome/extensions/systemd-manager/default.nix
```diff
@@ -0,0 +1,36 @@
...
+    owner = "hardpixel";
+    repo = "systemd-manager";
+    rev = "v${version}";
+    hash = "sha256-JecSIRj582jJWdrCQYBWFRkIhosxRhD3BxSAy8/0nVw=";
+  };
+
+  installPhase = ''
+    mkdir -p $out/share/gnome-shell/extensions/
+    mv systemd-manager@hardpixel.eu $out/share/gnome-shell/extensions/
```
doronbehar, line 21: Maybe use `install`? Never fully understood why that is considered better, but I often see people preferring it. Also it makes the expression more concise.

pkgs/desktops/gnome/extensions/systemd-manager/default.nix
```diff
@@ -0,0 +1,36 @@
...
+  '';
```
jtojnar, line 22: There are also [some Polkit files](https://github.com/hardpixel/systemd-manager#authorization) but we probably do not want to install them to default location since they loosen security. Maybe copy them to `share/doc/systemd-manager` directory for easier enabling and mention them in `longDescription`. Though the policy also FHS path so that would need to be patched.

pkgs/desktops/gnome/extensions/systemd-manager/default.nix
doronbehar, line 12:
```suggestion
  # Upstream doesn't post new versions in extensions.gnome.org anymore, see also:
  # https://github.com/hardpixel/systemd-manager/issues/19
  src = fetchFromGitHub {
```

pkgs/desktops/gnome/extensions/extensions.json
jtojnar: I wonder if it is to bypass the review. I have skimmed the source and did not notice anything immediately suspicious.

pkgs/desktops/gnome/extensions/systemd-manager/default.nix
```diff
@@ -0,0 +1,36 @@
...
+
+  passthru = {
+    extensionUuid = "systemd-manager@hardpixel.eu";
+    extensionPortalSlug = "systemd-manager";
+    updateScript = nix-update-script { };
```
doronbehar, line 27: Is that really needed? I think both `nix-update` and `nixpkgs-update` will be able to update this package even without this - there is nothing non-trivial with the GitHub project - there are both releases and tags there.

pkgs/desktops/gnome/extensions/extensions.json
doronbehar: > I wonder if it is to bypass the review. I have skimmed the source and did not notice anything immediately suspicious.

You mean to say that perhaps the author of the extension is not uploading version to e.g.o to bypass the review of the gnome extensions website?

pkgs/desktops/gnome/extensions/systemd-manager/default.nix
linsui: It seems install can't do that. https://stackoverflow.com/questions/19136297/how-to-copy-directories-into-a-directory-using-install-in-bash

pkgs/desktops/gnome/extensions/systemd-manager/default.nix
doronbehar: > I add the rule to `security.polkit.extraConfig` directly. Would it be better to source that file from this package directly?

I read this in the polkit service:

```nix
    # The polkit daemon reads action/rule files
    environment.pathsToLink = [ "/share/polkit-1" ];
```

I think it means that if the package would provide files in that directory - it should work if it the polkit service is enabled and the package is `environment.systemPackage`.

doronbehar: Pushed a suggestion regarding the polkit policy.

linsui:
Is this ready?

doronbehar:
> Is this ready?

As [I said here](https://github.com/NixOS/nixpkgs/pull/321469#discussion_r1648793351), I pushed 2 commits that fix the polkit files dilemma. If you think these changes are good @linsui then yes I think we can merge this.

linsui:
Your changes are good to me. :)

linsui:
How can I override this package? I thought this makes manually packaged extensions not overridable.

https://github.com/NixOS/nixpkgs/blob/153864b4ef7cfaa89ec5386028aa3ac51089f07d/pkgs/desktops/gnome/extensions/default.nix#L75

Changing this to 

```nix
    (extensions: extensions // (import ./manuallyPackaged.nix { inherit callPackage; }))
```

works.

doronbehar:
You are right... Also `.overrideScope` is not available. What do you think about @linsui's suggestion @jtojnar ?

jtojnar:
I think @linsui's suggestion makes sense.

PR #321484, opened 2024-06-21
fetchFromGitHub: use the API tarball endpoint for `private`

### LINKED ISSUE #321481: Fine-grained GitHub personal access tokens cannot be used with pkgs.fetchFromGitHub
### Describe the bug
[GitHub has two types of access token](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/managing-your-personal-access-tokens). Fine-grained access tokens have a number of advantages over the classic tokens, notably that they can be configured to only have access to specific private repositories, but currently using them doesn't work with `pkgs.fetchFromGitHub`.

### Steps To Reproduce
1. Set up a private repository in GitHub that has at least one commit on the default branch.
2. Set up a fine-grained personal access token at https://github.com/settings/personal-access-tokens/new with:
    - Any name, description and expiration
    - "Only select repositories" access to the private repository set up above
    - Select "Read-only" access to every permission that has that option.
3. Set up your Nix daemon to have `NIX_GITHUB_PRIVATE_USERNAME` set to your GitHub username, and `NIX_GITHUB_PRIVATE_PASSWORD` set to the token you just created. I've done this by creating the file `/etc/nixos/nix-daemon-environment` with the following content:

    ```
    NIX_GITHUB_PRIVATE_USERNAME=me-and
    NIX_GITHUB_PRIVATE_PASSWORD=github_pat_...
    ```

    and adding the following to my `/etc/nixos/configuration.nix`:

    ```
    systemd.services.nix-daemon.serviceConfig.EnvironmentFile = "/etc/nixos/nix-daemon-environment";
    ```
4. Make sure the Nix daemon picks up the new environment (with the above, you'd need a `nixos-rebuild switch`, possibly followed by `systemctl nix-daemon.service restart`).
5. Run `nix-build -E 'let pkgs = import <nixpkgs> {}; in pkgs.fetchFromGitHub { owner = "<username>"; repo = "<repository>"; rev = "HEAD"; hash = ""; }'`
6. Note you get a 404 error from curl.

### Expected behavior
Rather than a 404 error, the repository downloads and complains about the mismatched hash.

This does work if (a) you use `forceFetchGit = true` in the arguments to `fetchFromGitHub`, (b) you use a classic access token, or (c) you patch `fetchFromGitHub` to download using the API tarball endpoint. I have a PR I'm going to submit shortly that does (c).

### Additional context
This seems like missing GitHub function, and I've also raised a ticket with GitHub Support (<https://support.github.com/ticket/personal/0/2851266>, although I suspect that's not visible to anyone but me).

### Notify maintainers

No maintainers for this module AFAICS, so I'm tagging folk who seem responsible for relevant bits of code.

@wegank @SuperSandro2000 

### Metadata

 - system: `"x86_64-linux"`
 - host os: `Linux 5.15.153.1-microsoft-standard-WSL2, NixOS, 24.05 (Uakari), 24.05.675.805a384895c6`
 - multi-user?: `yes`
 - sandbox: `relaxed`
 - channels(root): `"home-manager-24.05.tar.gz, nixos-24.05, nixos-unstable, nixos-wsl"`

---

me-and:
GitHub currently has two kinds of personal access token: "classic" and "fine-grained".  Fine-grained personal access tokens, as the name suggests, allow much more control over what the token can and cannot do, and in particular allow users to specify which repositories the token should provide access to.

Unfortunately, fine-grained tokens don't allow access to repository archive tarballs for private repositories at (say) https://github.com/me-and/private-demo/archive/HEAD.tar.gz. Fortunately, the GitHub API endpoint does provide this access, and also works with classic tokens and -- for public repositories -- no token at all.

Fixes #321481

## Description of changes

Change `pkgs.fetchFromGitHub` to use the GitHub API for downloading packages.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/build-support/fetchgithub/default.nix
+++ b/pkgs/build-support/fetchgithub/default.nix
@@ -74,15 +74,16 @@ let
       # Use the archive URI for non-private repos, as the API endpoint has
       # relatively restrictive rate limits for unauthenticated users.
       url =
-        let
-          endpoint = "/repos/${owner}/${repo}/tarball/${revWithTag}";
-        in
-        if private
-        then
-          if githubBase == "github.com"
-          then "https://api.github.com${endpoint}"
-          else "https://${githubBase}/api/v3${endpoint}"
-        else "${baseUrl}/archive/${revWithTag}.tar.gz";
+        if private then
+          let
+            endpoint = "/repos/${owner}/${repo}/tarball/${revWithTag}";
+          in
+          if githubBase == "github.com" then
+            "https://api.github.com${endpoint}"
+          else
+            "https://${githubBase}/api/v3${endpoint}"
+        else
+          "${baseUrl}/archive/${revWithTag}.tar.gz";
       extension = "tar.gz";
 
       passthru = {
--- a/pkgs/build-support/fetchgithub/default.nix
+++ b/pkgs/build-support/fetchgithub/default.nix
@@ -43,17 +43,19 @@ let
     else if fetchzip ? override then fetchzip.override { withUnzip = false; }
     else fetchzip;
   privateAttrs = lib.optionalAttrs private {
-    netrcPhase = ''
-      if [ -z "''$${varBase}USERNAME" -o -z "''$${varBase}PASSWORD" ]; then
-        echo "Error: Private fetchFromGitHub requires the nix building process (nix-daemon in multi user mode) to have the ${varBase}USERNAME and ${varBase}PASSWORD env vars set." >&2
-        exit 1
-      fi
-      cat > netrc <<EOF
-      machine ${githubBase}
-              login ''$${varBase}USERNAME
-              password ''$${varBase}PASSWORD
-      EOF
-    '';
+    netrcPhase =
+      let machineName = if githubBase == "github.com" then "api.github.com" else githubBase;
+      in ''
+        if [ -z "''$${varBase}USERNAME" -o -z "''$${varBase}PASSWORD" ]; then
+          echo "Error: Private fetchFromGitHub requires the nix building process (nix-daemon in multi user mode) to have the ${varBase}USERNAME and ${varBase}PASSWORD env vars set." >&2
+          exit 1
+        fi
+        cat > netrc <<EOF
+        machine ${machineName}
+                login ''$${varBase}USERNAME
+                password ''$${varBase}PASSWORD
+        EOF
+      '';
     netrcImpureEnvVars = [ "${varBase}USERNAME" "${varBase}PASSWORD" ];
   };
 
@@ -66,7 +68,22 @@ let
       inherit tag rev deepClone fetchSubmodules sparseCheckout fetchLFS; url = gitRepoUrl;
     } // lib.optionalAttrs (leaveDotGit != null) { inherit leaveDotGit; }
     else {
-      url = "${baseUrl}/archive/${revWithTag}.tar.gz";
+      # Use the API endpoint for private repos, as the archive URI doesn't
+      # support access with GitHub's fine-grained access tokens.
+      #
+      # Use the archive URI for non-private repos, as the API endpoint has
+      # relatively restrictive rate limits for unauthenticated users.
+      url =
+        let
+          endpoint = "/repos/${owner}/${repo}/tarball/${revWithTag}";
+        in
+        if private
+        then
+          if githubBase == "github.com"
+          then "https://api.github.com${endpoint}"
+          else "https://${githubBase}/api/v3${endpoint}"
+        else "${baseUrl}/archive/${revWithTag}.tar.gz";
+      extension = "tar.gz";
 
       passthru = {
         inherit gitRepoUrl;


me-and:
Significantly, the new URLs are based on the GitHub API documentation at <https://docs.github.com/en/rest/repos/contents?apiVersion=2022-11-28#download-a-repository-archive-tar> and <https://docs.github.com/en/enterprise-server@3.13/rest/repos/contents?apiVersion=2022-11-28#download-a-repository-archive-tar>, and I've tested the behaviour with my own repositories on GitHub, but I don't have access to a GitHub Enterprise instance to be able to confirm things work as expected there.

me-and:
Force-push comes with hopefully-slightly-clearer code and slightly more detailed commit message; there shouldn't be any substantive changes.

SuperSandro2000 (commented):
LGTM but I didn't find the time to roughly test this

me-and:
GitHub Support have now confirmed the current behaviour is expected and won't be fixed; the fact that classic access tokens provided access to the archive tarballs under github.com rather than api.github.com was "incidental rather than intentional".

That means this isn't a workaround but a required fix for fine-grained access tokens to work in this scenario.

instantepiphany:
@me-and Whats next for this PR? I've hit this issue also, would love to see your fix in `master`.

Looks like GH thinks there is a merge conflict, but also it says the changes can be cleanly merged 🤔 

When I click "Resolve conflicts" I get a 404

pkgs/build-support/fetchgithub/default.nix
```diff
@@ -55,7 +57,13 @@ let
       inherit rev deepClone fetchSubmodules sparseCheckout; url = gitRepoUrl;
     } // lib.optionalAttrs (leaveDotGit != null) { inherit leaveDotGit; }
     else {
-      url = "${baseUrl}/archive/${rev}.tar.gz";
```
Mic92, line 58: This means the api endpoint would be always used, even if no token is provided. Correct?
Would this not mean that public CI that does not access private repositories and therefore has no token would run into API limits?

me-and: Correct, although I would be very surprised if there is a different rate limit applied to this new URL than to the previous one.

Mic92: The other one does not have rate limits. This would also break builds if many people are behind a shared ip address. i.e. this would break in github action runner.

me-and:
> @me-and Whats next for this PR? I've hit this issue also, would love to see your fix in `master`.
> 
> Looks like GH thinks there is a merge conflict, but also it says the changes can be cleanly merged 🤔
> 
> When I click "Resolve conflicts" I get a 404

There was definitely a merge conflict; I've just rebased onto master again, so I think this should be good to merge as soon as someone with the authority to do so comes along…

philiptaron:
I'll give this a look today.

philiptaron:
@Mic92's [comment here](https://github.com/NixOS/nixpkgs/pull/321484#discussion_r1895497204) makes me think that this whole endeavor isn't possible: 

> The other one does not have rate limits. This would also break builds if many people are behind a shared IP address. i.e. this would break in GitHub action runner.

Unless there's a way to cause retries on GitHub rate limiting, so that the derivation doesn't fail, isn't this a regression with a wide impact?

I get the motivation. Maybe we need a `fetchFromGitHubApi`?

me-and:
> @Mic92's [comment here](https://github.com/NixOS/nixpkgs/pull/321484#discussion_r1895497204) makes me think that this whole endeavor isn't possible: 
> 
> > The other one does not have rate limits. This would also break builds if many people are behind a shared IP address. i.e. this would break in GitHub action runner.
> 
> Unless there's a way to cause retries on GitHub rate limiting, so that the derivation doesn't fail, isn't this a regression with a wide impact?
> 
> I get the motivation. Maybe we need a `fetchFromGitHubApi`?

I think `fetchFromGitHubApi` is overkill. We can use the API endpoint if and only if `private` is set. I see three cases there:

- The mainline case with a public repo, which will work exactly the same as it does now.
- The case I was trying to solve for, with fine-grained access tokens, which will work in the only way it can.
- The case with the standard access tokens, which will change from the current behaviour to using the API endpoint, but as it's necessarily an authenticated connection, the API limits are high enough that it's safe to ignore them.

The alternative (which I think is still preferable to a separate function) would be to add a new `api` parameter or similar, but I don't think it's necessary to add a new parameter when the existing `private` parameter seems to provide all the information we need.

me-and:
I meant to add: I'll update this PR with a new version with the above behaviour in the next day or so.

me-and:
@philiptaron I think this is now ready to re-review and merge.

pkgs/build-support/fetchgithub/default.nix
```diff
@@ -43,17 +43,19 @@ let
...
-      machine ${githubBase}
-              login ''$${varBase}USERNAME
-              password ''$${varBase}PASSWORD
-      EOF
-    '';
+    netrcPhase =
+      let machineName = if githubBase == "github.com" then "api.github.com" else githubBase;
+      in ''
+        if [ -z "''$${varBase}USERNAME" -o -z "''$${varBase}PASSWORD" ]; then
+          echo "Error: Private fetchFromGitHub requires the nix building process (nix-daemon in multi user mode) to have the ${varBase}USERNAME and ${varBase}PASSWORD env vars set." >&2
```
Eveeifyeve, line 50: Just a Suggestion to varBase Password to PATs (Personal Access Token) github api password is deprecated as stated by github
> Beginning August 13, 2021, we will no longer accept account passwords when authenticating Git operations on GitHub.com.

[see the follow blog post for more infomation](https://github.blog/security/application-security/token-authentication-requirements-for-git-operations/)

_**You don't have to take this seriously it's just a suggestion**_

pkgs/build-support/fetchgithub/default.nix
```diff
@@ -43,17 +43,19 @@ let
...
+          exit 1
+        fi
+        cat > netrc <<EOF
+        machine ${machineName}
+                login ''$${varBase}USERNAME
+                password ''$${varBase}PASSWORD
```
Eveeifyeve, line 56:
```suggestion
      in ''
        if [ -z "''$${varBase}USERNAME" -o -z "''$${varBase}PAT" ]; then
          echo "Error: Private fetchFromGitHub requires the nix building process (nix-daemon in multi user mode) to have the ${varBase}USERNAME and ${varBase}PAT env vars set." >&2
          exit 1
        fi
        cat > netrc <<EOF
        machine ${machineName}
                login ''$${varBase}USERNAME
                password ''$${varBase}PASSWORD
```

Just a Suggestion to varBase Password to PATs (Personal Access Token) github api password is deprecated as stated by github
> Beginning August 13, 2021, we will no longer accept account passwords when authenticating Git operations on GitHub.com.

[see the follow blog post for more infomation](https://github.blog/security/application-security/token-authentication-requirements-for-git-operations/)

_**You don't have to take this seriously it's just a suggestion**_

philiptaron:
> @philiptaron I think this is now ready to re-review and merge.

I'll take a look. The scope is now gives me less heartburn. I'll try to use it to make a Nix derivation for a private repository I have...

pkgs/build-support/fetchgithub/default.nix
me-and: I'd rather not: that change would break existing configurations, as `...PASSWORD` has been the variable name for the access token for a long time. And it's still accurate from a protocol perspective: the value gets put in the password part of the HTTP request. The blog post is just that GitHub no longer accepts a GitHub password in the password field.

h0nIg:
LGTM IMHO

me-and:
> Missing enhancement: most? github enterprise instances have SAML/OIDC protection and redirects in front of the instance. So fetchurl will never reach the download and will get redirected to the login page.
> 
> Would it be a good enhancement to call the tarball API without "private" being set and a different flag instead? "private" requires special ENV to get set and i would love to see support for nix.conf netrc-file or access-tokens which may get used

I think that's a separate enhancement, so warrants its own issue and/or PR rather than trying to roll it into this PR.

pkgs/build-support/fetchgithub/default.nix
```diff
@@ -66,7 +68,22 @@ let
...
+      url =
+        let
+          endpoint = "/repos/${owner}/${repo}/tarball/${revWithTag}";
+        in
+        if private
+        then
+          if githubBase == "github.com"
+          then "https://api.github.com${endpoint}"
+          else "https://${githubBase}/api/v3${endpoint}"
+        else "${baseUrl}/archive/${revWithTag}.tar.gz";
```
philiptaron, line 85: Could scope it like this:

```suggestion
        if private then
          let
            endpoint = "/repos/${owner}/${repo}/tarball/${revWithTag}";
          in
          if githubBase == "github.com" then
            "https://api.github.com${endpoint}"
          else
            "https://${githubBase}/api/v3${endpoint}"
        else
          "${baseUrl}/archive/${revWithTag}.tar.gz";
```

What do you think? It's all the same in the end, just a narrower scope this way.

pkgs/build-support/fetchgithub/default.nix
```diff
@@ -43,17 +43,19 @@ let
...
-              login ''$${varBase}USERNAME
-              password ''$${varBase}PASSWORD
-      EOF
-    '';
+    netrcPhase =
+      let machineName = if githubBase == "github.com" then "api.github.com" else githubBase;
+      in ''
+        if [ -z "''$${varBase}USERNAME" -o -z "''$${varBase}PASSWORD" ]; then
+          echo "Error: Private fetchFromGitHub requires the nix building process (nix-daemon in multi user mode) to have the ${varBase}USERNAME and ${varBase}PASSWORD env vars set." >&2
+          exit 1
```
philiptaron, line 51: I found your directions on how to do this in [the linked issue](https://github.com/NixOS/nixpkgs/issues/321481#issue-2366305172) really helpful. I wonder if there's a place in our documentation universe that they could be landed, since the route to making them be set isn't exactly paved.

philiptaron (approved):
I tested both the private repository case and also the public repository case. Everything worked nicely with this PR, and I experienced the symptoms of #321481 when I tried them again on master. This is a nice upgrade for the `private` usecase.

I plan on merging this today unless there are objections from the others on the PR.

pkgs/build-support/fetchgithub/default.nix
me-and: I think this is probably a bit clearer, yes! I'm not going to be able to make the change and update the PR until tomorrow, but I *think* you should be able to make the change yourself before merging if you want to get it done before then.

Eveeifyeve: I feel the scope could be a bit more neater, but it looks good.

pkgs/build-support/fetchgithub/default.nix
philiptaron: I'm fine with a future PR.

philiptaron (approved):

PR #321519, opened 2024-06-21
maintainers: drop lheckemann

lheckemann:
If the open letter is dismissed as irrelevant because "none of the people signing it left", I guess it's time for me to leave.

### DIFF
--- a/maintainers/maintainer-list.nix
+++ b/maintainers/maintainer-list.nix
@@ -11310,12 +11310,6 @@
     github = "LogicalOverflow";
     githubId = 5919957;
   };
-  lheckemann = {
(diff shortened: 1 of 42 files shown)


vcunat:
Honestly I'm rather annoyed by the showy leavings/ultimata.  (by which I don't mean yours actually)

I believed that the substance of the letter is among things to be "addressed" by the assembly.  That is an ongoing process.  I'm sure that it won't be easy.

Anyway, thanks for your contributions.  This line sounds empty I guess, but thanks anyway.

imincik:
I am very sorry to see you leaving Linus . Thanks for everything.

PR #321521, opened 2024-06-21
vmctl: init at v0.99-unstable-2024-05-14

Panky-codes:
## Description of changes
Vmctl is a command line tool focused on NVMe testing in QEMU. vmctl provides easy config options to test different NVMe configurations easily in QEMU.

Github: https://github.com/SamsungDS/vmctl
## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/by-name/vm/vmctl/package.nix
+++ b/pkgs/by-name/vm/vmctl/package.nix
@@ -1,0 +1,70 @@
+{
+  lib,
+  stdenvNoCC,
+  fetchFromGitHub,
+  openssh,
+  socat,
+  gawk,
+  cloud-utils,
+  cdrtools,
+  qemu,
+  qemu-utils,
+  coreutils,
+  getopt,
+  makeWrapper,
+}:
+
+stdenvNoCC.mkDerivation {
+  pname = "vmctl";
+  version = "v0.99-unstable-2024-05-14";
+
+  src = fetchFromGitHub {
+    owner = "SamsungDS";
+    repo = "vmctl";
+    rev = "5b6b7084b8cba06b474c0e020df965237f2c832c";
+    hash = "sha256-yDgaY2RJXBjWkMSQb4JaJzMGLFzowfOGixSRzzv2iIk=";
+  };
+
+  dontBuild = true;
+
+  postPatch = ''
+    substituteInPlace vmctl \
+      --replace 'BASEDIR="$(dirname "$(readlink -f "''${BASH_SOURCE[0]}")")"' 'BASEDIR="${placeholder "out"}"'
+  '';
+
+  nativeBuildInputs = [ makeWrapper ];
+
+  installPhase = ''
+    runHook preInstall
+
+    install -Dm555 vmctl -t "$out/bin"
+    wrapProgram "$out/bin/vmctl" \
+      --set PATH "${
+        lib.makeBinPath [
+          openssh
+          socat
+          gawk
+          cloud-utils
+          cdrtools
+          qemu
+          qemu-utils
+          coreutils
+          getopt
+        ]
+      }"
+    cp -r {cmd,common,contrib,lib} $out
+
+    runHook postInstall
+  '';
+
+  meta = {
+    description = "Command line tool focused on NVMe testing in QEMU";
+    homepage = "https://github.com/SamsungDS/vmctl";
+    license = lib.licenses.gpl3;
+    maintainers = with lib.maintainers; [ panky ];
+    platforms = [
+      "x86_64-linux"
+      "aarch64-linux"
+    ];
+  };
+}
--- a/maintainers/maintainer-list.nix
+++ b/maintainers/maintainer-list.nix
@@ -15207,6 +15207,12 @@
     githubId = 79252025;
     name = "Nicolas Benes";
   };
+  panky = {
+    email = "dev@pankajraghav.com";
+    github = "Panky-codes";
+    githubId = 33182938;
+    name = "Pankaj";
+  };
   paperdigits = {
     email = "mica@silentumbrella.com";
     github = "paperdigits";


djacu (changes requested):
Looks pretty good. I'd recommend two changes.

1. Split up your commit so the first one adds you as a maintainer and the second one adds the package.
2. Since this package is new, I'd recommend formatting your `package.nix` file with the official formatter, [nixfmt-rfc-style](https://search.nixos.org/packages?channel=unstable&show=nixfmt-rfc-style&from=0&size=50&sort=relevance&type=packages&query=nixfmt-rfc-style). You can get it quickly into your shell with `nix-shell -p nixfmt-rfc-style` and run `nixfmt <path to your package.nix>`. Here is diff when I run it locally.

```
$ git diff
diff --git a/pkgs/by-name/vm/vmctl/package.nix b/pkgs/by-name/vm/vmctl/package.nix
index 885a9e619ffc..1e378d044c2a 100644
--- a/pkgs/by-name/vm/vmctl/package.nix
+++ b/pkgs/by-name/vm/vmctl/package.nix
@@ -1,4 +1,10 @@
-{ lib, stdenv, bash, shellcheck, fetchFromGitHub}:
+{
+  lib,
+  stdenv,
+  bash,
+  shellcheck,
+  fetchFromGitHub,
+}:
 
 stdenv.mkDerivation {
   pname = "vmctl";
@@ -11,7 +17,10 @@ stdenv.mkDerivation {
     sha256 = "sha256-yDgaY2RJXBjWkMSQb4JaJzMGLFzowfOGixSRzzv2iIk=";
   };
 
-  nativeBuildInputs = [ bash shellcheck ];
+  nativeBuildInputs = [
+    bash
+    shellcheck
+  ];
 
   installPhase = ''
     mkdir -p $out/bin
@@ -27,6 +36,9 @@ stdenv.mkDerivation {
     homepage = "https://github.com/SamsungDS/vmctl";
     license = lib.licenses.gpl3;
     maintainers = with lib.maintainers; [ panky ];
-    platforms = [ "x86_64-linux" "aarch64-linux"];
+    platforms = [
+      "x86_64-linux"
+      "aarch64-linux"
+    ];
   };
 }
```

djacu:
Result of `nixpkgs-review pr 321521` run on x86_64-linux [1](https://github.com/Mic92/nixpkgs-review)
<details>
  <ul>
    <li>vmctl</li>
  </ul>
</details>

Panky-codes:
@djacu Thanks for the comments. Updated the PR now :)

djacu:
Result of `nixpkgs-review pr 321521` run on x86_64-linux [1](https://github.com/Mic92/nixpkgs-review)
<details>
  <ul>
    <li>vmctl</li>
  </ul>
</details>

djacu (approved):
LGTM :rocket:

pkgs/by-name/vm/vmctl/package.nix
```diff
@@ -0,0 +1,44 @@
...
+  ];
+
+  installPhase = ''
+    mkdir -p $out/bin
+    cp vmctl $out/bin
+    cp -r cmd $out/bin
+    cp -r common $out/bin
+    cp -r contrib $out/bin
+    cp -r lib $out/bin
+  '';
```
h7x4, line 32:
```suggestion
  installPhase = ''
    runHook preInstall
  
    mkdir -p $out/bin
    cp -r {vmctl,cmd,common,contrib,lib} $out/bin
    
    runHook postInstall
  '';
```

pkgs/by-name/vm/vmctl/package.nix
```diff
@@ -0,0 +1,44 @@
...
+
+  src = fetchFromGitHub {
+    owner = "SamsungDS";
+    repo = "vmctl";
+    rev = "5b6b7084b8cba06b474c0e020df965237f2c832c";
+    sha256 = "sha256-yDgaY2RJXBjWkMSQb4JaJzMGLFzowfOGixSRzzv2iIk=";
+  };
+
+  nativeBuildInputs = [
+    bash
```
h7x4, line 21: Could you elaborate on the reason behind putting bash here? It's unusual to list bash as a native input

pkgs/by-name/vm/vmctl/package.nix
```diff
@@ -0,0 +1,44 @@
...
+    shellcheck
```
h7x4, line 22: Looking at the upstream `Makefile`, it seems like this is used in the `check` target, which I think is built during our `checkPhase`. Maybe it would fit better in `nativeCheckInputs`?

pkgs/by-name/vm/vmctl/package.nix
```diff
@@ -0,0 +1,44 @@
...
+  ];
+
+  installPhase = ''
+    mkdir -p $out/bin
+    cp vmctl $out/bin
+    cp -r cmd $out/bin
+    cp -r common $out/bin
+    cp -r contrib $out/bin
+    cp -r lib $out/bin
```
h7x4, line 31: It seems weird to me that all of these are going to `$out/bin`. I suppose it is because the entrypoint looks for the paths relative to itself? Do you think we could patch the code to look out of `bin`, so that scripts like "ssh" and "kill" doesn't end up in the users `$PATH`?

pkgs/by-name/vm/vmctl/package.nix
```diff
@@ -0,0 +1,44 @@
...
+  lib,
+  stdenv,
+  bash,
+  shellcheck,
+  fetchFromGitHub,
+}:
+
+stdenv.mkDerivation {
+  pname = "vmctl";
+  version = "1";
```
h7x4, line 11: Is there an actual release for this anywhere? I only see tag `v0.99`

pkgs/by-name/vm/vmctl/package.nix
Panky-codes: Looks good.

pkgs/by-name/vm/vmctl/package.nix
Panky-codes: It is a bash script. Actually, bash is more like a runtime dependency and not a build input. So can we just assume it is present instead of explicitly mentioning it?

h7x4: That would mean it's not native (i.e. if you were cross building from x64 to arm64, you'd want to end up with the arm64 bash, not the x64 one), and it also happens to get fixed by default in the `fixupPhase` where there is a script that replaces the shebangs with nix-specific ones. You can probably just drop it :)

pkgs/by-name/vm/vmctl/package.nix
Panky-codes: This is a good point. But the problem is, the check target is triggered even if I do just a `make`. I think I should patch the Makefile to not make it the default target somehow. I need check the Makefile syntax how to do it.

Panky-codes: Something like this:
```
diff --git a/Makefile b/Makefile
index b06e6f4..184bd20 100644
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,11 @@
 SHELLCHECK := $(shell command -v shellcheck 2>/dev/null)
 
+default: all
+
+all:
+
+.PHONY: all check
+
 check:
 ifndef SHELLCHECK
        $(error "cannot find shellcheck; install to run check")
```

h7x4: Maybe the `buildPhase` isn't even needed? It's not like the makefile is doing anything else than running shellcheck. Maybe you could specify these?

```nix
dontBuild = true;
doCheck = true;
```

pkgs/by-name/vm/vmctl/package.nix
Panky-codes: Hmm, at the moment everything id done relative to `BASEDIR`, and that is why I put it all in the bin folder. So you think we can patch "${BASEDIR}" with something like `${nixpath-vmctl}/lib` in the scripts?

pkgs/by-name/vm/vmctl/package.nix
Panky-codes: Much better idea :D

pkgs/by-name/vm/vmctl/package.nix
Panky-codes: ssh and kill will not be  in `$PATH` because it will in `$out/bin/cmd/ssh|kill`, so technically it should be fine?

pkgs/by-name/vm/vmctl/package.nix
Panky-codes: I think the main developer stopped doing release sometime ago. I have asked him to restart it. But for now I consider this stable, so started with version `1`.

pkgs/by-name/vm/vmctl/package.nix
h7x4: Adding a `../` would suffice, but you could override `BASEDIR` to be `${placeholder "out"}/` if you'd like.

pkgs/by-name/vm/vmctl/package.nix
h7x4: I see. In that case, I don't think we should mark it as version `1`. Even if the code itself happens to be stable on the main branch, being accurate in the version tagging is important for other reasons, e.g. for not notifying other distros about non-existent updates in [repology](https://repology.org/). By our [version scheme](https://github.com/NixOS/nixpkgs/blob/master/pkgs/README.md#package-naming), the version should be `v0.99-unstable-2024-05-14`

pkgs/by-name/vm/vmctl/package.nix
```diff
@@ -0,0 +1,44 @@
+{
+  lib,
+  stdenv,
+  bash,
+  shellcheck,
+  fetchFromGitHub,
+}:
+
+stdenv.mkDerivation {
```
h7x4, line 9: By the way, you can probably get by without a C toolchain here.

```suggestion
  stdenvNoCC,
  bash,
  shellcheck,
  fetchFromGitHub,
}:

stdenvNoCC.mkDerivation {
```

pkgs/by-name/vm/vmctl/package.nix
Panky-codes: In the code, we get BASEDIR and use "${BASEDIR}" subsequently. I was thinking of using substituteInPlace to replace "${BASEDIR}" with ".." . I think dollar symbol is confusing nix. How do we get subsitituteInPlace to replace the whole `${BASEDIR}` with `..`? Probably a noob question.

h7x4: Ah, sorry. I didn't intend for you to replace the entirety of BASEDIR with `..`, it was more like "if BASEDIR always points to `$out/bin` and does so in a dynamic way, then maybe we can override `BASEDIR` to be `$BASEDIR/..` somehow".

As for the dollar symbol thing, nix uses `${}` to interpolate it's own variables. You can write `''${}` to escape the nix interpolation and leave it for bash.

Seeing as how `BASEDIR` is only set once, maybe something like this would work?

```
postPatch = ''
  substituteInPlace vmctl \
    --replace 'BASEDIR="$(dirname "$(readlink -f "''${BASH_SOURCE[0]}")")"' 'BASEDIR="${placeholder "out"}"'
'';
```

Panky-codes: Yes, that looks good to me. Thanks for all the comments :)

Panky-codes:
Thanks @h7x4 and @djacu for the review comments :)

pkgs/by-name/vm/vmctl/package.nix
```diff
@@ -0,0 +1,50 @@
...
+    repo = "vmctl";
+    rev = "5b6b7084b8cba06b474c0e020df965237f2c832c";
+    sha256 = "sha256-yDgaY2RJXBjWkMSQb4JaJzMGLFzowfOGixSRzzv2iIk=";
+  };
+
+  dontBuild = true;
+  doCheck = true;
+
+  nativeCheckInputs = [ shellcheck ];
+
```
SuperSandro2000, line 24:
```suggestion
```

We don't want to fail this package if shellcheck gets updated and adds new checks.

Panky-codes: Hmm, so we just remove it and do ` dontCheck = true;` ?

Panky-codes: or just remove `docheck = true;`

pkgs/by-name/vm/vmctl/package.nix
```diff
@@ -0,0 +1,47 @@
...
+  };
+
+  dontBuild = true;
+
+  installPhase = ''
+    runHook preInstall
+
+    mkdir -p $out/bin
+    cp vmctl $out/bin
+    cp -r {cmd,common,contrib,lib} $out
```
h7x4, line 27: The program seems to have a few runtime dependencies that we should add to its `$PATH`. Upstream README seems to specify `ssh` and `socat`, but I found at least `tail`, `awk`, `cloud-localds` and `mkisofs` as well. I think `tail` (`coreutils`) might be given, but the others are probably not. Do you happen to know about any other runtime dependencies?

Remember to add the programs to args up top, as well as adding `makeWrapper` to `nativeBuildInputs` (and args)

```suggestion
    wrapProgram "$out/bin/vmctl" \
      --prefix PATH : "${lib.makeBinPath [ ssh socat gawk cloud-utils cdrtools ]}"

    cp -r {cmd,common,contrib,lib} $out
```

h7x4 (commented):
Just a few more comments

pkgs/by-name/vm/vmctl/package.nix
Panky-codes: You are right. It also needs `qemu` and `qemu-tools`. I will also add that

Panky-codes: @h7x4 I found this gist about wrapProgram [link](https://gist.github.com/CMCDragonkai/9b65cbb1989913555c203f4fa9c23374). Should I use --set instead of --prefix?

Plus, should I make part of postFixup or is it fine if it is a part of installPhase?

Panky-codes: I used `set` instead `prefix`. It also helped find me all runtime dependencies. :)

pkgs/by-name/vm/vmctl/package.nix
```diff
@@ -0,0 +1,70 @@
...
+            socat
+            gawk
+            cloud-utils
+            cdrtools
+            qemu
+            qemu-utils
+            coreutils
+            getopt
+          ]
+        }"
```
h7x4, line 54: The first line has an odd indent

```suggestion
    wrapProgram "$out/bin/vmctl" \
      --set PATH "${
        lib.makeBinPath [
          openssh
          socat
          gawk
          cloud-utils
          cdrtools
          qemu
          qemu-utils
          coreutils
          getopt
        ]
      }"
```

h7x4:
I think this is starting to look pretty good

pkgs/by-name/vm/vmctl/package.nix
Panky-codes: Hmm, I used [nixfmt-rfc-style](https://search.nixos.org/packages?channel=unstable&show=nixfmt-rfc-style&from=0&size=50&sort=relevance&type=packages&query=nixfmt-rfc-style) to format it. I don't why it didn't catch it!

h7x4 (approved):
- [x] verified github id: https://api.github.com/user/33182938

Feel free to ping me for merge once ofborg is done with eval and 2 linux builds.

PR #321591, opened 2024-06-21
nixos/gdm: allow logging in via fingerprint

JohnRTitor:
## Description of changes

Lets GDM/GNOME users to login via fingerprint if fprintd service is enabled.

Cherry-picked from #306338. Have not tested but should work fine. Originally borrowed from https://gitlab.gnome.org/GNOME/gdm/-/blob/main/data/pam-arch/gdm-fingerprint.pam

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/nixos/modules/services/x11/display-managers/gdm.nix
+++ b/nixos/modules/services/x11/display-managers/gdm.nix
@@ -321,6 +321,22 @@ in
         session   include       login
       '';
 
+      login.fprintAuth = mkIf config.services.fprintd.enable false;
+      gdm-fingerprint.text = mkIf config.services.fprintd.enable ''
+        auth       required                    pam_shells.so
+        auth       requisite                   pam_nologin.so
+        auth       requisite                   pam_faillock.so      preauth
+        auth       required                    ${pkgs.fprintd}/lib/security/pam_fprintd.so
+        auth       optional                    pam_permit.so
+        auth       required                    pam_env.so
+        auth       [success=ok default=1]      ${pkgs.gnome.gdm}/lib/security/pam_gdm.so
+
+        account    include                     login
+
+        password   required                    pam_deny.so
+
+        session    include                     login
+      '';
     };
 
   };


JohnRTitor:
Please have a look as well @LeSuisse, changes are mostly same as amaxine's PR

I removed [manual addition](https://github.com/NixOS/nixpkgs/compare/b3e54964bee0f0c87ad1e57558de2d1fbb3147b5..9d41fe6fcc4df838a56b1cfb2512b65e5e655958) of `gnome-keyring.so` so to go with `security.pam.services.gdm-fingerprint.enableGnomeKeyring = true;`

Pandapip1 (approved):
Diff LGTM. Have not tested myself.

JohnRTitor:
Thanks, I will keep it open for a while for others to provide feedback on it.

schuelermine:
Wanted to note that fingerprint login was already possible before, it just didn’t work well (fingerprint would immediately fail, then reactivate and block all password login attempts until it failed again)

schuelermine:
I’m using a Framework Laptop 16 and following this patch fingerprint login has become more inconsistent than before.

JohnRTitor:
Could be related to https://github.com/NixOS/nixpkgs/blob/09b3443ecb2c6a897a867ab421ba8234b939d6e3/nixos/modules/services/x11/display-managers/gdm.nix#L324, could you mkForce it to true and tell us the result? 

Fingerprint support on NixOS needs wider testing. Would love if you could help.

schuelermine:
doing it now

schuelermine:
doing this prevents me from logging in via password, and also GDM repeatedly shakes the password box and displays an error about fingerprint login not working (but when I touch the finger I can log in)

schuelermine:
note: that was the behavior on the lock screen
it was also the behavior before this force (but after this PR) but it would refuse fingerprint and accept password

schuelermine:
I will test a fresh login later

nixos/modules/services/x11/display-managers/gdm.nix
```diff
@@ -321,6 +321,22 @@ in
...
 
+      login.fprintAuth = mkIf config.services.fprintd.enable false;
+      gdm-fingerprint.text = mkIf config.services.fprintd.enable ''
+        auth       required                    pam_shells.so
+        auth       requisite                   pam_nologin.so
+        auth       requisite                   pam_faillock.so      preauth
+        auth       required                    ${pkgs.fprintd}/lib/security/pam_fprintd.so
+        auth       optional                    pam_permit.so
+        auth       required                    pam_env.so
+        auth       [success=ok default=1]      ${pkgs.gnome.gdm}/lib/security/pam_gdm.so
```
jtojnar, line 332: I am seeing the following in logs:

gdm-fingerprint][1033624]: PAM bad jump in stack

Probably caused by the `default=1`, which should skip one `auth` instruction when `pam_gdm` gives anything but success, there are no further `auth`instructions.

nixos/modules/services/x11/display-managers/gdm.nix
```diff
@@ -321,6 +321,22 @@ in
         session   include       login
       '';
 
+      login.fprintAuth = mkIf config.services.fprintd.enable false;
+      gdm-fingerprint.text = mkIf config.services.fprintd.enable ''
+        auth       required                    pam_shells.so
+        auth       requisite                   pam_nologin.so
+        auth       requisite                   pam_faillock.so      preauth
+        auth       required                    ${pkgs.fprintd}/lib/security/pam_fprintd.so
+        auth       optional                    pam_permit.so
```
jtojnar, line 330: I started looking into pam two weeks ago and according to my current understanding, this line does not do anything.

jtojnar:
> Could be related to
> 
> https://github.com/NixOS/nixpkgs/blob/09b3443ecb2c6a897a867ab421ba8234b939d6e3/nixos/modules/services/x11/display-managers/gdm.nix#L324
>
> , could you mkForce it to true and tell us the result?

I do not think that is desirable.

Not exactly sure what would happen but at best, I would expect it to lead to the same behaviour as before (user is not prompted for password until fingerprint auth fails). And at worst, two pam modules would ask for the fingerprint at the same time, which sounds like a fun way to get new bugs.

nixos/modules/services/x11/display-managers/gdm.nix
jtojnar: `security.pam.services.gdm-fingerprint.enableGnomeKeyring = true;` does not appear to do anything for me.

But I am now pretty sure invoking it manually as in https://github.com/NixOS/nixpkgs/commit/b3e54964bee0f0c87ad1e57558de2d1fbb3147b5 would be fine.

If we want to make it conditional, we would need to replace the one after the jump with the `pam_permit.so` line when disabled.

jtojnar:
Opened https://github.com/NixOS/nixpkgs/pull/324347 with a fix, also adding gnome-keyring.

PR #321601, opened 2024-06-21
Re-add grab-site with updates to dependencies for python 3.12

pastly:
## Description of changes

Re-adds and bumps the version of ludios_wpull as necessary for grab-site to work with python 3.12. Consequently bumps tornado to 6.4 and adds `packaging` as a dependency to yapsy (like it's supposed to be!).

The ludios_wpull version update sounds huge, but it's just one real bump. No changelog in the project, but [see release notes](https://github.com/ArchiveTeam/ludios_wpull/releases).

Tornado [release notes page](https://www.tornadoweb.org/en/stable/releases.html). It's only a bump of two versions. I stopped at 6.4.0 instead of going to 6.4.1 so that the nix mitmproxy tests wouldn't barf at a tornado version greater than 6.4 (I was running `nix-build -A pkgs.python3Packages.tornado.passthru.tests` as docs suggested`)

I'm running the above `passthru.tests` command now ... but it's seemingly going to take a long time since tornado is a popular library. 

Do I need to run it myself? Do I need to run it for more versions of python?

Thanks

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [x] `sandbox = true` (it's Linux)
- [ ] Tested, as applicable:

---


pkgs/development/python-modules/yapsy/default.nix
```diff
@@ -4,9 +4,10 @@
   fetchFromGitHub,
   setuptools,
   pytestCheckHook,
+  python,
 }:
 
-buildPythonPackage {
+with python.pkgs; buildPythonPackage {
```
kirillrdy, line 10: can you explain the reason for this change ?

pkgs/tools/backup/grab-site/default.nix
```diff
@@ -0,0 +1,47 @@
...
+  pname = "grab-site";
+  version = "2.2.7";
+
+  disabled = pythonOlder "3.12";
+
+  src = fetchFromGitHub {
+    rev = version;
+    owner = "ArchiveTeam";
+    repo = "grab-site";
+    sha256 = "sha256-tf8GyFjya3+TVc2VjlY6ztfjCJgof6tg4an18pz+Ig8=";
```
kirillrdy, line 17:
```suggestion
    hash = "sha256-tf8GyFjya3+TVc2VjlY6ztfjCJgof6tg4an18pz+Ig8=";
```

pkgs/tools/backup/grab-site/default.nix
```diff
@@ -0,0 +1,47 @@
...
+  };
+
+  postPatch = ''
+    substituteInPlace setup.py \
+      --replace '"wpull @ https://github.com/ArchiveTeam/ludios_wpull/tarball/master#egg=wpull-${ludios_wpull.version}"' '"wpull"'
```
kirillrdy, line 22:
```suggestion
      --replace-fail '"wpull @ https://github.com/ArchiveTeam/ludios_wpull/tarball/master#egg=wpull-${ludios_wpull.version}"' '"wpull"'
```

pkgs/tools/backup/grab-site/default.nix
```diff
@@ -0,0 +1,47 @@
+{ lib
```
kirillrdy, line 1: all new packages must be added to by-name hierarchy  

https://github.com/NixOS/nixpkgs/tree/master/pkgs/by-name#readme

pkgs/tools/backup/grab-site/default.nix
```diff
@@ -0,0 +1,47 @@
...
+    manhole
+    lmdb
+    autobahn
+    fb-re2
+    websockets
+    faust-cchardet
+  ];
+
+  checkPhase = ''
+    export PATH=$PATH:$out/bin
```
kirillrdy, line 37: isn't  $out is still empty during checkPhase ?

pkgs/tools/backup/grab-site/default.nix
```diff
@@ -0,0 +1,47 @@
...
+    bash ./tests/offline-tests
+  '';
+
+  meta = with lib; {
+    description = "Crawler for web archiving with WARC output";
+    homepage = "https://github.com/ArchiveTeam/grab-site";
+    license = licenses.mit;
+    platforms = platforms.all;
```
kirillrdy, line 45: platforms is not necessary for buildPythonApplication, unless it differs from python3.platforms

pkgs/tools/backup/grab-site/default.nix
pastly: That seems to be incorrect, as the `bash ./tests/offline-tests` on the next line fails to execute, complaining about `grab-site` not being in the path. So for now I'm leaving the `PATH` line here.

pkgs/development/python-modules/yapsy/default.nix
pastly: It's `packaging` dependency needs it; In an upcoming commit I've removed the `with python.pkgs` and instead used `python.pkgs.packaging` as the dependency. Thanks.

pastly:
Oh, I should probably cc @ivan as they were the original maintainer of these packages, I'm re-adding them [after their removal](https://github.com/NixOS/nixpkgs/pull/267001), and as the are now, I didn't list a maintainer. 

Would you like me to add you back?

ivan:
Yes, I would like to be in the `maintainers`, thanks.

pkgs/development/python-modules/ludios_wpull/default.nix
```diff
@@ -0,0 +1,68 @@
...
+# self: super: { pkg = super.pkg.overridePythonAttrs (oldAttrs: { ... }); }
+# Applied after defaultOverrides
+, packageOverrides ? self: super: {}
+}:
+
+let
+  defaultOverrides = [
+    # override the version of packages pinned in pyproject.toml
+    (self: super: {
+      sqlalchemy = super.sqlalchemy.overridePythonAttrs (oldAttrs: rec {
```
kirillrdy, line 27: overrides can not be done for python modules, because it would introduce clashes with other dependencies

overrides can be done for applications. from my initial look wpull is intended to be used as an application and not python module ( eg import )

if that's the case then it needs to be moved to by-name https://github.com/NixOS/nixpkgs/blob/master/pkgs/by-name/README.md

pkgs/development/python-modules/ludios_wpull/default.nix
```diff
@@ -0,0 +1,68 @@
...
+    # Put packageOverrides at the start so they are applied after defaultOverrides
+    packageOverrides = lib.foldr lib.composeExtensions (self: super: { }) ([ packageOverrides ] ++ defaultOverrides);
+  };
+in with py.pkgs; buildPythonPackage rec {
+  pname = "ludios_wpull";
+  version = "5.0.3";
+
+  disabled = pythonOlder "3.12";
+
+  format = "pyproject";
```
kirillrdy, line 47:
```suggestion
  pyproject = true;
```

this will use pypa to build package

pkgs/development/python-modules/ludios_wpull/default.nix
```diff
@@ -0,0 +1,68 @@
...
+  propagatedBuildInputs = [ chardet dnspython html5-parser lxml namedlist psutil setuptools py.pkgs.sqlalchemy tornado yapsy ];
+
+  # Test suite has tests that fail on all platforms
+  doCheck = false;
+
+  meta = {
+    description = "Web crawler; fork of wpull used by grab-site";
+    homepage = "https://github.com/ArchiveTeam/ludios_wpull";
+    license = lib.licenses.gpl3;
+    maintainers = with maintainers; [ ivan ];
```
kirillrdy, line 65:
```suggestion
    maintainers = with lib.maintainers; [ ivan ];
```

pkgs/development/python-modules/yapsy/default.nix
```diff
@@ -26,6 +27,8 @@ buildPythonPackage {
 
   pythonImportsCheck = [ "yapsy" ];
 
+  propagatedBuildInputs = [ python.pkgs.packaging ];
```
kirillrdy, line 30: you can get `packaging` via arguments to this derivation

pkgs/development/python-modules/ludios_wpull/default.nix
```diff
@@ -0,0 +1,68 @@
...
+          hash = "sha256-UIQlYsiZk1fzxDxKHNTBL06Ua/xPZQ15H5coEdkWc90=";
+        };
+      });
+    })
+  ];
+  py = python.override {
+    # Put packageOverrides at the start so they are applied after defaultOverrides
+    packageOverrides = lib.foldr lib.composeExtensions (self: super: { }) ([ packageOverrides ] ++ defaultOverrides);
+  };
+in with py.pkgs; buildPythonPackage rec {
```
kirillrdy, line 41:
```suggestion
in with py.pkgs; buildPythonApplication rec {
```

pkgs/by-name/gr/grab-site/package.nix
```diff
@@ -0,0 +1,47 @@
+{ lib
+, python312
```
kirillrdy, line 2: this should be set in all-packages.nix

https://github.com/NixOS/nixpkgs/blob/master/pkgs/by-name/README.md#changing-implicit-attribute-defaults

pastly: I believe I have addressed this one correctly. It's the main thing I doubt myself on regarding your latest feedback.

I didn't do it like your link instructs b/c it seemed like other packages in all-packages.nix with the same problem were using `python312Packages.callPackage` to solve it.

pkgs/development/python-modules/ludios_wpull/default.nix
pastly: Moved, thanks. And thank for your patience.

pkgs/by-name/gr/grab-site/package.nix
pastly: ... okay that didn't work. I've done it like these docs say. I finally figured out the right combo of `python___` stuff.

kirillrdy:
@pastly so currently this PR causes a lot of rebuilds, which means it should target staging branch.

staging already has tornado 6.4.1, so you can drop this patch

to read more about branches workflow https://github.com/NixOS/nixpkgs/blob/master/CONTRIBUTING.md#staging

please note a gotcha when changing target branch for a PR https://github.com/NixOS/nixpkgs/blob/master/CONTRIBUTING.md#rebasing-between-branches-ie-from-master-to-staging

kirillrdy:
tornado 6.4.1 is now on master, i recommend to target master for this PR

don't forget the gotchata with changing target branches

ivan:
@pastly thanks for working on this, do you want someone else to take this up, or was it accidentally closed?

pastly:
> @pastly thanks for working on this, do you want someone else to take this up, or was it accidentally closed, or did you intend to make a new PR for `master`?

The process was discouraging and I've temporarily/permanently tabled the project I was working on using these packages. 

No plans to make a PR for master, but anyone else is welcome to do so.

PR #321663, opened 2024-06-22
amdgpu: add kernelModule.patches option

Atemu:
## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/nixos/modules/services/hardware/amdgpu-kernel-module.nix
+++ b/nixos/modules/services/hardware/amdgpu-kernel-module.nix
@@ -1,0 +1,42 @@
+# TODO there should probably be a generic mechanism to patch any in-kernel module like this
+{
+  stdenv,
+  kernel, # The kernel to patch
+  patches ? [ ],
+}:
+
+stdenv.mkDerivation {
+  pname = "amdgpu-kernel-module-customised";
+  inherit (kernel)
+    src
+    version
+    postPatch
+    nativeBuildInputs
+    modDirVersion
+    ;
+  patches = kernel.patches or [ ] ++ patches;
+
+  modulePath = "drivers/gpu/drm/amd/amdgpu";
+
+  buildPhase = ''
+    BUILT_KERNEL=${kernel.dev}/lib/modules/$modDirVersion/build
+
+    cp $BUILT_KERNEL/Module.symvers $BUILT_KERNEL/.config ${kernel.dev}/vmlinux ./
+
+    make "-j$NIX_BUILD_CORES" modules_prepare
+    make "-j$NIX_BUILD_CORES" M=$modulePath modules
+  '';
+
+  installPhase = ''
+    make \
+      INSTALL_MOD_PATH="$out" \
+      XZ="xz -T$NIX_BUILD_CORES" \
+      M="$modulePath" \
+      modules_install
+  '';
+
+  meta = {
+    description = "AMDGPU kernel module";
+    inherit (kernel.meta) license;
+  };
+}
--- a/nixos/modules/services/hardware/amdgpu.nix
+++ b/nixos/modules/services/hardware/amdgpu.nix
@@ -117,6 +117,9 @@ in
   };
 
   meta = {
-    maintainers = with lib.maintainers; [ johnrtitor ];
+    maintainers = with lib.maintainers; [
+      johnrtitor
+      atemu
+    ];
   };
 }
--- a/nixos/modules/services/hardware/amdgpu.nix
+++ b/nixos/modules/services/hardware/amdgpu.nix
@@ -42,6 +42,36 @@ in
         This allows for an earlier modeset to apply the preferred resolution in
         the beginning of the initramfs phase rather than after it.
       '';
+
+      patches = lib.mkOption {
+        type = with lib.types; listOf path;
+        default = [ ];
+        description = ''
+          Patches to apply to the kernel for the `amdgpu` kernel module build.
+
+          This is intended for applying small patches concerning only the
+          `amdgpu` module's internals without needing to rebuild the entire
+          kernel.
+
+          The patches are applied to the entire kernel tree but only the
+          `amdgpu` module will actually be built and used. You should therefore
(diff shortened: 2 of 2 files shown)


nixos/modules/services/hardware/amdgpu-kernel-module.nix
```diff
@@ -0,0 +1,42 @@
+# TODO there should probably be a generic mechanism to patch any in-kernel module like this
```
JohnRTitor, line 1: I am not really keen on the idea, which problems would it solve specifically for `amdgpu` module? 

If anything like this to be introduced, I think it should be standardised and available for all modules. Perhaps @K900 wants to take a look here as well.

K900:
Agreed. This could be some sort of a generic mechanism if we really want it to exist.

nixos/modules/services/hardware/amdgpu-kernel-module.nix
Atemu: I had two issues that I needed to patch the amdgpu module for so far:

1. SteamVR needs to be able to create a high-priority queue but you can't give it CAP_SYS_NICE because we run Steam inside of a userns. I must patch the amdgpu kernel module to allow any process to create such a queue (and potentially DOS the system but idc).
2. The Framework 16 Laptop has a quirk and the display brightness does not go as low as it should and the lowest is too bright for dim environments. Until FW releases a firmware fix for this (which would be the proper fix), the module must be patched to add the quirk.

I'd also like this to be a standard option but I think that can come later. No need to perfect it in the V1.

K900: I really, really don't want to be setting this precedent.

Atemu: While I could certainly see this being a footgun, I don't really see any great danger here; could you elaborate?

I was also planning on making the unprivileged high-priority queue patch into a stand-alone option and providing some hacks in the Steam module to make the SteamVR setup experience smoother.

https://github.com/Atemu/nixos-config/blob/93b9546d1698286367bdb8826a3d827d5527128c/modules/gaming/module.nix#L116-L126

nyabinary: I think a generic mechanism would be nice, is there any way to unstall this?

Scrumplex: +1 on the generic mechanism.

I think a challenge here could be to make it easy to configure for users (i.e. `boot.kernelModules."amdgpu".patches = [...]`), as we would somehow need to map module names to module paths within the kernel source tree.

Atemu: Building a good interface here would be hard as some modules that are conceptually one unit may be constituted of multiple individual module files.

That's at least part of the reason why I limited my work to AMDGPU first. I'd prefer if we just merged this as iteration 1 and came up with a good design for a generic mechanism in iteration 2.

I know it's hard but we need to put a limit on our perfectionism sometimes ;)

@K900 could you share your doubts about "setting this precedent"?

K900: Mostly my concern is that we're giving users a tool that's very sharp and very easy to hold wrong, and the failure mode is not great.

K900: Also, the initial goal of this, that being the capability issues, should really be fixed somewhere at a higher level (possibly in bwrap itself even?)

Scrumplex: > Also, the initial goal of this, that being the capability issues, should really be fixed somewhere at a higher level (possibly in bwrap itself even?)

See https://github.com/NixOS/nixpkgs/issues/217119

In short:
Applications inside of user namespaces can never have effective file capabilities. This is just a design decision of the kernel. There have been numerous attempts to move away from capabilities specifically for high priority DRM contexts (which is what SteamVR needs) but none of them have made it into the kernel yet.

JohnRTitor (approved):
Now that I think about it, it might be good to merge this PR as an initial option and then try to look for a general solution.

LGTM approved, if you are still working on it, could you rebase it?

nixos/modules/services/hardware/amdgpu.nix
```diff
@@ -39,10 +34,25 @@ in
       };
     };
 
+    kernelModule = {
+      inInitrd = lib.mkEnableOption ''
```
Scrumplex, line 38: I don't feel like `kernelModule.inInitrd` is a good name for this option. But I also don't have a better idea than just going back to `initrd.enable`

JohnRTitor: Agreed.

nixos/modules/services/hardware/amdgpu.nix
```diff
@@ -39,10 +34,58 @@ in
...
+        installing the `amdgpu` kernelModule into the initrd; making it
+        available in stage 1 of the boot process.
+
+        This allows for an earlier modeset to apply the preferred resolution in
+        the beginning of the initramfs phase rather than after it.
+      '';
+
+      patches = lib.mkOption {
```
JohnRTitor, line 46: `hardware.amdgpu.modulePatches` can be used while keeping the current structure.

JohnRTitor (commented):
Since you have rebased the PR, and @Atemu has been inactive, I assume you want to adopt this PR? That's fine with me. Or you can also clone this branch and open a new PR.

nixos/modules/services/hardware/amdgpu.nix
JohnRTitor: Agreed.

PR #320202, opened 2024-06-16
ollama: remove rocblas override

Tungsten842:
## Description of changes

Fixes: https://discourse.nixos.org/t/rocblas-6-0-2-failure-to-build-with-ollama-acceleration-rocm-is-broken-with-24-05-upgrade/47055

Ollama works just fine without the rocblas override, not sure why it was there in the first place.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
- For non-Linux: Is sandboxing enabled in `nix.conf`? (See [Nix manual](https://nixos.org/manual/nix/stable/command-ref/conf-file.html))
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:
  - [NixOS test(s)](https://nixos.org/manual/nixos/unstable/index.html#sec-nixos-tests) (look inside [nixos/tests](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests))
  - and/or [package tests](https://github.com/NixOS/nixpkgs/blob/master/pkgs/README.md#package-tests)
  - or, for functions and "core" functionality, tests in [lib/tests](https://github.com/NixOS/nixpkgs/blob/master/lib/tests) or [pkgs/test](https://github.com/NixOS/nixpkgs/blob/master/pkgs/test)
  - made sure NixOS tests are [linked](https://nixos.org/manual/nixpkgs/unstable/#ssec-nixos-tests-linking) to the relevant packages
- [ ] Tested compilation of all packages that depend on this change using `nix-shell -p nixpkgs-review --run "nixpkgs-review rev HEAD"`. Note: all changes have to be committed, also see [nixpkgs-review usage](https://github.com/Mic92/nixpkgs-review#usage)
- [x] Tested basic functionality of all binary files (usually in `./result/bin/`)
- [24.11 Release Notes](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2411.section.md) (or backporting [23.11](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2311.section.md) and [24.05](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2405.section.md) Release notes)
  - [ ] (Package updates) Added a release notes entry if the change is major or breaking
[...]

PR #321664, opened 2024-06-22
nixos/ollama: add option for `HSA_OVERRIDE_GFX_VERSION`

abysssol:
## Description of changes

There are several GPUs that ROCm doesn't officially support but will work correctly if ROCm is directed to treat the GPU as a different one that is supported and has a similar architecture.

This can be done by setting `HSA_OVERRIDE_GFX_VERSION`. Ollama has documentation on this topic: https://github.com/ollama/ollama/blob/main/docs/gpu.md#amd-radeon

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [x] Tested, as applicable:
  - [x] `nix build -f pkgs/top-level/release.nix release-checks`
  - [x] `nix build -f default.nix ollama.tests`
- [x] Tested compilation of all packages that depend on this change using `nix run nixpkgs#nixpkgs-review -- rev HEAD`. Note: all changes have to be committed, also see [nixpkgs-review usage](https://github.com/Mic92/nixpkgs-review#usage)

---

### DIFF
--- a/pkgs/by-name/ol/ollama/package.nix
+++ b/pkgs/by-name/ol/ollama/package.nix
@@ -109,12 +109,6 @@ let
     ];
   };
 
-  runtimeLibs = lib.optionals enableRocm [
-    rocmPath
-  ] ++ lib.optionals enableCuda [
-    linuxPackages.nvidia_x11
-  ];
-
   appleFrameworks = darwin.apple_sdk_11_0.frameworks;
   metalFrameworks = [
     appleFrameworks.Accelerate
@@ -122,6 +116,17 @@ let
     appleFrameworks.MetalKit
     appleFrameworks.MetalPerformanceShaders
   ];
+
+  runtimeLibs = lib.optionals enableRocm [
+    rocmPath
+  ] ++ lib.optionals enableCuda [
+    linuxPackages.nvidia_x11
+  ];
+  wrapperOptions = builtins.concatStringsSep " " ([
+    "--suffix LD_LIBRARY_PATH : '/run/opengl-driver/lib:${lib.makeLibraryPath runtimeLibs}'"
+  ] ++ lib.optionals enableRocm [
+    "--set-default HIP_PATH '${rocmPath}'"
+  ]);
 
 
   goBuild =
@@ -183,10 +188,7 @@ goBuild ((lib.optionalAttrs enableRocm {
   '' + lib.optionalString (enableRocm || enableCuda) ''
     # expose runtime libraries necessary to use the gpu
     mv "$out/bin/ollama" "$out/bin/.ollama-unwrapped"
-    makeWrapper "$out/bin/.ollama-unwrapped" "$out/bin/ollama" ${
-      lib.optionalString enableRocm
-        ''--set-default HIP_PATH '${rocmPath}' ''} \
-      --suffix LD_LIBRARY_PATH : '/run/opengl-driver/lib:${lib.makeLibraryPath runtimeLibs}'
+    makeWrapper "$out/bin/.ollama-unwrapped" "$out/bin/ollama" ${wrapperOptions}
   '';
 
   ldflags = [
--- a/nixos/modules/services/misc/ollama.nix
+++ b/nixos/modules/services/misc/ollama.nix
@@ -98,7 +98,23 @@ in
             - otherwise defaults to `false`
           - `false`: disable GPU, only use CPU
           - `"rocm"`: supported by most modern AMD GPUs
+            - may require overriding gpu type with `services.ollama.rocmOverrideGfx`
+              if rocm doesn't detect your AMD gpu
           - `"cuda"`: supported by most modern NVIDIA GPUs
+        '';
+      };
+      rocmOverrideGfx = lib.mkOption {
+        type = types.nullOr types.str;
+        default = null;
+        example = "10.3.0";
+        description = ''
+          Override what rocm will detect your gpu model as.
+          For example, make rocm treat your RX 5700 XT (or any other model)
+          as an RX 6900 XT using a value of `"10.3.0"` (gfx 1030).
+
+          This sets the value of `HSA_OVERRIDE_GFX_VERSION`. See [ollama's docs](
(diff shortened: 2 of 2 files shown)


drupol (dismissed):

pkgs/by-name/ol/ollama/package.nix
```diff
@@ -185,7 +185,8 @@ goBuild ((lib.optionalAttrs enableRocm {
     mv "$out/bin/ollama" "$out/bin/.ollama-unwrapped"
     makeWrapper "$out/bin/.ollama-unwrapped" "$out/bin/ollama" ${
       lib.optionalString enableRocm
-        ''--set-default HIP_PATH '${rocmPath}' ''} \
+        ''--set-default HIP_PATH '${rocmPath}' \
+          --set-default HSA_OVERRIDE_GFX_VERSION 0 ''} \
```
Aleksanaa, line 189: What does 0 mean here?

drupol: Ho good catch !

abysssol: It's essentially garbage. It could just as well be a "1" or "." or "-". I added it because of its interaction with rocblas compiled without tensile files.

See #321335 and #320202 for details, tldr rocblas compiled with tensile files (large lazy loaded gpu specific data (and/or libraries?)) takes a lot of processing, memory, and storage to compile, and hydra won't cache it because it's too large or something like that. Most ollama users are unable to compile it, their machines choking and even crashing. So, rocblas was recently changed back to not compile tensile files, which works fine for some gpus, but for others (including mine), rocm becomes unable to detect the gpu unless HSA_OVERRIDE_GFX_VERSION is set.

I very recently figured out that the actual value of HSA_OVERRIDE_GFX_VERSION doesn't actually matter for my gpu (6950 xt), and any nonempty string will enable rocm for gfx 1030, while an empty string or being unset will cause ollama to only use cpu. Thus the garbage value of "0" causes rocm to work correctly for me without any further configuration, and my hope is that this will also apply to other users, allowing them to use rocm without having to figure out some arcane override (it's not like I really care about my own setup working with ollama, I don't actually even use ollama).

abysssol:
@kira-bruneau since your gpu doesn't work with rocm without HSA_OVERRIDE_GFX_VERSION, could you test ollama-rocm from this pr and see if rocm works correctly without you setting HSA_OVERRIDE_GFX_VERSION manually?

See my [earlier comment](https://github.com/NixOS/nixpkgs/pull/321664#discussion_r1649727925) for details, but essentially HSA_OVERRIDE_GFX_VERSION needs to be unset (**not** just set to an empty string) for the default value to be set in ollama's wrapper script. In bash thats `unset HSA_OVERRIDE_GFX_VERSION`, in fish it's `set -e HSA_OVERRIDE_GFX_VERSION`.

Alternatively, you could set HSA_OVERRIDE_GFX_VERSION to "0", the same as what the wrapper would use as a default, if that's more convenient.

You may also need to set `OLLAMA_HOST=127.0.0.1:11435` if you already have the ollama service running.

I'm hoping since both our gpus aren't used by rocm without setting HSA_OVERRIDE_GFX_VERSION, and mine works even when setting it to a garbage value, your gpu will also work without needing to manually define HSA_OVERRIDE_GFX_VERSION as a valid gfx version. If so, then that's good evidence that many other users' gpus will behave similarly, making using rocm easier for many more people.

pkgs/by-name/ol/ollama/package.nix
Aleksanaa: > rocm becomes unable to detect the gpu unless HSA_OVERRIDE_GFX_VERSION is set

Could you add this to a comment near it?

kira-bruneau:
Oh I just tried and ended up with a coredump:

```
Error: llama runner process has terminated: signal: aborted (core dumped) error:Could not initialize Tensile host: No devices found
```

I think it would be better to leave it up to the user to configure this override.

abysssol:
@kira-bruneau I see ... that's unfortunate. Thank you for testing it!

drupol (approved):
LGTM

kira-bruneau:
Result of `nixpkgs-review pr 321664` run on x86_64-linux [1](https://github.com/Mic92/nixpkgs-review)
<details>
  <ul>
    <li>nixos-install-tools</li>
  </ul>
</details>
<details>
  <ul>
    <li>ollama-cuda</li>
    <li>ollama-rocm</li>
  </ul>
</details>

kira-bruneau (approved):

caffineehacker:
FYI - I think this may have broken existing users that set environmentVariables.HSA_OVERRIDE_GFX_VERSION since this now overwrites that value rather than defaulting to what the user had already set.

I believe using mkMerge will work correctly, but I'll need to do some testing. I'll try to get a PR tonight if I have time, but anyone else can feel free to fix this sooner.

Aleksanaa:
> FYI - I think this may have broken existing users that set environmentVariables.HSA_OVERRIDE_GFX_VERSION since this now overwrites that value rather than defaulting to what the user had already set.

It only sets the value when the new option is set to something other than `null`.

caffineehacker:
> > FYI - I think this may have broken existing users that set environmentVariables.HSA_OVERRIDE_GFX_VERSION since this now overwrites that value rather than defaulting to what the user had already set.
> 
> It only sets the value when the new option is set to something other than `null`.

Unfortunately that's not how lib.mkIf works. I believe you need to use lib.mkMerge if you're using lib.mkIf. For example, try running this in nix repl:

```
nix-repl> ({ b = 1; } // { b = lib.mkIf false 2; }).b
{ _type = "if"; condition = false; content = 2; }
```

b is no longer 1 even though lib.mkIf had false as the first argument. I won't have time to send a PR tonight, but I should tomorrow.

PR #248334, opened 2023-08-10
nixos/nextcloud: Add ensureUsers option

onny:
## Description of changes

Aiming to get rid of the <code>config</code> option in the Nextcloud module, having them all in the free form type <code>extraOptions</code> which can be renamed to <code>settings</code> in accordance to RFC42 in a later PR.

Instead of defining a single admin user with <code>config.adminuser</code> and <code>config.adminpassFile</code> switch to a more generic approach which is used by other modules, having a <code>ensureUsers</code> attribute set.

```
services.nextcloud = {
  enable = true;
  package = pkgs.nextcloud27;
  hostName = "localhost";
  config = {
    adminpassFile = "${pkgs.writeText "adminpass" "hunter2"}";
  };
  database.createLocally = true;
  ensureUsers = {
    user1 = {
      email = "user1@localhost";
      passwordFile = /secrets/user1.secret;
    };
    user2 = {
      email = "user1@localhost";
      passwordFile = /secrets/user1.secret;
    };
  };
};
```

For now I would like to only add the `ensureUsers` option to add additional users. In a later PR I would like to manage admin user configuration with this option.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
- [ ] For non-Linux: Is `sandbox = true` set in `nix.conf`? (See [Nix manual](https://nixos.org/manual/nix/stable/command-ref/conf-file.html))
- [x] Tested, as applicable:
  - [NixOS test(s)](https://nixos.org/manual/nixos/unstable/index.html#sec-nixos-tests) (look inside [nixos/tests](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests))
  - and/or [package tests](https://nixos.org/manual/nixpkgs/unstable/#sec-package-tests)
  - or, for functions and "core" functionality, tests in [lib/tests](https://github.com/NixOS/nixpkgs/blob/master/lib/tests) or [pkgs/test](https://github.com/NixOS/nixpkgs/blob/master/pkgs/test)
  - made sure NixOS tests are [linked](https://nixos.org/manual/nixpkgs/unstable/#ssec-nixos-tests-linking) to the relevant packages
[...]

PR #321771, opened 2024-06-22
nixos/nextcloud: Remove --preserve-env in sudo

Mynacol:
## Description of changes

This helps supporting sudo-rs, which currently does not implement the --preserve-env flag and probably won't so in the foreseeable future [1].

The replacement just sets both environment variables behind the sudo invocation, which works with exec and both sudo implementations.

[1] https://github.com/memorysafety/sudo-rs/issues/129

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/nixos/modules/services/web-apps/nextcloud.nix
+++ b/nixos/modules/services/web-apps/nextcloud.nix
@@ -91,10 +91,10 @@ let
     cd ${webroot}
     sudo=exec
     if [[ "$USER" != nextcloud ]]; then
-      sudo='exec /run/wrappers/bin/sudo -u nextcloud --preserve-env=NEXTCLOUD_CONFIG_DIR --preserve-env=OC_PASS'
+      sudo='exec /run/wrappers/bin/sudo -u nextcloud'
     fi
-    export NEXTCLOUD_CONFIG_DIR="${datadir}/config"
-    $sudo \
+    $sudo ${pkgs.coreutils}/bin/env \
+      NEXTCLOUD_CONFIG_DIR="${datadir}/config" \
       ${phpCli} \
       occ "$@"
   '';


Ma27:
Would be fine by me.

Shawn8901:
It seems to break the `nextcloud-occ` when being run as nextcloud user

```log
[nextcloud@tank:~]$ nextcloud-occ
/run/current-system/sw/bin/nextcloud-occ: line 8: /nix/store/266vrngrbyyc2hnn8j7lak2y4nm70sa2-nextcloud-29.0.2-with-apps/NEXTCLOUD_CONFIG_DIR=/persist/var/lib/nextcloud/config: No such file or directory

[nextcloud@tank:~]$
```

Ma27:
To be clear, I agree with the idea, I haven't taken a close look whether the new invocation is actually correct currently :grin:

Shawn8901:
From my side all good 😊, I am also the option that it's a nice idea to keep compatible with sudo-rs if possible, so I wanted to test it out.

Wrapper invocation for non nextcloud users work fine.

SuperSandro2000:
Wouldn't this expose the pass env in the exec args and be displayed in programs like htop?

Why can't they implement this simple feature? This could be done in a few lines shell script.

Ma27:
> Wouldn't this expose the pass env in the exec args and be displayed in programs like htop?

Dammit, I forgot about OC_PASS. However, is there any reason this even exists? Just checked the sources and it's only used for adding a user or resetting someone's password, however other alternatives are also OK.

Can't remember having it used a single time.

Shawn8901:
> From my side all good 😊, I am also the option that it's a nice idea to keep compatible with sudo-rs if possible, so I wanted to test it out.
> 
> Wrapper invocation for non nextcloud users work fine.

To make it compatible with the `exec` command we have either to do 

```
$sudo env NEXTCLOUD_CONFIG_DIR="${datadir}/config"   OC_PASS="$OC_PASS"
```

or 

```
NEXTCLOUD_CONFIG_DIR="${datadir}/config"   OC_PASS="$OC_PASS" $sudo
```

Tho in case we keep OC_PASS , we should not expose it.

OC_PASS is used for `--password-from-env` (like @Ma27 already mentioned) --> https://docs.nextcloud.com/server/latest/admin_manual/configuration_server/occ_command.html
That could be a nice thing for creating users declarative (in case we want to do that at some point as we could also query the available users and create it from a password file if not existing or so [just wild guessing for what we could use it in our context]).

Ma27:
> That could be a nice thing for creating users declarative

Just want to mention here that if this suggestion is to be some kind of ensure*-style option, then it isn't declarative and I'm strongly against adding it to this module. See also https://github.com/NixOS/nixpkgs/pull/248334

If the VM tests still pass with OC_PASS removed (i.e. installing Nextcloud still works without it), I'd suggest to just drop it.

Shawn8901:
yeah, then we should remove it, totally agreeing from my end.

nixos/modules/services/web-apps/nextcloud.nix
```diff
@@ -91,10 +91,10 @@ let
     cd ${webroot}
     sudo=exec
     if [[ "$USER" != nextcloud ]]; then
-      sudo='exec /run/wrappers/bin/sudo -u nextcloud --preserve-env=NEXTCLOUD_CONFIG_DIR --preserve-env=OC_PASS'
+      sudo='exec /run/wrappers/bin/sudo -u nextcloud'
     fi
-    export NEXTCLOUD_CONFIG_DIR="${datadir}/config"
-    $sudo \
+    $sudo env \
```
Ma27, line 96: This won't work if you invoke `nextcloud-occ` in a custom systemd unit since those require you to set $PATH explicitly (otherwise it'll be empty).

In other words, I think it makes sense to do `${pkgs.coreutils}/bin/env` here.

Ma27 (approved):

PR #321785, opened 2024-06-22
libgnome-keyring: Unify

jtojnar:
## Description of changes

No need to have two API compatible copies of this ancient and deprecated library.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---


pkgs/development/haskell-modules/hackage-packages.nix
```diff
@@ -126016,22 +126016,21 @@ self: {
...
        sha256 = "044bbgy8cssi1jc8wwb0kvxpw6d7pwxackkzvw7p9r8ybmgv4d0b";
        libraryHaskellDepends = [ base bytestring text time ];
-       librarySystemDepends = [ gnome-keyring ];
+       librarySystemDepends = [ libgnome-keyring ];
        libraryPkgconfigDepends = [ libgnome-keyring ];
        libraryToolDepends = [ c2hs ];
        description = "Bindings for libgnome-keyring";
        license = lib.licenses.gpl3Only;
        badPlatforms = lib.platforms.darwin;
-     }) {inherit (pkgs.gnome) gnome-keyring;
```
jtojnar, line 126033: Opened https://github.com/NixOS/nixpkgs/pull/321785 for that.

jtojnar: Opened https://github.com/NixOS/cabal2nix/pull/625 for that.

pkgs/desktops/gnome/core/libgnome-keyring/default.nix
```diff
@@ -13,9 +13,23 @@ stdenv.mkDerivation (finalAttrs: {
 
   outputs = [ "out" "dev" ];
 
+  strictDeps = true;
   propagatedBuildInputs = [ glib gobject-introspection dbus libgcrypt ];
   nativeBuildInputs = [ pkg-config intltool ];
 
+  configureFlags = [
+    # not ideal to use -config scripts but it's not possible switch it to pkg-config
+    # binaries in dev have a for build shebang
```
doronbehar, line 22: I don't understand the end of this sentence, but for sure this word is missing:

```suggestion
    # not ideal to use -config scripts but it's not possible to switch it to pkg-config
```

pkgs/desktops/gnome/core/libgnome-keyring/default.nix
```diff
@@ -13,9 +13,23 @@ stdenv.mkDerivation (finalAttrs: {
 
   outputs = [ "out" "dev" ];
 
+  strictDeps = true;
   propagatedBuildInputs = [ glib gobject-introspection dbus libgcrypt ];
   nativeBuildInputs = [ pkg-config intltool ];
```
doronbehar, line 18: Maybe we should try to use `autoreconfHook`?

pkgs/top-level/all-packages.nix
```diff
@@ -22032,7 +22032,7 @@ with pkgs;
 
   libglibutil = callPackage ../development/libraries/libglibutil { };
 
-  libgnome-keyring = callPackage ../development/libraries/libgnome-keyring { };
+  libgnome-keyring = gnome.libgnome-keyring;
```
doronbehar, line 22035: > Those libraries actually have the same API.

Hmm I tend to think maybe we should use a `lib.warn` alias instead? Maybe there are other differences we are not aware of?

pkgs/top-level/aliases.nix
```diff
@@ -673,6 +673,7 @@ mapAliases ({
   libcxxabi = throw "'libcxxabi' was merged into 'libcxx'"; # Converted to throw 2024-03-08
   libdwarf_20210528 = throw "'libdwarf_20210528' has been removed because it is not used in nixpkgs, move to libdwarf"; # Added 2024-03-23
   libgme = game-music-emu; # Added 2022-07-20
+  libgnome-keyring3 = libgnome-keyring; # Added 2024-06-22
```
doronbehar, line 676: Maybe this should be part of the previous commit.

doronbehar (commented):
Looks good overall.

pkgs/desktops/gnome/core/libgnome-keyring/default.nix
jtojnar: I do not want to touch this. It should die already.

pkgs/desktops/gnome/core/libgnome-keyring/default.nix
jtojnar: Same here, I just cherry-picked @Artturin’s commit and gave up on it.

I believe it is supposed to be parsed “binaries in dev have (a (for build) shebang)”

pkgs/top-level/all-packages.nix
jtojnar: Well, one difference is that the newer version marks everything deprecated.

pkgs/top-level/aliases.nix
jtojnar: I think a separate commit would be cleanest but I just cannot be bothered.

pkgs/top-level/all-packages.nix
doronbehar: > Well, one difference is that the newer version marks everything deprecated so it can cause problems if some project has deprecations disabled. But I do not expect that to happen – most consumers do not even include the headers, they just dlopen the library. And I would expect sizeable portion to be unused – the library has been deprecated for over 10 years.
> 
> Here is the abidiff for peace of mind:

OK Great it's nice to get confidence that this won't break anything, but still, the reason to add a `lib.warn` is to distribute the update - that's our role as a distribution IMO (same arguments as in the ["How to introduce a breaking version bump" discourse thread](https://discourse.nixos.org/t/how-to-introduce-a-breaking-version-bump/46746)).

pkgs/desktops/gnome/core/libgnome-keyring/default.nix
doronbehar: > I do not want to touch this. It should die already.

Oh I think the 1st commit's title is misleading - you are in fact editing the expression of the attribute `libgnome-keyring`, and not `libgnome-keyring3`.

doronbehar:
If the 1st commit has only a misleading title, but correct contents, it can be removed IMO.

pkgs/desktops/gnome/core/libgnome-keyring/default.nix
jtojnar: It is modifying `gnome.libgnome-keyring` (`libgnome-keyring3`) at the time. The package is only renamed later.

pkgs/top-level/all-packages.nix
jtojnar: I do not think there is a consensus on `lib.warn` yet so I decided to go with status quo for `aliases.nix` and not add it here.

jtojnar:
> If the 1st commit has only a misleading title, but correct contents, it can be removed IMO.

Not sure why you think it an be removed.

My primary concern when deciding on commit contents is ease of reviewing and similar `git blame` archaeology. As a result, I tend to make commits minimal, to be reviewed one-by-one. Each commit applied individually (after all dependencies were already applied) should also keep the tree in a good state (again to keep bisection working), which enforces topological ordering on the commits.

pkgs/desktops/gnome/core/libgnome-keyring/default.nix
doronbehar: > I do not want to touch this. It should die already.

This is the comment that confused me BTW - why should it die?

doronbehar:
> My primary concern when deciding on commit contents is ease of reviewing and similar `git blame` archaeology.

Yea I agree with this goal, I just got confused between the two version of the `libgnome-keyring` package - I thought you edit one of them, and right afterwards delete them.

pkgs/desktops/gnome/core/libgnome-keyring/default.nix
doronbehar: > Same here, I just cherry-picked @Artturin’s commit and gave up on it.
> 
> I believe it is supposed to be parsed “binaries in dev have (a (for build) shebang)”

@Artturin gave their :+1: for rephrasing that sentence as you said.

pkgs/desktops/gnome/core/libgnome-keyring/default.nix
jtojnar: It adds extra maintenance work just to support few apps that could not be bothered to migrate to `libsecret` in the 10+ years libgnome-keyring has been deprecated.

jtojnar:
The guidelines require an existing attribute name as the commit header tag, so renames should be `new: rename from old`. I agree that it can be a bit confusing.

pkgs/desktops/gnome/core/libgnome-keyring/default.nix
doronbehar: > It adds extra maintenance work just to support few apps that could not be bothered to migrate to `libsecret` in the 10+ years libgnome-keyring has been deprecated.

Ahh OK so `libgnome-keyring` is replaced by `libgnome-keyring3` which should really be replaced by `libsecret`.

Still though, if we don't have that many packages still depending on `libgnome-keyring` or `libgnome-keyring3`, I'd make the effort to try to add `autoreconfHook`, but I don't care too much about this old library to insist on that :).

PR #234651, opened 2023-05-28
lib.extendMkDerivation: init

ShamrockLee:
###### Description of changes

This PR introduces a unified approach to implementing build helpers that support fixed-point arguments and bring such support to existing build helpers.

The fixed-point arguments support in `stdenv.mkDerivation` is introduced in #119942, and there are ongoing attempts to make other build helpers support such syntax (`buildRustPackage` refactor #194475,  `buildGoModule` refactor #225051). The overlay style `overrideAttrs` brought by the `stdenv.mkDerivation`  change can be used to implement the functionality, which is adopted by the `buildRustPackage` refactor and the previous version of the `buildGoModule` refactor. The challenge of such an approach is that the whole [set pattern](https://nixos.org/manual/nix/stable/language/constructs.html#functions) matching the input arguments are degenerated into a single identifier, making it hard to see from the source code which attributes to the build helper accepts.

The new Nixpkgs Library function, `lib.extendMkDerivation` accepts a base build helper and an attribute overlay (an overlay in the form `finalAttrs: args: <attrsToUpdate>`), and returns a new build helper by extending the base build helper with the attribute overlay via its `<pkg>.overrideAttrs`.

The following is the definition of an example build helper, `mkLocalDerivation`:

```nix
lib.extendMkDerivation {
  constructDrv = stdenv.mkDerivation;
  extendDrvArgs =
    finalAttrs:
    {
      preferLocalBuild ? true,
      allowSubstitute ? false,
      ...
    }@args:

    # No need of `// args` here.
    # The whole set of input arguments is passed in advance.
    {
      # Attributes to update
      inherit preferLocalBuild allowSubstitute;
    };
}
```

For existing build helpers with arguments that cannot be passed to the base build helper, `lib.extendMkDerivation` provides an attribute `removedAttrNames` to specify arguments not to pass down to the base build helper.

```nix
lib.extendMkDerivation {
  constructDrv = stdenv.mkDerivation;
  extendDrvArgs =
    finalAttrs:
    {
      preferLocalBuild ? true,
      allowSubstitute ? false,
      specialArg ? (_: false),
      ...
    }@args:

    {
      # Arguments to pass
      inherit preferLocalBuild allowSubstitute;
      # Some expressions involving specialArg
      greeting = if specialArg "hi" then "hi" else "hello";
    };

  excludeDrvArgNames = [
    "specialArg"
  ];
}
```

[...]

PR #225051, opened 2023-04-06
buildGoModule: Fix overriding with overlay-style stdenv

ShamrockLee:
###### Description of changes

The Go package builder `buildGoModule` is refactored to adopt the overlay-style `stdenv.mkDerivation` (#119942) and to fix overriding. One can now override `vendorHash` using `overrideAttrs`. This addresses https://github.com/NixOS/nixpkgs/issues/86349#issuecomment-763914147

The `goModules` and `vendorHash` attributes are moved out of `passthru` as they are already exposed as `stdenv.mkDerivation` attributes. They can still be accessed as `pet.vendorHash`, but no longer `pet.passthru.vendorHash`. The builder now only adds `go` to `passthru`.

To address the issue of `vendorHash = null` through `overrideAttrs` may break the overriding of `goModules`, this PR also provides an overridable `passthru.overrideModAttrs` that will be applied to `goModules` when `vendorHash` is not `null`.

Closes #86349

Comparison to similar PRs:

In this PR, the diff of `pkgs/build-support/go/module.nix` is shorter than #171586.

###### Things done

- Built on platform(s)
  - [X] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
- [ ] For non-Linux: Is `sandbox = true` set in `nix.conf`? (See [Nix manual](https://nixos.org/manual/nix/stable/command-ref/conf-file.html))
- [ ] Tested, as applicable:
  - [NixOS test(s)](https://nixos.org/manual/nixos/unstable/index.html#sec-nixos-tests) (look inside [nixos/tests](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests))
  - and/or [package tests](https://nixos.org/manual/nixpkgs/unstable/#sec-package-tests)
  - or, for functions and "core" functionality, tests in [lib/tests](https://github.com/NixOS/nixpkgs/blob/master/lib/tests) or [pkgs/test](https://github.com/NixOS/nixpkgs/blob/master/pkgs/test)
  - made sure NixOS tests are [linked](https://nixos.org/manual/nixpkgs/unstable/#ssec-nixos-tests-linking) to the relevant packages
- [ ] Tested compilation of all packages that depend on this change using `nix-shell -p nixpkgs-review --run "nixpkgs-review rev HEAD"`. Note: all changes have to be committed, also see [nixpkgs-review usage](https://github.com/Mic92/nixpkgs-review#usage)
[...]

PR #321791, opened 2024-06-22
buildGoModule: accept fixed-point function argument

ghost:
Make it possible to use `buildGoModule (finalAttrs: {` instead of `buildGoModule rec {`.

I'm hoping this is a zero-rebuild change.

Related: #225051 #234651 #317559 

---

### DIFF
--- a/pkgs/build-support/go/module.nix
+++ b/pkgs/build-support/go/module.nix
@@ -1,6 +1,7 @@
 { go, cacert, git, lib, stdenv }:
 
-{ name ? "${args'.pname}-${args'.version}"
+let postArgsFunc =
+{ name ? "${prevAttrs.pname}-${prevAttrs.version}"
 , src
 , nativeBuildInputs ? [ ]
 , passthru ? { }
@@ -17,7 +18,7 @@
   # if vendorHash is null, then we won't fetch any dependencies and
   # rely on the vendor folder within the source.
 , vendorHash ? throw (
-    if args'?vendorSha256 then
+    if prevAttrs?vendorSha256 then
       "buildGoModule: Expect vendorHash instead of vendorSha256"
     else
       "buildGoModule: vendorHash is missing"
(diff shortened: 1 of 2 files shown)


pkgs/build-support/go/module.nix
Aleksanaa, line 1: Please split formatting to another commit.

Aleksanaa:
Support for `finalAttrs` is good, but replacing all occurrence, especially those only using `rev = version`, doesn't bring any immediate benefit. See https://github.com/NixOS/nixpkgs/issues/310373

Aleksanaa:
> Support for `finalAttrs` is good, but replacing all occurrence of `rec`, especially those only using `rev = version`, doesn't bring any immediate benefit. See #310373

ehmry:
The immediate benefit is that there is consistency, and consistency perpetuates more consistency because people look at other packages for examples.

ehmry:
I did this because its good to have the option of using `finalAttrs` and because it makes `buildGoModule` consistent with what we are recommending elsewhere.

Changing packages that are using `rec` shows that it can be done without effecting outputs, but it isn't necessary so the second commit doesn't need to be merged.

zowoq (dismissed):
> Please split formatting to another commit.

ehmry:
There are actually quite a few places where overriding `version` and `src` later isn't going to be correct, see `git grep -n -- '-X=g' pkgs`.

Mic92:
> The immediate benefit is that there is consistency, and consistency perpetuates more consistency because people look at other packages for examples.

A different problem is that it has a negative impact on eval performance - So I don't think we should use this everywhere in nixpkgs, but only if we need it.

philiptaron:
@emhry mind if I force-push this to fix the merge conflict?

ehmry:
@philiptaron yes, fine by me

pkgs/build-support/go/module.nix
```diff
@@ -156,8 +155,12 @@ let
     # error: empty hash requires explicit hash algorithm
     outputHashAlgo = if vendorHash == "" then "sha256" else null;
   }).overrideAttrs overrideModAttrs;
-
-  package = stdenv.mkDerivation (args // {
+in
+  (builtins.removeAttrs prevAttrs [
+    "overrideModAttrs"
+    "vendorSha256"
+    "vendorHash"
```
ShamrockLee, line 162:
```suggestion
    "vendorSha256"
    "vendorHash"
```

Now, we pass `vendorHash` directly as a derivation attribute. As for the long-deprecated `vendorSha256`, we should be able to ignore it, as people who haven't migrated to `vendorHash` would get an error. The only thing we need to filter out is `overrideModAttrs`.

ShamrockLee:
This commit needs rebasing, as #225051 is merged.

Issue #84312, opened 2020-04-05
fetchFromGitHub is not reproducible when export-subst is used


veprbl:
To investigate #71948 I made a following setup to try to catch the hash changes: A following script:
```bash
#!/bin/bash
  
. ~/.nix-profile/etc/profile.d/nix.sh

BASEDIR="$(dirname "$0")"
"$HOME"/.nix-profile/bin/nix-prefetch-url "$BASEDIR"/nixpkgs -A emacsPackages.lua-mode.src 1> "$BASEDIR"/`date +%j-%Y`.out 2> "$BASEDIR"/`date +%j-%Y`.err
```
Was run daily using cron, the `"$BASEDIR"/nixpkgs` was an immutable nixpkgs checkout. There were two observed changes:
```
# grep path *.err
010-2020.err:path is '/nix/store/lj9l6grfbh1d5pxs1kccbnsk6snmb4q4-source'
<snip>
037-2020.err:path is '/nix/store/lj9l6grfbh1d5pxs1kccbnsk6snmb4q4-source'
038-2020.err:path is '/nix/store/5cd6xbb5a31a1d9a0iyn3kbxxad8xd5v-source'
<snip>
092-2020.err:path is '/nix/store/5cd6xbb5a31a1d9a0iyn3kbxxad8xd5v-source'
093-2020.err:path is '/nix/store/iss0dg74shwiph6yaagfwl8b36p4yspn-source'
094-2020.err:path is '/nix/store/iss0dg74shwiph6yaagfwl8b36p4yspn-source'
095-2020.err:path is '/nix/store/iss0dg74shwiph6yaagfwl8b36p4yspn-source'
```
The first one looks like a version bump, but the second one looks more interesting:
```diff
diff -aru /nix/store/5cd6xbb5a31a1d9a0iyn3kbxxad8xd5v-source/lua-mode.el /nix/store/iss0dg74shwiph6yaagfwl8b36p4yspn-source/lua-mode.el
--- /nix/store/5cd6xbb5a31a1d9a0iyn3kbxxad8xd5v-source/lua-mode.el      1969-12-31 19:00:01.000000000 -0500
+++ /nix/store/iss0dg74shwiph6yaagfwl8b36p4yspn-source/lua-mode.el      1969-12-31 19:00:01.000000000 -0500
@@ -36,7 +36,7 @@
 
 ;; This field is expanded to commit SHA, date & associated heads/tags during
 ;; archive creation.
[...]

PR #321796, opened 2024-06-22
python3Packages.amaranth-boards: fix fetchFromGitHub hash stability

tpwrules:
cc @pbsds @mweinelt 

## Things done

- Built on platform(s)
  - [X] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/development/python-modules/amaranth-boards/default.nix
+++ b/pkgs/development/python-modules/amaranth-boards/default.nix
@@ -19,7 +19,9 @@ buildPythonPackage rec {
     owner = "amaranth-lang";
     repo = "amaranth-boards";
     rev = "aba2300dc83216523e1c98fdb22471cb4bac5027";
-    hash = "sha256-IMhNMkOf6dg4Djyi39TDTpU0S3TQM1+e6TBoyiITaos=";
+    # these files change depending on git branch status
+    postFetch = "rm -f $out/.git_archival.txt $out/.gitattributes";
+    hash = "sha256-jldXyMJtuSGcZKmtwpZBYrR/UBe4ufblPYRYpBmReM8=";
   };
 
   nativeBuildInputs = [


pkgs/development/python-modules/amaranth-boards/default.nix
```diff
@@ -19,7 +19,9 @@ buildPythonPackage rec {
     owner = "amaranth-lang";
     repo = "amaranth-boards";
     rev = "aba2300dc83216523e1c98fdb22471cb4bac5027";
-    hash = "sha256-IMhNMkOf6dg4Djyi39TDTpU0S3TQM1+e6TBoyiITaos=";
+    # these files change depending on git branch status
+    postFetch = "rm -f $out/.git_archival.txt $out/.gitattributes";
```
Aleksanaa, line 23: I found the context https://github.com/NixOS/nixpkgs/issues/84312

This is soooo bad.

tpwrules: Did not realize this was known to nixpkgs, thanks for linking. Yeah it's a really unfortunate mis-feature.

whitequark: I'm actually pretty sad about the removal of `.git_archival.txt` since it means that anyone using tarballs gets an archive which cannot be installed without additional knowledge (of what the version number is supposed to be). pdm-backend doesn't support this one way or another, but if it did I would have preferred to keep it...

tpwrules: I don’t know if there were improvements to be made to the actual implementation, but the version being not fully dependent on the git hash is strange to me to say the least.

whitequark: I am fairly confident there is a way to implement `.git_archival.txt` which would not change the tarball iff your versioning strategy (tag assignment) does not result in retroactively changing the version for a normal git checkout (which is also very bad for obvious reasons).

I am also fairly confident that export-subst is too crude of a mechanism to do that, so I do sympathize with both the Nix folks in particular, and with anyone complaining about reproducibility in general. I have just also had unpleasant conversations with people who reasonably expect tarballs to be buildable as-is, so it's being caught between a rock and a hard place.

pbsds:
@ofborg build python311Packages.amaranth-boards python312Packages.amaranth-boards python311Packages.amaranth-boards.passthru.tests python312Packages.amaranth-boards.passthru.tests

pkgs/development/python-modules/amaranth-boards/default.nix
lf-: fyi yes you can soundly do this, with the invariant just changing to "you don't tag parents of commits someone put in nixpkgs" which is a lot looser of an invariant i think. you want `%(describe:abbrev=10)`. it will only catch annotated tags.

whitequark: Thanks! That would interact really poorly with our existing git history in Amaranth since all of the tags are not annotated. Can I safely replace unannotated tags with annotated ones?

lf-: I think it should be perfectly sound to do to future ones and only then swap the substitution spec. Also there's a tags=true option on that descriptor if you want to also consider unannotated tags. and there's a filter you can use to which tags it considers. see PRETTY FORMATS in git-log(1).

I think annotating existing tags is probably safe as well. I expect it is likely safe for our code as well. 

I've not looked at the code but you may have to adjust the stuff consuming it because it will give you v1.0-50-abcdef0123 instead of just tag names on the current commit; in the case of it being a tag however it will just give you the tag itself.

whitequark: This will require a patch to `pdm-backend` which neither me nor Frost Ming seem to feel like doing, but it's very useful to have this for future reference; thanks!

PR #321871, opened 2024-06-23
emacs and org: backport a security fix

jian-lin:
## Description of changes

- emacs.pkgs.org: backport a security fix from 9.7.5
- emacs28: backport a security fix from Emacs 29.4

CC macport maintainer @Atemu 

## Things done

- Built on platform(s)
  - [X] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/applications/editors/emacs/elisp-packages/elpa-devel-packages.nix
+++ b/pkgs/applications/editors/emacs/elisp-packages/elpa-devel-packages.nix
@@ -64,6 +64,26 @@ self: let
         '';
       });
 
+      org = super.org.overrideAttrs (old: {
+        dontUnpack = false;
+        patches = old.patches or [ ] ++ [
+          # security fix backported from 9.7.5
+          (pkgs.fetchpatch {
+            url = "https://git.savannah.gnu.org/cgit/emacs/org-mode.git/patch/?id=f4cc61636947b5c2f0afc67174dd369fe3277aa8";
+            hash = "sha256-bGgsnTSn6SMu1J8P2BfJjrKx2845FCsUB2okcIrEjDg=";
+            stripLen = 1;
+          })
+        ];
+        postPatch = old.postPatch or "" + "\n" + ''
+          pushd ..
+          local content_directory=${old.ename}-${old.version}
+          src=$PWD/$content_directory.tar
+          tar --create --verbose --file=$src $content_directory
+          popd
+        '';
+        dontBuild = true;
+      });
+
       pq = super.pq.overrideAttrs (old: {
         buildInputs = (old.buildInputs or [ ]) ++ [ pkgs.postgresql ];
       });
--- a/pkgs/applications/editors/emacs/elisp-packages/elpa-devel-generated.nix
+++ b/pkgs/applications/editors/emacs/elisp-packages/elpa-devel-generated.nix
@@ -4453,10 +4453,10 @@
       elpaBuild {
         pname = "org";
         ename = "org";
-        version = "9.7pre0.20240521.135840";
+        version = "9.7pre0.20240530.133120";
         src = fetchurl {
-          url = "https://elpa.gnu.org/devel/org-9.7pre0.20240521.135840.tar";
-          sha256 = "06j7kjbpd390i0kqvvnlkpdzqkwiicbdgjymif338l0qx2kl4sg5";
+          url = "https://elpa.gnu.org/devel/org-9.7pre0.20240530.133120.tar";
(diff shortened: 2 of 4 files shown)


AndersonTorres (approved):

pkgs/applications/editors/emacs/elisp-packages/elpa-devel-packages.nix
```diff
@@ -64,6 +64,25 @@ self: let
...
+        patches = old.patches or [ ] ++ [
+          # security fix backported from 9.7.5
+          (pkgs.fetchpatch {
+            url = "https://git.savannah.gnu.org/cgit/emacs/org-mode.git/patch/?id=f4cc61636947b5c2f0afc67174dd369fe3277aa8";
+            hash = "sha256-bGgsnTSn6SMu1J8P2BfJjrKx2845FCsUB2okcIrEjDg=";
+            stripLen = 1;
+          })
+        ];
+        postPatch = old.postPatch or "" + "\n" + ''
+          cd ..
```
Atemu, line 78: Doesn't this break the build?

jian-lin: why do you think so?  You can try it out.

jian-lin:
@ofborg build emacs.pkgs.elpaPackages.org emacs.pkgs.elpaDevelPackages.org
emacs28.pkgs.elpaPackages.org
emacs28.pkgs.elpaDevelPackages.org

pkgs/applications/editors/emacs/elisp-packages/elpa-devel-packages.nix
Atemu: Don't we change the build dir for all other phases here? How is that supposed to work?

jian-lin: > How is that supposed to work?

It actually works, which is verified by the CI.

> Don't we change the build dir for all other phases here?

For `elpaBuild`, working directory has no effect.  It just need a right `src` environment variable pointing to a "good" tarball.

Atemu: It'd still feel cleaner to me to pusd/popd here, even if the build technically works without. Overriders could get confused for instance.

jian-lin: We can use `pushd ..` and `popd` here.  However, in that case, deleting the unpacked source directory is not allowed to avoid the failure of `popd`.  In addition, we have to disable the `buildPhase`.

I just think the `pushd` and `popd` is not worth the effect.  Just `cd ..` is good enough.

jian-lin: > Overriders could get confused for instance.

Please note that by default `dontUnpack` is true, which means there is no unpacked directory.  So `cd ..` does not change anything actually.

jian-lin: Pushed the `pushd`/`popd` version, which has its own side effects.

Atemu (approved):
Diff LGTM.

@ofborg build emacs.pkgs.elpaPackages.org emacs.pkgs.elpaDevelPackages.org
emacs28.pkgs.elpaPackages.org
emacs28.pkgs.elpaDevelPackages.org

jian-lin:
I think ofborg remembers what needs to `build`.

pkgs/applications/editors/emacs/elisp-packages/elpa-devel-packages.nix
Atemu: I see, I was missing this context. Feel free to revert to the previous version.

jian-lin:
Since it is a security fix, let's not wait for the slow darwin CI.

pkgs/applications/editors/emacs/elisp-packages/elpa-devel-packages.nix
hraban: fwiw for future reference: you can use a subshell to sidestep all of this:

```bash
(
  cd some/dir
)
```

This way you don't have to juggle directory stacks and you don't have to worry about anything which potentially exits early. (technically it also survives the subshell deleting the working directory of the parent shell tho at that point I'm not sure it's still a good idea 😂)

PR #198526, opened 2022-10-29
[nixos-container] Make nixos containers use unified cgroupsv2

farnoy:
###### Description of changes

Resolves #196413

systemd-nspawn uses FHS paths to detect the installed version of systemd to check if it's recent enough to support unified cgroups. This fails on NixOS as we don't have these `.so` files in `/lib`, `/usr/lib`, and so on.

Fortunately, we found an env var that can be used to help make the right decision without installing these files.

A regression test is included that checks for the output of `systemctl status`, to confirm the absence of the `cgroupsv1` taint flag.

###### Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
- [ ] For non-Linux: Is `sandbox = true` set in `nix.conf`? (See [Nix manual](https://nixos.org/manual/nix/stable/command-ref/conf-file.html))
- [ ] Tested, as applicable:
  - [NixOS test(s)](https://nixos.org/manual/nixos/unstable/index.html#sec-nixos-tests) (look inside [nixos/tests](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests))
  - and/or [package tests](https://nixos.org/manual/nixpkgs/unstable/#sec-package-tests)
  - or, for functions and "core" functionality, tests in [lib/tests](https://github.com/NixOS/nixpkgs/blob/master/lib/tests) or [pkgs/test](https://github.com/NixOS/nixpkgs/blob/master/pkgs/test)
  - made sure NixOS tests are [linked](https://nixos.org/manual/nixpkgs/unstable/#ssec-nixos-tests-linking) to the relevant packages
- [ ] Tested compilation of all packages that depend on this change using `nix-shell -p nixpkgs-review --run "nixpkgs-review rev HEAD"`. Note: all changes have to be committed, also see [nixpkgs-review usage](https://github.com/Mic92/nixpkgs-review#usage)
- [ ] Tested basic functionality of all binary files (usually in `./result/bin/`)
- [22.11 Release Notes (or backporting 22.05 Release notes)](https://github.com/NixOS/nixpkgs/blob/master/CONTRIBUTING.md#generating-2211-release-notes)
  - [ ] (Package updates) Added a release notes entry if the change is major or breaking
  - [ ] (Module updates) Added a release notes entry if the change is significant
  - [ ] (Module addition) Added a release notes entry if adding a new NixOS module
[...]

PR #321935, opened 2024-06-23
nixos/nspawn: cgroupsv2 by default

Ma27:
## Description of changes
That way you also get features like cgroup accounting for systemd units when using systemd-nspawn w/o classic nixos containers.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/nixos/modules/system/boot/systemd/nspawn.nix
+++ b/nixos/modules/system/boot/systemd/nspawn.nix
@@ -127,6 +127,9 @@ in {
         })
         {
           systemd.targets.multi-user.wants = [ "machines.target" ];
+          systemd.services."systemd-nspawn@".environment = {
+            SYSTEMD_NSPAWN_UNIFIED_HIERARCHY = mkDefault "1";
+          };
         }
       ];
 }


arianvp:
I'm surprised we have to explicitly enable this?

Ma27:
@arianvp

> I'm surprised we have to explicitly enable this?

I basically do the NixOS container setup here with Boot=false and Parameters set to the stage2 `init` script of a built nspawn container[1]. With that, `systemctl status` complained about cgroupsv1 being used (and cgroup accounting didn't work as a consequence for instance).

According to the sources[2] it's attempted to determine whether the installation supports that (i.e. the systemd version in it is sufficiently recent). This may work with other distros with a layout systemd expects, but I strongly suspect that the detection just fails with a NixOS in it and thus nspawn falls back to cgroupsv1.

[1] The main difference is that each container has its own store here, but that's not related since nixos-containers solved the issue the same way.
[2] https://github.com/systemd/systemd/blob/59a7019fee57b36a3deb470af09c0573de6f4a47/src/nspawn/nspawn.c#L531-L560

ElvishJerricco:
> but I strongly suspect that the detection just fails with a NixOS in it

Could we fix *that* instead?

Ma27:
@ElvishJerricco here's the relevant code-part: https://github.com/systemd/systemd/blob/main/src/nspawn/nspawn-util.c#L19-L30

In my case the fix would be to add `/nix/var/nix/profiles/system/systemd/lib/...` to that list. For my use-case (nspawn with custom store inside) this will work fine. For the normal nixos-container use-case (shared store) it probably won't (but for that we already have the env var set).

Would you accept a patch doing that (bringing this up because my use-case is rather special I think) to pkgs.systemd?

ElvishJerricco:
Hm, `/nix/var/nix/profiles/system/systemd` is not a path we would normally use. Normally we'd use `/run/current-system`, as this *definitely* refers to the intended system derivation, which could be any generation from any profile (or no profile at all), not just the latest from the `system` profile. So I'm not sure if that's correct or not for this particular use case. @flokli do you have any thoughts?

arianvp:
Can we run the activation script outside of the container before we invoke nspawn? Just like we do with systemd stage1/stage2?

flokli:
https://github.com/NixOS/nixpkgs/pull/198526 did set this in [nixos/modules/virtualisation/nixos-containers.nix](https://github.com/NixOS/nixpkgs/pull/198526/files#diff-9f35243fa1e67cb77387bca245d3bc5011b08ed64326772fb9190a35df438cab)`, so doing the same here is probably fine. I'd still like this to be a bit smarter.

Considering we also do bind-mount `/nix/var/nix/profiles/system` into the container in both places, extending systemd to look there is probably fine (?)

ElvishJerricco:
@arianvp 

> Can we run the activation script outside of the container before we invoke nspawn? Just like we do with systemd stage1/stage2?

Well, the issue there is that the activation script creates the `/run/current-system` symlink, but I'm guessing `/run` is going to be mounted over with tmpfs when the container starts, clearing anything activation had put there.

Ma27:
> Considering we also do bind-mount /nix/var/nix/profiles/system into the container in both places, extending systemd to look there is probably fine (?)

@flokli fwiw while this will work in my case (I don't do bind mounts, each container has its own store), I don't think this would work for the bind-mount case since systemd will look from the outside what systemd is in the container directory at a very early stage (see link I posted above), i.e. without bind-mounts in place.

flokli:
Following https://github.com/NixOS/nixpkgs/pull/321935#issuecomment-2212503606, this is applying the same mechanism already present in NixOS containers. Any further patch to get the detection done more reliably applies to both places. No need to hold this PR back.

PR #321723, opened 2024-06-22
libimobiledevice: pull `gcc-14` fix pending upstream inclusion

trofi:
## Description of changes

Without the change both `gcc-14` and `clag` `stdenv`s fil the build as:

        $ nix build --impure --expr 'with import ./. {}; libimobiledevice.override {stdenv = gcc14Stdenv; }' -L
        ...
        libimobiledevice> afcclient.c: In function 'print_file_info':
        libimobiledevice> afcclient.c:488:17: error: implicit declaration of function 'strftime' [8;;https://gcc.gnu.org/onlinedocs/gcc-14.1.0/gcc/Warning-Options.html#index-Wimplicit-function-declaration-Wimplicit-function-declaration8;;]
        libimobiledevice>   488 |                 strftime(timebuf, 64, "%d %h %Y %H:%M:%S", localtime(&t));
        libimobiledevice>       |                 ^~~~~~~~
        libimobiledevice> afcclient.c:75:1: note: include '<time.h>' or provide a declaration of 'strftime'
        libimobiledevice>    74 | #include <libimobiledevice-glue/utils.h>
        libimobiledevice>   +++ |+#include <time.h>
        libimobiledevice>    75 |

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
- For non-Linux: Is sandboxing enabled in `nix.conf`? (See [Nix manual](https://nixos.org/manual/nix/stable/command-ref/conf-file.html))
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:
  - [NixOS test(s)](https://nixos.org/manual/nixos/unstable/index.html#sec-nixos-tests) (look inside [nixos/tests](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests))
  - and/or [package tests](https://github.com/NixOS/nixpkgs/blob/master/pkgs/README.md#package-tests)
  - or, for functions and "core" functionality, tests in [lib/tests](https://github.com/NixOS/nixpkgs/blob/master/lib/tests) or [pkgs/test](https://github.com/NixOS/nixpkgs/blob/master/pkgs/test)
  - made sure NixOS tests are [linked](https://nixos.org/manual/nixpkgs/unstable/#ssec-nixos-tests-linking) to the relevant packages
[...]

PR #321957, opened 2024-06-23
mercurial: fix cargoDeps hash

superherointj:
mercurial: fix cargoDeps hash

image

Found this problem at [libimobiledevice build](https://github.com/NixOS/nixpkgs/pull/321723#issuecomment-2184693488).

I don't know yet why cargoDeps hash changed. But this fixes builds for other packages.

Current best guess:

* https://github.com/NixOS/nixpkgs/pull/321960#issuecomment-2185099697


mweinelt:
Fixed via https://github.com/NixOS/nixpkgs/pull/321998

Issue #189851, opened 2022-09-05
Flatpak apps can't launch the default browser


cizra:
### Describe the bug
Flatpak apps (like Thunderbird) can't launch the default browser

### Steps To Reproduce
1. `flatpak --user install org.mozilla.Thunderbird`
3. Verify you have a default browser: `xdg-settings get default-web-browser` - I'm getting `firefox.desktop` as output
4. Open Thunderbird. Click on any link in a mail or somewhere in UI (there's the "Donate" link on one of the tabs shown by Thunderbird on a clean install)

### Expected behavior
Default browser opens.

### Screenshots
![image](https://user-images.githubusercontent.com/2131991/188449978-dcbae4ac-47d5-4a67-9fa5-6127d0a7571c.png)

### Additional context
```
  programs.sway.enable = true;
  services.flatpak.enable = true;
  xdg.portal = {
    wlr.enable = true;
    extraPortals = [ pkgs.xdg-desktop-portal-gtk ];
    enable = true;
  };
```
To make sure my user settings aren't messed up, I created a new `test` user in my system. It too exhibited the buggy behavior.

### Notify maintainers
@jtojnar 

### Metadata
```console
[user@system:~]$ nix-shell -p nix-info --run "nix-info -m"
 - system: `"x86_64-linux"`
 - host os: `Linux 5.10.126, NixOS, 21.11 (Porcupine)`
 - multi-user?: `yes`
 - sandbox: `yes`
 - version: `nix-env (Nix) 2.3.16`
 - channels(root): `"nixos-21.11.337975.eabc3821918, unstable-22.11pre405560.2da64a81275"`
 - nixpkgs: `/nix/var/nix/profiles/per-user/root/channels/nixos`
```

cizra:
Reading the source you provided, I see that OpenURI portal calls xdp_dbus_impl_app_chooser_call_choose_application. Where's that implemented? It seems this is the failing point.

Here's an exceprt from dbus-monitor:
```
method call time=1662371825.457775 sender=:1.4 -> destination=:1.7 serial=151 path=/org/freedesktop/portal/desktop; interface=org.freedesktop.impl.portal.AppChooser; member=ChooseApplication
   object path "/org/freedesktop/portal/desktop/request/1_31/t"
   string "org.mozilla.Thunderbird"
   string ""
   array [
   ]
   array [
      dict entry(
         string "content_type"
         variant             string "x-scheme-handler/https"
      )
      dict entry(
         string "uri"
[...]

PR #93764, opened 2020-07-24
nixos/displayManager: add XDG_SESSION_ID to systemd user environment

evenbrenden:
This is a revamp of @misuzu's https://github.com/NixOS/nixpkgs/pull/67751, with SLiM changes omitted and xss-lock changes added.

###### Motivation for this change

xss-lock needs `XDG_SESSION_ID` to respond to `loginctl lock-session` (and possibly other session operations such as idle hint management). Tested with LightDM.

###### Things done

Add `XDG_SESSION_ID` to the list of imported environment variables when starting systemctl.

Inspired by home-manager, add `importVariables` configuration.

Set session to `XDG_SESSION_ID` when running xss-lock as a service.

- [ ] Tested using sandboxing ([nix.useSandbox](https://nixos.org/nixos/manual/options.html#opt-nix.useSandbox) on NixOS, or option `sandbox` in [`nix.conf`](https://nixos.org/nix/manual/#sec-conf-file) on non-NixOS linux)
- Built on platform(s)
   - [X] NixOS
   - [ ] macOS
   - [ ] other Linux distributions
- [ ] Tested via one or more NixOS test(s) if existing and applicable for the change (look inside [nixos/tests](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests))
- [ ] Tested compilation of all pkgs that depend on this change using `nix-shell -p nixpkgs-review --run "nixpkgs-review wip"`
- [ ] Tested execution of all binary files (usually in `./result/bin/`)
- [ ] Determined the impact on package closure size (by running `nix path-info -S` before and after)
- [ ] Ensured that relevant documentation is up to date
- [X] Fits [CONTRIBUTING.md](https://github.com/NixOS/nixpkgs/blob/master/.github/CONTRIBUTING.md).

PR #322000, opened 2024-06-23
nixos/xserver: expose services.xserver.displayManager.importedVariables

JohnRTitor:
## Description of changes

This option allows user to import specific environment variables into systemd user environment, this is useful situations where opening links into another app does not work. This is frequently the case with flatpak apps.

~After this PR, we can utilise this option for example in [sway module](https://github.com/NixOS/nixpkgs/blob/a71e967ef3694799d0c418c98332f7ff4cc5f6af/nixos/modules/programs/wayland/sway.nix#L125) and [hyprland module](https://github.com/NixOS/nixpkgs/blob/a71e967ef3694799d0c418c98332f7ff4cc5f6af/nixos/modules/programs/wayland/hyprland.nix#L66), which employ other mechanisms for importing crucial variables like PATH, into the systemd environment.~

This option only affects XSessions and NOT Wayland.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
- [ ] Tested, as applicable:

---

### DIFF
--- a/nixos/modules/services/x11/display-managers/default.nix
+++ b/nixos/modules/services/x11/display-managers/default.nix
@@ -210,10 +210,14 @@ in
 
       importedVariables = mkOption {
         type = types.listOf (types.strMatching "[a-zA-Z_][a-zA-Z0-9_]*");
-        visible = false;
         description = ''
-          Environment variables to import into the systemd user environment.
+          Environment variables to import into the systemd user environment
+          when starting a graphical session.
         '';
+        example = [
+          "PATH"
+          "XDG_DATA_DIRS"
+        ];
       };
 
     };


fufexan (commented):
From what I can see, `importedVariables` already does all the work, it's just not visible. Maybe changing its visibility and setting the default to the value already set in `config` is enough.

JohnRTitor:
Indeed, though I did not want to touch the defaults. Sure.

fufexan (commented):
We should also move https://github.com/NixOS/nixpkgs/blob/2a2796c2bc87c5689383217d0a5b1cc293342aae/nixos/modules/services/x11/display-managers/default.nix#L228-L236 to the `default` value in the option definition.

JohnRTitor:
Won't moving them to default make them easily overridable, which can cause trouble, for example if someone just specifies:

```nix
services.xserver.displayManager.importedVariables = [ "PATH" ];
```

Would the default values be applied?

JohnRTitor:
And I tried testing with this, it does not work. Are the values even being applied in a Wayland session?

```
  programs.hyprland.systemd.setPath = lib.mkForce false;
  services.xserver.updateDbusEnvironment = true;
  services.xserver.displayManager.importedVariables = [
    "PATH"
    "XDG_DATA_DIRS"
  ];
```

Tried on the Hyprland version on nixos-unstable.

jtojnar:
Yes, this is only invoked by Xsession wrapper, which is not used by Wayland session.

JohnRTitor:
Can we implement something like this for Wayland, by running the same commands when starting a Wayland session?

ElvishJerricco:
> I am not sure if the default values will be applied then.

You're right, they won't. If you put things in the option's `default`, then any definition will override and wipe out those things.

JohnRTitor:
I am tempted to merge this as it is, even though it does not impact Wayland session like I originally intended it would.

All I did is make it visible and added an example here.

fufexan:
Perhaps we could make `importedVariables` readonly, and add an `extraVariables` user-settable option.

ElvishJerricco:
@fufexan Why is that better than just doing it this way? That's just a more complicated way of doing the same thing that also can't be overridden if someone wants to.

JohnRTitor:
That was my original intention with this PR, yes, but then I realised variables defined here and in other modules will be merged into a list by the module system. So is there even a need to? 
https://github.com/NixOS/nixpkgs/blob/560f0817fc6018a895d73d28a401d3a5cd132d38/nixos/modules/services/x11/display-managers/default.nix#L232-L240

fufexan:
I assumed the option was not visible for a reason (users should probably not change it). I have no issue with merging it as is.

jtojnar:
I am not convinced the [PR that introduced this option](https://github.com/NixOS/nixpkgs/pull/93764) was desirable in the first place. If a service needs certain environment variable, it should be passed to it explicitly.

Most services probably do not need `XDG_SESSION_ID`. I just checked Fedora (created user service that runs `env` and set `XDG_SESSION_ID` in `/etc/profile.d`) and it does not pass that to services either.

Also, we already have extremely bloated user service environment compared to Fedora – ¿perhaps from `pam_env` being used for `environment.sessionVariables`?

JohnRTitor:
> I am not convinced the [PR that introduced this option](https://github.com/NixOS/nixpkgs/pull/93764) was desirable in the first place. If a service needs certain environment variable, it should be passed to it explicitly.

I wanted to introduce this option as a workaround of https://github.com/NixOS/nixpkgs/issues/189851. 

Importing PATH to systemd, usually fixes it.

jtojnar:
I think `PATH` and `XDG_DATA_DIRS` might be one of those things that make sense to introduce unconditionally, to reproduce `/usr/bin` and `/usr/share` being available to user services on FHS distros out of the box.

But as we noticed above, this option will not help anyone – Wayland desktops will not run the Xsession wrapper, and the X-based desktops typically invoke [`dbus-update-activation-environment`](https://dbus.freedesktop.org/doc/dbus-update-activation-environment.1.html) themselves (which is what we probably would want instead of `systemd import-environment`).

A proper solution might be replacing the use of `pam_env` in `environment.sessionVariables` with systemd [environment generators](https://www.freedesktop.org/software/systemd/man/latest/systemd.environment-generator.html). That way, services would have the environment variables without needing to import anything.

JohnRTitor:
> A proper solution might be replacing the use of `pam_env` in `environment.sessionVariables` with systemd [environment generators](https://www.freedesktop.org/software/systemd/man/latest/systemd.environment-generator.html). That way, services would have the environment variables without needing to import anything.

Thanks for this, I'll need to get an idea on how to do this though, would love pointers from the systemd team!

JohnRTitor:
Closing. I don't see any point in keeping this open, as I use Wayland and this only affects X.

PR #322004, opened 2024-06-23
lib.network: first/last and next addresses in a network

woojiq:
### Description of changes

Depends on #318712 and #326765

* Add functions to find the first, last and next addresses in a network using
bitwise operations.
* Write unit tests for such functions.

Mentor: @Janik-Haag 

### Things done
- [x] Tested, as applicable:
* [NixOS test(s)](https://nixos.org/manual/nixos/unstable/index.html#sec-nixos-tests) (look inside [nixos/tests](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests))
* and/or [package tests](https://nixos.org/manual/nixpkgs/unstable/#sec-package-tests)
* or, for functions and "core" functionality, tests in [lib/tests](https://github.com/NixOS/nixpkgs/blob/master/lib/tests) or [pkgs/test](https://github.com/NixOS/nixpkgs/blob/master/pkgs/test)
- [x] Formatted using nixfmt-rfc-style

Add a 👍 [reaction](https://github.blog/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to [pull requests you find important](https://github.com/NixOS/nixpkgs/pulls?q=is%3Aopen+sort%3Areactions-%2B1-desc).

### DIFF
--- a/lib/network/default.nix
+++ b/lib/network/default.nix
@@ -154,5 +154,36 @@ in
         address = _ipv6.calculateLastAddress addrV6._address addrV6.prefixLength;
       in
       _ipv6.makeIpv6Type address addrV6.prefixLength;
+
+    /**
+      Calculates the next address in an IPv6 subnet. If this is the last address, null is returned.
+
+      # Type
+
+      ```
+      nextAddress :: ipv6AddrAttrs -> (ipv6AddrAttrs | null)
+      ```
+
+      # Examples
+
+      ```nix
+      nextAddress (fromString "2001:DB8::ffff/64")
+      => {
+        address = "2001:db8:0:0:0:0:1:0";
+        ...
+      }
+      ```
+
+      # Arguments
+
+      addrV6
+      : IPv6 type produced by a parser like `ipv6.fromString`.
+    */
+    nextAddress =
+      addrV6:
+      let
+        address = _ipv6.calculateNextAddress addrV6._address addrV6.prefixLength;
+      in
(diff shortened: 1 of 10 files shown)


lib/network/default.nix
```diff
@@ -0,0 +1,65 @@
...
+        lastAddress = "2001:db8:ffff:ffff:ffff:ffff:ffff:ffff";
+        nextAddress = "2001:db8:0:0:0:0:1:0";
+      }
+      ```
+
+      # Arguments
+
+      - [addr] An IPv6 address with optional prefix length.
+    */
+    fromString =
```
Janik-Haag, line 38: Instead of modifying `fromString`, we should have something like `lib.ipv6.firstAddress` and `lib.ipv6.lastAddress`.

woojiq: Hmm, can you explain roughly how you want to use the library? From my point of view, I expect something like `fromString <addr>` and then the user takes whatever data they need from the returned attr set.

Janik-Haag: I guess that's fair, esp. since if we implement it correctly the values are lazily evaluated, which means it doesn't really increase the cost. But I still think we should expose the function to end users because it's more ergonomic to call `lib.ipv6.lastAddress "f00::c0ffee/64"` vs `(lib.ipv6.fromString "f00::c0ffee/64").lastAddress`. There will also be functions in the future that aren't really incorporatable into `fromString`, for example imagine something like `lib.ipv6.arithmetic "f00::c0ffee/64" (-50)`

PR #322029, opened 2024-06-23
Revert "roc-toolkit: 0.3.0 -> 0.4.0"

superherointj:
Revert "roc-toolkit: 0.3.0 -> 0.4.0"

Reverts NixOS/nixpkgs#320870

Broke pipewire.

PR #320870, opened 2024-06-18
roc-toolkit: 0.3.0 -> 0.4.0

noblepayne:
## Description of changes
Update `roc-toolkit` from 0.3.0 to 0.4.0:
+ [GitHub Release](https://github.com/roc-streaming/roc-toolkit/releases/tag/v0.4.0)
+ [Changelog](https://roc-streaming.org/toolkit/docs/development/changelog.html#version-0-4-0-jun-14-2024)
+ [Git Diff](https://github.com/roc-streaming/roc-toolkit/compare/v0.3.0..v0.4.0)

Changes:
+ Bump version and update hash.
+ Added libsndfile.

Testing:
+ Built and ran main binaries.
+ Built and ran pipewire with roc module.

## Things done
- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
- For non-Linux: Is sandboxing enabled in `nix.conf`? (See [Nix manual](https://nixos.org/manual/nix/stable/command-ref/conf-file.html))
  - [ ] `sandbox = relaxed`
  - [x] `sandbox = true`
- [ ] Tested, as applicable:
  - [NixOS test(s)](https://nixos.org/manual/nixos/unstable/index.html#sec-nixos-tests) (look inside [nixos/tests](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests))
  - and/or [package tests](https://github.com/NixOS/nixpkgs/blob/master/pkgs/README.md#package-tests)
  - or, for functions and "core" functionality, tests in [lib/tests](https://github.com/NixOS/nixpkgs/blob/master/lib/tests) or [pkgs/test](https://github.com/NixOS/nixpkgs/blob/master/pkgs/test)
  - made sure NixOS tests are [linked](https://nixos.org/manual/nixpkgs/unstable/#ssec-nixos-tests-linking) to the relevant packages
- [ ] Tested compilation of all packages that depend on this change using `nix-shell -p nixpkgs-review --run "nixpkgs-review rev HEAD"`. Note: all changes have to be committed, also see [nixpkgs-review usage](https://github.com/Mic92/nixpkgs-review#usage)
- [x] Tested basic functionality of all binary files (usually in `./result/bin/`)
[...]

PR #322021, opened 2024-06-23
lz4: patch cmake to generate unified target

tobim:
Fixes #322003.

## Description of changes

Integrated changes from:
* https://github.com/lz4/lz4/pull/1372
* https://github.com/lz4/lz4/pull/1413

Dropped a patch for the `make` based build that is not used any more.

Added myself as a maintainer of the package.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
- For non-Linux: Is sandboxing enabled in `nix.conf`? (See [Nix manual](https://nixos.org/manual/nix/stable/command-ref/conf-file.html))
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:
  - [NixOS test(s)](https://nixos.org/manual/nixos/unstable/index.html#sec-nixos-tests) (look inside [nixos/tests](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests))
  - and/or [package tests](https://github.com/NixOS/nixpkgs/blob/master/pkgs/README.md#package-tests)
  - or, for functions and "core" functionality, tests in [lib/tests](https://github.com/NixOS/nixpkgs/blob/master/lib/tests) or [pkgs/test](https://github.com/NixOS/nixpkgs/blob/master/pkgs/test)
  - made sure NixOS tests are [linked](https://nixos.org/manual/nixpkgs/unstable/#ssec-nixos-tests-linking) to the relevant packages
- [ ] Tested compilation of all packages that depend on this change using `nix-shell -p nixpkgs-review --run "nixpkgs-review rev HEAD"`. Note: all changes have to be committed, also see [nixpkgs-review usage](https://github.com/Mic92/nixpkgs-review#usage)
- [ ] Tested basic functionality of all binary files (usually in `./result/bin/`)
- [24.11 Release Notes](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2411.section.md) (or backporting [23.11](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2311.section.md) and [24.05](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2405.section.md) Release notes)
  - [ ] (Package updates) Added a release notes entry if the change is major or breaking
  - [ ] (Module updates) Added a release notes entry if the change is significant
[...]

PR #322151, opened 2024-06-24
pipewire: fix module-roc-sink explicity specifying sender packet encoding

superherointj:
pipewire: fix module-roc-sink explicity specifying sender packet encoding

CyberShadow [reports success](https://github.com/NixOS/nixpkgs/pull/320870#issuecomment-2185964928) with this patch. I haven't tested.
Per K900 [comment](https://github.com/NixOS/nixpkgs/pull/322029#issuecomment-2185317650), failing passthru test is assumed to be unrelated.

Fixes regression:
https://github.com/NixOS/nixpkgs/pull/320870#issuecomment-2185279926

Thanks CyberShadow for the fix.

Draft: Needs to target `next-staging`. Because it's a [fix to hydra builds](https://github.com/NixOS/nixpkgs/blob/master/CONTRIBUTING.md#staging). I suppose.

### DIFF
--- a/pkgs/development/libraries/pipewire/default.nix
+++ b/pkgs/development/libraries/pipewire/default.nix
@@ -1,6 +1,7 @@
 { stdenv
 , lib
 , fetchFromGitLab
+, fetchpatch
 , python3
 , meson
 , ninja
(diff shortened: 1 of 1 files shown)


CyberShadow (approved):

superherointj:
@mweinelt @K900  I'm trying to understand if this should target `staging-next`. It's a fix for hydra builds. You just re-targeted a PR from staging to staging-next. So I wonder if this is the same case. It seems so. Because this is fixing pipewire at master.

superherointj:
Thanks @K900. And I'm sorry for the trouble.

mweinelt:
> You have https://github.com/NixOS/nixpkgs/pull/322021#issuecomment-2186372509. So I wonder if this is the same case

If it fixes a known regression on staging-next, then it can go into the current cycle.

PR #295155, opened 2024-03-11
hareHook: init

onemoresuza:
## Description of changes

The `hareHook` package does the following:

1. Sets `HARECACHE` to `$(mktemp -d)`;
2. Allows for cross-compilation out of the box, by wrapping the hare binary in a script that passes the `-a` flag for it;
3. Provides the `hareBuildType` attribute, which accepts either `release` (default) or `debug` for passing the `-R` for the wrapper script;
4. Removes the pollution of logs by passing the `-q` flag to the wrapper script;
5. Provides the `hare-native` script for when there's a need to run compiled binaries on the build process; and
6. Appends Hare's stdlib to `HAREPATH` only after all third party libraries are added, so that it may be extended or shadowed by them.

### TODO

- [x] Fix cross-compilation (Fixed, thanks to @uninsane in https://github.com/onemoresuza/nixpkgs/pull/1#issue-2182138217.)
	The hook does not seem to pass the `buildPlatform` `hare` binary when
	cross-compiling, but the one from `hostPlatform`.
- [x] Add an entry to the nixpkgs manual.

## Things done

- Built on platform(s)
  - [X] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
- For non-Linux: Is sandboxing enabled in `nix.conf`? (See [Nix manual](https://nixos.org/manual/nix/stable/command-ref/conf-file.html))
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [x] Tested, as applicable:
  - [NixOS test(s)](https://nixos.org/manual/nixos/unstable/index.html#sec-nixos-tests) (look inside [nixos/tests](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests))
  - and/or [package tests](https://nixos.org/manual/nixpkgs/unstable/#sec-package-tests)
  - or, for functions and "core" functionality, tests in [lib/tests](https://github.com/NixOS/nixpkgs/blob/master/lib/tests) or [pkgs/test](https://github.com/NixOS/nixpkgs/blob/master/pkgs/test)
  - made sure NixOS tests are [linked](https://nixos.org/manual/nixpkgs/unstable/#ssec-nixos-tests-linking) to the relevant packages
[...]

PR #271597, opened 2023-12-02
python3.pkgs.buildPythonPackage: Separate runtime & build time dependencies

adisbladis:
## Description of changes

See #272178 for more context.

### Dependency separation

Adds a separation between Python runtime dependencies, Python build-time dependencies & non-Python dependencies arguments to `buildPythonPackage`.
It's important to have these distinctions so we can solve our long standing issues re dependency propagation in Python.
With this change in place it's possible to implement a solution for https://github.com/NixOS/nixpkgs/issues/170577.

It's similar to https://github.com/NixOS/nixpkgs/pull/102613 (a subset, really) but with the notable changes that:
- Uses PEP-621 standard names/idioms
  #102613 introduces the argument `requiredPythonModules`, but with Python metadata being standardised I'd like to be more closely aligned with that.
- Only implements the API that enables changing the behaviour
  It's probably going to take a while to refactor all of nixpkgs to use the appropriate arguments.
  This change is purposely minimal in the hopes that we can move forward with it more quickly.

For now `dependencies` are always mapped onto `propgatedBuildInputs`, so there is no change in behaviour in the immediate term.
2 packages are migrated as a PoC, one is using `format = "pyproject"` and the other is `setuptools`.

### Interface changes

This also unifies regular dependencies with optional dependencies:
``` nix
{
  passthru.optional-dependencies.socks = [ pysocks ];
  nativeBuildInputs = [ poetry-core ];
  propagatedBuildInputs = [ requests ];
}
```
becomes
``` nix
{
  optional-dependencies.socks = [ pysocks ];
  build-system = [ poetry-core ];
  dependencies = [ requests ];
}
```
which looks much more like what a standard `pyproject.toml` file contains.
I think the interface change alone is a pretty good improvement.

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
- For non-Linux: Is sandboxing enabled in `nix.conf`? (See [Nix manual](https://nixos.org/manual/nix/stable/command-ref/conf-file.html))
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:
  - [NixOS test(s)](https://nixos.org/manual/nixos/unstable/index.html#sec-nixos-tests) (look inside [nixos/tests](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests))
[...]

PR #322192, opened 2024-06-24
python3Packages.listparser: init at 0.20

uninsane:
this package was previously available via `gnome-feeds.passthru.listparser` but was [removed](https://github.com/NixOS/nixpkgs/pull/320801/files) when no longer needed by gnome-feeds. make it available as a standalone package, because the OPML lists it works with are usable by many (other) feed readers.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/development/python-modules/listparser/default.nix
+++ b/pkgs/development/python-modules/listparser/default.nix
@@ -1,0 +1,55 @@
+{
+  fetchFromGitHub,
+  lib,
+  lxml,
+  nix-update-script,
+  poetry-core,
+  pypaBuildHook,
+  pypaInstallHook,
+  pytestCheckHook,
+  requests,
+  stdenv,
+}:
+
+stdenv.mkDerivation (finalAttrs: {
+  pname = "listparser";
+  version = "0.20";
+
+  src = fetchFromGitHub {
+    owner = "kurtmckee";
+    repo = "listparser";
+    rev = "v${finalAttrs.version}";
+    hash = "sha256-eg9TrjDgvHsYt/0JQ7MK/uGc3KK3uGr3jRxzG0FlySg=";
+  };
+
+  nativeBuildInputs = [
+    poetry-core
+    pypaBuildHook
+    pypaInstallHook
+  ];
+
+  propagatedBuildInputs = [
+    requests
+    lxml
+  ];
+
+  nativeCheckInputs = [
+    pytestCheckHook
+  ];
+
+  pythonImportsCheck = [
+    "listparser"
+  ];
+
+  doCheck = true;
+
+  passthru.updateScript = nix-update-script { };
+
+  meta = with lib; {
+    description = "Parse OPML subscription lists in Python";
+    homepage = "https://github.com/kurtmckee/listparser";
+    license = licenses.lgpl3Plus;
+    maintainers = [ maintainers.colinsane ];
+    platforms = platforms.linux;
+  };
+})
--- a/pkgs/top-level/python-packages.nix
+++ b/pkgs/top-level/python-packages.nix
@@ -6943,6 +6943,8 @@ self: super: with self; {
     python3 = python;
   });
 
+  listparser = toPythonModule (callPackage ../development/python-modules/listparser { });
+
   lit = callPackage ../development/python-modules/lit { };
 
   litellm = callPackage ../development/python-modules/litellm { };


pkgs/development/python-modules/listparser/default.nix
```diff
@@ -0,0 +1,48 @@
...
+  poetry-core,
+  pytestCheckHook,
+  requests,
+}:
+
+buildPythonPackage rec {
+  pname = "listparser";
+  version = "0.20";
+
+  format = "pyproject";
```
h7x4, line 16:
```suggestion
  pyproject = true;
```

pkgs/development/python-modules/listparser/default.nix
```diff
@@ -0,0 +1,48 @@
...
+
+  nativeBuildInputs = [
+    poetry-core
+  ];
+
+  nativeCheckInputs = [
+    pytestCheckHook
+  ];
+
+  propagatedBuildInputs = [
```
h7x4, line 33:
```suggestion
  dependencies = [
```

pkgs/development/python-modules/listparser/default.nix
```diff
@@ -0,0 +1,48 @@
...
+  format = "pyproject";
+
+  src = fetchFromGitHub {
+    owner = "kurtmckee";
+    repo = "listparser";
+    rev = "v${version}";
+    hash = "sha256-eg9TrjDgvHsYt/0JQ7MK/uGc3KK3uGr3jRxzG0FlySg=";
+  };
+
+  nativeBuildInputs = [
```
h7x4, line 25:
```suggestion
  build-system = [
```

pkgs/development/python-modules/listparser/default.nix
uninsane:
```console
$ rg 'propagatedBuildInputs =' pkgs/development/python-modules -l  | wc -l
4374
$ rg 'dependencies =' pkgs/development/python-modules -l  | wc -l
1467
```

is `dependencies` a new thing? otherwise, AFAICT `propagatedBuildInputs` would seem to be the preferred method.

h7x4: Yes, all of my suggestions are only changes towards the relatively new way of doing these. Shouldn't have any functional impact

uninsane: switched to `stdenv.mkDerivation`, where only the standard attributes are available.

h7x4: What was the reason behind the switch?

uninsane: i'm asking the opposite question ([in this thread](https://github.com/NixOS/nixpkgs/pull/271597)). packaging is only scalable because the idioms you learn from your first package apply to your second package, and soon enough it only takes 10 minutes to write a package even if your first package took a day.

that only works because we share idioms across much of nixpkgs. language-specific derivation wrappers already work against that and newer languages to nix ([zig](https://github.com/NixOS/nixpkgs/pull/241741#issuecomment-1627467997), [hare](https://github.com/NixOS/nixpkgs/pull/295155)) are trending toward stock `stdenv.mkDerivation` with a minimal amount of language-specific logic delegated to the derivation author in light of that (as little as asking them to add `nativeBuildInputs = [ zig.hook ]` to their package).

the changes i'm learning of here are contrary to all of that. it's not worth it for the 1000s of nixpkgs-maintainers-who-might-someday-want-to-package-a-python-library to learn _even more_ language-specific things whose specificity don't (yet?) seem to have much benefit to the packager.

pkgs/top-level/python-packages.nix
```diff
@@ -6943,6 +6943,8 @@ self: super: with self; {
     python3 = python;
   });
 
+  listparser = toPythonModule (callPackage ../development/python-modules/listparser { });
```
SuperSandro2000, line 6946:
```suggestion
  listparser = callPackage ../development/python-modules/listparser { };
```

pkgs/development/python-modules/listparser/default.nix
```diff
@@ -0,0 +1,55 @@
...
+    owner = "kurtmckee";
+    repo = "listparser";
+    rev = "v${finalAttrs.version}";
+    hash = "sha256-eg9TrjDgvHsYt/0JQ7MK/uGc3KK3uGr3jRxzG0FlySg=";
+  };
+
+  nativeBuildInputs = [
+    poetry-core
+    pypaBuildHook
+    pypaInstallHook
```
SuperSandro2000, line 28:
```suggestion
```

pkgs/development/python-modules/listparser/default.nix
```diff
@@ -0,0 +1,55 @@
...
+  nix-update-script,
+  poetry-core,
+  pypaBuildHook,
+  pypaInstallHook,
+  pytestCheckHook,
+  requests,
+  stdenv,
+}:
+
+stdenv.mkDerivation (finalAttrs: {
```
SuperSandro2000, line 14:
```suggestion
buildPythonPackage {
```

SuperSandro2000 (changes requested):

natsukium:
> @SuperSandro2000 You might want to read earlier comment threads here for context

There is no need. See https://github.com/NixOS/nixpkgs/pull/271597#issuecomment-2187849983

pkgs/development/python-modules/listparser/default.nix
SuperSandro2000: > switched to `stdenv.mkDerivation`, where only the standard attributes are available.

That is wrong. python packages under python-modules have to use buildPythonPackage unless they are build outputs of other packages or have special exceptions which is not the case here. That you don't like the naming doesn't free you of that.

SuperSandro2000:
> @SuperSandro2000 You might want to read earlier comment threads here for context

After finding the resolved review thread and thinking a bit about it, I could finally understand why the change was done. The justification for it is not sufficient and just hacks around our future plans, making things more complicated than they need to be. The package here should and must use buildPythonPackage and I personally do not have a preference if the new or old names are used but it is probably better to use the newer ones.

That there is a personal disliking for the naming doesn't allow that we make a special exception for a package where there is no reason to do so. I also had other naming suggestions in the PR where it was done but matching upstreams python code is a good justification.

uninsane:
@natsukium thanks for taking the time to weigh in: that did clear up the direction of python packaging in nixpkgs for me.

i understand the limitations with `propagatedBuildInputs`. i don't quite understand the need for `build-system` yet, after reading the discussions i could reach from https://github.com/NixOS/nixpkgs/pull/271597#issue-2021868316. but that's OK: it's clear enough that the direction of these things is settled. OTOH the idioms i use when packaging for myself v.s. packaging python libraries for nixpkgs are different enough now (and will grow more different in the future) that trying to do both doesn't make a whole lot of sense.

unless something pops up that i really, really want to share, i'm going to keep it simple and just package these smaller libraries locally (where they're still public, at least, just without the accessibility/reach of nixpkgs).

PR #322215, opened 2024-06-24
vscode-extensions.jeff-hykin.better-nix-syntax: init 1.1.5

superherointj:
vscode-extensions.jeff-hykin.better-nix-syntax: init 1.1.5

Possible review? @ratsclub @znaniye


znaniye (approved):
LGTM!

PR #322394, opened 2024-06-25
elpa: mark broken

superherointj:
elpa: mark broken

* https://hydra.nixos.org/build/263906391/nixlog/1

CC Maintainer @markuskowa


markuskowa (approved):
Fair enough. I have been trying to repair it since January and have failed on multiple attempts. Git bisect only lands a merge commit, making it even harder to find the cause.

PR #260193, opened 2023-10-10
minimal-bootstrap: component packages for `bootstrap-tools`

emilytrau:
## Description of changes

Packages many components of `bootstrap-tools` as statically linked binaries. The only missing tool is a `gcc` that targets `glibc`, I'm still working on understanding how compiler/libc split works 😅

Related #227914

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
- [ ] For non-Linux: Is `sandbox = true` set in `nix.conf`? (See [Nix manual](https://nixos.org/manual/nix/stable/command-ref/conf-file.html))
- [x] Tested, as applicable:
  - [NixOS test(s)](https://nixos.org/manual/nixos/unstable/index.html#sec-nixos-tests) (look inside [nixos/tests](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests))
  - and/or [package tests](https://nixos.org/manual/nixpkgs/unstable/#sec-package-tests)
  - or, for functions and "core" functionality, tests in [lib/tests](https://github.com/NixOS/nixpkgs/blob/master/lib/tests) or [pkgs/test](https://github.com/NixOS/nixpkgs/blob/master/pkgs/test)
  - made sure NixOS tests are [linked](https://nixos.org/manual/nixpkgs/unstable/#ssec-nixos-tests-linking) to the relevant packages
- [ ] Tested compilation of all packages that depend on this change using `nix-shell -p nixpkgs-review --run "nixpkgs-review rev HEAD"`. Note: all changes have to be committed, also see [nixpkgs-review usage](https://github.com/Mic92/nixpkgs-review#usage)
- [x] Tested basic functionality of all binary files (usually in `./result/bin/`)
- [23.11 Release Notes](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2311.section.md) (or backporting [23.05 Release notes](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2305.section.md))
  - [ ] (Package updates) Added a release notes entry if the change is major or breaking
  - [ ] (Module updates) Added a release notes entry if the change is significant
  - [ ] (Module addition) Added a release notes entry if adding a new NixOS module
- [x] Fits [CONTRIBUTING.md](https://github.com/NixOS/nixpkgs/blob/master/CONTRIBUTING.md).

PR #322422, opened 2024-06-25
stage0-posix: move to top-level

GetPsyched:
## Description of changes

Towards https://github.com/ngi-nix/projects/issues/1349

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---


pkgs/top-level/all-packages.nix
```diff
@@ -27720,6 +27720,8 @@ with pkgs;
 
   smimesign = callPackage ../os-specific/darwin/smimesign { };
 
+  stage0-posix = recurseIntoAttrs (callPackage ../os-specific/linux/stage0-posix { });
```
GetPsyched, line 27723: TODO:
- Now that `stage0-posix` is discoverable, do all of the derivations within it have proper meta attributes?
- Are there extraneous attributes?

GetPsyched:
We did not verify that this new package set ends up in `packages.json` because we couldn't figure out how to build "the tarball."

emilytrau:
Apologies, but I must disagree with this approach. minimal-bootstrap was designed to be self-contained and not to be re-used in parts. It was made non-discoverable at the top-level on purpose to prevent use of bootstrapping intermediate packages that may be broken or incomplete. Bootstrap components should also not be considered a stable interface for external dependencies.

In place of this may I suggest creating a second version of stage0-posix as a top level package.

Ericson2314:
Yeah https://github.com/ngi-nix/ngipkgs/issues/304 doesn't reflect my (albeit vague) understanding of what remains on this work.

@emilytrau If you could write an issue saying what is left to do, that would be very useful! It's some badly needed context we've been sorely lacking for a long time, I think.

fricklerhandwerk:
@Emilytrau the idea was to merely expose mes and friends to make it appear in search an enable people to run it for playing around, really to raise awareness that these things exist in Nixpkgs. My understanding was that we shouldn't expose any intermediate builds, but what exactly is wrong with having mes, kaem, hex0 accessible?

emilytrau:
The versions of these packages inside minimal-bootstrap are not designed to the standards of a regular nixpkgs package. Once such issue for example - is that binaries are unable to be placed in `$out/bin` as expected. The derivations for these tools are not considered discrete and independent packages themselves, and only work within the context minimal-bootstrap. In addition minimal-bootstrap is intended to provide no API surface or stability guarantees beyond a final "tarball" of tools. This hard separation is intended to allow minimal-bootstrap to evolve freely and independently.

End users who wish to play around with stage0 or mes may be better served with a dedicated package built on top of the regular stdenv that can enable more features, better platform support, and timely updates.

fricklerhandwerk:
Yes, that was original idea for this change - not move code but expose separate packages built with stdenv (and possibly add some hints that the actual bootstrap works differently).

Ericson2314:
https://github.com/NixOS/nixpkgs/pull/260193#issuecomment-2267227955 here @philiptaron volunteered to finish the bootstrap. I think that is much more valuable than exposing mes or whatever — no one cares about those packages in isolation, just the finished project.

Right now it feels like we have an arch under construction just missing the keystone, which is quite frustrating.

philiptaron:
Yes, I'm very interested in continuing the minimal bootstrap feature. I've got a day job, so it'll be slower going than I'd like, but my commitment is to make incremental progress. I'd love co-contributors, since I also have lots to learn about this endeavor.

philiptaron:
> Yes, I'm very interested in continuing the minimal bootstrap feature. I've got a day job, so it'll be slower going than I'd like, but my commitment is to make incremental progress. I'd love co-contributors, since I also have lots to learn about this endeavor.

Update from two months down the road, no incremental progress made: I think this task is beyond the 5-10 hours a week I have to contribute to nixpkgs. It's incredibly important, but please don't let my drop in and drop out dampen things.

Issue #280535, opened 2024-01-12
Unable to use Filebeat service due to hard-coded default.


fifthecho:
### Describe the bug
The Filebeat service definition has a hardcoded default of an output to an elasticsearch instance on localhost. If I want to have a filebeat output to logstash, however, the Filebeat service [only supports one output](https://discuss.elastic.co/t/more-than-one-namespace-configured-accessing-output/146646) and with my configured output and the default, the service will not start.

### Steps To Reproduce
A configuration of: 
```
services.filebeat = {
    enable = true;
    settings = {
      output = {
        logstash = {
          hosts = ["logstash5044"];
        };
      };
    };
    inputs = {
      journald.id = "everything";
      log = {
        enabled = true;
        paths = [
          "/var/log/*.log"
        ];
      };
    };
```

Generates the following file:
```
{
  "filebeat": {
    "inputs": [
      {
        "id": "everything",
        "type": "journald"
      },
      {
        "enabled": true,
        "paths": [
          "/var/log/*.log"
        ],
        "type": "log"
      }
    ],
    "modules": []
  },
  "output": {
    "elasticsearch": {
      "hosts": [
        "127.0.0.1:9200"
      ]
    },
    "logstash": {
      "hosts": [
        "logstash:5044"
      ]
    }
  }
}

```

and a config of 

```
services.filebeat = {
    enable = true;
    settings = {
      output = {
        logstash = {
          hosts = ["logstash:5044"];
        };
        elasticsearch = {
          hosts = [];
        }
      };
    };
    inputs = {
      journald.id = "everything";
      log = {
        enabled = true;
        paths = [
          "/var/log/*.log"
        ];
      };
    };
  };
```
generates a config of 

```
{
  "filebeat": {
    "inputs": [
      {
        "id": "everything",
        "type": "journald"
      },
      {
        "enabled": true,
        "paths": [
          "/var/log/*.log"
        ],
        "type": "log"
      }
    ],
    "modules": []
  },
  "output": {
    "elasticsearch": {
      "hosts": []
    },
    "logstash": {
      "hosts": [
        "logstash:5044"
      ]
    }
  }
}

```

Neither is a config that Filebeats will accept with an error of
`Exiting: error unpacking config data: more than one namespace configured accessing 'output' (source:'/var/lib/filebeat/filebeat.yml')`

### Expected behavior
There shouldn't be a default so that a valid config file can be generated with an output other than ElasticSearch

### Metadata
[...]

PR #322425, opened 2024-06-25
nixos/filebeat: Improve options

felbinger:
## Description of changes

The implemented nixos options for filebeat could be improved.

I had trouble finding a way to disable the elasticsearch output in the first place, which was the main reason to open this pr, see: https://github.com/NixOS/nixpkgs/issues/280535

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [x] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [x] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/nixos/modules/services/logging/filebeat.nix
+++ b/nixos/modules/services/logging/filebeat.nix
@@ -22,15 +22,14 @@ let
   json = pkgs.formats.json { };
 in
 {
+  meta.maintainers = with lib.maintainers; [ felbinger ];
   options = {
 
     services.filebeat = {
 
       enable = mkEnableOption "filebeat";
 
-      package = mkPackageOption pkgs "filebeat" {
-        example = "filebeat7";
-      };
+      package = mkPackageOption pkgs "filebeat" { example = "filebeat7"; };
 
       inputs = mkOption {
         description = ''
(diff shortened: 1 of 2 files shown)


felbinger:
I'm currently thinking about a solution with a submodule for outputs, which defaults to null. This would at least provide the config tree for the option search...

felbinger:
@SuperSandro2000 Is it possible to set default values for `attrOf submodule`, which can be set like `outputs.<name>.enable` ?
I tried setting the default value [(see code)](https://github.com/NixOS/nixpkgs/pull/322425/files#diff-73d78e5309cf1e59e89c5b4a2495f15e8de4164eea0d6ebc140cfcce7f16500fR134-R154), but I don't think this is going to work.

Do you see any other way to accomplish this without removing the default option (which again is a breaking change)?

Is there also a simpler solution to filter the outputs attrs to only get the enabled [(see code)](https://github.com/NixOS/nixpkgs/pull/322425/commits/fd2bd02e53954d6f161d3dd13dca26663cfbc9a6#diff-73d78e5309cf1e59e89c5b4a2495f15e8de4164eea0d6ebc140cfcce7f16500fR245-R246)?

nixos/modules/services/logging/filebeat.nix
```diff
@@ -41,23 +47,28 @@ in
...
-        type = types.attrsOf (types.submodule ({ name, ... }: {
-          freeformType = json.type;
-          options = {
-            type = mkOption {
-              type = types.str;
-              default = name;
-              description = ''
-                The input type.
+        default = { };
+        type = types.attrsOf (
```
SuperSandro2000, line 51: I don't think the module system will now properly merge this.

I think this could be similar to https://github.com/NixOS/nixpkgs/commit/490ba58dd7a5687a5cde953161a5bcb198261012 but not fully sure.

felbinger: Looks like it is working. I tested a couple configurations and just had one issue with an assertion. Also this is nothing I changed in this commit, it was just changed due to new formatting standard which has been introduced in https://github.com/NixOS/nixpkgs/pull/326407.

PR #322496, opened 2024-06-25
binaryninja: init at 4.0.0

Rookeur:
## Things done

- Built on platform(s)
  - [x] x86_64-linux

---

### DIFF
--- a/pkgs/by-name/bi/binary-ninja/package.nix
+++ b/pkgs/by-name/bi/binary-ninja/package.nix
@@ -1,0 +1,80 @@
+{ stdenv
+, lib
+, fetchurl
+, unzip
+, autoPatchelfHook
+, dbus
+, fontconfig
+, freetype
+, fvwm3
+, glxinfo
+, libGLU
+, libxkbcommon
+, wayland
+, x11basic
+, xcbproto
+, xcbutil
+, xcbutilimage
+, xcbutilkeysyms
+, xcbutilrenderutil
+, xcbutilwm
+, xcbutilxrm
+}:
+stdenv.mkDerivation (finalAttrs: {
+  pname = "binary-ninja";
+  version = "4.0.0";
+
+  src = fetchurl {
+    url = "https://web.archive.org/web/20240625203027/https://cdn.binary.ninja/installers/binaryninja_free_linux.zip";
+    hash = "sha256-Q9CR3UVPzFzCHFrMxIdGrQ0RqL0L566MBZnGG/3nOB4=";
+  };
+
+  nativeBuildInputs = [
+    unzip
+    autoPatchelfHook
+  ];
+
+  buildInputs = [
+    dbus
+    fontconfig
+    freetype
+    fvwm3
+    glxinfo
+    libGLU
+    libxkbcommon
+    stdenv.cc.cc.lib
+    wayland
+    x11basic
+    xcbproto
+    xcbutil
+    xcbutilimage
+    xcbutilkeysyms
+    xcbutilrenderutil
+    xcbutilwm
+    xcbutilxrm
+  ];
+
+  installPhase = ''
+    runHook preInstall
+    mkdir -p $out/
+    cp -R . $out/
+
+    mkdir $out/bin
+    ln -s $out/binaryninja $out/bin/binaryninja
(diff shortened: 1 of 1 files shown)


lib/licenses.nix
```diff
@@ -149,6 +149,12 @@ in mkLicense lset) ({
     fullName = " BitTorrent Open Source License v1.1";
   };
 
+  binaryninja = {
+    fullName = "Binaryninja Non-Commercial License Agreement";
+    url = "https://docs.binary.ninja/about/license.html";
+    free = false;
+  };
```
alyssais, line 156: We can probably just inline this in the package, since it's unlikely to ever be used in any other package.

Rookeur: fixed

pkgs/by-name/bi/binary-ninja/package.nix
```diff
@@ -0,0 +1,81 @@
...
+      description = "Interactive decompiler, disassembler, debugger";
+      homepage = "https://binary.ninja/";
+      license = {
+        fullName = "Binaryninja Non-Commercial License Agreement";
+        url = "https://docs.binary.ninja/about/license.html";
+        free = false;
+      };
+      mainProgram = "binaryninja";
+      maintainers = with maintainers; [ rookeur ];
+      platforms = platforms.linux;
```
alyssais, line 79: Every architecture?

Rookeur: no linux x86_64

pkgs/by-name/bi/binary-ninja/package.nix
```diff
@@ -0,0 +1,80 @@
...
+    description = "Interactive decompiler, disassembler, debugger";
+    homepage = "https://binary.ninja/";
+    license = {
+      fullName = "Binaryninja Non-Commercial License Agreement";
+      url = "https://docs.binary.ninja/about/license.html";
+      free = false;
+    };
+    mainProgram = "binaryninja";
+    maintainers = with maintainers; [ rookeur ];
+    platforms = intersectLists platforms.linux platforms.x86;
```
alyssais, line 78: This would also include 32-bit.  I think you just want

```suggestion
    platforms = [ "x86_64-linux" ];
```

Rookeur: exactly thank you !

alyssais (commented):
I was notified about this because of my CODEOWNERS entry for licenses.nix.  The license and the other stuff I happened to notice look fine to me now, but I don't have much experience with non-source packages, so somebody else should still take a look.

AkechiShiro:
Hello @Rookeur are you still working on this PR and waiting for new reviews?

Rookeur:
Hi @AkechiShiro 

I am waiting for new reviews 😄

AkechiShiro:
I'll be trying but the version seems to be not precise enough, not sure if there is a way to grep the version string from inside the binary

The cdn URL seems to update silently the archive, so it's a bit of a pain, I believe currently there is version 4.1

Rookeur:
absolutely that is why i based my work on what others did and i noticed that many people use web archive

AkechiShiro:
Yeah that's good I guess to pin the version but I dont know how much it is present in Nixpkgs as a use case for proprietary binaries, I'm not sure how binaryninja is versioned on the AUR for instance

FliegendeWurst (approved):

GaetanLepage (commented):
nit

pkgs/by-name/bi/binary-ninja/package.nix
```diff
@@ -0,0 +1,80 @@
+{ stdenv
+, lib
+, fetchurl
```
wolfgangwalther, line 3: This would require a run of `treefmt` before merge today.

wolfgangwalther (changes requested):
@rookeur are you still active and would like to push this over the finish line? You already have an approval from @FliegendeWurst, so I guess this just needs to be rebased, formatted the `with lib` nit applied - and then checked whether everything still works.

eilvelia:
The binary ninja package was merged in #360314

PR #322512, opened 2024-06-25
Development shell with a pinned nixfmt

infinisil:
## Description of changes

This is a decently-sized step towards having a fully nixfmt-formatted Nixpkgs (https://github.com/NixOS/nixpkgs/issues/322520) :tada:. I'd like others in the @NixOS/nix-formatting to approve this.

This PR makes it such that the existing workflow to check Nix formatting gives clear instructions on how to fix unformatted files.

It does so by introducing a `shell.nix` file for Nixpkgs that provides users with a pinned `nixfmt` version that matches CI.

Along with it comes a workflow to check that `nix-build shell.nix` works on `x86_64-linux` and `aarch64-darwin` (the two relevant platforms available in GitHub Actions), such that it doesn't break randomly

This PR introduces _2_ new entries in the root of the directory tree (`shell.nix` and `ci` for some extra CI bits), exciting!

## Things done
- [x] Successfully ran `nix-shell` and `nixfmt pkgs/development/cuda-modules` in it, a directory that's already enforced to be formatted
- [x] Wrote some docs
- [x] Tested formatting CI failure: https://github.com/tweag/nixpkgs/actions/runs/9671086737/job/26681017292?pr=95
  - And successful: https://github.com/tweag/nixpkgs/actions/runs/9671095373/job/26681045327?pr=95
- [x] Tested nix-shell CI: https://github.com/tweag/nixpkgs/actions/runs/9671095369?pr=95
- [x] Added myself to codeowners of new files, and the security team for critical ones
- [x] Set myself as code owner of the formatter workflow file (can't do the formatting team because it doesn't have write access)

---

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -13,6 +13,7 @@
 
 # CI
 /.github/workflows @NixOS/Security @Mic92 @zowoq
+/.github/workflows/check-nix-format.yml @infinisil
 /ci @infinisil
 
 # Develompent support
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -15,8 +15,9 @@
 /.github/workflows @NixOS/Security @Mic92 @zowoq
 /ci @infinisil
 
-# EditorConfig
+# Develompent support
 /.editorconfig @Mic92 @zowoq
+/shell.nix @infinisil @NixOS/Security
 
 # Libraries
 /lib                        @infinisil
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -11,8 +11,9 @@
 # This also holds true for GitHub teams. Since almost none of our teams have write
 # permissions, you need to list all members of the team with commit access individually.
 
-# GitHub actions
+# CI
 /.github/workflows @NixOS/Security @Mic92 @zowoq
+/ci @infinisil
 
 # EditorConfig
 /.editorconfig @Mic92 @zowoq
--- a/.github/workflows/check-nix-format.yml
+++ b/.github/workflows/check-nix-format.yml
@@ -40,14 +40,14 @@ jobs:
           NIX_FMT_PATHS_BSD: pkgs/os-specific/bsd
           NIX_FMT_PATHS_MPVSCRIPTS: pkgs/applications/video/mpv/scripts
           # Format paths related to the Nixpkgs CUDA ecosystem.
-          NIX_FMT_PATHS_CUDA: |
+          NIX_FMT_PATHS_CUDA: |-
             pkgs/development/cuda-modules
             pkgs/test/cuda
             pkgs/top-level/cuda-packages.nix
-          NIX_FMT_PATHS_MAINTAINERS: |
+          NIX_FMT_PATHS_MAINTAINERS: |-
             maintainers/maintainer-list.nix
             maintainers/team-list.nix
-          NIX_FMT_PATHS_K3S: |
+          NIX_FMT_PATHS_K3S: |-
             nixos/modules/services/cluster/k3s
             nixos/tests/k3s
             pkgs/applications/networking/cluster/k3s
--- a/.github/workflows/check-nix-format.yml
+++ b/.github/workflows/check-nix-format.yml
@@ -19,13 +19,18 @@ jobs:
         with:
           # pull_request_target checks out the base branch by default
           ref: refs/pull/${{ github.event.pull_request.number }}/merge
+      - name: Get Nixpkgs revision for nixfmt
+        run: |
+          # pin to a commit from nixpkgs-unstable to avoid e.g. building nixfmt
+          # from staging
+          # This should not be a URL, because it would allow PRs to run arbitrary code in CI!
+          rev=$(jq -r .rev ci/pinned-nixpkgs.json)
+          echo "url=https://github.com/NixOS/nixpkgs/archive/$rev.tar.gz" >> "$GITHUB_ENV"
       - uses: cachix/install-nix-action@8887e596b4ee1134dae06b98d573bd674693f47c # v26
         with:
           # explicitly enable sandbox
           extra_nix_config: sandbox = true
-          # fix a commit from nixpkgs-unstable to avoid e.g. building nixfmt
-          # from staging
-          nix_path: nixpkgs=https://github.com/NixOS/nixpkgs/archive/4b455dc2048f73a79eb3713f342369ff58f93e0b.tar.gz
+          nix_path: nixpkgs=${{ env.url }}
       - name: Install nixfmt
         run: "nix-env -f '<nixpkgs>' -iAP nixfmt-rfc-style"
       - name: Check that Nix files are formatted according to the RFC style
--- a/.github/workflows/check-nix-format.yml
+++ b/.github/workflows/check-nix-format.yml
@@ -56,6 +56,7 @@ jobs:
           NIX_FMT_PATHS_BUILD_SUPPORT_PHP: pkgs/build-support/php
         # Iterate over all environment variables beginning with NIX_FMT_PATHS_.
         run: |
+          unformattedPaths=()
           for env_var in "${!NIX_FMT_PATHS_@}"; do
             readarray -t paths <<< "${!env_var}"
             if [[ "${paths[*]}" == "" ]]; then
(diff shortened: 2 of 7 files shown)


.github/CODEOWNERS
```diff
@@ -11,11 +11,15 @@
...
+# CI
 /.github/workflows @NixOS/Security @Mic92 @zowoq
+/.github/workflows/check-nix-format.yml @infinisil
+/ci @infinisil
 
-# EditorConfig
+# Develompent support
 /.editorconfig @Mic92 @zowoq
+/shell.nix @infinisil
+/.envrc @infinisil
```
mweinelt, line 22:
```suggestion
/shell.nix @infinisil @NixOS/Security
/.envrc @infinisil @NixOS/Security
```

I think it would be good to involve the security team.

infinisil: Done!

Frontear:
Not a fan of a `.envrc` being added by default. My reason is fairly simple, I like to put my `GITHUB_TOKEN` for `nixpkgs-review --post-result`, and I have a global ignore to prevent my `nixpkgs` fork from seeing it, so that whenever I `git add -A` any changes, they ignore the .envrc.

Having it default included means its tracked, which means per-user secrets have a chance to end up accidentally in a commit. Sure this is a user error and isn't necessarily nixpkgs responsibility, but I honestly think I'd prefer it being left out and left to the user to create and maintain.

.github/CODEOWNERS
```diff
@@ -33,6 +37,7 @@
 /lib/tests/modules.sh       @infinisil @roberth
 /lib/tests/modules          @infinisil @roberth
 
+
```
drupol, line 40: Unrelated changes

.github/workflows/check-nix-format.yml
```diff
@@ -19,13 +19,18 @@ jobs:
         with:
           # pull_request_target checks out the base branch by default
           ref: refs/pull/${{ github.event.pull_request.number }}/merge
+      - name: Get Nixpkgs revision for nixfmt
+        run: |
+          # pin a commit from nixpkgs-unstable to avoid e.g. building nixfmt
```
drupol, line 24:
```suggestion
          # pin to a commit from nixpkgs-unstable to avoid e.g. building nixfmt
```

.github/workflows/check-shell.yml
```diff
@@ -0,0 +1,29 @@
+name: Check shell.nix
```
drupol, line 1:
```suggestion
name: "Check shell"
```

To make it consistent with the others

ci/README.md
```diff
@@ -0,0 +1,12 @@
+# CI support files
+
+This directory contains files to support CI, such as [GitHub Actions](https://github.com/NixOS/nixpkgs/tree/master/.github/workflows) and [Ofborg](https://github.com/nixos/ofborg).
+This is in contrast with [`maintainers/scripts`](`../maintainers/scripts`) which is for human use instead.
+
+## Pinned Nixpkgs
+
+CI may need certain packages from Nixpkgs.
+In order to ensure that the needed packages are generally available without building,
+[`pinned-nixpkgs.json`](./pinned-nixpkgs.json) contains pins a Nixpkgs version tested by Hydra.
```
drupol, line 10:
```suggestion
[`pinned-nixpkgs.json`](./pinned-nixpkgs.json) contains a pinned Nixpkgs version tested by Hydra.
```

drupol (commented):
Cool!

A few nits and it's good for me.

infinisil:
@Frontear It's very standard to commit the `.envrc` file, I don't think it's worth everybody having to add it manually for such edge cases. Here are some workarounds you can use:
- Unset `GITHUB_TOKEN` and install the GitHub CLI (`pkgs.github-cli`) instead, which [nixpkgs-review can use](https://github.com/Mic92/nixpkgs-review/blob/b85f5b097bf23480c0d532269aaa50154ed88b06/nixpkgs_review/cli/__init__.py#L147-L150) to get the token
- Set the token in `~/.config/direnv/direnvrc` instead, which gets loaded before `.envrc`
- Use `git update-index --assume-unchanged .envrc` for Git to never try adding it

Frontear:
I've always believed envrc shouldnt be commited due to it holding environment secrets, unless im conflating the role of .env with .envrc. In any case thats fair that everyone needing to manually make one is tedious.

shell.nix
```diff
@@ -0,0 +1,29 @@
+# A shell to get tooling for Nixpkgs development
+#
+# Note: We intentionally don't use Flakes here,
+# because it would require copying all of Nixpkgs every time the shell is refreshed.
```
Mic92, line 4: Can you clarify where this copy would be created? I only see a fetchTarball below, so this also copies nixpkgs bascially.

MattSturgeon: Maybe the comment is referring to how flake repos are copied to the nix store before being evaluated?

infinisil: Yeah, running `nix develop` will copy the entire tree into the store, and when you change _any_ file in Nixpkgs and run `nix develop` again, it creates another copy.

This might work on smaller repos, but with Nixpkgs, these copies are like 500MB each.

I'll improve the wording a bit.

Mic92: Ah, yeah. In that case, we would need to put the flake.nix in a subdirectory.

infinisil:
Since I don't want this PR to focus on the direnv part, I split that into a separate PR that depends on this one: https://github.com/NixOS/nixpkgs/pull/322650

shell.nix
infinisil: @Mic92 That doesn't work, it still copies the entire Git repo

tomberek (approved):

dasJ (approved):

.github/CODEOWNERS
Frontear, line 1: Unrelated to this pr but removing @\samueldr might be a good idea at some point.

Frontear (approved):

ci/update-pinned-nixpkgs.sh
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env nix-shell
+#!nix-shell -i bash -p jq
+
+set -euo pipefail
+
+# https://stackoverflow.com/a/246128
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+
+repo=https://github.com/nixos/nixpkgs
```
0x4A6F, line 9:
```diff
-repo=https://github.com/nixos/nixpkgs
+repo=https://github.com/NixOS/nixpkgs
```

0x4A6F (approved):

.github/CODEOWNERS
MattSturgeon: Wasn't that done in https://github.com/NixOS/nixpkgs/pull/323228?

Frontear: Then this must be using an older revision of nixpkgs before the change, because on this branch he is still listed (see line 121 122)

MattSturgeon: This PR was started 4 days ago. https://github.com/NixOS/nixpkgs/pull/323228 was merged 14hrs ago.

ci/update-pinned-nixpkgs.sh
infinisil: GitHub also works with the lowercase version, so I don't think this matters nor worth changing, even if only to set a precedent for less bikeshedding :P

infinisil:
Since this has been approved by 3 other members of the [formatting team](https://nixos.org/community/teams/formatting/), let's go ahead with merging this ;)

infinisil:
Manual backport of this is in https://github.com/NixOS/nixpkgs/pull/326630

infinisil:
@alyssais Oh damn, yeah that's something that didn't occur to me in advance. Looks like `nix-shell . -A $pkg` works, which seems satisfactory to me. But we could also make `nix-shell -A pkgs.$pkg` work using `passthru.pkgs = import ./.` in `shell.nix`.

alyssais:
Learning to add a . isn't too bad, so I think it would only be worth doing something if it didn't break people's workflows at all, but it's still a shame it broke.

raboof:
> @alyssais Oh damn, yeah that's something that didn't occur to me in advance. Looks like `nix-shell . -A $pkg` works, which seems satisfactory to me.

Does it? It doesn't seem to for me (CppNix 2.24.9). `nix-shell default.nix -A $pkg` or `nix-shell '<nixpkgs/default.nix>' -A $pkg` does work.

alyssais:
Yeah, it got even worse with Nix 2.24: https://github.com/NixOS/nix/issues/11699

roberth:
nix#11699 does fix five reported issues, so it would be unfortunate to regress those now.
While I'm open to refining the new rules for `nix-shell`/`shell.nix`, they would almost certainly make the final logic more complicated and harder to understand, but probably only fix half of this `nix-shell -A` issue.

For these reasons I think it is preferable to fix it completely on the Nixpkgs side
- https://github.com/NixOS/nixpkgs/pull/353240

Issue #99014, opened 2020-09-28
xorg.fontbhlucidatypewriter100dpi: gentoo says it's public domain, nixpkgs says it's unfree - who's right?


mkg20001:
Getting the following error

```
error: Package ‘font-bh-lucidatypewriter-100dpi-1.0.3’ in /home/maciej/Projekte/nixpkgsv/pkgs/servers/x11/xorg/default.nix:241 has an unfree license (‘unfree’), refusing to evaluate.
```

xorg/overrides.nix says it's unfree

```
    # unfree, possibly not redistributable
    unfree = [
      # no license, just a copyright notice
      "fontbhlucidatypewriter100dpi"
      "fontbhlucidatypewriter75dpi"
      "fontdaewoomisc"

      # unclear license, "permission to use"?
      "fontjismisc"
    ];
```

gentoo package says it's public domain

https://packages.gentoo.org/packages/media-fonts/font-bh-lucidatypewriter-100dpi

Now the question is: Which one is correct? Should we just "go with the flow"?

jonringer:
related: #99136

rnhmjoj:
This came up while I was working on converting core X11 fonts to newer formats.
Before distributing modifications which we are not allowed to make I checked what the actual license was and discovered many fonts were actually proprietary with only a permission to "use" or "display" and distribution given to the X.org.

Regarding lucidatypewriter, the [source code](https://gitlab.freedesktop.org/xorg/font/bh-lucidatypewriter-75dpi/-/blob/master/COPYING) doesn't contain any license information except a copyright notice, from which I assumed the author reserver all their rights. I don't known where gentoo is getting that information. I'll try asking around.

rnhmjoj:
So, I asked at #xorg and they have no idea. 
I've found this [document](https://www.x.org/releases/X11R7.7/doc/xorg-docs/License.html) which contains a license agreement with B&H (the designers) for the use of the Luxi fonts but there is no mention of other B&H fonts.
My theory is that this agreement is included here because it was made directly with the (defunct) X Consortium in 2001 when Luxi was released but the Lucida fonts are much older (1984) so the agreement must have been with an older organization (MIT, [OSF](https://en.wikipedia.org/wiki/Open_Software_Foundation), whatever) and someone never bother to copy the legal notice when the font was included into X.

In any case, since licenses for Lucida are being sold by Linotype, I doubt that it's in the public domain and can be considered free, by the FSF definition of it.

[...]

PR #322521, opened 2024-06-26
xorg.*: add licensing information

quantenzitrone:
## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/lib/licenses.nix
+++ b/lib/licenses.nix
@@ -1259,6 +1259,11 @@ in mkLicense lset) ({
     fullName = "X11 License";
   };
 
+  x11DistributeModifications = {
+    spdxId = "X11-distribute-modifications-variant";
+    fullName = "X11 License Distribution Modification Variant";
+  };
+
   xfig = {
     spdxId = "Xfig";
     fullName = "xfig";
--- a/lib/licenses.nix
+++ b/lib/licenses.nix
@@ -134,6 +134,11 @@ in mkLicense lset) ({
     fullName = "Apache License 2.0 with LLVM Exceptions";
   };
 
+  bitstreamCharter = {
+    spdxId = "Bitstream-Charter";
+    fullName = "Bitstream Charter Font License";
+  };
+
   bitstreamVera = {
     spdxId = "Bitstream-Vera";
     fullName = "Bitstream Vera Font License";
@@ -222,6 +227,11 @@ in mkLicense lset) ({
   bsdProtection = {
     spdxId = "BSD-Protection";
     fullName = "BSD Protection License";
+  };
+
+  bsdSourceCode = {
+    spdx = "BSD-Source-Code";
+    fullName = "BSD Source Code Attribution";
   };
 
   bsl11 = {
@@ -615,6 +625,11 @@ in mkLicense lset) ({
     fullName = "Historic Permission Notice and Disclaimer";
   };
 
+  hpndDocSell = {
+    spdxId = "HPND-doc-sell";
+    fullName = "Historical Permission Notice and Disclaimer - documentation sell variant";
+  };
+
   hpndSellVariant = {
     fullName = "Historical Permission Notice and Disclaimer - sell variant";
     spdxId = "HPND-sell-variant";
@@ -858,6 +873,11 @@ in mkLicense lset) ({
     fullName = "Enlightenment License (e16)";
   };
 
+  mitOpenGroup = {
+    spdxId = "MIT-open-group";
+    fullName = "MIT Open Group variant";
+  };
+
   mit0 = {
     spdxId = "MIT-0";
     fullName = "MIT No Attribution";
@@ -937,6 +957,11 @@ in mkLicense lset) ({
   nposl3 = {
     spdxId = "NPOSL-3.0";
     fullName = "Non-Profit Open Software License 3.0";
+  };
+
+  ntp = {
+    spdxId = "NTP";
+    fullName = "NTP License";
   };
 
   nvidiaCuda = {
--- a/lib/licenses.nix
+++ b/lib/licenses.nix
@@ -843,13 +843,11 @@ in mkLicense lset) ({
     url = "https://opensource.org/licenses/MirOS";
   };
 
-  # spdx.org does not (yet) differentiate between the X11 and Expat versions
-  # for details see https://en.wikipedia.org/wiki/MIT_License#Various_versions
   mit = {
     spdxId = "MIT";
     fullName = "MIT License";
   };
-  # https://spdx.org/licenses/MIT-feh.html
+
   mit-feh = {
     spdxId = "MIT-feh";
     fullName = "feh License";
--- a/lib/licenses.nix
+++ b/lib/licenses.nix
@@ -45,6 +45,16 @@ in mkLicense lset) ({
     free = false;
   };
 
+  adobeUtopia = {
+    spdxId = "Adobe-Utopia";
+    fullName = "Adobe Utopia Font License";
+    # permission to use and share is granted, but not explicitly to improve, but
+    # trademark and attribution only have to appear in "unmodified copies", which
+    # could mean that modified copies are allowed. Fedora allows this license.
+    # https://docs.fedoraproject.org/en-US/legal/allowed-licenses/#_allowed_font_licenses
+    free = true;
+  };
+
   afl20 = {
     spdxId = "AFL-2.0";
     fullName = "Academic Free License v2.0";
@@ -395,6 +405,11 @@ in mkLicense lset) ({
   cecill-c = {
     spdxId = "CECILL-C";
     fullName  = "CeCILL-C Free Software License Agreement";
+  };
+
+  cronyx = {
+    spdxId = "Cronyx";
+    fullName = "Cronyx License";
   };
 
   cpal10 = {
--- a/lib/licenses.nix
+++ b/lib/licenses.nix
@@ -1049,7 +1049,7 @@ in mkLicense lset) ({
   };
 
   purdueBsd = {
-    fullName = " Purdue BSD-Style License"; # also know as lsof license
+    fullName = "Purdue BSD-Style License"; # also known as lsof license
     url = "https://enterprise.dejacode.com/licenses/public/purdue-bsd";
   };
 
--- a/lib/licenses.nix
+++ b/lib/licenses.nix
@@ -628,7 +628,7 @@ in mkLicense lset) ({
   # Intel's license, seems free
   iasl = {
     spdxId = "Intel-ACPI";
-    fullName = "iASL";
+    fullName = "Intel ACPI Software License Agreement";
     url = "https://old.calculate-linux.org/packages/licenses/iASL";
   };
 
--- a/lib/licenses.nix
+++ b/lib/licenses.nix
@@ -879,7 +879,6 @@ in mkLicense lset) ({
     # - SPDX Lists the license (meaning it is free according to their inclusion policy)
     # - Fedora and other Linux Distributions allow this License:
     # - https://docs.fedoraproject.org/en-US/legal/allowed-licenses/#_allowed_font_licenses (accessed 2024-06-27)
-    # - Wikipedia says the license is "commercial": https://en.wikipedia.org/w/index.php?title=Lucida&oldid=1182121399
     free = true;
   };
 
--- a/lib/licenses.nix
+++ b/lib/licenses.nix
@@ -867,6 +867,22 @@ in mkLicense lset) ({
     fullName = "Lucent Public License v1.02";
   };
 
+  lucidaBitmapFonts = {
+    spxId = "Lucida-Bitmap-Fonts";
+    fullName = "Lucida Bitmap Fonts License";
+    # We consider this as free which is debatable, however
+    # > use […] in individual and commercial software
+    # usage in software implies redistribution
+    # > Users are granted […] to use the trademark […] if the fonts […] are not modified […].
+    # inplies that modification is allowed
+    # Other Opinions:
+    # - SPDX Lists the license (meaning it is free according to their inclusion policy)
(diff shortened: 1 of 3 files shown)


quantenzitrone:
this is already like 8 hours of work
and I'm not even half way done lol

pkgs/servers/x11/xorg/overrides.nix
```diff
@@ -1,109 +1,725 @@
...
+        # some unknown permission notices that grant all 4 freedoms
+        free
+        # quite a few files carry this copyright notice:
+        # > (c) Copyright 1995 FUJITSU LIMITED
+        # > This is source code modified by FUJITSU LIMITED under the Joint
+        # > Development Agreement for the CDE/Motif PST.
+        # commit 5e7d589697755a70fb22d85c6a1ae82b39843e53 removes 2 files
+        # with the same copyright notice and states:
+        # > This is unpublished proprietary source code of FUJITSU LIMITED
+        unfree
```
jopejoe1, line 669: I just did a grep on the source we are pulling, and the only reference to `unpublished proprietary` I can find is in the Changelog. Maybe verify from what source you're checking.

quantenzitrone: Yes, "This is unpublished proprietary source code of FUJITSU LIMITED" is only part of the changelog.
My assumption was that this applies to all files that have the same Copyright statement.

emilazy:
Just noting what I said in Matrix: SPDX (which generally only includes FOSS licenses) lists the [`Adobe-Utopia`](https://spdx.org/licenses/Adobe-Utopia.html) and [`Lucida-Bitmap-Fonts`](https://spdx.org/licenses/Lucida-Bitmap-Fonts.html) licences that are considered non‐free here, and Fedora explicitly includes them in their [Allowed Font Licenses](https://docs.fedoraproject.org/en-US/legal/allowed-licenses/#_allowed_font_licenses) list, which has [strict FOSS criteria](https://docs.fedoraproject.org/en-US/legal/license-approval/#_licenses_allowed_for_fonts). Many legacy licences that are generally considered acceptably FOSS (if unfortunately vague by modern standards) by the lawyers have similarly vague wording, and especially `Lucida-Bitmap-Fonts` seems to clearly permit derivative works to me.

While correctly recording licence information is a good thing, I think we should require very strong documentation or evidence to consider software that the sticklers at both Fedora and Debian explicitly consider FOSS to be unfree, and in my opinion this PR errs much too far on the side of considering vague but clearly intended as permissive licences to be restrictive. I’m not sure what the deal with that Fujitsu code is, but I think it should be raised upstream with X.org for clarification rather than assuming that they must have gotten it wrong and breaking the entire OS.

emilazy:
Ah, sorry; I see that the Utopia case has already been corrected. The interpretation of the Lucida licence is still incorrect, however:

> A royalty-free, nonexclusive trademark license to refer to the code and output as "OPEN LOOK" compatible is available from AT&T if, and only if, the appearance of the icons or glyphs is not changed in any manner except as absolutely necessary to accommodate the standard resolution of the screen or other output device …

Modifications are only restricted for use of the “OPEN LOOK” trademark licence. I don’t know if that trademark is even still valid, but it’s not relevant to the *copyright* licence that determines FOSS status. See also: Mozilla’s (historically quite stringent) restrictions on use of the Firefox trademark. The fact that the trademark licence follows a more blanket copyright grant, and includes additional conditions on modification, makes the intent to grant *copyright* permission for derivative works clear.

pkgs/servers/x11/xorg/overrides.nix
emilazy: The removed Fujitsu files have the following copyright notice:

```
XCOMM Copyright 1995 by FUJITSU LIMITED
XCOMM This is source code modified by FUJITSU LIMITED under the Joint
XCOMM Development Agreement for the CDEnext PST.
XCOMM This is unpublished proprietary source code of FUJITSU LIMITED
```

The ones still present in the libX11 tree have this one:

```
 * Copyright 1995 by FUJITSU LIMITED
 * This is source code modified by FUJITSU LIMITED under the Joint
 * Development Agreement for the CDE/Motif PST.
```

This differs both in omitting the questioned “This is unpublished proprietary source code of FUJITSU LIMITED” notice and in the name of the cited Joint Development Agreement (though presumably there is some relation). I checked a few files with these notices and they all were in files that already contained FOSS licence terms attached to a Toshiba copyright notice (perhaps their partner in the Joint Development Agreement?). Therefore, I believe they function as simple notices of Fujitsu’s copyright in the files, rather than being indications of a lack of accompanying copyright licence. I don’t think there is any reason to believe that this code is non‐free, but I would again suggest raising it with upstream first if you have concerns about it.

lib/licenses.nix
```diff
@@ -867,6 +867,22 @@ in mkLicense lset) ({
...
+    # We consider this as free which is debatable, however
+    # > use […] in individual and commercial software
+    # usage in software implies redistribution
+    # > Users are granted […] to use the trademark […] if the fonts […] are not modified […].
+    # inplies that modification is allowed
+    # Other Opinions:
+    # - SPDX Lists the license (meaning it is free according to their inclusion policy)
+    # - Fedora and other Linux Distributions allow this License:
+    # - https://docs.fedoraproject.org/en-US/legal/allowed-licenses/#_allowed_font_licenses (accessed 2024-06-27)
+    # - Wikipedia says the license is "commercial": https://en.wikipedia.org/w/index.php?title=Lucida&oldid=1182121399
```
emilazy, line 882: In the Wikipedia article this is expanded on as follows:

> Lucida is still licensed commercially through the TUG store[5] as well through their own web store.[6] The fonts are occasionally updated.

Indeed, the TUG site [will license you OpenType and Type 1](https://www.tug.org/store/lucida/index.html) Lucida fonts which are presumably not available under a FOSS licence. However, being based on the same typeface design does not mean that the copyright in the fonts is the same; indeed, in the US typeface designs and bitmap fonts are not copyrightable at all, whereas complex vector fonts are treated as software and are eligible for copyright protection.

I don’t think Wikipedia is expressing any opinion on the licence text of this specific distribution of Lucida bitmap fonts; it does not mention the bitmap fonts or their inclusion in X11 at all.

quantenzitrone:
I'll squash the commits when I'm done, until I'll just append, so reviewers know what changed and only have to review the new changes

pkgs/servers/x11/xorg/overrides.nix
quantenzitrone: oh i kind of assumed the last sentence was a commentary by the commiter
dumb of me

quantenzitrone: https://gitlab.freedesktop.org/xorg/lib/libx11/-/issues/217

emilazy: In my view, there’s no reason to treat this differently to e.g.

```c
/*
 * Copyright 2024 Author 1
 * Copyright 2024 Author 2
 *
 * […standard MIT licence text…]
 */
```

It’s true that the licence text says TOSHIBA specifically, but only in terms of e.g. non‐advertisement and warranty disclaimer, not in terms of the basic copyright grant. The copyright notice being in a separate comment block, without any statement of differing licence terms or reservation of rights, doesn’t indicate any negation of the licence terms.

You could object that there’s no explicit statement that Fujitsu also releases its code under that licence, but that also applies to the very conventional example I gave – whether the owner of the copyright releases their work under the licence isn’t really determined by that, although I would consider contributing to a file with an existing FOSS licence notice without explicitly reserving rights to be an indicator of intent in that regards. For instance, GitHub pull requests like this one are assumed to be licensed under the project’s licence despite usually not containing any notice of such because the Terms of Service obligate the submitting user to agree to this (and also presumably because the intent is also just obvious, regardless of your feelings on the enforceability of website ToSes). I presume that there were agreements between Toshiba and Fujitsu that lead to it being released under a FOSS licence to begin with – either at the time it was written or as part of a later opening up of the code.

I think under the general presumption (that we can’t avoid in a huge distribution project such as this) that code which says it is licensed a certain way should be treated that way unless there is strong reason to believe otherwise, there’s no cause for concern here, especially given that we know that Fujitsu did explicitly put a proprietary disclaimer on other files that they may not have wanted released. (Although, even then I sort of suspect that it was just a hold‐over notice that they forgot to remove when deciding to release code, and only removed from the modern codebase out of caution because the pre‐Unicode locale files were obsolete anyway.) Maybe someone on the X.org side knows the full history here about what was going on 30 years ago though…

lib/licenses.nix
```diff
@@ -387,6 +407,11 @@ in mkLicense lset) ({
     fullName  = "CeCILL-C Free Software License Agreement";
   };
 
+  cronyx = {
+    spdxId = "Cronyx";
+    fullName = "Cronyx License";
+  };
```
emilazy, line 413:
```
This software may be used, modified, copied, distributed, and sold,
in both source and binary form provided that the above copyright
and these terms are retained. Under no circumstances is the author
responsible for the proper functioning of this software, nor does
the author assume any responsibility for damages incurred with its use.
```

Seems obviously FOSS?

emilazy: Oh, I guess `free = true;` is the default? Sorry!

pkgs/servers/x11/xorg/overrides.nix
```diff
@@ -1,109 +1,727 @@
...
+      license = lib.licenses.unfree;
+    };
+  });
+
+  fontbhlucidatypewriter100dpi = prev.fontbhlucidatypewriter100dpi.overrideAttrs (attrs: {
+    meta = attrs.meta // {
+      # only includes a copyright
+      license = lib.licenses.unfree;
+    };
+  });
```
emilazy, line 218: These have previously been covered here: https://github.com/NixOS/nixpkgs/issues/99014. See especially https://github.com/NixOS/nixpkgs/issues/99014#issuecomment-827859023. It looks like it may be a casualty of repository splitting. I think these are as fine as the other Lucida fonts, and Debian and Fedora have them marked accordingly.

pkgs/servers/x11/xorg/overrides.nix
```diff
@@ -1,109 +1,727 @@
...
+    meta = attrs.meta // {
+      license = lib.licenses.cronyx;
+    };
+  });
+
+  fontcursormisc = prev.fontcursormisc.overrideAttrs (attrs: {
+    meta = attrs.meta // {
+      # whatever this is supposed to mean:
+      # "These ""glyphs"" are unencumbered"
+      license = lib.licenses.unfree;
```
emilazy, line 294: Pretty standard *ye olde style* somewhere‐between‐permissive‐and‐public‐domain notice. Probably not even copyrightable in the country of origin (remember that typefaces and bitmap fonts are considered too functional to copyright in the US).

You should cross‐reference all of these you can with [Fedora’s package specification](https://src.fedoraproject.org/rpms/xorg-x11-fonts/blob/21908b7629fd3c15cf24b8f0fb63a221e0a3d655/f/xorg-x11-fonts.spec); in this case:

```
# SPDX: LicenseRef-Fedora-UltraPermissive
Source16:   https://www.x.org/pub/individual/font/font-cursor-misc-1.0.3.tar.bz2
```

pkgs/servers/x11/xorg/overrides.nix
```diff
@@ -134,171 +752,146 @@ self: super:
...
+        mit
+        mitOpenGroup
+        smlnj
+        hpndSellVariant
+        bsdOriginalUC
+        bsd2
+        # lib/font/fontfile/gunzip.c
+        # > intended for inclusion in X11 public releases.
+        # as well as the last license in COPYING grants no freedoms
+        unfree
```
emilazy, line 872: In the same way that submitting code to Nixpkgs intended for inclusion in its public releases constitutes obvious intent to license your code under the terms those public releases are released under, I’m quite sure the same applies here.

lib/licenses.nix
quantenzitrone: yes i only included `free = true` with an explaination text for licenses that have debateable freeness

quantenzitrone: yes i only included `free = true` with an explaination text for licenses that have debateable freeness

quantenzitrone: yes i only included `free = true` with an explaination text for licenses that have debateable freeness

quantenzitrone: my github phone app told me that there was an error while posting my comment, so i retried a few times resulting 5 comments (i deleted 4 of them now)

pkgs/servers/x11/xorg/overrides.nix
quantenzitrone: i see your point, however i would interpret
> intended for inclusion in X11 public releases.
as "exclusive license to redistribute in X11 public releases"
are xorg packages in nixpkgs public releases of X11? maybe

emilazy: I think that interpretation is reading too much into the notice, but perhaps it’d be a good idea to report this upstream as well? It’s at least more ambiguous than the Fujitsu stuff.

quantenzitrone:
i just noticed that the [spdx license identification tool](https://tools.spdx.org/app/check_license/) gives a lot of false results
e.g. any x11-like license that does not specifically say XFree86 or X Consortium is detected as X11-distribute-modifications-variant, even if the license doesn't have the "distribute modifications" added.

jopejoe1:
For license identification i recommend using [SPDX license diff browser extension](https://addons.mozilla.org/en-US/firefox/addon/spdx-license-diff/) or [ScanCode](https://github.com/nexB/scancode-toolkit)

PR #319134, opened 2024-06-11
vue-language-server: init at 2.0.21

friedow:
## Description of changes

This adds the official vue-language-server (called @vue/language-server in npm) package. This is needed for example for the neovim lsp integration (https://github.com/vuejs/language-tools?tab=readme-ov-file#community-integration).

Furethermore this PR removes the deprecated nodePackages.@volar/vue-language-server.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
- For non-Linux: Is sandboxing enabled in `nix.conf`? (See [Nix manual](https://nixos.org/manual/nix/stable/command-ref/conf-file.html))
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:
  - [NixOS test(s)](https://nixos.org/manual/nixos/unstable/index.html#sec-nixos-tests) (look inside [nixos/tests](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests))
  - and/or [package tests](https://github.com/NixOS/nixpkgs/blob/master/pkgs/README.md#package-tests)
  - or, for functions and "core" functionality, tests in [lib/tests](https://github.com/NixOS/nixpkgs/blob/master/lib/tests) or [pkgs/test](https://github.com/NixOS/nixpkgs/blob/master/pkgs/test)
  - made sure NixOS tests are [linked](https://nixos.org/manual/nixpkgs/unstable/#ssec-nixos-tests-linking) to the relevant packages
- [ ] Tested compilation of all packages that depend on this change using `nix-shell -p nixpkgs-review --run "nixpkgs-review rev HEAD"`. Note: all changes have to be committed, also see [nixpkgs-review usage](https://github.com/Mic92/nixpkgs-review#usage)
- [ ] Tested basic functionality of all binary files (usually in `./result/bin/`)
- [24.11 Release Notes](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2411.section.md) (or backporting [23.11](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2311.section.md) and [24.05](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2405.section.md) Release notes)
  - [ ] (Package updates) Added a release notes entry if the change is major or breaking
[...]

Issue #316908, opened 2024-06-03
`pnpm.configHook` / `npmHooks.npm{Install,Build}Hook`: Support `pnpm-workspace.yaml` projects


doronbehar:
Per the issue mentioned here:

https://github.com/NixOS/nixpkgs/pull/290715#issuecomment-2144543728

And [the docs](https://nixos.org/manual/nixpkgs/unstable/#javascript-pnpm-sourceRoot) (link will be correct once https://github.com/NixOS/nixpkgs/pull/290715 reaches the channels).

We need to think about a way to support projects that use `pnpm-workspace.yaml`.

cc @Scrumplex .

nathanscully:
I did a test for workspaces using the exiting build and linked it in #290715 the other week: https://github.com/NixOS/nixpkgs/pull/290715#issuecomment-2118801278. It's probably non-ideal as its using the inbuilt pnpm filter spread syntax (` pnpm run --filter server... build`) but it does work. 
Not sure if we can get away just documenting that for now rather than having to change the tooling to much to get the first PR in.

doronbehar:
For `bash-language-server`, I tried in the `buildPhase`:

```
pnpm --filter server run compile
pnpm run --filter server compile
```

And all of them gave me:

```
No projects matched the filters in "/build/source"
```

Which is peculiar... Running without a `--filter server` gave me the same errors as using `npmHooks.npmBuildHook` gave me...

nathanscully:
This seems to work for me. Can you try it? Might need to move around the out folders but it builds. 

```
{
  lib,
  stdenv,
  fetchFromGitHub,
  pnpm_9,
  nodejs,
  npmHooks,
  writeScriptBin,
}:

stdenv.mkDerivation (finalAttrs: {
  pname = "bash-language-server";
  version = "5.3.3";

  src = fetchFromGitHub {
    owner = "bash-lsp";
    repo = "bash-language-server";
    rev = "server-${finalAttrs.version}";
    hash = "sha256-4+imeVjosEspFhWzjtUp2IZx7aZIx56g8M03V3dEGVs=";
  };

  nativeBuildInputs = [
    nodejs
    pnpm_9.configHook
  ];
  pnpmDeps = pnpm_9.fetchDeps {
    inherit (finalAttrs) pname version src;
    hash = "sha256-W7BrUh0tGoG5vTuR3hXDbUTJFDiKET+QvIvCwewfZ2I=";
  };
  preBuild = ''
[...]

PR #238886, opened 2023-06-21
wrangler: autopatchelf bundled binaries

kirillrdy:
###### Description of changes

hopefully fixes https://github.com/NixOS/nixpkgs/issues/235329

###### Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [x] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
- [ ] For non-Linux: Is `sandbox = true` set in `nix.conf`? (See [Nix manual](https://nixos.org/manual/nix/stable/command-ref/conf-file.html))
- [ ] Tested, as applicable:
  - [NixOS test(s)](https://nixos.org/manual/nixos/unstable/index.html#sec-nixos-tests) (look inside [nixos/tests](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests))
  - and/or [package tests](https://nixos.org/manual/nixpkgs/unstable/#sec-package-tests)
  - or, for functions and "core" functionality, tests in [lib/tests](https://github.com/NixOS/nixpkgs/blob/master/lib/tests) or [pkgs/test](https://github.com/NixOS/nixpkgs/blob/master/pkgs/test)
  - made sure NixOS tests are [linked](https://nixos.org/manual/nixpkgs/unstable/#ssec-nixos-tests-linking) to the relevant packages
- [ ] Tested compilation of all packages that depend on this change using `nix-shell -p nixpkgs-review --run "nixpkgs-review rev HEAD"`. Note: all changes have to be committed, also see [nixpkgs-review usage](https://github.com/Mic92/nixpkgs-review#usage)
- [ ] Tested basic functionality of all binary files (usually in `./result/bin/`)
- [23.11 Release Notes](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2311.section.md) (or backporting [23.05 Release notes](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2305.section.md))
  - [ ] (Package updates) Added a release notes entry if the change is major or breaking
  - [ ] (Module updates) Added a release notes entry if the change is significant
  - [ ] (Module addition) Added a release notes entry if adding a new NixOS module
- [ ] Fits [CONTRIBUTING.md](https://github.com/NixOS/nixpkgs/blob/master/CONTRIBUTING.md).

PR #322573, opened 2024-06-26
wrangler: init at 3.62.0

### LINKED ISSUE #322571: Build failure: wrangler
### Steps To Reproduce

1. run nixos flake build, with nixpkgs/master, which worked yesterday and doesn't today
2. build fails on workerd fixed derivation that is pulled from npm

### Build log

```
error: hash mismatch in fixed-output derivation '/nix/store/ycbcryc6xbvrlqg9ndfn30gvrrm9jqzj-workerd-linux-64-1.20240610.1.tgz.drv':
         specified: sha512-E8mj+HPBryKwaJAiNsYzXtVjKCL0KvUBZbtxJxlWM4mLSQhT+uwGT3nydb/hFY59rZnQgZslw0oqEWht5TEYiQ==
            got:    sha512-2zDcadR7+Gs9SjcMXmwsMji2Xs+yASGNA2cEHDuFc4NMUup+eL1mkzxc/QzvFjyBck98e92rBjMZt2dVscpGKg==
```

### Additional context

may relate to #319134 based on commit histor

### Metadata

```console
 user  nix-shell -p nix-info --run "nix-info -m"

 - system: `"x86_64-linux"`
 - host os: `Linux 6.6.35, NixOS, 24.11 (Vicuna), 24.11.20240625.e6e4cd9`
 - multi-user?: `yes`
 - sandbox: `yes`
 - version: `nix-env (Nix) 2.23.0`
 - nixpkgs: `/nix/store/2365ibi8vdmml11l7sdiclpi2hanxn5n-source`
```

I do have a bunch of experimental flags enabled, they could impact something..
```
"auto-allocate-uids"
"ca-derivations"
"cgroups"
"dynamic-derivations"
"fetch-closure"
"flakes"
"git-hashing"
"nix-command"
"parse-toml-timestamps"
"read-only-local-store"
"recursive-nix"
"verified-fetches"
```

---

dezren39:
## Description of changes

updates wrangler to be standalone package.

previously it was available but only as part of the nodePackages set.

initially i had just updated hashes because wrangler build broke.

wrangler build had other issues even after hash update,

and it was also suggested that it's preferred to use the new way to handle npm builds.

wrangler is published to npm without a package-lock,

it uses a pnpm build with workspaces as part of the cloudflare/workers-sdk monorepo.

some dependencies of wrangler are part of the monorepo and are unavailable on npm.

the pnpm helper functions such as `pnpm.fetchDeps` do not support monorepos directly.

this deployment works around that by copying a large part of the pnpm cache (hard links in node_modules into the cache) directly into the workflow. 

because pnpm dependency resolution can sometimes pull dev dependencies and other utilities,

a few dev dependencies that are also binaries are removed in the build script.

- `typescript`
- `eslint`
- `prettier`

there may be other dependencies which could conflict other nixos builds.

fixes #322571

may relate to #319134

## Things done

i confirmed this builds on my nixos framework 13 intel 11 gen,

as opposed to before when build threw.

- Built on platform(s)
  - [x] x86_64-linux
  - [x] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---


dezren39:
@SuperSandro2000 @lelgenio @a-kenji @friedow can you please review

dezren39:
this makes wrangler build as far as nix is concerned but the binary itself still has some issues

```
Error: Cannot find module 'unenv'
```

Eventually this probably should be resolved too if it happens on other machines, but I'm not entirely sure the reasons/solution for that issue.

dezren39:
@ofborg build wrangler

lelgenio (changes requested):
You could also update the hash for the other systems:
Using `nix-hash to-sri --type sha512 "$(nix-prefetch-url --type sha512 <URL>)"`, I get:

| System | Hash |
|--------|--------|
|linux-64 | `sha512-2zDcadR7+Gs9SjcMXmwsMji2Xs+yASGNA2cEHDuFc4NMUup+eL1mkzxc/QzvFjyBck98e92rBjMZt2dVscpGKg==`|
 |linux-arm64 | `sha512-7y41rPi5xmIYJN8CY+t3RHnjLL0xx/WYmaTd/j552k1qSr02eTE2o/TGyWZmGUC+lWnwdPQJla0mXbvdqgRdQg==`|
 |darwin-64 | `sha512-YanZ1iXgMGaUWlleB5cswSE6qbzyjQ8O7ENWZcPAcZZ6BfuL7q3CWi0t9iM1cv2qx92rRztsRTyjcfq099++XQ==`|
 |darwin-arm64 | `sha512-bRe/y/LKjIgp3L2EHjc+CvoCzfHhf4aFTtOBkv2zW+VToNJ4KlXridndf7LvR9urfsFRRo9r4TXCssuKaU+ypQ==`|

dezren39:
I didn't know you could do that. I'll update

pkgs/development/node-packages/overrides.nix
```diff
@@ -457,7 +457,7 @@ final: prev: {
         version = workerdVersion;
         src = fetchurl {
           url = "https://registry.npmjs.org/@cloudflare/workerd-linux-64/-/workerd-linux-64-${workerdVersion}.tgz";
-          sha512 = "sha512-E8mj+HPBryKwaJAiNsYzXtVjKCL0KvUBZbtxJxlWM4mLSQhT+uwGT3nydb/hFY59rZnQgZslw0oqEWht5TEYiQ==";
+          sha512 = "sha512-2zDcadR7+Gs9SjcMXmwsMji2Xs+yASGNA2cEHDuFc4NMUup+eL1mkzxc/QzvFjyBck98e92rBjMZt2dVscpGKg==";
```
friedow, line 460: This broke because of the way wrangler is updated. Every time you run the [generate.sh](https://github.com/NixOS/nixpkgs/blob/master/pkgs/development/node-packages/node-packages.json) it will update the [lock file](https://github.com/NixOS/nixpkgs/blob/master/pkgs/development/node-packages/node-packages.nix) which potentially breaks this package because this hash is not updated.

Therefore I'd prefer a fix where you use `buildNpmPackage` to build this as a standalone package, add yourself as a maintainer for the package, remove it from the [node-packages.json](https://github.com/NixOS/nixpkgs/blob/master/pkgs/development/node-packages/node-packages.json) and remove these overrides here. This would be a change very similar to the one I did for the [vue-language-server](https://github.com/NixOS/nixpkgs/pull/319134), so you can take inspiration from there.

friedow (changes requested):

pkgs/development/node-packages/overrides.nix
dezren39: I will take a look and see what I can do. The general idea makes sense to me, seems like the right long-term answer, but I'm not entirely prepared for reworking the whole package. I'll look at the linked pr again and maybe can work something up.

dezren39:
@ofborg build wrangler

pkgs/development/node-packages/overrides.nix
friedow: Just let me know if you are struggling at any point :)

dezren39: OK so I got a start on this, here is a pr into this pr. Once I'm done testing I'll merge into here if this is still open, otherwise I'll make a new one.

https://github.com/dezren39/nixpkgs/pull/1/files

My first roadblock is the not all dependencies are on npm. `@cloudflare/cli` isn't an npm package, it's available in the github repo in `packages/cli`.

https://github.com/cloudflare/workers-sdk/tree/main/packages/cli

```
npm ERR! code E404
npm ERR! 404 Not Found - GET https://registry.npmjs.org/@cloudflare%2fcli - Not found
npm ERR! 404
npm ERR! 404  '@cloudflare/cli@1.1.1' is not in this registry.
npm ERR! 404
npm ERR! 404 Note that you can also install from a
npm ERR! 404 tarball, folder, http url, or git url.
```

I'm not sure how best to resolve this part. I did open an issue with cloudflare/workers-sdk asking if they could add this to npm, but maybe there is some other function or method I could use to resolve this on the nixpkgs side?

https://github.com/cloudflare/workers-sdk/issues/6172

I think the function could resolve this dependency if it was listed using github in the `package.json`, but I'm not entirely clear on that.

I had to add `--legacy-peer-deps` in order to get past an issue where my react version was too new for a dependency, 
```
npm ERR! Found: react@18.3.1
npm ERR! node_modules/react
npm ERR!   dev react@"^18.2.0" from the root project
npm ERR!
npm ERR! Could not resolve dependency:
npm ERR! peer react@"^15.0.0-0 || ^16.0.0-0 || ^17.0.0-0" from @cloudflare/types@6.29.0
npm ERR! node_modules/@cloudflare/types
npm ERR!   dev @cloudflare/types@"^6.18.4" from the root project
```
I'm not sure if there are negative consequences of adding `--legacy-peer-deps` to the npm install command.

I also added `--save-dev` because all the wrangler examples included that flag. This is another one I'm unsure the consequences of. It doesn't seem to make a difference for my current issue, but it may matter once it's actually building.

friedow: If I understand the package correctly https://github.com/cloudflare/workers-sdk holds all code to build `wrangler` including https://github.com/cloudflare/workers-sdk/tree/main/packages/cli. So instead of fetching `src` for `buildNpmPackage` from npm you can fetch it from github as described in the `buildNpmPackage` example: https://github.com/NixOS/nixpkgs/blob/master/doc/languages-frameworks/javascript.section.md#buildnpmpackage-javascript-buildnpmpackage. Btw, this readme is a great read anyways.

dezren39: i'll try that! thank you

SuperSandro2000:
It seems that workerd is important to you. Please move it out of nodePackages to prevent such regressions in the future. see #229475

dezren39:
@ofborg build wrangler

dezren39:
@ofborg build wrangler

pkgs/development/node-packages/overrides.nix
dezren39: found there is no package-lock because this project uses pnpm but they also use pnpm workspaces running head first into

https://github.com/NixOS/nixpkgs/blob/master/doc/languages-frameworks/javascript.section.md#dealing-with-sourceroot-javascript-pnpm-sourceroot

dezren39:
@ofborg build wrangler

dezren39:
@ofborg build wrangler

dezren39:
issues with pnpm monorepo, looking at suggestions from #316908 which may have a path forward

seanrmurphy:
> issues with pnpm monorepo, looking at suggestions from #316908 which may have a path forward

I was looking at this a week or two ago - this is what I've come up with - this resulted in a successful `wrangler` build but when I tried to run it the paths in the script were not correct and it prob didn't know where to look for all `pnpm` dependencies - I guess they should somehow also be copied to `$out` somewhere. It's not working yet, but it might be helpful:

```
{
  lib,
  stdenv,
  fetchFromGitHub,
  pnpm_9,
  nodejs,
  npmHooks,
  writeScriptBin,
}:

  pname = "wrangler";
  version = "3.62.0";

  src = fetchFromGitHub {
    owner = "cloudflare";
    repo = "workers-sdk";
    rev = "wrangler@${finalAttrs.version}";
    hash = "sha256-/4iIkvSn85fkRggmIha2kRlW0MEwvzy0ZAmIb8+LpZQ=";
  };

  nativeBuildInputs = [
    nodejs
    pnpm_9.configHook
  ];

  pnpmDeps = pnpm_9.fetchDeps {
    inherit (finalAttrs) pname version src;
    hash = "sha256-aTTaiGXm1WYwmy+ljUC9yO3qtvN20SA+24T83dWYrI0=";
  };

  # @cloudflare/vitest-pool-workers wanted to run a server as part of the build process 
  # so I simply removed it 
  postBuild = ''
    rm -fr packages/vitest-pool-workers
    NODE_ENV="production" pnpm run build
  '';

  # this does install what gets executed, but it has a lot of dependencies on stuff which was 
  # imported so it does not work as is...
  # there seem to be a couple of issues:
  # 1. the path which is used during the build process seems to be part of the executable - sigh
  # 2. the pnpm stuff which was pulled in during the build process is not put anywhere...
  installPhase = ''
    #find . -name wrangler -print
    runHook preInstall
    mkdir -p $out/bin
    cp -r node_modules/.pnpm/node_modules/.bin/* $out/bin
  '';

})
```

maintainers/maintainer-list.nix
```diff
@@ -18037,6 +18043,12 @@
     githubId = 11613056;
     name = "Scott Dier";
   };
+  seanrmurphy = {
```
dezren39, line 18046: @seanrmurphy i added you here and as a maintainer on the wrangler derivation because you put a good bit of work into it, please let me know if you'd rather i remove this or change it.

dezren39:
> I did a little more on this and this variant seems to somehow work (in the sense that it is possible to print version, login and list dbs) but it feels very brittle and I really don't have much understanding of any model assumed by `pnpm` (ie what should go where) so I guess it can be done better. In particular, this copies an entire `node_modules` dir which is used in the build process across a monorepo which prob is adding too much to the resulting package.

getting closer, now i'm looking at this when i run the derivation as part of a larger nixos deploy

```
 > error: collision between `/nix/store/8h5s7xk9pq5i0nqv5qhs6pn4vw8mrj5f-typescript-5.5.2/lib/node_modules/typescript/package.json' and `/nix/store/pxji292jk5b1mqz45vyfhhgi9x8j723r-wrangler-3.62.0/lib/node_modules/typescript/package.json'
```

i'm guessing i pull in typescript as part of the build or something, but i'm not sure if that's a good solution. there are probably other npm deps after typescript. i'm definitely way out past my ski's here, but i'll try hacking a bit at this on my laptop for a while.

dezren39:
i hacked my way backward into a working build
the following ?devDependencies? gave me issues for some reason.
```
typescript
eslint
prettier
```
i have removed them from the node_modules in the script directly, this is pretty hacky and makes me feel bad.

I think pnpm may be pulling these in for one reason or another.

it seems to be 'fine' without them, it may be better if anyone knows either

- a way to force pnpm not to include devdependencies
  - https://github.com/pnpm/pnpm/issues/2411  
- a way to provide pnpm the dependencies that already exist as packages
  - i think there is an obvious way in nix to do this but i'm not sure

i've got no idea what i'm doing.

dezren39:
@ofborg build wrangler

pkgs/by-name/wr/wrangler/package.nix
```diff
@@ -0,0 +1,90 @@
...
+  };
+
+  nativeBuildInputs = [
+    nodejs
+    pnpm_9.configHook
+  ];
+
+  pnpmDeps = pnpm_9.fetchDeps {
+    inherit (finalAttrs) pname version src;
+    hash = "sha256-aTTaiGXm1WYwmy+ljUC9yO3qtvN20SA+24T83dWYrI0=";
```
dezren39, line 30: how do you determine these hash? i'm trying to understand if an update script is no longer needed

dezren39: well, i found that pnpm doesn't have a helper for determining this, so i did this.
```
nix-build -E "with import <nixpkgs> {}; callPackage ./package.nix {}" 2>&1 | grep "got: " | awk '{print $NF}')
```
which i'm sure will cause no issues.

maintainers/maintainer-list.nix
seanrmurphy: Fine with me - thanks.

Note that I don't think I'm in the maintainers list as yet, so when this PR is complete, it prob needs a specific commit which adds me to the maintainers list.

pkgs/by-name/wr/wrangler/package.nix
seanrmurphy: > how do you determine the pnpmDeps hash?

I use my usual approach of inserting a bogus hash and letting the tools tell me the correct one - I probably should use a more adult approach ;-)

I changed it to this (just changed the first letter to `X`):

```
    hash = "sha256-XTTaiGXm1WYwmy+ljUC9yO3qtvN20SA+24T83dWYrI0=";
```

and got this output when I did a build:

```
error: hash mismatch in fixed-output derivation '/nix/store/l1qcljcgammj59kdx0ng7y0303798g7d-wrangler-pnpm-deps.drv':
         specified: sha256-XTTaiGXm1WYwmy+ljUC9yO3qtvN20SA+24T83dWYrI0=
            got:    sha256-aTTaiGXm1WYwmy+ljUC9yO3qtvN20SA+24T83dWYrI0=
```

dezren39:
@ofborg build wrangler

dezren39:
Result of `nixpkgs-review pr 322573` run on x86_64-linux [1](https://github.com/Mic92/nixpkgs-review)
<details>
  <ul>
    <li>wrangler</li>
  </ul>
</details>

seanrmurphy:
> i hacked my way backward into a working build the following ?devDependencies? gave me issues for some reason.
> 
> ```
> typescript
> eslint
> prettier
> ```

I didn't have those packages installed; I guess that's why I didn't encounter an issue here.

> i have removed them from the node_modules in the script directly, this is pretty hacky and makes me feel bad.

It seems there are `pnpm why` and `pnpm remove` commands which could be used for helping to understand what is really required here and removing if the packages are already available but it does look like a bit of effort to dig deep into this.

> I think pnpm may be pulling these in for one reason or another.
> 
> it seems to be 'fine' without them, it may be better if anyone knows either
> 
> * a way to force pnpm not to include devdependencies
>   
>   * [`pnpm install --prod` installs devDependencies pnpm/pnpm#2411](https://github.com/pnpm/pnpm/issues/2411)
> * a way to provide pnpm the dependencies that already exist as packages
>   
>   * i think there is an obvious way in nix to do this but i'm not sure
> 
> i've got no idea what i'm doing. i feel like there should be a better way to handle the pnpm cache and node_modules, but i'm not sure. this is the first time in years i've built a javascript project, and i'd never used pnpm before this. kind of hoped pnpm.fetchDeps would do all this but i'm guessing the workspace thing is a blocker there. I also am wondering if there are some other flags that should be used in the build process.

I'm pretty much the same - the biggest challenge I think is trying to understand the model provided by `pnpm` and how it can fit with the `nix` model.

A couple of further points:
- I guess the `update.sh` script is redundant, right?
- Imo, it's prob worth experimenting a little with limiting the build by using the filters - I don't know how these work really and how they could affect the build, eg `pnpm --filter wrangler build` - I can look into this a little
- one thing I didn't really like in the build process is that it seems that a build of `wrangler` seems to depend on a build of `miniflare` which is assumed to be inside the same repo (and is not explicitly versioned) - this results in symlinks from the `wrangler` `node_modules` directory to the `miniflare` directory in the repo which is something that seems brittle

seanrmurphy:
Things are changing pretty fast here, so I'll just make a note that I had a mistake in my previous input regarding the location of `miniflare` - it needs to go here:

```
cp -r packages/miniflare $out/lib/packages
```

I also had to include the `templates` directory so now I have this block (which contains the correct `miniflare` content from above):

```
  installPhase = ''
    runHook preInstall
    mkdir -p $out/bin $out/lib $out/lib/packages/wrangler
    cp -r node_modules $out/lib
    cp -r packages/wrangler/bin $out/lib/packages/wrangler
    cp -r packages/wrangler/wrangler-dist $out/lib/packages/wrangler
    cp -r packages/wrangler/node_modules $out/lib/packages/wrangler
    cp -r packages/wrangler/templates $out/lib/packages/wrangler
    cp -r packages/miniflare $out/lib/packages
    cp $wranglerScript $out/bin/wrangler
    chmod a+x $out/bin/wrangler
    substituteInPlace $out/bin/wrangler --replace-warn /bin/sh ${pkgs.bash}/bin/sh
    substituteInPlace $out/bin/wrangler --replace-warn WRANGLER_PATH $out 
  '';
```

With this, I can log in, list dbs, list deployments, make a deployment and prob do most stuff regarding API interaction.

The outstanding issue for me is running `wrangler dev` - there is an issue with bringing in a version of `workerd-linux-64` which assumes dynamic linking...

image

Another thing which is not v cool about this approach is the amount of space consumed:

image

dezren39:
yeah that's a lot of gb. i did find the miniflare thing while trying to get it to compile but hadn't found the issue with `wrangler-dev` and `workerd`. i'm not entirely sure how to solve that, i'm guessing we would need to package it and then bring it into this package?

seanrmurphy:
> yeah that's a lot of gb. i did find the miniflare thing while trying to get it to compile but hadn't found the issue with `wrangler dev` and `workerd`. i'm not entirely sure how to solve that, i'm guessing we would need to package it and then bring it into this package?

I don't know how to do this - I did spend a few minutes looking at `workerd` but it's v complex. It's all based on Bazel which has its own ideas about dependencies and setting up build tools etc - there has been work on making nix and Bazel play nice together, but taking a complex project like `workerd` and changing its build from vanilla Bazel to some nixified Bazel looks quite non trivial and I guess there are reasons it has not been done before.

That said, having a nixified build of `workerd` is something to be aimed for.

What's unclear to me is how the wrangler build within the previous `node-packages` file does not seem to suffer from this - somehow, it must be able to run a version of `workerd` but it doesn't seem to be built by nix...

In other news, I did a simple check on the 1.3G of files in the derivation - seems there is some quite long tail of packages with the largest one being about 120M; also, there seems to be a lot of duplication of node packages with many packages being present multiple times with different versions.

seanrmurphy:
Maybe this can be reused: https://github.com/NixOS/nixpkgs/pull/238886

seanrmurphy:
Ok, so I finally have a working version. I don't think it's pretty but it seems to work - the derivation is below.

Points:
- the main point was to elf patch the `workerd` binary
- elf patching also patches quite a few other binaries that it found under the `node_modules` directory - perhaps some of these can be removed
- I had to add a few libraries to support the elf patching - I took this from #238886 but also had to add `musl` and `xorg.libX11` as these were required by some of the binaries which were found
- I also found that I needed to copy another directory from source (`workers-tsconfig`)
- I bumped to 3.63 which only required changing a couple of hashes, which was nice

Things which should be done here:
- testing - it seems to work for my context, but I definitely don't use all of the functionality of `wrangler` so perhaps it still breaks in places
- review the dependencies for `wrangler` and `miniflare` to see if they point at some other directories which should be copied across (ie ones which are not versioned but have relative paths within the source repo)
- add support for other systems - as I'm on nixos, I just focused on patching for x86_64 but this might need to be generalized; there is explicit consideration of different systems in #238886
- see if some of the content which is brought in by the `pnpm.fetchDeps` can be removed somehow - there seem to be many duplicate packages (with different versions)

We should prob get some feedback from others in the nix community regarding whether this approach (which I'm pretty sure we can make work) is too ugly/hacky etc - it's certainly not an easy general solution to `pnpm` monorepo projects.

Would love to have your feedback @dezren39 !

image

```
{
  lib,
  pkgs,
  stdenv,
  fetchFromGitHub,
  pnpm_9,
  nodejs,
  npmHooks,
  writeScriptBin,
}:

  pname = "wrangler";
  version = "3.63.0";

  src = fetchFromGitHub {
    owner = "cloudflare";
    repo = "workers-sdk";
    rev = "wrangler@${finalAttrs.version}";
    # hash = "sha256-X4iIkvSn85fkRggmIha2kRlW0MEwvzy0ZAmIb8+LpZQ=";
    hash = "sha256-lctvXOFs7/Lz8oQ4Xn+CUOnxX+qPVUrKeWDpLGXd6Zo=";
  };

  buildInputs = [ 
    pkgs.llvmPackages.libcxx 
    pkgs.llvmPackages.libunwind 
    pkgs.musl
    pkgs.xorg.libX11
  ] ++ lib.optional stdenv.isLinux pkgs.autoPatchelfHook;

  nativeBuildInputs = [
    nodejs
    pnpm_9.configHook
    pkgs.autoPatchelfHook
  ]; 

  pnpmDeps = pnpm_9.fetchDeps {
    inherit (finalAttrs) pname version src;
    #hash = "sha256-XTTaiGXm1WYwmy+ljUC9yO3qtvN20SA+24T83dWYrI0=";
    hash = "sha256-ylbJubwPUy8vNdSbssFojl6m2qf6mnFN6tpA0w2X9tU=";
  };

  preBuild = ''
    addAutoPatchelfSearchPath lib/node_modules/.pnpm/@cloudflare+workerd-linux-64@1.20240620.1/node_modules/@cloudflare/workerd-linux-64/bin/ 
  '';

  # @cloudflare/vitest-pool-workers wanted to run a server as part of the build process 
  # so I simply removed it 
  postBuild = ''
    rm -fr packages/vitest-pool-workers
    NODE_ENV="production" pnpm --filter miniflare run build
    NODE_ENV="production" pnpm --filter wrangler run build
    # pnpm --offline deploy --frozen-lockfile  --ignore-script  --filter=bash-language-server server-deploy
  '';

  # this was taken from a previous script which was generated somehow
  wranglerScript = pkgs.writeText "wrangler" ''
        #!/bin/sh
        basedir=$(dirname "$(echo "$0" | sed -e 's,\\,/,g')")

        case `uname` in
            *CYGWIN*) basedir=`cygpath -w "$basedir"`;;
        esac

        if [ -z "$NODE_PATH" ]; then
          export NODE_PATH="WRANGLER_PATH/lib/node_modules:WRANGLER_PATH/lib/packages/wrangler/node_modules"
        else
          export NODE_PATH="WRANGLER_PATH/lib/node_modules:WRANGLER_PATH/lib/packages/wrangler/node_modules:$NODE_PATH"
        fi
        if [ -x "$basedir/node" ]; then
          exec "$basedir/node"  "WRANGLER_PATH/lib/packages/wrangler/bin/wrangler.js" "$@"
        else
          exec node  "WRANGLER_PATH/lib/packages/wrangler/bin/wrangler.js" "$@"
        fi
      '';

  # I'm sure this is suboptimal but it seems to work. Points:
  # - when build is run in the original repo, no specific executable seems to be generated; you run the resulting build with pnpm run start 
  # - this means we need to add a dedicated script - perhaps it is possible to create this frok the workers-sdk dir, but I don't know how to do this
  # - the bulid process builds a version of miniflare which is used by wrangler; for this reason, the miniflare package is copied also
  # - pnpm stores all content in the top-level node_modules directory, but it is linked to from a node_modules directory inside wrangler
  # - as they are linked via symlinks, the relative location of them on the filesystem should be maintained
  installPhase = ''
    runHook preInstall
    mkdir -p $out/bin $out/lib $out/lib/packages/wrangler
    cp -r node_modules $out/lib
    cp -r packages/wrangler/bin $out/lib/packages/wrangler
    cp -r packages/wrangler/wrangler-dist $out/lib/packages/wrangler
    cp -r packages/wrangler/node_modules $out/lib/packages/wrangler
    cp -r packages/wrangler/templates $out/lib/packages/wrangler
    cp -r packages/wrangler/templates $out/lib/packages/wrangler
    cp -r packages/miniflare $out/lib/packages
    cp -r packages/workers-tsconfig $out/lib/packages
    cp $wranglerScript $out/bin/wrangler
    chmod a+x $out/bin/wrangler
    substituteInPlace $out/bin/wrangler --replace-warn /bin/sh ${pkgs.bash}/bin/sh
    substituteInPlace $out/bin/wrangler --replace-warn WRANGLER_PATH $out 
  '';

})
```

dezren39:
@seanrmurphy good find!! i just committed the changes you made and applied them to my local machine, i'm also on x86_64 nixos and don't really know how best to test for the other common architectures.

@kirillrdy @wegank tagging you in case you have opinions on our attempts at making wrangler a standalone package. part of the complexity is that this package is part of a monorepo, which pnpm fetchDeps don't directly support, and has other complex binaries like workerd to bundle in.

it would help if someone particularly knowledgeable about pnpm in nix could review, there's a ton of hard links to files using relative locations and i'm not sure how to best make this work while not wasting a ton of space.

dezren39:
@seanrmurphy i think we should see what happens with #323493 because it contains the missing functions for pnpm workspaces

seanrmurphy:
> @seanrmurphy i think we should see what happens with #323493 because it contains the missing functions for pnpm workspaces

Makes sense.

More generally, there is an issue here that `wrangler` is updated pretty much every week - I don't think `nixpkgs` can keep up with this cadence. Could make sense to put, say, a flake on flakehub which could be updated more easily/frequently if necessary.

Would be good to have some idea of how frequently this should be updated - my guess would be that, say, a monthly update would be about the right balance between keeping a modern version of `wrangler` and not loading the nixpkgs system too much.

Another point: not sure how much testing we should do before removing the old one - did you consider, say, keeping the old one (which is more or less trusted but perhaps a bit older and not easy to control individually) and adding, say `wranglerx` which is this new version, keeping it there for a few months and then removing the old one subsequently?

dezren39:
@seanrmurphy my main issue with the old one is it also was broken at least on my machine and when I started this pr it was just to update hash so that it would build at all (but still broken for me, something about unenv)

The update script looked like it technically worked, making the update process easier, but at least for a while would need to be tested after each update. If this were to merge I could commit to manually make PR updates for hashes roughly every month for some time.

pkgs/by-name/wr/wrangler/package.nix
```diff
@@ -0,0 +1,110 @@
...
+    addAutoPatchelfSearchPath lib/node_modules/.pnpm/@cloudflare+workerd-linux-64@1.20240620.1/node_modules/@cloudflare/workerd-linux-64/bin/
+    '';
+
+  # @cloudflare/vitest-pool-workers wanted to run a server as part of the build process
+  # so I simply removed it
+  postBuild = ''
+    rm -fr packages/vitest-pool-workers
+    NODE_ENV="production" pnpm --filter miniflare run build
+    NODE_ENV="production" pnpm --filter wrangler run build
+    # pnpm --offline deploy --frozen-lockfile  --ignore-script  --filter=bash-language-server server-deploy
```
pyrox0, line 54: This line doesn't need to be here, right?

pkgs/by-name/wr/wrangler/package.nix
```diff
@@ -0,0 +1,110 @@
...
+    cp -r packages/wrangler/bin $out/lib/packages/wrangler
+    cp -r packages/wrangler/wrangler-dist $out/lib/packages/wrangler
+    cp -r packages/wrangler/node_modules $out/lib/packages/wrangler
+    cp -r packages/wrangler/templates $out/lib/packages/wrangler
+    cp -r packages/wrangler/templates $out/lib/packages/wrangler
+    cp -r packages/miniflare $out/lib/packages
+    cp -r packages/workers-tsconfig $out/lib/packages
+    cp $wranglerScript $out/bin/wrangler
+    chmod a+x $out/bin/wrangler
+    substituteInPlace $out/bin/wrangler --replace-warn /bin/sh ${pkgs.bash}/bin/sh
```
pyrox0, line 98: Why not just replace it in the wrangler script generation above? Also, using `patchShebangs` would be more ideal, as this is what it's designed to do.

pkgs/top-level/all-packages.nix
```diff
@@ -7542,6 +7542,7 @@ with pkgs;
   timeline = callPackage ../applications/office/timeline { };
 
   tsm-client = callPackage ../tools/backup/tsm-client { };
+
```
pyrox0, line 7545: Please remove this.

pkgs/by-name/wr/wrangler/package.nix
dezren39: It definitely doesn't, I think it's there because that is what the normal example is for what a standard pnpm nixos build "should" look like. This should be removed before merging but also we need to figure out if what we have above this line is actually "good enough".

pkgs/by-name/wr/wrangler/package.nix
```diff
@@ -0,0 +1,109 @@
...
+  # @cloudflare/vitest-pool-workers wanted to run a server as part of the build process
+  # so I simply removed it
+  postBuild = ''
+    rm -fr packages/vitest-pool-workers
+    NODE_ENV="production" pnpm --filter miniflare run build
+    NODE_ENV="production" pnpm --filter wrangler run build
+    # pnpm --offline deploy --frozen-lockfile  --ignore-script  --filter=bash-language-server server-deploy
+    '';
+
+  # this was taken from a previous script which was generated somehow
```
doronbehar, line 57: Could you expand on that a little bit? Generated where?

pkgs/by-name/wr/wrangler/package.nix
```diff
@@ -0,0 +1,109 @@
...
+  wranglerScript = pkgs.writeText "wrangler" ''
+        #!${pkgs.bash}/bin/sh
+        basedir=$(dirname "$(echo "$0" | sed -e 's,\\,/,g')")
+
+        case `uname` in
+            *CYGWIN*) basedir=`cygpath -w "$basedir"`;;
+        esac
```
doronbehar, line 64: I'm not sure we should even try to support cygwin.

pkgs/by-name/wr/wrangler/package.nix
```diff
@@ -0,0 +1,109 @@
...
+
+        if [ -z "$NODE_PATH" ]; then
+            export NODE_PATH="WRANGLER_PATH/lib/node_modules:WRANGLER_PATH/lib/packages/wrangler/node_modules"
+        else
+            export NODE_PATH="WRANGLER_PATH/lib/node_modules:WRANGLER_PATH/lib/packages/wrangler/node_modules:$NODE_PATH"
+        fi
```
doronbehar, line 70: Is this executable eventually designed to interact with the calling environment's `$NODE_PATH`?

pkgs/by-name/wr/wrangler/package.nix
```diff
@@ -0,0 +1,109 @@
...
+        if [ -x "$basedir/node" ]; then
```
doronbehar, line 71: This will always be true in case of NixOS.

pkgs/by-name/wr/wrangler/package.nix
```diff
@@ -0,0 +1,109 @@
...
+  postBuild = ''
+    rm -fr packages/vitest-pool-workers
+    NODE_ENV="production" pnpm --filter miniflare run build
+    NODE_ENV="production" pnpm --filter wrangler run build
+    # pnpm --offline deploy --frozen-lockfile  --ignore-script  --filter=bash-language-server server-deploy
+    '';
+
+  # this was taken from a previous script which was generated somehow
+  wranglerScript = pkgs.writeText "wrangler" ''
+        #!${pkgs.bash}/bin/sh
```
doronbehar, line 59: Extra indentation.

pkgs/by-name/wr/wrangler/package.nix
```diff
@@ -0,0 +1,109 @@
...
+  # so I simply removed it
+  postBuild = ''
+    rm -fr packages/vitest-pool-workers
+    NODE_ENV="production" pnpm --filter miniflare run build
+    NODE_ENV="production" pnpm --filter wrangler run build
+    # pnpm --offline deploy --frozen-lockfile  --ignore-script  --filter=bash-language-server server-deploy
+    '';
+
+  # this was taken from a previous script which was generated somehow
+  wranglerScript = pkgs.writeText "wrangler" ''
```
doronbehar, line 58: Use `pkgs.writeShellScript`.

pkgs/by-name/wr/wrangler/package.nix
```diff
@@ -0,0 +1,109 @@
...
+    cp -r packages/wrangler/node_modules $out/lib/packages/wrangler
+    cp -r packages/wrangler/templates $out/lib/packages/wrangler
+    cp -r packages/wrangler/templates $out/lib/packages/wrangler
+    cp -r packages/miniflare $out/lib/packages
+    cp -r packages/workers-tsconfig $out/lib/packages
+    cp $wranglerScript $out/bin/wrangler
+    chmod a+x $out/bin/wrangler
+    substituteInPlace $out/bin/wrangler --replace-warn WRANGLER_PATH $out
+  '';
+  passthru.updateScript = ./update.sh;
```
doronbehar, line 100: This update script doesn't do anything that a trivial `nix-update` command wouldn't do. Please remove it to simplify the package.

pkgs/by-name/wr/wrangler/package.nix
```diff
@@ -0,0 +1,109 @@
...
+
+  meta = {
+    description = "Command-line interface for all things Cloudflare Workers";
+    homepage = "https://github.com/cloudflare/workers-sdk#readme";
+    license = "MIT OR Apache-2.0";
```
doronbehar, line 105:
```suggestion
    license = with lib.licenses; [ mit apsl20 ];
```

pkgs/by-name/wr/wrangler/package.nix
```diff
@@ -0,0 +1,109 @@
...
+  fetchFromGitHub,
+  pnpm_9,
+  nodejs,
+  npmHooks,
+  writeScriptBin
+}:
+let
+  srcHash = "sha256-/4iIkvSn85fkRggmIha2kRlW0MEwvzy0ZAmIb8+LpZQ=";
+  pnpmDepsHash = "sha256-aTTaiGXm1WYwmy+ljUC9yO3qtvN20SA+24T83dWYrI0=";
+in
```
doronbehar, line 14: There is no real reason to simply put these hashes inline where they always appear at.

pkgs/by-name/wr/wrangler/package.nix
```diff
@@ -0,0 +1,109 @@
...
+
+        if [ -z "$NODE_PATH" ]; then
+            export NODE_PATH="WRANGLER_PATH/lib/node_modules:WRANGLER_PATH/lib/packages/wrangler/node_modules"
+        else
+            export NODE_PATH="WRANGLER_PATH/lib/node_modules:WRANGLER_PATH/lib/packages/wrangler/node_modules:$NODE_PATH"
+        fi
+        if [ -x "$basedir/node" ]; then
+            exec "$basedir/node" "WRANGLER_PATH/lib/packages/wrangler/bin/wrangler.js" "$@"
+        else
+            exec node "WRANGLER_PATH/lib/packages/wrangler/bin/wrangler.js" "$@"
```
doronbehar, line 74: In general, it may very well be that this script can be generated with `makeWrapper`, see an example at the expression of `bash-language-server`'s `installPhase`.

pkgs/by-name/wr/wrangler/package.nix
```diff
@@ -0,0 +1,109 @@
...
+    cp -r packages/wrangler/bin $out/lib/packages/wrangler
+    cp -r packages/wrangler/wrangler-dist $out/lib/packages/wrangler
+    cp -r packages/wrangler/node_modules $out/lib/packages/wrangler
+    cp -r packages/wrangler/templates $out/lib/packages/wrangler
+    cp -r packages/wrangler/templates $out/lib/packages/wrangler
+    cp -r packages/miniflare $out/lib/packages
+    cp -r packages/workers-tsconfig $out/lib/packages
+    cp $wranglerScript $out/bin/wrangler
+    chmod a+x $out/bin/wrangler
+    substituteInPlace $out/bin/wrangler --replace-warn WRANGLER_PATH $out
```
doronbehar, line 98: This little hack will also won't be needed.

doronbehar (changes requested):
This PR is a mess. Please rebase it to 3 commits:

```
maintainers: add seanrmurphy
maintainers: add dezren39
wrangler: init at 3.62.0
nodePackages.wrangler: remove
```

Also make sure that after the next reviews, the PR will always have these 4 commits.

pkgs/by-name/wr/wrangler/package.nix
dezren39: ok i've removed the update script, what is the nix-update command that would update the version automatically? i've reviewed the readme and --help of nix-update and don't see an equivalent.

pkgs/by-name/wr/wrangler/package.nix
dezren39: this is useful for the update script which you said was unneeded. i'll remove. how would you recommend updating?

pkgs/by-name/wr/wrangler/package.nix
dezren39: what's the solution with this one? can we reference $out in wranglerScript?

dezren39:
@GrahamcOfBorg build wrangler

dezren39:
@ofborg build wrangler

pkgs/by-name/wr/wrangler/package.nix
seanrmurphy: I copied this from a previous version of `wrangler` somehow but now I can't find it; at the time, the point was to see if we could get it working at all, so it was a quick and dirty way to move forward.

Having looked a little more carefully, I think it could be possible to use the `wrangler.js` file directly which is put in `bin` (ie renaming it to `wrangler`)...I can try this out @dezren39 and it could make things a bit cleaner.

seanrmurphy: I see this has been completely changed so I guess there is nothing to do here - thantks @dezren39!

seanrmurphy:
Great job cleaning this up @dezren39!

We can also check if this is necessary:
```
  preBuild = ''
    addAutoPatchelfSearchPath lib/node_modules/.pnpm/@cloudflare+workerd-linux-64@1.20240620.1/node_modules/@cloudflare/workerd-linux-64/bin/
  '';
```

I added this to specify the location of the binary that needs to be patched but I subsequently found that the patching applies to all binaries found under (iirc) `lib` - we could try removing this.

dezren39:
@seanrmurphy it still builds when i removed autopatch altogether and that seems to have worked.. but should i have just removed the preBuild step? what do you think?

dezren39:
@ofborg build wrangler

seanrmurphy:
> @seanrmurphy it still builds when i removed autopatch altogether and that seems to have worked.. but should i have just removed the preBuild step? what do you think? guessing one of those subcommands might fail if we don't have the autopatch

The autopatch is necessary to run `workerd` - without the autopatch, it will not be able to find the libraries necessary to run `workerd` and the error I highlighted above should manifest when `wrangler dev` is run.

It can be built without the autopatch, because the 'build' is mostly just copying and downloading files afaict. I believe functionalities that involve querying the API should prob be fine without the autopatching but anything which is dependent on the binaries could fail.

I think the `preBuild` step can be removed - it's somehow a directive to tell autopatch where to look for files; by default it looks under `lib` but it was not clear to me if it performs a full search under this directory or simply looks one level deep. For this reason, I explicitly specified a full path. When running it, it was clear that it does a full tree search under `lib` and hence I don't think it's necessary to specify this path.

dezren39:
@seanrmurphy ill add back autopatch, makes sense to me

pkgs/by-name/wr/wrangler/package.nix
```diff
@@ -0,0 +1,90 @@
...
+    llvmPackages.libunwind
+    musl
+    xorg.libX11
+  ] ++ lib.optional stdenv.isLinux autoPatchelfHook;
+
+  nativeBuildInputs = [
+    nodejs
+    pnpm_9.configHook
+    makeWrapper
+    autoPatchelfHook
```
doronbehar, line 35: It'd be nice to put a comment explaining this. IIRC there was a comment here in one of the revisions of the PR.

P.S The issue with that dependency which requires `patchelf` is also causing the build to fail on platforms other then `x86_64-linux` - which warrants a `meta.platforms` line. Please mention that in the comment here, or near the `meta.platformts` line.

doronbehar (changes requested):
Looks almost perfect! Well done :).

dezren39:
@ofborg build wrangler

dezren39:
@ofborg build wrangler

pkgs/by-name/wr/wrangler/package.nix
```diff
@@ -0,0 +1,107 @@
...
+  lib,
+  stdenv,
+  fetchFromGitHub,
+  makeWrapper,
+  nodejs,
+  pnpm_9,
+  # cpp is required for building workerd.
+  # workerd is used for some wrangler subcommands.
+  # non-linux platforms may experience errors
+  # on certain subcommands due to this issue.
```
doronbehar, line 11: Please write comments only once :). Near the buildInputs is a better place IMO.

pkgs/by-name/wr/wrangler/package.nix
```diff
@@ -0,0 +1,107 @@
...
+    description = "Command-line interface for all things Cloudflare Workers";
+    homepage = "https://github.com/cloudflare/workers-sdk#readme";
+    license = with lib.licenses; [
+      mit
+      apsl20
+    ];
+    maintainers = with lib.maintainers; [
+      seanrmurphy
+      dezren39
+    ];
```
doronbehar, line 104: Not setting `platforms`?

pkgs/by-name/wr/wrangler/package.nix
```diff
@@ -0,0 +1,107 @@
...
+    # non-linux platforms may experience errors
+    # on certain subcommands due to this issue.
+    lib.optional stdenv.isLinux
+    [
+      autoPatchelfHook
+      llvmPackages.libcxx
+      llvmPackages.libunwind
+      musl
+      xorg.libX11
+    ]
```
doronbehar, line 40: Missing `;`.

pkgs/by-name/wr/wrangler/package.nix
```diff
@@ -0,0 +1,107 @@
...
+    rev = "wrangler@${finalAttrs.version}";
+    hash = "sha256-/4iIkvSn85fkRggmIha2kRlW0MEwvzy0ZAmIb8+LpZQ=";
+  };
+
+  buildInputs =
+    # cpp is required for building workerd.
+    # workerd is used for some wrangler subcommands.
+    # non-linux platforms may experience errors
+    # on certain subcommands due to this issue.
+    lib.optional stdenv.isLinux
```
doronbehar, line 33: I think it'd be more correct to add these inputs unconditionally and use the `meta.platforms` attribute.

pkgs/by-name/wr/wrangler/package.nix
```diff
@@ -0,0 +1,107 @@
...
+    [
+      autoPatchelfHook
```
doronbehar, line 35: Doesn't belong here. Belongs only to `nativeBuildInputs`.

pkgs/by-name/wr/wrangler/package.nix
```diff
@@ -0,0 +1,107 @@
...
+      makeWrapper
+      nodejs
+      pnpm_9.configHook
+    ]
+    ++
+    # cpp is required for building workerd.
+    # workerd is used for some wrangler subcommands.
+    # non-linux platforms may experience errors
+    # on certain subcommands due to this issue.
+    lib.optional stdenv.isLinux autoPatchelfHook;
```
doronbehar, line 53: Same here regarding the conditional - not needed if `meta.platforms is set correctly.

pkgs/by-name/wr/wrangler/package.nix
```diff
@@ -0,0 +1,107 @@
...
+
+  pnpmDeps = pnpm_9.fetchDeps {
+    inherit (finalAttrs) pname version src;
+    hash = "sha256-aTTaiGXm1WYwmy+ljUC9yO3qtvN20SA+24T83dWYrI0=";
+  };
```
doronbehar, line 58: I think it'd be better to put this near the `src`, as the hashes will be near each other and it will be a bit harder to miss if in a future update this hash won't be updated. Also, it makes the diffs a bit smaller this way when the hashes are near each other.

doronbehar (changes requested):

dezren39:
@ofborg build wrangler

pkgs/by-name/wr/wrangler/package.nix
dezren39: on it

doronbehar:
Also please don't push so frequently - you are exhausting CI. Push only after you have tested it builds locally and after you have gone through all the unresolved review comments at least twice and made sure they are all addressed.

pkgs/by-name/wr/wrangler/package.nix
dezren39: ok i can do that, i just didn't want to make wrangler limited on mac, afaik they normally can download and use wrangler just fine, but the normal install is just pulling the binary from npm or something. i figured it would be better to at least let them use the commands that do work but i can see either way.

doronbehar: > ok i can do that, i just didn't want to make wrangler limited on mac, afaik they normally can download and use wrangler just fine, but the normal install is just pulling the binary from npm or something. i figured it would be better to at least let them use the commands that do work but i can see either way.

Yes but it would be too hard to maintain that, especially if you don't have a Darwin machine to test it locally. When a darwin user will want to use this package, they'll make the effort don't worry.

pkgs/by-name/wr/wrangler/package.nix
doronbehar: > Please write comments only once :). Near the buildInputs is a better place IMO.

Unresolved.

pkgs/by-name/wr/wrangler/package.nix
doronbehar: > Doesn't belong here. Belongs only to `nativeBuildInputs`.

Unresolved.

pkgs/by-name/wr/wrangler/package.nix
doronbehar: > Not setting `platforms`?

Unresolved.

pkgs/by-name/wr/wrangler/package.nix
doronbehar: > Same here regarding the conditional - not needed if `meta.platforms is set correctly.

Unresolved.

doronbehar:
> Also please don't push so frequently - you are exhausting CI. Push only after you have tested it builds locally and after you have gone through all the unresolved review comments at least twice and made sure they are all addressed.

@dezren39 you are not behaving as I asked you. Not only that, you are marking conversations as resolved without really resolving them. I won't be helping you to get this in if you won't behave.

pkgs/by-name/wr/wrangler/package.nix
```diff
@@ -0,0 +1,96 @@
...
+    # cpp is required for building workerd.
+    # workerd is used for some wrangler subcommands.
+    # non-linux platforms may experience errors
+    # on certain subcommands due to this issue.
+    [
+      llvmPackages.libcxx
+      llvmPackages.libunwind
+      musl
+      xorg.libX11
+    ];
```
doronbehar, line 39: Please format this again with `nixfmt-rfc-style`.

pkgs/by-name/wr/wrangler/package.nix
```diff
@@ -0,0 +1,96 @@
...
+    license = with lib.licenses; [
+      mit
+      apsl20
+    ];
+    maintainers = with lib.maintainers; [
+      seanrmurphy
+      dezren39
+    ];
+    mainProgram = "wrangler";
+    platforms = lib.platforms.linux;
```
doronbehar, line 94: Incorrect, this includes `aarch64-linux`. Please use:

```suggestion
    platforms = [ "x86_64-linux" ];
```

doronbehar (changes requested):

pkgs/by-name/wr/wrangler/package.nix
dezren39: i already used nixfmt. i believe it's the rfc-style already.
just to be sure, i used the instructions from nixfmt github
`nix-env -f https://github.com/NixOS/nixfmt/archive/master.tar.gz -i`
and installed an entirely different copy. it had the same results.

pkgs/by-name/wr/wrangler/package.nix
```diff
@@ -0,0 +1,96 @@
...
+  pnpmDeps = pnpm_9.fetchDeps {
+    inherit (finalAttrs) pname version src;
+    hash = "sha256-aTTaiGXm1WYwmy+ljUC9yO3qtvN20SA+24T83dWYrI0=";
+  };
+
+  buildInputs =
+    # cpp is required for building workerd.
+    # workerd is used for some wrangler subcommands.
+    # non-linux platforms may experience errors
+    # on certain subcommands due to this issue.
```
doronbehar, line 33: On a 2nd thought, perhaps this comment would fit better above the `platforms` entry.

pkgs/by-name/wr/wrangler/package.nix
doronbehar: Hmm I'm pretty sure that after after you'll move that comment nixfmt will format it. The best way to run it though is to use `direnv` (with a recent enough checkout that includes a `.envrc` file).

DataHearth:
Looks awesome 👏! Hopes this can be shipped soon 👌👌

doronbehar:
Result of `nixpkgs-review pr 322573` run on x86_64-linux [1](https://github.com/Mic92/nixpkgs-review)
<details>
  <ul>
    <li>wrangler</li>
  </ul>
</details>

doronbehar (approved):

PR #321559, opened 2024-06-21
python311Packages.jaxlib: fix bazel hash

GaetanLepage:
## Description of changes

Should fix https://github.com/NixOS/nixpkgs/issues/296737...

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [x] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
- For non-Linux: Is sandboxing enabled in `nix.conf`? (See [Nix manual](https://nixos.org/manual/nix/stable/command-ref/conf-file.html))
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:
  - [NixOS test(s)](https://nixos.org/manual/nixos/unstable/index.html#sec-nixos-tests) (look inside [nixos/tests](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests))
  - and/or [package tests](https://github.com/NixOS/nixpkgs/blob/master/pkgs/README.md#package-tests)
  - or, for functions and "core" functionality, tests in [lib/tests](https://github.com/NixOS/nixpkgs/blob/master/lib/tests) or [pkgs/test](https://github.com/NixOS/nixpkgs/blob/master/pkgs/test)
  - made sure NixOS tests are [linked](https://nixos.org/manual/nixpkgs/unstable/#ssec-nixos-tests-linking) to the relevant packages
- [x] Tested compilation of all packages that depend on this change using `nix-shell -p nixpkgs-review --run "nixpkgs-review rev HEAD"`. Note: all changes have to be committed, also see [nixpkgs-review usage](https://github.com/Mic92/nixpkgs-review#usage)
- [ ] Tested basic functionality of all binary files (usually in `./result/bin/`)
- [24.11 Release Notes](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2411.section.md) (or backporting [23.11](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2311.section.md) and [24.05](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2405.section.md) Release notes)
  - [ ] (Package updates) Added a release notes entry if the change is major or breaking
  - [ ] (Module updates) Added a release notes entry if the change is significant
  - [ ] (Module addition) Added a release notes entry if adding a new NixOS module
[...]

PR #322625, opened 2024-06-26
python3Packages.haystack-ai: init at 2.2.3, remove farm-haystack

edwtjo:
## Description of changes

Introduces `haystack-ai` which replaces the current package `farm-haystack`

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [x] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/development/python-modules/haystack-ai/default.nix
+++ b/pkgs/development/python-modules/haystack-ai/default.nix
@@ -90,7 +90,7 @@
 }:
 
 buildPythonPackage rec {
-  pname = "farm-haystack";
+  pname = "haystack-ai";
   version = "2.2.3";
   pyproject = true;
 
--- a/pkgs/top-level/python-packages.nix
+++ b/pkgs/top-level/python-packages.nix
@@ -4151,8 +4151,6 @@ self: super: with self; {
 
   farama-notifications = callPackage ../development/python-modules/farama-notifications { };
 
-  farm-haystack = callPackage ../development/python-modules/farm-haystack { };
-
   fastai = callPackage ../development/python-modules/fastai { };
 
   fastapi = callPackage ../development/python-modules/fastapi { };
@@ -5407,6 +5405,8 @@ self: super: with self; {
   haversine = callPackage ../development/python-modules/haversine { };
 
   hawkauthlib = callPackage ../development/python-modules/hawkauthlib { };
+
+  haystack-ai = callPackage ../development/python-modules/haystack-ai { };
 
   hcloud = callPackage ../development/python-modules/hcloud { };
 
--- a/pkgs/top-level/python-aliases.nix
+++ b/pkgs/top-level/python-aliases.nix
@@ -163,6 +163,7 @@ mapAliases ({
   face_recognition_models = face-recognition-models; # added 2022-10-15
   factory_boy = factory-boy; # added 2023-10-08
   fake_factory = throw "fake_factory has been removed because it is unused and deprecated by upstream since 2016."; # added 2022-05-30
+  farm-haystack = throw "farm-haystack was removed in favor of haystack-ai"; # added 2024-07-03
   faster-fifo = throw "faster-fifo has been removed since it was an unused leaf package"; # added 2024-07-02
   fastnlo_toolkit = fastnlo-toolkit; # added 2024-01-03
   fastpair = throw "fastpair is unmaintained upstream and has therefore been removed"; # added 2024-05-01


jopejoe1:
Regarding the jaxlib hash update please see https://github.com/NixOS/nixpkgs/pull/321559 and the discussion that happens after it was merged.

pkgs/development/python-modules/jaxlib/default.nix
```diff
@@ -374,10 +374,10 @@ let
       sha256 =
         (
           if cudaSupport then
-            { x86_64-linux = "sha256-vUoAPkYKEnHkV4fw6BI0mCeuP2e8BMCJnVuZMm9LwSA="; }
+            { x86_64-linux = "sha256-vAF5JwpINIp4pn1tFFl5059m/8/hn2cmEia04h9hHAw="; }
```
JohnRTitor, line 377: Hash update without a version change?

pkgs/top-level/python-packages.nix
```diff
@@ -5384,6 +5382,8 @@ self: super: with self; {
 
   hawkauthlib = callPackage ../development/python-modules/hawkauthlib { };
 
+  haystack-ai = callPackage ../development/python-modules/haystack-ai { };
```
JohnRTitor, line 5385: Perhaps `farm-haystack` can be aliased to `haystack-ai`?

pkgs/development/python-modules/haystack-ai/default.nix
```diff
@@ -90,18 +96,25 @@
   weaviate-client,
 }:
 
+let
+  inherit (lib) optionals;
```
JohnRTitor, line 100: Umm, let's use `lib.optionals` explicitly, there is like three usage anyway

JohnRTitor (commented):
Please add release notes since this is a breaking change.
Ideally the old package could just be aliased the new one.

Please update the PR title accordingly

pkgs/development/python-modules/haystack-ai/default.nix
```diff
@@ -1,11 +1,15 @@
-{
-  lib,
-  buildPythonPackage,
-  fetchFromGitHub,
-  pythonRelaxDepsHook,
+{ lib, buildPythonPackage, fetchPypi, pythonRelaxDepsHook,
```
JohnRTitor, line 1: Please format according to `nixfmt-rfc-style`

pkgs/top-level/python-packages.nix
JohnRTitor:
```suggestion
  haystack-ai = callPackage ../development/python-modules/haystack-ai {
    pytorch = python3Packages.torch;
  };
```

Eval fix

edwtjo:
> Please add release notes since this is a breaking change. Ideally the old package could just be aliased the new one.
> 
> Please update the PR title accordingly

The PR title is correct, the jaxlib update was cruft which I forgot to remove.

edwtjo:
> The jaxlib hash seems to be unreproducible please see #321559 for more information on it.

Removed the jaxlib update.

JohnRTitor (dismissed):
Eval is failing, there is no `src.rev` in the new derivation, additionally new packages should be in `pkgs/by-name`

EDIT: Oh, Python packages should be in `python-packages`, that's fine then

edwtjo:
> Eval is failing, there is no `src.rev` in the new derivation, additionally new packages should be in `pkgs/by-name`

Yes, already noticed it and addressed it

happysalada:
Why did you decide de fetch from pypi against fetching the source from github ?
Just out of curiosity, the changes look good.

nixos/doc/manual/release-notes/rl-2411.section.md
```diff
@@ -25,6 +25,8 @@
 
 ## Backward Incompatibilities {#sec-release-24.11-incompatibilities}
 
+- `farm-haystack` has been deprecated since it will no longer receive updates and users are advised to upgrade to `haystack-ai`.
```
natsukium, line 28: We do not usually list the removal of packages, at least python packages, in the release notes.
Please add the alias to `pkgs/top-level/python-aliases.nix` instead.

pkgs/development/python-modules/haystack-ai/default.nix
```diff
@@ -1,11 +1,19 @@
...
   lib,
   buildPythonPackage,
-  fetchFromGitHub,
+  fetchPypi,
   pythonRelaxDepsHook,
+
+  # Feature flags
+  withJax ? false,
+  withTorch ? true,
+  withTensorflow ? false,
```
natsukium, line 10: The python module does not allow such flags.
This is because it causes unnecessary rebuilds and is uncontrollable by downstream packages.
This package does not need to provide anything, and users should add any backend they want to their environment.

natsukium (dismissed):

natsukium:
This is a python module, so please change your commit message and PR title to `python311Packages.haystack-ai:` from `haystack-ai:`

nixos/doc/manual/release-notes/rl-2411.section.md
edwtjo: This shouldn't be aliased since the package name changes; it is in effect a new package even though it provides similar functionality. @JohnRTitor and @natsukium please reach consensus whether this should have release notes.

pkgs/development/python-modules/haystack-ai/default.nix
edwtjo: What do you mean, there are other python modules using `withFlag` or `enableFlag` in the tree. Do you believe it is preferable to do `overrideAttrs` to manipulate `propagatedBuildInputs` in your python environment, instead of regular `override`. To me that seems like more work, or is this now some policy in nixpkgs to which I'm unaware?

edwtjo:
> Why did you decide de fetch from pypi against fetching the source from github ? Just out of curiosity, the changes look good.

I've had a "rule" in the back of my mind that python packages should use pypi instead of source repos when they're packaged, but I cannot find any such recommendation in nixpkgs or where I got that into my mind so :shrug: I can change back to github if you want of course

nixos/doc/manual/release-notes/rl-2411.section.md
JohnRTitor: I am fine with aliasing this, if the change is mostly backwards-compatible, else we should mention it in the release notes.

natsukium: Once again, the python module should not be listed in the nixos release notes, but in `pkgs/top-level/python-aliases.nix`. This is a leaf package, so there is no need to add it to the release notes.
I may have used the word "alias" incorrectly, but just throw an exception with the reason.

pkgs/development/python-modules/haystack-ai/default.nix
natsukium: > there are other python modules using withFlag or enableFlag in the tree.

They are used as applications or are old or have not received proper review.

> Do you believe it is preferable to do overrideAttrs to manipulate propagatedBuildInputs in your python environment, instead of regular override.

We have a function called `overridePythonAttrs` that is used to properly handle dependencies.
https://github.com/NixOS/nixpkgs/blob/master/doc/languages-frameworks/python.section.md#overriding-python-packages-overriding-python-packages

For this package, you don't even need overridePythonAttrs, just add the necessary backends to your python environment.

natsukium:
Result of `nixpkgs-review pr 322625` at 368f4c41 run on x86_64-linux [1](https://github.com/Mic92/nixpkgs-review)
<details>
  <summary>2 packages marked as broken and skipped:</summary>
  <ul>
    <li>python311Packages.haystack-ai</li>
    <li>python312Packages.haystack-ai</li>
  </ul>
</details>

natsukium (approved):
LGTM, thanks.

PR #322641, opened 2024-06-26
xcode-wrapper: allow xcodebuild to fail in version check

mroi:
With the current beta of Xcode 16.0, the `xcode-wrapper` derivation fails with error code 141. No additional output is generated. I traced this back to `xcodebuild -version` failing when its output is piped to `head`. I think this is because head closes xcodebuild’s stdout after the first line, which xcodebuild does not like.

## Description of changes

Fixed by allowing xcodebuild to fail within the pipe used for version number extraction.

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [x] x86_64-darwin
  - [ ] aarch64-darwin
  - [x] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/development/mobile/xcodeenv/compose-xcodewrapper.nix
+++ b/pkgs/development/mobile/xcodeenv/compose-xcodewrapper.nix
@@ -29,7 +29,7 @@ stdenv.mkDerivation {
     ln -s "${xcodeBaseDir}/Contents/Developer/Platforms/iPhoneSimulator.platform/Developer/SDKs"
 
     # Check if we have the xcodebuild version that we want
-    currVer=$($out/bin/xcodebuild -version | head -n1)
+    currVer=$($out/bin/xcodebuild -version | head -n1 || true)
     ${if allowHigher then ''
     if [ -z "$(printf '%s\n' "${version}" "$currVer" | sort -V | head -n1)""" != "${version}" ]
     '' else ''


jakubgs (changes requested):
No, why would you do that?

mroi:
With the Xcode 16 beta, the Nixpkgs `xcode-wrapper` fails to build.

`xcodebuild` does indeed return a nonzero exit code when output is piped to `head -n1`. I guess it does not handle the fact that its stdout is closed when it wants to output its second line. We could try and get Apple to fix this, but I thought this can be an acceptable fix as we don’t really care about the `xcodebuild` exit code as long as the version number it outputs is the expected one.

jakubgs:
As always apple coming out with the most mind boggling behaviors that make no sense.

I'd rather not just ignore Xcode errors. Could we not just pipe it through `cat` first so it reads all lines?
```sh
    currVer=$($out/bin/xcodebuild -version | cat | head -n1)
```
Based on your description that should not fail right?

mroi:
I investigated some more and the problem is not specific to `xcodebuild` or macOS at all. [It also happens on Linux with other tools.](https://stackoverflow.com/questions/19120263/)

tl;dr: When piping output to a tool that may exit prematurely, without consuming all its input, the process on the left of the pipe receives a `SIGPIPE` and the standard reaction leads to exit code 141.

So piping to `head -n1` within a script with pipefail enabled is brittle and may break as it does here. This should work reliably:

```
versionInfo=$($out/bin/xcodebuild -version)
currVer="${versionInfo%%$'\n'*}"  # Xcode version is in the first line
```

Do you think this is better than masking the exit code with `|| true`? It should be more robust, as it checks the exit code, but conveys the intention a little less clearly.

mroi:
This also works, maybe the best compromise? Do you have a preference?

```
currVer=$(head -n1 <(xcodebuild -version))
```

jakubgs:
But I don't see `pipefail` enabled in that `buildCommand`. Where are you getting `pipefail` from there?

Also, I see no such behavior from Xcode 15.1.

mroi:
> But I don't see `pipefail` enabled in that `buildCommand`. Where are you getting `pipefail` from there?

I think it’s [default with the `stdenv` builder.](https://github.com/NixOS/nixpkgs/blob/3e6461ff2de9badd8eb67a2e07b5747585544efd/pkgs/stdenv/generic/setup.sh#L5)

> Also, I see no such behavior from Xcode 15.1.

Me neither. I guess this is either consistent luck (the situation is inherently racy), or between the versions things got split in `xcodebuild` from one `printf` for both lines to one per line or something like that.

jakubgs:
>I think it’s [default with the stdenv builder.](https://github.com/NixOS/nixpkgs/blob/3e6461ff2de9badd8eb67a2e07b5747585544efd/pkgs/stdenv/generic/setup.sh#L5)

Oh, interesting, that's good to know.

Well, I tested it with pipefail and can't reproduce it:
```
 > cat test.sh 
#!/usr/bin/env bash
set -ee
set -o pipefail
xcodebuild -version | head -n1

 > ./test.sh  
Xcode 15.1

 > echo $?
0
```
Can you with the newer version?

mroi:
Thanks for the script. With the Xcode beta, I can even demonstrate it’s racy:
```
Michael@carpo:~/Desktop > cat test.sh 
#!/usr/bin/env bash
set -ee
set -o pipefail
xcodebuild -version | head -n1

Michael@carpo:~/Desktop > ./test.sh 
Xcode 16.0

Michael@carpo:~/Desktop > echo $?
0

Michael@carpo:~/Desktop > ./test.sh 
Xcode 16.0

Michael@carpo:~/Desktop > echo $?
141
```

siddarthkay:
Indeed this seems to be a problem with Xcode 16 beta.
```
./test.sh
Xcode 16.0
echo $?
141
./test.sh
Xcode 16.0
echo $?
141
```

after switching locally to stable Xcode 15.4

```
./test.sh
Xcode 15.4
echo $?
0
./test.sh
Xcode 15.4
echo $?
0
./test.sh
Xcode 15.4
echo $?
0
```

mroi:
Current Nixpkgs no longer contains the incriminated line.
This was fixed by 15d5ce3687725ed41921289a1ab8fdbe1a69e5e4.

PR #322642, opened 2024-06-26
wayneko: init at 0-unstable-2024-03-29

fgaz:
## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/by-name/wa/wayneko/package.nix
+++ b/pkgs/by-name/wa/wayneko/package.nix
@@ -1,0 +1,47 @@
+{
+  lib,
+  stdenv,
+  fetchFromSourcehut,
+  pixman,
+  pkg-config,
+  wayland,
+  wayland-scanner,
+}:
+
+stdenv.mkDerivation {
+  pname = "wayneko";
+  version = "0-unstable-2024-03-29";
+
+  src = fetchFromSourcehut {
+    owner = "~leon_plickat";
+    repo = "wayneko";
+    rev = "c1919dc3a7e610d30e4c06efaa5af85941f27d86";
+    hash = "sha256-2cbEcDK6WZPe4HvY1pxmZVyDAj617VP1l0Gn7uSlNaE=";
+  };
+
+  postPatch = ''
+    substituteInPlace Makefile \
+      --replace-fail -Werror ""
+  '';
+
+  nativeBuildInputs = [
+    pkg-config
+    wayland-scanner
+  ];
+
+  buildInputs = [
+    pixman
+    wayland
+  ];
+
+  makeFlags = [ "PREFIX=${placeholder "out"}" ];
+
+  meta = {
+    description = "Neko on Wayland";
+    homepage = "https://sr.ht/~leon_plickat/wayneko";
+    license = lib.licenses.gpl3Only;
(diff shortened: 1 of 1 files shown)


donovanglover (approved):
- [x] Makefile installs man pages and shell completions
- [x] License seems correct
- [x] Nix code looks clean
- [x] Checked `src.rev` and `version`
- [x] Checked [repology](https://repology.org/project/wayneko/badges) for other distributions packaging wayneko
- [x] Checked `meta.description` formatting
- [x] Tested and works with Hyprland

LGTM

pkgs/by-name/wa/wayneko/package.nix
```diff
@@ -0,0 +1,46 @@
...
+  src = fetchFromSourcehut {
+    owner = "~leon_plickat";
+    repo = "wayneko";
+    rev = "c1919dc3a7e610d30e4c06efaa5af85941f27d86";
+    hash = "sha256-2cbEcDK6WZPe4HvY1pxmZVyDAj617VP1l0Gn7uSlNaE=";
+  };
+
+  postPatch = ''
+    substituteInPlace Makefile \
+      --replace-warn -Werror ""
```
SuperSandro2000, line 23:
```suggestion
      --replace-fail -Werror ""
```
we probably want to notice if that moves somewhere else or is entirely removed

pinpox:
Result of `nixpkgs-review pr 322642` run on x86_64-linux [1](https://github.com/Mic92/nixpkgs-review)
<details>
  <ul>
    <li>wayneko</li>
  </ul>
</details>

pkgs/by-name/wa/wayneko/package.nix
fgaz: Is there some guidance about the `--replace` variants somewhere? My reasoning for `warn` here is that even if this replace fails it isn't going to affect the package in any way. At most it's going to fail to build. Using `fail` here can prevent automatic updates for no actual benefit.

On the other hand, if I was for example replacing a hardcoded path of a runtime dependency I would use `--replace-fail`, because a failed replace would silently produce a broken package.

fgaz: Ping

fgaz: Ping

Aleksanaa: IMO `--replace-fail` should be used in all situations when the code to be patched is definite

fgaz: What do you mean by definite/indefinite? Aren't all patches "definite"?

Aleksanaa: I mean "we know exactly what we are patching". If we are patching a set of unspecified files with common characteristics, it would make sense to use warn or even success. But if the patch fails for a single, specific file after it is updated, the least bad result is we get a dead code. But it is likely that we already need additional processing. Here, failure makes it easier for us to see the problem in time.

PR #322653, opened 2024-06-26
dosbox-staging: refactor

AndersonTorres:
- reorder lists ASCIIbetically
- split outputs
- add tests and updateScript

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/by-name/do/dosbox-staging/package.nix
+++ b/pkgs/by-name/do/dosbox-staging/package.nix
@@ -1,14 +1,14 @@
 {
   lib,
-  stdenv,
-  fetchFromGitHub,
-  fetchpatch,
   SDL2,
   SDL2_image,
   SDL2_net,
   alsa-lib,
   darwin,
+  fetchFromGitHub,
+  fetchpatch,
   fluidsynth,
+  gitUpdater,
   glib,
   gtest,
   iir1,
@@ -18,7 +18,6 @@
   libmt32emu,
   libogg,
   libpng,
-  zlib-ng,
   libpulseaudio,
   libslirp,
   libsndfile,
@@ -28,7 +27,9 @@
   opusfile,
   pkg-config,
   speexdsp,
-  nix-update-script,
+  stdenv,
+  testers,
+  zlib-ng,
 }:
 
 stdenv.mkDerivation (finalAttrs: {
@@ -60,6 +61,9 @@ stdenv.mkDerivation (finalAttrs: {
 
   buildInputs =
     [
+      SDL2
+      SDL2_image
+      SDL2_net
       fluidsynth
       glib
       iir1
@@ -69,15 +73,12 @@ stdenv.mkDerivation (finalAttrs: {
       libmt32emu
       libogg
       libpng
-      zlib-ng
       libpulseaudio
       libslirp
       libsndfile
       opusfile
-      SDL2
-      SDL2_image
-      SDL2_net
       speexdsp
+      zlib-ng
     ]
     ++ lib.optionals stdenv.isLinux [ alsa-lib ]
     ++ lib.optionals stdenv.isDarwin (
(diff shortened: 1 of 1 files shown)


AndersonTorres:
@ofborg build pkgsMusl.dosbox-staging

pkgs/by-name/do/dosbox-staging/package.nix
```diff
@@ -89,29 +90,40 @@ stdenv.mkDerivation (finalAttrs: {
...
+  outputs = [ "out" "man" ];
+
   postInstall = ''
     install -Dm644 $src/contrib/linux/dosbox-staging.desktop $out/share/applications/
   '';
 
+  # Rename binary, add a wrapper, and copy manual to avoid conflict with
+  # original dosbox. Doing it this way allows us to work with frontends and
+  # launchers that expect the binary to be named dosbox, but get out of the way
+  # of vanilla dosbox if the user desires to install that as well.
```
superherointj, line 102:
```suggestion
  # Rename binary, add a wrapper, and copy manual to avoid conflict with
  # original dosbox. Doing it this way allows us to work with frontends and
  # launchers that expect the binary to be named dosbox, but get out of the way
  # of vanilla dosbox if the user desires to install that as well.
```

I've received reverse feedback on this from some other commiter. Maybe @SuperSandro2000 ? (It's hard to remember the source, I remember the issue.)

Like, the comment should be inside the phase because it's a bash comment.

pkgs/by-name/do/dosbox-staging/package.nix
```diff
@@ -1,14 +1,14 @@
 {
   lib,
```
superherointj, line 2: Why `lib` is breaking `ASCIIbetical` order?

pkgs/by-name/do/dosbox-staging/package.nix
```diff
@@ -89,29 +90,40 @@ stdenv.mkDerivation (finalAttrs: {
...
-    # Rename binary, add a wrapper, and copy manual to avoid conflict with
-    # original dosbox. Doing it this way allows us to work with frontends and
-    # launchers that expect the binary to be named dosbox, but get out of the
-    # way of vanilla dosbox if the user desires to install that as well.
     mv $out/bin/dosbox $out/bin/dosbox-staging
     makeWrapper $out/bin/dosbox-staging $out/bin/dosbox
 
-    # Create a symlink to dosbox manual instead of copying it
-    pushd $out/share/man/man1/
+    pushd $man/share/man/man1/
```
superherointj, line 107: Thanks for doing this. Found the `$man` output at containers. So much wasted space.

pkgs/by-name/do/dosbox-staging/package.nix
```diff
@@ -89,29 +90,40 @@ stdenv.mkDerivation (finalAttrs: {
       ]
     );
 
+  outputs = [ "out" "man" ];
```
superherointj, line 93: Phases order wise, I don't know where to put `outputs`. @SuperSandro2000

pkgs/by-name/do/dosbox-staging/package.nix
AndersonTorres: Historical and aesthetic reasons.

`lib` was (almost) always first since it was split from stdenv.
And since it is almost omnipresent an rarely subjected to overriding, it makes some sense to put it first.

pkgs/by-name/do/dosbox-staging/package.nix
AndersonTorres: I am more or less following nixpkgs-hammering. Complex attributes (in the sense of "composed by other attributes" - lists, sets &c.) come first.

maybe right after src?

pkgs/by-name/do/dosbox-staging/package.nix
superherointj: My bias is for alphabetical ordering, but I can understand the need of `ASCIIbetical`. Since it is the alphabetical ordering that computers are usually doing.

A new person, with no historical context, can understand alphabetical ordering. But having to explain some historical context for the person getting `lib` right (whatever that is) is extra work. I'd rather not have to.

pkgs/by-name/do/dosbox-staging/package.nix
AndersonTorres: I don't like Bash comments.

Roughly speaking, Nix is the high level lang, while Bash is assembly.
Bash comments should be restricted to Bash specificities.
Since this comment is about a decision about installation, it should be on a higher level.

Also, since this is a string that will be passed to the "assembler", Nix will rebuild this if the comment is modified. However, the inferior Bash process will happily ignore the comment.

And we end with a strange situation where the superior process requires an unneeded rebuild to the inferior one!

pkgs/by-name/do/dosbox-staging/package.nix
AndersonTorres: `lib` is mostly off anyway. It is, roughly speaking just a collection of "macros".
It is not used as a `buildInput` or anything that affects the build.

This is mostly aesthetic at this point.

superherointj (approved):

AndersonTorres:
I am impressed that pkgsMusl.dosbox-staging built!

PR #322733, opened 2024-06-26
lib: Document stability guarantee

infinisil:
## Description of changes

Could be better, but this is a good start :)

Prompted by @Janik-Haag, ping @roberth for review

---

### DIFF
--- a/lib/README.md
+++ b/lib/README.md
@@ -57,6 +57,17 @@ Don't have multiple changes in one PR, instead split it up into multiple ones.
 
 This keeps the conversation focused and has a higher chance of getting merged.
 
+### Ensure backwards compatibility
+
+`lib` is the effective standard library for Nix, both for Nixpkgs and third-parties.
+Because of that it's important to make sure that the interface isn't changed in a breaking way.
+
+The only exceptions are:
+- Breaking changes to interfaces that aren't in a release yet
+- Breaking changes that are unlikely to cause problems for users
+- Removal of functions that have been deprecated for at least one release
+- Breaking changes to functions/sub-libraries that are explicitly marked as experimental
+
 ### Name the interface appropriately
 
 When introducing new names to the interface, such as new function, or new function attributes,


lib/README.md
```diff
@@ -57,6 +57,11 @@ Don't have multiple changes in one PR, instead split it up into multiple ones.
 
 This keeps the conversation focused and has a higher chance of getting merged.
 
+### Ensure backwards compatibility
+
+`lib` is the effective standard library for Nix, both for Nixpkgs and third-parties.
+Because of that it's important to make sure that the interface isn't changed in a breaking way.
```
Janik-Haag, line 63: Uhm, this seems a bit extreme. There will always be some factors causing a little change, sometimes even external ones. I think a better policy would be to avoid breaking changes if possible, but allow them for every new stable release if they are documented and reasoned accordingly.

infinisil: It's very rare that a breaking change is necessary, we haven't had a real one in the past 2 releases (and I didn't write release notes before that):
- [24.05](https://nixos.org/manual/nixos/stable/release-notes#sec-release-24.05-nixpkgs-lib): None at all
- [23.11](https://nixos.org/manual/nixos/stable/release-notes#sec-release-23.11-nixpkgs-lib): Just removal of deprecated functions and a _very_ minor one to `lib.foldl'` and friends

I think it's fine to say "no breaking changes", but not insist on it when there's a good case for it.

infinisil: @pennae I guess I can clarify that removing deprecated functions is fine after some time (but removing functions without deprecating them is not).

K900: Then that should be spelled out very clearly in the document, along with a deprecation process. "No API changes ever EVER" is not something we want, and not something we can achieve even if we did want it.

infinisil: The docs I'm writing here aren't something new, I've been maintaining `lib` with these guidelines for years now, and it hasn't caused any problems yet

infinisil: I think it would be helpful to align on what me mean by "breaking change". On that note I've essentially written a very formal definition of it in [RFC 33](https://github.com/Infinisil/rfcs/blob/deprecation/rfcs/0033-deprecation.md#types-of-deprecation), it's a bit overly formal, but please take a look! In such a framework, I don't consider `lib.mdDoc` to have been a breaking change.

But I also agree that there's a use-case for explicitly experimental features where any breaking change is okay, kind of like [RFC 82 (`lib.experimental`)](https://github.com/NixOS/rfcs/pull/82) and [my experiments with it](https://github.com/tweag/nixpkgs/commit/4df58c50e1f0b0169dfa73461ce4573760b43001). Imo it's all about setting expectations, making sure that users know if they can rely on something or not.

> insisting that lib is infinitely stable while pkgs gets to be broken left, right and center is absurd.

Not sure what you're trying to say with this. I'm not claiming `lib` to be "infinitely" stable (see the exceptions I wrote down), and neither did I saying anything about the stability of `pkgs` here. Personally I _am_ also a strong proponent of ensuring the rest of Nixpkgs to be fairly stable (I even [proposed a bachelor thesis topic](https://infinisil.com/nix-api-tracker.pdf) for this!), but that's not on my priorities right now.

infinisil: I updated the list of exceptions in this PR to also include that it's fine to break parts explicitly marked as experimental. We even [have some precedent for that](https://nixos.org/manual/nixpkgs/stable/#sec-functions-library-gvariant) (though no breaking change ever happened iirc).

Ericson2314: I think some parts more are more stable, and some parts are less stable (`lib.system`!), and actually it might be less contentious to skip straight to @infinisil's longer term goal of splitting out `lib` from Nixpkgs.

The `lib.system` part would have to stay behind in Nixpkgs, which is good! The remainder, living in the separate, much smaller repo, would naturally be subject to more scrutiny, which is also good!

roberth: I suppose the confusion here is about where the "interface" ends. Hyrum's law* does apply to `lib`, so one could conclude that the implementation is also not allowed to change in any observable way.
> \* With a sufficient number of users of an API, it does not matter what you promise in the contract: all observable behaviors of your system will be depended on by somebody.

However, this is now clarified in the exceptions below.
As a reader I can now conclude that it is about the _intended_ interface, and not the technical interface or all observable behaviors.

@infinisil has done a great job at maintaining `lib`, and he always listens to contributors, so I really don't get why the tone changed.

numinit: Splitting out `lib` would be nice. It would let users import it into flakes instead of needing to pull in all of nixpkgs (which I just had to do the other day) 👀

eclairevoyant: @numinit There is https://github.com/nix-community/nixpkgs.lib in today's world :)

Ericson2314: @eclairevoyant that's a good start :), but the real milestone is when Nixpkgs *uses* such a repo.

eclairevoyant: Sure. I was simply addressing numinit's usecase.

lib/README.md
```diff
@@ -57,6 +57,17 @@ Don't have multiple changes in one PR, instead split it up into multiple ones.
...
+
+The only exceptions are:
+- Breaking changes to interfaces that aren't in a release yet
+- Breaking changes that are unlikely to cause problems for users
+- Removal of functions that have been deprecated for at least one release
```
eclairevoyant, line 68: Does this mean it needs to be deprecated for 6 calendar months, or that it was deprecated before any previous release and functions can be removed the day after the release it was deprecated in?

lib/README.md
numinit: Neat, thanks. Still glad to see the latter case too :-)

lib/README.md
infinisil: They can be removed immediately following the release branch-off, so that the release will have the warning, while the next one won't, I'll clarify this.

lib/README.md
Janik-Haag: Splitting lib into a separate repository sounds like a bad idea. The "beauty" of nixpkgs is that you can easily do large refactors, bisecting, etc... there is a reason why most similar projects are structured in a monorepo like fashion (we can and should consider doing regular repository exports, maybe with something like [josh](https://github.com/josh-project/josh/)). What we should imho do is unify the entry points for lib and not make them gaslight you, currently there are like 4 different ones with at least one changing the behavior of `lib.evalModules` without telling you. But all of this is off-topic and beside the point, providing a longer stability guarantee for lib than we do for modules is ridiculous.

roberth: > changing the behavior of `lib.evalModules` without telling you

:eyes:  I think you're referring to the NixOS library `nixos/lib`, and I'd be guilty of that. We could rename that function there if you think it's confusing. To be clear that library is supposed to be NixOS-as-a-library, and not a clone of Nixpkgs `lib`.

As for the flake `lib`, that's definitely a problem. I'd be in favor of moving the Nixpkgs `lib` to `(getFlake "nixpkgs").lib.lib`, so that `lib.nixos` isn't presented as a Nixpkgs `lib` sublibrary.

What's the other entrypoint?

Janik-Haag: > 👀 I think you're referring to the NixOS library nixos/lib, and I'd be guilty of that. We could rename that function there if you think it's confusing. To be clear that library is supposed to be NixOS-as-a-library, and not a clone of Nixpkgs lib.

Yes I'm and yes I think having different names would make sense here, e.g. lib.evalModules and lib.evalNixOS or something similar. E.g. I was talking with someone about the `minimalModules` feature flag some time ago, and differentiating between `lib/` and `nixos/lib` was a bit confusing here.

> As for the flake lib, that's definitely a problem. I'd be in favor of moving the Nixpkgs lib to (getFlake "nixpkgs").lib.lib, so that lib.nixos isn't presented as a Nixpkgs lib sublibrary.

> What's the other entrypoint?

IIRC there is the flake one, the `lib/` one, the lib that gets passed to you inside a nixos module and `pkgs.lib`

> infinisil has done a great job at maintaining lib, and he always listens to contributors, so I really don't get why the tone changed.

I want to disagree with this, I know at least one case where people opted to contribute to nixos/lib/ instead of lib/ and multiple cases where people out right started to avoid contributing to nixpkgs because of the dreadful experiences they had trying to contribute to lib/

lib/README.md
roberth: I'd prefer for deprecations to start after old stable EOL, especially if it doesn't have the new alternative. That adds one month, but provides a guarantee that third party libraries work with all currently supported versions of lib, ensuring that maintainers don't get caught in this sequence of events:
- users report deprecation warning
- maintainer acts on it
- users report evaluation error on a supported Nixpkgs release
- frustrated maintainer has to revert or write a workaround

Alternatively, we could require a backporting of the new thing to stable / old stable to mitigate compatibility issues, but that's slightly more effort.

PR #322735, opened 2024-06-26
pass: add tail extension

Pablito2020:
## Description of changes

Added support for the "tail" extension in pass. The [AUR has this package](https://aur.archlinux.org/packages/pass-extension-tail), and when transitioning from Arch to Nix, I was surprised it wasn't in nixpkgs. So, I added support for it 😄.

You can now add the extension with:

`
pass.withExtensions (ext: [ext.pass-tail]);
`
This is especially useful with the Home Manager module, particularly with [this attribute of the password-store module](https://github.com/nix-community/home-manager/blob/master/modules/programs/password-store.nix#L20). Consequently, I didn't add pass-tail to all-packages.nix, as I didn't see any pass plugins there either.

## Things done

- Built on platform(s)
  - [X] x86_64-linux
  - [X] aarch64-linux
  - [X] x86_64-darwin
  - [X] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [X] Tested, as applicable:

### DIFF
--- a/pkgs/development/python-modules/pepit/default.nix
+++ b/pkgs/development/python-modules/pepit/default.nix
@@ -1,0 +1,53 @@
+{
+  lib,
+  buildPythonPackage,
+  fetchFromGitHub,
+  setuptools,
+  cvxpy,
+  numpy,
+  pandas,
+  scipy,
+  matplotlib,
+  pytestCheckHook,
+}:
+
+buildPythonPackage rec {
+  pname = "pepit";
+  version = "0.3.2";
+  pyproject = true;
+
+  src = fetchFromGitHub {
+    owner = "PerformanceEstimation";
+    repo = "PEPit";
+    rev = version;
+    hash = "sha256-Gdymdfi0Iv9KXBNSbAEWGYIQ4k5EONnbyWs+99L5D/A=";
+  };
+
+  build-system = [
+    setuptools
+  ];
+
+  dependencies = [
+    cvxpy
+    numpy
+    pandas
+    scipy
+    matplotlib
+  ];
+
+  nativeCheckInputs = [
+    pytestCheckHook
+  ];
+
+  pythonImportsCheck = [
+    "PEPit"
+  ];
+
+  meta = {
+    description = "Performance Estimation in Python";
+    changelog = "https://pepit.readthedocs.io/en/latest/whatsnew/${version}.html";
+    homepage = "https://pepit.readthedocs.io/";
+    license = lib.licenses.mit;
+    maintainers = with lib.maintainers; [ wegank ];
+  };
+}
--- a/pkgs/top-level/python-packages.nix
+++ b/pkgs/top-level/python-packages.nix
@@ -9890,6 +9890,8 @@ self: super: with self; {
 
   pep8-naming = callPackage ../development/python-modules/pep8-naming { };
 
+  pepit = callPackage ../development/python-modules/pepit { };
+
   peppercorn = callPackage ../development/python-modules/peppercorn { };
 
   percol = callPackage ../development/python-modules/percol { };


pkgs/tools/security/pass/extensions/tail.nix
```diff
@@ -0,0 +1,30 @@
+{ lib, stdenv, fetchFromGitHub }:
+
+stdenv.mkDerivation rec {
```
drupol, line 3: I think we should use the `finalAttrs` pattern here.

```nix
mkDerivation rec {
   foo = 1;
   bar = foo + 1;
}
```

pkgs/tools/security/pass/extensions/tail.nix
```diff
@@ -0,0 +1,30 @@
...
+  pname = "pass-tail";
+  version = "1.2.0";
+
+  src = fetchFromGitHub {
+    owner = "palortoff";
+    repo = "pass-extension-tail";
+    rev = "v${version}";
+    sha256 = "RyPXW4Lgx6GdF9j+zEJqKRGxinDwPCmF9WXJ5yW4Pcc=";
```
drupol, line 11: Prefer `hash = ...` to `sha256 = ...` Both currently work, but the former will make it easier if the algorithm ever needs to change.

pkgs/tools/security/pass/extensions/tail.nix
```diff
@@ -0,0 +1,30 @@
...
+  };
+
+  buildInputs = [ ];
```
drupol, line 14: Remove it

pkgs/tools/security/pass/extensions/tail.nix
```diff
@@ -0,0 +1,30 @@
...
+
+  dontBuild = true;
+
+  installFlags = [
+    "PREFIX=$(out)"
```
drupol, line 19: Isn't this automatically set?

pkgs/tools/security/pass/extensions/tail.nix
```diff
@@ -0,0 +1,30 @@
...
+    "BASHCOMPDIR=$(out)/share/bash-completion/completions"
+  ];
+
+  meta = with lib; {
```
drupol, line 23: You can easily remove `with lib;` here.

drupol (changes requested):

pkgs/tools/security/pass/extensions/tail.nix
Pablito2020: Didn't know rec was an "anti-pattern", now I see it. Will change it

pkgs/tools/security/pass/extensions/tail.nix
Pablito2020: I added it because if not, I would have to repeat it on each attribute:

```nix
  meta = {
    description = "A pass extension to avoid printing the password to the console";
    homepage = "https://github.com/palortoff/pass-extension-tail";
    license = lib.licenses.gpl3;
    maintainers = with lib.maintainers; [ pablito2020 ];
  };
```

So I thought that adding it made the code "cleaner":

```nix 
meta = with lib; {
    description = "A pass extension to avoid printing the password to the console";
    homepage = "https://github.com/palortoff/pass-extension-tail";
    license = licenses.gpl3;
    maintainers = with maintainers; [ pablito2020 ];
    platforms = platforms.unix;
};

```

I can agree with the "with maintainers" one.

pkgs/tools/security/pass/extensions/tail.nix
Pablito2020: True. Removed "dontBuild".

drupol: `dontBuild = true;` is fine, it's more the `PREFIX=$(out)`, have you tried building this without it?

pkgs/tools/security/pass/extensions/tail.nix
drupol: Yes, the idea is to completely remove it when it is in front of a set, just like it's done on Nixos/modules since a couple of months now.

pkgs/tools/security/pass/extensions/tail.nix
Pablito2020: It shows the error that it can't find /usr (of course, the Makefile searchs for it if PREFIX isn't defined (https://github.com/palortoff/pass-extension-tail/blob/master/Makefile#L1))

```
       > Running phase: installPhase
       > install flags: SHELL=/nix/store/agkxax48k35wdmkhmmija2i2sxg8i7ny-bash-5.2p26/bin/bash BASHCOMPDIR=\$\(out\)/share/bash-completion/completions install
       > install: cannot create directory '/usr': Permission denied
```

pkgs/tools/security/pass/extensions/tail.nix
Pablito2020: Okey, may I ask why? I mean in my eyes adding "with lib;" makes the code easier to read, but I can see that when you have a lot of attributes you don't know which ones are comming from lib and which ones are not.

PR #316309, opened 2024-06-01
fedimint: init at 0.3.2-rc.0

dpc:
## Description of changes

Add Fedimint package.

[Fedimint](https://fedimint.org/) is a module based system for building federated applications. It is designed to be a trust-minimized, censorship-resistant, and private alternative to centralized applications.

Fedimint is sizeable Rust project (>90k lines of Rust), in a big workspace that shares most dependencies. Build creates multiple binaries, intended to run on a different systems (cli, daemon, LN gateway with cli and daemon, etc.), so to avoid rebuilding the project multiple times, I went with the multiple outputs route.

I'm one of the main contributors to Fedimint and maintain our Nix Flake based CI system, etc. so I'm planning to maintain this package. Once this is accepted, I'd like to follow up with NixOS modules that are already part of our flake.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
- For non-Linux: Is sandboxing enabled in `nix.conf`? (See [Nix manual](https://nixos.org/manual/nix/stable/command-ref/conf-file.html))
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:
  - [NixOS test(s)](https://nixos.org/manual/nixos/unstable/index.html#sec-nixos-tests) (look inside [nixos/tests](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests))
  - and/or [package tests](https://github.com/NixOS/nixpkgs/blob/master/pkgs/README.md#package-tests)
  - or, for functions and "core" functionality, tests in [lib/tests](https://github.com/NixOS/nixpkgs/blob/master/lib/tests) or [pkgs/test](https://github.com/NixOS/nixpkgs/blob/master/pkgs/test)
  - made sure NixOS tests are [linked](https://nixos.org/manual/nixpkgs/unstable/#ssec-nixos-tests-linking) to the relevant packages
- [ ] Tested compilation of all packages that depend on this change using `nix-shell -p nixpkgs-review --run "nixpkgs-review rev HEAD"`. Note: all changes have to be committed, also see [nixpkgs-review usage](https://github.com/Mic92/nixpkgs-review#usage)
- [x] Tested basic functionality of all binary files (usually in `./result/bin/`)
[...]

PR #322815, opened 2024-06-27
nixos/fedimintd: init services

dpc:
## Description of changes

Follow-up to #316309 adding nixos module and a basic vmtest.

Based on https://github.com/NixOS/nixpkgs/blob/nixos-24.05/nixos/modules/services/networking/bitcoind.nix , which has similar requirements (like being able to run multiple instances for different Bitcoin networks) and is a closely related project.

I also tried to mimic what I've seen in #314440 because I happen to watch that PR.

This is a first time I'm adding NixOS module to nixpkgs, so please be patient with me.

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/nixos/modules/services/networking/fedimintd.nix
+++ b/nixos/modules/services/networking/fedimintd.nix
@@ -1,0 +1,304 @@
+{
+  config,
+  lib,
+  pkgs,
+  ...
+}:
+let
+  inherit (lib)
+    concatLists
+    filterAttrs
+    mapAttrs'
+    mapAttrsToList
+    mkEnableOption
+    mkIf
+    mkOption
+    mkOverride
+    mkPackageOption
+    nameValuePair
+    recursiveUpdate
+    types
+    ;
+
+  fedimintdOpts =
+    {
+      config,
+      lib,
+      name,
+      ...
+    }:
+    {
+      options = {
+        enable = mkEnableOption "fedimintd";
+
+        package = mkPackageOption pkgs "fedimint" { };
+
+        environment = mkOption {
+          type = types.attrsOf types.str;
+          description = "Extra Environment variables to pass to the fedimintd.";
+          default = {
+            RUST_BACKTRACE = "1";
+          };
+          example = {
+            RUST_LOG = "info,fm=debug";
+            RUST_BACKTRACE = "1";
+          };
+        };
+
+        p2p = {
+          openFirewall = mkOption {
+            type = types.bool;
+            default = true;
+            description = "Opens port in firewall for fedimintd's p2p port";
+          };
+          port = mkOption {
+            type = types.port;
+            default = 8173;
+            description = "Port to bind on for p2p connections from peers";
+          };
+          bind = mkOption {
+            type = types.str;
+            default = "0.0.0.0";
+            description = "Address to bind on for p2p connections from peers";
+          };
+          url = mkOption {
+            type = types.str;
+            example = "fedimint://p2p.myfedimint.com";
+            description = ''
+              Public address for p2p connections from peers
+            '';
+          };
+        };
+        api = {
+          openFirewall = mkOption {
+            type = types.bool;
+            default = false;
+            description = "Opens port in firewall for fedimintd's api port";
+          };
+          port = mkOption {
+            type = types.port;
+            default = 8174;
+            description = "Port to bind on for API connections relied by the reverse proxy/tls terminator.";
+          };
+          bind = mkOption {
+            type = types.str;
+            default = "127.0.0.1";
+            description = "Address to bind on for API connections relied by the reverse proxy/tls terminator.";
+          };
+          url = mkOption {
+            type = types.str;
+            description = ''
+              Public URL of the API address of the reverse proxy/tls terminator. Usually starting with `wss://`.
+            '';
+          };
+        };
+        bitcoin = {
+          network = mkOption {
+            type = types.str;
+            default = "signet";
+            example = "bitcoin";
+            description = "Bitcoin network to participate in.";
+          };
+          rpc = {
+            url = mkOption {
+              type = types.str;
+              default = "http://127.0.0.1:38332";
+              example = "signet";
+              description = "Bitcoin node (bitcoind/electrum/esplora) address to connect to";
+            };
+
+            kind = mkOption {
+              type = types.str;
+              default = "bitcoind";
+              example = "electrum";
+              description = "Kind of a bitcoin node.";
+            };
+
+            secretFile = mkOption {
+              type = types.nullOr types.path;
+              default = null;
+              description = ''
+                If set the URL specified in `bitcoin.rpc.url` will get the content of this file added
+                as an URL password, so `http://user@example.com` will turn into `http://user:SOMESECRET@example.com`.
+
+                Example:
+
+                `/etc/nix-bitcoin-secrets/bitcoin-rpcpassword-public` (for nix-bitcoin default)
+              '';
+            };
+          };
+        };
+
+        consensus.finalityDelay = mkOption {
+          type = types.ints.unsigned;
+          default = 10;
+          description = "Consensus peg-in finality delay.";
+        };
+
+        dataDir = mkOption {
+          type = types.path;
+          default = "/var/lib/fedimintd-${name}/";
+          readOnly = true;
+          description = ''
+            Path to the data dir fedimintd will use to store its data.
+            Note that due to using the DynamicUser feature of systemd, this value should not be changed
+            and is set to be read only.
+          '';
+        };
+
+        nginx = {
+          enable = mkOption {
+            type = types.bool;
+            default = false;
+            description = ''
+              Whether to configure nginx for fedimintd
+            '';
+          };
+          fqdn = mkOption {
+            type = types.str;
+            example = "api.myfedimint.com";
+            description = "Public domain of the API address of the reverse proxy/tls terminator.";
+          };
+          config = mkOption {
+            type = types.submodule (
+              recursiveUpdate (import ../web-servers/nginx/vhost-options.nix {
+                inherit config lib;
+              }) { }
+            );
+            default = { };
+            description = "Overrides to the nginx vhost section for api";
+          };
+        };
+      };
+    };
+in
+{
+  options = {
+    services.fedimintd = mkOption {
+      type = types.attrsOf (types.submodule fedimintdOpts);
+      default = { };
+      description = "Specification of one or more fedimintd instances.";
+    };
+  };
+
+  config =
+    let
+      eachFedimintd = filterAttrs (fedimintdName: cfg: cfg.enable) config.services.fedimintd;
+      eachFedimintdNginx = filterAttrs (fedimintdName: cfg: cfg.nginx.enable) eachFedimintd;
+    in
+    mkIf (eachFedimintd != { }) {
+
+      networking.firewall.allowedTCPPorts = concatLists (
+        mapAttrsToList (
+          fedimintdName: cfg:
+          (lib.optional cfg.api.openFirewall cfg.api.port ++ lib.optional cfg.p2p.openFirewall cfg.p2p.port)
+        ) eachFedimintd
+      );
+
+      systemd.services = mapAttrs' (
+        fedimintdName: cfg:
+        (nameValuePair "fedimintd-${fedimintdName}" (
+          let
+            startScript = pkgs.writeShellScript "fedimintd-start" (
+              (
+                if cfg.bitcoin.rpc.secretFile != null then
+                  ''
+                    secret=$(${pkgs.coreutils}/bin/head -n 1 "${cfg.bitcoin.rpc.secretFile}")
+                    prefix="''${FM_BITCOIN_RPC_URL%*@*}"  # Everything before the last '@'
+                    suffix="''${FM_BITCOIN_RPC_URL##*@}"  # Everything after the last '@'
+                    FM_BITCOIN_RPC_URL="''${prefix}:''${secret}@''${suffix}"
+                  ''
+                else
+                  ""
+              )
+              + ''
+                exec ${cfg.package}/bin/fedimintd
+              ''
+            );
+          in
+          {
+            description = "Fedimint Server";
+            documentation = [ "https://github.com/fedimint/fedimint/" ];
+            wantedBy = [ "multi-user.target" ];
+            environment = lib.mkMerge [
+              {
+                FM_BIND_P2P = "${cfg.p2p.bind}:${toString cfg.p2p.port}";
+                FM_BIND_API = "${cfg.api.bind}:${toString cfg.api.port}";
+                FM_P2P_URL = cfg.p2p.url;
+                FM_API_URL = cfg.api.url;
+                FM_DATA_DIR = cfg.dataDir;
+                FM_BITCOIN_NETWORK = cfg.bitcoin.network;
+                FM_BITCOIN_RPC_URL = cfg.bitcoin.rpc.url;
+                FM_BITCOIN_RPC_KIND = cfg.bitcoin.rpc.kind;
+              }
+              cfg.environment
+            ];
+            serviceConfig = {
+              DynamicUser = true;
+
+              StateDirectory = "fedimintd-${fedimintdName}";
+              StateDirectoryMode = "0700";
+              ExecStart = startScript;
+
+              Restart = "always";
+              RestartSec = 10;
+              StartLimitBurst = 5;
+              UMask = "007";
+              LimitNOFILE = "100000";
+
+              LockPersonality = true;
+              MemoryDenyWriteExecute = true;
+              NoNewPrivileges = true;
+              PrivateDevices = true;
+              PrivateMounts = true;
+              PrivateTmp = true;
+              ProtectClock = true;
+              ProtectControlGroups = true;
+              ProtectHostname = true;
+              ProtectKernelLogs = true;
+              ProtectKernelModules = true;
+              ProtectKernelTunables = true;
+              ProtectSystem = "full";
+              RestrictAddressFamilies = [
+                "AF_INET"
+                "AF_INET6"
+              ];
+              RestrictNamespaces = true;
+              RestrictRealtime = true;
+              SystemCallArchitectures = "native";
+              SystemCallFilter = [
+                "@system-service"
+                "~@privileged"
+              ];
+            };
+          }
+        ))
+      ) eachFedimintd;
+
+      services.nginx.virtualHosts = mapAttrs' (
+        fedimintdName: cfg:
+        (nameValuePair cfg.nginx.fqdn (
+          lib.mkMerge [
+            cfg.nginx.config
+
+            {
+              # Note: we want by default to enable OpenSSL, but it seems anything 100 and above is
+              # overriden by default value from vhost-options.nix
+              enableACME = mkOverride 99 true;
+              forceSSL = mkOverride 99 true;
+              # Currently Fedimint API only support JsonRPC on `/ws/` endpoint, so no need to handle `/`
+              locations."/ws/" = {
+                proxyPass = "http://127.0.0.1:${toString cfg.api.port}/";
+                proxyWebsockets = true;
+                extraConfig = ''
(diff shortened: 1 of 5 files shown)


nixos/tests/fedimintd.nix
```diff
@@ -0,0 +1,34 @@
...
+        fqdn = "example.com";
+      };
+      api = {
+        fqdn = "example.com";
+      };
+    };
+  };
+
+
+  testScript = { nodes, ... }: ''
```
SuperSandro2000, line 27:
```suggestion
  };

  testScript = { nodes, ... }: ''
```

nixos/tests/fedimintd.nix
```diff
@@ -0,0 +1,34 @@
...
+    start_all()
+
+    machine.wait_for_unit("fedimintd-mainnet.service")
+    machine.wait_for_open_port(${toString nodes.machine.services.fedimintd.mainnet.api.port})
+  '';
+}
+
```
SuperSandro2000, line 34: There is one new line to many here 😅

nixos/modules/services/networking/fedimintd.nix
```diff
@@ -0,0 +1,357 @@
...
+        nameValuePair "${cfg.group}" { }
+      ))
+      eachFedimintd;
+
+    services.nginx.virtualHosts = mapAttrs'
+      (fedimintdName: cfg: (
+        nameValuePair cfg.api.fqdn (lib.mkMerge [
+
+          cfg.nginx.config
+
```
SuperSandro2000, line 338:
```suggestion
          cfg.nginx.config
```

nixos/modules/services/networking/fedimintd.nix
```diff
@@ -0,0 +1,357 @@
...
+          {
+            enableACME = mkDefault true;
+            forceSSL = mkDefault true;
+            locations."/ws/" = {
+              proxyPass = "http://127.0.0.1:${builtins.toString cfg.api.port}/";
+              proxyWebsockets = true;
+              extraConfig = "proxy_pass_header Authorization;";
```
SuperSandro2000, line 345:
```suggestion
              extraConfig = ''
                proxy_pass_header Authorization;
              '';
```
otherwise it might get concatinated wrongly when being merged

nixos/modules/services/networking/fedimintd.nix
```diff
@@ -0,0 +1,357 @@
...
+    services.nginx.virtualHosts = mapAttrs'
+      (fedimintdName: cfg: (
+        nameValuePair cfg.api.fqdn (lib.mkMerge [
+
+          cfg.nginx.config
+
+          {
+            enableACME = mkDefault true;
+            forceSSL = mkDefault true;
+            locations."/ws/" = {
```
SuperSandro2000, line 342: Don't we also need to handle /?

nixos/modules/services/networking/fedimintd.nix
```diff
@@ -0,0 +1,357 @@
...
+                SystemCallArchitectures = "native";
+                SystemCallFilter = [
+                  "@system-service"
+                  "~@privileged"
+                ];
+                StateDirectory = "fedimintd-${fedimintdName}";
+                StateDirectoryMode = "0700";
+                ExecStart = startScript;
+
+                # Hardening measures
```
SuperSandro2000, line 303:
```suggestion
```
there are also hardening options above this comment

nixos/modules/services/networking/fedimintd.nix
```diff
@@ -0,0 +1,357 @@
...
+                cfg.extraEnvironment
+              ]);
+              serviceConfig = {
+                User = cfg.user;
+                Group = cfg.group;
+
+                Restart = "always";
+                RestartSec = 10;
+                StartLimitBurst = 5;
+                UMask = "077";
```
SuperSandro2000, line 280: Why not allow the group to read data?

nixos/modules/services/networking/fedimintd.nix
```diff
@@ -0,0 +1,357 @@
...
+              );
+            in
+            {
+              description = "Fedimint Server";
+              documentation = [ "https://github.com/fedimint/fedimint/" ];
+              wantedBy = [ "multi-user.target" ];
+              environment = lib.mkMerge ([
+                {
+                  FM_BIND_P2P = "${cfg.p2p.bind}:${builtins.toString cfg.p2p.port}";
+                  FM_BIND_API = "${cfg.api.bind}:${builtins.toString cfg.api.port}";
```
SuperSandro2000, line 263:
```suggestion
                  FM_BIND_P2P = "${cfg.p2p.bind}:${toString cfg.p2p.port}";
                  FM_BIND_API = "${cfg.api.bind}:${toString cfg.api.port}";
```

nixos/modules/services/networking/fedimintd.nix
```diff
@@ -0,0 +1,357 @@
...
+        (fedimintdName: cfg:
+          (
+            if cfg.api.openFirewall then [
+              cfg.api.port
+            ] else [ ]
+          ) ++ (
+            if cfg.p2p.openFirewall then [
+              cfg.p2p.port
+            ] else [ ]
+          )
```
SuperSandro2000, line 231:
```suggestion
          (
            lib.optional cfg.api.openFirewall cfg.api.port
            ++ lib.optional cfg.p2p.openFirewall cfg.p2p.port
          )
```

nixos/modules/services/networking/fedimintd.nix
```diff
@@ -0,0 +1,357 @@
...
+
+    assertions = flatten
+      (mapAttrsToList
+        (fedimintdName: cfg: [
+          {
+            assertion = cfg.p2p.address != null;
+            message = ''
+              `services.fedimintd.${fedimintdName}.p2p.address` must be set to address reachable by other peers.
+
+              Example: `fedimint://p2p.mymint.org`.
```
SuperSandro2000, line 206: Address could here also refer to an ip address. Maybe we should use URI?

nixos/modules/services/networking/fedimintd.nix
```diff
@@ -0,0 +1,357 @@
...
+            '';
+          }
+          {
+            assertion = cfg.api.address != null;
+            message = ''
+              `services.fedimintd.${fedimintdName}.api.address` must be set to address reachable by the clients, with TLS terminated by external service (typically nginx), and relayed to the fedimintd bind address.
+
+              Example: `wss://api.mymint.org`.
```
SuperSandro2000, line 214: Same here, We should probably use URI

nixos/modules/services/networking/fedimintd.nix
```diff
@@ -0,0 +1,357 @@
...
+        readOnly = true;
+        description = ''
+          Path to the data dir fedimintd will use to store its data.
+          Note that due to using the DynamicUser feature of systemd, this value should not be changed
+          and is set to be read only.
+        '';
+      };
+
+      nginx = {
+        enable = mkEnableOption "fedimint";
```
SuperSandro2000, line 170: We should probably change the description here to `Wether to configure nginx for fedimint` or similar

nixos/modules/services/networking/fedimintd.nix
```diff
@@ -0,0 +1,357 @@
...
+        config = mkOption {
+          type = types.nullOr (types.submodule (
+            recursiveUpdate (import ../web-servers/nginx/vhost-options.nix { inherit config lib; }) {
+              options.serverName = {
+                default = "radicle-${config.networking.hostName}.${config.networking.domain}";
+                defaultText = "radicle-\${config.networking.hostName}.\${config.networking.domain}";
```
SuperSandro2000, line 176: Copy paste error

nixos/modules/services/networking/fedimintd.nix
```diff
@@ -0,0 +1,357 @@
...
+        };
+        port = mkOption {
+          type = types.port;
+          default = 8174;
+          description = "Port to bind on for API connections relied by the reverse proxy/tls terminator.";
+        };
+        bind = mkOption {
+          type = types.str;
+          default = "127.0.0.1";
+          description = "Address to bind on for API connections relied by the reverse proxy/tls terminator. Usually starting with `fedimint://`";
```
SuperSandro2000, line 97: The default is not really matching the description

nixos/modules/services/networking/fedimintd.nix
```diff
@@ -0,0 +1,357 @@
...
+        description = "The user as which to run fedimintd.";
+      };
+
+      group = mkOption {
+        type = types.str;
+        default = config.user;
+        description = "The group as which to run fedimintd.";
+      };
+
+      extraEnvironment = mkOption {
```
SuperSandro2000, line 38:
```suggestion
      environment = mkOption {
```

nixos/modules/services/networking/fedimintd.nix
dpc: Not really. But I added a comment to explain it.

nixos/modules/services/networking/fedimintd.nix
dpc: Just removed and sorted it all.

nixos/modules/services/networking/fedimintd.nix
dpc: Good point. I went with `url`, as we do require `protocol://name`

nixos/modules/services/networking/fedimintd.nix
dpc: IIUC no other way than to convert to `mkOption` for that, so did that.

nixos/modules/services/networking/fedimintd.nix
dpc: :see_no_evil: Oops.

dpc:
@SuperSandro2000 Lots of useful things I didn't catch. Thank you so much.

dpc:
Pleeeease? :D

h7x4:
I don't have time to review today, but I'll see what I can do tomorrow. Ping me if nothing has happened by friday :)

nixos/modules/services/networking/fedimintd.nix
```diff
@@ -0,0 +1,363 @@
...
+        home = cfg.dataDir;
+        isSystemUser = true;
+      })
+    ) eachFedimintd;
+
+    users.groups = mapAttrs' (fedimintdName: cfg: (nameValuePair "${cfg.group}" { })) eachFedimintd;
+
+    services.nginx.virtualHosts = mapAttrs' (
+      fedimintdName: cfg:
+      (nameValuePair cfg.api.fqdn (
```
h7x4, line 341: This is a bit of an interesting trap. Further up, you allow the `fqdn` to be `null`, presumably to be able to set a conditional default. But there's a quirk in the nix language where if it's given `null` as a key in an attrset, it will silently delete that attribute.

```
nix-repl> { "hello" = 1; }
{ hello = 1; }

nix-repl> { ${null} = 1; }
{ }
```

This could mean that people who unknowingly forget to set `fqdn` suddenly doesn't find their nginx entries, despite explicitly setting `nginx.enable`.

I think the best way to fix this, is to set `fqdn`s type to `types.str` (and also `url`), not allowing null values, and setting a conditional default in `config`. That way, the module system will complain if you haven't set the value during eval. This is also more semantically correct, as `null` isn't really a valid value if you want the service to work (if I have understood correctly at least).

```nix
options.<...> = {
  fqdn = mkOption {
    type = types.str; # no null
    example = "api.myfedimint.com";
    description = "Public domain of the API address of the reverse proxy/tls terminator.";
  };
  url = mkOption {
    type = types.str; # no null
    # no default
    description = ''
      Public URL of the API address of the reverse proxy/tls terminator. Usually starting with `wss://`.
      Typically you want to override `fqdn` instead.
    '';
  };
};

config = {
  # note: no need for the url assertions anymore.

  services.fedimintd = mapAttrs (name: value: {
    api.url = lib.mkIf (value.p2p.url != null) (lib.mkDefault "fedimint://${value.p2p.fqdn}");
    p2p.url = lib.mkIf (value.api.url != null) (lib.mkDefault "wss://${value.api.fqdn}");
  }) eachFedimintd;
};
```

I'm not certain this doesn't lead to infinite recursion. It might be better to drop the magic defaults and let people do the magic themselves.

nixos/modules/services/networking/fedimintd.nix
```diff
@@ -0,0 +1,363 @@
...
+      (nameValuePair "fedimintd-${fedimintdName}" {
+        name = cfg.user;
+        group = cfg.group;
+        description = "Fedimint daemon user";
+        home = cfg.dataDir;
+        isSystemUser = true;
+      })
+    ) eachFedimintd;
+
+    users.groups = mapAttrs' (fedimintdName: cfg: (nameValuePair "${cfg.group}" { })) eachFedimintd;
```
h7x4, line 337: It would be great if we could make systemd's `DynamicUser` option work for this, so we don't need so many hardcoded system users with dynamic UID/GIDs. Could you try dropping the users, setting `systemd.services.<name>.serviceConfig.DynamicUser = true`, and see if it still works normally? Considering the UMask and the StateDirectory privileges, it seems likely to work

nixos/modules/services/networking/fedimintd.nix
```diff
@@ -0,0 +1,363 @@
...
+      (nameValuePair cfg.api.fqdn (
+        lib.mkMerge [
+          cfg.nginx.config
+
+          {
+            enableACME = mkDefault true;
+            forceSSL = mkDefault true;
+            # Currently Fedimint API only support JsonRPC on `/ws/` endpoint, so no need to handle `/`
+            locations."/ws/" = {
+              proxyPass = "http://127.0.0.1:${builtins.toString cfg.api.port}/";
```
h7x4, line 350: Small nit, not blocking

```suggestion
              proxyPass = "http://127.0.0.1:${toString cfg.api.port}/";
```

nixos/modules/services/networking/fedimintd.nix
```diff
@@ -0,0 +1,363 @@
...
+          message = ''
+            `services.fedimintd.${fedimintdName}.api.url` must be set to address reachable by the clients, with TLS terminated by external service (typically nginx), and relayed to the fedimintd bind address.
+
+            Example: `wss://api.mymint.org`.
+          '';
+        }
+      ]) eachFedimintd
+    );
+
+    networking.firewall.allowedTCPPorts = flatten (
```
h7x4, line 238: Slightly more efficient since you know the list nesting only is one layer deep. Same for `assertions` maybe.

```suggestion
    networking.firewall.allowedTCPPorts = concatLists (
```

nixos/modules/services/networking/fedimintd.nix
```diff
@@ -0,0 +1,363 @@
...
+            and is set to be read only.
+          '';
+        };
+
+        nginx = {
+          enable = mkOption {
+            type = types.bool;
+            default = false;
+            description = ''
+              Wether to configure nginx for fedimintd
```
h7x4, line 187:
```suggestion
              Whether to configure nginx for fedimintd
```

nixos/modules/services/networking/fedimintd.nix
```diff
@@ -0,0 +1,363 @@
...
+                Example:
+
+                `/etc/nix-bitcoin-secrets/bitcoin-rpcpassword-public` (for nix-bitcoin default)
+              '';
+            };
+          };
+        };
+
+        consensus.finalityDelay = mkOption {
+          type = types.number;
```
h7x4, line 166: Just doublechecking that this in fact can be either a float and int, both positive and negative? Maybe `types.numbers.nonnegative` or `types.ints.unsigned` would be more fitting?

nixos/modules/services/networking/fedimintd.nix
```diff
@@ -0,0 +1,363 @@
...
+    mkIf
+    mkOption
+    mkPackageOption
+    nameValuePair
+    recursiveUpdate
+    types
+    ;
+
+  eachFedimintd = filterAttrs (fedimintdName: cfg: cfg.enable) config.services.fedimintd;
+  eachFedimintdNginx = filterAttrs (fedimintdName: cfg: cfg.nginx.enable) eachFedimintd;
```
h7x4, line 24: Not a blocker, but it would be great if you could move these down into a let block under config, for better code locality. Had to scroll up and down a few times to look at these. Something like this?

```nix
config = let
  eachFedimintd = filterAttrs (fedimintdName: cfg: cfg.enable) config.services.fedimintd;
  eachFedimintdNginx = filterAttrs (fedimintdName: cfg: cfg.nginx.enable) eachFedimintd;
in mkIf (eachFedimintd != { }) {
  ...
};
```

h7x4 (commented):
Here's a few comments. Maybe you could add a release note as well?

nixos/modules/services/networking/fedimintd.nix
```diff
@@ -0,0 +1,363 @@
...
+            default = false;
+            description = ''
+              Wether to configure nginx for fedimintd
+            '';
+          };
+          config = mkOption {
+            type = types.submodule (
+              recursiveUpdate (import ../web-servers/nginx/vhost-options.nix { inherit config lib; }) {
+                options.serverName = {
+                  default = "fedimintd-${name}";
```
dpc, line 194: This needs to be the fqdn

nixos/modules/services/networking/fedimintd.nix
```diff
@@ -0,0 +1,363 @@
...
+    users.groups = mapAttrs' (fedimintdName: cfg: (nameValuePair "${cfg.group}" { })) eachFedimintd;
+
+    services.nginx.virtualHosts = mapAttrs' (
+      fedimintdName: cfg:
+      (nameValuePair cfg.api.fqdn (
+        lib.mkMerge [
+          cfg.nginx.config
+
+          {
+            enableACME = mkDefault true;
```
dpc, line 346: This gets overwritten somewhere. Maybe use a higher level?

nixos/modules/services/networking/fedimintd.nix
dpc: I didn't know about these. `types.ints.unsigned` is perfect.

dpc:
@h7x4 I think I addressed all issues, yours (thank you!) and some other ones I found while setting up configurations based on this PR.

dpc:
I'm running `nix run nixpkgs#nixfmt nixos/modules/services/networking/fedimintd.nix`, and it doesn't make any differences.

otech47:
ack

h7x4:
> I'm running nix run nixpkgs#nixfmt nixos/modules/services/networking/fedimintd.nix, and it doesn't make any changes anymore, so why is nixfmt-check step not like me?

I believe `nixpkgs#nixfmt` still refers to the old formatter (aliased to `nixfmt-classic`). Try using the `nixfmt` executable from the `nixfmt-rfc-style` package, maybe manually via `nix shell nixpkgs#nixfmt-rfc-style`

dpc:
`nix run nixpkgs#nixfmt-rfc-style nixos/modules/services/networking/fedimintd.nix` did it. I thought I was already using `nixfmt-rfc-style` the first time it failed, so I confused myself believing I need to use the classic one.

h7x4 (approved):
Okay, I think it's good to go. I have a few more comments, nothing important.

nixos/modules/services/networking/fedimintd.nix
```diff
@@ -0,0 +1,316 @@
...
+        };
+      };
+    };
+in
+{
+  options = {
+    services.fedimintd = mkOption {
+      type = types.attrsOf (types.submodule fedimintdOpts);
+      default = { };
+      description = "Specification of one or more fedimintd instances.";
```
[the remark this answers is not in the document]
dpc: I'm afraid we don't have a good place to link to yet, relying on Dockerfile templates and other instructions still under development. Hopefully I can update it in the future.

h7x4:
Great stuff, thank you for the contribution!

PR #322906, opened 2024-06-27
openblas: support CPUs with more than 64 cores

drupol:
## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [x] Tested, as applicable:

---

### DIFF
--- a/pkgs/development/libraries/science/math/openblas/default.nix
+++ b/pkgs/development/libraries/science/math/openblas/default.nix
@@ -225,7 +225,7 @@ stdenv.mkDerivation (finalAttrs: {
       CC = "${stdenv.cc.targetPrefix}${if stdenv.cc.isClang then "clang" else "cc"}";
       PREFIX = placeholder "out";
       OPENBLAS_INCLUDE_DIR = "${placeholder "dev"}/include";
-      NUM_THREADS = 64;
+      NUM_THREADS = 128;
       INTERFACE64 = blas64;
       NO_STATIC = !enableStatic;
       NO_SHARED = !enableShared;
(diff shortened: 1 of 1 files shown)


pkgs/development/libraries/science/math/openblas/default.nix
```diff
@@ -219,7 +219,7 @@ stdenv.mkDerivation (finalAttrs: {
       CC = "${stdenv.cc.targetPrefix}${if stdenv.cc.isClang then "clang" else "cc"}";
       PREFIX = placeholder "out";
       OPENBLAS_INCLUDE_DIR = "${placeholder "dev"}/include";
-      NUM_THREADS = 64;
+      NUM_THREADS = 128;
```
SomeoneSerge, line 232: These seem like they're inconvenient to override in an overlay by the way

drupol: What do you mean? Do you want me to add a toplevel parameter?

SomeoneSerge: Maybe not yet (let's not do anything before we find we need it), just a general observation that if I had to override this locally I'd have to `builtins.filter` the makeFlags...

drupol: Oh because it's a list? Yeah indeed... but I'm not going to change how `makeFlags` and `buildFlags` are supposed to work.

Also, it would be interesting to know how these flags behaves when there's something like this: `makeFlags = [ "FOO=1" "BAR=2" "FOO=3" ];`. Does `FOO` is equal to `1` or `3` ?

PR #322980, opened 2024-06-27
nixos/virtualisation: increase priority for libvirt NSS modules

### LINKED ISSUE #322022: libvirtd NSS module not working (nscd, nsncd, libnss, libc)
### Describe the bug

I was happy to learn that in NixOS 24.05 the virtualization module gained [NSS module support for libvirtd](https://github.com/NixOS/nixpkgs/commit/412543ddd83f4ee235ee32772afc19483aaf83ab).

However, I found it not working: `getent hosts <hostname>` does not resolve to an IP address, even though the VM host name and IP is present in libvirt.

### Steps To Reproduce

With the following NixOS configuration:

```nix
virtualisation = {
    libvirtd = {
      enable = true;
      nss = {
        enable = true;
        enableGuest = true;
      };
    };
  };
```

The following entry in `/etc/nsswitch.conf` is correctly created:

```
hosts:     mymachines resolve [!UNAVAIL=return] files myhostname libvirt libvirt_guest dns
```

The path to `libnss_libvirt.so.2` and `libnss_libvirt_guest.so.2` is also correctly added to the `LD_LIBRARY_PATH` environment variable of the `nscd.service`, as evidenced by `systemctl show nscd`:

```
Environment=LD_LIBRARY_PATH=/nix/store/gzvzqc3lyya4xwjrplmfk3x6l160rr62-libvirt-10.0.0/lib:/nix/store/xjiifrz7ha6s29gp0p0j3w0155phxmia-systemd-255.6/lib
```

The data source of `libnss_libvirt*` [are plain text files](https://github.com/libvirt/libvirt/blob/v10.4.0/tools/nss/libvirt_nss.c#L134) in `/var/lib/libvirt/dnsmasq`, which makes it a lot easier to reproduce the issue because no actual VMs have to be created using libvirt. Here are two files that can be used for reproduction (**Note:** The `expiry-time` below is significant, and should lie in your future, otherwise the name lookups will not work.)

`/var/lib/libvirt/dnsmasq/virbr0.macs`:

```
[
  {
    "domain": "debian11",
    "macs": [
      "52:54:00:60:dc:fe"
    ]
  }
] 
```

`/var/lib/libvirt/dnsmasq/virbr0.status`:

```
[
  {
    "ip-address": "192.168.122.8",
    "mac-address": "52:54:00:60:dc:fe",
    "hostname": "debian11",
    "expiry-time": 1819155914
  }
] 
```

The actual querying of the name can now be done with `getent hosts debian11`. **This does not return a result (exit code 2).**

### Expected behavior

`getent hosts debian11` should return the IP address `192.168.122.8`.

### Additional context

Having more than one directory in `LD_LIBRARY_PATH` does not work.

Having just one directory in `LD_LIBRARY_PATH`, the libvirt one, does work.

To verify those two statements, do the following steps as `root`.

Please take extra care below to distinguish between `nscd` and `nsncd`.

```sh
systemctl stop nscd
nix-shell -p nsncd
LD_LIBRARY_PATH="/nix/store/gzvzqc3lyya4xwjrplmfk3x6l160rr62-libvirt-10.0.0/lib" NSNCD_WORKER_COUNT=1 NSNCD_HANDOFF_TIMOUT=10 strace -f nsncd 2>&1 | grep dnsmasq # adapt the libvirt path to your Nix store
```

In a different terminal, the name resolution is successful:

```
getent hosts debian11
# ->192.168.122.8   debian11
```

`strace` in the first terminal shows that the data source `/var/lib/libvirt/ndsmasq` is actually consulted:

```
[pid  9358] openat(AT_FDCWD, "/var/lib/libvirt/dnsmasq/", O_RDONLY|O_NONBLOCK|O_CLOEXEC|O_DIRECTORY) = 8
[pid  9358] openat(AT_FDCWD, "/var/lib/libvirt/dnsmasq//virbr0.status", O_RDONLY) = 8
[pid  9358] openat(AT_FDCWD, "/var/lib/libvirt/dnsmasq/", O_RDONLY|O_NONBLOCK|O_CLOEXEC|O_DIRECTORY) = 8
[pid  9358] openat(AT_FDCWD, "/var/lib/libvirt/dnsmasq//virbr0.macs", O_RDONLY) = 9
[pid  9358] openat(AT_FDCWD, "/var/lib/libvirt/dnsmasq//virbr0.status", O_RDONLY) = 8
[pid  9358] openat(AT_FDCWD, "/var/lib/libvirt/dnsmasq/", O_RDONLY|O_NONBLOCK|O_CLOEXEC|O_DIRECTORY) = 8
[pid  9358] openat(AT_FDCWD, "/var/lib/libvirt/dnsmasq//virbr0.status", O_RDONLY) = 8
```

Now, repeat the same experiment, but add the second path to `LD_LIBRARY_PATH` (the actual value from `systemd show nscd`), e.g.
`LD_LIBRARY_PATH="/nix/store/gzvzqc3lyya4xwjrplmfk3x6l160rr62-libvirt-10.0.0/lib:/nix/store/xjiifrz7ha6s29gp0p0j3w0155phxmia-systemd-255.6/lib"` to the `nsncd` invocation.

This time, `grep` does not match any lines and `getent hosts debian11` does not return a result.

**Note:** If you pass `--service libvirt` (or similar) to `getent`, this will bypass the call to nscd and attempt to go to the libnss libraries directly, so this will always fail on NixOS (unless the needed shared libraries are in scope).

I do not know enough about the actual NSS mechanism inside glibc to understand why `libnss_libvirt.so.2` and `libnss_libvirt_guest.so.2` are not consulted.

Since `nsncd` does not do actual lookups, and only calls into glibc (`gethostbyname()` or such), this looks to me like a glibc issue, but this is currently out of my depth to analyze.

In any case, the NixOS feature as I think it was intended to work, does not, hence I think that this is a bug.

I'm happy to provide more details if needed.

### Notify maintainers

Sorry for notifying so many of you but this issue looks like it could be interconnected between several parts.

nixos/modules/virtualization/libvirtd.nix: @jchv 

nixos/modules/config/nsswitch.nix: @dasJ 

nsncd: @flokli  @picnoir 

libvirt: @fpletz @globin @lovesegfault 

### Metadata

```console

 - system: `"x86_64-linux"`
 - host os: `Linux 6.6.33, NixOS, 24.05 (Uakari), 24.05.20240615.752c634`
 - multi-user?: `yes`
 - sandbox: `yes`
 - nixpkgs: `/nix/store/mcwr2j04fikfsrsaq76f4bviinvl6zql-source`
```

---

michaelfranzl:
When `services.resolved` is enabled, then `resolve [!UNAVAIL=return]` is added to `system.nssDatabases.hosts` with priority 501, which prevents lower-priority NSS modules from running unless systemd-resolved is not available.

Quoting from `man nss-resolve`:

> To activate the NSS module, add "resolve [!UNAVAIL=return]" to the line
> starting with "hosts:" in /etc/nsswitch.conf. Specifically, it is
> recommended to place "resolve" early in /etc/nsswitch.conf's "hosts:"
> line. It should be before the "files" entry, since systemd-resolved
> supports /etc/hosts internally, but with caching. To the contrary, it
> should be after "mymachines", to give hostnames given to local VMs and
> containers precedence over names received over DNS. Finally, we
> recommend placing "dns" somewhere after "resolve", to fall back to
> nss-dns if systemd-resolved.service is not available.

Note that the man page (just) recommends "early" and means with this "before the 'files' and 'dns' entries". It does not insist on being first or excluding other modules.

For this reason, libvirt NSS modules should run before the `resolve` module. They should come right next to `mymachines` because both are conceptually very similar -- they resolve local VMs/containers.

Since the data source of the libvirt NSS modules are local plain text files (see source code of the libvirt NSS module), no performance impact is expected form this raise of priorities.

Other NSS modules in NixOS also explicitly set their priority, which is why this change increases consistency.

Fixes #322022

## Description of changes

The NSS modules `libvirt` and `libvirt_guest` now run even when `services.resolved` is enabled.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [x] Tested, as applicable:

---

### DIFF
--- a/nixos/modules/system/boot/systemd.nix
+++ b/nixos/modules/system/boot/systemd.nix
@@ -489,7 +489,7 @@ in
     system.nssModules = [ cfg.package.out ];
     system.nssDatabases = {
       hosts = (mkMerge [
-        (mkOrder 400 ["mymachines"]) # 400 to ensure it comes before resolve (which is mkBefore'd)
+        (mkOrder 400 ["mymachines"]) # 400 to ensure it comes before resolve (which is 501)
         (mkOrder 999 ["myhostname"]) # after files (which is 998), but before regular nss modules
       ]);
       passwd = (mkMerge [
--- a/nixos/modules/virtualisation/libvirtd.nix
+++ b/nixos/modules/virtualisation/libvirtd.nix
@@ -545,9 +545,10 @@ in
     };
 
     system.nssModules = optional (cfg.nss.enable or cfg.nss.enableGuest) cfg.package;
-    system.nssDatabases.hosts = builtins.concatLists [
-      (optional cfg.nss.enable "libvirt")
-      (optional cfg.nss.enableGuest "libvirt_guest")
+    system.nssDatabases.hosts = mkMerge [
+      # ensure that the NSS modules come between mymachines (which is 400) and resolve (which is 501)
+      (mkIf cfg.nss.enable (mkOrder 430 [ "libvirt" ]))
+      (mkIf cfg.nss.enableGuest (mkOrder 432 [ "libvirt_guest" ]))
     ];
   };
 }


jchv (approved):
LGTM. Seems to work, of course, and seems to be closer to the general recommendation anyways. I ran the libvirtd test and it still passes as expected.

flokli:
Please include the test for the NSS lookup into the libvirt VM test, so we can make sure this doesn't regress unnoticed.

flokli:
Fair enough. Running two tests just to that is probably a bit overkill. I assume you'll notice once it regresses ;-)

michaelfranzl:
> Please include the test for the NSS lookup into the libvirt VM test, so we can make sure this doesn't regress unnoticed.

@flokli I did now in https://github.com/NixOS/nixpkgs/pull/323728

FYI @jchv

PR #323056, opened 2024-06-28
cudaPackages: make getOutput work again (again)

SomeoneSerge:
(cherry picked from commit a2954dea377e749d9cff5f71f3815330a5907ee9)

## Description of changes

PROS: Recovers the correct behaviour of `getOutput`, empties the fat symlink farm `$out` output, hopefully reduces the size of saxpy's closure enough for `.#apptainer.gpuChecks.saxpy` to work again, hopefully fixes the issue where packages accidentally use `${cuda_cudart.stubs}/lib/libcuda.so` instead of `addDriverRunpath.driverLink`;

CONS: In the current form, there's no `cuda_nvcc.dev` anymore, so one can't substitute `cuda_nvcc`'s headers without also deploying gcc.

## Build status

- [x] `magma`
- [x] `ucx`, openmpi`
- [x] `torch`
- [x] `blender`
- [x] `opencv`
- [x] `tensorflow`
- [x] `jax`
- [x] `ollama`
- [x] `llama-cpp`
- [x] `python3Packages.faster-whisper`
- [x] `python3Packages.opensfm`
- [x] `faiss`

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/by-name/gp/gpt4all/package.nix
+++ b/pkgs/by-name/gp/gpt4all/package.nix
@@ -47,12 +47,9 @@ stdenv.mkDerivation (finalAttrs: {
   ] ++ lib.optionals cudaSupport (
       with cudaPackages;
       [
-        cuda_cccl.dev
-        cuda_cudart.dev
-        cuda_cudart.lib
-        cuda_cudart.static
-        libcublas.dev
-        libcublas.lib
+        cuda_cccl
+        cuda_cudart
+        libcublas
       ]);
 
   cmakeFlags = [
--- a/pkgs/by-name/ol/ollama/package.nix
+++ b/pkgs/by-name/ol/ollama/package.nix
@@ -101,12 +101,12 @@ let
   };
 
   cudaToolkit = buildEnv {
-    name = "cuda-toolkit";
-    ignoreCollisions = true; # FIXME: find a cleaner way to do this without ignoring collisions
+    name = "cuda-merged";
     paths = [
-      cudaPackages.cudatoolkit
-      cudaPackages.cuda_cudart
-      cudaPackages.cuda_cudart.static
+      (lib.getBin (cudaPackages.cuda_nvcc.__spliced.buildHost or cudaPackages.cuda_nvcc))
+      (lib.getLib cudaPackages.cuda_cudart)
+      (lib.getOutput "static" cudaPackages.cuda_cudart)
+      (lib.getLib cudaPackages.libcublas)
     ];
   };
 
@@ -140,10 +140,6 @@ in
 goBuild ((lib.optionalAttrs enableRocm {
   ROCM_PATH = rocmPath;
   CLBlast_DIR = "${clblast}/lib/cmake/CLBlast";
-}) // (lib.optionalAttrs enableCuda {
-  CUDA_LIB_DIR = "${cudaToolkit}/lib";
-  CUDACXX = "${cudaToolkit}/bin/nvcc";
-  CUDAToolkit_ROOT = cudaToolkit;
 }) // {
   inherit pname version src vendorHash;
 
@@ -151,6 +147,8 @@ goBuild ((lib.optionalAttrs enableRocm {
     cmake
   ] ++ lib.optionals enableRocm [
     rocmPackages.llvm.bintools
+  ] ++ lib.optionals enableCuda [
+    cudaPackages.cuda_nvcc
   ] ++ lib.optionals (enableRocm || enableCuda) [
     makeWrapper
   ] ++ lib.optionals stdenv.isDarwin
@@ -160,6 +158,7 @@ goBuild ((lib.optionalAttrs enableRocm {
     (rocmLibs ++ [ libdrm ])
   ++ lib.optionals enableCuda [
     cudaPackages.cuda_cudart
+    cudaPackages.libcublas
   ] ++ lib.optionals stdenv.isDarwin
     metalFrameworks;
 
--- a/pkgs/development/cuda-modules/cuda/overrides.nix
+++ b/pkgs/development/cuda-modules/cuda/overrides.nix
@@ -166,13 +166,17 @@ filterAndCreateOverrides {
     };
 
   cuda_nvcc =
-    {
-      backendStdenv,
-      cuda_cudart,
-      lib,
-      setupCudaHook,
-    }:
+    { backendStdenv, setupCudaHook }:
     prevAttrs: {
+      # Merge "bin" and "dev" into "out" to avoid circular references
+      outputs = builtins.filter (
+        x:
+        !(builtins.elem x [
+          "dev"
+          "bin"
+        ])
+      ) prevAttrs.outputs;
+
       # Patch the nvcc.profile.
       # Syntax:
       # - `=` for assignment,
@@ -230,8 +234,7 @@ filterAndCreateOverrides {
       };
     };
 
-  cuda_nvprof =
-    { cuda_cupti }: prevAttrs: { buildInputs = prevAttrs.buildInputs ++ [ cuda_cupti.lib ]; };
+  cuda_nvprof = { cuda_cupti }: prevAttrs: { buildInputs = prevAttrs.buildInputs ++ [ cuda_cupti ]; };
 
   cuda_demo_suite =
     {
--- a/pkgs/development/cuda-modules/cuda/overrides.nix
+++ b/pkgs/development/cuda-modules/cuda/overrides.nix
@@ -44,7 +44,7 @@ filterAndCreateOverrides {
     }:
     prevAttrs: {
       buildInputs = prevAttrs.buildInputs ++ [
-        libcublas.lib
+        libcublas
         numactl
         rdma-core
       ];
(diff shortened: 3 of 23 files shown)


Atemu:
In the discourse post you mentioned that one now needs to do `${getBin cuda_nvcc}/bin/nvcc`. If you put the `bin` output first in the list however, `${cuda_nvcc}/bin/nvcc` will continue to work and you also get the `bin` output by default if you `nix-build`.

I also ran into this with ffmpeg and that's the fix. I put `out` last in the list because, in ffmpeg's case, it's entirely empty.

SomeoneSerge:
>  now needs to do ${getBin cuda_nvcc}/bin/nvcc. If you put the bin output first in the list however, ${cuda_nvcc}/bin/nvcc will continue to work and you also get the bin output by default if you nix-build.

The problem is there is now a certain automation in place related to https://github.com/ConnorBaker/cuda-redist-find-features/, and I'd much prefer to address that properly. Also, I actually don't want people to use `${cuda_nvcc}/bin/nvcc`, I want `getExe'` to be the only public interface

pkgs/development/cuda-modules/generic-builders/manifest.nix
SomeoneSerge, line 1: @sameula @ConnorBaker software I cared about (and some which I didn't) builds fine; I'd like to keep this simple: merge as is, wait for the fallout. I'd still appreciate if either of you had a look

pkgs/by-name/gp/gpt4all/package.nix
```diff
@@ -47,12 +47,9 @@ stdenv.mkDerivation (finalAttrs: {
...
       [
-        cuda_cccl.dev
-        cuda_cudart.dev
-        cuda_cudart.lib
-        cuda_cudart.static
-        libcublas.dev
-        libcublas.lib
+        cuda_cccl
+        cuda_cudart
+        libcublas
```
samuela, line 52: I'm confused, quite possibly out of the loop on this: Why are we going back to the coarse outputs? Didn't we recently invest effort into splitting these outputs?

pkgs/by-name/ol/ollama/package.nix
```diff
@@ -101,12 +101,12 @@ let
...
-    name = "cuda-toolkit";
-    ignoreCollisions = true; # FIXME: find a cleaner way to do this without ignoring collisions
+    name = "cuda-merged";
     paths = [
-      cudaPackages.cudatoolkit
-      cudaPackages.cuda_cudart
-      cudaPackages.cuda_cudart.static
+      (lib.getBin (cudaPackages.cuda_nvcc.__spliced.buildHost or cudaPackages.cuda_nvcc))
+      (lib.getLib cudaPackages.cuda_cudart)
+      (lib.getOutput "static" cudaPackages.cuda_cudart)
```
samuela, line 108: what's the advantage of `(lib.getOutput "static" cudaPackages.cuda_cudart)` over `cudaPackages.cuda_cudart.static`?

pkgs/development/cuda-modules/cuda/overrides.nix
```diff
@@ -90,16 +90,16 @@ filterAndCreateOverrides {
...
       brokenConditions = prevAttrs.brokenConditions // {
         "libnvjitlink missing (CUDA >= 12.0)" =
-          !(cudaAtLeast "12.0" -> (libnvjitlink != null && libnvjitlink.lib != null));
+          !(cudaAtLeast "12.0" -> (libnvjitlink != null && libnvjitlink != null));
       };
     };
 
   # TODO(@connorbaker): cuda_cudart.dev depends on crt/host_config.h, which is from
-  # cuda_nvcc.dev. It would be nice to be able to encode that.
+  # (getDev cuda_nvcc. It would be nice to be able to encode that.
```
samuela, line 102:
```suggestion
  # (getDev cuda_nvcc). It would be nice to be able to encode that.
```

pkgs/development/cuda-modules/generic-builders/manifest.nix
```diff
@@ -290,42 +290,23 @@ backendStdenv.mkDerivation (finalAttrs: {
...
+  postPhases = [ "postPatchelf" ];
   postPatchelf = ''
-    mkdir -p "$out"
-    for output in $(getAllOutputNames); do
-      if [[ "$output" != "out" ]]; then
-        ${meta.getExe lndir} "''${!output}" "$out"
-      fi
-    done
+    true
   '';
```
samuela, line 305: can we just remove `postPatchelf`?

pkgs/development/libraries/science/math/faiss/default.nix
```diff
@@ -33,19 +33,15 @@ let
...
-    ];
-  };
+  cudaComponents = with cudaPackages; [
+    cuda_cudart # cuda_runtime.h
+    libcublas
+    libcurand
+    cuda_cccl
+
+    # cuda_profiler_api.h
+    (cudaPackages.cuda_profiler_api or cudaPackages.cuda_nvprof)
```
samuela, line 43: why is an or necessary here? (i know this logic already existed before, but just curious)

pkgs/by-name/gp/gpt4all/package.nix
SomeoneSerge: We invested an effort into splitting the outputs, but then we had a ton of packages use the symlink farm from `$out` which rendered the effort futile: a store path that references `libcudart.so` through the symlink farm ends up pulling the `.a`s in.

What you see in the highlighted diff instead is inputs that `mkDerivation` automatically resolves into the (small) `dev` outputs which _propagate_ all other outputs. The build closure expands, huge (it'll include static archives). The runtime closure, however, retains references to `.lib`s but not to `.dev`s and `.static`s. My expectation: runtime closures shrink.

pkgs/development/cuda-modules/generic-builders/manifest.nix
SomeoneSerge: Empty `postPatchelf` makes `postPhases` error out; no `postPatchelf` means regression for a consumer overriding `postPatchelf` but not re-setting `postPhases`

pkgs/by-name/ol/ollama/package.nix
SomeoneSerge: The advantage of `getX` in general is that if we refactor the generator (`cuda-find-redist-features`) and, say, merge `bin` into `out`, the `getBin` consumers don't have to adapt.

`getOutput "static"` specifically is maybe somewhat useless: it allows us either to merge `static` into `out` (why would we; makes more sense to fallback to lib even), or accommodate removing static libraries altogether...  but then there's a matter of limiting the public interface

pkgs/development/libraries/science/math/faiss/default.nix
SomeoneSerge: One of these didn't exist pre cuda-11-something, and I suspect that the other didn't exist in cuda-11-something-plus-one?

pkgs/by-name/gp/gpt4all/package.nix
samuela: > which rendered the effort futile: a store path that references libcudart.so through the symlink farm ends up pulling the .as in.

This behavior only happens when the symlink farm includes static stuff though, right? Eg if I do `symlinkJoin { cudaPackages.foo.lib }` it will not pull in any static archives IIUC?

> The build closure expands, huge (it'll include static archives). The runtime closure, however, retains references to .libs but not to .devs and .statics. My expectation: runtime closures shrink.

I see. Smaller runtime closures are always welcome.

SomeoneSerge: We could still trade some complexity for smaller _build_ closures, but do we want to?

SomeoneSerge: I'm still concerned about packages potentially retaining a reference to `$dev` though. With torch and jax moving towards JIT/AOT compilation, they just might do that...

Honestly, ideally we'd have a completely virtual output for `nix-support/propagated-build-inputs`, and arrows from it to all other inputs; having `dev` accommodate both headers and `nix-support/propagated-build-inputs` is very constraining

SomeoneSerge:
Checking some closure says before/after the PR:

```nix
with import <nixpkgs> { };

dockerTools.buildLayeredImage {
  name = "torch";
  contents = [
    (python3.withPackages (ps: [
      ps.torch
      ps.torchaudio
      ps.torchvision
    ]))
  ];
}
```

```
❯ nom build --impure -f torch-image.nix -o torch-pre.img
❯ readlink torch-pre.img | xargs du -hs
3.2G    /nix/store/jy09wcj71jh05jncv9z5vhlsc6qsf898-torch.tar.gz
❯ nom build --impure -I nixpkgs=flake:$PWD -f torch-image.nix -o torch-post.img
❯ readlink torch-post.img | xargs du -hs
3.2G    /nix/store/1h7pifhxq6yfir50kwm414j4v1xnlan4-torch.tar.gz
```

I think we're OK

SomeoneSerge:
Smaller closure for jax:

```
❯ nom build --impure -f jax-image.nix -I nixpkgs=flake:github:NixOS/nixpkgs/master -o jax-pre.img
❯ readlink jax-pre.img | xargs du -hs
2.9G    /nix/store/7iamp7aa95q4ddnmkm99pzwik8dgy2i2-torch.tar.gz
❯ nom build --impure -I nixpkgs=flake:$PWD -f jax-image.nix -o jax-post.img
❯ readlink jax-post.img | xargs du -hs
2.3G    /nix/store/q2s38xfsn3wbnrl32x58my55j8xsgp6k-torch.tar.gz


❯ cat jax-image.nix 
with import <nixpkgs> { };

dockerTools.buildLayeredImage {
  name = "torch";
  contents = [
    (python3.withPackages (ps: [
      ps.jax
      ps.jaxlib
    ])) 
  ];
}
❯ cat ~/.config/nixpkgs/config.nix 
{
  cudaSupport = true;
  cudaCapabilities = [
    "7.0"
    "8.0"
    "8.6"
  ];
  cudaEnableForwardCompat = false;
  allowUnfreePredicate =
    p:
    builtins.all (
      license:
      license.free
      || builtins.elem license.shortName [
        "CUDA EULA"
        "cuDNN EULA"
        "NVidia OptiX EULA"
      ]
    ) (if builtins.isList p.meta.license then p.meta.license else [ p.meta.license ]);
  permittedInsecurePackages = [
    "electron-24.8.6"
    "zotero-6.0.27"
  ];
}
```

SomeoneSerge:
@samuela I'll presume you've no objections and I'm ready to clean up if I make a mess:) @ConnorBaker I hope you're doing well with your flat search

pkgs/by-name/ol/ollama/package.nix
```diff
@@ -140,17 +140,15 @@ in
 goBuild ((lib.optionalAttrs enableRocm {
   ROCM_PATH = rocmPath;
   CLBlast_DIR = "${clblast}/lib/cmake/CLBlast";
-}) // (lib.optionalAttrs enableCuda {
-  CUDA_LIB_DIR = "${cudaToolkit}/lib";
```
yshui, line 144: ollama needs this for cuda support.

SomeoneSerge: The CUDA build succeeds without this

PR #323094, opened 2024-06-28
Revert "trivial: make pkgs.writeText always fixed output"

infinisil:
Reverts the last commit of NixOS/nixpkgs#318872

See https://github.com/NixOS/nixpkgs/pull/318872#issuecomment-2196168990


alyssais (approved):

alyssais:
Oh, would you mind explaining a bit more in the commit message itself?  It's good to not have to make people follow links (which may or may not exist in future) to see why a change was reverted.

infinisil:
Good point, done :)

alyssais (approved):

roberth (approved):
This also wouldn't work for dependencies on content-addressed ("floating") outputs, because you only get a placeholder at evaluation time.
The realised CA path only becomes available at build time, so it's not possible to predict the hash part of any floating outputs that are interpolated in the `text`, and that makes predicting the hash of the whole `text` impossible.

Issue #323030, opened 2024-06-27
haskellPackages.arbtt: man pages missing


martinvonwittich:
### Describe the bug

I've added `haskellPackages.arbtt` to my `environment.systemPackages`. `arbtt` itself works fine:

```
martin@desktop ~ % arbtt-dump | head -n1
2024-06-26 11:56:36 (0ms inactive):
```

But the man pages are all missing:

```
martin@desktop ~ % man arbtt-dump
No manual entry for arbtt-dump
martin@desktop ~ % man -k arbtt
arbtt: nothing appropriate.
```

As far as I can tell, upstream provides them in DocBook format: https://github.com/nomeata/arbtt/blob/master/doc/arbtt.xml and also provides a Makefile to build them: https://github.com/nomeata/arbtt/blob/52a00557fff1606152f78ebf7292780430ba884d/doc/Makefile
so this should hopefully be easily fixable.

### Steps To Reproduce

1. Install `haskellPackages.arbtt`.
1. Attempt to read a man page for any of the `arbtt*` binaries.

### Expected behavior
Man pages should be provided.

### Notify maintainers

@rvl

### Metadata
Please run `nix-shell -p nix-info --run "nix-info -m"` and paste the result.

```console
martin@desktop ~ % nix-shell -p nix-info --run "nix-info -m"
 - system: `"x86_64-linux"`
 - host os: `Linux 6.9.3, NixOS, 24.05 (Uakari), 24.05.20240531.63dacb4`
 - multi-user?: `yes`
 - sandbox: `yes`
 - version: `nix-env (Nix) 2.18.2`
 - nixpkgs: `/nix/store/wzx1ba5hqqfa23vfrvqmfmkpj25p37mr-source`
```

---

Add a :+1: [reaction] to [issues you find important].

[reaction]: https://github.blog/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/
[issues you find important]: https://github.com/NixOS/nixpkgs/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions-%2B1-desc

rvl:
Thanks for reporting @martinvonwittich. I have opened PR #323117 to fix it.

PR #321948, opened 2024-06-23
haskellPackages: update stackage and hackage

maralorn:
### This Merge

This PR is the regular merge of the `haskell-updates` branch into `master`.

This branch is being continually built and tested by hydra at https://hydra.nixos.org/jobset/nixpkgs/haskell-updates. You may be able to find an up-to-date Hydra build report at [cdepillabout/nix-haskell-updates-status](https://github.com/cdepillabout/nix-haskell-updates-status).

We roughly aim to merge these `haskell-updates` PRs at least once every two weeks. See the @NixOS/haskell [team calendar](https://cloud.maralorn.de/apps/calendar/p/H6migHmKX7xHoTFa) for who is currently in charge of this branch.

### haskellPackages Workflow Summary

Our workflow is currently described in [`pkgs/development/haskell-modules/HACKING.md`](https://github.com/NixOS/nixpkgs/blob/haskell-updates/pkgs/development/haskell-modules/HACKING.md).

The short version is this:
* We regularly update the Stackage and Hackage pins on `haskell-updates` (normally at the beginning of a merge window).
* The community fixes builds of Haskell packages on that branch.
* We aim at at least one merge of `haskell-updates` into `master` every two weeks.
* We only do the merge if the [`mergeable`](https://hydra.nixos.org/job/nixpkgs/haskell-updates/mergeable) job is succeeding on hydra.
* If a [`maintained`](https://hydra.nixos.org/job/nixpkgs/haskell-updates/maintained) package is still broken at the time of merge, we will only merge if the maintainer has been pinged 7 days in advance. (If you care about a Haskell package, become a maintainer!)

More information about Haskell packages in nixpkgs can be found [in the nixpkgs manual](https://nixos.org/manual/nixpkgs/unstable/#haskell).

---

This is the follow-up to #318363. Come to [#haskell:nixos.org](https://matrix.to/#/#haskell:nixos.org) if you have any questions.

PR #323117, opened 2024-06-28
haskellPackages.arbtt: Build and install manpages

rvl:
This is for merging into the `haskell-updates` branch (PR #321948).

It builds and installs the man pages for arbtt.

They were reported missing in #323030.


maralorn:
As a fellow arbtt user I am doubly grateful.

PR #323133, opened 2024-06-28
quickemu: fix Samba shares

michaeladler:
## Description of changes

This makes it possible to share your home dir with a Windows guest.

* use `qemu_full` for the correct `smbd` path; the fallback path `/usr/sbin/smbd` fails on NixOS [1]
* make the `smbd` binary available to quickemu, enabling it to start qemu with Samba support [2]

[1] https://github.com/qemu/qemu/blob/28b8a57ad63670aa0ce90334523dc552b13b4336/meson.build#L2217
[2] https://github.com/quickemu-project/quickemu/blob/626591bca87e47c545d8fff9a9be0be24f26c3d8/quickemu#L1083

## Things done

- Built on platform(s)
  - [X] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/by-name/qu/quickemu/package.nix
+++ b/pkgs/by-name/qu/quickemu/package.nix
@@ -14,7 +14,7 @@
   pciutils,
   procps,
   python3,
-  qemu,
+  qemu_full,
   socat,
   spice-gtk,
   swtpm,
@@ -42,7 +42,7 @@ let
     pciutils
     procps
     python3
-    qemu
+    qemu_full
     socat
     swtpm
     util-linux
@@ -73,6 +73,7 @@ stdenv.mkDerivation (finalAttrs: {
       -e '/OVMF_CODE_4M.fd/s|ovmfs=(|ovmfs=("${OVMF.firmware}","${OVMF.variables}" |' \
       -e '/cp "''${VARS_IN}" "''${VARS_OUT}"/a chmod +w "''${VARS_OUT}"' \
       -e 's/Icon=.*qemu.svg/Icon=qemu/' \
+      -e 's,\[ -x "\$(command -v smbd)" \],true,' \
       quickemu
   '';
 


flexiondotorg (approved):
Thanks for this change. I am also adding this to the flake in the upstream project.

NickCao:
> make the smbd binary available to quickemu, enabling it to start qemu with Samba support [2]

We shall patch out the command check instead.

michaeladler:
I have patched out the command check as suggested by @NickCao.

pkgs/by-name/qu/quickemu/package.nix
```diff
@@ -73,6 +74,7 @@ stdenv.mkDerivation (finalAttrs: {
       -e '/OVMF_CODE_4M.fd/s|ovmfs=(|ovmfs=("${OVMF.firmware}","${OVMF.variables}" |' \
       -e '/cp "''${VARS_IN}" "''${VARS_OUT}"/a chmod +w "''${VARS_OUT}"' \
       -e 's/Icon=.*qemu.svg/Icon=qemu/' \
+      -e 's,\$(command -v smbd),${samba}/bin/smbd,' \
```
flexiondotorg, line 77: `$[samba}` should be `${pkgs.samba}`

flexiondotorg (changes requested):

flexiondotorg:
It doesn't look like this technique will work; with this patch applied, `quickemu` now fails to start VMs using SPICE because qemu is also looking for `smbd` in a fixed location.

```
qemu-system-x86_64: -netdev user,hostname=windows-server-2022,hostfwd=tcp::22220-:22,smb=/home/martin/Public,id=nic: Could not find '/usr/sbin/smbd', please install it
```

flexiondotorg (approved):
Tested in a local package on NixOS 24.05 and ran an Ubuntu 24.04 guest VM via a Quickemu with SPICE. I was able to connect to smb://10.0.2.4/qemu

pkgs/by-name/qu/quickemu/package.nix
```diff
@@ -73,6 +74,7 @@ stdenv.mkDerivation (finalAttrs: {
       -e '/OVMF_CODE_4M.fd/s|ovmfs=(|ovmfs=("${OVMF.firmware}","${OVMF.variables}" |' \
       -e '/cp "''${VARS_IN}" "''${VARS_OUT}"/a chmod +w "''${VARS_OUT}"' \
       -e 's/Icon=.*qemu.svg/Icon=qemu/' \
+      -e 's,\$(command -v smbd),${pkgs.samba}/bin/smbd,' \
```
NickCao, line 77: Well I think I failed to make it clear, the whole check can be hardcoded to true, since it's qemu's job to find smbd, not quickemu's.

michaeladler: True but if qemu fails to find smbd it will fail to start. Thus it makes sense for quickemu to figure out if qemu supports smbd. In our case we know it does, so yes, the check can be set to true. I've applied your suggestion and rebased the branch.

flexiondotorg (approved):
We can be certain that via Nix smbd is going to be available so I agree with this change for the Nix packaging.

eclairevoyant:
Want to point out this adds an entire gig to the closure, only 1/5 of which is related to samba :upside_down_face:

NickCao:
> Want to point out this adds an entire gig to the closure, only 1/5 of which is related to samba 🙃

But we don't have a qemu_smb attribute cached on hydra.

eclairevoyant:
Any dependencies of something that's cached would also end up cached, barring license restrictions and brokenness and so on.
Anyway, I overrode it in my own config, just pointing it out.

NickCao:
>  Anyway, I overrode it in my own config, just pointing it out.

Ideally we should teach qemu to accept a flag for path to samba.

Issue #308283, opened 2024-05-01
Tracking: generating completion/manual by running compiled program breaks cross compilation


Aleksanaa:
Some programs generates shell completion or manual page with some command line args, e.g. `--completion`, and we usually do this to install completions:

```sh
    installShellCompletion --cmd vrc-get \
      --bash <($out/bin/vrc-get completion bash) \
      --fish <($out/bin/vrc-get completion fish) \
      --zsh <($out/bin/vrc-get completion zsh)
```

However, this process requires executing the resulting binary, which is not possible when running cross compilation.

There should be much more packages out there having this, but they don't do any special handling of cross compilation, so they will very probably fail.

I suggest we add `stdenv.buildPlatform.canExecute stdenv.hostPlatform` toggle, like in other places where we avoid executing in cross compilation, in https://github.com/NixOS/nixpkgs/pull/308055

Another PR https://github.com/NixOS/nixpkgs/pull/307018 built the package targeting `buildPlatform` first, by using `buildPackages.cert-viewer`, which looks good but can cause excessive builds (But if `buildPlatform` is cached, we have a better situation)

sky1e:
https://github.com/NixOS/nixpkgs/blob/1207d1b4ac411eda37b97d91b2d535e20a6cd6fd/pkgs/build-support/setup-hooks/install-shell-files.sh#L92-L95
The `installShellCompletion` command documentation should be updated to not recommend this.

Princemachiavelli:
I think we could also run the targetPlatform binary in a user-mode qemu wrapper? It's very slow but probably enough for just generating the shell auto-completion or man pages.

SFrijters:
Just found this issue while doing some random package fixups after https://github.com/NixOS/nixpkgs/pull/289517 : failing to generate non-empty completions will cause a build failure, so adding a guard for cross-compilation becomes more necessary.

Some extra discussion about how to possibly handle this happened in https://github.com/NixOS/nixpkgs/pull/316749 .

jcaesar:
Noticed this while working on #323161.

Looking around a bit,
```console
$ rg -lU 'installShellCompletion(?s).*<\(\$out/bin' | wc -l
329
```
is definitely more than the handful @SuperSandro2000 predicted [here](https://github.com/NixOS/nixpkgs/pull/289517#pullrequestreview-2092500881). (Looking at a small random sample, it seems most packages don't guard against this at all.)

[...]

PR #289517, opened 2024-02-17
installShellCompletion: add sanity check

SFrijters:
## Description of changes

When implementing https://github.com/NixOS/nixpkgs/pull/288279 I first made a typo (i.e. calling `<($out/bin/espflash completion bash)` instead of `<($out/bin/espflash completions bash)`) and I didn't notice it until I manually checked the generated completion files - any errors when using the `<(...)` syntax are apparently silently ignored.

I tried to add some error handing to it but it looked a bit tricky to fix without unindended breakage. But maybe having this simple sanity check would be nice.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
- For non-Linux: Is sandboxing enabled in `nix.conf`? (See [Nix manual](https://nixos.org/manual/nix/stable/command-ref/conf-file.html))
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:
  - [NixOS test(s)](https://nixos.org/manual/nixos/unstable/index.html#sec-nixos-tests) (look inside [nixos/tests](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests))
  - and/or [package tests](https://nixos.org/manual/nixpkgs/unstable/#sec-package-tests)
  - or, for functions and "core" functionality, tests in [lib/tests](https://github.com/NixOS/nixpkgs/blob/master/lib/tests) or [pkgs/test](https://github.com/NixOS/nixpkgs/blob/master/pkgs/test)
  - made sure NixOS tests are [linked](https://nixos.org/manual/nixpkgs/unstable/#ssec-nixos-tests-linking) to the relevant packages
- [ ] Tested compilation of all packages that depend on this change using `nix-shell -p nixpkgs-review --run "nixpkgs-review rev HEAD"`. Note: all changes have to be committed, also see [nixpkgs-review usage](https://github.com/Mic92/nixpkgs-review#usage)
- [ ] Tested basic functionality of all binary files (usually in `./result/bin/`)
[...]

PR #323161, opened 2024-06-28
rust: make pkgsCross.wasi32.rustPlatform.buildRustPackage work

jcaesar:
## Description of changes

This PR makes it possible to e.g. build and run rust packages like `nix shell .#pkgsCross.wasi32.charasay -c charasay.wasm say moo` (assuming you have `wasm32-wasi` in your emulated systems).

Without this PR, building the cross rustc fails with
```
thread 'main' panicked at src/core/build_steps/compile.rs:370:17:
Target "wasm32-unknown-wasi" does not have a "wasi-root" ```
```

I think this is neat, but it admittedly isn't super useful, as a lot of packages fail to compile due to failing C dependencies or some common dependencies assuming that all wasm is for the web. I've tested the 469 rust packages in by-name, and only 60¹ compile. Even those that do compile might fail at runtime, e.g. due to std::thread not being supported and panicking when attempting to create a thread (like alejandra).

I'm only aware of two mildly useful things that work, namely rq and qrtool. But I think it's nice to have this working.

¹ aarch64-esr-decoder, aba, action-validator, adrs, bunbun, cargo-autoinherit, cargo-bloat, cargo-bump, cargo-shear, cargo-swift, catppuccin-whiskers, catppuccinifier-cli, cdwe, circom, clippy-sarif, cmd-wrapped, commitmsgfmt, diesel-cli-ext, dipc, elf-info, emacs-lsp-booster, esbuild-config, french-numbers, galerio, git-upstream, hadolint-sarif, hvm, hyperlink, icnsify, ifrextractor-rs, keedump, kickstart, ldproxy, little_boxes, ludtwig, lutgen, majima, mcumgr-client, mdsh, ndstrim, obsidian-export, okolors, parallel-disk-usage, pdfrip, polylux2pdfpc, preserves-tools, prettypst, protoc-gen-rust-grpc, qrtool, rq, rs-tftpd, rusti-cal, rusty-diceware, sarif-fmt, shellcheck-sarif, tera-cli, thud, toml-cli, typstfmt, unison-fsmonitor, wit-bindgen

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [x] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/lib/systems/examples.nix
+++ b/lib/systems/examples.nix
@@ -353,6 +353,7 @@ rec {
 
   wasi32 = {
     config = "wasm32-unknown-wasi";
+    rustc.config = "wasm32-wasip1";
     useLLVM = true;
   };
 
--- a/pkgs/build-support/rust/lib/default.nix
+++ b/pkgs/build-support/rust/lib/default.nix
@@ -14,18 +14,23 @@ rec {
   # passed on a build=x86_64/host=aarch64 compilation.
   envVars = let
 
-    ccForBuild = "${pkgsBuildHost.stdenv.cc}/bin/${pkgsBuildHost.stdenv.cc.targetPrefix}cc";
-    cxxForBuild = "${pkgsBuildHost.stdenv.cc}/bin/${pkgsBuildHost.stdenv.cc.targetPrefix}c++";
+    # should match pkgs/development/compilers/rust/rustc.nix
+    prefixForStdenv = stdenv: "${stdenv.cc}/bin/${stdenv.cc.targetPrefix}";
+    ccForStdenv = stdenv: "${prefixForStdenv stdenv}${if (stdenv.cc.isClang or false) then "clang" else "gcc"}";
+    cxxForStdenv = stdenv: "${prefixForStdenv stdenv}${if (stdenv.cc.isClang or false) then "clang++" else "g++"}";
 
-    ccForHost = "${stdenv.cc}/bin/${stdenv.cc.targetPrefix}cc";
-    cxxForHost = "${stdenv.cc}/bin/${stdenv.cc.targetPrefix}c++";
+    ccForBuild = ccForStdenv pkgsBuildHost.stdenv;
+    cxxForBuild = cxxForStdenv pkgsBuildHost.stdenv;
+
+    ccForHost = ccForStdenv stdenv;
+    cxxForHost = cxxForStdenv stdenv;
 
     # Unfortunately we must use the dangerous `pkgsTargetTarget` here
     # because hooks are artificially phase-shifted one slot earlier
     # (they go in nativeBuildInputs, so the hostPlatform looks like
     # a targetPlatform to them).
-    ccForTarget = "${pkgsTargetTarget.stdenv.cc}/bin/${pkgsTargetTarget.stdenv.cc.targetPrefix}cc";
-    cxxForTarget = "${pkgsTargetTarget.stdenv.cc}/bin/${pkgsTargetTarget.stdenv.cc.targetPrefix}c++";
+    ccForTarget = ccForStdenv pkgsTargetTarget.stdenv;
+    cxxForTarget = cxxForStdenv pkgsTargetTarget.stdenv;
 
     rustBuildPlatform = stdenv.buildPlatform.rust.rustcTarget;
     rustBuildPlatformSpec = stdenv.buildPlatform.rust.rustcTargetSpec;
--- a/pkgs/development/compilers/rust/rustc.nix
+++ b/pkgs/development/compilers/rust/rustc.nix
@@ -84,8 +84,8 @@ in stdenv.mkDerivation (finalAttrs: {
   # Reference: https://github.com/rust-lang/rust/blob/master/src/bootstrap/configure.py
   configureFlags = let
     prefixForStdenv = stdenv: "${stdenv.cc}/bin/${stdenv.cc.targetPrefix}";
-    ccPrefixForStdenv = stdenv: "${prefixForStdenv stdenv}${if (stdenv.cc.isClang or false) then "clang" else "cc"}";
-    cxxPrefixForStdenv = stdenv: "${prefixForStdenv stdenv}${if (stdenv.cc.isClang or false) then "clang++" else "c++"}";
+    ccPrefixForStdenv = stdenv: "${prefixForStdenv stdenv}${if (stdenv.cc.isClang or false) then "clang" else "gcc"}";
+    cxxPrefixForStdenv = stdenv: "${prefixForStdenv stdenv}${if (stdenv.cc.isClang or false) then "clang++" else "g++"}";
     setBuild  = "--set=target.${stdenv.buildPlatform.rust.rustcTarget}";
     setHost   = "--set=target.${stdenv.hostPlatform.rust.rustcTarget}";
     setTarget = "--set=target.${stdenv.targetPlatform.rust.rustcTarget}";
--- a/pkgs/development/compilers/rust/rustc.nix
+++ b/pkgs/development/compilers/rust/rustc.nix
@@ -225,6 +225,10 @@ in stdenv.mkDerivation (finalAttrs: {
     substituteInPlace compiler/rustc_target/src/spec/*/*.rs \
       --replace-quiet '"rust-lld"' '"lld"'
 
+    # Figuring out when this should be False is a rustc FIXME. Nixpkgs needs it false.
+    substituteInPlace compiler/rustc_target/src/spec/targets/wasm32_wasi*.rs \
+      --replace LinkSelfContainedDefault::True LinkSelfContainedDefault::False
+
     ${optionalString (!withBundledLLVM) "rm -rf src/llvm"}
 
     # Useful debugging parameter
--- a/pkgs/development/compilers/rust/rustc.nix
+++ b/pkgs/development/compilers/rust/rustc.nix
@@ -167,6 +167,8 @@ in stdenv.mkDerivation (finalAttrs: {
     "${setHost}.musl-root=${pkgsBuildHost.targetPackages.stdenv.cc.libc}"
   ] ++ optionals stdenv.targetPlatform.isMusl [
     "${setTarget}.musl-root=${pkgsBuildTarget.targetPackages.stdenv.cc.libc}"
+  ] ++ optionals stdenv.targetPlatform.isWasi [
+    "${setTarget}.wasi-root=${pkgsBuildTarget.targetPackages.stdenv.cc.libc}"
   ] ++ optionals stdenv.targetPlatform.rust.isNoStdTarget [
     "--disable-docs"
   ] ++ optionals (stdenv.hostPlatform.isDarwin && stdenv.hostPlatform.isx86_64) [
@@ -262,7 +264,7 @@ in stdenv.mkDerivation (finalAttrs: {
   buildInputs = [ openssl ]
     ++ optionals stdenv.hostPlatform.isDarwin [ libiconv Security zlib ]
     ++ optional (!withBundledLLVM) llvmShared.lib
-    ++ optional (useLLVM && !withBundledLLVM) [
+    ++ optional (useLLVM && !withBundledLLVM && !stdenv.targetPlatform.isWasi) [
       llvmPackages.libunwind
       # Hack which is used upstream https://github.com/gentoo/gentoo/blob/master/dev-lang/rust/rust-1.78.0.ebuild#L284
       (runCommandLocal "libunwind-libgcc" {} ''
--- a/pkgs/build-support/bintools-wrapper/default.nix
+++ b/pkgs/build-support/bintools-wrapper/default.nix
@@ -235,6 +235,12 @@ stdenvNoCC.mkDerivation {
         basename=$(basename "$variant")
         wrap $basename ${./ld-wrapper.sh} $variant
       done
+    '' + optionalString targetPlatform.isWasi ''
+      wrap ${targetPrefix}wasm-ld ${./ld-wrapper.sh} $ldPath/${targetPrefix}wasm-ld
+      # clang doesn't properly normalize the target triple[1] before using it as a search path,
+      # and rustc passes wasm32-wasi as `-target`, which gets used as is. Work around.
+      # [1] https://github.com/llvm/llvm-project/blob/ad7aeb0ff58ebd29f68adb85c64e8010639e2a76/clang/lib/Driver/Driver.cpp#L6187
+      ln -s $out/bin/${targetPrefix}wasm-ld $out/bin/wasm32-wasi-wasm-ld
     '';
 
   strictDeps = true;
--- a/pkgs/development/compilers/rust/1_81.nix
+++ b/pkgs/development/compilers/rust/1_81.nix
@@ -32,18 +32,21 @@
 let
   llvmSharedFor =
     pkgSet:
-    pkgSet.llvmPackages_18.libllvm.override (
-      {
-        enableSharedLibraries = true;
-      }
-      // lib.optionalAttrs (stdenv.targetPlatform.useLLVM or false) {
-        # Force LLVM to compile using clang + LLVM libs when targeting pkgsLLVM
-        stdenv = pkgSet.stdenv.override {
-          allowedRequisites = null;
-          cc = pkgSet.llvmPackages_18.clangUseLLVM;
-        };
-      }
-    );
+    if !stdenv.targetPlatform.isWasi then
+      pkgSet.llvmPackages_18.libllvm.override (
+        {
+          enableSharedLibraries = true;
+        }
+        // lib.optionalAttrs (stdenv.targetPlatform.useLLVM or false) {
+          # Force LLVM to compile using clang + LLVM libs when targeting pkgsLLVM
+          stdenv = pkgSet.stdenv.override {
+            allowedRequisites = null;
+            cc = pkgSet.llvmPackages_18.clangUseLLVM;
+          };
+        }
+      )
+    else
+      pkgSet.llvmPackages_18.libllvm;
 in
 import ./default.nix
   {
@@ -59,7 +62,7 @@ import ./default.nix
 
     # Expose llvmPackages used for rustc from rustc via passthru for LTO in Firefox
     llvmPackages =
-      if (stdenv.targetPlatform.useLLVM or false) then
+      if ((stdenv.targetPlatform.useLLVM or false) && !stdenv.targetPlatform.isWasi) then
         callPackage (
           {
             pkgs,
--- a/pkgs/development/libraries/wasilibc/default.nix
+++ b/pkgs/development/libraries/wasilibc/default.nix
@@ -51,6 +51,11 @@ stdenv.mkDerivation {
 
   preFixup = ''
     ln -s $share/share/undefined-symbols.txt $out/lib/wasi.imports
+    # rustc build expects lib at the path where a make install would put them without the SYSROOT_* overrides
+    TARGET_TRIPLE=wasm32-wasip1
+    ln -s $out/lib/ $out/lib/$TARGET_TRIPLE
+    ln -s $dev/include $dev/include/$TARGET_TRIPLE
+    ln -s $share/share $share/share/$TARGET_TRIPLE
   '';
 
   passthru.tests = {


pkgs/development/compilers/rust/rustc.nix
```diff
@@ -145,14 +146,19 @@ in stdenv.mkDerivation (finalAttrs: {
...
     "${setBuild}.musl-root=${pkgsBuildBuild.targetPackages.stdenv.cc.libc}"
   ] ++ optionals stdenv.hostPlatform.isMusl [
     "${setHost}.musl-root=${pkgsBuildHost.targetPackages.stdenv.cc.libc}"
   ] ++ optionals stdenv.targetPlatform.isMusl [
     "${setTarget}.musl-root=${pkgsBuildTarget.targetPackages.stdenv.cc.libc}"
+  ] ++ optionals stdenv.targetPlatform.isWasi [
+    "${setTarget}.wasi-root=${runCommandLocal "wasi-sysroot" {} ''
+      mkdir -p $out/lib
+      ln -s ${pkgsBuildTarget.targetPackages.stdenv.cc.libc}/lib $out/lib/wasm32-wasi
+    ''}"
```
jcaesar, line 161: This workaround is required because wasilibc [overrides](https://github.com/NixOS/nixpkgs/blob/8fa00a61925cf7fe2c302b8d0d9bb5119a7388bb/pkgs/development/libraries/wasilibc/default.nix#L33) the default path/folder structure it's being installed to, but rust expects that structure to exist. The rustc build copies the files from that folder, so this derivation won't get referenced in the closure.

Alternatively this could be done by
1. creating a symlink with the right folder stucture in wasilibc's postInstall (which isn't pretty either and would cause a rebuild of firefox)
2. creating some tmp directory with the right folder structure during rustc's postPatch (but that would mean this workaround must be spread over two different places in this file, which I'd prefer to avoid.

So I think this is the prettiest way of handling that.

pkgs/build-support/rust/lib/default.nix
```diff
@@ -23,7 +23,9 @@ rec {
 
     ccForHost = "${stdenv.cc}/bin/${stdenv.cc.targetPrefix}cc";
     cxxForHost = "${stdenv.cc}/bin/${stdenv.cc.targetPrefix}c++";
-    linkerForHost = if shouldUseLLD stdenv.targetPlatform
+    linkerForHost = if stdenv.targetPlatform.isWasi
+      then "${pkgsBuildHost.llvmPackages.bintools-unwrapped}/bin/${stdenv.cc.targetPrefix}wasm-ld"
```
jcaesar, line 27: The problem here is that
1. wasm-ld isn't included by bintools-wrapper. And even if add that, rustc seems to pass `-flavor wasm` to the linker, even though its path ends in `wasm-ld`.)
2. `ld.lld` from `bintools-unwrapped` would also work, but from what I understand, it works because rustc passes `-flavor wasm`, which is a legacy option. `ld.lld` from `bintools` chokes on that argument.
3. Using  `"${stdenv.cc}/bin/${stdenv.cc.targetPrefix}clang"` as a linker could work (must be "clang", not "cc" as used in `ccForHost`), but I have been unable to figure out how to explain to clang that the linker it should use is not named `wasm-ld` but `wasm32-unknown-wasi-wasm-ld` (it does have that in its search path already).

jcaesar:
I've re-verified that this is the best solution I can come up with. I've added comments on the  @Kranzes, if you only removed your reviewer status because this was still a draft, can I ask you to have a look at some point?

lib/systems/examples.nix
```diff
@@ -353,6 +353,7 @@ rec {
 
   wasi32 = {
     config = "wasm32-unknown-wasi";
+    rustc.config = "wasm32-wasi";
```
alyssais, line 356: The Rust documentation says "undergoing a [rename to wasm32-wasip1](https://github.com/rust-lang/compiler-team/issues/607)".  Should we use the new name?  Also, should we use the threads variant?

pkgs/build-support/rust/lib/default.nix
alyssais: Wouldn't using a wrapped linker be necessary to find dependencies?  What linker does Rust use by default?

pkgs/development/compilers/rust/rustc.nix
alyssais: > creating a symlink with the right folder stucture in wasilibc's postInstall like [so](https://github.com/NixOS/nixpkgs/compare/8aa5aa70d620106e0be3cab003d5cd0c49fae874..d57e347cee6e74743e019b0b30bb2a3230837b96) (which isn't pretty either and would mean I have to touch something that's in the build path of firefox, which I'd like to avoid)

I think this makes more sense.  If Rust expects it here, other things might too.  This is not an urgent change, so there's definitely no reason avoiding a one-off Firefox rebuild should be a consideration.

lib/systems/examples.nix
jcaesar: That's only available from Rust 1.78, nixpkgs is still on 1.77. Btw, wasm32-wasi will go away January 2025. [cf this timeline](https://blog.rust-lang.org/2024/04/09/updates-to-rusts-wasi-targets.html#timeline)

pkgs/build-support/rust/lib/default.nix
jcaesar: > What linker does Rust use by default?

Either `"${stdenv.cc}/bin/${stdenv.cc.targetPrefix}cc"` if that's clang or ld.lld otherwise, both are wrapped, from what I can see.
> Wouldn't using a wrapped linker be necessary to find dependencies?

It's possible that there isn't a single rust package that depends on nixpkgs-built dependencies makes it to linking. Most of them do depend on stuff that doesn't work on wasi, like openssl or gui-related libraries. So I can't answer that question. Why using a wrapped linker breaks certainly needs some investigation… Not sure I have time for that.

pkgs/development/compilers/rust/rustc.nix
jcaesar: If you say "other things too", I should probably create that symlink for the headers, too. Done.

jcaesar:
Thank you for the review.

(The reason I wasn't particularly keen on touching wasilibc is that the following takes ages to build. The floorp failure also happens on master, so that's unrelated.)

---

Result of `nixpkgs-review pr 323161` run on x86_64-linux [1](https://github.com/Mic92/nixpkgs-review)
<details>
  <summary>1 package blacklisted:</summary>
  <ul>
    <li>nixos-install-tools</li>
  </ul>
</details>
<details>
  <summary>3 packages failed to build:</summary>
  <ul>
    <li>floorp</li>
    <li>floorp-unwrapped</li>
    <li>floorp-unwrapped.debug</li>
  </ul>
</details>
<details>
  <summary>32 packages built:</summary>
  <ul>
    <li>adoptopenjdk-icedtea-web</li>
    <li>betterbird</li>
    <li>betterbird-unwrapped</li>
    <li>betterbird-unwrapped.debug</li>
    <li>eyewitness</li>
    <li>firefox</li>
    <li>firefox-beta</li>
    <li>firefox-beta-unwrapped</li>
    <li>firefox-beta-unwrapped.debug</li>
    <li>firefox-beta-unwrapped.symbols</li>
    <li>firefox-devedition</li>
    <li>firefox-devedition-unwrapped</li>
    <li>firefox-devedition-unwrapped.debug</li>
    <li>firefox-devedition-unwrapped.symbols</li>
    <li>firefox-esr</li>
    <li>firefox-esr-115-unwrapped</li>
    <li>firefox-esr-115-unwrapped.debug</li>
    <li>firefox-esr-115-unwrapped.symbols</li>
    <li>firefox-mobile</li>
    <li>firefox-unwrapped</li>
    <li>firefox-unwrapped.debug</li>
    <li>firefox-unwrapped.symbols</li>
    <li>firefoxpwa</li>
    <li>librewolf</li>
    <li>librewolf-unwrapped</li>
    <li>librewolf-unwrapped.debug</li>
    <li>sitespeed-io</li>
    <li>slimerjs</li>
    <li>thunderbird</li>
    <li>thunderbird-unwrapped</li>
    <li>thunderbird-unwrapped.debug</li>
    <li>thunderbird-unwrapped.symbols</li>
  </ul>
</details>

lib/systems/examples.nix
alyssais: … and 1.78 on master

jcaesar: Must have missed that it hit master... Anyway, rebased on master and updated. Thank you.

Problem is that now

pkgs/build-support/rust/lib/default.nix
Mic92: What are the observed symptoms of it not working? Not that if you se NIX_DEBUG than it will also what flags it injects into the build.

Mic92: I would at least leave a comment why we are doing this for now, potentially combined with the error message.

jcaesar: @Mic92 Thank you for the `NIX_DEBUG` hint. Getting rust to print the linker command line in the unwrapped case was a bit more tricky, but with both, I was able to confirm what's going on.

The wrapper script was passing hardening flags (`-z relro -z now`) to wasm-ld, which 

* cause lld to not recognize the `-flavor wasm` arguments
* don't make sense on wasm (`warning: unknown -z value: now`) anyway

I'm not sure what's best here, I don't think hardening makes a lot of sense when compiling to wasm anyway, but I don't know of a way to disable hardening for an entire target. What I opted for is to instead add `hardening_unsupported_flags+=" relro bindnow"` in the wrapper script on wasi.

---
For reference:

The output the linker wrapper is:
```
error: linking with `/nix/store/6plp6qym01fbp05lwa1rh4ym50vnw1ng-wasm32-unknown-wasi-llvm-binutils-wrapper-17.0.6/bin/wasm32-unknown-wasi-wasm-ld` failed: exit status: 1
  |
  = note: LC_ALL="C" PATH="/nix/store/m9dh8g3r6pvbzsxz6b94136l38wmkvgz-wasm32-unknown-wasi-rustc-1.78.0/lib/rustlib/x86_64-unknown-linux-gnu/bin:/nix/store/m9dh8g3r6pvbzsxz6b94136l38wmkvgz-wasm32-unknown-wasi-rustc-1.78.0/lib/rustlib/x86_64-unknown-linux-gnu/bin/self-contained:/nix/store/708hwm
  = note: HARDENING: disabled flags: pie
          HARDENING: Is active (not completely disabled with "all" flag)
          HARDENING: enabling relro
          HARDENING: enabling bindnow
          extra flags before to /nix/store/5gghgw5dkhw3dx862wynrd26sm3k8xbp-llvm-binutils-17.0.6/bin/wasm32-unknown-wasi-wasm-ld:
            -z
            relro
            -z
            now
          original flags to /nix/store/5gghgw5dkhw3dx862wynrd26sm3k8xbp-llvm-binutils-17.0.6/bin/wasm32-unknown-wasi-wasm-ld:
            -flavor
            wasm
            --rsp-quoting=posix
            --export
            __main_void
            -z
            stack-size=1048576
            --stack-first
            --allow-undefined
            --no-demangle
            /nix/store/m9dh8g3r6pvbzsxz6b94136l38wmkvgz-wasm32-unknown-wasi-rustc-1.78.0/lib/rustlib/wasm32-wasip1/lib/self-contained/crt1-command.o
            /build/source/target/wasm32-wasip1/release/deps/rq-9dc42ad07b27356d.rq.2f192189ed26cc41-cgu.0.rcgu.o
            -L
            /build/source/target/wasm32-wasip1/release/deps
            -L
            /build/source/target/release/deps
            -L
            /nix/store/m9dh8g3r6pvbzsxz6b94136l38wmkvgz-wasm32-unknown-wasi-rustc-1.78.0/lib/rustlib/wasm32-wasip1/lib
            -l
            c
            /nix/store/m9dh8g3r6pvbzsxz6b94136l38wmkvgz-wasm32-unknown-wasi-rustc-1.78.0/lib/rustlib/wasm32-wasip1/lib/libcompiler_builtins-dae8cdcfb0b90f99.rlib
            -L
            /nix/store/m9dh8g3r6pvbzsxz6b94136l38wmkvgz-wasm32-unknown-wasi-rustc-1.78.0/lib/rustlib/wasm32-wasip1/lib
            -L
            /nix/store/m9dh8g3r6pvbzsxz6b94136l38wmkvgz-wasm32-unknown-wasi-rustc-1.78.0/lib/rustlib/wasm32-wasip1/lib/self-contained
            -o
            /build/source/target/wasm32-wasip1/release/deps/rq-9dc42ad07b27356d.wasm
            --gc-sections
            -O3
          extra flags after to /nix/store/5gghgw5dkhw3dx862wynrd26sm3k8xbp-llvm-binutils-17.0.6/bin/wasm32-unknown-wasi-wasm-ld:
            -L/nix/store/56c3y4h4ba85fxz0876ld9qd8l3ww2hy-libcxx-static-wasm32-unknown-wasi-17.0.6/lib
            -L/nix/store/02cizbm0w4akb46mapwb0in8jcr10srp-wasilibc-static-wasm32-unknown-wasi-21/lib
          wasm32-unknown-wasi-wasm-ld: error: unknown argument: -flavor
          wasm32-unknown-wasi-wasm-ld: warning: unknown -z value: relro
          wasm32-unknown-wasi-wasm-ld: warning: unknown -z value: now
          wasm32-unknown-wasi-wasm-ld: error: cannot open wasm: No such file or directory
```

This is what I was trying with:

```diff
diff --git a/pkgs/build-support/bintools-wrapper/default.nix b/pkgs/build-support/bintools-wrapper/default.nix
index e7fcf173c602..1c955fdd8d0d 100644
--- a/pkgs/build-support/bintools-wrapper/default.nix
+++ b/pkgs/build-support/bintools-wrapper/default.nix
@@ -239,6 +239,8 @@ stdenvNoCC.mkDerivation {
         basename=$(basename "$variant")
         wrap $basename ${./ld-wrapper.sh} $variant
       done
+    '' + optionalString targetPlatform.isWasi ''
+      wrap ${targetPrefix}wasm-ld ${./ld-wrapper.sh} $ldPath/${targetPrefix}wasm-ld
     '';
 
   strictDeps = true;
diff --git a/pkgs/build-support/rust/lib/default.nix b/pkgs/build-support/rust/lib/default.nix
index b52eecccbb2f..7f66bf01c1e5 100644
--- a/pkgs/build-support/rust/lib/default.nix
+++ b/pkgs/build-support/rust/lib/default.nix
@@ -24,7 +24,7 @@ rec {
     ccForHost = "${stdenv.cc}/bin/${stdenv.cc.targetPrefix}cc";
     cxxForHost = "${stdenv.cc}/bin/${stdenv.cc.targetPrefix}c++";
     linkerForHost = if stdenv.targetPlatform.isWasi
-      then "${pkgsBuildHost.llvmPackages.bintools-unwrapped}/bin/${stdenv.cc.targetPrefix}wasm-ld"
+      then "${pkgsBuildHost.llvmPackages.bintools}/bin/${stdenv.cc.targetPrefix}wasm-ld"
       else if shouldUseLLD stdenv.targetPlatform
       && !stdenv.cc.bintools.isLLVM
       then "${pkgsBuildHost.llvmPackages.bintools}/bin/${stdenv.cc.targetPrefix}ld.lld"
diff --git a/pkgs/development/tools/rq/default.nix b/pkgs/development/tools/rq/default.nix
index f8ed521ad2cc..589d347dbac5 100644
--- a/pkgs/development/tools/rq/default.nix
+++ b/pkgs/development/tools/rq/default.nix
@@ -24,6 +24,9 @@ rustPlatform.buildRustPackage rec {
     rm build.rs
   '';
 
+  env.NIX_DEBUG = "1";
+  env.RUSTC_LOG = "rustc_codegen_ssa::back::link=info";
+
   VERGEN_SEMVER = version;
 
   meta = with lib; {
```

alyssais: Hardening flags are already platform-dependent, so I think you'd just need to add a case for wasmi.

alyssais: By "by default", I mean what linker does a rustc you'd get from rustup use in this case?

jcaesar: Ah, I see. @alyssais For the docker.io/library/rust docker image (which is currently on 1.79) it defaults to `rust-lld -flavor wasm`.

pkgs/build-support/rust/build-rust-package/default.nix
```diff
@@ -45,7 +45,7 @@
 , checkFeatures ? buildFeatures
 , useNextest ? false
 # Enable except on aarch64 pkgsStatic, where we use lld for reasons
-, auditable ? !cargo-auditable.meta.broken && !(stdenv.hostPlatform.isStatic && stdenv.hostPlatform.isAarch64 && !stdenv.hostPlatform.isDarwin)
+, auditable ? !cargo-auditable.meta.broken && (!stdenv.hostPlatform.isStatic || !stdenv.hostPlatform.isAarch64 || stdenv.hostPlatform.isDarwin) && !stdenv.hostPlatform.isWasi
```
alyssais, line 48: I think the motivation for the logic here is more difficult to understand with the ! distributed inside.  Why change that?

Would probably also be clearer to move the wasi case before the complicated one with the parentheses, and maybe start a new line after each top-level `&&`.

jcaesar: I wasn't able to wrap my head around what I wanted to do without putting this into negation normal form (personal preference? double negation is mean). The motivation is to disable auditable in "this *or* that" situation, yet the top level operator is an &&.

I considered doing this

```diff
diff --git a/pkgs/build-support/rust/build-rust-package/default.nix b/pkgs/build-support/rust/build-rust-package/default.nix
index 36e0408cc198..d85820ab536a 100644
--- a/pkgs/build-support/rust/build-rust-package/default.nix
+++ b/pkgs/build-support/rust/build-rust-package/default.nix
@@ -44,9 +44,14 @@
 , buildFeatures ? [ ]
 , checkFeatures ? buildFeatures
 , useNextest ? false
-# Enable except on aarch64 pkgsStatic, where we use lld for reasons
-, auditable ? !cargo-auditable.meta.broken && !(stdenv.hostPlatform.isStatic && stdenv.hostPlatform.isAarch64 && !stdenv.hostPlatform.isDarwin)
-
+, auditable ? !( # rust binaries are made auditable except for any of these:
+    # - cargo-auditable is broken
+    cargo-auditable.meta.broken
+    # - we're on wasi
+    || stdenv.hostPlatform.isWasi
+    # - were on static aarch64 except darwin
+    || (stdenv.hostPlatform.isStatic && stdenv.hostPlatform.isAarch64 && !stdenv.hostPlatform.isDarwin)
+  )
 , depsExtraArgs ? {}

 # Toggles whether a custom sysroot is created when the target is a .json file.
 ```

but not doing this at all just causes a little warning at the end of the build

```
WARNING: target 'wasm32-wasip1' is not supported by 'cargo auditable'!
The build will continue, but no audit data will be injected into the binary.
```

Ignoring that warning and not touching this line at all seems like the lesser evil.

File: pkgs/build-support/rust/lib/default.nix (hunk 2 above)
Comment (jcaesar) [reply pkgs/build-support/rust/lib/default.nix]: (I'm sorry, I force-pushed over this commit again, I mistakenly thought not changing the linker was going to work. But it does have to be set to lld.)

File: pkgs/build-support/rust/build-rust-package/default.nix (hunk 4 above)
Comment (alyssais) [reply pkgs/build-support/rust/build-rust-package/default.nix]: Agreed.

File: pkgs/development/compilers/rust/rustc.nix [hunk 5]
```diff
@@ -19,6 +19,7 @@
 , fastCross
 , lndir
 , makeWrapper
+, runCommandLocal
```
alyssais, line 22: No longer used.

alyssais (changes requested):

pkgs/development/compilers/rust/rustc.nix
jcaesar: Thank you.

alyssais:
I tried building pkgsCross.wasi32.qrtool, but it failed to build wasm32-unknown-wasi-rustc-1.78.0.

```
thread 'main' panicked at src/core/build_steps/compile.rs:400:17:
Target "wasm32-unknown-wasi" does not have a "wasi-root" key in Config.toml
Build completed unsuccessfully in 0:00:13
```

jcaesar:
@alyssais Could you tell me exactly what you're trying to build? The following work for me (should be the same thing):

```
nix build  -vL github:NixOS/nixpkgs/56df69358a7c06f415c33c3cb9543e41b872c220#pkgsCross.wasi32.rq
nix build  -vL github:NixOS/nixpkgs/pull/323161/head#pkgsCross.wasi32.rq
```

(qrtool will fail in installPhase because of #289517 / #308283. Try rq instead if the above does get past the failure you saw in rustc.)

alyssais:
Looks like I might have mixed up which branch I was building — sorry!

pkgs/build-support/rust/lib/default.nix
```diff
@@ -23,8 +23,9 @@ rec {
 
     ccForHost = "${stdenv.cc}/bin/${stdenv.cc.targetPrefix}cc";
     cxxForHost = "${stdenv.cc}/bin/${stdenv.cc.targetPrefix}c++";
-    linkerForHost = if shouldUseLLD stdenv.targetPlatform
-      && !stdenv.cc.bintools.isLLVM
+    # For wasi: rustc passes lld-flavored args unless the compiler is named "clang". So use lld.
```
alyssais, line 26: Do you know which code in rustc does this check?

jcaesar: After some digging. I think I found the [relevant part](https://github.com/rust-lang/rust/blob/adeb79d3f523dfc11a8e9db64ddfc5f546f650bb/compiler/rustc_target/src/spec/mod.rs#L314-L324). So `clang` or `gcc` as names would be recognized, but not `cc` (as set in `ccForHost`). I've revised the comment accordingly.
(Defaulting to lld without being wrapped in a compiler is [here](https://github.com/rust-lang/rust/blob/adeb79d3f523dfc11a8e9db64ddfc5f546f650bb/compiler/rustc_target/src/spec/base/wasm.rs#L90).)

Side notes:
* Since `cc` isn't recognized as anything by rust, the current setup for `ccForHost` is maybe not the best in general.
* [This doccoment](https://doc.rust-lang.org/beta/nightly-rustc/rustc_target/spec/enum.LinkerFlavor.html) has some helpful insights.

alyssais: > * Since `cc` isn't recognized as anything by rust, the current setup for `ccForHost` is maybe not the best in general…?

Yeah, that's what I was wondering.

alyssais: It might also make sense to teach Rust to understand "cc", as it's not exactly a Nixpkgs-specific convention.

jcaesar: I think I read somewhere that not recognizing cc is intentional. Somebody felt that two letters is too little to go on? I might be mistaken, my memory on this is rather hazy. (In any case, this is a finicky part of rustc. Just looking at the amount of "fix" [here](https://github.com/rust-lang/rust/blame/498553fc04f6a3fdc53412320f4e913bc53bc267/compiler/rustc_codegen_ssa/src/back/link.rs#L1315-L1322). I'd rather not touch it.)

In any case, that's not something I intend to fix here. ^^

jcaesar: There are other options?

In case you're thinking of just doing
```nix
    linkerForHost = if stdenv.cc.bintools.isLLVM
      then "${stdenv.cc}/bin/${stdenv.cc.targetPrefix}clang"
      else "${stdenv.cc}/bin/${stdenv.cc.targetPrefix}gcc";
    # resp. doing this for all of cc/cxx/linkerForHost
```

that won't work without further adjustments either:

```
  = note: clang: error: unable to execute command: Executable "wasm-ld" doesn't exist!
```

And I really don't now what those adjustments are, or whether they're benign. I've looked into whether clang can somehow be taught a different linker command before, but couldn't find out anything.

(This would also severely increase the blast radius of this PR…)

jcaesar: I dug around a bit (for options), here's a few things I found:

1. It's unlikely that rustc upstream will accept patches that recognize a `-cc` suffix as a compiler wrapping a linker, reason being that `-cc` might be anything, so it's unclear what arguments can be passed to it
2. cargo-auditable passes linker flags with `-Wl,` unconditionally ([src](https://github.com/rust-secure-code/cargo-auditable/blob/38b37332dfa3601f90608c1565fbd6e63bd6b08a/cargo-auditable/src/rustc_wrapper.rs#L64)) meaning we can't just use a plain linker as `linkerForHost`, it has to be invoked through a compiler as long as we want to use cargo-auditable.
3. 1 and 2 combined mean that some `if stdenv.cc.bintools.isLLVM then clang else gcc` is necessary for `linkerFor…`.
4. Using clang as a linker may work, but it's mechanism for finding wasm-ld is not nice: It searches paths passed with `-B` for `<triple>-wasm-ld` and `wasm-ld` ([src](https://github.com/llvm/llvm-project/blob/ad7aeb0ff58ebd29f68adb85c64e8010639e2a76/clang/lib/Driver/Driver.cpp#L6219-L6226)). However, it doesn't properly normalize the triple, and just uses the `-target wasm32-wasi` argument it gets from rustc as is. Clang itself doesn't think that's a good state and does have a [TODO](https://github.com/llvm/llvm-project/blob/ad7aeb0ff58ebd29f68adb85c64e8010639e2a76/clang/lib/Driver/Driver.cpp#L6187) about this. 
  (This might also be fixable in rustc, which does seem to have logic to not directly pass its own target name. But I havent looked into how that works.)
5. The way the linker gets passed from bintools to clang is:
   * LLVM bintools gets compiled
   * bintools-wrapper wraps it (currently not including wasm-ld) ([src](https://github.com/NixOS/nixpkgs/blob/999be9b3390bbff39f6adb9335f215c2b035744a/pkgs/build-support/bintools-wrapper/default.nix#L242-L245))
   * cc-wrapper symlinks that into ints own bin directory ([src](https://github.com/NixOS/nixpkgs/blob/f8b69469ae63cd405b22bff2a949ba5721a21017/pkgs/build-support/cc-wrapper/default.nix#L371-L374))
   * cc-wrapper's `add-flags.sh` passes the `-B` flag to its own bin dir ([src](https://github.com/NixOS/nixpkgs/blob/d1ae288c68a3a3455e46cbd7c35c0babb3485c79/pkgs/build-support/cc-wrapper/add-flags.sh#L30-L31))
6. 4+5 could in theory be worked around like [this](https://github.com/user-attachments/files/16284747/changerustlinker.patch.txt), but that's a far more terrible hack than what I've tried to do here.

So, the only way I see to get this to work cleanly is to submit a fix to clang for triple normalization, and then submit a PR to nixpkgs staging that switches to `if stdenv.cc.bintools.isLLVM then clang else gcc` (oh the rebuilds, I don't have hardware to test this), and then finally get back to this. I think that's above the level of difficulty I can manage.  

If you've got any advice on how else I could approach this, please do tell.

pkgs/development/compilers/rust/rustc.nix
```diff
@@ -146,14 +146,16 @@ in stdenv.mkDerivation (finalAttrs: {
     # Since fastCross only builds std, it doesn't make sense (and
     # doesn't work) to build a linker.
     "--disable-llvm-bitcode-linker"
-  ] ++ optionals (stdenv.isLinux && !stdenv.targetPlatform.isRedox) [
+  ] ++ optionals (stdenv.isLinux && !stdenv.targetPlatform.isRedox && !stdenv.targetPlatform.isWasi) [
```
Artturin, line 149: Can be dropped after https://github.com/NixOS/nixpkgs/pull/343059

Artturin: It's been merged btw

jcaesar: Thank you for letting me know, but I'm a bit at a loss with the changes to `rust/lib/default.nix`. (I have an idea, but it'll require mass rebuilds. :/)

jcaesar:
Here I'd thought I could finally make progress because found a decent workaround for the mess in the rust build environment setup (namely: do actually use linkers that are either named `*-gcc` or `*-clang` (just like actually passed to rustc). There's a longer rationale in the commit message.)

And already, three new messes have appeared.
* There's now three wasi targets (-p1, p1-threads, -p2), and that isn't being dealt with anywhere yet, at all
* rustc now uses llvm fired clang, which doesn't build for wasi. (I don't fully understand what's going on)
* rustc started to expect wasi targets to use a built-in linker (i.e. placed at `…/lib/rustlib/x86_64-unknown-linux-gnu/bin/gcc-ld/`). Fixed by a single substituteInPlace, but :/. (Placing a linker in that path is possible, but annoying since `x86_64-unknown-linux-gnu` is a symlink to a different derivation thanks to fastCross)

I think I'll let this rest.

PR #323200, opened 2024-06-28
nixos/scion: use RuntimeDirectory instead of StateDirectory

MatthewCroughan:
## Description of changes

This PR doesn't cause any issues for existing deployments if there are any. It's just a re-routing of where the runtime files are placed, to prevent keeping them persistent on the disk.

It was wrong to use `StateDirectory` to keep the `scion-control` and `scion-router` runtime databases on disk for the next run. I observed that doing this means a reboot, or power outage can corrupt the temporary runtime databases for the next service start, leading `scion ping` and other functionality to stop working permanently, since those files are not managed in an atomic manner by the golang code.

This is supposed to be stateless, which is what `RuntimeDirectory` provides, as those dirs get cleaned up on each service start, leaving nothing behind.

You can see in the following `ls` that I've ended up with a 0 byte file `control.trust.db-wal`:

```
[root@t480:/var/lib/private]# ls -lah scion-*
scion-control:
total 178K
drwxr-xr-x 2 scion-control scion   11 Jun 28 18:02 .
drwx------ 9 root          root     9 Jun 28 18:01 ..
-rw-r--r-- 1 scion-control scion  16K Jun 28 18:02 control.beacon.db
-rw-r--r-- 1 scion-control scion  32K Jun 28 18:12 control.beacon.db-shm
-rw-r--r-- 1 scion-control scion  89K Jun 28 18:12 control.beacon.db-wal
-rw-r--r-- 1 scion-control scion  48K Jun 28 18:02 control.path.db
-rw-r--r-- 1 scion-control scion  32K Jun 28 18:13 control.path.db-shm
-rw-r--r-- 1 scion-control scion 974K Jun 28 18:13 control.path.db-wal
-rw-r--r-- 1 scion-control scion 572K Jun 28 18:02 control.trust.db
-rw-r--r-- 1 scion-control scion  32K Jun 28 18:02 control.trust.db-shm
-rw-r--r-- 1 scion-control scion    0 Jun 28 18:02 control.trust.db-wal
```

Which leads to the following failure log on the next service start:

```
Jun 28 17:55:50 t480 scion-control[1960]: 2024-06-28 16:55:50.093142+0000        ERROR               beaconing/writer.go:324        Unable to register segment        {"debug_id": "49dcfbb8", "seg_type": "down", "addr": "17-ffaa:0:1108,CS", "err": {"msg": "resolving SVC address", "cause": "context deadline exceeded", "stacktrace": ["github.com/scionproto/scion/pkg/grpc.(*QUICDialer).Dial github.com/scionproto/scion/pkg/grpc/dialer.go:148", "github.com/scionproto/scion/control/beaconing/grpc.Registrar.RegisterSegment github.com/scionproto/scion/control/beaconing/grpc/register.go:34", "github.com/scionproto/scion/control/beaconing.(*remoteWriter).startSendSegReg.func1 github.com/scionproto/scion/control/beaconing/writer.go:323"]}}
Jun 28 17:55:50 t480 scion-control[1960]: 2024-06-28 16:55:50.093271+0000        ERROR               beaconing/writer.go:110        Unable to register        {"debug_id": "49dcfbb8", "seg_type": "down", "err": {"msg": "no beacons registered", "stacktrace": ["github.com/scionproto/scion/control/beaconing.(*RemoteWriter).Write github.com/scionproto/scion/control/beaconing/writer.go:191", "github.com/scionproto/scion/control/beaconing.(*WriteScheduler).run github.com/scionproto/scion/control/beaconing/writer.go:124", "github.com/scionproto/scion/control/beaconing.(*WriteScheduler).Run github.com/scionproto/scion/control/beaconing/writer.go:109", "github.com/scionproto/scion/private/periodic.(*Runner).onTick github.com/scionproto/scion/private/periodic/periodic.go:213", "github.com/scionproto/scion/private/periodic.(*Runner).runLoop github.com/scionproto/scion/private/periodic/periodic.go:195", "github.com/scionproto/scion/private/periodic.StartWithMetrics.func1 github.com/scionproto/scion/private/periodic/periodic.go:145"], "candidates": 1}}
```

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/nixos/modules/services/networking/scion/scion-control.nix
+++ b/nixos/modules/services/networking/scion/scion-control.nix
@@ -12,13 +12,13 @@ let
       reconnect_to_dispatcher = true;
     };
     beacon_db = {
-      connection = "/var/lib/scion-control/control.beacon.db";
+      connection = "/run/scion-control/control.beacon.db";
     };
     path_db = {
-      connection = "/var/lib/scion-control/control.path.db";
+      connection = "/run/scion-control/control.path.db";
     };
     trust_db = {
-      connection = "/var/lib/scion-control/control.trust.db";
+      connection = "/run/scion-control/control.trust.db";
     };
     log.console = {
       level = "info";
@@ -35,7 +35,7 @@ in
       example = literalExpression ''
         {
           path_db = {
-            connection = "/var/lib/scion-control/control.path.db";
+            connection = "/run/scion-control/control.path.db";
           };
           log.console = {
             level = "info";
@@ -62,7 +62,7 @@ in
         DynamicUser = true;
         Restart = "on-failure";
         BindPaths = [ "/dev/shm:/run/shm" ];
-        StateDirectory = "scion-control";
+        RuntimeDirectory = "scion-control";
       };
     };
   };
--- a/nixos/modules/services/networking/scion/scion-daemon.nix
+++ b/nixos/modules/services/networking/scion/scion-daemon.nix
@@ -12,10 +12,10 @@ let
       reconnect_to_dispatcher = true;
     };
     path_db = {
-      connection = "/var/lib/scion-daemon/sd.path.db";
+      connection = "/run/scion-daemon/sd.path.db";
     };
     trust_db = {
-      connection = "/var/lib/scion-daemon/sd.trust.db";
+      connection = "/run/scion-daemon/sd.trust.db";
     };
     log.console = {
       level = "info";
@@ -32,7 +32,7 @@ in
       example = literalExpression ''
         {
           path_db = {
-            connection = "/var/lib/scion-daemon/sd.path.db";
+            connection = "/run/scion-daemon/sd.path.db";
           };
           log.console = {
             level = "info";
@@ -57,7 +57,7 @@ in
         ExecStart = "${pkgs.scion}/bin/scion-daemon --config ${configFile}";
         Restart = "on-failure";
         DynamicUser = true;
-        StateDirectory = "scion-daemon";
+        RuntimeDirectory = "scion-daemon";
       };
     };
   };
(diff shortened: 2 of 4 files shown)


fricklerhandwerk:
Please add the rationale to the commit message, otherwise it's not clear from the history why that change is made.

MatthewCroughan:
Done @fricklerhandwerk

MatthewCroughan:
According to @matzf this is not supposed to happen, but this kind of error happens in the context of SCIONLab Vagrant deployments all the time, and is documented here https://docs.scionlab.org/content/faq/troubleshooting.html#border-router-fails-to-start

For now, we should keep it as a `RuntimeDirectory` in systemd until it can be demonstrated safe and to not lead to these issues, at which point switching back to `StateDirectory` is simple.

Matzf suggested it could be an issue with service start order, or waiting for interfaces/IPs. But that 0 byte file has me wondering if it's not just an issue in how the upstream code read/writes the files on disk.

matzf:
To clarify, my telediagnosis referred to the somewhat generic error messages in the control service log that you shared (here and in the chat) that indicate some connectivity issues during the SCION path discovery processes. Such failures of the underlying network can occur for any number of reasons; in SCIONLab, starting the router before establishing VPN connection is a particularly frequent case of such a connectivity problem.

Corrupt databases was not at all what I was referring to.
Furthermore, I'm not sure whether a 0 byte write-ahead-log "wal" file is (necessarily) a corrupt database. This could also "just" mean that there have not been any writes to the database since the last checkpoint.
While it's of course not impossible that the files do end up being corrupt, sqlite has an excellent track record for not creating invalid files even on a hard shutdown.
It's quite possible that the information contained in the database is entirely valid, but leads to different behavior which in your case leads to the path discovery not working. Good that you found a workaround for your problem by removing these cache files, even if this might not address the actual cause.

MatthewCroughan:
Continuing conversation in the scion dev channel, there was a mention that there are two sqlite implementations that can be switched between as documented here https://docs.scion.org/en/latest/dev/build.html#options

The two options are "real sqlite from c sources" and "pure golang sqlite". The latter is in use by default and could be less reliable than the C sqlite, leading to the above issues.

So I'm experimenting and testing both implementations to see if this issue still occurs.

PR #256230, opened 2023-09-20
GPU access in the sandbox

SomeoneSerge:
## Description of changes

The PR introduces a way to expose devices and drivers (e.g. GPUs and `/run/opengl-driver/lib`) in the Nix build's sandbox for specially marked derivations. The approach is that described in https://github.com/NixOS/nixpkgs/issues/225912: we introduce a pre-build-hook which would scan the `.drv` being built for the presence of a chosen tag in the `requiredSystemFeatures` attribute and, upon a match, instruct Nix to mount a list of device-related paths. E.g. one could mark the derivation as `mkDerivation { ...; requiredSystemFeatures = [ "cuda" ]; }` to indicate that it would need to use CUDA during the build. Untagged derivations would still observe "pure" environment without any devices. Further, the `requiredSystemFeatures` attribute is special, Nix uses it to determine whether a host is capable of building the derivation, and which remote builders to choose. A host that hasn't been set up with `nix.settings.system-features` (`system-features` in `nix.conf(5)`) would refuse to build the marked derivation

One application of this hook is to implement GPU tests as normal derivations (cf. the torch example in the diff).
Another use could be to use Nix to run ML/DL/NumberCrunching pipelines that utilize GPUs (e.g. https://github.com/SomeoneSerge/pkgs/blob/5bf5ea2b16696e70ade5a68063042dcf2e8f033b/python-packages/by-name/om/omnimotion/package.nix#L270-L291). Similarly, one could write derivations that render things using OpenGL or Vulkan

How to test:

1. Take a NixOS machine with an nvidia GPU and `hardware.opengl.enable = true`.
2. Try building `python3Packages.torch.tests.cudaAvailable`, observe it's refused evaluation
3. Rebulid with `programs.nix-required-mounts.enable = true` (

    <details>

    ```
    # Note: nix-required-sandbox-paths extends system-features,
    # but they won't be merged with the nixos defaults, because they got a different priority
    # -> set manually:
    nix.settings.system-features = [
      "nixos-test"
      "benchmark"
      "big-parallel"
      "kvm"
    ];
    programs.nix-required-sandbox-paths.enable = true;
    ```
   
    </details>
4. Build `python3Packages.torch.gpuChecks.cudaAvailable` successfully
5. Try dropping `cuda` from `gpuChecks.cudaAvailable`'s `requiredSystemFeatures`, and observe the test fail

CC @NixOS/cuda-maintainers 

## Alternatives and related work

[...]

PR #323249, opened 2024-06-28
ollama: remove dependency on`linuxPackages.nvidia_x11`, use `autoAddDriverRunpath` instead

abysssol:
## Description of changes

`linuxPackages.nvidia_x11` is a driver, and isn't intended to be directly linked like a library. `autoAddDriverRunpath` should be able to make ollama work without it.

Due to my pc being broken at the moment, I'm borrowing a weak laptop that isn't capable of building/testing ollama. I therefore require help in testing that ollama-rocm and ollama-cuda actually detect and use the gpu, and don't fall back to cpu. I really appreciate any help with this!

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/by-name/ol/ollama/package.nix
+++ b/pkgs/by-name/ol/ollama/package.nix
@@ -7,6 +7,7 @@
 , overrideCC
 , makeWrapper
 , stdenv
+, addDriverRunpath
 
 , cmake
 , gcc12
@@ -14,8 +15,8 @@
 , libdrm
 , rocmPackages
 , cudaPackages
-, linuxPackages
 , darwin
+, autoAddDriverRunpath
 
 , nixosTests
 , testers
@@ -118,16 +119,16 @@ let
     appleFrameworks.MetalPerformanceShaders
   ];
 
-  runtimeLibs = lib.optionals enableRocm [
-    rocmPath
-  ] ++ lib.optionals enableCuda [
-    linuxPackages.nvidia_x11
-  ];
-  wrapperOptions = builtins.concatStringsSep " " ([
-    "--suffix LD_LIBRARY_PATH : '/run/opengl-driver/lib:${lib.makeLibraryPath runtimeLibs}'"
+  wrapperOptions = [
+    # ollama embeds llama-cpp binaries which actually run the ai models
+    # these llama-cpp binaries are unaffected by the ollama binary's DT_RUNPATH
+    # LD_LIBRARY_PATH is temporarily required to use the gpu
+    # until these llama-cpp binaries can have their runpath patched
+    "--suffix LD_LIBRARY_PATH : '${addDriverRunpath.driverLink}/lib'"
   ] ++ lib.optionals enableRocm [
     "--set-default HIP_PATH '${rocmPath}'"
-  ]);
+  ];
+  wrapperArgs = builtins.concatStringsSep " " wrapperOptions;
 
 
   goBuild =
@@ -153,6 +154,7 @@ goBuild ((lib.optionalAttrs enableRocm {
     rocmPackages.llvm.bintools
   ] ++ lib.optionals (enableRocm || enableCuda) [
     makeWrapper
+    autoAddDriverRunpath
   ] ++ lib.optionals stdenv.isDarwin
     metalFrameworks;
 
@@ -188,8 +190,7 @@ goBuild ((lib.optionalAttrs enableRocm {
     mv "$out/bin/app" "$out/bin/.ollama-app"
   '' + lib.optionalString (enableRocm || enableCuda) ''
     # expose runtime libraries necessary to use the gpu
-    mv "$out/bin/ollama" "$out/bin/.ollama-unwrapped"
-    makeWrapper "$out/bin/.ollama-unwrapped" "$out/bin/ollama" ${wrapperOptions}
+    wrapProgram "$out/bin/ollama" ${wrapperArgs}
   '';
 
   ldflags = [
--- a/nixos/modules/services/misc/ollama.nix
+++ b/nixos/modules/services/misc/ollama.nix
@@ -5,9 +5,6 @@ let
   cfg = config.services.ollama;
   ollamaPackage = cfg.package.override {
     inherit (cfg) acceleration;
-    linuxPackages = config.boot.kernelPackages // {
-      nvidia_x11 = config.hardware.nvidia.package;
-    };
   };
 in
 {


drupol (approved):
Diff lgtm

mrdev023:
I try you PR on my OS after update it

mrdev023:
Strange in  my side

```nix
disabledModules = [ "services/misc/ollama.nix" ];

  imports = let
    nixpkgsPr = builtins.fetchTarball {
      url = "https://github.com/abysssol/nixpkgs/archive/ollama-driver-runpath.tar.gz";
      sha256 = "1ixfvdpi2v4r9yrkvqnfk9whs8lyjhrkdph47bcznh8ak9aipr8p";
    };
  in
  [
    (import "${nixpkgsPr}/nixos/modules/services/misc/ollama.nix")
  ];
```

Seems to use default nvidia pacakge and from my OS. 550 instead of 555. 
I think, i misconfigured the patches

drupol:
I think you need to update `services.ollama.package` perhaps?

abysssol:
@mrdev023 Maybe add something like this to the imports?
``` nix
({ ... }: {
  services.ollama.package = (import nixpkgsPr { }).ollama;
})
```

mrdev023:
> ```nix
> package = (import nixpkgsPr { }).ollama;
> ```

I'm using a flake so i get following errors

```nix
 … while evaluating the attribute 'value'

         at /nix/store/bi5zxc0v5g6ylygdwyqzh280sccg3ykb-source/lib/modules.nix:809:9:

          808|     in warnDeprecation opt //
          809|       { value = builtins.addErrorContext "while evaluating the option `${showOption loc}':" value;
             |         ^
          810|         inherit (res.defsFinal') highestPrio;

       error: attribute 'currentSystem' missing

       at /nix/store/gnnnpxkb36m3cphyjv5hw0is8ki52567-source/pkgs/top-level/impure.nix:17:43:

           16|   # (build, in GNU Autotools parlance) platform.
           17|   localSystem ? { system = args.system or builtins.currentSystem; }
```

abysssol:
``` nix
({ pkgs, ... }: {
  services.ollama.package = (import nixpkgsPr { import (pkgs) system; config.allowUnfree = true; }).ollama;
})
```

mrdev023:
> ```nix
> package = (import nixpkgsPr { import (pkgs) system; config.allowUnfree = true; }).ollama;
> ```
> 
> or if you have `system` in scope:
> 
> ```nix
> package = (import nixpkgsPr { import system; config.allowUnfree = true; }).ollama;
> ```

Thx, i searching how to configure allowUnfree ^^ 

This > package = (import nixpkgsPr { import (pkgs) system; config.allowUnfree = true; }).ollama; works (For building)

mrdev023:
It seems to work in my side

ollama version is 0.1.47

NAME         	ID          	SIZE  	PROCESSOR	UNTIL              
llama3:latest	365c0bd3c000	5.4 GB	100% GPU 	4 minutes from now

abysssol:
good to hear, thank you for testing it. hopefully rocm still works too

mrdev023:
> good to hear, thank you for testing it. hopefully rocm still works too

Thx for you help ^^

pkgs/by-name/ol/ollama/package.nix
```diff
@@ -118,16 +118,12 @@ let
...
 
-  runtimeLibs = lib.optionals enableRocm [
-    rocmPath
-  ] ++ lib.optionals enableCuda [
-    linuxPackages.nvidia_x11
-  ];
-  wrapperOptions = builtins.concatStringsSep " " ([
-    "--suffix LD_LIBRARY_PATH : '/run/opengl-driver/lib:${lib.makeLibraryPath runtimeLibs}'"
+  wrapperOptions = [
+    "--suffix LD_LIBRARY_PATH : '/run/opengl-driver/lib'"
```
SomeoneSerge, line 122: 1. `/run/opengl-driver/lib` should be `${addDriverRunpath.driverLink}/lib` to simplify future migrations
1. Does `ollama` actually need `LD_LIBRARY_PATH`? I see it's a native program, is it not enough tthat `addDriverRunpath` patches dt `DT_RUNPATH`s?

abysssol: > Does `ollama` actually need `LD_LIBRARY_PATH`?

I don't really know. I normally would have tested that locally, but am unable to do so at the moment, so I was more cautious and decided to leave it in case it was necessary.

Since you seem to think it's unnecessary, I just pushed some code removing the wrapper. Hopefully rocm and cuda both still work right, though I will need others to test both to see if they use the gpu.

@mrdev023 Could you test the new version to see if cuda still works right?

mrdev023: I'm prepare my conf to test it

mrdev023: I see this

```
llama3:latest	365c0bd3c000	5.4 GB	100% GPU 	4 minutes from now	
```

But is slow and use CPU and not GPU. I'll try with switch / boot instead of test

mrdev023: I confirm it's a regretion, when i rollback and return back to the global settings 

```nix
services.ollama = {
      enable = true;

      acceleration = if nvidiaEnabled then "cuda" else null;
    };
```

It works well

mrdev023: I don't know if yesterday i'll try to use chat to check the speed.

And i don't know if is a configuration regression or ollama regression because ollama ps display it use GPU at 100% but it's wrong

mrdev023: Currently in nixos-unstable, i'm using ollama version 0.1.45

SomeoneSerge: Ok, I created a reproducer: https://gist.github.com/SomeoneSerge/309d4e26fdf0af328968cfcfac33347e

Basically, `ollama serve` does successfully load `libcuda.so` using `DT_RUNPATH`: https://gist.github.com/SomeoneSerge/309d4e26fdf0af328968cfcfac33347e#file-run-ollama-log-L111

However, it also generates another executable on the fly (maybe during `ollama run`), and that executable is incorrect and fails to load `libcuda.so`:
```
        61:	 search path=/nix/store/hzh8w7jkkqf6wlngvy0rb26bxjzphcpz-cuda-toolkit/lib/glibc-hwcaps/x86-64-v3:...:/nix/store/avkdrmgnn4c0b4sqnlgbn0w464cavh33-cuda_cudart-12.2.140/lib
                		(RUNPATH from file /build/ollama1724294767/runners/cuda_v12/ollama_llama_server)
```

SomeoneSerge: @abysssol let's temporarily recover `LD_LIBRARY_PATH` in the wrapper and add a comment that it's to affect the "runner" executables created by ollama on the fly (or whatever that means, I'm not familiar with ollama)

SomeoneSerge: Btw I'd be grateful if you could help me wrap this into a proper test. Right now I'm confused about ollama's api, with the driver issue addressed it still fails with:

```
llama_new_context_with_model: graph splits = 2
⠋ INFO [main] model loaded | tid="140737352847360" timestamp=1719755151
time=2024-06-30T13:45:51.840Z level=INFO source=server.go:599 msg="llama runner started in 0.91 seconds"
[GIN] 2024/06/30 - 13:45:51 | 200 |  1.071189722s |       127.0.0.1 | POST     "/api/generate"
[GIN] 2024/06/30 - 13:45:51 | 400 |       32.77µs |       127.0.0.1 | POST     "/api/chat"
  File "/build/.attr-1h44fmdnc6dwaj0f8gdbv6z0mrdzkgx9p7zjc2girsm2j4cbi4g2", line 43, in <module>
    r.raise_for_status()
  File "/nix/store/srfswwz4ma22cqjmydjb36l40ki0rmak-python3-3.11.9-env/lib/python3.11/site-packages/requests/models.py", line 1024, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
requests.exceptions.HTTPError: 400 Client Error: Bad Request for url: http://127.0.0.1:11434/api/chat
```

...at https://gist.github.com/SomeoneSerge/8207f77c78e92e787d6d1114b01ae51c#file-llama-2-7b-nix-L108-L117

SomeoneSerge: Maybe they're not really generated on the fly: https://github.com/ollama/ollama/blob/c1218199cfe82eda35f5e4a8031eee28f01ebf75/llm/payload.go#L162. Maybe we could patchelf them during the build if we did it early enough

abysssol: > Maybe they're not really generated on the fly

No, they're not. They are compiled and compressed by `go generate` from ollama's custom c++ code which wraps llama-cpp's code/api, then they're embedded into the ollama binary. Ollama merely extracts them and runs them.

Ollama implements its gpu support by building multiple of these binaries, for example one that supports cuda, another that supports rocm, another that supports cpus that support a certain feature/instruction, and another fallback cpu build that is expected to work on almost all x86 cpus. At runtime, it tries running the binaries with gpu support, and if they fail it falls back to the binaries with cpu support.

This is what makes using the version of llama-cpp packaged in nix so difficult. Ollama has significant amounts of custom code that would need to be patched/spliced into llama-cpp's source, and it might require some modification to make it work right in a different environment (if ollama's custom files have to be moved, paths that refer to them would have to be updated, etc). Ollama also maintains patches for llama-cpp that are specific to particular versions of llama-cpp, so any update of llama-cpp's source would require those patches to be inspected, maybe removed or updated, which is one reason why ollama doesn't use the latest version of llama-cpp.

This applies to #323672: "reus[ing] the prebuilt llama-cpp from Nixpkgs" isn't really possible because ollama doen't use stock llama-cpp in the first place. However, I do think separating the `go generate` step into a separate derivation is still a good idea, and shouldn't be too difficult.

All of this is why I still haven't tried to replace the vendored llama-cpp with nix's llama-cpp package: it's been too intimidating, and seemed to require far more effort than I wanted to put in, especially when ollama works fine without doing so. In the long term, it might make sense to integrate into nix's llama-cpp package, but I would expect significant changes to the llama-cpp package to be required, and using the same version of llama-cpp's source as ollama expects will probably always be necessary.

> Maybe we could patchelf them during the build if we did it early enough

I think the script run by `go generate` would have to be patched to achieve that, though I'm not sure where exactly in the script that would have to go.

If I can find the right place, what would I have to put there? Should I just use patchelf to add `libcuda.so` to the binary's runpath, or is there some way to use whatever autoAddDriverRunpath does on arbitrary binaries?

> Btw I'd be grateful if you could help me wrap this into a proper test

I can't see anything obviously wrong at the moment; I'll look closer when I get the chance.

I never realized that gpu support was possible in the nixos test vms. Good to see that it actually is, though I wonder how that interacts with local test execution. If the local machine doesn't support cuda, do the tests that require cuda simply not run?

SomeoneSerge: > If I can find the right place, what would I have to put there? Should I just use patchelf to add libcuda.so to the binary's runpath, or is there some way to use whatever autoAddDriverRunpath does on arbitrary binaries?

Something like `addDriverRunpath cuda_v12_runner.so` (warning: a purely made up name) after `go generate`

> This applies to https://github.com/NixOS/nixpkgs/issues/323672: "reus[ing] the prebuilt llama-cpp from Nixpkgs" isn't really possible because ollama doen't use stock llama-cpp in the first place

I think this might be feasible after all, but let's not get ahead of ourselves.


For now, let's just recover the `wrapProgram` logic and rely on `LD_LIBRARY_PATH` to handle the embedded runners

SomeoneSerge: > If the local machine doesn't support cuda, do the tests that require cuda simply not run?

Nix would reject the build with a message "error: a 'x86_64-linux' with features {cuda} is required to build ...drv, ..."

> I never realized that gpu support was possible in the nixos test vms.

This one doesn't involve VMs, just simple derivations built in the sandbox. This is experimental and a recent introduction: https://github.com/NixOS/nixpkgs/pull/256230/commits/79a7186f1ce8d94b0c136a7cc7c3e2e31facc794

SomeoneSerge: The `${addDriverRunpath.driverLink}/lib` comment still applies:)

SomeoneSerge: As well as

> add a comment that it's to affect the "runner" executables

SomeoneSerge, line 122:
```suggestion
    # Ollama embeds additional "runner" binaries which do not yet know
    # how to patchelf. They're unaffected by ollama's DT_RUNPATH,
    # so we temporarily use an LD_LIBRARY_PATH wrapper
    "--suffix LD_LIBRARY_PATH : '${addDriverRunpath.driverLink}/lib'"
```

SomeoneSerge:
Thank you @abysssol!

abysssol:
I don't think rocm was tested. I might have to add something like  `"--suffix LD_LIBRARY_PATH : '${rocmPath}'"` to `wrapperOptions` if it doesn't work.

PR #323256, opened 2024-06-28
sile: 0.14.17 → 0.15.5

alerque:
The upstream release is a big change from a pure Lua based CLI to Rust. The Rust CLI still includes a Lua VM that loads the rest of the system and the whole build is still autotools based, so *most* of the current derivation should apply including LuaRock dependencies. The only significant change is the addition of some Rust tooling.

See upstream release notes for [v0.15.0](https://github.com/sile-typesetter/sile/releases/tag/v0.15.0), [v0.15.1](https://github.com/sile-typesetter/sile/releases/tag/v0.15.1), [v0.15.2](https://github.com/sile-typesetter/sile/releases/tag/v0.15.2), [v0.15.3](https://github.com/sile-typesetter/sile/releases/tag/v0.15.3), [v0.15.4](https://github.com/sile-typesetter/sile/releases/tag/v0.15.4), and [v0.15.5](https://github.com/sile-typesetter/sile/releases/tag/v0.15.5).

Most of this is copied from the project's Flake which has adapted to the changes during the development process. The one thing *not* reflected here yet is the ability to change the Lua version by selecting a different package. I wasn't quite sure how to adapt that out of the flake into this derivation.

### DIFF
--- a/pkgs/tools/typesetting/sile/default.nix
+++ b/pkgs/tools/typesetting/sile/default.nix
@@ -1,47 +1,54 @@
-{ lib
-, stdenv
-, darwin
-, fetchurl
-, makeWrapper
-, pkg-config
-, poppler_utils
-, gitMinimal
-, harfbuzz
-, icu
-, fontconfig
-, lua
-, libiconv
-, makeFontsConf
-, gentium
-, runCommand
-, sile
+{
+  lib,
+  stdenv,
+  darwin,
+  fetchurl,
+  makeWrapper,
+  pkg-config,
+  poppler_utils,
+  gitMinimal,
+  harfbuzz,
+  icu,
+  fontconfig,
+  lua,
+  libiconv,
+  makeFontsConf,
+  gentium,
+  runCommand,
+  sile,
 }:
 
 let
-  luaEnv = lua.withPackages(ps: with ps; [
-    cassowary
-    cldr
-    cosmo
-    fluent
-    linenoise
-    loadkit
-    lpeg
-    lua-zlib
-    lua_cliargs
-    luaepnf
-    luaexpat
-    luafilesystem
-    luarepl
-    luasec
-    luasocket
-    luautf8
-    penlight
-    vstruct
-  ] ++ lib.optionals (lib.versionOlder lua.luaversion "5.2") [
-    bit32
-  ] ++ lib.optionals (lib.versionOlder lua.luaversion "5.3") [
-    compat53
-  ]);
+  luaEnv = lua.withPackages (
+    ps:
+    with ps;
+    [
+      cassowary
+      cldr
+      cosmo
+      fluent
+      linenoise
+      loadkit
+      lpeg
+      lua-zlib
+      lua_cliargs
+      luaepnf
+      luaexpat
+      luafilesystem
+      luarepl
+      luasec
+      luasocket
+      luautf8
+      penlight
+      vstruct
+    ]
+    ++ lib.optionals (lib.versionOlder lua.luaversion "5.2") [
+      bit32
+    ]
+    ++ lib.optionals (lib.versionOlder lua.luaversion "5.3") [
+      compat53
+    ]
+  );
 in
 
 stdenv.mkDerivation (finalAttrs: {
@@ -69,9 +76,7 @@ stdenv.mkDerivation (finalAttrs: {
     icu
     fontconfig
     libiconv
-  ]
-  ++ lib.optional stdenv.hostPlatform.isDarwin darwin.apple_sdk.frameworks.AppKit
-  ;
+  ] ++ lib.optional stdenv.hostPlatform.isDarwin darwin.apple_sdk.frameworks.AppKit;
   passthru = {
     # So it will be easier to inspect this environment, in comparison to others
     inherit luaEnv;
@@ -79,20 +84,27 @@ stdenv.mkDerivation (finalAttrs: {
     tests.test = lib.optionalAttrs (!(stdenv.hostPlatform.isDarwin && stdenv.hostPlatform.isAarch64)) (
       runCommand "sile-test"
         {
-          nativeBuildInputs = [ poppler_utils sile ];
+          nativeBuildInputs = [
+            poppler_utils
+            sile
+          ];
           inherit (finalAttrs) FONTCONFIG_FILE;
-        } ''
-        output=$(mktemp -t selfcheck-XXXXXX.pdf)
-        echo "<sile>foo</sile>" | sile -o $output -
-        pdfinfo $output | grep "SILE v${finalAttrs.version}" > $out
-      '');
+        }
+        ''
+          output=$(mktemp -t selfcheck-XXXXXX.pdf)
+          echo "<sile>foo</sile>" | sile -o $output -
+          pdfinfo $output | grep "SILE v${finalAttrs.version}" > $out
+        ''
+    );
   };
 
-  postPatch = ''
-    patchShebangs build-aux/*.sh
-  '' + lib.optionalString stdenv.hostPlatform.isDarwin ''
-    sed -i -e 's|@import AppKit;|#import <AppKit/AppKit.h>|' src/macfonts.m
-  '';
+  postPatch =
+    ''
+      patchShebangs build-aux/*.sh
+    ''
+    + lib.optionalString stdenv.hostPlatform.isDarwin ''
+      sed -i -e 's|@import AppKit;|#import <AppKit/AppKit.h>|' src/macfonts.m
+    '';
 
   NIX_LDFLAGS = lib.optionalString stdenv.hostPlatform.isDarwin "-framework AppKit";
 
@@ -118,7 +130,12 @@ stdenv.mkDerivation (finalAttrs: {
     done
   '';
 
-  outputs = [ "out" "doc" "man" "dev" ];
+  outputs = [
+    "out"
+    "doc"
+    "man"
+    "dev"
+  ];
 
   meta = with lib; {
     description = "Typesetting system";
@@ -135,7 +152,10 @@ stdenv.mkDerivation (finalAttrs: {
     homepage = "https://sile-typesetter.org";
     changelog = "https://github.com/sile-typesetter/sile/raw/v${finalAttrs.version}/CHANGELOG.md";
     platforms = platforms.unix;
-    maintainers = with maintainers; [ doronbehar alerque ];
+    maintainers = with maintainers; [
+      doronbehar
+      alerque
+    ];
     license = licenses.mit;
     mainProgram = "sile";
   };
--- a/pkgs/tools/typesetting/sile/default.nix
+++ b/pkgs/tools/typesetting/sile/default.nix
@@ -2,16 +2,20 @@
   # Nix specific packaging and flake tooling
   darwin,
   fetchurl,
-  gitMinimal,
   lib,
   makeFontsConf,
   makeWrapper,
   runCommand,
+  rustPlatform,
   stdenv,
+  zstd,
 
   # Upstream build time dependencies
+  cargo,
+  jq,
   pkg-config,
   poppler_utils,
+  rustc,
 
   # Upstream run time dependencies
   fontconfig,
@@ -32,7 +36,6 @@ let
     [
       cassowary
       cldr
-      cosmo
       fluent
       linenoise
       loadkit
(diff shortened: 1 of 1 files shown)


pkgs/tools/typesetting/sile/default.nix
```diff
@@ -42,56 +44,69 @@ let
...
+  in stdenv.mkDerivation (finalAttrs: {
   pname = "sile";
-  version = "0.14.17";
+  version = "0.15.4";
 
   src = fetchurl {
-    url = "https://github.com/sile-typesetter/sile/releases/download/v${finalAttrs.version}/sile-${finalAttrs.version}.tar.xz";
-    sha256 = "sha256-f4m+3s7au1FoJQrZ3YDAntKJyOiMPQ11bS0dku4GXgQ=";
+    url = "https://github.com/sile-typesetter/sile/releases/download/v${finalAttrs.version}/sile-${finalAttrs.version}.tar.zst";
+    sha256 = "0n13kv1pabqccjycswcfafhks54l9ck6k8dfkd436jggharkgn1m";
```
eclairevoyant, line 54: generally SRI hash is preferred (you could get this hash from e.g., `nix-hash --to-sri --type sha256 0n13kv1pabqccjycswcfafhks54l9ck6k8dfkd436jggharkgn1m`)

`zstd` is moved in here to facilitate `fetchCargoTarball`

```suggestion
    hash = "sha256-Ndg3s4LvSTNIm66haSZLlBQ9oVOOcc28ZAwvdcOeI1g=";
    nativeBuildInputs = [ zstd ];
```

pkgs/tools/typesetting/sile/default.nix
```diff
@@ -42,56 +44,69 @@ let
...
   nativeBuildInputs = [
-    gitMinimal
-    pkg-config
     makeWrapper
+    pkg-config
+    rustPlatform.cargoSetupHook
+    jq
+    cargo
+    rustc
+    zstd
```
eclairevoyant, line 64:
```suggestion
```

pkgs/tools/typesetting/sile/default.nix
```diff
@@ -42,56 +44,69 @@ let
...
   ];
+
+  # In Nixpkgs, we don't copy the Cargo.lock file from the repo to Nixpkgs'
+  # repo, but we inherit src, and specify a hash (it is a fixed output
+  # derivation). `nix-update` and `nixpkgs-update` should be able to catch that
+  # hash and update it as well when performing updates.
+  cargoDeps = rustPlatform.importCargoLock {
+    lockFile = ./Cargo.lock;
+  };
```
eclairevoyant, line 73: Ironically your code contradicts the comments and vendors `Cargo.lock`. But we don't need to vendor it here.

```suggestion
  cargoDeps = rustPlatform.fetchCargoTarball {
    inherit (finalAttrs) src;
    dontConfigure = true;
    hash = "sha256-FD2otvk92/99AT3BEmfbID8j8MFIicshcocPZalPTHQ=";
  };
```

pkgs/tools/typesetting/sile/Cargo.lock
eclairevoyant, line 1: With the above changes, this file can be deleted.

eclairevoyant (changes requested):
This will also need a rebase on current `master`, due to the change to `meta.description`

pkgs/tools/typesetting/sile/default.nix
alerque: Thanks! It looks like the current invocation for that hash conversion is `nix hash convert --hash-algo sha256`. I'm a bit surprised `nix-prefetch-url` doesn't just output the preferred format though by default.

pkgs/tools/typesetting/sile/default.nix
alerque: That comment came from code copied out of the upstream Flake, but I had to tinker with it so much I lost track of what it was saying. I was so confused by the apparent need to vendor this (I'm glad to find it is not actually needed).

Where did you get this hash?

pkgs/tools/typesetting/sile/default.nix
eclairevoyant: fair point, I've now created a feature req here: https://github.com/NixOS/nix/issues/10990

pkgs/tools/typesetting/sile/default.nix
eclairevoyant: I used the trust-on-first-use method, i.e. set `hash = "";`, build, and get the hash from the error. There might be a more elegant way that I don't know of.

pkgs/tools/typesetting/sile/default.nix
alerque: For some reason this did not work for me, taking it out of here and adding it into the fetchurl hunk above made the build start erroring out about not being able to execute zstd.

pkgs/tools/typesetting/sile/default.nix
alerque: Fair enough. Because I didn't understand it I tried using the hash of the Cargo.lock file from upstream, but that turned out *not* to work and gave me this hash in the error ;-) Given that the lock file is in the upstream source tarball and there is already a checksum on that, it seems like a bit of redundant work here. Whatever this is actually hashing should be derived from the already checksumed source inputs.

pkgs/tools/typesetting/sile/default.nix
```diff
@@ -1,27 +1,29 @@
 { lib
 , stdenv
 , darwin
-, fetchurl
+, fetchurl, zstd
 , makeWrapper
 , pkg-config
-, poppler_utils
```
doronbehar, line 7: I'm a bit confused about all of those formatting and reorder changes that enlarge the diff (for so no real reason so it seems). I also noticed your other reformat PR upstream:

https://github.com/sile-typesetter/sile/pull/2081

What is the reason behind all of this changes? Which expression should lead the formatting standard? BTW we have nowadays a consensus sort of regarding formatting, it is contained in the package [`nixfmt-rfc-style`](https://search.nixos.org/packages?channel=unstable&show=nixfmt-rfc-style&from=0&size=15&sort=relevance&type=packages&query=nixfmt).

Whatever decision you'll go with, I'd suggest to apply a `nixfmt` + consistent argument reordering, and then perform the changes related to the update itself - it will make the PR a bit easier to review. Another option is to discard any of the formatting changes, and minimize the diff as much as possible, which is also legitimate.

pkgs/tools/typesetting/sile/default.nix
```diff
@@ -42,56 +44,67 @@ let
...
-stdenv.mkDerivation (finalAttrs: {
+  in stdenv.mkDerivation (finalAttrs: {
   pname = "sile";
-  version = "0.14.17";
+  version = "0.15.4";
 
   src = fetchurl {
-    url = "https://github.com/sile-typesetter/sile/releases/download/v${finalAttrs.version}/sile-${finalAttrs.version}.tar.xz";
-    sha256 = "sha256-f4m+3s7au1FoJQrZ3YDAntKJyOiMPQ11bS0dku4GXgQ=";
+    url = "https://github.com/sile-typesetter/sile/releases/download/v${finalAttrs.version}/sile-${finalAttrs.version}.tar.zst";
```
doronbehar, line 53: Just out of curiosity, why change to `zst`?

doronbehar (changes requested):

pkgs/tools/typesetting/sile/default.nix
alerque: I've always wanted to use ZST, the reason for XV in the first place was that the GitHub Actions runners we were using to build the source had a version of GNU Autotools that did not support the zstd extension out of the box. Using XZ was less resistance. With more recent LTS of runners available by default that is no longer a limitation, and the generated configure scripts work on older systems as well so there were no remaining reasons *not* to use ZST which has in my mind superior properties.

Then the whole snafu with XZ being backdoored happened and we had folks complaining about the use of XZ. The complaints don't have much merit since the compromise did not affect generated archives, but since I'd always wanted ZST anyway just switching seemed like the easy move. I switched a bunch of my projects based on that logic.

pkgs/tools/typesetting/sile/default.nix
alerque: The re-ordering was because both the package here and the upstream flake had different orders making them hard to diff, and *neither* order made any sense to me. Re-ordering them both grouping by type and then alphabetically brought some rhyme and reason to them. I'll look into `nixfmt`.

doronbehar:
Firendly ping @alerque .

pkgs/tools/typesetting/sile/default.nix
alerque: I have added some comments to the dependency lists marking logical groups, then sorted them alphabetically in each group. I've also reformatted the whole thing with `nixfmt-rfc-style`, and then separately bumped the version. That should make the version bump related changes much easier to review.

Sorry about the delay, life just got a bit chaotic to put it mildly.

doronbehar (commented):
Aish I wish I could have commented about this earlier...

doronbehar:
@alerque there are multiple issues that cause this to not work. I tried to fix some of them myself, but I encountered further issues. Could you please rebase your Nixpkgs fork, and debug the build of it here? My best attempt at fixing the Nixpkgs build encountered:

```
sile> /nix/store/rw1sbc7ya3abg2p2fbkpmijzn6ylna9d-curl-8.10.1-bin/bin/curl -fsSL https://software.sil.org/downloads/r/gentium/GentiumPlus-5.000.zip -o .sources/GentiumPlus-5.000.zip
sile> curl: (6) Could not resolve host: software.sil.org
```

My attempt (draft): https://github.com/NixOS/nixpkgs/pull/353607

pkgs/tools/typesetting/sile/default.nix
```diff
@@ -1,97 +1,125 @@
...
+  fetchurl,
+  lib,
+  makeFontsConf,
+  makeWrapper,
+  runCommand,
+  rustPlatform,
+  stdenv,
+  zstd,
+
+  # Upstream build time dependencies
```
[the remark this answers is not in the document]
doronbehar: > Haha what does "upstream" means in this context :)?

Oh I'm sorry , I thought I was reviewing https://github.com/sile-typesetter/sile/pull/2081 .

doronbehar:
> Aish I wish I could have commented about this earlier...

When I wrote that I meant https://github.com/sile-typesetter/sile/pull/2081 too, for which I have the same comments...

alerque:
> @alerque there are multiple issues that cause this to not work. I tried to fix some of them myself, but I encountered further issues. Could you please rebase your Nixpkgs fork, and debug the build of it here? My best attempt at fixing the Nixpkgs build encountered:
> 
> ```
> sile> /nix/store/rw1sbc7ya3abg2p2fbkpmijzn6ylna9d-curl-8.10.1-bin/bin/curl -fsSL https://software.sil.org/downloads/r/gentium/GentiumPlus-5.000.zip -o .sources/GentiumPlus-5.000.zip
> sile> curl: (6) Could not resolve host: software.sil.org
> ```
> 
> My attempt (draft): #353607

The only issue preventing building your draft is regenerating the manual from scratch. I do not think this should be done in the package derivation anyway. The upstream source releases have the manual PDF already generated. Also it doesn't get installed anywhere by default, so why go out of your way to generate it from scratch? Building the manual requires online access to downloading dozens of fonts—it was never intended for distros to need to do this so the sources are not setup for that to be anything other than a developer target.

Drop the `--with-manual` configure flag and `graphviz-nox` and `ghostscript_headless` dependencies and your derivation should build fine.

If there is a sensible place for a derivation to install PDF manuals for software we can configure the build to install the existing PDF to that location, but I don't even know where that would be.

doronbehar:
> Drop the `--with-manual` configure flag and `graphviz-nox` and `ghostscript_headless` dependencies and your derivation should build fine (at least it did for me locally).

OK OK, I haven't noticed you have even removed the commented `#--with-manual` `./configure` flag...

> If there is a sensible place for a derivation to install PDF manuals for software we can configure the build to install the existing PDF to that location, but I don't even know where that would be.

PDF docs are not that common, but I'd go for `$out/share/doc/sile/manual.pdf`. Is there a `./configure` flag that can tell the generated `Makefile` where to install it?

doronbehar:
OK I also tested my update PR on a Darwin machine, and I was even able to cleanup the `NIX_LDFLAGS` variable and pretty much all the hooks that bothered me long ago, along a few `nativeBuildInputs`. Please see https://github.com/NixOS/nixpkgs/pull/353607#pullrequestreview-2412751869 .

alerque:
Deferring to #353607.

PR #323394, opened 2024-06-29
alfis: 0.8.4 -> 0.8.4-unstable-2024-03-08

superherointj:
alfis: 0.8.4 -> 0.8.4-unstable-2024-03-08

I don't care about this package. Only fixing to lower failed builds at hydra:
- https://hydra.nixos.org/build/263764679/nixlog/1

CC @misuzu

(Not moving to `by-name` because there are extra arguments at top-level.)

### DIFF
--- a/pkgs/applications/blockchains/alfis/default.nix
+++ b/pkgs/applications/blockchains/alfis/default.nix
@@ -14,27 +14,23 @@
 
 rustPlatform.buildRustPackage rec {
   pname = "alfis";
-  version = "0.8.4";
+  version = "0.8.4-unstable-2024-03-08";
 
   src = fetchFromGitHub {
     owner = "Revertron";
     repo = "Alfis";
-    rev = "v${version}";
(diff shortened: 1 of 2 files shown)


pkgs/applications/blockchains/alfis/default.nix
```diff
@@ -1,62 +1,70 @@
...
-  meta = with lib; {
+  meta = {
     description = "Alternative Free Identity System";
     homepage = "https://alfis.name";
-    license = licenses.agpl3Only;
-    maintainers = with maintainers; [ misuzu ];
-    platforms = platforms.unix;
+    license = lib.licenses.agpl3Only;
+    maintainers = with lib.maintainers; [ misuzu ];
+    platforms = lib.platforms.unix;
```
misuzu, line 67: gui version doesn't build on macos.
```suggestion
    platforms = lib.platforms.unix;
    broken = withGui && stdenv.isDarwin;
```

misuzu (approved):

PR #323426, opened 2024-06-29
sageWithDoc: python 3.12 fixes

collares:
## Description of changes

gcc 13.3.0 seems to exploit some undefined behaviour around setjmp/longjmp to do some optimizations that completely break Sage's GAP interface, see https://gcc.gnu.org/bugzilla/show_bug.cgi?id=114872 for the upstream investigation and https://trofi.github.io/posts/312-the-sagemath-saga.html for @trofi's analysis. This only affects Python 3.12.

Unfortunately, the minimal upstream patch at https://github.com/sagemath/sage/pull/37951, which was supposed to address the issue, is not enough to fix the issue on Nixpkgs: GCC still emits code in which an always-null pointer is dereferenced with no null checks. Arch [is also seeing crashes even with the upstream patch applied](https://github.com/sagemath/sage/issues/37026#issuecomment-2131373734). As a stopgap, we compile the relevant file with `-O1` (see `gap-element-crash.patch`)

I also tried other patches similar to https://github.com/sagemath/sage/pull/37951, like having volatile `v0`, `v1` and `v2` variables and populating the relevant variables in the 1-arg, 2-arg and 3-arg cases (see [here](https://github.com/NixOS/nixpkgs/blob/beacd5135b73bebbf6bd6efc9cc6a6839805e415/pkgs/applications/science/math/sage/patches/gap-element-crash.patch)), but that makes crashes *more* frequent, presumably GCC can then easily infer that  `__pyx_t_4`  (a temporary used in Cython-generated code to hold expressions of the form `<GapElement>a[i]`) is always non-null when jumping to the error handling cleanup code without longjmp. It then removes the null check, but it also performs other optimizations which notice that `__pyx_t_4` is always zero. The case where we jump to the error handling code after longjmp, which is what Sage tests, is undefined behaviour and leads to a crash.

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/applications/science/math/sage/patches/gap-element-crash.patch
+++ b/pkgs/applications/science/math/sage/patches/gap-element-crash.patch
@@ -1,0 +1,18 @@
+diff --git a/src/sage/libs/gap/element.pyx b/src/sage/libs/gap/element.pyx
+index b2b0681c052..b2e5807392c 100644
+--- a/src/sage/libs/gap/element.pyx
++++ b/src/sage/libs/gap/element.pyx
+@@ -35,6 +35,13 @@ from sage.structure.coerce cimport coercion_model as cm
+ ### helper functions to construct lists and records ########################
+ ############################################################################
+ 
++
++cdef extern from *:
++    """
++    #pragma GCC optimize("O1")
++    """
++
++
+ cdef Obj make_gap_list(sage_list) except NULL:
+     """
+     Convert Sage lists into Gap lists
--- a/pkgs/applications/science/math/sage/sage-env.nix
+++ b/pkgs/applications/science/math/sage/sage-env.nix
@@ -5,6 +5,7 @@
 , sage-docbuild
 , env-locations
 , gfortran
+, ninja
 , bash
 , coreutils
 , gnused
@@ -63,6 +64,7 @@ let
     "@sage-local@/build"
     pythonEnv
     gfortran # for inline fortran
+    ninja # for inline fortran via numpy.f2py
     stdenv.cc # for cython
     bash
     coreutils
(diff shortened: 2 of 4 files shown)


collares:
@ofborg build sageWithDoc

collares:
@ofborg build sage

pkgs/applications/science/math/sage/patches/gap-element-crash.patch
```diff
@@ -0,0 +1,18 @@
...
+--- a/src/sage/libs/gap/element.pyx
++++ b/src/sage/libs/gap/element.pyx
+@@ -35,6 +35,13 @@ from sage.structure.coerce cimport coercion_model as cm
+ ### helper functions to construct lists and records ########################
+ ############################################################################
+ 
++
++cdef extern from *:
++    """
++    #pragma GCC optimize("O1")
```
collares, line 12: This is a big hack, but Sage is missing a lot of `volatile` variable qualifiers and https://github.com/sagemath/sage/pull/37951 wasn't enough to stop the crashes with Python 3.12 and the newest GCC. I will keep investigating this, but this seems like an acceptable compromise for now. I don't think `clang` implements the exact same optimizations, so a GCC-only fix seems OK.

7c6f434c (approved):
Looks good on reading through

collares:
Thanks for the review!

PR #323473, opened 2024-06-29
move neovim-remote to pkgs/by-name

teto:
- **neovim-remote: move to pkgs/by-name**
- **neovim-gtk: move to pkgs/by-name**


Aleksanaa:
🥱

PR #323493, opened 2024-06-30
pnpm.fetchDeps: Add workspace and custom pnpm config commands support

pyrox0:
## Description of changes
Adds `pnpmWorkspace` and `extraPnpmCommands` options for pnpm.fetchDeps calls

These can be used to either use a specific pnpm workspace for installation, and to run custom pnpm config commands during dependency fetching. Some packages may need a setting such as `dedupe-peer-dependents` set to a non-default value, and so this allows them to set those when needed.

See the discussion in #309100 (starting at https://github.com/NixOS/nixpkgs/pull/309100#discussion_r1658477855) as to why these are useful.

I've also updated the documentation appropriately.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [x] Tested, as applicable:

---

### DIFF
--- a/doc/languages-frameworks/javascript.section.md
+++ b/doc/languages-frameworks/javascript.section.md
@@ -346,11 +346,11 @@ NOTE: It is highly recommended to use a pinned version of pnpm (i.e. `pnpm_8` or
 
 In case you are patching `package.json` or `pnpm-lock.yaml`, make sure to pass `finalAttrs.patches` to the function as well (i.e. `inherit (finalAttrs) patches`.
 
+`pnpm.configHook` supports adding additional `pnpm install` flags via `pnpmInstallFlags` which can be set to a Nix string array.
+
 #### Dealing with `sourceRoot` {#javascript-pnpm-sourceRoot}
 
-NOTE: Nixpkgs pnpm tooling doesn't support building projects with a `pnpm-workspace.yaml`, or building monorepos. It maybe possible to use `pnpm.fetchDeps` for these projects, but it may be hard or impossible to produce a binary from such projects ([an example attempt](https://github.com/NixOS/nixpkgs/pull/290715#issuecomment-2144543728)).
-
-If the pnpm project is in a subdirectory, you can just define `sourceRoot` or `setSourceRoot` for `fetchDeps`. Note, that projects using `pnpm-workspace.yaml` are currently not supported, and will probably not work using this approach.
+If the pnpm project is in a subdirectory, you can just define `sourceRoot` or `setSourceRoot` for `fetchDeps`.
 If `sourceRoot` is different between the parent derivation and `fetchDeps`, you will have to set `pnpmRoot` to effectively be the same location as it is in `fetchDeps`.
 
 Assuming the following directory structure, we can define `sourceRoot` and `pnpmRoot` as follows:
@@ -374,6 +374,55 @@ Assuming the following directory structure, we can define `sourceRoot` and `pnpm
   # by default the working directory is the extracted source
   pnpmRoot = "frontend";
 ```
+
+#### PNPM Workspaces {#javascript-pnpm-workspaces}
+
+If you need to use a PNPM workspace for your project, then set `pnpmWorkspace = "<workspace project name>"` in your `pnpm.fetchDeps` call,
+which will make PNPM only install dependencies for that workspace package.
+
+For example:
+
+```nix
+...
+pnpmWorkspace = "@astrojs/language-server";
+pnpmDeps = pnpm.fetchDeps {
+  inherit (finalAttrs) pnpmWorkspace;
+  ...
+}
+```
+
+The above would make `pnpm.fetchDeps` call only install dependencies for the `@astrojs/language-server` workspace package.
+Note that you do not need to set `sourceRoot` to make this work.
+
+Usually in such cases, you'd want to use `pnpm --filter=$pnpmWorkspace build` to build your project, as `npmHooks.npmBuildHook` probably won't work. A `buildPhase` based on the following example will probably fit most workspace projects:
+
+```nix
+buildPhase = ''
+  runHook preBuild
+
+  pnpm --filter=@astrojs/language-server build
+
+  runHook postBuild
+'';
+```
+
+#### Additional PNPM Commands and settings {#javascript-pnpm-extraCommands}
+
+If you require setting an additional PNPM configuration setting (such as `dedupe-peer-dependents` or similar),
+set `prePnpmInstall` to the right commands to run. For example:
+
+```nix
+prePnpmInstall = ''
+  pnpm config set dedupe-peer-dependants false
+'';
+pnpmDeps = pnpm.fetchDeps {
+  inherit (finalAttrs) prePnpmInstall;
+  ...
+};
+```
+
+In this example, `prePnpmInstall` will be run by both `pnpm.configHook` and by the `pnpm.fetchDeps` builder.
+
 
 ### Yarn {#javascript-yarn}
 
--- a/pkgs/development/tools/pnpm/fetch-deps/default.nix
+++ b/pkgs/development/tools/pnpm/fetch-deps/default.nix
@@ -14,6 +14,8 @@
     {
       hash ? "",
       pname,
+      pnpmWorkspace ? "",
+      prePnpmInstall ? "",
       ...
     }@args:
     let
@@ -29,6 +31,7 @@
             outputHash = "";
             outputHashAlgo = "sha256";
           };
+      installFlags = lib.optionalString (pnpmWorkspace != "")  "--filter=${pnpmWorkspace}";
     in
     stdenvNoCC.mkDerivation (finalAttrs: (
       args'
@@ -58,11 +61,14 @@
           pnpm config set side-effects-cache false
           # As we pin pnpm versions, we don't really care about updates
           pnpm config set update-notifier false
+          # Run any additional pnpm configuration commands that users provide.
+          ${prePnpmInstall}
           # pnpm is going to warn us about using --force
           # --force allows us to fetch all dependencies including ones that aren't meant for our host platform
           pnpm install \
               --force \
               --ignore-scripts \
+              ${installFlags} \
               --frozen-lockfile
 
           runHook postInstall
--- a/pkgs/development/tools/pnpm/fetch-deps/default.nix
+++ b/pkgs/development/tools/pnpm/fetch-deps/default.nix
@@ -62,7 +62,7 @@
           # --force allows us to fetch all dependencies including ones that aren't meant for our host platform
           pnpm install \
               --force \
-              --ignore-script \
+              --ignore-scripts \
               --frozen-lockfile
 
           runHook postInstall
--- a/pkgs/development/tools/pnpm/fetch-deps/default.nix
+++ b/pkgs/development/tools/pnpm/fetch-deps/default.nix
@@ -60,7 +60,10 @@
           pnpm config set update-notifier false
           # pnpm is going to warn us about using --force
           # --force allows us to fetch all dependencies including ones that aren't meant for our host platform
-          pnpm install --frozen-lockfile --ignore-script --force
+          pnpm install \
+              --force \
+              --ignore-script \
+              --frozen-lockfile
 
           runHook postInstall
         '';
--- a/pkgs/development/tools/pnpm/fetch-deps/pnpm-config-hook.sh
+++ b/pkgs/development/tools/pnpm/fetch-deps/pnpm-config-hook.sh
@@ -24,7 +24,11 @@ pnpmConfigHook() {
 
     echo "Installing dependencies"
 
-    pnpm install --offline --frozen-lockfile --ignore-script
+    pnpm install \
+        --offline \
+        --ignore-script \
+        --frozen-lockfile
+
 
     echo "Patching scripts"
 
--- a/pkgs/development/tools/pnpm/fetch-deps/pnpm-config-hook.sh
+++ b/pkgs/development/tools/pnpm/fetch-deps/pnpm-config-hook.sh
@@ -26,7 +26,7 @@ pnpmConfigHook() {
 
     pnpm install \
         --offline \
-        --ignore-script \
+        --ignore-scripts \
         --frozen-lockfile
 
 
--- a/pkgs/development/tools/pnpm/fetch-deps/pnpm-config-hook.sh
+++ b/pkgs/development/tools/pnpm/fetch-deps/pnpm-config-hook.sh
@@ -24,9 +24,15 @@ pnpmConfigHook() {
 
     echo "Installing dependencies"
 
+    if [[ -n "$pnpmWorkspace" ]]; then
+        pnpmInstallFlags+=("--filter=$pnpmWorkspace")
+    fi
+    runHook prePnpmInstall
+
     pnpm install \
         --offline \
         --ignore-scripts \
+        "${pnpmInstallFlags[@]}" \
         --frozen-lockfile
 
 
--- a/pkgs/by-name/as/astro-language-server/package.nix
+++ b/pkgs/by-name/as/astro-language-server/package.nix
@@ -1,0 +1,69 @@
+{
+  lib,
+  stdenv,
+  fetchFromGitHub,
+  pnpm_8,
+  nodejs_22,
+}:
+
+stdenv.mkDerivation (finalAttrs: {
+  pname = "astro-language-server";
+  version = "2.10.0";
+
+  src = fetchFromGitHub {
+    owner = "withastro";
+    repo = "language-tools";
+    rev = "@astrojs/language-server@${finalAttrs.version}";
+    hash = "sha256-WdeQQaC9AVHT+/pXLzaC6MZ6ddHsFSpxoDPHqWvqmiQ=";
+  };
+
+  pnpmDeps = pnpm_8.fetchDeps {
+    inherit (finalAttrs)
+      pname
+      version
+      src
+      pnpmWorkspace
+      prePnpmInstall
+      ;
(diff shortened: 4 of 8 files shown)


doc/languages-frameworks/javascript.section.md
```diff
@@ -375,6 +373,37 @@ Assuming the following directory structure, we can define `sourceRoot` and `pnpm
   pnpmRoot = "frontend";
 ```
 
+#### PNPM Workspaces
```
doronbehar, line 376:
```suggestion
#### PNPM Workspaces {#javascript-pnpm-workspaces}
```

doc/languages-frameworks/javascript.section.md
```diff
@@ -375,6 +373,37 @@ Assuming the following directory structure, we can define `sourceRoot` and `pnpm
...
+...
+pnpmDeps = pnpm.fetchDeps {
+  ...
+  pnpmWorkspace = "@astrojs/language-server";
+}
+
+The above would make the fetchDeps call only install dependencies for the `@astrojs/language-server` workspace package.
+Note that you do not to set `sourceRoot` to make this work.
+
+#### Additional PNPM Commands
```
doronbehar, line 393:
```suggestion
#### Additional PNPM Commands {#javascript-pnpm-extraCommands}
```

doc/languages-frameworks/javascript.section.md
```diff
@@ -375,6 +373,37 @@ Assuming the following directory structure, we can define `sourceRoot` and `pnpm
...
+If you require setting an additional PNPM configuration setting(such as `dedupe-peer-dependents` or similar),
+set `extraPnpmCommands` to the right commands to run. For example:
+```nix
+...
+pnpmDeps = pnpm.fetchDeps {
+  ...
+  extraPnpmCommands = ''
```
doronbehar, line 400: I'd name this `prePnpmInstall`, to make it similar to the `preInstall` hook, which can't be used in this example.

Another option, which can be implemented in addition is an interface like this:

```nix
# ...
pnpmDeps = pnpm.fetchDeps {
  # ...
  extraPnpmConfig = {
    dedupe-peer-dependants = "false";
  };
};
```

This should be easily implemented via something like this (not tested):

```nix
let 
  prePnpmInstall = lib.pipe [
    (lib.mapAttrsToList (n:v: "pnpm config set ${n} ${builtins.toString v}"))
    (lib.concatStringsSep "\n")
  ] extraConfig;
in
  stdenvNoCC.mkDerivation # ...
  # ...
    installPhase = ''
      # ...
    '';
  # ...
```

You could also combine both interfaces by allowing both a string `prePnpmInstall` and a declerative `extraPnpmConfig` attribute set.

doronbehar:
On a 2nd thought, this PR is a good opportunity to test this new interface. I also had motivation to work on this feature for `bash-language-server` and to fix #316908 , so I pushed 2 more commits that will also close #309100 .

doc/languages-frameworks/javascript.section.md
doronbehar: > Another option, which can be implemented in addition is an interface like this:

After playing with this interface a bit I reached the conclusion that it will be a bit confusing to have one declarative argument that only fits the `pnpm.fetchDeps` arguments, and that can't be used as an attribute set argument to the `mkDerivation` builder - because it is an attribute set.

Hence I renamed the string option to `prePnpmInstall` and I'm marking this conversation as resolved.

pkgs/development/tools/pnpm/fetch-deps/default.nix
```diff
@@ -29,6 +31,9 @@
             outputHash = "";
             outputHashAlgo = "sha256";
           };
+      extraPnpmInstallFlags =
+        if pnpmWorkspace != "" then "--filter=${pnpmWorkspace}"
+        else "";
```
getchoo, line 36:
```suggestion
      installFlags = lib.optional (pnpmWorkspace != "")  "--filter=${pnpmWorkspace}";
```
this would also be nice as similar to above, callers would be able to use standard `mkDerivation` variables rather than unique ones found here

pkgs/development/tools/pnpm/fetch-deps/default.nix
```diff
@@ -58,9 +63,11 @@
           pnpm config set side-effects-cache false
           # As we pin pnpm versions, we don't really care about updates
           pnpm config set update-notifier false
+          # Run any additional pnpm configuration commands that users provide.
+          ${prePnpmInstall}
```
getchoo, line 67:
```suggestion
```
see above. this wouldn't be required anymore

pkgs/development/tools/pnpm/fetch-deps/default.nix
```diff
@@ -58,9 +63,11 @@
...
           # pnpm is going to warn us about using --force
           # --force allows us to fetch all dependencies including ones that aren't meant for our host platform
-          pnpm install --frozen-lockfile --ignore-script --force
+          pnpm install --frozen-lockfile --ignore-script --force ${extraPnpmInstallFlags}
```
getchoo, line 70:
```suggestion
          pnpm install --frozen-lockfile --ignore-script --force "$installFlags"
```

pkgs/development/tools/pnpm/fetch-deps/pnpm-config-hook.sh
```diff
@@ -24,7 +24,12 @@ pnpmConfigHook() {
 
     echo "Installing dependencies"
 
-    pnpm install --offline --frozen-lockfile --ignore-script
+    if [[ -n "$pnpmWorkspace" ]]; then
+        extraPnpmInstallFlags="--filter=$pnpmWorkspace"
```
getchoo, line 28:
```suggestion
        pnpmInstallFlags+=("--filter=$pnpmWorkspace")
```

pkgs/development/tools/pnpm/fetch-deps/pnpm-config-hook.sh
```diff
@@ -24,7 +24,12 @@ pnpmConfigHook() {
...
+    fi
+    runHook prePnpmInstall
+
+    pnpm install --offline --frozen-lockfile --ignore-script "$extraPnpmInstallFlags"
```
getchoo, line 32:
```suggestion
    pnpm install --offline --frozen-lockfile --ignore-script "$pnpmInstallFlags"
```

getchoo (changes requested):

pkgs/development/tools/pnpm/fetch-deps/pnpm-config-hook.sh
doronbehar: Nice, thanks!

pkgs/development/tools/pnpm/fetch-deps/default.nix
doronbehar: > this would also be nice as similar to above, this would also be nice as similar to above, callers would be able to use standard `mkDerivation` variables rather than unique ones found here

I agree that `installFlags` is a better variable name (pushed this change), but I'm not sure how that helps someone override this variable with an interface similar to `mkDerivation`...

pkgs/development/tools/pnpm/fetch-deps/default.nix
doronbehar: Another option would be to `runHook prePnpmInstall` here.

pkgs/development/tools/pnpm/fetch-deps/pnpm-config-hook.sh
doronbehar: Nice, this also requires to document this variable and to allow others to use it. I revised the first commit to include this change, and I tested and documented this interface.

pkgs/development/tools/pnpm/fetch-deps/default.nix
```diff
@@ -14,6 +14,8 @@
     {
       hash ? "",
       pname,
+      pnpmWorkspace ? "",
+      prePnpmInstall ? "",
```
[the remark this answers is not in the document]
doronbehar: > i think it would be much better to just use a built in phase here
> 
> i.e., if someone wants to set specific options or run arbitrary commands, they can just use `postConfigure`/`postInstall`/etc from `mkDerivation` rather than this (single) unique phase
> 
> this would require splitting up some of the internal phases here, namely when `HOME`, `store-dir`, and other defaults are applied - they should probably in the `configurePhase`

Hmm you have a point, but still though, even if I'd simply move the `runHook preInstall` line after `HOME` is set, which would allow to perform such an override, I'd have to do something like this:

```nix
  # ...
  pnpmDeps = (pnpm_8.fetchDeps {
    inherit (finalAttrs) pname version src pnpmWorkspace;
    hash = "sha256-n7HTd/rKxJdQKnty5TeOcyvBU9j/EClQ9IHqbBaEwQE=";
  }).overrideAttrs(_: {
    preInstall = finalAttrs.prePnpmInstall;
  });

  prePnpmInstall = ''
    pnpm config set dedupe-peer-dependents false
  '';
  # ...
})
```

Seems much more ugly. Even if we'll just rename the argument to `preInstall`, such that the interface would look like this:

```nix
  # ...
  pnpmDeps = pnpm_8.fetchDeps {
    inherit (finalAttrs) pname version src pnpmWorkspace;
    preInstall = finalAttrs.prePnpmInstall;
    hash = "sha256-n7HTd/rKxJdQKnty5TeOcyvBU9j/EClQ9IHqbBaEwQE=";
  };

  prePnpmInstall = ''
    pnpm config set dedupe-peer-dependents false
  '';
  # ...
})
```

It is still a bit weird IMO how these arguments are named differently, since they serve exactly the same purpose.

If you weren't thinking of either of these interfaces, I didn't fully understand you.

doc/languages-frameworks/javascript.section.md
```diff
@@ -375,6 +373,54 @@ Assuming the following directory structure, we can define `sourceRoot` and `pnpm
...
+pnpmDeps = pnpm.fetchDeps {
+  inherit (finalAttrs) pnpmWorkspace;
+  ...
+}
+```
+
+The above would make the fetchDeps call only install dependencies for the `@astrojs/language-server` workspace package.
+Note that you do not to set `sourceRoot` to make this work.
+
+Usually in such cases, you'd want to use `pnpm --filter=$pnpmWorkspace build` to build your project, as `npmHooks.npmBuildHook` will probably won't work. A `buildPhase` based on the following example will probably fit most workspace projects:
```
SuperSandro2000, line 395: Why can't we detect pnpmWorkspace and set that automatically in the hook?

pkgs/development/tools/pnpm/fetch-deps/pnpm-config-hook.sh
```diff
@@ -24,7 +24,12 @@ pnpmConfigHook() {
 
     echo "Installing dependencies"
 
-    pnpm install --offline --frozen-lockfile --ignore-script
+    if [[ -n "$pnpmWorkspace" ]]; then
+        extraPnpmInstallFlags="--filter=$pnpmWorkspace"
+    fi
```
SuperSandro2000, line 29: If we set this already, why are we doing the buildPhases above by hand? I am confused

dezren39:
@doronbehar would this be something we could use for wrangler in #322573 ? should we wait for this to possibly merge in and then update the pr with these new commands? the pr builds and works fine, but uses quite a bit of disk and needs a bit more testing.

doc/languages-frameworks/javascript.section.md
dezren39, line 395:
```suggestion
Usually in such cases, you'd want to use `pnpm --filter=$pnpmWorkspace build` to build your project, as `npmHooks.npmBuildHook` probably won't work. A `buildPhase` based on the following example will probably fit most workspace projects:
```

pkgs/development/tools/pnpm/fetch-deps/pnpm-config-hook.sh
doronbehar: > If we set this already, why are we doing the buildPhases above by hand? I am confused

You are confused because (surprisingly) we don't have a `pnpm.buildHook`. If the project you try to package doesn't have a specific workspace you are interested in, using `npmHooks.npmBuildHook` probably will do the job. With a specific workspace you pretty much have to use `pnpm --filter=.. build`. Perhaps we should add `pnpm.buildHook`.

doc/languages-frameworks/javascript.section.md
doronbehar: > Why can't we detect pnpmWorkspace and set that automatically in the hook?

See my comment regarding the `pnpm.buildHook` idea below.

thenbe:
Just a heads up that pnpm had a [bug](https://github.com/pnpm/pnpm/issues/6300) where attempting to install deps using `--filter` wasn't behaving as expected in certain cases.

The issue was fixed in pnpm `v9.5.0`. nixos-unstable is currently on `v9.4.0` (an update [PR](https://github.com/NixOS/nixpkgs/pull/325837) to `v9.6.0` is pending).

doronbehar:
Thanks @thenbe for notifying. Now (after https://github.com/NixOS/nixpkgs/pull/325837 is merged) this is rebased. I will merge this after CI is green.

doronbehar:
There are a few unresolved review conversations above, so I let @SuperSandro2000 and @getchoo a chance to reply to them for a few days.

doc/languages-frameworks/javascript.section.md
```diff
@@ -375,6 +375,55 @@ Assuming the following directory structure, we can define `sourceRoot` and `pnpm
...
+...
+pnpmWorkspace = "@astrojs/language-server";
+pnpmDeps = pnpm.fetchDeps {
+  inherit (finalAttrs) pnpmWorkspace;
+  ...
+}
+```
+
+The above would make `pnpm.fetchDeps` call only install dependencies for the `@astrojs/language-server` workspace package.
+Note that you do not to set `sourceRoot` to make this work.
```
gepbird, line 395:
```suggestion
Note that you do not need to set `sourceRoot` to make this work.
```

doc/languages-frameworks/javascript.section.md
```diff
@@ -375,6 +375,55 @@ Assuming the following directory structure, we can define `sourceRoot` and `pnpm
...
+prePnpmInstall = ''
+  pnpm config set dedupe-peer-dependants false
+'';
+pnpmDeps = pnpm.fetchDeps {
+  inherit (finalAttrs) prePnpmInstall;
+  ...
+};
+```
+
+In this example, the will be run by both `pnpm.configHook` and by the `pnpm.fetchDeps` builder.
```
gepbird, line 424:
```suggestion
In this example, `prePnpmInstall` will be run by both `pnpm.configHook` and by the `pnpm.fetchDeps` builder.
```

gepbird (changes requested):

doronbehar:
Thanks for the suggestions @gepbird! Both of them fixed.

pkgs/development/tools/pnpm/fetch-deps/default.nix
```diff
@@ -58,9 +61,11 @@
           pnpm config set side-effects-cache false
           # As we pin pnpm versions, we don't really care about updates
           pnpm config set update-notifier false
+          # Run any additional pnpm configuration commands that users provide.
+          ${prePnpmInstall}
           # pnpm is going to warn us about using --force
           # --force allows us to fetch all dependencies including ones that aren't meant for our host platform
-          pnpm install --frozen-lockfile --ignore-script --force
+          pnpm install --frozen-lockfile --ignore-script --force ${installFlags}
```
niklaskorz, line 68: I know this is technically unrelated to this PR, but there is a typo here, the parameter is called `--ignore-scripts`: https://pnpm.io/cli/install#--ignore-scripts

pkgs/development/tools/pnpm/fetch-deps/pnpm-config-hook.sh
```diff
@@ -24,7 +24,12 @@ pnpmConfigHook() {
 
     echo "Installing dependencies"
 
-    pnpm install --offline --frozen-lockfile --ignore-script
+    if [[ -n "$pnpmWorkspace" ]]; then
+        pnpmInstallFlags+=("--filter=$pnpmWorkspace")
+    fi
+    runHook prePnpmInstall
+
+    pnpm install --offline --frozen-lockfile --ignore-script "${pnpmInstallFlags[@]}"
```
niklaskorz, line 32: Same here (`--ignore-scripts`)

doronbehar:
Thanks @niklaskorz for the review :) I fixed that in separate commits before the main commits of the PR. I will merge this when CI is green.

ck3d:
This PR broke the package bash-language-server, which has been fixed with #333701.

ck3d:
The resulting dependencies for the packages `bash-language-server` and `astro-language-server` do not change if `pnpmWorkspace` is set. The hashes and the content of the dependencies is the same, regardless of the value of `pnpmWorkspace`.
I think this PR needs rework for the parameter `pnpmWorkspace`.

doronbehar:
> The resulting dependencies for the packages `bash-language-server` and `astro-language-server` do not change if `pnpmWorkspace` is set. The hashes and the content of the dependencies is the same, regardless of the value of `pnpmWorkspace`. I think this PR needs rework for the parameter `pnpmWorkspace`.

Thanks for noticing that. I suspect that what me and @pyrox0 experienced back when we initially worked on this PR, was a related to [this upstream pnpm issue](https://github.com/pnpm/pnpm/issues/6300). Perhaps indeed now the workspace support could be simplified. I'm pretty sure though, that for `astro-language-server`, using `pnpm config set dedupe-peer-dependants false` is required, and support for `prePnpmInstall` was added here as well.

Hence for now, I think we'll have to wait for more packages that use pnpm workspaces to aggregate in Nixpkgs, and we can revise this topic.

PR #323549, opened 2024-06-30
lib: add `filterMap`

nrdxp:
## Description of changes

Needed this enough times I thought it deserved to be in the standard library.

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/lib/lists.nix
+++ b/lib/lists.nix
@@ -404,6 +404,45 @@ rec {
   concatMap = builtins.concatMap;
 
   /**
+    Map, removing null values.
+
+    # Type
+
+    ```
+    filterMap :: (a -> (null | b)) -> [a] -> [b]
+    ```
+
+    # Examples
+    :::{.example}
+    ## `lib.lists.filterMap` usage example
+
+    ```nix
+    let f = x:
+      if "udp" == x.proto
+      then x.port
+      else null;
+    in filterMap f [
+      {
+        dns.proto  = "udp";
+        dns.port  = 53;
+      }
+      {
+        http.proto = "tcp";
+        http.port = 80;
+      }
+    ]
+    => [ 53 ]
+    ```
+
+    :::
+  */
+  filterMap = f:
+    let g = acc: x:
+      let y = f x;
+      in if y == null then acc else [y] ++ acc;
+    in builtins.foldl' g [];
+
+  /**
     Flatten the argument into a single list; that is, nested lists are
     spliced into the top-level lists.
 
--- a/lib/lists.nix
+++ b/lib/lists.nix
@@ -406,6 +406,10 @@ rec {
   /**
     Map, removing null values.
 
+    `filterMap` is a convenience function. Use it when `null` is never a
+    valid return item; otherwise, use the more flexible `concatMap`,
+    especially in generic code.
+
     # Type
 
     ```
--- a/lib/default.nix
+++ b/lib/default.nix
@@ -88,7 +88,7 @@ let
       recurseIntoAttrs dontRecurseIntoAttrs cartesianProduct cartesianProductOfSets
       mapCartesianProduct updateManyAttrsByPath listToAttrs hasAttr getAttr isAttrs intersectAttrs removeAttrs;
     inherit (self.lists) singleton forEach map foldr fold foldl foldl' imap0 imap1
-      filter ifilter0 concatMap flatten remove findSingle findFirst any all count
+      filter ifilter0 concatMap filterMap flatten remove findSingle findFirst any all count
       optional optionals toList range replicate partition zipListsWith zipLists
       reverseList listDfs toposort sort sortOn naturalSort compareLists take
       drop sublist last init crossLists unique allUnique intersectLists


lib/lists.nix
```diff
@@ -403,6 +403,45 @@ rec {
...
+    => [ 53 ]
+    ```
+
+    :::
+  */
+  filterMap = f:
+    let g = x: acc:
+      let y = f x;
+      in if isNull y then acc else [y] ++ acc;
+    in lib.foldr g [];
```
roberth, line 443: We're going to be strict in the whole thing anyway, realistically, so we might as well `foldl'` to be quicker and constant-stack.
At that point, it can be compared to a solution based on `concatMap`. That's a builtin, so it should be pretty fast.

lib/lists.nix
```diff
@@ -403,6 +403,45 @@ rec {
...
+    ]
+    => [ 53 ]
+    ```
+
+    :::
+  */
+  filterMap = f:
+    let g = x: acc:
+      let y = f x;
+      in if isNull y then acc else [y] ++ acc;
```
roberth, line 442:
```suggestion
      in if y == null then acc else [y] ++ acc;
```

Same number of AST nodes, but simpler operations.

```
$ bench -- 'nix eval --impure --expr "let lib = import ./lib; inherit (lib) foldl'\'' range; inherit (builtins) isNull; in foldl'\'' (a: b: a || isNull b) false (range 0 10000000)"';                                   bench -- 'nix eval --impure --expr "let lib = import ./lib; inherit (lib) foldl'\'' range; inherit (builtins) isNull; in foldl'\'' (a: b: a || b == null) false (range 0 10000000)"'
benchmarking nix eval --impure --expr "let lib = import ./lib; inherit (lib) foldl' range; inherit (builtins) isNull; in foldl' (a: b: a || isNull b) false (range 0 10000000)"
time                 2.014 s    (NaN s .. 2.022 s)
                     1.000 R²   (1.000 R² .. 1.000 R²)
mean                 2.010 s    (2.008 s .. 2.012 s)
std dev              2.497 ms   (1.883 ms .. 2.958 ms)
variance introduced by outliers: 19% (moderately inflated)

benchmarking nix eval --impure --expr "let lib = import ./lib; inherit (lib) foldl' range; inherit (builtins) isNull; in foldl' (a: b: a || b == null) false (range 0 10000000)"
time                 1.938 s    (1.915 s .. 1.952 s)
                     1.000 R²   (1.000 R² .. 1.000 R²)
mean                 1.941 s    (1.938 s .. 1.944 s)
std dev              3.180 ms   (1.610 ms .. 4.395 ms)
variance introduced by outliers: 19% (moderately inflated)

```

roberth (commented):
I have multiple reasons not to want this function, but these can be mitigated and I believe the balance is in favor of having it.

I prefer to always use `concatMap` because then I can avoid thinking about the union type `b | null`, which is something I like to avoid because union types don't compose losslessly like sum types or tagged unions.
It's just a _slightly_ more widely applicable solution because of this.

I'd also like to keep the `lib` interface [small](https://wiki.haskell.org/Fairbairn_threshold), but I believe in this case the downside of "added noise" is offset by users simply finding *a* solution. We can also use its documentation to help them learn about `concatMap`.

So I think the mitigation here is to immediately offer `concatMap` as an alternative for use in generic code. For example:

`filterMap` is a convenience function that is less capable than `concatMap`; only use it if you know that `null` will never need to be returned as an item. This might not be knowable in generic code, in which case it is better to use `concatMap`.

lib/lists.nix
roberth: Also strictness is not an issue. The type tag is compared first, and there won't be any need to evaluate any nested values, because those are only evaluated if the type tag matches, and if it matches, it is `null`, which has no nested values.

lib/lists.nix
```diff
@@ -403,6 +403,45 @@ rec {
   */
   concatMap = builtins.concatMap;
 
+  /**
+    Map, removing null values.
+
+    # Type
+
+    ```
+    filterMap :: (a -> [b]) -> [a] -> [b]
```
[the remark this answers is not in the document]
nrdxp: indeed, I copied it as a skeleton for the doc and forgot to update. Forced pushed a fix.

lib/lists.nix
nrdxp: are you refering to `lib.foldl'` or `builtins.foldl'`. I assume the former?

roberth: Good question. I think in this case we don't need the `lib.foldl'` improvement, because we don't need to "evaluating `[]` first" does nothing.

Normally it is a good idea to use `lib` instead of `builtins`, but in this case we don't need it, and we can squeeze a tiny bit of extra performance. It really only makes sense for functions in a heavily used and/or performance critical library like `lib`  to even consider to use the `builtins` function instead of the usually better `lib`.

lib/lists.nix
```diff
@@ -403,6 +403,45 @@ rec {
   */
   concatMap = builtins.concatMap;
 
+  /**
+    Map, removing null values.
+
+    # Type
+
+    ```
+    filterMap :: (a -> b) -> [a] -> [b]
```
roberth, line 412:
```suggestion
    filterMap :: (a -> (null | b)) -> [a] -> [b]
```

I think we can reasonably convey a bit more info here.

We don't have a real type language, but `|` is for union types in TypeScript and Scala, which is what this is meant to refer to. Other associations with `|` tend to also work.

lib/lists.nix
```diff
@@ -403,6 +403,49 @@ rec {
...
+      {
+        http.proto = "tcp";
+        http.port = 80;
+      }
+    ]
+    => [ 53 ]
+    ```
+
+    :::
+  */
```
roberth, line 442:
```suggestion

    # Strictness

    If all of your list is used, and it has no recursion between items, ignore this section.

    Like `filterAttrs`, this function is strict in the items, which means that unlike e.g. `map`, `filterMap` can only return after evaluating each item.
    This can be a problem when accessing list items by index or with `head`, in situations where not all items are well defined, or when a list item is defined in terms of another list item in a strict manner.

  */
```

Probably more relevant for `filterAttrs` anyway, but in case it helps someone.

roberth (commented):
Also ping @infinisil for a review

bb010g (changes requested):
I would much rather see `lib` have a
```nix
filterMap =
  pred: f: list:
  filter pred (map f list);
```
than a version that uses `null`, with my main reason being that it mirrors a potential
```nix
filterMapAttrs =
  pred: f: attrs:
  listToAttrs (
    concatMap (
      name:
      let
        value = f name attrs.${name};
      in
      if pred name value then [ { inherit name value; } ] else [ ]
    ) (attrNames attrs)
  );
```
that can return an attrset that's still lazy in its values. You mention that that `filterAttrs` is strict in the items, but `lib.filterAttrs` is strict in attr names and *not* strict in attr values. Forcing values is less avoidable in a list, but I'm writing some code that filters attrsets based solely on their names (which are required to be eager anyways) right now. I especially think that there's value in a `builtins.filterMapAttrs` with the above signature (note how even though attr names don't change, `listToAttrs` still has to be used in a library implementation) and I wouldn't want `lib.filterMap` to be that different.

Your `filterMap` semantics would be `filterMap (value: value != null)` under my proposed signature.

I'd also like to see some bench results for your current `filterMap` implementation using `foldl'` compared to an implementation using `filter` & `map` and an implementation using `concatMap`.

roberth:
@bb010g Your function behaves like `pred: f: attrs: mapAttrs f (filterAttrs pred) attrs`, i.e. map after filter. Shouldn't it be `mapFilterAttrs` then?
Also I'm not sure we should have it because it doesn't pass the [Fairbairn threshold](https://wiki.haskell.org/Fairbairn_threshold).
`filterMap` is also close to that threshold, but at least does a bit more than just sequencing two functions, specifically because everything happens in a single callback. That's useful because you can have a `let` binding in the callback for both the purpose of filtering as well as transforming. (Solving similar problem to the module system types' distinct `check` and `merge` functions.)


> You mention that that `filterAttrs` is strict in the items

I agree it is a little too hand-wavy what I [wrote](https://github.com/NixOS/nixpkgs/pull/323549#discussion_r1674056956).

Regarding the implementation of a function like `filterMapAttrs`, `concatMapAttrs` might be a good choice to build on. Or become a good choice, because I'd consider adding that functions to the `builtins`. It seems to be most general form of attrset transforming functions after `zipAttrsWith`, which adds list allocations that `concatMapAttrs` wouldn't need.
As for strictness, `concatMapAttrs` works fine, because you can return the relevant part of the attrset while the "map" part of the computation is only performed when the attribute value is evaluated. Also if you use `concatMapAttrs` directly instead of via `filterMapAttrs`, you get to write `let` bindings that are used by both the "predicate" logic and the "map" logic, similar to what this PR achieves.

nrdxp:
> I'd also like to see some bench results for your current `filterMap` implementation using `foldl'` compared to an implementation using `filter` & `map` and an implementation using `concatMap`.

`filter` and `map` basically causes you to iterate the list twice, so I can't imagine it would be competitive, performance wise. foldl' and concatMap are both builtins, however, so their comparison might be a bit more interesting. I might be able to put together a quick and dirty benchmark.

PR #273170, opened 2023-12-09
nix-channel: do not set empty nix-path when disabling channels

oxalica:
## Description of changes

This reverts the line introduced by #242098.

An empty nix-path in nix.conf will disable NIX_PATH environment variable entirely, which is not necessarily implied by users who want to disable nix channels. NIX_PATH also has some usages in tools like nixos-rebuild or just as user aliases.

**EDIT: It seems that empty nix-path also makes config `nix.nixPath` have no effect, because NIX_PATH is always ignored.**

That change is surprising and debatable, and also caused breakages in nixpkgs-review and user configs.

See:
- https://github.com/NixOS/nixpkgs/pull/242098/files#r1269891427
- https://github.com/Mic92/nixpkgs-review/issues/343

Not sure what to test. 

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
- For non-Linux: Is sandboxing enabled in `nix.conf`? (See [Nix manual](https://nixos.org/manual/nix/stable/command-ref/conf-file.html))
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:
  - [NixOS test(s)](https://nixos.org/manual/nixos/unstable/index.html#sec-nixos-tests) (look inside [nixos/tests](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests))
  - and/or [package tests](https://nixos.org/manual/nixpkgs/unstable/#sec-package-tests)
  - or, for functions and "core" functionality, tests in [lib/tests](https://github.com/NixOS/nixpkgs/blob/master/lib/tests) or [pkgs/test](https://github.com/NixOS/nixpkgs/blob/master/pkgs/test)
  - made sure NixOS tests are [linked](https://nixos.org/manual/nixpkgs/unstable/#ssec-nixos-tests-linking) to the relevant packages
- [ ] Tested compilation of all packages that depend on this change using `nix-shell -p nixpkgs-review --run "nixpkgs-review rev HEAD"`. Note: all changes have to be committed, also see [nixpkgs-review usage](https://github.com/Mic92/nixpkgs-review#usage)
- [ ] Tested basic functionality of all binary files (usually in `./result/bin/`)
[...]

PR #242098, opened 2023-07-07
nixos: Disable nix-channel

roberth:
###### Motivation

This fixes a warning that's been bothering me.

    warning: Nix search path entry '/nix/var/nix/profiles/per-user/root/channels' does not exist, ignoring

###### Description of changes

- Allow users to remove nix-channel and its accompanying files. Opt-in.

  `configuration.nix`:

       nix.channel.enable = false;

- Extends the installer test case to make sure users can switch from a traditional config to a flake.

###### Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
- [ ] For non-Linux: Is `sandbox = true` set in `nix.conf`? (See [Nix manual](https://nixos.org/manual/nix/stable/command-ref/conf-file.html))
- [x] Tested, as applicable:
  - [NixOS test(s)](https://nixos.org/manual/nixos/unstable/index.html#sec-nixos-tests) (look inside [nixos/tests](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests))
  - and/or [package tests](https://nixos.org/manual/nixpkgs/unstable/#sec-package-tests)
  - or, for functions and "core" functionality, tests in [lib/tests](https://github.com/NixOS/nixpkgs/blob/master/lib/tests) or [pkgs/test](https://github.com/NixOS/nixpkgs/blob/master/pkgs/test)
  - made sure NixOS tests are [linked](https://nixos.org/manual/nixpkgs/unstable/#ssec-nixos-tests-linking) to the relevant packages
- [ ] Tested compilation of all packages that depend on this change using `nix-shell -p nixpkgs-review --run "nixpkgs-review rev HEAD"`. Note: all changes have to be committed, also see [nixpkgs-review usage](https://github.com/Mic92/nixpkgs-review#usage)
- [ ] Tested basic functionality of all binary files (usually in `./result/bin/`)
- [23.11 Release Notes](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2311.section.md) (or backporting [23.05 Release notes](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2305.section.md))
  - [ ] (Package updates) Added a release notes entry if the change is major or breaking
  - [ ] (Module updates) Added a release notes entry if the change is significant
  - [ ] (Module addition) Added a release notes entry if adding a new NixOS module
[...]

PR #323613, opened 2024-06-30
nix-channel: do not set empty nix-path when disabling channels

CyberShadow:
## Description of changes

Continuation of https://github.com/NixOS/nixpkgs/pull/273170.

An empty nix-path in nix.conf will disable NIX_PATH environment variable entirely, which is not necessarily implied by users who want to disable nix channels. NIX_PATH also has some usages in tools like nixos-rebuild or just as user aliases.

That change is surprising and debatable, and also caused breakages in nixpkgs-review and user configs.

See:
- https://github.com/NixOS/nixpkgs/pull/242098/files#r1269891427
- https://github.com/Mic92/nixpkgs-review/issues/343
- https://github.com/NixOS/nix/pull/10998

Changes since https://github.com/NixOS/nixpkgs/pull/273170: implements @roberth's suggestion to warn when channel data still exists, as Nix (in the absence of a `NIX_PATH`) will still use it.

In this circumstance, Nix will now complain:

```
warning: Nix search path entry '/root/.nix-defexpr/channels' does not exist, ignoring
warning: Nix search path entry '/nix/var/nix/profiles/per-user/root/channels' does not exist, ignoring
```

It is a little annoying, but also in a way reassuring.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [x] Tested, as applicable:

---

### DIFF
--- a/nixos/modules/config/nix-channel.nix
+++ b/nixos/modules/config/nix-channel.nix
@@ -12,6 +12,7 @@ let
     mkDefault
     mkIf
     mkOption
+    stringAfter
     types
     ;
 
@@ -94,10 +95,29 @@ in
       NIX_PATH = cfg.nixPath;
     };
 
-    nix.settings.nix-path = mkIf (! cfg.channel.enable) (mkDefault "");
-
     systemd.tmpfiles.rules = lib.mkIf cfg.channel.enable [
       ''f /root/.nix-channels - - - - ${config.system.defaultChannel} nixos\n''
     ];
+
+    system.activationScripts.no-nix-channel = mkIf (!cfg.channel.enable)
+      (stringAfter [ "etc" "users" ] ''
+        if [ -e "/root/.nix-defexpr/channels" ]; then
+            echo "WARNING: /root/.nix-defexpr/channels exists, but channels have been disabled." 1>&2
+            echo "Due to https://github.com/NixOS/nix/issues/9574, Nix may still use these channels when NIX_PATH is unset." 1>&2
+            echo "Delete the above directory to prevent this." 1>&2
+        fi
+        if [ -e "/nix/var/nix/profiles/per-user/root/channels" ]; then
+            echo "WARNING: /nix/var/nix/profiles/per-user/root/channels exists, but channels have been disabled." 1>&2
+            echo "Due to https://github.com/NixOS/nix/issues/9574, Nix may still use these channels when NIX_PATH is unset." 1>&2
+            echo "Delete the above directory to prevent this." 1>&2
+        fi
+        getent passwd | while IFS=: read -r _ _ _ _ _ home _ ; do
+            if [ -n  "$home" -a -e "$home/.nix-defexpr/channels" ]; then
+                echo "WARNING: $home/.nix-defexpr/channels exists, but channels have been disabled." 1>&2
+                echo "Due to https://github.com/NixOS/nix/issues/9574, Nix may still use these channels when NIX_PATH is unset." 1>&2
+                echo "Delete the above directory to prevent this." 1>&2
+            fi
+        done
+      '');
   };
 }
--- a/nixos/modules/config/nix-channel.nix
+++ b/nixos/modules/config/nix-channel.nix
@@ -101,23 +101,25 @@ in
 
     system.activationScripts.no-nix-channel = mkIf (!cfg.channel.enable)
       (stringAfter [ "etc" "users" ] ''
+        explainChannelWarning=0
         if [ -e "/root/.nix-defexpr/channels" ]; then
-            echo "WARNING: /root/.nix-defexpr/channels exists, but channels have been disabled." 1>&2
-            echo "Due to https://github.com/NixOS/nix/issues/9574, Nix may still use these channels when NIX_PATH is unset." 1>&2
-            echo "Delete the above directory to prevent this." 1>&2
+            warn '/root/.nix-defexpr/channels exists, but channels have been disabled.'
+            explainChannelWarning=1
         fi
         if [ -e "/nix/var/nix/profiles/per-user/root/channels" ]; then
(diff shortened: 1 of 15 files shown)


SuperSandro2000 (approved):

roberth:
@fricklerhandwerk and I are making progress on a fix in Nix proper, and while I wish we could do just that, not this, and keep it simple, realistically NixOS has to deal with the buggy behavior for some time to come.
We can - later - make the warning conditional or tone it down depending on `config.nix.package.version`.

All that context aside, one small question.

> In this circumstance, Nix will now complain:
> 
> ```
> warning: Nix search path entry '/root/.nix-defexpr/channels' does not exist, ignoring
> warning: Nix search path entry '/nix/var/nix/profiles/per-user/root/channels' does not exist, ignoring
> ```
> 
> It is a little annoying, but also in a way reassuring.

Do I understand correctly that `NIX_PATH` is set to `""` by the configuration, and it would take for example `unset NIX_PATH` in a `shellHook` to cause these warnings to be printed?
That seems beneficial, because the intent of `unset NIX_PATH` is probably to ignore external `NIX_PATH` configuration such as system channels, and probably not "I don't care; use the system stuff", which is quite the opposite of what they probably want!

nixos/modules/config/nix-channel.nix
```diff
@@ -94,10 +95,22 @@ in
...
+            echo "WARNING: /root/.nix-defexpr/channels exists, but channels have been disabled." 1>&2
+            echo "Due to https://github.com/NixOS/nix/issues/9574, Nix may still use these channels when NIX_PATH is unset." 1>&2
+            echo "Delete the above directory to prevent this." 1>&2
+        fi
+        if [ -e "/nix/var/nix/profiles/per-user/root/channels" ]; then
+            echo "WARNING: /nix/var/nix/profiles/per-user/root/channels exists, but channels have been disabled." 1>&2
+            echo "Due to https://github.com/NixOS/nix/issues/9574, Nix may still use these channels when NIX_PATH is unset." 1>&2
+            echo "Delete the above directory to prevent this." 1>&2
+        fi
+      '');
```
roberth, line 114: Shouldn't we also check for user channels?

CyberShadow:
> Do I understand correctly that `NIX_PATH` is set to `""` by the configuration

Yes.

> and it would take for example `unset NIX_PATH` in a `shellHook` to cause these warnings to be printed?

The warnings are printed regardless of whether `NIX_PATH` is set. They're printed by the NixOS activation script.

The circumstance that I ran into where Nix was using an unintended default was running things with `env -i`, but I imagine the situation you described applies as well.

> That seems beneficial, because the intent of `unset NIX_PATH` is probably to ignore external `NIX_PATH` configuration such as system channels, and probably not "I don't care; use the system stuff", which is quite the opposite of what they probably want!

Agreed! In my case, I expected the script to fail with `'nixpkgs' was not found in the Nix search path`, instead of using the channels from when I installed NixOS :)

nixos/modules/config/nix-channel.nix
CyberShadow: Yes, good idea. Even though we can't enumerate all possible values of `HOME` that Nix could be invoked with, we could at least check the home directories in the user catalog. Added.

roberth: That easily catches 99% of users in my estimate, and that's a big win. :+1:

roberth (approved):
I've adjusted the warnings a bit (plus added the infra to make it nice; sorry for the scope creep), and I've done some testing, with the NixOS test as well as manually with a VM.

I think we may have an issue, which is that since recently NixOS sets a `NIX_PATH=nixpkgs=flake:nixpkgs` anyway, which I think is really bad, but that's a separate issue.

This PR is good now! :rocket:

nixos/modules/config/nix-channel.nix
```diff
@@ -94,10 +95,31 @@ in
...
+            explainChannelWarning=1
+        fi
+        if [ -e "/nix/var/nix/profiles/per-user/root/channels" ]; then
+            warn "/nix/var/nix/profiles/per-user/root/channels exists, but channels have been disabled."
+            explainChannelWarning=1
+        fi
+        getent passwd | while IFS=: read -r _ _ _ _ _ home _ ; do
+            if [ -n  "$home" -a -e "$home/.nix-defexpr/channels" ]; then
+                warn "$home/.nix-defexpr/channels exists, but channels have been disabled." 1>&2
+                explainChannelWarning=1
```
CyberShadow, line 116: This is set in a subshell so it will not be visible from outside the loop.
Flipping the redirect to be `while ... done < <(getent passwd)`, or using `shopt -s lastpipe` (bash-ism) should fix it.

Say, shouldn't ShellCheck have flagged this?

CyberShadow (commented):
Good stuff!

nixos/modules/config/nix-channel.nix
roberth: ohh..
I was only able to shellcheck a small part.
It'd be good to shellcheck the whole activation script, but that's a much larger exercise

roberth: I'll put this string in a file. Let's see

roberth: Shellcheck caught it too, but thank you for spotting this!

roberth:
@ofborg build nixosTests.installer.switchToFlake
@ofborg build nixosTests.nix-channel
@ofborg build nixosTests.activation-lib

CyberShadow:
> and it would take for example `unset NIX_PATH` in a `shellHook` to cause these warnings to be printed?

Apologies, I've only just now realized that you meant the warnings printed by Nix, not the ones printed by the activation script added here.

And... I can no longer recall how I got Nix to print those warnings. :facepalm:

PR #323616, opened 2024-06-30
python311Packages.scipy: attempt to fix Darwin build

doronbehar:
## Description of changes

cc @vcunat .

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [x] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/development/python-modules/scipy/default.nix
+++ b/pkgs/development/python-modules/scipy/default.nix
@@ -5,6 +5,7 @@
   fetchpatch,
   fetchurl,
   writeText,
+  xcbuild,
   python,
   buildPythonPackage,
   cython,
@@ -98,6 +99,12 @@ buildPythonPackage {
     pkg-config
     wheel
     setuptools
+  ] ++ lib.optionals stdenv.isDarwin [
+    # Minimal version required according to:
+    # https://github.com/scipy/scipy/blob/v1.14.0/scipy/meson.build#L185-L188
+    (xcbuild.override {
+      sdkVer = "13.3";
+    })
   ];
 
   buildInputs = [
--- a/pkgs/development/python-modules/scipy/default.nix
+++ b/pkgs/development/python-modules/scipy/default.nix
@@ -6,7 +6,6 @@
   fetchurl,
   writeText,
   python,
-  pythonOlder,
   buildPythonPackage,
   cython,
   gfortran,
@@ -22,7 +21,6 @@
   numpy,
   pybind11,
   pooch,
-  libxcrypt,
   xsimd,
   blas,
   lapack,
@@ -108,7 +106,7 @@ buildPythonPackage {
     pybind11
     pooch
     xsimd
-  ] ++ lib.optionals (pythonOlder "3.9") [ libxcrypt ];
+  ];
 
   propagatedBuildInputs = [ numpy ];
 
--- a/pkgs/development/python-modules/scipy/default.nix
+++ b/pkgs/development/python-modules/scipy/default.nix
@@ -23,6 +23,10 @@
   pybind11,
   pooch,
   xsimd,
+  # Upstream has support for using Darwin's Accelerate package. However this
+  # requires a Darwin user to work on a nice way to do that via an override.
+  # See:
+  # https://github.com/scipy/scipy/blob/v1.14.0/scipy/meson.build#L194-L211
   blas,
   lapack,
 


vcunat:
This builds for me on aarch64-darwin.  With both pythons; BTW 3.12 is the default now.

pkgs/development/python-modules/scipy/default.nix
```diff
@@ -100,6 +104,12 @@ buildPythonPackage {
     pkg-config
     wheel
     setuptools
+    # A shim xcrun executable, to fool upstream's meson.build that we have or
+    # on a macos with new enough SDK. We don't have the real executable
+    # available as a real package in Nixpkgs.
+    (writeShellScriptBin "xcrun" ''
+      echo "13.3"
+    '')
```
vcunat, line 112: Why is this not conditioned by `stdenv.isDarwin`?

doronbehar: > Why is this not conditioned by `stdenv.isDarwin`?

Hmm I was thinking about it, but it doesn't hurt at all to put it in the Linux build as well, as this executable is not even being run on a Linux build. If you find a good choice to condition this I won't mind to, it will make the diff a bit more ugly due to the `nixfmt` requirement...

doronbehar:
BTW what do you think about the time out pytest failure reported by ofborg?

```
FAILED lib/python3.12/site-packages/scipy/optimize/tests/test_milp.py::test_milp_timeout_16545[options0-Time limit reached. (HiGHS Status 13:] - assert None is not None
```

It's not present on Hydra right?

vcunat:
Probably a transient issue, maybe the CI machine was more loaded at that point.  Scipy on linux seems OK on Hydra's last eval: https://hydra.nixos.org/eval/1807322?filter=scipy#tabs-still-succeed

pkgs/development/python-modules/scipy/default.nix
vcunat: It felt a bit weird to have it on linux.  And the thousands of linux rebuilds are also a consideration at this point.

doronbehar: > It felt a bit weird to have it on linux. And the thousands of linux rebuilds are also a consideration at this point.

That's a good reason. I also fixed the `pythonOlder` commit such that it won't trigger linux rebuilds.

pkgs/development/python-modules/scipy/default.nix
```diff
@@ -100,6 +103,13 @@ buildPythonPackage {
     pkg-config
     wheel
     setuptools
+  ] ++ lib.optionals stdenv.isDarwin [
+    # A shim xcrun executable, to fool upstream's meson.build that we have or
+    # on a macos with new enough SDK. We don't have the real executable
+    # available as a real package in Nixpkgs.
+    (writeShellScriptBin "xcrun" ''
+      echo "13.3"
+    '')
```
wegank, line 112: Does `(xcbuild.override { sdkVer = "13.3"; })` work?

doronbehar: > Does `(xcbuild.override { sdkVer = "13.3"; })` work?

I'm not sure it will evaluate, [this `xcbuild`](https://github.com/NixOS/nixpkgs/blob/staging-next/pkgs/development/tools/xcbuild/default.nix) expression doesn't accept an `sdkVer` argument, and according to `git grep` there are no places where such overrides are used.

wegank: The argument is defined [in the wrapper](https://github.com/NixOS/nixpkgs/blob/staging-next/pkgs/development/tools/xcbuild/wrapper.nix), and you'll see there's an `xcrun = writeTextFile something` just below.

doronbehar: > The argument is defined [in the wrapper](https://github.com/NixOS/nixpkgs/blob/staging-next/pkgs/development/tools/xcbuild/wrapper.nix?rgh-link-date=2024-07-01T07%3A17%3A55Z), and you'll see there's an `xcrun = writeTextFile something` just below.

OK great, indeed it is better then my script.

Issue #318674, opened 2024-06-10
Build failure: rustc


reckenrode:
### Steps To Reproduce

1. `nix build 'github:NixOS/nixpkgs?ref=afa876d6feea138d23310cc19662b0d3364570d9#rustc'` (tested only on Darwin).

### Build log

https://gist.github.com/reckenrode/53c181f0c8fff37707e719734250024c

### Additional context

Possibly due to https://github.com/NixOS/nixpkgs/pull/316761. https://github.com/NixOS/nixpkgs/pull/317273 may be related to the fix, but it targets machine flags not hardening flags.

Clang does [provide][1] `hardeningUnsupportedFlagsByTargetPlatform`, but if Rust is not invoking a wasm32-unknown-unknown target clang, I don’t think that will apply.

### Notify maintainers

@risicle @alyssais

### Metadata

Please run `nix-shell -p nix-info --run "nix-info -m"` and paste the result.

```console
[user@system:~]$ nix-shell -p nix-info --run "nix-info -m"
 - system: `"aarch64-darwin"`
 - host os: `Darwin 23.5.0, macOS 14.5`
 - multi-user?: `no`
 - sandbox: `yes`
 - version: `nix-env (Nix) 2.18.2`
 - channels(root): `"nixpkgs"`
 - nixpkgs: `/etc/nix/inputs/nixpkgs`
```

I actually have the Darwin sandbox disabled, so I don’t know what’s up with `nix-info`.

---

Add a :+1: [reaction] to [issues you find important].

[reaction]: https://github.blog/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/
[issues you find important]: https://github.com/NixOS/nixpkgs/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions-%2B1-desc

[1]: https://github.com/NixOS/nixpkgs/blob/cd7ec2419a9a98e9119d64f5b0cbaa069e53c298/pkgs/development/compilers/llvm/common/clang/default.nix#L141-L143

alyssais:
Sounds like maybe we should not enable this by default for clang stdenv until we have a mechanism to filter Clang hardening flags by target?

risicle:
[...]

PR #307880, opened 2024-04-30
cctools: 973.0.1 -> 1010.6

reckenrode:
## Description of changes

This PR updates cctools and ld64. It also fixes related breakage. It is currently set to draft until 24.05 is released. I will be rebasing weekly against staging, resolving any merge conflicts then. Before updating the ld64 branch, I build my configs and the Darwin channel blockers to confirm no regressions.

The following notes are copied from https://discourse.nixos.org/t/darwin-updates-news/42249/10.

### Packaging Changes

* cctools, ld64, and libtapi are now built from Apple’s [OSS distributions](https://github.com/apple-oss-distributions) repo instead of the [cctools-port](https://github.com/tpoechtrager/cctools-port) and [apple-libtapi](https://github.com/tpoechtrager/apple-libtapi) repos.
* These versions (cctools 1010.6, ld64 951.9, libtapi 1500.0.12.3) correspond to the versions shipped with Xcode 15. cctools and ld64 are current as of Xcode 15.3. Packages should no longer have to patch around missing flags or feature support.
* Restoring Linux support is a WIP. I have made some changes to use non-private and non-portable APIs, but I am currently focused on getting Darwin updates first.
* The packages have been moved to the by-name hierarchy. Aliases are provided to the old names, but there are some breaking changes (see below).
* cctools is now only cctools. This is technically a breaking change (see below). ld64 has been moved to its own, separate package.
* darwin.binutils is now the package that defines the mix of cctools and LLVM tools. It has been update with additional tools. See below for the breakdown of what it contains.
* The build system has been replaced with Meson. I did this because dealing with xcbuild stinks, and these packages eventually need to support Linux for cross-compilation.
* corecrypto and CommonCrypto hashing APIs are reimplemented using OpenSSL on both Darwin (and eventually) Linux. cctools-port only does this on non-Darwin platforms.
* libtapi has been split into separate out and lib outputs. Most packages only need libtapi.dylib, which is ~500 KiB compared to ~30 MiB for the tapi binary.
* libtapi is linked against zlib. I don’t know what it uses zlib to do, but the upstream libtapi is also linked against zlib, and I want to match what upstream is doing.

_Note:_ ld-prime is will not be packaged because it is not currently included in the source releases. While I think it’s unlikely, ld-prime will be added separately as a new package should the source be released.

### New Features

[...]

PR #323679, opened 2024-07-01
WIP x64 darwin default to sdk 11

paparodeo:
## Description of changes

changes to make x86-64 darwin default to sdk 11.

the flags used to check for the sdk just assumed that x64 darwin was correlated with sdk 10.12. that is no longer the case with this change tho the updated flags can be improved.

testing: built stdenv on x64 darwin

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---


paparodeo:
cc @emilazy

emilazy:
I support aligning `x86_64-darwin` and `aarch64-darwin` here, but don’t think we should do it until @reckenrode’s SDK rework happens.

paparodeo:
> I support aligning `x86_64-darwin` and `aarch64-darwin` here, but don’t think we should do it until `@reckenrode’s` SDK rework happens.

ok. i built stdenv then tried inkscape but that dies on rustc https://github.com/NixOS/nixpkgs/issues/318674. I don't think that the SDK rework and this change are mutually exclusive -- or at least i do not think the sdk rework has started.  keeping x64 darwin at 10.12 should be possible with this change by leaving ` darwinSdkVersion = final.sdkVer or (if final.isAarch64 then "11.0" else "10.12");` in `lib/systems/default.nix`.

reckenrode:
The SDK refactor is going to consolidate the separate SDK hierarchies and provide a common pattern for building SDKs. I’m using GCC as a point of reference, which does this pretty well. The Darwin stdenv will see extensive changes. All the separate code paths will be removed. The more extreme changes will remove frameworks and rely on `SDKROOT`.

But I need to land https://github.com/NixOS/nixpkgs/pull/307880 and https://github.com/NixOS/nixpkgs/pull/256590 first before I even look at making SDK changes.

paparodeo:
i still don't understand how this change conflicts with a sdk refactor that has not started. IIRC, when this change was proposed in late march by annalee the sdk refactor that was going to use gcc as a point of reference was still right around the corner.

most of what this change is doing is changing `stdenv.isDarwin && stdenv.isx86_64` into `stdenv.isDarwin && lib.versionOlder stdenv.hostPlatform.darwinSdkVersion "11"`, or at least that is the idea, so a modification to set darwinSdkVersion in lib/systems/default.nix will change the default sdk.

these changes could go into master today with xnu refactor and dropping the  cc-wrapper and lib/systems/default.nix change and cause 0 rebuilds. i don't see how that would conflict with a future sdk refactor either.

nonetheless, i don't want to spend time on a change that is going nowhere so closing.

reckenrode:
The “conflict” would be losing the test case of having multiple SDKs and needing to support overriding the SDK version. If it got bumped in the meantime, it wouldn’t be my preference, but it wouldn’t stop that work.

> IIRC, when this change was proposed in late march by annalee the sdk refactor that was going to use gcc as a point of reference was still right around the corner.

I’m sorry. It’s targeting the 24.11 release, so I’ve prioritized getting the ld64 and the Darwin cross-compilation PRs done first. I have only so much bandwidth, and I need to be careful to avoid burnout. (I also made a misstep of starting work on the clang 18 update while I was blocked during the 24.05 release freeze, but I put that on hold to focus on this other stuff.)

paparodeo:
I don't see how this change prohibits testing multiple sdks -- i see how changing the default for x64 back to 10.12  will require a stdenv rebuild tho i am assuming that an sdk refactor would do that anyway.

i'm not trying to be critical of the sdk refactor not being far enough along, more just trying to be realistic about timelines. if waiting for sdk refactor i don't think that trying x64 darwin at sdk 11 happens till after 24.11.

reckenrode:
My primary (albeit unlikely) concern would be whether switching the default back even worked. A more realistic concern would be packages removing their overrides.

> i'm not trying to be critical of the sdk refactor not being far enough along, more just trying to be realistic about timelines. if waiting for sdk refactor i don't think that trying x64 darwin at sdk 11 happens till after 24.11.

And I’m not trying to slow progress with vaporware promises. 🙂

Honestly, I don’t know if updating the SDK version would possible today even if the refactor were done. When there was talk of bumping to the minimum supported version to 10.13 (for ld64 before I found a workaround), the release team wanted to put an announcement in the 24.05 release before changing the minimum version. As far as I know, that announcement didn’t make the 24.05 release notes, so I’ve been assuming the refactor would not be accompanied by a minimum version bump.

> the sdk refactor / rewrite seems much more beneficial.

I’ve been calling it a refactor because the first stage would be landing the 10.12.4 and 11.3 SDKs in the common pattern and as compatible as possible attribute-wise, but it’s definitely a rewrite.

emilazy:
> The “conflict” would be losing the test case of having multiple SDKs and needing to support overriding the SDK version. If it got bumped in the meantime, it wouldn’t be my preference, but it wouldn’t stop that work.

There’s some stuff that wants the 12.0 SDK, right? Would it be more or less fuss to add that to test those kinds of things in the rework, as opposed to recreating the current SDK set?

Honestly I’d personally be really happy if we could make do with just one SDK, but I don’t know if that’s remotely viable. Still, if it’d be possible to land this without interfering significantly with your SDK rework I’d be in support of it. `x86_64-darwin` is in its twilight and the less divergence it has with `aarch64-darwin` the more practical it’ll be to keep it going as long as we can. I suspect the amount of breakage even on older macOS versions will be manageable, especially if we accept best‐effort patches to fix it.

paparodeo:
i am unable to re-open this PR so i created a new one: https://github.com/NixOS/nixpkgs/pull/324155 
i was able to drop the change to `cc-wrapper` by using an unwrapped clang in the xnu header build. also, i left out the change to `lib/systems/default.nix` which changes the default sdk to "11.0" for x64 darwin.

PR #323696, opened 2024-07-01
lianad: init at 6.0

plebhash:
## Description of changes

the `liana` package currently only supports the `liana-gui` executable, while it is desired to also support headless use-cases

this PR adds a `withGui` attribute, which is used to choose between:
- provide `liana-gui` executable if `withGui` is `true`
- provide `lianad` + `liana-cli` executables if `withGui` is `false` (plus a `$out/etc/liana/config.toml` based on [`lianad_config_example.toml`](https://github.com/wizardsardine/liana/blob/master/contrib/lianad_config_example.toml))

cc @dunxen

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/maintainers/maintainer-list.nix
+++ b/maintainers/maintainer-list.nix
@@ -16330,6 +16330,12 @@
     githubId = 358550;
     name = "Philip Lykke Carlsen";
   };
+  plebhash = {
+    name = "plebhash";
+    email = "plebhash@proton.me";
+    github = "plebhash";
+    githubId = 147345153;
+  };
   pleshevskiy = {
     email = "dmitriy@pleshevski.ru";
     github = "pleshevskiy";
--- a/pkgs/by-name/li/liana/package.nix
+++ b/pkgs/by-name/li/liana/package.nix
@@ -38,7 +38,7 @@ let
 in
 rustPlatform.buildRustPackage rec {
   pname = "liana";
-  version = "6.0";
+  version = "6.0"; # keep in sync with lianad
 
   src = fetchFromGitHub {
     owner = "wizardsardine";
--- a/pkgs/by-name/li/lianad/package.nix
+++ b/pkgs/by-name/li/lianad/package.nix
@@ -1,0 +1,43 @@
+{
+  lib,
+  stdenv,
+  rustPlatform,
+  fetchFromGitHub,
+  udev,
+}:
+
+rustPlatform.buildRustPackage rec {
+  pname = "lianad";
+  version = "6.0"; # keep in sync with liana
+
+  src = fetchFromGitHub {
+    owner = "wizardsardine";
+    repo = "liana";
+    rev = "v${version}";
+    hash = "sha256-LLDgo4GoRTVYt72IT0II7O5wiMDrvJhe0f2yjzxQgsE=";
+  };
+
+  cargoHash = "sha256-a4hLtDXnEeTa0e1LcMkEPKEqGBp5bzWseq5Pe5ZYF1M=";
+
+  buildInputs = [ udev ];
+
+  postInstall = ''
+    install -Dm0644 ./contrib/lianad_config_example.toml $out/etc/liana/config.toml
+  '';
+
+  # bypass broken unit tests
+  doCheck = false;
+
+  meta = {
+    mainProgram = "lianad";
+    description = "Bitcoin wallet leveraging on-chain timelocks for safety and recovery";
+    homepage = "https://wizardsardine.com/liana";
+    license = lib.licenses.bsd3;
+    maintainers = [
+      lib.maintainers.dunxen
+      lib.maintainers.plebhash
+    ];
+    platforms = lib.platforms.linux;
+    broken = stdenv.isAarch64;
+  };
+}


pkgs/by-name/li/liana/default.nix
dunxen, line 1: I don't think you meant to rename this. It breaks CI.

pkgs/top-level/all-packages.nix
```diff
@@ -21553,6 +21553,14 @@ with pkgs;
...
   jxrlib = callPackage ../development/libraries/jxrlib { };
 
+  liana = callPackage ../by-name/li/liana {
+    withGui = true;
+  };
+
+  lianad = callPackage ../by-name/li/liana {
+    withGui = false;
+  };
+
```
dunxen, line 21563: nit: we could probably just oneline and group these by dropping the newline inbetween as
```suggestion
  liana = callPackage ../by-name/li/liana { withGui = true; };
  lianad = callPackage ../by-name/li/liana { withGui = false; };
```

pkgs/by-name/li/liana/default.nix
```diff
@@ -61,27 +63,29 @@ rustPlatform.buildRustPackage rec {
...
       exec = "liana-gui";
       icon = "liana";
       desktopName = "Liana";
       comment = meta.description;
     })
-  ];
+  ] else null;
 
-  doCheck = true;
+  doCheck = if withGui then true else false;
```
dunxen, line 85: Is something failing for `lianad`?

pkgs/by-name/li/liana/default.nix
```diff
@@ -61,27 +63,29 @@ rustPlatform.buildRustPackage rec {
     systemd
   ];
 
-  sourceRoot = "${src.name}/gui";
+  sourceRoot = if withGui then "${src.name}/gui" else null;
```
dunxen, line 66: Would the `sourceRoot` not be `"${src.name}"` for lianad instead of null?

dunxen (commented):
Thanks for opening this!

Left some comments, and it seems that only `lianad` gets built without `liana-cli` when `withGui` is `false`.

pkgs/by-name/li/liana/default.nix
plebhash: I renamed it because as soon as I added the entries to `all-packages.nix`, I started getting the following error:
```
$ nix-build -A liana
error:

         at /home/plebhash/develop/nixpkgs/lib/customisation.nix:263:8:

          262|
          263|     in if missingArgs == {}
             |        ^
          264|        then makeOverridable f allArgs

       … while calling the 'listToAttrs' builtin

         at /home/plebhash/develop/nixpkgs/lib/attrsets.nix:647:5:

          646|     set:
          647|     listToAttrs (concatMap (name: let v = set.${name}; in if pred name v then [(nameValuePair name v)] else []) (attrNames set));
             |     ^
          648|

       error: opening file '/home/plebhash/develop/nixpkgs/pkgs/by-name/li/liana/default.nix': No such file or directory

```

pkgs/by-name/li/liana/default.nix
plebhash: yes, if I revert this back to `doCheck = true;` and do `nix-build -A lianad`:

```
failures:

---- config::tests::config_directory stdout ----
thread 'config::tests::config_directory' panicked at src/config.rs:482:13:
assertion failed: filepath.as_path().starts_with("/home/")


failures:
    config::tests::config_directory

test result: FAILED. 50 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 4.07s

error: test failed, to rerun pass `--lib`
error: builder for '/nix/store/78if09d6hy9sr6ymcw6m1c6xzdh0si4z-lianad-5.0.drv' failed with exit code 101;
       last 10 log lines:
       > assertion failed: filepath.as_path().starts_with("/home/")
       > note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
       >
       >
       > failures:
       >     config::tests::config_directory
       >
       > test result: FAILED. 50 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 4.07s
       >
       > error: test failed, to rerun pass `--lib`
       For full logs, run 'nix log /nix/store/78if09d6hy9sr6ymcw6m1c6xzdh0si4z-lianad-5.0.drv'.

```

pkgs/by-name/li/liana/default.nix
plebhash: according to the [docs on `rustPlatform`](https://github.com/NixOS/nixpkgs/blob/master/doc/languages-frameworks/rust.section.md#compiling-non-rust-packages-that-include-rust-code-compiling-non-rust-packages-that-include-rust-code):

> sourceRoot: when the `Cargo.lock`/`Cargo.toml` are in a subdirectory, `sourceRoot` specifies the relative path to these files.

so the rationale here was: if `withGui` is `false`, then we avoid defining `sourceRoot`

if I change this line to:
```
  sourceRoot = if withGui then "${src.name}/gui" else "${src.name}";
```

compilation also works, so I think either way should be fine.

plebhash:
> it seems that only `lianad` gets built without `liana-cli` when `withGui` is false

@dunxen are you sure? this is what I get here:
```
$ nix-build -A lianad
/nix/store/jjldm51jm0jss2lnac841v4725hbjcg1-lianad-5.0
$ tree result        
result
├── bin
│   ├── liana-cli
│   └── lianad
└── etc
    └── liana
        └── config.toml

4 directories, 3 files

```

pkgs/by-name/li/liana/default.nix
dunxen: That's fine if it's default at root. Can leave it null.

pkgs/by-name/li/liana/default.nix
plebhash: based on my understand of what you said here https://github.com/fort-nix/nix-bitcoin/pull/708#issuecomment-2199512798

it seems we should move it out of `by-name`?

pkgs/by-name/li/liana/default.nix
plebhash: based on my understanding of what you said here https://github.com/fort-nix/nix-bitcoin/pull/708#issuecomment-2199512798

it seems we should move it out of `by-name`?

pkgs/by-name/li/liana/default.nix
```diff
@@ -14,6 +14,7 @@
 , systemd
 , vulkan-loader
 , xorg
+, withGui
```
dunxen, line 17:
```suggestion
, withGui ? true
```

pkgs/top-level/all-packages.nix
dunxen: Actually we only need `lianad` here as that is the virtual package:

`lianad = callPackage liana.override { withGui = false; };`

but the filename change above should be reverted.

pkgs/by-name/li/liana/default.nix
dunxen: Ok we can leave it then :+1:

pkgs/by-name/li/liana/default.nix
dunxen: Nah let's not move. See the other comments :)

dunxen:
> > it seems that only `lianad` gets built without `liana-cli` when `withGui` is false
> 
> @dunxen are you sure? this is what I get here:
> 
> ```
> $ nix-build -A lianad
> /nix/store/jjldm51jm0jss2lnac841v4725hbjcg1-lianad-5.0
> $ tree result        
> result
> ├── bin
> │   ├── liana-cli
> │   └── lianad
> └── etc
>     └── liana
>         └── config.toml
> 
> 4 directories, 3 files
> ```

You are correct. I mischecked the result.

pkgs/top-level/all-packages.nix
dunxen: You can see a similar setup for `lshw-gui` and `lshw` (just inverted bools)

plebhash: looks like CI [doesn't like this approach](https://github.com/NixOS/nixpkgs/actions/runs/9753108716/job/26917825105?pr=323696#step:7:16):
```
- Attribute `pkgs.lianad` is a new top-level package using `pkgs.callPackage ... { /* ... */ }`.
  Please define it in pkgs/by-name/li/lianad/package.nix instead.
  See `pkgs/by-name/README.md` for more details.
  Since the second `callPackage` argument is not `{ }`, the manual `callPackage` in pkgs/top-level/all-packages.nix is still needed.

This PR introduces additional instances of discouraged patterns as listed above. Merging is discouraged but would not break the base branch.
```

dunxen: Ah yes sorry, didn't delete the `callPackage` in my comment after copying the original.

Thanks!

dunxen (approved):

dunxen (changes requested):
Oops. Didn't see the PR title (I cannot edit it). Should be:

`lianad: init at 5.0`

dunxen (approved):
LGTM. Thanks!

Aleksanaa (commented):
it looks a bit weird because you are effectively making two derivation (which are different except for version and source.) into one. This may not be a good idea.

Good idea:

1. Use callPackages to declare multiple packages.
2. Just build them separately, and add a comment to remind keeping version the same
3. Same to 2 but add assert to keep version the same. This may break r-ryantm, however

dunxen:
> it looks a bit weird because you are effectively making two derivation (which are different except for version and source.) into one. This may not be a good idea.
> 
> Good idea:
> 
>     1. Use callPackages to declare multiple packages.
> 
>     2. Just build them separately, and add a comment to remind keeping version the same
> 
>     3. Same to 2 but add assert to keep version the same. This may break r-ryantm, however

I agree in the current state it is not as discoverable / clear in the source what's going on, but I'm not sure which of those options more closely aligns to other packages that are in the same situation (see for example `lshw` vs `bitcoin` and a few others). Seems there are a few different methods.

I'm unfortunately unaware of any documentation around this practice, but would love to keep consistent and still have this maintainable without breaking r-ryantm.

Either way, if this should be a separate package declaration, I'll help if we need to keep them in sync manually without an `assert` (i.e. Option 2)

Aleksanaa:
> see for example lshw vs bitcoin and a few others

That's `withGui`, just an option to decide whether to include gui or not, and is often passed as an argument to build tools. The `withGui` in this package is actually building another package. They have different sourceRoots and different dependencies. Calling it `withGui` is quite misleading.

Yes, there are more cursed examples, see `catppuccin` (although I do want to remove that stuff for a long time)

dunxen:
> That's `withGui`, just an option to decide whether to include gui or not, and is often passed as an argument to build tools. The `withGui` in this package is actually building another package. They have different sourceRoots and different dependencies. Calling it `withGui` is quite misleading.
> 
> Yes, there are more cursed examples, see `catppuccin` (although I do want to remove that stuff for a long time)

Right, got it. Thanks @Aleksanaa for clearing that up! So let's not add to the cursed list then :)

I'm inclined to go with Option 2 and have a separate package declaration in `pkgs/by-name/li/lianad/package.nix`. @plebhash, is this alright with you? You can add me and yourself as maintainers there. Then you can just drop the diff in `all-packages.nix` :)

We can investigate option 3 at some later stage.

dunxen (changes requested):
Just changing the approval status to make it clear we're splitting this package.

dunxen (changes requested):
Thanks for updating this!
Just need to remove some GUI-related libs and it should be good ❤️

dunxen:
Once this goes in, we can handle the v6.0 update <3

pkgs/by-name/li/lianad/package.nix
```diff
@@ -0,0 +1,73 @@
...
+    pkg-config
+    cmake
+    copyDesktopItems
+    makeWrapper
+  ];
+
+  buildInputs = [
+    fontconfig
+    systemd
+  ];
```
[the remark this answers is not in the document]
plebhash: I passed a fine comb and noticed that there was actually a lot more stuff that could be removed without affecting the build. I'm committing the bare minimum to get `lianad` + `liana-cli` to compile. That actually boils down to quite few dependencies, since all we need is a simple rust build.

Here's the diff of everything I'm removing:
```diff
diff --git a/pkgs/by-name/li/lianad/package.nix b/pkgs/by-name/li/lianad/package.nix
index 88502c9a1f61..1cb805a86639 100644
--- a/pkgs/by-name/li/lianad/package.nix
+++ b/pkgs/by-name/li/lianad/package.nix
@@ -2,34 +2,8 @@
 , stdenv
 , rustPlatform
 , fetchFromGitHub
-, pkg-config
-, cmake
-, copyDesktopItems
-, makeDesktopItem
-, makeWrapper
-, expat
-, fontconfig
-, freetype
-, libGL
-, systemd
-, vulkan-loader
-, xorg
 }:
 
-let
-  runtimeLibs = [
-    expat
-    fontconfig
-    freetype
-    freetype.dev
-    libGL
-    vulkan-loader
-    xorg.libX11
-    xorg.libXcursor
-    xorg.libXi
-    xorg.libXrandr
-  ];
-in
 rustPlatform.buildRustPackage rec {
   pname = "lianad";
   version = "5.0"; # keep in sync with liana
@@ -43,18 +17,6 @@ rustPlatform.buildRustPackage rec {
 
   cargoSha256 = "sha256-6r8G4sO/aDO5VyJbmbdFz3W2b8ZmulXJZtnz8IAwcaU=";
 
-  nativeBuildInputs = [
-    pkg-config
-    cmake
-    copyDesktopItems
-    makeWrapper
-  ];
-
-  buildInputs = [
-    fontconfig
-    systemd
-  ];
-
   postInstall = ''
     install -Dm0644 ./contrib/lianad_config_example.toml $out/etc/liana/config.toml
   '';

```

This diff makes me wonder if some of these components could be also removed from the GUI package. For example, I don't get the point of `systemd`, since I can't identify anything that specifies an actual `systemd` service on the GUI package.

But that is probably work for a separate PR (maybe while upgrading for 6.0?).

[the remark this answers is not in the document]
dunxen: The GUI requires all of those (could be one or two that are not required anymore, but at the time it was the bare minimum).

I'm curious about not needing `systemd` as an input. I recall it being required for USB devices to work at all. Would you mind checking that you can connect hardware without it?

Sorry for the back and forth.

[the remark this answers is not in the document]
dunxen: And yes, I'll make sure to see if we can now trim down any other dependencies for GUI in the 6.0 update PR! :)

pkgs/by-name/li/liana/package.nix
dunxen, line 1: optional: Feel free to add yourself as a maintainer on this package too.

pkgs/by-name/li/lianad/package.nix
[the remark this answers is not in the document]
dunxen: Yeah I'll investigate a bit more but let's leave `systemd` there :)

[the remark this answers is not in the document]
plebhash: sounds good!

should I focus on any further changes?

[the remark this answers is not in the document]
dunxen: Nope, when you add systemd back I'll accept. You can also add yourself as a maintainer to `liana` as well as `lianad`.

pkgs/by-name/li/lianad/package.nix
```diff
@@ -0,0 +1,35 @@
...
+  '';
+
+  doCheck = false;
+
+  meta = with lib; {
+    mainProgram = "lianad";
+    description = "A Bitcoin wallet leveraging on-chain timelocks for safety and recovery";
+    homepage = "https://wizardsardine.com/liana";
+    license = licenses.bsd3;
+    maintainers = with maintainers; [ dunxen plebhash ];
```
dunxen, line 31: If you add yourself in as a maintainer, you also have to have an entry for yourself, `plebhash` in https://github.com/NixOS/nixpkgs/blob/master/maintainers/maintainer-list.nix

dunxen (changes requested):
You also have to have an entry for yourself, `plebhash` in https://github.com/NixOS/nixpkgs/blob/master/maintainers/maintainer-list.nix. That's what is causing the last breaking check. :)

pkgs/by-name/li/lianad/package.nix
plebhash: done

maintainers/maintainer-list.nix
```diff
@@ -15815,6 +15815,11 @@
     githubId = 358550;
     name = "Philip Lykke Carlsen";
   };
+  plebhash = {
+    email = "plebhash@proton.me";
+    github = "plebhash";
+    githubId = 147345153;
+  };
```
dunxen, line 15822: `name` must be defined. It's fine to just put your nym there like this.

```suggestion
  plebhash = {
    name = "plebhash";
    email = "plebhash@proton.me";
    github = "plebhash";
    githubId = 147345153;
  };
```

dunxen (changes requested):

dunxen (approved):

dunxen (approved):

pkgs/by-name/li/lianad/package.nix
```diff
@@ -0,0 +1,38 @@
...
+
+  cargoSha256 = "sha256-6r8G4sO/aDO5VyJbmbdFz3W2b8ZmulXJZtnz8IAwcaU=";
+
+  buildInputs = [ systemd ];
+
+  postInstall = ''
+    install -Dm0644 ./contrib/lianad_config_example.toml $out/etc/liana/config.toml
+  '';
+
+  doCheck = false;
```
Aleksanaa, line 27: Please add reason in comment

pkgs/by-name/li/lianad/package.nix
```diff
@@ -0,0 +1,38 @@
...
+  version = "5.0"; # keep in sync with liana
+
+  src = fetchFromGitHub {
+    owner = "wizardsardine";
+    repo = "liana";
+    rev = "v${version}";
+    hash = "sha256-RkZ2HSN7IjwN3tD0UhpMeQeqkb+Y79kSWnjJZ5KPbQk=";
+  };
+
+  cargoSha256 = "sha256-6r8G4sO/aDO5VyJbmbdFz3W2b8ZmulXJZtnz8IAwcaU=";
```
Aleksanaa, line 19:
```suggestion
  cargoHash = "sha256-6r8G4sO/aDO5VyJbmbdFz3W2b8ZmulXJZtnz8IAwcaU=";
```

pkgs/by-name/li/lianad/package.nix
```diff
@@ -0,0 +1,38 @@
...
+
+  postInstall = ''
+    install -Dm0644 ./contrib/lianad_config_example.toml $out/etc/liana/config.toml
+  '';
+
+  doCheck = false;
+
+  meta = with lib; {
+    mainProgram = "lianad";
+    description = "A Bitcoin wallet leveraging on-chain timelocks for safety and recovery";
```
Aleksanaa, line 31:
```suggestion
    description = "Bitcoin wallet leveraging on-chain timelocks for safety and recovery";
```

pkgs/by-name/li/lianad/Cargo.lock
Aleksanaa, line 1: Please delete this

maintainers/maintainer-list.nix
Aleksanaa, line 1: Please split this into commit `maintainers: add plebhash`.

pkgs/by-name/li/liana/package.nix
Aleksanaa, line 1: Please split this into commit `liana: add notice to keep version in sync with lianad`

pkgs/by-name/li/lianad/package.nix
```diff
@@ -0,0 +1,38 @@
...
+    rev = "v${version}";
+    hash = "sha256-RkZ2HSN7IjwN3tD0UhpMeQeqkb+Y79kSWnjJZ5KPbQk=";
+  };
+
+  cargoSha256 = "sha256-6r8G4sO/aDO5VyJbmbdFz3W2b8ZmulXJZtnz8IAwcaU=";
+
+  buildInputs = [ systemd ];
+
+  postInstall = ''
+    install -Dm0644 ./contrib/lianad_config_example.toml $out/etc/liana/config.toml
```
Aleksanaa, line 24: Are you sure it will read this (without wrapping with some `XDG_` environment variables?)

pkgs/by-name/li/lianad/package.nix
```diff
@@ -0,0 +1,38 @@
...
+  '';
+
+  doCheck = false;
+
+  meta = with lib; {
```
Aleksanaa, line 29: Better to avoid `with lib`

pkgs/by-name/li/lianad/Cargo.lock
plebhash: I put this here to follow the same pattern of the `liana` package:

https://github.com/NixOS/nixpkgs/blob/master/pkgs/by-name/li/liana/Cargo.lock

pkgs/by-name/li/lianad/package.nix
plebhash: this is also just following the `liana` package:

https://github.com/NixOS/nixpkgs/blob/8764d898c4f365d98ef77af140b32c6396eb4e02/pkgs/by-name/li/liana/package.nix#L85

pkgs/by-name/li/lianad/package.nix
plebhash: can you elaborate? my testing rationale is based on the contents of `result` after running `nix-build -A lianad`:

```
result
├── bin
│   ├── liana-cli
│   └── lianad
└── etc
    └── liana
        └── config.toml
```

dunxen:
Need to run `nixfmt`.

dunxen (approved):

dunxen (changes requested):
Needs rebase and should init at 6.0 now. I had to update `liana` as 5.0 had build errors with rust 1.80: https://github.com/NixOS/nixpkgs/pull/333059

pkgs/by-name/li/liana/package.nix
dunxen: @plebhash not sure if you saw this comment above to split this.

pkgs/by-name/li/lianad/package.nix
```diff
@@ -0,0 +1,43 @@
...
+    rev = "v${version}";
+    hash = "sha256-LLDgo4GoRTVYt72IT0II7O5wiMDrvJhe0f2yjzxQgsE=";
+  };
+
+  cargoHash = "sha256-a4hLtDXnEeTa0e1LcMkEPKEqGBp5bzWseq5Pe5ZYF1M=";
+
+  buildInputs = [ systemd ];
+
+  postInstall = ''
+    install -Dm0644 ./contrib/lianad_config_example.toml $out/etc/liana/config.toml
```
dunxen, line 25: Just using `udev` works now. I confirmed on the latest update of `liana`.

```suggestion
{
  lib,
  stdenv,
  rustPlatform,
  fetchFromGitHub,
  udev,
}:

rustPlatform.buildRustPackage rec {
  pname = "lianad";
  version = "6.0"; # keep in sync with liana

  src = fetchFromGitHub {
    owner = "wizardsardine";
    repo = "liana";
    rev = "v${version}";
    hash = "sha256-LLDgo4GoRTVYt72IT0II7O5wiMDrvJhe0f2yjzxQgsE=";
  };

  cargoHash = "sha256-a4hLtDXnEeTa0e1LcMkEPKEqGBp5bzWseq5Pe5ZYF1M=";

  buildInputs = [ udev ];

  postInstall = ''
    install -Dm0644 ./contrib/lianad_config_example.toml $out/etc/liana/config.toml
```

dunxen (changes requested):
Thanks! Last few comments :crossed_fingers:

dunxen (commented):
First commit message should be ｀lianad: init at 5.0｀

dunxen (approved):

Aleksanaa (approved):

PR #323718, opened 2024-07-01
ed: a huge refactor

AndersonTorres:
## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---


pkgs/by-name/ed/ed/package.nix
```diff
@@ -1,9 +1,40 @@
-{ lib
-, fetchurl
+{
+  lib,
+  callPackage,
+  lzip,
+  runtimeShellPackage,
+  stdenv,
+  # Configurable options
+  sources ? callPackage ./sources.nix { },
```
jtojnar, line 8: What is the point of this compared to `overideAttrs` and using `finalAttrs` pattern?

pkgs/top-level/all-packages.nix
```diff
@@ -30252,9 +30252,6 @@ with pkgs;
 
   ecs-agent = callPackage ../applications/virtualization/ecs-agent { };
 
-  inherit (recurseIntoAttrs (callPackage ../applications/editors/ed { }))
-    ed edUnstable;
```
jtojnar, line 30256: `edUnstable` should go to aliases as `throw`.

pkgs/by-name/ed/ed/package.nix
AndersonTorres: I was writing an expression that would build both edUnstable and ed. To facilitate this, I have factored sources acquisition in a separate file.

pkgs/top-level/all-packages.nix
AndersonTorres: Done.

pkgs/by-name/ed/ed/package.nix
jtojnar: I thought that `edUnstable` was removed.

AndersonTorres: Yes, I removed it because it does not exist anymore  (I verified it today). So I removed the detached sources.nix.

pkgs/by-name/ed/ed/package.nix
```diff
@@ -0,0 +1,61 @@
...
+  runtimeShellPackage,
+  stdenv,
+  testers,
+}:
+
+# Note: this package is used for bootstrapping fetchurl, and thus cannot use
+# fetchpatch! All mutable patches (generated by GitHub or cgit) that are needed
+# here should be included directly in Nixpkgs as files; further, those patches
+# should be included in the directory (or subdirectories) where this package file
+# belongs, given that ed now follows RFC 140.
```
jtojnar, line 14: Do we need to specify the patch location? Using the same directory as the package was the convention even before RFC 140. And I would assume most Nixpkgs contributors to be familiar with Nixpkgs conventions – if they are not, the CI will catch it anyway.

AndersonTorres: Given that
- I have found some expressions that refer to files outside directory 
  - SDL and gettext, that I am trying to migrate
- Many people are not well used to read the manuals and announcements
  - Things like "Why I can't use `qt6.callPackage ../by-name/. . .`?" -  or worse,  "Have you tried `qt6.callPackage ../by-name/. . .`?" - are commonplace

I preferred to be redundant.

pkgs/by-name/ed/ed/package.nix
```diff
@@ -0,0 +1,61 @@
...
+  lib,
+  fetchurl,
+  lzip,
+  runtimeShellPackage,
+  stdenv,
+  testers,
+}:
+
+# Note: this package is used for bootstrapping fetchurl, and thus cannot use
+# fetchpatch! All mutable patches (generated by GitHub or cgit) that are needed
```
SuperSandro2000, line 11:
```suggestion
# fetchpatch! All mutable patches (obtained by GitHub or cgit) that are needed
```

pkgs/by-name/ed/ed/package.nix
SuperSandro2000: > * I have found some expressions that refer to files outside directory
>       
>       * SDL and gettext, that I am trying to migrate

I think those are very special cases and we should look at them individually. 


I don't think we need to explicitly mention it, but it also can't hurt. Maybe we can slim down the comment. No strong feelings.

pkgs/by-name/ed/ed/package.nix
AndersonTorres: Done

pkgs/by-name/ed/ed/package.nix
AndersonTorres: This convention of "writing everything related to a package in the same directory" is not really enforced.
The only place it is enforced is the by-name hierarchy.

And I know how difficult is to convince people about following unwritten and non-enforceable conventions.

Aleksanaa: > I preferred to be redundant.

Then should we also add these three lines of comments to all other packages in by-name?

AndersonTorres: > > I preferred to be redundant.
> 
> Then should we also add these three lines of comments to all other packages in by-name?

How many of them can't use `fetchPatch` because they are used to bootstrap the fetchers?

AndersonTorres: Reworded to not cite RFC 140

jtojnar (approved):
Did not test but looks good.

PR #323743, opened 2024-07-01
nixos/glance: init module

drupol:
## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/nixos/modules/services/web-apps/glance.nix
+++ b/nixos/modules/services/web-apps/glance.nix
@@ -1,0 +1,141 @@
+{
+  config,
+  lib,
+  pkgs,
+  ...
+}:
+let
+  cfg = config.services.glance;
+
+  inherit (lib)
+    mkEnableOption
+    mkPackageOption
+    mkOption
+    mkIf
+    getExe
+    types
+    ;
+
+  settingsFormat = pkgs.formats.yaml { };
+in
+{
+  options.services.glance = {
+    enable = mkEnableOption "glance";
+    package = mkPackageOption pkgs "glance" { };
+
+    settings = mkOption {
+      type = types.submodule {
+        freeformType = settingsFormat.type;
+        options = {
+          server = {
+            host = mkOption {
+              description = "Glance bind address";
+              default = "127.0.0.1";
+              example = "0.0.0.0";
+              type = types.str;
+            };
+            port = mkOption {
+              description = "Glance port to listen on";
+              default = 8080;
+              example = 5678;
+              type = types.port;
+            };
+          };
+          pages = mkOption {
+            type = settingsFormat.type;
+            description = ''
+              List of pages to be present on the dashboard.
+
+              See <https://github.com/glanceapp/glance/blob/main/docs/configuration.md#pages--columns>
+            '';
+            default = [
+              {
+                name = "Calendar";
+                columns = [
+                  {
+                    size = "full";
+                    widgets = [ { type = "calendar"; } ];
+                  }
+                ];
+              }
+            ];
+            example = [
+              {
+                name = "Home";
+                columns = [
+                  {
+                    size = "full";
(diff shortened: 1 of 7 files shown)


drupol:
@ofborg build glance.tests.service

Aleksanaa:
See https://github.com/NixOS/nixpkgs/tree/master/nixos#commit-conventions

drupol:
@ofborg build glance.tests.service

devusb (approved):
Tested including configuration using `settings` -- seemed to work well -- LGTM. Thanks -- this is a cool one!

nixos/modules/services/web-apps/glance.nix
```diff
@@ -0,0 +1,135 @@
...
+    };
+
+    port = lib.mkOption {
+      description = "Glance port to listen on";
+      default = 8080;
+      type = types.port;
+    };
+
+    settings = mkOption {
+      type = settingsFormat.type;
```
h7x4, line 39: Maybe we could place these options within `settings`? It would remove the need for the `recursiveUpdate` further down.

```suggestion
    settings = mkOption {
      type = lib.types.submodule {
			  freeformType = settingsFormat.type;
			  options = {
          host = lib.mkOption {
            description = "Glance bind address";
            default = "127.0.0.1";
            type = types.str;
          };

          port = lib.mkOption {
            description = "Glance port to listen on";
            default = 8080;
            type = types.port;
          };
        };
			};
```

drupol: I thought about doing that, but I really wanted to keep `host` and `port` out of the `glance` configuration file. IMO, it seems better doing so, but perhaps I am totally wrong.

h7x4: By all means :) I guess I should also clarify that this is a completely non-blocking suggestion.

But I'm interested to hear your thoughts. What are the merits of keeping these outside `settings`? Is it better for discoverability, or for similarity with other modules?

The only downside I can see is a scenario where someone has an existing configuration and rewrites it to nix, only to put everything in settings and be surprised that the `host` and `port` is overwritten by the `default` values on the other options. I guess it's unlikely, since most people probably use https://search.nixos.org or something similar to discover options, but I know I have been bitten by similar overwriting mechanisms before.

drupol: OK cool.

This PR introduce a new service. That service has a configuration file and is bound to an address and a port.

For transparency, I definitely think that those 2 network settings should be clearly separated from the config file.

When thinking twice about your proposal, it actually could be a good idea because `port` and `host` are still separated... but I'm having trouble seeing how to make it work since those settings needs to be wrapped under `server.port` and `server.host` in the `glance` config file.

Have you tried your proposal, is it working?

h7x4: I tried it out with this patch: https://gist.github.com/h7x4/caf28ce4eda95ef3b4aa3bd5f9349d3c

I also edited the default `pages` value (with example) to exist within the submodule, so that it doesn't disappear the moment you set anything within `services.glance.settings`.

drupol: Thank you, I merged it :) now I understand your point and I totally agree with it! TIL !

drupol:
@devusb Do you mind making a review? I integrated the good changes suggested by @h7x4 !

devusb:
@drupol looks great -- adjusted my test configuration to the new pattern, worked well including configuration of a page with some widgets on a non-standard port and host.

devusb (approved):

PR #269356, opened 2023-11-23
Ensure all jobset attributes evaluate, and check that in CT

ghost:
#### Description

aka ["eval checks go brrrr"](https://github.com/NixOS/nixpkgs/pull/269403)

Please review the commits individually.  Produced as part of the non-Hydra tooling used to validate

- https://github.com/NixOS/nixpkgs/pull/268097

#### Description of changes

This PR adds `release-attrpaths-superset.nix`, which walks the entire nixpkgs tree and emits a superset of all release attrnames in only 44 seconds on a 3ghz CPU, using less than 5 gbytes of memory.  By comparison, on the same CPU the `nix-env -qaP` hack used by ofborg on every PR submission requires waiting **25 minutes** and peaks at 58 gbytes, even with checkMeta turned off.  Full details below.  This is:

- 33x faster (or 2.9% of the elapsed time)
- 12.5x less memory (or 8.0% of the peak memory usage)

If all you need is the attrnames -- for example to launch a mass rebuild -- then you're done.  If you want to replace the ofborg `outpaths.nix` step, see the follow-up PR:

- https://github.com/NixOS/nixpkgs/pull/269403

#### Validation

To confirm that the `release-attrpaths-superset.nix` really are a superset of the slower method, run

```
nix-build pkgs/test/release/release-attrpaths-superset.nix -A failures
```

#### Additional CT ([Continuous Testing](https://en.wikipedia.org/wiki/Continuous_testing)) check

This PR adds `recurseForRelease=false` where necessary within nixpkgs in order to make this possible -- it screens out infinite recursions and several non-`tryEval`-catchable failures (for example, [passing a mistyped argument to a builtin](https://nixos.org/manual/nix/unstable/language/builtins#builtins-tryEval), which is an uncatchable error) and infinite recursions.

Finding all the places where it was necessary to add this was a lot of tedious work, so to make sure that work is not undone, `lib/tests/release.nix` now verifies that `release-attrpaths-superset.nix` evaluates successfully.

Although it doesn't really fit in `lib/tests`, there isn't anywhere else it can be placed and still be run by ofborg on every commit.  If some other such place materializes in the future I would be happy to relocate it, but since ofborg is out-of-tree this PR should not block on that happening.

#### Fixes

- https://github.com/NixOS/nixpkgs/issues/269357

PR #323854, opened 2024-07-01
lib.types: remove __attrsFailEvaluation

philiptaron:
## Description of changes

The test (`nix-build pkgs/test/release/default.nix`) continues to pass without this preventative measure.

This line was added in #269356.

## Things done

- [x] Tested, as applicable:
  - [x] `nix-build pkgs/test/release/default.nix`

### DIFF
--- a/lib/types.nix
+++ b/lib/types.nix
@@ -73,7 +73,6 @@ let
 
   outer_types =
 rec {
-  __attrsFailEvaluation = true;
   isType = type: x: (x._type or "") == type;
 
   setType = typeName: value: value // {


lf-:
sorry, I'm done with the nix organization. can you request someone else for reviews? thanks!

roberth:
Hi @lf-, is there something I could do?
Organization was always an overstatement, so I like to focus on achieving things together with others in the community.

@philiptaron 
As for these PRs @philiptaron made, they seem to be part of an incomplete effort by someone who's left GitHub, so we could either finish that work or remove it. Until someone wants to lead such an effort, I think it's better to keep the code clean, and it's easy enough to re-add something like this when needed.

lf-:
nope, sorry. #323887 I've done my best to make things work but at some point you just have to call it for the health of all parties involved.

philiptaron:
> As for these PRs you made, they seem to be part of an incomplete effort by someone who's left GitHub, so we could either finish that work or remove it. Until someone wants to lead such an effort, I think it's better to keep the code clean, and it's easy enough to re-add something like this when needed.

Thanks for merging. The ghost was (as you likely know) `@amjoseph-nixpkgs`, and the work was attempting to make the eval checks faster. I think this is a good goal, and the test (`pkgs/test/release/default.nix`) means it _could_ be checked in CI. I see no evidence that it currently is, but I definitely could be missing something.

The set of PRs is motivated by the same sentiment you expressed -- "it's better to keep the code clean" -- and if the test passes with the code removed, all the better.

PR #323866, opened 2024-07-01
vcpkg{,-tool}: miscellaneous fixups

h7x4:
## Description of changes

**Summary:**

- Moved most of the wrapper from vcpkg to vcpkg-tool directly (so you can rely on vcpkg-tool with an external VCPKG_ROOT and not have to repeat the patches)
- Ensure all of the arguments in the (now) vcpkg-tool wrapper is overridable, using envvars prefixed with `NIX` to communicate that these are nix specific.
- Patch some of the vcpkg cmake scripts, to conform the compiler detection to the executable names used in `pkgsCross`.
- Add some missing runtime deps for vcpkg-tool (these are listed as direct dependencies at https://github.com/microsoft/vcpkg/blob/101cc9a69a1061969caf4b73579a34873fdd60fe/scripts/bootstrap.sh#L88-L91 as well as https://github.com/microsoft/vcpkg/blob/101cc9a69a1061969caf4b73579a34873fdd60fe/scripts/azure-pipelines/linux/provision-image.sh#L27)

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/by-name/vc/vcpkg-tool/package.nix
+++ b/pkgs/by-name/vc/vcpkg-tool/package.nix
@@ -5,6 +5,7 @@
 , cacert
 , cmake
 , cmakerc
+, curl
 , fmt
 , git
 , gzip
@@ -12,6 +13,7 @@
 , ninja
 , openssh
 , python3
+, unzip
 , zip
 , zstd
 , extraRuntimeDeps ? []
@@ -55,12 +57,14 @@ stdenv.mkDerivation (finalAttrs: {
     runtimeDeps = [
       cacert
       cmake
+      curl
       git
       gzip
       meson
       ninja
       openssh
       python3
+      unzip
       zip
       zstd
     ] ++ extraRuntimeDeps;
--- a/pkgs/by-name/vc/vcpkg-tool/package.nix
+++ b/pkgs/by-name/vc/vcpkg-tool/package.nix
@@ -16,6 +16,8 @@
 , unzip
 , zip
 , zstd
+, runCommand
+, writeText
 , extraRuntimeDeps ? []
 , doWrap ? true
 }:
@@ -160,6 +162,20 @@ stdenv.mkDerivation (finalAttrs: {
     mv "$out/bin/vcpkg" "$out/bin/.vcpkg-wrapped"
     install -Dm555 "$vcpkgWrapperPath" "$out/bin/vcpkg"
   '';
+
+  passthru.tests = {
+    testWrapper = runCommand "vcpkg-tool-test-wrapper" { buildInputs = [ finalAttrs.finalPackage ];  } ''
+      export NIX_VCPKG_DEBUG_PRINT_ENVVARS=true
+      vcpkg --x-packages-root="test" --x-install-root="test2" contact > "$out"
+
+      cat "$out" | head -n 4 | diff - ${writeText "vcpkg-tool-test-wrapper-expected" ''
+        NIX_VCPKG_DOWNLOADS_ROOT = /homeless-shelter/.vcpkg/root/downloads
+        NIX_VCPKG_BUILDTREES_ROOT = /homeless-shelter/.vcpkg/root/buildtrees
+        NIX_VCPKG_PACKAGES_ROOT = test
+        NIX_VCPKG_INSTALL_ROOT = test2
+      ''}
+    '';
+  };
 
   meta = {
     description = "Components of microsoft/vcpkg's binary";
--- a/pkgs/by-name/vc/vcpkg-tool/package.nix
+++ b/pkgs/by-name/vc/vcpkg-tool/package.nix
@@ -183,7 +183,7 @@ stdenv.mkDerivation (finalAttrs: {
     homepage = "https://github.com/microsoft/vcpkg-tool";
     changelog = "https://github.com/microsoft/vcpkg-tool/releases/tag/${finalAttrs.src.rev}";
     license = lib.licenses.mit;
-    maintainers = with lib.maintainers; [ guekka gracicot ];
+    maintainers = with lib.maintainers; [ guekka gracicot h7x4 ];
     platforms = lib.platforms.all;
   };
 })
--- a/pkgs/by-name/vc/vcpkg-tool/package.nix
+++ b/pkgs/by-name/vc/vcpkg-tool/package.nix
@@ -1,6 +1,7 @@
 { lib
 , stdenv
 , fetchFromGitHub
+, runtimeShell
 , cacert
 , cmake
 , cmakerc
@@ -14,6 +15,7 @@
 , zip
 , zstd
 , extraRuntimeDeps ? []
+, doWrap ? true
 }:
 stdenv.mkDerivation (finalAttrs: {
   pname = "vcpkg-tool";
@@ -46,12 +48,10 @@ stdenv.mkDerivation (finalAttrs: {
   ];
 
 
-  # vcpkg needs two directories to write to that is independent of installation directory.
-  # Since vcpkg already creates $HOME/.vcpkg/ we use that to create a root where vcpkg can write into.
   passAsFile = [ "vcpkgWrapper" ];
   vcpkgWrapper = let
     # These are the most common binaries used by vcpkg
-    # Extra binaries can be added via overlay when needed
+    # Extra binaries can be added through override when needed
     runtimeDeps = [
       cacert
       cmake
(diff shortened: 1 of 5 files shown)


Guekka:
This looks good, thank you! I've been wanting to add XDG compliance for a while
I assume you have tested the changes?

Nitpick: I think `substituteInPlace` is recommended over sed

h7x4:
> This looks good, thank you! I've been wanting to add XDG compliance for a while
> I assume you have tested the changes?

Admittedly, I have not tested with this exact patch. I created the patch out of a bunch of local overrides in a project. I'll try building it with overlayed changes from this PR.

> Nitpick: I think `substituteInPlace` is recommended over sed

Just to be clear, `substituteInPlace` does not allow for regexes. Would you be okay with unrolling `(as|gcc|g++)` to 3 similar copies of `--replace-fail`?

Guekka:
> Just to be clear, substituteInPlace does not allow for regexes. Would you be okay with unrolling (as|gcc|g++) to 3 similar copies of --replace-fail?

Oh, right. I only suggested it because I think it's common in nixpkgs. But actually, I think it might be more obvious this way, so let's keep it like this if you're fine with it

I'll approve once I get a chance to test it, should be pretty soon

h7x4:
On a second note, this is a pretty breaking change since we now do not allow use of `${vcpkg-tool}/bin/vcpkg` with the 4 flags we are setting (I know I at least have used this as an escape hatch in the past, to avoid the overrides in `${vcpkg}/bin/vcpkg`). I think we should write a release note, and maybe document the use of these 4 variables somewhere.

I'm also open to do some kind of magic, where we detect the incoming flags from outside the wrappeer and drop the overrides accordingly. I think it would make for a more native-feeling, seamless experience. But I'm not very confident in my bash skills :sweat_smile: 

I'll mark the PR as draft in the meantime.

gracicot:
@h7x4 This is a breaking change indeed, but I'm open to not break too much if there's benefit. The magic flags seems okay to me. It's not too magic, it just let the users override default locations we set up for them.

Also if `vcpkg-tool` is wrapped by default, I would also provide `vcpkg-tool-unwrapped`, as with other build tools that have a wrapper

For example, I might type such command to manage a private registry:
```
vcpkg --x-builtin-ports-root=./ports --x-builtin-registry-versions-dir=./versions x-add-version --all --verbose
```

Having a unwrapped version would make sense for such scripts

h7x4:
I have added a new part to the wrapper that will loop over the args and detect user-provided versions of the args we override. This should make the UX better, while still ensuring the tool doesn't write to the nix store.

pkgs/by-name/vc/vcpkg-tool/package.nix
```diff
@@ -47,23 +49,100 @@ stdenv.mkDerivation (finalAttrs: {
...
+      }
+    ];
   in ''
-    wrapProgram $out/bin/vcpkg --prefix PATH ${lib.makeBinPath runtimeDeps}
+    #!${runtimeShell}
+
+    NIX_VCPKG_WRITABLE_PATH=''${NIX_VCPKG_WRITABLE_PATH:-''${XDG_CACHE_HOME+"$XDG_CACHE_HOME/vcpkg"}}
+    NIX_VCPKG_WRITABLE_PATH=''${NIX_VCPKG_WRITABLE_PATH:-''${HOME+"$HOME/.vcpkg/root"}}
+    NIX_VCPKG_WRITABLE_PATH=''${NIX_VCPKG_WRITABLE_PATH:-''${TMP}}
+    NIX_VCPKG_WRITABLE_PATH=''${NIX_VCPKG_WRITABLE_PATH:-'/tmp'}
```
Guekka, line 109: For someone who doesn't know well bash, this was not easy to grasp. What do you think of something like this? Longer, but more explicit
```suggestion
    get_vcpkg_path() {
      if [[ -n $NIX_VCPKG_WRITABLE_PATH ]]; then
          echo "$NIX_VCPKG_WRITABLE_PATH"
      elif [[ -n $XDG_CACHE_HOME ]]; then
          echo "$XDG_CACHE_HOME/vcpkg"
      elif [[ -n $HOME ]]; then
          echo "$HOME/.vcpkg/root"
      elif [[ -n $TMP ]]; then
          echo "$TMP"
      else
          echo "/tmp"
      fi
    }

    NIX_VCPKG_WRITABLE_PATH=$(get_vcpkg_path)
```

pkgs/by-name/vc/vcpkg-tool/package.nix
```diff
@@ -47,23 +49,100 @@ stdenv.mkDerivation (finalAttrs: {
...
+                ((i++))
+              fi
+              ;;
+          '';
+        in lib.concatMapStringsSep "\n" f argsWithDefault}
+        *)
+          FINAL_NONMODIFIED_ARGS+=(''${ARGS[i]})
+          ;;
+      esac
+    done
```
Guekka, line 136: This logic is (necessarily) complex. I'm wondering if it's worth writing a test

Guekka (commented):
I think your script has the right idea

pkgs/by-name/vc/vcpkg-tool/package.nix
h7x4: I think it would be nice both for testing and for debugging to provide an environment variable that will debug print the final cli flags in the wrapper before calling `vcpkg`. We could create a test where we set this envvar, and run the tool with a command such as `contact` or `--version`, and then verify that the resulting flags from the debug output looks sane.

What do you think?

Guekka: Sounds good to me

h7x4: I've added a basic test. It should be able to report any syntactic failures of the wrapper in case the script suddenly doesn't run at all, as well as basic functionality errors.

pkgs/by-name/vc/vcpkg-tool/package.nix
```diff
@@ -47,31 +51,139 @@ stdenv.mkDerivation (finalAttrs: {
...
+      elif [[ -n $TMP ]]; then
+          echo "$TMP"
+      else
+          echo "/tmp"
+      fi
+    }
+
+    NIX_VCPKG_WRITABLE_PATH=$(get_vcpkg_path)
+
+    ${lib.concatMapStringsSep "\n" ({ env, default, ... }: ''${env}=''${${env}-"${default}"}'') argsWithDefault}
```
Guekka, line 124: If I'm not mistaken, this expression is repeated thrice and should be extracted to a function

h7x4: We actually use it 4 places at this point. But I am not too sure whether it makes for a good abstraction.

The only thing common between the invocations is that we apply some function for all the arguments with defaults, the function returns a string, and then concatenate with `\n` between. The intent behind the functions are very different. While something like `applyFuncForArgsWithDefaultAndConcat` might save some space, it doesn't save any evaluation time, nor do I think it makes it much clearer to the reader.

But I'd be happy if we could find a better expression

h7x4:
@gracicot: added a `vcpkg-tool-unwrapped` attribute in top-level.

I've added a piece of documentation that will end up in the nixpkgs manual. Built with `nix build .#htmlDocs.nixpkgsManual.x86_64-linux` and verified that the formatting looks sane (but I'd like some verification that the content is reasonable as well)

I can't think of any more changes that I'd like to do before merging this, so I'll undraft the PR. Happy if I could get some more reviews :)

Guekka (approved):
Looks good to me, thank you!

gracicot:
Overlay triplets doesn't seem to work for me. I'm investigating

pkgs/by-name/vc/vcpkg-tool/package.nix
```diff
@@ -47,31 +51,139 @@ stdenv.mkDerivation (finalAttrs: {
...
+    done
+
+    if [ "''${NIX_VCPKG_DEBUG_PRINT_ENVVARS-'false'}" = 'true' ]; then
+      ${lib.concatMapStringsSep "\n" ({ env, ... }: "  " + ''echo "${env} = ''$${env}"'') argsWithDefault}
+      echo ""
+    fi
+
+    exec -a "$0" "${placeholder "out"}/bin/.vcpkg-wrapped" \
+    ${lib.concatMapStringsSep "\n" ({ arg, env, ... }: "  " + ''${arg}="''$${env}" \'') argsWithDefault}
+      "$FINAL_NONMODIFIED_ARGS"
```
gracicot, line 158: Arguments sent from `scripts/buildsystems/vcpkg.cmake` are not forwarded, leading to overlay triplets and target triplets to not work

gracicot (changes requested):

pkgs/by-name/vc/vcpkg/package.nix
```diff
@@ -16,27 +16,29 @@ stdenvNoCC.mkDerivation (finalAttrs: {
...
+  nativeBuildInputs = [ makeWrapper ];
+
+  postPatch = ''
+    substituteInPlace scripts/toolchains/linux.cmake \
+      --replace-fail "aarch64-linux-gnu-as"  "aarch64-unknown-linux-gnu-as" \
+      --replace-fail "aarch64-linux-gnu-gcc" "aarch64-unknown-linux-gnu-gcc" \
+      --replace-fail "aarch64-linux-gnu-g++" "aarch64-unknown-linux-gnu-g++" \
+      --replace-fail "arm-linux-gnueabihf-as"  "armv7l-unknown-linux-gnueabihf-as" \
+      --replace-fail "arm-linux-gnueabihf-gcc" "armv7l-unknown-linux-gnueabihf-gcc" \
+      --replace-fail "arm-linux-gnueabihf-g++" "armv7l-unknown-linux-gnueabihf-g++"
```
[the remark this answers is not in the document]
h7x4: While the binary names might not be nixpkgs-only, I believe they differ from most distros. Ubuntu at least uses the ones specified upstream. I think we should leave them be.

pkgs/by-name/vc/vcpkg-tool/package.nix
h7x4: Thanks for looking into it. I'll have a closer look soon to see if we can make the proposed changes work somehow

pkgs/by-name/vc/vcpkg-tool/package.nix
```diff
@@ -47,31 +51,139 @@ stdenv.mkDerivation (finalAttrs: {
...
+      fi
+    }
+
+    NIX_VCPKG_WRITABLE_PATH=$(get_vcpkg_path)
+
+    ${lib.concatMapStringsSep "\n" ({ env, default, ... }: ''${env}=''${${env}-"${default}"}'') argsWithDefault}
+
+    export PATH="${lib.makeBinPath runtimeDeps}''${PATH:+":$PATH"}"
+
+    ARGS=( "$@" )
```
gracicot, line 128: The problem seems to originate from here. I changed the script to output the values. Here's what I did:
```
ARGS=( "$@" )
echo "REALLY_ALL_ARGS $@"
echo "ARGS ${ARGS}"
```
Here's the output I see in my project:
```
REALLY_ALL_ARGS install --triplet wasm32-unknown-emscripten --vcpkg-root /Users/grc/Workspace/github/nixpkgs/result/share/vcpkg --x-wait-for-lock --x-manifest-root=/Users/grc/Workspace/software/pixmob/apps/visualizer/visualizer-web/packages/vtx --x-install-root=/Users/grc/Workspace/software/pixmob/apps/visualizer/visualizer-web/packages/vtx/build/dev-emscripten/vcpkg_installed --host-triplet=aarch64-darwin-clang --overlay-triplets=/Users/grc/Workspace/software/pixmob/apps/visualizer/visualizer-web/packages/vtx/triplets
ARGS install
```
Here `ARGS` should not be just `install`, it should be an array with all the args above, if I understood the scripts correctly.

pkgs/by-name/vc/vcpkg-tool/package.nix
gracicot: Yes! I'm really looking forward those. It's a good cleanup that was overdue.

pkgs/by-name/vc/vcpkg-tool/package.nix
h7x4: This seems to be an issue indeed. `$ARRAY` only yields the first element of the array, while `${ARRAY[@]}` yields all elements. I seem to have overlooked it for `FINAL_NONMODIFIED_ARGS`, making it so only the first of the non-special arguments are forwarded from the wrapper. I'll push a fix. Does it resolve the overlay issues as well?

gracicot: The updated version seems to work as expected. I approve the changes!

gracicot (approved):

PR #324058, opened 2024-07-02
lib.systems: do not confuse isStatic with hasSharedLibraries

tie:
## Description of changes

* isStatic: packages should be statically linked against the dependencies when possible.

* hasSharedLibraries: the platform supports dynamic linking. For example, if set to `false`, things like dlopen do not work. If the package *requires* dynamic linking (e.g. loads plugins), it should be considered unavailable on such platform.

That is, these are two related but distinct concepts. We shouldn’t assume that statically linked system does not support dynamic libraries. For example, GNU/Linux (i.e. glibc) and macOS binaries can be statically linked but have to use dynamic program loader (dyld or ld.so).

Note that it is reasonable to set `hasSharedLibraries = false` on Linux since the entire system is built from source (we can use C library that does not require shared libraries, e.g. musl) and there is a useful subset of packages that works without shared libraries.

This is still a bit ambiguous and naming could be better, but I think it’s a step in the right direction.

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/doc/stdenv/meta.chapter.md
+++ b/doc/stdenv/meta.chapter.md
@@ -101,9 +101,28 @@ For example, a package which requires dynamic linking and cannot be linked stati
 ```nix
 {
   meta.platforms = lib.platforms.all;
+  meta.badPlatforms = [{ hasSharedLibraries = false; }];
+}
+```
+
+or
+
+```nix
+{
+  meta.platforms = lib.platforms.all;
   meta.badPlatforms = [ lib.systems.inspect.platformPatterns.isStatic ];
 }
 ```
+
+In the examples above, `lib.systems.inspect.platformPatterns.isStatic` is an alias for `{ isStatic = true; }` platform pattern. The difference between `isStatic` and `hasSharedLibraries` is that:
+
+* isStatic: packages should be statically linked against the dependencies when
+  possible. Dynamic linking may or may not be available.
+
+* hasSharedLibraries: the platform supports dynamic linking. For example, if set
+  to `false`, things like dlopen do not work. If the package *requires* dynamic
+  linking (e.g. must load plugins dynamically), it should be considered
+  unavailable on such platform.
 
 The [`lib.meta.availableOn`](https://github.com/NixOS/nixpkgs/blob/b03ac42b0734da3e7be9bf8d94433a5195734b19/lib/meta.nix#L95-L106) function can be used to test whether or not a package is available (i.e. buildable) on a given platform.
 Some packages use this to automatically detect the maximum set of features with which they can be built.
--- a/lib/systems/inspect.nix
+++ b/lib/systems/inspect.nix
@@ -164,6 +164,10 @@ rec {
   # apply only to the `parsed` field.
 
   platformPatterns = mapAttrs (_: p: { parsed = {}; } // p) {
+    # Matches platforms that are statically linked. That does not imply that
+    # dynamic linking is not supported.
     isStatic = { isStatic = true; };
+    # Matches platforms that support dynamic linking.
+    hasSharedLibraries = { hasSharedLibraries = true; };
   };
 }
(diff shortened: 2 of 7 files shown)


doc/stdenv/meta.chapter.md
```diff
@@ -98,13 +98,32 @@ Hydra will never create prebuilt binaries for these platform types, even if they
...
 ```
 
+In the examples above, `lib.systems.inspect.platformPatterns.isStatic` is an alias for `{ isStatic = true; }` platform pattern. The difference between `isStatic` and `hasSharedLibraries` is that:
+
+* isStatic: packages should be statically linked against the dependencies when
+  possible.
+
+* hasSharedLibraries: the platform supports dynamic linking. For example, if set
+  to `false`, things like dlopen do not work. If the package *requires* dynamic
+  linking (e.g. loads plugins), it should be considered unavailable on such
```
roberth, line 124:
```suggestion
  linking (e.g. must load plugins dynamically), it should be considered unavailable on such
```

I don't want to spend too many words on this, but we can hint a bit:

must: plugins may be optional.
dynamically: maybe the plugins can be linked statically; always plugged in. Linux allows this for its modules.

lib/systems/inspect.nix
```diff
@@ -164,6 +164,10 @@ rec {
   # apply only to the `parsed` field.
 
   platformPatterns = mapAttrs (_: p: { parsed = {}; } // p) {
+    # Matches platforms that are statically linked. That does not mean that
```
roberth, line 167:
```suggestion
    # Matches platforms that are statically linked. That does not imply that
```

doc/stdenv/meta.chapter.md
```diff
@@ -98,13 +98,32 @@ Hydra will never create prebuilt binaries for these platform types, even if they
...
 {
   meta.platforms = lib.platforms.all;
   meta.badPlatforms = [ lib.systems.inspect.platformPatterns.isStatic ];
 }
 ```
 
+In the examples above, `lib.systems.inspect.platformPatterns.isStatic` is an alias for `{ isStatic = true; }` platform pattern. The difference between `isStatic` and `hasSharedLibraries` is that:
+
+* isStatic: packages should be statically linked against the dependencies when
+  possible.
```
roberth, line 120:
```suggestion
  possible. Dynamic linking may or may not be available.
```

roberth (commented):
Just proofreading. I'm not an expert on the topic, but I think you're making sense.

roberth:
`hasSharedLibraries` sounds like a property of a package, rather than a property of the platform. Would `hasDynamicLinker` be better?
(again, not an expert, so use your own judgement)

pkgs/top-level/stage.nix
```diff
@@ -269,14 +274,15 @@ let
...
-    # Currently uses Musl on Linux (couldn’t get static glibc to work).
+    # Fully static packages without shared libraries support.
+    # Currently uses Musl on Linux.
     pkgsStatic = nixpkgsFun ({
       overlays = [ (self': super': {
         pkgsStatic = super';
       })] ++ overlays;
       crossSystem = {
         isStatic = true;
+        hasSharedLibraries = false;
```
alyssais, line 285: Does dlopen not work in pkgsStatic?  IIRC, we can build shared objects for aarch64, but not x86_64, or something like that.

tie: Hm, perhaps renaming `isStatic` to `hasStaticLibraries` (in addition to `hasSharedLibraries`) and adding a new `hasDynamicLinker` attribute would be better (with appropriate asserts, i.e. `hasStaticLibraries || hasSharedLibraries`, `hasSharedLibraries -> hasDynamicLinker`)?

alyssais: I don't think that would make sense, because every environment has static libraries, including the default ones.

PR #324071, opened 2024-07-02
lib/systems: use execline’s exec instead of runtimeShell

tie:
## Description of changes

Changes platform’s emulator to `execline`’s `exec` program instead of `runtimeShell`. This makes it easier to use emulator in contexts where only a program path is accepted, but also avoids some POSIX shell quirks (see [POSIX.1-2024](https://pubs.opengroup.org/onlinepubs/9799919799/utilities/V3_chap02.html#tag_19_09_01_06:~:text=due%20to%20an%20error%20equivalent%20to%20the%20%5BENOEXEC%5D%20error%2C,shell%20invoked%20with%20the%20command%20name%20as%20its%20first%20operand)).

Tested on aarch64-linux with
```
nix build --file pkgs/top-level/release-small.nix
nix build --file . librsvg
```

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [x] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/lib/systems/default.nix
+++ b/lib/systems/default.nix
@@ -282,8 +282,11 @@ let
             };
             wine = (pkgs.winePackagesFor "wine${toString final.parsed.cpu.bits}").minimal;
           in
+          # Note: we guarantee that the return value is either `null` or a path
+          # to an emulator program. That is, if an emulator requires additional
+          # arguments, a wrapper should be used.
           if pkgs.stdenv.hostPlatform.canExecute final
-          then "${pkgs.runtimeShell} -c '\"$@\"' --"
+          then "${pkgs.execline}/bin/exec"
           else if final.isWindows
           then "${wine}/bin/wine${optionalString (final.parsed.cpu.bits == 64) "64"}"
           else if final.isLinux && pkgs.stdenv.hostPlatform.isLinux && final.qemuArch != null


tie:
@Ericson2314, @alyssais, @roberth, ping 👀
The PR has been ready for review for 3 weeks…

roberth:
Sorry about that. I can't always keep up.

Did you try `env --`?
The `--` syntax is also available on darwin.

If that's not a good fit, I'm quite fond of this little program, which I think should always have existed; a simple building block like the `true` and `false` programs, so then I would suggest to make it a proper package, which is easy to find for users when they encounter it in an error message.

tie:
The issue with `${coreutils}/bin/env --` is that, ideally, we want selectEmulator to return an _executable path_ instead of an unstructured string with some extra arguments appended—this just makes it easier to use in some contexts. We’d need some wrapping code to hide the arguments in selectEmulator, so might as well use a program that does execvp and nothing else for `canExecute` branch to avoid unnecessary layers of abstractions and indirection.

That said, I’m not really sure why emulator is a property of the platform and not a package set, i.e. `stdenv.hostPlatform.emulator buildPackages` instead of `buildPackages.emulatorFor stdenv.hostPlatform` with `buildPackages.hostEmulator` defined for convenience.

>so then I would suggest to make it a proper package, which is easy to find for users when they encounter it in an error message.

Sounds good to me, should it be in `by-name` or `build-support`?

roberth:
I see. 

> I’m not really sure why emulator is a property of the platform and not a package set

Me neither. It couples `lib`, which was supposed to be quite independent, to the package set.
We're already trying to get rid of functions in platforms so that equality works, so yeah, I think you're on to something.



> `by-name` or `build-support`?

I see no reason not to use `by-name`, because it can be a proper package.

I did notice that it was had two names, `doexec` and `exec-emulator`.
I think a name like `doexec` makes it more widely applicable.
What do you think of a name like `bin-exec` or `exec-args`?

tie:
Moved to `pkgs/by-name/bi/bin-exec`.

alyssais:
I think we could use the `exec` binary from the execline package.

alyssais:
> this just makes it easier to use in some contexts

What contexts?  Are we sure it's worth it?

Ericson2314 (approved):

alyssais (changes requested):
I think we probably don't need to duplicate the [exec](https://skarnet.org/software/execline/exec.html) program we already have packaged — the package containing it is tiny.

tie:
>I think we probably don't need to duplicate the [exec](https://skarnet.org/software/execline/exec.html) program we already have packaged — the package containing it is tiny.

@alyssais, sounds reasonable. Another motivation for a `bin-exec` program was to allow arguments starting with `-` without wrappers (see https://github.com/NixOS/nixpkgs/pull/324071#issuecomment-2247771353), but currently this won’t work for other emulators (e.g. IIRC wine doesn’t support `--`) so we might as well use execline (although it’s a bit weird that it has its own execvpe implementation).

alyssais (approved):

tie:
@roberth, ping 😅

roberth:
It's probably ok, but how do we test this?
A code comment about that would be helpful, making it easy for any reviewer to test this and merge PRs like this confidently with little effort.

tie:
>It's probably ok, but how do we test this?

@roberth, I don’t think we have any specific tests for emulators right now. I’ve tested the PR with `nix build --file pkgs/top-level/release-small.nix` rebuild on `aarch64-linux` and also `nix build --file . librsvg` (that [uses an emulator in native case](https://github.com/NixOS/nixpkgs/blob/669ce8b640c2ac856f714897aaffc5feb614ad57/pkgs/development/libraries/librsvg/default.nix#L153-L163) as well), but it’s a very brute-force way to test a change.

>>I’m not really sure why emulator is a property of the platform and not a package set
>
>Me neither. It couples lib, which was supposed to be quite independent, to the package set.
>We're already trying to get rid of functions in platforms so that equality works, so yeah, I think you're on to something.

I currently have a few non-trivial refactor PRs open and I don’t want to take on another tree wide change, but I’d like work on moving emulator selection functions from lib to pkgs (and add some tests along the way) once I’ll have enough time for this, unless somebody picks this up before me.

roberth (approved):

Artturin:
Adds 110 paths to drv deps in some cases https://github.com/NixOS/nixpkgs/pull/351768

PR #272177, opened 2023-12-05
buildPythonPackage: Add support for running tests in a separate derivation

adisbladis:
## Description of changes
Introduces a new argument to `buildPythonPackage` called `separateChecks`.

A `passthru.tests.python` attribute is added where the tests are run separately from building the package.

By removing check inputs from the main derivation we can drastically reduce build closures & reduce issues with circular dependencies.
For more motivations see the related tracking issue.

This work is a part of #272178.

### Implications

With tests no longer running in the same derivation as the build it can be harder to find issues early.

OTOH debug cycles can be cut shorter as you don't need to wait for a package rebuild to find out that a test failed because it lacked networking access.
With tests running separately I expect iteration cycles to be quicker, once the new workflow has been adopted.

`passthru.tests` is evaluated by OfBorg. Checks for nixpkgs contributions should be fine.
It's possible we may want to add another Hydra attrset so it can find tests, or make Hydra smart enough to explore `passthru.tests`.

In this PR only one derivation has been converted as a PoC. Locally I have converted many more and also found some issues.
This seems to work out fine for most pure Python packages, but Pandas was problematic. I think that particular issue is because of a bug in pytest that I will take a closer look at.

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
- For non-Linux: Is sandboxing enabled in `nix.conf`? (See [Nix manual](https://nixos.org/manual/nix/stable/command-ref/conf-file.html))
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:
  - [NixOS test(s)](https://nixos.org/manual/nixos/unstable/index.html#sec-nixos-tests) (look inside [nixos/tests](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests))
  - and/or [package tests](https://nixos.org/manual/nixpkgs/unstable/#sec-package-tests)
  - or, for functions and "core" functionality, tests in [lib/tests](https://github.com/NixOS/nixpkgs/blob/master/lib/tests) or [pkgs/test](https://github.com/NixOS/nixpkgs/blob/master/pkgs/test)
  - made sure NixOS tests are [linked](https://nixos.org/manual/nixpkgs/unstable/#ssec-nixos-tests-linking) to the relevant packages
[...]

PR #18143, opened 2016-08-31
Python: split buildPythonPackage into two functions

FRidh:
###### Motivation for this change

Currently we have only `buildPythonPackage` for setuptools and wheels installations. However, we still have quite some packages that do not use setuptools or wheels. This PR provides another function, `mkPythonDerivation`, that 
- sets `pythonPath` so `buildEnv` will function
- adds the `python` interpreter which was typically missing
- wraps the Python programs

thus solving issues like described in https://github.com/NixOS/nixpkgs/issues/17679.

A rebuild is needed.

PR #324124, opened 2024-07-02
`buildPythonPackage` , `buildPythonApplication`: take `installCheck`-related attributes directly

ShamrockLee:
## Description of changes

This draft PR enables `buildPythonPackage` and `buildPythonApplicaiton` to directly take the `installCheck`- related attribute instead of specifying it using `check`- associated arguments.

Following the community consensus to merge `lib.extendMkDerivation`, there are ongoing efforts to make `buildPythonPackage` and `buildPythonApplication` take  fixed-point arguments (`(finalAttrs: { })`) and supports the unified overrider `<pkg>.overrideAttrs`.

As the finalAttrs are provided by stdenv.mkDerivation, it contains `doInstallCheck` and `installCheckInputs` instead of `doCheck` and `checkInput`. Should `buildPythonPackage` continues to take only the `checkPhase`-related arguments and map them to the `installCheckPhase`-related ones, it creates inconsistency between the specification and reference of the same attribute.

For example, with fixed-point arguments and `overrideAttrs`, there will be package definitions like:

```nix
{
  buildPythonPackage
}:
buildPythonPackage (finalAttrs: {
  doCheck = false;
  checkInputs = lib.optionals finalAttrs.doInstallCheck [
    # installCheckInputs elements
  ];
  passthru = {
    tests = {
      python-check = finalAttrs.finalPackage.overrideAttrs {
        doInstallCheck = true;
      };
    };
  };
});
```

This PR will also fulfill the to-do stated in [the `mk-python-derivation` comment](https://github.com/NixOS/nixpkgs/blob/2d49e5c54ef320820066a3269321dc21957cdb99/pkgs/development/interpreters/python/mk-python-derivation.nix#L340) to deprecate the use of `checkPhase` to specify `installCheckPhase`, and use `installCheckPhase` directly.

This PR depends on
- #379637

which is a staging change.

After the depending PR merged and this PR rebased, it should create only a few rebuilds.

## Things done

- Built on platform(s)
  - [X] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/doc/languages-frameworks/python.section.md
+++ b/doc/languages-frameworks/python.section.md
@@ -127,7 +127,7 @@ buildPythonPackage rec {
     pluggy
   ];
 
-  nativeCheckInputs = [ hypothesis ];
+  nativeInstallCheckInputs = [ hypothesis ];
 
   meta = {
     changelog = "https://github.com/pytest-dev/pytest/releases/tag/${version}";
@@ -153,8 +153,11 @@ The `buildPythonPackage` mainly does four things:
   environment variable and add dependent libraries to script's `sys.path`.
 * In the [`installCheck`](#ssec-installCheck-phase) phase, `${python.interpreter} -m pytest` is run.
 
-By default tests are run because [`doCheck = true`](#var-stdenv-doCheck). Test dependencies, like
-e.g. the test runner, should be added to [`nativeCheckInputs`](#var-stdenv-nativeCheckInputs).
+By default, tests are run because [`doInstallCheck = true`](#var-stdenv-doInstallCheck). Test dependencies, such as
+the test runner, should be added to [`nativeInstallCheckInputs`](#var-stdenv-nativeInstallCheckInputs).
+
+Python packages and applications constructed with `buildPythonPackage` or `buildPythonApplication` don't have [`checkPhase`](#ssec-check-phase).
+While `doCheck`, `checkPhase`, `checkInputs`, and `nativeCheckInputs` are aliases to the `installCheck`-related attributes for historical reason, we encourage using the `installCheck`-related attributes.
 
 By default `meta.platforms` is set to the same value
 as the interpreter unless overridden otherwise.
@@ -217,8 +220,8 @@ because their behaviour is different:
 * `buildInputs ? []`: Build and/or run-time dependencies that need to be
   compiled for the host machine. Typically non-Python libraries which are being
   linked.
-* `nativeCheckInputs ? []`: Dependencies needed for running the [`checkPhase`](#ssec-check-phase). These
-  are added to [`nativeBuildInputs`](#var-stdenv-nativeBuildInputs) when [`doCheck = true`](#var-stdenv-doCheck). Items listed in
+* `nativeInstallCheckInputs ? []`: Dependencies needed for running the [`installCheckPhase`](#ssec-installCheck-phase). These
+  are added to [`nativeBuildInputs`](#var-stdenv-nativeBuildInputs) when [`doInstallCheck = true`](#var-stdenv-doInstallCheck). Items listed in
   `tests_require` go here.
 * `dependencies ? []`: Aside from propagating dependencies,
   `buildPythonPackage` also injects code into and wraps executables with the
@@ -928,7 +931,7 @@ buildPythonPackage rec {
   build-system = [ setuptools ];
 
   # has no tests
-  doCheck = false;
+  doInstallCheck = false;
 
   pythonImportsCheck = [
     "toolz.itertoolz"
@@ -950,7 +953,7 @@ it accepts a set. In this case the set is a recursive set, `rec`. One of the
 arguments is the name of the package, which consists of a basename (generally
 following the name on PyPI) and a version. Another argument, `src` specifies the
 source, which in this case is fetched from PyPI using the helper function
-`fetchPypi`. The argument `doCheck` is used to set whether tests should be run
+`fetchPypi`. The argument `doInstallCheck` is used to set whether tests should be run
 when building the package. Since there are no tests, we rely on [`pythonImportsCheck`](#using-pythonimportscheck)
 to test whether the package can be imported. Furthermore, we specify some meta
 information. The output of the function is a derivation.
@@ -985,7 +988,7 @@ with import <nixpkgs> { };
       build-system = [ python313.pkgs.setuptools ];
 
       # has no tests
-      doCheck = false;
+      doInstallCheck = false;
 
       meta = {
         homepage = "https://github.com/pytoolz/toolz/";
@@ -1024,7 +1027,7 @@ Our example, `toolz`, does not have any dependencies on other Python packages or
 - `dependencies` - For Python runtime dependencies.
 - `build-system` - For Python build-time requirements.
 - [`buildInputs`](#var-stdenv-buildInputs) - For non-Python build-time requirements.
-- [`nativeCheckInputs`](#var-stdenv-nativeCheckInputs) - For test dependencies
+- [`nativeInstallCheckInputs`](#var-stdenv-nativeInstallCheckInputs) - For test dependencies
 
 Dependencies can belong to multiple arguments, for example if something is both a build time requirement & a runtime dependency.
 
@@ -1067,7 +1070,7 @@ buildPythonPackage rec {
     python-dateutil
   ];
 
-  nativeCheckInputs = [ pytestCheckHook ];
+  nativeInstallCheckInputs = [ pytestCheckHook ];
 
   meta = {
     changelog = "https://github.com/blaze/datashape/releases/tag/${version}";
@@ -1079,8 +1082,8 @@ buildPythonPackage rec {
 ```
 
 We can see several runtime dependencies, `numpy`, `multipledispatch`, and
-`python-dateutil`. Furthermore, we have [`nativeCheckInputs`](#var-stdenv-nativeCheckInputs) with `pytestCheckHook`.
-`pytestCheckHook` is a test runner hook and is only used during the [`checkPhase`](#ssec-check-phase) and is
+`python-dateutil`. Furthermore, we have [`nativeInstallCheckInputs`](#var-stdenv-nativeInstallCheckInputs) with `pytestCheckHook`.
+`pytestCheckHook` is a test runner hook and is only used during the [`installCheckPhase`](#ssec-installCheck-phase) and is
 therefore not added to `dependencies`.
 
 In the previous case we had only dependencies on other Python packages to consider.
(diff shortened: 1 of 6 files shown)


doc/languages-frameworks/python.section.md
```diff
@@ -144,8 +144,10 @@ The `buildPythonPackage` mainly does four things:
   environment variable and add dependent libraries to script's `sys.path`.
 * In the [`installCheck`](#ssec-installCheck-phase) phase, `${python.interpreter} -m pytest` is run.
 
-By default tests are run because [`doCheck = true`](#var-stdenv-doCheck). Test dependencies, like
-e.g. the test runner, should be added to [`nativeCheckInputs`](#var-stdenv-nativeCheckInputs).
+By default tests are run because [`doInstallCheck = true`](#var-stdenv-doInstallCheck). Test dependencies, like
+e.g. the test runner, should be added to [`nativeInstallCheckInputs`](#var-stdenv-nativeInstallCheckInputs).
+
+Python packages and applicationst doesn't have [`checkPhase`](#ssec-check-phase). `doCheck`, `checkPhase`, `checkInputs`, and `nativeCheckInputs` are aliases to the `installCheck`-related attributes for historical reason, but using them instead of the `installCheck`-related variables are obsolete and discouraged.
```
afh, line 150: Minor spelling and grammar suggestion:
```suggestion
Python packages and applications don't have [`checkPhase`](#ssec-check-phase). `doCheck`, `checkPhase`, `checkInputs`, and `nativeCheckInputs` are aliases to the `installCheck`-related attributes for historical reason, but using them instead of the `installCheck`-related variables are obsolete and discouraged.
```

doc/languages-frameworks/python.section.md
```diff
@@ -1936,9 +1938,9 @@ Both are also exported in `nix-shell`.
 It is recommended to test packages as part of the build process.
 Source distributions (`sdist`) often include test files, but not always.
 
-The best practice today is to pass a test hook (e.g. pytestCheckHook, unittestCheckHook) into nativeCheckInputs.
-This will reconfigure the checkPhase to make use of that particular test framework.
-Occasionally packages don't make use of a common test framework, which may then require a custom checkPhase.
+The best practice today is to pass a test hook (e.g. pytestCheckHook, unittestCheckHook) into nativeInstallCheckInputs.
+This will reconfigure the installCheckPhase to make use of that particular test framework.
+Occasionally packages don't make use of a common test framework, which may then require a custom installCheckPhase.
```
afh, line 1943: Minor formatting and linking suggestions:
```suggestion
The best practice today is to pass a test hook (e.g. `pytestCheckHook`, `unittestCheckHook`) into [`nativeInstallCheckInputs`](#var-stdenv-nativeInstallCheckInputs).
This will reconfigure the `installCheckPhase` to make use of that particular test framework.
Occasionally packages don't make use of a common test framework, which may then require a custom `installCheckPhase`.
```

afh (commented):
Thanks for continuing the work on this, @ShamrockLee, please find below some minor suggestions…

mweinelt:
I believe to understand the distinction between `checkPhase` and `installCheckPhase` in the meaning of the stdenv chapter in the manual. What I don't understand is why Python adopted one over the other, and what the motivation and benefit of rolling that decision back is.

ShamrockLee:
@mweinelt The issue is that we are
- specifying the `installCherkPhase` attributes using the `checkPhase` arguments of `buildPython*` and
- pretending that we are using `checkPhase` in the corresponding section of Nixpkgs Manual.

It is not about the choice between `chcekPhase` and `installCheckPhase`, but to inform the choice to users of the bud helpers.

mweinelt:
Why then not move us into `checkPhase` if it is all the same? I don't see the value in adopting another lingo when a whole bunch of packages have already adopted it.

ShamrockLee:
That's indeed another way to go.

The decision to implement `doCheck` with  `doInstallCheck` dates back to 2016, in commit 3e05cce97ca3ea9eee42a17c8bc0b89345b5d0b6 made by @FRidh in #18143. The comment mentioning the long-term goal to move to `installCheckPhase` was added in 2019 in commit f7e28bf5d8181926e600a222cb70180519d09726.

natsukium:
> The issue is that we are
>
> - specifying the installCherkPhase attributes using the checkPhase arguments of buildPython* and
> - pretending that we are using checkPhase in the corresponding section of the Nixpkgs Manual.
>
> It is not about choosing between chcekPhase and installCheckPhase but about informing users of the build helpers about this choice.

The manual notes that checkPhase is used instead of installCheckPhase.
However, I do not know why we adopted checkPhase and not installCheckPhase as it was.
Perhaps because installCheckPhase was rarely used and unfamiliar to most people?

> The [checkPhase](https://github.com/NixOS/nixpkgs/blob/master/doc/languages-frameworks/python.section.md#ssec-check-phase) for python maps to the installCheckPhase on a normal derivation. This is due to many python packages not behaving well to the pre-installed version of the package. Version info, and natively compiled extensions generally only exist in the install directory, and thus can cause issues when a test suite asserts on that behavior.

https://github.com/NixOS/nixpkgs/blob/master/doc/languages-frameworks/python.section.md#testing-python-packages-testing-python-packages

FRidh:
> I get that `installCheckPhase` runs post-install and is more useful for the reasons mentioned.
> 
> The frustrating thing here is, that we simplified the lingo to `build-system` and `dependencies`, only to move from `checkInputs` to `nativeCheckInputs`, and now to `nativeInstallCheckInputs`. I get that all of these bits and pieces have meaning, but we'll be dealing with the resulting churn for the next two releases at least.
> 
> And then there is the matter of [nix-community/nix-init#419](https://github.com/nix-community/nix-init/pull/419) 🫠

Build system and dependencies are essentially another layer on top. The underlying stdenv atttributes will always be needed for all other cases.

I think motivation at the time was consistency. Probably it was not given much thought.

nixos/doc/manual/release-notes/rl-2411.section.md
```diff
@@ -193,6 +193,10 @@
 
 - The `stackclashprotection` hardening flag has been added, though disabled by default.
 
+- `buildPythonPackage` and `buildPythonApplication` now take `doCheck`-related attributes (`doInstallCheck`, `installCheckPhase`, `installCheckInputs`, `nativeInstallCheckInputs`) directly, which precedes the old approach of specifying them via the `check`-related arguments (`doCheck`, `checkPhase`, `checkInputs`, `nativeCheckInputs`).
+
+    Existing Python packages and applications will stay unchanged as long as they don't specify `doCheck` and `doInstallCheck` differently. Such specification is irrelevent both before and after this change, but the effect would be different.
```
mweinelt, line 198:
```suggestion
- The `buildPythonPackage` and `buildPythonApplication` functions now consume the `installCheckPhase` attributes (`doInstallCheck`, `installCheckPhase`, `installCheckInputs`, `nativeInstallCheckInputs`) directly. `checkPhase` attributes (`doCheck`, `checkPhase`, `checkInputs`, `nativeCheckInputs`) have been aliases to their `installCheckPhase` counterparts since at least 2016 and their usage is now deprecated.
```

pkgs/development/interpreters/python/mk-python-derivation.nix
```diff
@@ -312,15 +312,15 @@ let
...
     nativeInstallCheckInputs = [
     ] ++ optionals (format' == "setuptools") [
       # Longer-term we should get rid of this and require
       # users of this function to set the `installCheckPhase` or
       # pass in a hook that sets it.
       setuptoolsCheckHook
-    ] ++ nativeCheckInputs;
-    installCheckInputs = checkInputs;
+    ] ++ nativeInstallCheckInputs;
+    inherit installCheckInputs;
```
mweinelt, line 323:
```suggestion

    inherit installCheckInputs;
```

pkgs/development/interpreters/python/python2/mk-python-derivation.nix
```diff
@@ -209,15 +209,15 @@ let
...
     nativeInstallCheckInputs = [
     ] ++ lib.optionals (format == "setuptools") [
       # Longer-term we should get rid of this and require
       # users of this function to set the `installCheckPhase` or
       # pass in a hook that sets it.
       setuptoolsCheckHook
-    ] ++ nativeCheckInputs;
-    installCheckInputs = checkInputs;
+    ] ++ nativeInstallCheckInputs;
+    inherit installCheckInputs;
```
mweinelt, line 220:
```suggestion

    inherit installCheckInputs;
```

natsukium:
After some thought, I'm personally against the migration to `installCheckPhase` for now.
It has been over six months since we started the switch to `build-system/dependencies`, and this migration is painful.

Indeed, this change is correct and should be done to ensure consistency with the phase definition.
However, it will be more difficult to encourage contributors to migrate just for that reason, even if it is backwards compatible.

Regarding `installCheckPhase`, I feel that many of its functions have delegated `passthru.tests`, such as `tester.testVersion`.
It is no exception in python, where https://github.com/NixOS/nixpkgs/pull/272177 tries to separate build and test.
I don't know how it will finally be implemented, but it seems that it would be less burdensome if this PR change were adopted at the same time as the separation.

ShamrockLee:
> After some thought, I'm personally against the migration to `installCheckPhase` for now.
> It has been over six months since we started the switch to `build-system/dependencies`, and this migration is painful.
>
> Indeed, this change is correct and should be done to ensure consistency with the phase definition.
> However, it will be more difficult to encourage contributors to migrate just for that reason, even if it is backwards compatible.

Thank you for sharing your precious experiences about attribute renaming.

Your right. The change wouldn't be helpful to the community if the pain caused by the mass renaming outweighs the benefits of the name correctness.

> Regarding `installCheckPhase`, I feel that many of its functions have delegated `passthru.tests`, such as `tester.testVersion`.
> It is no exception in python, where #272177 tries to separate build and test.
> I don't know how it will finally be implemented, but it would be less burdensome if this PR change were adopted at the same time as the separation.

Delegating checks to `passthru.test` (as package tests) sounds like the right way to go. Let's push that one first.

If this is not the change we want to make now, should we close this pull request or mark it as a draft?

SandaruKasa:
Is it still/already relevant?

ShamrockLee:
This PR is surpassed by PR #469804, closing.

PR #469804 tells `makeOverridablePythonAttrs` (i.e., `<pkg>.overridePythonAttrs`) to handle both `checkPhase` and `installCheckPhase`, while passing only `installCheckPhase` to `stdenv.mkDerivation`.
Such approach is much more flexible and backward-compatible while ensuring the attribute correctness of `finalAttrs`.

I'll move the documentation change from this PR to PR #469804 or subsequent PRs.

- #469804

@SandaruKasa

PR #324155, opened 2024-07-02
treewide: change various flags to allow x64 darwin to default to sdk 11.0 when ready

paparodeo:
## Description of changes

continuation of https://github.com/NixOS/nixpkgs/pull/323679 which i closed and can not re-open.

with this change and plus a modification to `lib/systems/default.nix` x64 darwin will default to sdk 11.0.

the first commit modifies various conditionals to not assume that x86_64-darwin uses sdk 10.12 and instead checks the darwinSdkVersion. Additionally, darwin.xnu is modified so x64 darwin uses unwrapped clang when generating header files, as the headers require a 32 bit arch which is not allowed when using cc-wrapper as it sets -arch to x86_64. this commit is sufficient for building stdenv when `darwinSdkVersion` is set to "11.0" in `lib/systems/default.nix`.

the second commit fixes evaluation errors which assume that x64 darwin can access the `apple_sdk.sdk` attribute which is only available as part of the 10.12 sdk so rather than checking for x64 darwin, condition on `darwinSdkVersion` is older than "11.0".

with these two commits and the following diff x64 darwin will default to using sdk 11.0
<details><summary>enable sdk 11 on all platforms</summary>

```diff
--- a/lib/systems/default.nix
+++ b/lib/systems/default.nix
@@ -251,7 +251,7 @@ let
         else null;
       # The canonical name for this attribute is darwinSdkVersion, but some
       # platforms define the old name "sdkVer".
-      darwinSdkVersion = final.sdkVer or (if final.isAarch64 then "11.0" else "10.12");
+      darwinSdkVersion = final.sdkVer "11.0";
       darwinMinVersion = final.darwinSdkVersion;
       darwinMinVersionVariable =
         if final.isMacOS then "MACOSX_DEPLOYMENT_TARGET"

```
</details>

testing: using the patch to enable sdk 11, built inkscape and verified removing overrideSDK for a few sampled packages (vcpkg-tool, wasmedge) works.

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [x] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

cc: @emilazy @reckenrode


pkgs/os-specific/darwin/pam-reattach/default.nix
```diff
@@ -19,10 +19,10 @@ stdenv.mkDerivation rec {
         "arm64"
     }"
     "-DENABLE_CLI=ON"
-  ] ++ lib.optional (!stdenv.isAarch64) "-DCMAKE_LIBRARY_PATH=${darwin.apple_sdk.sdk}/usr/lib";
+  ] ++ lib.optional  (darwin.apple_sdk ? sdk) "-DCMAKE_LIBRARY_PATH=${darwin.apple_sdk.sdk}/usr/lib";
 
   buildInputs = [ openpam ]
-    ++ lib.optional (!stdenv.isAarch64) darwin.apple_sdk.sdk;
+    ++ lib.optional (darwin.apple_sdk ? sdk) darwin.apple_sdk.sdk;
```
reckenrode, line 25: pam-reattach builds for me without adding the SDK, so I think it can be dropped. `CMAKE_OSX_ARCHITECTURES` can be defined as `"-DCMAKE_OSX_ARCHITECTURES=${stdenv.hostPlatform.darwinArch}"` (but GitHub won’t let me comment on those lines). This is the patch I used locally to build it (with some extra changes to `cmakeFlags`):

```patch
diff --git a/pkgs/os-specific/darwin/pam-reattach/default.nix b/pkgs/os-specific/darwin/pam-reattach/default.nix
index 4350865080f0..4a20d7dd93f8 100644
--- a/pkgs/os-specific/darwin/pam-reattach/default.nix
+++ b/pkgs/os-specific/darwin/pam-reattach/default.nix
@@ -12,17 +12,11 @@ stdenv.mkDerivation rec {
   };
 
   cmakeFlags = [
-    "-DCMAKE_OSX_ARCHITECTURES=${
-      if stdenv.hostPlatform.system == "x86_64-darwin" then
-        "x86_64"
-      else
-        "arm64"
-    }"
-    "-DENABLE_CLI=ON"
-  ] ++ lib.optional (!stdenv.isAarch64) "-DCMAKE_LIBRARY_PATH=${darwin.apple_sdk.sdk}/usr/lib";
+    (lib.cmakeFeature "CMAKE_OSX_ARCHITECTURES" stdenv.hostPlatform.darwinArch)
+    (lib.cmakeBool "ENABLE_CLI" true)
+  ];
 
-  buildInputs = [ openpam ]
-    ++ lib.optional (!stdenv.isAarch64) darwin.apple_sdk.sdk;
+  buildInputs = [ openpam ];
 
   nativeBuildInputs = [ cmake ];
 
```

pkgs/development/compilers/llvm/13/default.nix
```diff
@@ -236,7 +236,7 @@ in let
         # https://github.com/NixOS/nixpkgs/pull/194634#issuecomment-1272129132
         ++ lib.optional (
           stdenv.targetPlatform.isDarwin
-            && !stdenv.targetPlatform.isAarch64
+            && darwin.apple_sdk ? sdk
```
reckenrode, line 239: The LLVM checks can be replaced with `stdenv.targetPlatform.isDarwin && lib.versionOlder stdenv.targetPlatform.darwinSdkVersion "11.0"`. Even when using `overrideSDK` and after the refactor, that version will always match the version of the SDK in use.

paparodeo: most of the llvm checks, like this one, also check for `(lib.versionOlder darwin.apple_sdk.sdk.version "11.0")` which I assume can be dropped when using `lib.versionOlder stdenv.targetPlatform.darwinSdkVersion "11.0"`?

reckenrode: Also for what it’s worth, I’m pretty sure this will compile on the 10.12 SDK after the ld64 update. Libsystem pulls the affected header from ld64, which will be considerably newer than the 10.12 SDK. I actually consider that a bug, but there’s no nice way to fix it right now.

pkgs/development/libraries/webkitgtk/default.nix
```diff
@@ -151,7 +151,7 @@ stdenv.mkDerivation (finalAttrs: {
   ] ++ lib.optionals stdenv.isDarwin [
     libedit
     readline
-  ] ++ lib.optional (stdenv.isDarwin && !stdenv.isAarch64) (
+  ] ++ lib.optional (stdenv.isDarwin && apple_sdk ? sdk) (
```
reckenrode, line 154: Is this still necessary? `libproc.h` is included in the 10.12 SDK Libsystem and should be identical to this one.

pkgs/os-specific/darwin/apple-source-releases/xnu/default.nix
```diff
@@ -72,15 +72,19 @@ appleDerivation' (if headersOnly then stdenvNoCC else stdenv) (
   HOST_FLEX = "flex";
   HOST_BISON = "bison";
   HOST_GM4 = "m4";
-  MIGCC = "cc";
+  # use unwrapped clang to generate headers because wrapper is not compatible with a 32 bit -arch.
+  # aarch64 should likely do this as well and remove the --replace MACHINE_ARCH above
+  MIGCC = if stdenv.isx86_64 && lib.versionAtLeast stdenv.hostPlatform.darwinSdkVersion "11"
+    then "${lib.getBin buildPackages.stdenv.cc.cc}/bin/clang"
```
reckenrode, line 78: `mig` needs to target the build platform because it’s run at build time to generate headers.

```suggestion
    then "${lib.getBin pkgsBuildBuild.stdenv.cc.cc}/bin/clang"
```

pkgs/os-specific/darwin/apple-source-releases/xnu/default.nix
```diff
@@ -150,7 +154,9 @@ appleDerivation' (if headersOnly then stdenvNoCC else stdenv) (
     mv $out/Library/Frameworks/IOKit.framework $out/Library/PrivateFrameworks
   '';
 
-  appleHeaders = builtins.readFile (./. + "/headers-${arch}.txt");
+  appleHeaders = let
+    name = if stdenv.hostPlatform.darwinSdkVersion == "10.12" then "headers-10.12-${arch}.txt" else "headers-${arch}.txt";
```
reckenrode, line 158: Couldn’t this just be `headers-10.12.txt` and `headers-11.0.txt`?

pkgs/development/compilers/llvm/13/default.nix
reckenrode: > most of the llvm checks, like this one, also check for `(lib.versionOlder darwin.apple_sdk.sdk.version "11.0")` which I assume can be dropped when using `lib.versionOlder stdenv.targetPlatform.darwinSdkVersion "11.0"`?

Right. I only commented on this one, but all the identical checks could be replaced in the same way.

reckenrode:
Also pinging @toonn.

pkgs/os-specific/darwin/apple-source-releases/xnu/default.nix
paparodeo: it can. i thought that the naming would cause rebuilds but it doesn't so i can align everything to use headers-${sdk-version}-${arch}.txt

pkgs/os-specific/darwin/apple-source-releases/xnu/default.nix
reckenrode: I’d also add this can be done unconditionally. I do something similar in https://github.com/NixOS/nixpkgs/pull/256590/commits/37a87d0d1a5e9dbd1499e744bf68283da48eddb7, but I think this way is probably cleaner. I can’t remember why I didn’t specify a full path to for `MIGCC` in that commit.

pkgs/development/compilers/llvm/13/default.nix
paparodeo: thanks -- i miss-read.

pkgs/os-specific/darwin/apple-source-releases/xnu/default.nix
paparodeo: i don't have aarch64 to test on. i suspect that making it unconditional means that MACHINE_ARCH doesn't need to get patched for aarch64 as well, and thus heads the comment saying that the header build needs to use 32 bit arch. it does change the layouts for the x64 darwin structures so i assume that the aarch64 ones have mismatched structure sizes as well.

i'd like this to be part of a followup given the lack of hardware and this would trigger the commit going into staging.

pkgs/os-specific/darwin/pam-reattach/default.nix
paparodeo: this patch doesn't work for me on sdk 10.12 x64 darwin.
```
/tmp/nix-build-pam_reattach-1.3.drv-0/source/src/xpc.h:29:10: fatal error: 'xpc/xpc.h' file not found
#include <xpc/xpc.h>
```
so i think darwin.apple_sdk.sdk needs to be conditionally included for now.

pkgs/development/libraries/webkitgtk/default.nix
paparodeo: hmm, webkitgtk is marked as broken on darwin so not sure. i can add a comment stating it might not be necessary.

pkgs/os-specific/darwin/pam-reattach/default.nix
toonn: I think this is correct, xpc is currently patched out in the source releases I believe so getting it from the SDK is appropriate.

pkgs/os-specific/darwin/apple-source-releases/xnu/default.nix
```diff
@@ -150,7 +154,7 @@ appleDerivation' (if headersOnly then stdenvNoCC else stdenv) (
     mv $out/Library/Frameworks/IOKit.framework $out/Library/PrivateFrameworks
   '';
 
-  appleHeaders = builtins.readFile (./. + "/headers-${arch}.txt");
+  appleHeaders = builtins.readFile (./. + "/headers-${stdenv.hostPlatform.darwinSdkVersion}-${arch}.txt");
```
toonn, line 157: Will we ever want to split on more than the architecture? I'd say the goal is to unsplit these header checks eventually. If they're even still useful.

toonn (commented):
Most of it looks good to me, just a couple questions.

RossComputerGuy (commented):
LGTM, however I would like to get #325175 in before we merge any other LLVM changes to minimize rebasing and conflicts.

toonn:
@RossComputerGuy, how close is that PR from being marked ready?

RossComputerGuy:
@toonn it's ready, trying to figure out CI failures. Should be ready tonight or within the next couple days.

pkgs/os-specific/darwin/apple-source-releases/xnu/default.nix
reckenrode: The way to do these in the future probably requires us to inventory the upstream SDKs and compare our headers to those in the SDK. Otherwise, using headers + stubs with more than a few SDKs is going to get painful quickly.

(Or I wouldn’t worry too much about the current situation because it’s going to need to change.)

paparodeo:
> @toonn it's ready, trying to figure out CI failures. Should be ready tonight or within the next couple days.

so it is not ready.

RossComputerGuy:
> so it is not ready.

The actual work is done, there's just CI roadblocks. I checked using nix-diff and rebuild count should be 0. It sounds like the CI issue it's having should be simpler enough and only affects Darwin.

paparodeo:
> > so it is not ready.
> 
> The actual work is done, there's just CI roadblocks. I checked using nix-diff and rebuild count should be 0. It sounds like the CI issue it's having should be simpler enough and only affects Darwin.

it only effects the builds which have a stdenv based on llvm. i am not sure how you think it is good etiquette to think that you should be able to open up a non-critical PR that creates merge conflicts with this one and then request that this not get merged so you don't have to resolve the merge conflicts with the files you deleted so I can deal with fixing them in this PR. And your PR doesn't even work yet. Fixing the merge conflicts is not really a big deal for me but your request is just rude.

RossComputerGuy:
> Fixing the merge conflicts is not really a big deal for me but your request is just rude.

Sorry, it wasn't my intention to be rude. I'll merge this one first if nobody has anything left to address.

RossComputerGuy (approved):

paparodeo:
> > Fixing the merge conflicts is not really a big deal for me but your request is just rude.
> 
> Sorry, it wasn't my intention to be rude. I'll merge this one first if nobody has anything left to address.

it doesn't really matter. i overreacted.

wentasah:
Hi,

in the [nix-ros-overlay](https://github.com/lopsided98/nix-ros-overlay/issues/484) project, we're experiencing the following eval failure:
```
error: opening file '/nix/store/l3amk5lsakpc93him5kry24kax23sn4h-source/pkgs/os-specific/darwin/apple-source-releases/xnu/headers-10.12-arm64.txt': No such file or directory
```
Can somebody here suggests what's wrong?

paparodeo:
> Hi,
> 
> in the [nix-ros-overlay](https://github.com/lopsided98/nix-ros-overlay/issues/484) project, we're experiencing the following eval failure:
> 
> ```
> error: opening file '/nix/store/l3amk5lsakpc93him5kry24kax23sn4h-source/pkgs/os-specific/darwin/apple-source-releases/xnu/headers-10.12-arm64.txt': No such file or directory
> ```
> 
> Can somebody here suggests what's wrong?

arm64 doesnt support 10.12

reckenrode:
> Hi,
> 
> in the [nix-ros-overlay](https://github.com/lopsided98/nix-ros-overlay/issues/484) project, we're experiencing the following eval failure:
> 
> ```
> error: opening file '/nix/store/l3amk5lsakpc93him5kry24kax23sn4h-source/pkgs/os-specific/darwin/apple-source-releases/xnu/headers-10.12-arm64.txt': No such file or directory
> ```
> 
> Can somebody here suggests what's wrong?

Somehow `stdenv.hostPlatform.darwinSdkVersion` is being set to 10.12 on aarch64-darwin. The oldest version supported by aarch64-darwin is 11.0, which is the current default.

https://github.com/NixOS/nixpkgs/blob/8ce34b38ecba5be44fe665c6e779beadc1eae2d3/pkgs/os-specific/darwin/apple-source-releases/xnu/default.nix#L152

wentasah:
Thanks for quick answers. I don't think we pin the SDK versions anywhere. Who and how can do that? Can you see something interesting from the trace? Posting just the end, which seems to contain some package names, not just generic nixpkgs stuff:
```
      … in the condition of the assert statement
         at /nix/store/l3amk5lsakpc93him5kry24kax23sn4h-source/pkgs/build-support/cc-wrapper/default.nix:293:1:
          292| # Ensure bintools matches
          293| assert libc_bin == bintools.libc_bin;
             | ^
          294| assert libc_dev == bintools.libc_dev;

       … while calling the 'getAttr' builtin
         at <nix/derivation-internal.nix>:44:19:
           43|       value = commonAttrs // {
           44|         outPath = builtins.getAttr outputName strict;
             |                   ^
           45|         drvPath = strict.drvPath;

         at <nix/derivation-internal.nix>:34:12:
           33|
           34|   strict = derivationStrict drvAttrs;
             |            ^
           35|

       … while evaluating derivation 'Libsystem-1238.60.2'
         whose name attribute is located at /nix/store/l3amk5lsakpc93him5kry24kax23sn4h-source/pkgs/stdenv/generic/make-derivation.nix:336:7

       … while evaluating attribute 'installPhase' of derivation 'Libsystem-1238.60.2'
         at /nix/store/l3amk5lsakpc93him5kry24kax23sn4h-source/pkgs/os-specific/darwin/apple-source-releases/Libsystem/default.nix:36:3:
           35|
           36|   installPhase = ''
             |   ^
           37|     export NIX_ENFORCE_PURITY=

       … while calling the 'getAttr' builtin
         at <nix/derivation-internal.nix>:44:19:
           43|       value = commonAttrs // {
           44|         outPath = builtins.getAttr outputName strict;
             |                   ^
           45|         drvPath = strict.drvPath;

         at <nix/derivation-internal.nix>:34:12:
           33|
           34|   strict = derivationStrict drvAttrs;
             |            ^
           35|

       … while evaluating derivation 'xnu-3789.70.16'
         whose name attribute is located at /nix/store/l3amk5lsakpc93him5kry24kax23sn4h-source/pkgs/stdenv/generic/make-derivation.nix:336:7

       … while evaluating attribute 'appleHeaders' of derivation 'xnu-3789.70.16'
         at /nix/store/l3amk5lsakpc93him5kry24kax23sn4h-source/pkgs/os-specific/darwin/apple-source-releases/xnu/default.nix:152:3:
          151|
          152|   appleHeaders = builtins.readFile (./. + "/headers-${stdenv.hostPlatform.darwinSdkVersion}-${arch}.txt");
             |   ^
          153| } // lib.optionalAttrs headersOnly {

       … while calling the 'readFile' builtin
         at /nix/store/l3amk5lsakpc93him5kry24kax23sn4h-source/pkgs/os-specific/darwin/apple-source-releases/xnu/default.nix:152:18:
          151|
          152|   appleHeaders = builtins.readFile (./. + "/headers-${stdenv.hostPlatform.darwinSdkVersion}-${arch}.txt");
             |                  ^
          153| } // lib.optionalAttrs headersOnly {

       error: opening file '/nix/store/l3amk5lsakpc93him5kry24kax23sn4h-source/pkgs/os-specific/darwin/apple-source-releases/xnu/headers-10.12-arm64.txt': No such file or directory
```

emilazy:
Do you have a flake command that reliably reproduces the failure?

reckenrode:
I added tracing to your flake. It’s failing when it tries to evaluate armv7a-darwin.

wentasah:
I guess, this is some niche platform, which we can disable. But probably it should not look for `headers-*-arm64.txt`.

emilazy:
It’s because your flake includes `armv7a-darwin`, which is not a platform that has ever been supported or really makes much sense beyond unsupported iPhones. You should probably use something more like Nixpkgs’ `flakeExposed` list.

wentasah:
Sounds like a good idea. Thanks to all for help.

reckenrode:
armv7a-darwin is included in `lib.platforms.darwin` for some reason. Maybe it worked in the past, but I can’t imagine it’s worked anytime recently. macOS has never supported 32-bit ARM (and Apple Silicon doesn’t implement it). iOS dropped support in iOS 11, which was released in 2017. Apple stopped accepting 32-bit submissions to the App Store in 2015.

PR #303930, opened 2024-04-14
xpra: 4.4.6 -> 5.0.9

aqrln:
## Description of changes

Update [Xpra](https://xpra.org/) from 4.4.6 to 5.0.8.

Tested:
- Connecting from a remote client on macOS via ssh and running apps in seamless mode
- Running Xpra itself on NixOS: GIU works, running apps from the same host in seamless mode but in Xpra's X server works
- Running `xpra start` and `xpra shadow`: couldn't reproduce the issue reported by @kashw2 in https://github.com/NixOS/nixpkgs/issues/270566#issuecomment-1830737443 on my NixOS 24.05.git.f0154a9104b0 aarch64.

Not tested:
- Desktop mode
- Shadow mode aside from just starting the server and then stopping it
- Xpra display manager NixOS module

TODO: check the changelog and see if there are any breaking changes or incompatibilities in the major release. Add a release notes entry if necessary.

Closes: https://github.com/NixOS/nixpkgs/issues/270566

cc maintainers @offlinehacker @numinit @mvnetbiz 

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [x] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
- For non-Linux: Is sandboxing enabled in `nix.conf`? (See [Nix manual](https://nixos.org/manual/nix/stable/command-ref/conf-file.html))
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:
  - [NixOS test(s)](https://nixos.org/manual/nixos/unstable/index.html#sec-nixos-tests) (look inside [nixos/tests](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests))
  - and/or [package tests](https://github.com/NixOS/nixpkgs/blob/master/pkgs/README.md#package-tests)
  - or, for functions and "core" functionality, tests in [lib/tests](https://github.com/NixOS/nixpkgs/blob/master/lib/tests) or [pkgs/test](https://github.com/NixOS/nixpkgs/blob/master/pkgs/test)
  - made sure NixOS tests are [linked](https://nixos.org/manual/nixpkgs/unstable/#ssec-nixos-tests-linking) to the relevant packages
[...]

PR #324179, opened 2024-07-02
xpra: fix xdg-open path

### LINKED ISSUE #324175: Xpra has hardcoded broken path /usr/bin/xdg-open for open command on NixOS
### Describe the bug
Xpra supports forwarding `xdg-open` invocations from the server to the client. The Xpra client defaults to hardcoding `/usr/bin/xdg-open` as the command to invoke on the client, which does not exist in NixOS.

This bug does not impact hosts other than NixOS.

### Steps To Reproduce
1. `nix run nixpkgs#xpra -- attach ...`
2. Invoke `xdg-open` somehow.

### Expected behavior
The Xpra client should invoke the local `xdg-open`.

### Screenshots
N/A.

### Additional context
This could be fixed by setting the open command to just `xdg-open`, but it might match the intent better to change it to something like `${pkgs.xdg-utils}/bin/xdg-open` perhaps. Presumably, the point of doing it this way is to avoid any other copies of `xdg-open` that might appear on the `$PATH`, e.g. Xpra's own. I suppose we can't really rely on `/run/current-system/sw` to exist: even if we could on NixOS, we can't on other Linux distributions running Nix, where `/usr/bin/xdg-open` might make more sense. It seems to me that the best compromise, then, is to hardcode a path to `xdg-utils` copy of `xdg-open` and users can override this using the `--open-command` option if needed.

The open command that Xpra uses is configurable on the client-side, but the default is set as a platform-specific variable. On Linux, it is set in [`xpra/platform/posix/features.py`](https://github.com/Xpra-org/xpra/blob/dea16aead5be5276e4976f3a1fa4766a3c852d8e/xpra/platform/posix/features.py#L48). Because there are separate values for both Windows and Darwin hosts, patching the value to `xdg-utils` shouldn't ever break them (but of course it should still be a Linux-specific patch, otherwise it unnecessarily brings `xdg-utils` into the closure.)

### Notify maintainers

@offlinehacker @numinit @mvnetbiz

### Metadata

```console
 - system: `"x86_64-linux"`
 - host os: `Linux 6.9.2, NixOS, 24.11 (Vicuna), 24.11.20240627.b2852eb`
 - multi-user?: `yes`
 - sandbox: `yes`
 - version: `nix-env (Nix) 2.18.4`
 - channels(root): `"nixos"`
 - channels(john): `""`
 - nixpkgs: `/nix/store/qgbn0imyridkb9527v6gnv6z3jzzprb9-source`
```

---

jchv:
## Description of changes
Adds a patch to Xpra to move a hardcoded path to `/usr/bin/xdg-open` to `${xdg-utils}/bin/xdg-open`. The rationale for this specific choice is in the additional details at #324175. This fixes the forwarding of `xdg-open` over Xpra.

This patch will need to be updated slightly when Xpra is updated to 5.0 or later, since the `xposix` directory was renamed to `posix` in 5.0. So, assuming #303930 is merged first, this will need to be updated. The update should be trivial though and it won't need to be changed again for 6.0.

Closes #324175.

## Things done
- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/tools/X11/xpra/default.nix
+++ b/pkgs/tools/X11/xpra/default.nix
@@ -29,11 +29,13 @@
 , pango
 , pulseaudio
 , python3
+, stdenv
 , util-linux
 , which
 , x264
 , x265
 , xauth
+, xdg-utils
 , xorg
 , xorgserver
 }:
@@ -83,6 +85,12 @@ in buildPythonApplication rec {
     ./fix-41106.patch  # https://github.com/NixOS/nixpkgs/issues/41106
     ./fix-122159.patch # https://github.com/NixOS/nixpkgs/issues/122159
   ];
+
+  # Note: xposix is renamed to posix in v5.
+  postPatch = lib.optionalString stdenv.isLinux ''
+    substituteInPlace xpra/platform/xposix/features.py \
+      --replace-fail "/usr/bin/xdg-open" "${xdg-utils}/bin/xdg-open"
+  '';
 
   INCLUDE_DIRS = "${pam}/include";
 


pkgs/tools/X11/xpra/default.nix
```diff
@@ -84,6 +86,12 @@ in buildPythonApplication rec {
     ./fix-122159.patch # https://github.com/NixOS/nixpkgs/issues/122159
   ];
 
+  postPatch = lib.optionalString stdenv.isLinux ''
+    # Note: xposix is renamed to posix in v5.
+    substituteInPlace xpra/platform/xposix/features.py \
+      --replace "/usr/bin/xdg-open" "${xdg-utils}/bin/xdg-open"
```
numinit, line 92: --replace-quiet possibly?

https://nixos.org/manual/nixpkgs/stable/#fun-substitute-replace-quiet

numinit (changes requested):

pkgs/tools/X11/xpra/default.nix
eclairevoyant: we should use `--replace-fail`; if it fails, we want to know.

numinit: Fair enough! :+1:

numinit (approved):

pkgs/tools/X11/xpra/default.nix
```diff
@@ -84,6 +86,12 @@ in buildPythonApplication rec {
     ./fix-122159.patch # https://github.com/NixOS/nixpkgs/issues/122159
   ];
 
+  postPatch = lib.optionalString stdenv.isLinux ''
+    # Note: xposix is renamed to posix in v5.
```
Aleksanaa, line 90: Better to put this comment outside

PR #324183, opened 2024-07-03
avdump3: init at 8293_stable

kini:
## Description of changes

Add `avdump3`, a tool for scanning media files for audio/video metadata / fingerprinting, and usually uploading the resulting analysis to [AniDB](https://anidb.net) (a website for tracking anime).

Home page, such as it is: https://wiki.anidb.net/Avdump3
Partial source code, excluding some proprietary parts: https://github.com/DvdKhl/AVDump3
Repology page: https://repology.org/project/avdump3

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
- [x] Tested, as applicable:

---

### DIFF
--- a/pkgs/by-name/av/avdump3/package.nix
+++ b/pkgs/by-name/av/avdump3/package.nix
@@ -1,0 +1,59 @@
+{
+  lib,
+  stdenv,
+  fetchzip,
+  dotnet-runtime,
+  zlib,
+  runtimeShell,
+}:
+
+stdenv.mkDerivation rec {
+  pname = "avdump3";
+  version = "8293_stable";
+
+  src = fetchzip {
+    url = "https://cdn.anidb.net/client/avdump3/avdump3_8293_stable.zip";
+    hash = "sha256-H9Sn3I4S9CmymKIMHVagDy+7svHs285S3EJgYQo+ks0=";
+    stripRoot = false;
+  };
+
+  installPhase = ''
+    runHook preInstall
+    mkdir -p $out/share/avdump3 $out/bin
+    mv * $out/share/avdump3
+    cat > $out/bin/avdump3 <<EOF
+    #!${runtimeShell}
+    export LD_LIBRARY_PATH="${lib.makeLibraryPath [ zlib ]}:\$LD_LIBRARY_PATH"
+    exec ${dotnet-runtime}/bin/dotnet $out/share/avdump3/AVDump3CL.dll "\$@"
+    EOF
+    chmod +x $out/bin/avdump3
+    runHook postInstall
+  '';
+
+  dontPatchELF = true;
+
+  meta = {
+    mainProgram = "avdump3";
+    description = "Tool for extracting audio/video metadata from media files and uploading it to AniDB";
+    longDescription = ''
+      AVDump is a tool to extract meta information from media files while at the
+      same time calculating multiple hashes. Based on that information reports
+      can be generated in multiple forms. Of particular interest is the ability
+      to send those reports back to AniDB and thereby quickly filling in missing
+      metadata for new files.
+    '';
+    homepage = "https://wiki.anidb.net/Avdump3";
+    sourceProvenance = with lib.sourceTypes; [
+      binaryNativeCode
+      binaryBytecode
+    ];
+    # partial source code available under MIT license at https://github.com/DvdKhl/AVDump3
+    license = with lib.licenses; [
+      mit
+      unfree
+    ];
+    maintainers = with lib.maintainers; [ kini ];
+    # NOTE: aarch64-linux may also work but hasn't been tested; co-maintainers welcome.
+    platforms = [ "x86_64-linux" ];
+  };
+}


kini:
Result of `nixpkgs-review pr 324183` run on x86_64-linux [1](https://github.com/Mic92/nixpkgs-review)
<details>
  <ul>
    <li>avdump3</li>
  </ul>
</details>

pkgs/by-name/av/avdump3/package.nix
```diff
@@ -0,0 +1,43 @@
...
+  version = "8293_stable";
+
+  src = fetchzip {
+    url = "https://cdn.anidb.net/client/avdump3/avdump3_8293_stable.zip";
+    hash = "sha256-H9Sn3I4S9CmymKIMHVagDy+7svHs285S3EJgYQo+ks0=";
+    stripRoot = false;
+  };
+
+  installPhase = ''
+    mkdir -p $out/share/${pname} $out/bin
```
Aleksanaa, line 14: Don't reference pname here and there, that doesn't make life better (btw `${pname}` has 8 characters and `avdump3` has 7)

pkgs/by-name/av/avdump3/package.nix
```diff
@@ -0,0 +1,43 @@
...
+    #!/bin/sh
+    export LD_LIBRARY_PATH="${lib.makeLibraryPath [ zlib ]}:\$LD_LIBRARY_PATH"
+    exec ${dotnet-runtime}/bin/dotnet $out/share/${pname}/AVDump3CL.dll "\$@"
+    EOF
+    chmod +x $out/bin/${pname}
+  '';
+
+  dontPatchELF = true;
+
+  meta = with lib; {
```
Aleksanaa, line 26:
```suggestion
  meta = {
```

Better to avoid `with lib`

pkgs/by-name/av/avdump3/package.nix
```diff
@@ -0,0 +1,43 @@
...
+      can be generated in multiple forms. Of particular interest is the ability
+      to send those reports back to AniDB and thereby quickly filling in missing
+      metadata for new files.
+    '';
+    homepage = "https://wiki.anidb.net/Avdump3";
+    sourceProvenance = with sourceTypes; [ binaryNativeCode binaryBytecode ];
+    # partial source code available under MIT license at https://github.com/DvdKhl/AVDump3
+    license = with licenses; [ mit unfree ];
+    maintainers = with maintainers; [ kini ];
+    platforms = [ "x86_64-linux" "aarch64-linux" ];
```
Aleksanaa, line 41: Working for both platforms?

pkgs/by-name/av/avdump3/package.nix
kini: Sure. I wasn't doing that for conciseness but for flexibility in case the derivation needed to be cloned with a different name or something (I've seen this happen with a few packages in the past I think). But that's premature optimization so I'll get rid of it as you suggest.

pkgs/by-name/av/avdump3/package.nix
kini: `x86_64-linux` is working. I don't have an `aarch64-linux` machine to test, but I see there are `.so` files for it in the archive. Any suggestion what to do here? Just mark it as `x86_64-linux` exclusive for now and wait for some ARM user to come along and offer to test it on their platform?

kini: I asked for opinions in the NixOS matrix channel and the consensus seemed to be to restrict to `x86_64-linux` for now, so I updated the commit and pushed. Thanks for pointing it out!

pkgs/by-name/av/avdump3/package.nix
```diff
@@ -0,0 +1,44 @@
...
+  pname = "avdump3";
+  version = "8293_stable";
+
+  src = fetchzip {
+    url = "https://cdn.anidb.net/client/avdump3/avdump3_8293_stable.zip";
+    hash = "sha256-H9Sn3I4S9CmymKIMHVagDy+7svHs285S3EJgYQo+ks0=";
+    stripRoot = false;
+  };
+
+  installPhase = ''
```
Aleksanaa, line 13:
```
runHook preInstall
# Your code
runHook postInstall
```

pkgs/by-name/av/avdump3/package.nix
```diff
@@ -0,0 +1,44 @@
...
+    exec ${dotnet-runtime}/bin/dotnet $out/share/avdump3/AVDump3CL.dll "\$@"
+    EOF
+    chmod +x $out/bin/avdump3
+  '';
+
+  dontPatchELF = true;
+
+  meta = {
+    mainProgram = "avdump3";
+    description = "Tool for extracting audio/video metadata from media files and uploading it to AniDB";
```
Aleksanaa, line 28:
```suggestion
    description = "Extracting audio/video metadata from media files and uploading it to AniDB";
```

kini: Hmm, I don't think I agree with this one. Making the description a gerund (actually two conjoined gerunds, "extracting" and "uploading") sounds awkward to me. Plenty of packages' descriptions follow the pattern of "tool for", "program for", "app for", etc., even looking only at the relatively recent packages in `pkgs/by-name/`.

pkgs/by-name/av/avdump3/package.nix
kini: Oops, forgot to commit this change, sorry! Pushed now.

pkgs/by-name/av/avdump3/package.nix
Frontear: Suggesting an opposite change for this:

```diff
-description = "Tool for extracting audio/video metadata from media files and uploading it to AniDB";
+description = "A tool for extracting audio/video metadata from media files and uploading it to AniDB";
```

kini: At some point there were some style guidelines added to the READMEs: https://github.com/NixOS/nixpkgs/blob/master/pkgs/README.md#meta-attributes

According to those guidelines, the description shouldn't start with an article like "A" or "The". On the other hand the guidelines also say that the description should be a "sentence" but that would require a verb -- doesn't make much sense to me :)

Frontear:
Can you elaborate on why this much stuff is required in the outputs for this app? Can some of these not be linked via other inputs? Not particularly familiar with dotnet so I don't know much here

```console
$ tree ./results/avdump3
├── bin
│  └── avdump3
└── share
   └── avdump3
      ├── AVD3AniDBModule.dll
      ├── AVDump3CL
      ├── AVDump3CL.dll
      ├── AVDump3CL.exe
      ├── AVDump3CL.runtimeconfig.json
      ├── AVDump3Lib.dll
      ├── AVDump3NativeLib-aarch64.so
      ├── AVDump3NativeLib-x64.so
      ├── AVDump3NativeLib.dll
      ├── AVDump3UI.dll
      ├── BXmlLib.dll
      ├── ExtKnot.StringInvariants.dll
      ├── Ionic.Zlib.Core.dll
      ├── MediaInfo-aarch64.so
      ├── MediaInfo-x64.so
      ├── MediaInfo.dll
      ├── Microsoft.CodeAnalysis.CSharp.dll
      ├── Microsoft.CodeAnalysis.CSharp.Scripting.dll
      ├── Microsoft.CodeAnalysis.dll
      ├── Microsoft.CodeAnalysis.Scripting.dll
      ├── Microsoft.Extensions.DependencyInjection.Abstractions.dll
      ├── Microsoft.Extensions.DependencyInjection.dll
      ├── Microsoft.Extensions.Logging.Abstractions.dll
      ├── Microsoft.Extensions.Logging.dll
      ├── Microsoft.Extensions.Options.dll
      ├── Microsoft.Extensions.Primitives.dll
      ├── Newtonsoft.Json.dll
      ├── Serilog.dll
      └── Serilog.Extensions.Logging.dll
```

kini:
All of that is just directly extracted from the zip file provided by upstream. This is a closed source application so I wouldn't know how to regenerate those exact binaries from other inputs. Or maybe I'm misunderstanding what you're suggesting?

Frontear:
I'm just curious if all of these are necessary for the app, because its a lot of files and its increasing closure size. If it can be trimmed down that'd be awesome

pkgs/by-name/av/avdump3/package.nix
Frontear: What you have now should be fine

kini:
Are you sure AUR doesn't copy everything? Maybe I'm misreading [the PKGBUILD](https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=avdump3-bin) but it looks like they also copy all the files.

I suppose it would likely be safe to at least delete the aarch64-specific files when building on x86_64... But I hesitate to wantonly delete files without understanding their purpose.

pkgs/by-name/av/avdump3/package.nix
```diff
@@ -0,0 +1,59 @@
...
+    hash = "sha256-H9Sn3I4S9CmymKIMHVagDy+7svHs285S3EJgYQo+ks0=";
+    stripRoot = false;
+  };
+
+  installPhase = ''
+    runHook preInstall
+    mkdir -p $out/share/avdump3 $out/bin
+    mv * $out/share/avdump3
+    cat > $out/bin/avdump3 <<EOF
+    #!/bin/sh
```
Frontear, line 25: nit: i'm a little against this style of shebang, even if it _would_ work.

there's stylistic reasons and there's a logical reason as well. stylistically, i think having a direct path shebang is contrary to nixpkgs' whole goal of declarativeness. logically, while this shebang will have relatively consistent behavior on NixOS, on other distros that use Nix as a package manager `/bin/sh` can point to the literal `sh` executable, an older version of `bash`, something minimal like `dash`, etc etc.

Can you instead change this whole bit into a `writeShellScript` style derivation?

Frontear (changes requested):

pkgs/by-name/av/avdump3/package.nix
kini: I'm not sure what exactly you're suggesting, but would it be acceptable to simply replace `#!/bin/sh` with `#!${runtimeShell}` in this inline script? That seems the simplest, and it's a pattern that occurs elsewhere in nixpkgs.

kini: I implemented my above suggestion.

Frontear:
Testing the program seems to make my terminal cursor pointer disappear (the blinking '|' line). I don't know if that's an upstream issue or not, just figured I'd drop that here as an fyi.

kini:
I think it's an upstream issue. It seems to only occur when the program prints the usage text. If you actually run it successfully on some proper input (i.e. actually scan some files) it doesn't screw up the terminal cursor.

Aleksanaa (approved):

Aleksanaa:
Backport if you want.

kini:
Thanks for the merge!

PR #324217, opened 2024-07-03
86Box: Add optional wayland support

cjab:
86Box will run on wayland without these changes but mouse capturing will not work. Adding `extra-cmake-modules` and `wayland` as dependencies solves this problem.

I'm uncertain if compiling with these dependencies will cause issues for non-wayland users so the they are optional and controlled by the `enableWayland` flag.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/applications/emulators/86box/default.nix
+++ b/pkgs/applications/emulators/86box/default.nix
@@ -3,6 +3,7 @@
   lib,
   fetchFromGitHub,
   cmake,
+  extra-cmake-modules,
   pkg-config,
   makeWrapper,
   freetype,
@@ -19,10 +20,13 @@
   discord-gamesdk,
   libpcap,
   libslirp,
+  wayland,
+  wayland-scanner,
 
   enableDynarec ? with stdenv.hostPlatform; isx86 || isAarch,
   enableNewDynarec ? enableDynarec && stdenv.hostPlatform.isAarch,
   enableVncRenderer ? false,
+  enableWayland ? stdenv.isLinux,
   unfreeEnableDiscord ? false,
   unfreeEnableRoms ? false,
 }:
@@ -43,7 +47,7 @@ stdenv.mkDerivation (finalAttrs: {
     pkg-config
     makeWrapper
     qt5.wrapQtAppsHook
-  ];
+  ] ++ lib.optionals enableWayland [ extra-cmake-modules wayland-scanner ];
 
   buildInputs = [
     freetype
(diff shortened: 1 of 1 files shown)


pkgs/applications/emulators/86box/default.nix
```diff
@@ -19,10 +20,12 @@
   discord-gamesdk,
   libpcap,
   libslirp,
+  wayland,
 
   enableDynarec ? with stdenv.hostPlatform; isx86 || isAarch,
   enableNewDynarec ? enableDynarec && stdenv.hostPlatform.isAarch,
   enableVncRenderer ? false,
+  enableWayland ? false,
```
Artturin, line 28:
```suggestion
  enableWayland ? stdenv.isLinux,
```

The aur package always enables wayland so it should be fine https://aur.archlinux.org/packages/86box?all_deps=1#pkgdeps

Also add `86Box:` to the commit message.

pkgs/applications/emulators/86box/default.nix
```diff
@@ -43,7 +46,7 @@ stdenv.mkDerivation (finalAttrs: {
     pkg-config
     makeWrapper
     qt5.wrapQtAppsHook
-  ];
+  ] ++ lib.optional enableWayland extra-cmake-modules;
```
Artturin, line 49:
```suggestion
  ] ++ lib.optionals enableWayland [ extra-cmake-modules wayland-scanner ];
```

https://github.com/86Box/86Box/blob/c728f6cbf59735196e0ef69be514583c915e1d72/src/qt/CMakeLists.txt#L414

The reason it's working is that `strictDeps` isn't enabled

Artturin (approved):

pkgs/applications/emulators/86box/default.nix
cjab: Thanks @Artturin ! I applied this and your other suggestion but I just wanted to understand a bit more about `strictDeps`. Should `strictDeps` be enabled here? It seems like maybe it should be whenever possible? If you have any information on it or even just a link to some documentation that would be greatly appreciated 😄 .

Aleksanaa (approved):
Please reword commit to `_86box: Add optional wayland support`

jchv (approved):

pkgs/applications/emulators/86box/default.nix
Artturin: No docs but it's good to enable it when possible

PR #324248, opened 2024-07-03
xcodeenv: accept version & perform runtime checks

siddarthkay:
## Description of changes

- xcodewrapper nix derivation has been updated to now accept a list of acceptable versions.
- allowHigher is now removed
- this matches closely to what we use for building react-native with nix at status-mobile repo
ref -> https://github.com/status-im/status-mobile/blob/develop/nix/pkgs/xcodeenv/compose-xcodewrapper.nix
- The key change done here is that now xcodewrapper checks Xcode versions at runtime instead of build time. This helps us to show warning messages when underlying environment does not have the Xcode version we want to support.

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [ ] aarch64-linux
  - [x] x86_64-darwin
  - [x] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/development/mobile/xcodeenv/compose-xcodewrapper.nix
+++ b/pkgs/development/mobile/xcodeenv/compose-xcodewrapper.nix
@@ -1,19 +1,41 @@
-{ stdenv, lib }:
-{ version ? "11.1"
-, allowHigher ? false
-, xcodeBaseDir ? "/Applications/Xcode.app" }:
+{ lib,
+  stdenv,
+  writeShellScriptBin }:
+{ versions ? [ ] , xcodeBaseDir ? "/Applications/Xcode.app" }:
 
 assert stdenv.isDarwin;
+let
+  xcodebuildPath = "${xcodeBaseDir}/Contents/Developer/usr/bin/xcodebuild";
 
+  xcodebuildWrapper = writeShellScriptBin "xcodebuild" ''
+    currentVer="$(${xcodebuildPath} -version | awk 'NR==1{print $2}')"
+    wrapperVers=(${lib.concatStringsSep " " versions})
+
+    for ver in "''${wrapperVers[@]}"; do
+      if [[ "$currentVer" == "$ver" ]]; then
+        # here exec replaces the shell without creating a new process
+        # https://www.gnu.org/software/bash/manual/bash.html#index-exec
+        exec "${xcodebuildPath}" "$@"
+      fi
+    done
+
+    echo "The installed Xcode version ($currentVer) does not match any of the allowed versions: ${lib.concatStringsSep ", " versions}"
+    echo "Please update your local Xcode installation to match one of the allowed versions"
+    exit 1
+  '';
+in
 stdenv.mkDerivation {
-  pname = "xcode-wrapper${lib.optionalString allowHigher "-plus"}";
-  inherit version;
+  name = "xcode-wrapper-impure";
   # Fails in sandbox. Use `--option sandbox relaxed` or `--option sandbox false`.
   __noChroot = true;
   buildCommand = ''
     mkdir -p $out/bin
     cd $out/bin
-    ln -s /usr/bin/xcode-select
+    ${if versions == [ ] then ''
+    ln -s "${xcodebuildPath}"
+    '' else ''
+    ln -s "${xcodebuildWrapper}/bin/xcode-select"
+    ''}
     ln -s /usr/bin/security
     ln -s /usr/bin/codesign
     ln -s /usr/bin/xcrun
@@ -22,23 +44,9 @@ stdenv.mkDerivation {
     ln -s /usr/bin/lipo
     ln -s /usr/bin/file
     ln -s /usr/bin/rev
-    ln -s "${xcodeBaseDir}/Contents/Developer/usr/bin/xcodebuild"
     ln -s "${xcodeBaseDir}/Contents/Developer/Applications/Simulator.app/Contents/MacOS/Simulator"
 
     cd ..
     ln -s "${xcodeBaseDir}/Contents/Developer/Platforms/iPhoneSimulator.platform/Developer/SDKs"
-
-    # Check if we have the xcodebuild version that we want
-    currVer=$($out/bin/xcodebuild -version | head -n1)
-    ${if allowHigher then ''
-    if [ -z "$(printf '%s\n' "${version}" "$currVer" | sort -V | head -n1)""" != "${version}" ]
-    '' else ''
-    if [ -z "$(echo $currVer | grep -x 'Xcode ${version}')" ]
-    ''}
-    then
-        echo "We require xcodebuild version${if allowHigher then " or higher" else ""}: ${version}"
-        echo "Instead what was found: $currVer"
-        exit 1
-    fi
   '';
 }


jakubgs (approved):

jakubgs (changes requested):
Commit title should be prefixed with the package being changed.

siddarthkay:
@jakubgs : Done, thanks for taking a look

jakubgs (approved):

wegank:
@NixOS/darwin-maintainers

kashw2:
Result of `nixpkgs-review pr 324248` run on aarch64-darwin [1](https://github.com/Mic92/nixpkgs-review)

pkgs/development/mobile/xcodeenv/compose-xcodewrapper.nix
```diff
@@ -1,19 +1,39 @@
...
+    for ver in "''${wrapperVers[@]}"; do
+      if [[ "$currentVer" == "$ver" ]]; then
+        # here exec replaces the shell without creating a new process
+        # https://www.gnu.org/software/bash/manual/bash.html#index-exec
+        exec "${xcodebuildPath}" "$@"
+      fi
+    done
+
+    echo "The installed Xcode version ($currentVer) does not match any of the allowed versions: ${lib.concatStringsSep ", " versions}"
+    echo "Please update your local Xcode installation to match one of the allowed versions"
```
DontEatOreo, line 24: Could we use `lib.warn`?

DontEatOreo:
Result of `nixpkgs-review pr 324248` run on aarch64-darwin [1](https://github.com/Mic92/nixpkgs-review)

DontEatOreo:
Result of `nixpkgs-review pr 324248` run on x86_64-darwin [1](https://github.com/Mic92/nixpkgs-review)

pkgs/development/mobile/xcodeenv/compose-xcodewrapper.nix
reckenrode: I don’t think it can be. This script is invoking Xcode impurely, so the version isn’t known at evaluation time.

pkgs/development/mobile/xcodeenv/compose-xcodewrapper.nix
```diff
@@ -1,19 +1,39 @@
-{ stdenv, lib }:
-{ version ? "11.1"
-, allowHigher ? false
-, xcodeBaseDir ? "/Applications/Xcode.app" }:
+{ lib,
+  stdenv,
+  writeShellScriptBin }:
+{ versions ? [ "14.3" "15.1" "15.2" "15.3" ], xcodeBaseDir ? "/Applications/Xcode.app" }:
```
toonn, line 4: How did you select these specific versions?

pkgs/development/mobile/xcodeenv/compose-xcodewrapper.nix
```diff
@@ -1,19 +1,39 @@
...
-  inherit version;
+  pname = "xcode-wrapper-plus";
+  version = lib.concatStringsSep "," versions;
   # Fails in sandbox. Use `--option sandbox relaxed` or `--option sandbox false`.
   __noChroot = true;
   buildCommand = ''
     mkdir -p $out/bin
     cd $out/bin
-    ln -s /usr/bin/xcode-select
+    ln -s "${xcodebuildWrapper}/bin/xcode-select"
```
toonn, line 36: Doesn't this point to `xcodebuildWrapper.outPath/bin/xcode-select`? Wouldn't we want to get this from `xcodeBaseDir` if not `/usr/bin`?

toonn (commented):
I'm left with one fundamental question. What do we gain from limiting the accepted versions? It's an impure wrapper anyway so flexibility seems more important. I could see a use for allowing to constrain the version, as a user, but I don't see why Nixpkgs should constrain this.

pkgs/development/mobile/xcodeenv/compose-xcodewrapper.nix
siddarthkay: These are the last few working versions and they are considered only when versions parameter is not provided.

siddarthkay:
> What do we gain from limiting the accepted versions?

in any project limiting Xcode version prevents developers from installing bleeding edge versions of `Xcode` and having their app break. Earlier `xcode-wrapper` achieved this by providing one version and an allowHigher flag. However a versions list is more explicit and allows us more control over what are acceptable `Xcode` versions for any given project.

you can see how we use this wrapper here in status-mobile -> https://github.com/status-im/status-mobile/blob/develop/nix/overlay.nix#L60

pkgs/development/mobile/xcodeenv/compose-xcodewrapper.nix
toonn: It's not as if older versions of Xcode stop working when new versions are released, so at least older versions should be accepted.

toonn:
> in any project limiting Xcode version prevents developers from installing bleeding edge versions of `Xcode` and having their app break.

I'm fine with echoing a warning, I'm fine with providing a way for people to constrain the accepted versions and exiting if the right version isn't available.
I don't see the use of limiting allowable versions by default, that just seems like it introduces bookkeeping that has no business being part of Nixpkgs.

pkgs/development/mobile/xcodeenv/compose-xcodewrapper.nix
siddarthkay: they would be, you can pass whatever you wish in version argument, this list is just a fallback for when you don't pass an argument

pkgs/development/mobile/xcodeenv/compose-xcodewrapper.nix
```diff
@@ -22,23 +42,10 @@ stdenv.mkDerivation {
...
-
-    # Check if we have the xcodebuild version that we want
-    currVer=$($out/bin/xcodebuild -version | head -n1)
-    ${if allowHigher then ''
-    if [ -z "$(printf '%s\n' "${version}" "$currVer" | sort -V | head -n1)""" != "${version}" ]
-    '' else ''
-    if [ -z "$(echo $currVer | grep -x 'Xcode ${version}')" ]
-    ''}
-    then
-        echo "We require xcodebuild version${if allowHigher then " or higher" else ""}: ${version}"
```
siddarthkay, line 39: @toonn : look at what already exists in nixpkgs. this wrapper script served the business of moderating an xcode version. my changes just improve this a bit further.

siddarthkay:
> it introduces bookkeeping

yes i agree, but its important, pinning versions is the reason why you get predictable builds each time.
Don't get me started on nasty issues everyone has to face when Xcode releases a new version.

This way we are protected against that.

pkgs/development/mobile/xcodeenv/compose-xcodewrapper.nix
toonn: The original was version 11, that was either just the version of Xcode on the system at that time or it was an attempt to keep it compatible with the 11.0 SDK that aarch64-darwin stdenv is based on.

Constraining to other versions by default just doesn't seem helpful.

siddarthkay: I don't mind adding Xcode 11 to the list, as this is just a fallback when no version list is provided.

toonn:
How does a derivation not working when you update Xcode, which Nix has no control over, protect "us?" I could understand the argument if this was used internally in Nixpkgs but it can't because it's impure.

And you can guarantee no one will run into nasty compatibility issues between 14.3 and 15.3?

emilazy:
Is Status the only user of this code? If it’s specialized to their specific needs around version compatibility – which is fair enough – maybe it should just be split out into a repository of their own rather than being in Nixpkgs?

jakubgs:
Yes, we didn't introduce this package, it has been there for a while, no idea who else uses it.

@toonn 

>How does a derivation not working when you update Xcode, which Nix has no control over, protect "us?"

By verifying version at runtime rather than at build time. Since xcode version on the system can change AFTER the derivation has been instantiated the current way it works will ignore the xcode version change and continue working as if nothing happened, even though the version check would fail if the build was ran once again.

> I could understand the argument if this was used internally in Nixpkgs but it can't because it's impure.

No idea where else it's used.

>And you can guarantee no one will run into nasty compatibility issues between 14.3 and 15.3?

I cannot guarantee anything about behavior of xcode itself nor Apple. We have simply observed that there are cases where xcode version on the system changes and the `xcodewrapper` derivation does not detect this, since it's already built, and the xcode version is not an input, only allowed versions, so the cached build of `xcodewrapper` in Nix store is used, making the version check effective only at build time, and never again at runtime.

jakubgs:
@emilazy I can agree with that. We could use an empty list to indicate "no-limits" I guess, and have it as the default, so all versions are accepted, but also allow other to specify versions if need-be.

toonn:
Indeed, that's all I want. Checking at run-time is a great improvement so it's unfortunate this PR has been held back by the default version baggage. I take the blame for not effectively communicating this.

siddarthkay:
Alright I'll get rid of the version list.

pkgs/development/mobile/xcodeenv/compose-xcodewrapper.nix
jakubgs: `xcodeBaseDir` is used in the wrapper.

pkgs/development/mobile/xcodeenv/compose-xcodewrapper.nix
```diff
@@ -1,19 +1,43 @@
...
-  inherit version;
+  pname = "xcode-wrapper-plus";
+  version = if builtins.length versions > 0 then lib.concatStringsSep "," versions else "unwrapped";
   # Fails in sandbox. Use `--option sandbox relaxed` or `--option sandbox false`.
   __noChroot = true;
   buildCommand = ''
     mkdir -p $out/bin
     cd $out/bin
-    ln -s /usr/bin/xcode-select
+    ln -s "${xcodebuildWrapper}/bin/xcode-select"
```
jakubgs, line 40: Wouldn't this be simpler?
```suggestion
    ${if versions == [] then ''
    ln -s "${xcodeBaseDir}/Contents/Developer/usr/bin/xcodebuild"
    '' else ''
    ln -s "${xcodebuildWrapper}/bin/xcode-select"
    ''}
```

pkgs/development/mobile/xcodeenv/compose-xcodewrapper.nix
```diff
@@ -1,19 +1,43 @@
...
+  ''
+  else
+    writeShellScriptBin "xcodebuild" ''
+      exec "${xcodebuildPath}" "$@"
+  '';
+in
 stdenv.mkDerivation {
-  pname = "xcode-wrapper${lib.optionalString allowHigher "-plus"}";
-  inherit version;
+  pname = "xcode-wrapper-plus";
```
toonn, line 33: Since we don't have a version to include how about exchanging "plus" for "impure?"

pkgs/development/mobile/xcodeenv/compose-xcodewrapper.nix
```diff
@@ -1,19 +1,43 @@
...
+        exec "${xcodebuildPath}" "$@"
+      fi
+    done
+
+    echo "The installed Xcode version ($currentVer) does not match any of the allowed versions: ${lib.concatStringsSep ", " versions}"
+    echo "Please update your local Xcode installation to match one of the allowed versions"
+    exit 1
+  ''
+  else
+    writeShellScriptBin "xcodebuild" ''
```
toonn, line 28: Should we consider a symlink?

toonn (commented):
I'm OK with the current state, just some suggestions for improvement.

pkgs/development/mobile/xcodeenv/compose-xcodewrapper.nix
siddarthkay: indeed simpler, I've incorporated this change in my latest commit.

pkgs/development/mobile/xcodeenv/compose-xcodewrapper.nix
siddarthkay: yeah sure I think impure also sounds good

pkgs/development/mobile/xcodeenv/compose-xcodewrapper.nix
siddarthkay: Could you expand a bit on what you mean here?

pkgs/development/mobile/xcodeenv/compose-xcodewrapper.nix
jakubgs: Good suggestion.

jakubgs (approved):
Looks good.

pkgs/development/mobile/xcodeenv/compose-xcodewrapper.nix
toonn: The suggestion by jakubgs seems to have taken care of this already.

toonn (approved):
LGTM though I haven't tested this.

The only addition I'd consider is echoing some form of warning when `versions` is an empty list so that users learn about the need to specify their own version constraints. Maybe with a helpful message about why you might want to do so.

emilazy:
We could also just make `versions` required, if all downstream users want to restrict the versions, and save the code complexity of the branches?

siddarthkay:
I tested this PR on a derivation and found an issue with the symlink, I'm converting this PR to draft for the time being so that this does not get merged by anyone else.
I'll report here once that issue is resolved.

toonn:
@emilazy, no people might not ever hit a problem with their versions so an empty list is still a desireable value.

siddarthkay:
> issue with the `xcodebuild` symlink,

update: this issue is fixed, Thanks @jakubgs

pkgs/development/mobile/xcodeenv/compose-xcodewrapper.nix
```diff
@@ -1,18 +1,42 @@
...
-  inherit version;
+  pname = "xcode-wrapper-impure";
+  version = if versions != [ ] then lib.concatStringsSep "," versions else "";
   # Fails in sandbox. Use `--option sandbox relaxed` or `--option sandbox false`.
   __noChroot = true;
   buildCommand = ''
     mkdir -p $out/bin
     cd $out/bin
+    ${if versions == [] then ''
+    ln -s "${xcodeBaseDir}/Contents/Developer/usr/bin/xcodebuild"
```
jakubgs, line 36: You could just use `xcodebuildPath` here if you already have it set.

jakubgs (approved):

jakubgs (approved):

pkgs/development/mobile/xcodeenv/compose-xcodewrapper.nix
```diff
@@ -1,18 +1,42 @@
...
+    echo "The installed Xcode version ($currentVer) does not match any of the allowed versions: ${lib.concatStringsSep ", " versions}"
+    echo "Please update your local Xcode installation to match one of the allowed versions"
+    exit 1
+  '';
+in
 stdenv.mkDerivation {
-  pname = "xcode-wrapper${lib.optionalString allowHigher "-plus"}";
-  inherit version;
+  pname = "xcode-wrapper-impure";
+  version = if versions != [ ] then lib.concatStringsSep "," versions else "";
```
toonn, line 29: No need to guard this, the result for an empty list is `""`.

pkgs/development/mobile/xcodeenv/compose-xcodewrapper.nix
```diff
@@ -1,18 +1,42 @@
...
   # Fails in sandbox. Use `--option sandbox relaxed` or `--option sandbox false`.
   __noChroot = true;
   buildCommand = ''
     mkdir -p $out/bin
     cd $out/bin
+    ${if versions == [ ] then ''
+    ln -s "${xcodeBaseDir}/Contents/Developer/usr/bin/xcodebuild"
```
toonn, line 36: This is the same content as `xcodebuildPath` so I don't understand the branch here?

toonn (commented):
Was about to hit merge but two small questions.

pkgs/development/mobile/xcodeenv/compose-xcodewrapper.nix
jakubgs: I think he made a mistake and forgot to use `xcodebuildWrapper`.

pkgs/development/mobile/xcodeenv/compose-xcodewrapper.nix
siddarthkay: removed this entirely and instead of using `pname` with `version` I am now using `name`

pkgs/development/mobile/xcodeenv/compose-xcodewrapper.nix
```diff
@@ -1,18 +1,41 @@
...
+  name = "xcode-wrapper-impure";
   # Fails in sandbox. Use `--option sandbox relaxed` or `--option sandbox false`.
   __noChroot = true;
   buildCommand = ''
     mkdir -p $out/bin
     cd $out/bin
+    ${if versions == [ ] then ''
+    ln -s "${xcodebuildPath}"
+    '' else ''
+    ln -s "${xcodebuildWrapper}"
```
jakubgs, line 37: Shouldn't this be:
```suggestion
    ln -s "${xcodebuildWrapper}/bin/xcode-select"
```

siddarthkay: you're right!
This was an oversight on my part.
Thanks for taking a look.

siddarthkay:
@toonn : fwiw I tested this code in https://github.com/status-im/status-mobile and they indeed work for both cases.
However also happy to wait 1-2 days for others to test etc

toonn:
Good enough for me.

PR #324347, opened 2024-07-03
nixos/gdm: Fix fingerprint auth rules

jtojnar:
## Description of changes

We introduced the gdm-fingerprint.pam in 9d41fe6fcc4df838a56b1cfb2512b65e5e655958.

We used the [upstream Arch config] as a template, which contains an extended control field that jumps over **one** immediately-following `auth` rule unless `pam_gdm.so` succeeds.

But we decided to not include `pam_gnome_keyring.so` so there was no rule to skip over, resulting in a broken control flow and the PAM module failing with “PAM bad jump in stack”, breaking the fingerprint authentication in GDM.

Let’s actually add `pam_gnome_keyring.so`, like the Arch config does. Because we are creating the PAM file using the `text` option, `security.pam.services.gdm-fingerprint.enableGnomeKeyring` does not do anything so we need to do it manually.

For the case where gnome-keyring is not enabled, we could add a no-op rule like `optional pam_permit.so` after `pam_gdm.so` so that the branching always has something to jump over but it will be simpler to just make the both conditional. There are no further `auth` rules that could benefit from `pam_gdm.so` doing something so it should be fine.

Unlike in Arch, we are not going to invoke `pam_gnome_keyring.so` in a `session` rule since that is already done by the included `login` module.

Let’s also add a comment about disabling `fprintAuth`, remove no-op `optional pam_permit.so` rule and revert parts of https://github.com/NixOS/nixpkgs/pull/315874 that do not do anything.

[upstream Arch config]: https://gitlab.gnome.org/GNOME/gdm/-/blob/81ee658c11381912131dd4a29e84190f7f9cd039/data/pam-arch/gdm-fingerprint.pam

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/nixos/modules/services/x11/display-managers/gdm.nix
+++ b/nixos/modules/services/x11/display-managers/gdm.nix
@@ -6,6 +6,7 @@ let
 
   cfg = config.services.xserver.displayManager;
   gdm = pkgs.gnome.gdm;
+  pamCfg = config.security.pam.services;
   settingsFormat = pkgs.formats.ini { };
   configFile = settingsFormat.generate "custom.conf" cfg.gdm.settings;
 
@@ -331,7 +332,10 @@ in
         auth       requisite                   pam_faillock.so      preauth
         auth       required                    ${pkgs.fprintd}/lib/security/pam_fprintd.so
         auth       required                    pam_env.so
-        auth       [success=ok default=1]      ${pkgs.gnome.gdm}/lib/security/pam_gdm.so
+        ${lib.optionalString pamCfg.login.enableGnomeKeyring ''
+          auth       [success=ok default=1]      ${pkgs.gnome.gdm}/lib/security/pam_gdm.so
+          auth       optional                    ${pkgs.gnome-keyring}/lib/security/pam_gnome_keyring.so
+        ''}
 
         account    include                     login
 
--- a/nixos/modules/services/x11/display-managers/gdm.nix
+++ b/nixos/modules/services/x11/display-managers/gdm.nix
@@ -321,13 +321,15 @@ in
         session   include       login
       '';
 
+      # This would block password prompt when included by gdm-password.
+      # GDM will instead run gdm-fingerprint in parallel.
       login.fprintAuth = mkIf config.services.fprintd.enable false;
+
       gdm-fingerprint.text = mkIf config.services.fprintd.enable ''
         auth       required                    pam_shells.so
         auth       requisite                   pam_nologin.so
         auth       requisite                   pam_faillock.so      preauth
         auth       required                    ${pkgs.fprintd}/lib/security/pam_fprintd.so
-        auth       optional                    pam_permit.so
         auth       required                    pam_env.so
         auth       [success=ok default=1]      ${pkgs.gnome.gdm}/lib/security/pam_gdm.so
 
--- a/nixos/modules/services/desktops/gnome/gnome-keyring.nix
+++ b/nixos/modules/services/desktops/gnome/gnome-keyring.nix
@@ -35,15 +35,7 @@ in
 
     xdg.portal.extraPortals = [ pkgs.gnome-keyring ];
 
-    security.pam.services = lib.mkMerge [
-      {
-        login.enableGnomeKeyring = true;
-      }
-      (lib.mkIf config.services.xserver.displayManager.gdm.enable {
-        gdm-password.enableGnomeKeyring = true;
-        gdm-autologin.enableGnomeKeyring = true;
-      })
-    ];
+    security.pam.services.login.enableGnomeKeyring = true;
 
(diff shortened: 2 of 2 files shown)


nixos/modules/services/x11/display-managers/gdm.nix
```diff
@@ -321,21 +322,31 @@ in
...
         auth       requisite                   pam_faillock.so      preauth
         auth       required                    ${pkgs.fprintd}/lib/security/pam_fprintd.so
-        auth       optional                    pam_permit.so
         auth       required                    pam_env.so
         auth       [success=ok default=1]      ${pkgs.gnome.gdm}/lib/security/pam_gdm.so
+        ${lib.optionalString
+          pamCfg.gdm-fingerprint.enableGnomeKeyring
+          "auth       optional                    ${pkgs.gnome-keyring}/lib/security/pam_gnome_keyring.so"}
+        # So that pam_gdm has something to skip over in the default branch when gnome-keyring is disabled.
+        auth       optional                    pam_permit.so
```
Pandapip1, line 340:
```suggestion
        ${if pamCfg.gdm-fingerprint.enableGnomeKeyring then
          "auth       optional                    ${pkgs.gnome-keyring}/lib/security/pam_gnome_keyring.so" else
          # So that pam_gdm has something to skip over in the default branch when gnome-keyring is disabled.
          "auth       optional                    pam_permit.so"}
```
So that the `pam_permit.so` line is only included when it needs to be.

Pandapip1 (approved):
LGTM; one nitpick.

JohnRTitor:
We introduced `gdm-password.enableGnomeKeyring` in #315874 as well, can we make its PAM config more in line with [upstream](https://gitlab.gnome.org/GNOME/gdm/-/blob/main/data/pam-arch/gdm-password.pam)? 
https://github.com/NixOS/nixpkgs/blob/6cb20d80466067bc39c00a02d28f866f0db05965/nixos/modules/services/x11/display-managers/gdm.nix#L304-L309

nixos/modules/services/x11/display-managers/gdm.nix
jtojnar: Hmm, I am split on whether that is easier to read. But it gave me an idea – I wonder if it would make sense to remove the `pam_permit.so` altogether and use something like the following control for `pam_gdm.so`:

```nix
${if pamCfg.gdm-fingerprint.enableGnomeKeyring then "[success=ok default=1]" else "optional"}
```

jtojnar:
Good point, `gdm-password.enableGnomeKeyring = true;` and `gdm-autologin.enableGnomeKeyring = true;` won’t do anything either.

Though, I do not think it makes sense to copy the Arch PAM config as-is everywhere – we already have `pam_gnome_keyring.so` in `login` PAM module.

Actually, maybe, thanks to that, we might not need `pam_gnome_keyring.so` here at all.

I believe Arch skips the `pam_gnome_keyring.so` rule when `pam_gdm.so` fails. But since we were already calling `pam_gnome_keyring.so` in `pam.d/login` without issue even without running `pam_gdm.so`, it is probably fine to get rid of the conditional. After all, the only thing `pam_gdm.so` does is make sure passwords from early boot are available to later rules (same as `pam_systemd_loadkeys.so`, which should replace it). And I would expect the `optional pam_gnome_keyring.so` rule to just fail when a password is not available.

jtojnar:
Actually, never mind – the `gdm-fingerprint.pam` does not contain `auth include login` so we still need to add `pam_gnome_keyring.so` ourselves.

I have mostly reverted https://github.com/NixOS/nixpkgs/pull/315874 since it does not do anything.

We should probably replace `gdm-autologin` with https://gitlab.gnome.org/GNOME/gdm/-/blob/81ee658c11381912131dd4a29e84190f7f9cd039/data/pam-arch/gdm-autologin.pam but I won’t do it here so that we do not block fingerprint support further.

nixos/modules/services/x11/display-managers/gdm.nix
Pandapip1: I agree, that's preferable.

jtojnar: Decided to make both lines conditional instead, since that will be even cleaner. There are no further `auth` rules that could benefit from `pam_gdm.so` doing something.

JohnRTitor (approved):
LGTM

schuelermine:
seems to work fine on my machine when cherrypicked onto nixos-unstable

PR #247766, opened 2023-08-07
GNOME: 44 → 45

bobby285271:
This updates GNOME desktop environment and its underlying platform to [45 series](https://release.gnome.org/45), some of the highlights (pain points) are:

- gnome-shell{,-extensions} now using ESModules
- gsettings-desktop-schemas 45 dropped window shading support
- Bumped API version: libmutter-13, sysprof-6, libpeas-2
- New core apps: loupe 45 and snapshot 45
- GJS using spidermonkey 115
- ~~Cairo was bumped from a 4-year-old version (1.16.0) to 1.18.0~~ Done separately in https://github.com/NixOS/nixpkgs/pull/257301
- xdg-desktop-portal 1.17 reworked how portal implementations are loaded

Current running evals:

360bbbce42a8215b35f297f42d3c878cdd79b724: https://hydra.nixos.org/eval/1801912?filter=x86_64-linux&compare=1801877

0edd45851f4dad51e068de409028d795e7a0b93a: https://hydra.nixos.org/eval/1802015?compare=1802011

## Test/contribute to this PR

> This will be merged soon, if you wish to review please do.

To switch your system to this branch:

- Pass `-I nixpkgs=https://github.com/NixOS/nixpkgs/archive/refs/heads/gnome.tar.gz` flag to `nixos-rebuild`
- Use `nixos-rebuild build-vm` to create a VM if you don't wish to test on physical machine
- With flakes, `nix flake update --override-input nixpkgs github:NixOS/nixpkgs/gnome`

PR #160343, opened 2022-02-16
GNOME: 41 → 42

jtojnar:
https://discourse.gnome.org/t/gnome-42-released/9255
http://release.gnome.org/42
[Introducing GNOME 42 - YouTube](https://youtu.be/du-2QpWbiLU)

https://github.com/NixOS/nixpkgs/projects/44

This time starting a bit early.

-----

Main pain points:

- ~~libsoup3 migration https://gitlab.gnome.org/GNOME/libsoup/-/issues/218~~ does not look ready for this cycle, let’s stay on libsoup 2 when possible
- gtk4 migration https://gitlab.gnome.org/GNOME/Initiatives/-/issues/26

Processes can only link against one version of the library (even transitively), or there will be symbol clashes!

----

cc @NixOS/gnome 

https://hydra.nixos.org/jobset/nixpkgs/gnome

PR #324361, opened 2024-07-03
gnome.gnome-remote-desktop: re-enable tests

l0b0:
## Description of changes

Adds enough build inputs to ensure the built-in tests run.

This partially reverts commits efb08d9060b804c4d8525c42be2b960a53354a0a (#247766) / 1de0e97b3721f5361c738d13308123e02f563515 (#160343).

@bobby285271 @jtojnar This was done as a naive precursor to #266774, and I'm not sure whether there are any additional caveats I'm not aware of beyond the tests being flaky at the time they were disabled.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [x] Tested, as applicable:

---

### DIFF
--- a/pkgs/by-name/gn/gnome-remote-desktop/package.nix
+++ b/pkgs/by-name/gn/gnome-remote-desktop/package.nix
@@ -1,32 +1,32 @@
 {
-  stdenv,
-  lib,
-  fetchurl,
-  cairo,
-  meson,
-  ninja,
-  pkg-config,
-  python3,
   asciidoc,
-  wrapGAppsHook3,
+  cairo,
+  fdk_aac,
+  fetchurl,
+  freerdp3,
+  fuse3,
+  gdk-pixbuf,
   glib,
+  gnome,
+  lib,
+  libdrm,
   libei,
   libepoxy,
-  libdrm,
-  nv-codec-headers-11,
-  pipewire,
-  systemd,
-  libsecret,
   libnotify,
   libopus,
+  libsecret,
   libxkbcommon,
-  gdk-pixbuf,
-  freerdp3,
-  fdk_aac,
-  tpm2-tss,
-  fuse3,
-  gnome,
+  meson,
+  ninja,
+  nv-codec-headers-11,
+  pipewire,
+  pkg-config,
   polkit,
+  python3,
+  stdenv,
+  systemd,
+  tpm2-tss,
+  wrapGAppsHook3,
 }:
 
 stdenv.mkDerivation rec {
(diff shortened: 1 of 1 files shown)


pkgs/desktops/gnome/core/gnome-remote-desktop/default.nix
```diff
@@ -20,12 +20,18 @@
 , libopus
 , libxkbcommon
 , gdk-pixbuf
+, cmake
```
jtojnar: Are you sure `cmake` is needed? Meson supports it for finding dependencies but typically, `pkg-config` will be sufficient. See <https://github.com/jtojnar/nixpkgs-hammering/blob/master/explanations/meson-cmake.md> for more info.

pkgs/desktops/gnome/core/gnome-remote-desktop/default.nix
```diff
@@ -73,7 +85,6 @@ stdenv.mkDerivation rec {
     "-Dsystemd_system_unit_dir=${placeholder "out"}/lib/systemd/system"
     "-Dsystemd_sysusers_dir=${placeholder "out"}/lib/sysusers.d"
     "-Dsystemd_tmpfiles_dir=${placeholder "out"}/lib/tmpfiles.d"
-    "-Dtests=false" # Too deep of a rabbit hole.
```
jtojnar, line 76: This just re-enables building them. To actually run them, you need `doCheck = true`.

jtojnar:
> This partially reverts commits \[…\] 1de0e97b3721f5361c738d13308123e02f563515

Would not the second commit be re-applied rather than reverted?

pkgs/desktops/gnome/core/gnome-remote-desktop/default.nix
l0b0: It must've been included (or superseded) by another dependency, because the build used to fail because it was missing when I removed `"-Dtests=false"`. Removing…

l0b0:
> > This partially reverts commits […] [1de0e97](https://github.com/NixOS/nixpkgs/commit/1de0e97b3721f5361c738d13308123e02f563515)
> 
> Would not the second commit be re-applied rather than reverted?

Oh, I meant in the sense of reverting to running the tests (which turned out to require `doCheck = true;`).

pkgs/desktops/gnome/core/gnome-remote-desktop/default.nix
l0b0: Oh, good thing you mentioned this; the build was showing a whole bunch of tests running. Are different tests enabled by removing `-Dtests=false` and setting `doCheck = true;`?

l0b0:
The current issue seems to be

> Failed to start EGL thread: Missing extension 'EGL_EXT_platform_base'

Does this mean the test needs to run in a VM?

pkgs/desktops/gnome/core/gnome-remote-desktop/default.nix
jtojnar: Not sure what you have seen – no tests will run unless `doCheck` is `true`.

jtojnar:
> > Failed to start EGL thread: Missing extension 'EGL_EXT_platform_base'
> 
> Does this mean the test needs to run in a VM?

🤷‍♀️ Down the rabbit hole we go.

Maybe it is possible to use mesa software rendering but this is not something I am very familiar with.

PR #324386, opened 2024-07-03
ghc-8.10.7: use libffi-3.3

### LINKED ISSUE #324384: GHCi: ghc: freeHaskellFunctionPtr: not for me, guv!
Given this derivation:

```
{ pkgs ? import (builtins.fetchTarball "https://nixos.org/channels/nixos-24.05/nixexprs.tar.xz") {}
}:

with pkgs;

mkShell rec {
  name = "my-shell";

  buildInputs = [ haskell.packages.ghc8107.ghc ];
}
```

ghci will emit `freeHaskellFunctionPtr: not for me, guv!` for every key at its prompt. This is a bug and occurs on nixpkgs-24.05 and 23.11. I did not test `unstable`.

Why does this happen? Because GHC has very particular assumptions about libffi, when those assumptions don't hold we get this error  message from the RTS a particular GHC is built with. See https://gitlab.haskell.org/ghc/ghc/-/issues/20051 

Why did this happen? The ghc-8.10.7 derivation does not specify the version of libffi it pulls in. That version has crept from libffi-3.3 (which is what ghc 8.10.7 in gitlab was released with) to 3.4.4. The version itself is not problematic, what is problematic was the removal of a flag which incited this bug. This happened in [69c24679d695d06e22a544cbed3bd10feff63c5a](https://github.com/NixOS/nixpkgs/commit/69c24679d695d06e22a544cbed3bd10feff63c5a) which explains why this bug begins to occur in 23.11 and not in 23.05.

What's the fix? The fix is to use `libffi_3_3` instead of `libffi` in the ghc-8.10 derivation instead of just `libffi`. I verified that this solves the bug locally.

Just making this issue as a log. PR incoming...

doyougnu:
- Fixes #324384

- This is also the libffi version that GHC-8.10.7 was released with. See https://gitlab.haskell.org/ghc/ghc/-/issues/20051

### DIFF
--- a/pkgs/development/compilers/ghc/8.10.7.nix
+++ b/pkgs/development/compilers/ghc/8.10.7.nix
@@ -9,7 +9,9 @@
 , libiconv ? null, ncurses
 
 , # GHC can be built with system libffi or a bundled one.
-  libffi ? null
+  # we explicitly use libffi-3.3 here because 3.4 removes a flag that causes
+  # problems for ghc-8.10.7's RTS. See #324384.
+  libffi_3_3 ? null
 
 , useLLVM ? !(stdenv.targetPlatform.isx86
               || stdenv.targetPlatform.isPower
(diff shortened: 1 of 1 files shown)


maralorn:
I think this change is incomplete. It looks like you only edited the function arguments but not the usage sites of the variables? But maybe I am misunderstanding.

doyougnu:
ah! indeed I rushed.

sternenseemann:
Cherry-picked as 93d765be7e7db383c08b41bce4591b00abbfdbb1.

Issue #324306, opened 2024-07-03
Closure of ldc-built programs is too large


CyberShadow:
### Describe the bug
```nix
{ lib, writeText, stdenv, ldc }:
stdenv.mkDerivation {
  name = "ldc-test";
  src = writeText "test.d" ''
    module test;
    import std.stdio;
    void main() {
      writeln("Hello, world!");
    }
  '';
  dontUnpack = true;
  buildInputs = [ ldc ];
  buildPhase = ''
    ldc2 -of=test $src
  '';
  installPhase = ''
    mkdir -p $out/bin
    mv test $out/bin
  '';
}
```

Building and getting the closure size gives us a number around 1.3 GiB:

```console
$ nix --extra-experimental-features nix-command path-info --json -S $(nix-build --no-out-link -E '(import <nixpkgs> {}).callPackage ./package.nix {}') | jq '.[0].closureSize' 
1395772728
```

Graph:

![out](https://github.com/NixOS/nixpkgs/assets/160894/3dc64c6c-b831-490a-baea-7246d1f4d3d1)

It looks like there are some things there which might not need to be there, such as:
- the bootstrap LDC
- LLVM
- considering that the test program was compiled with neither debug information nor is dynamically linked to Phobos, LDC itself

### Notify maintainers

CC @ThomasMader @lionello @jtbx 

### Metadata
 - system: `"x86_64-linux"`
 - host os: `Linux 6.6.34, NixOS, 24.11 (Vicuna), 24.11pre641786.d603719ec6e2`
 - multi-user?: `yes`
 - sandbox: `yes`
 - version: `nix-env (Nix) 2.18.2`
 - nixpkgs: `/nix/store/ksj02zhy7qc4dmqi8g2ga8p9hj5173dv-nixpkgs-patched`

---

Add a :+1: [reaction] to [issues you find important].

[reaction]: https://github.blog/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/
[issues you find important]: https://github.com/NixOS/nixpkgs/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions-%2B1-desc

lionello:
Also see #111525

TomaSajt:
@jtbx if you have the time could you do something about the fact that `ldc` has references to `ldc-bootstrap`?

jtbx:
Yes! #393645

TomaSajt:
[...]

PR #288841, opened 2024-02-14
inochi-creator: init at 0.8.4, inochi-session: init at 0.8.3

TomaSajt:
## Description of changes

Closes https://github.com/NixOS/nixpkgs/issues/288706

This PR adds 2 packages: `inochi-session` `inochi-creator`

I have tested them with a basic setup, they seem to work fine.

---
## Notes

This app uses the D language. It uses the `ldc` compiler with the `dub` package manager and build system.

This PR originally contained the code which adds a generic builder for `dub` based packages, but that part got split off into another PR. This PR uses that builder.

As per the request of the author, I only use the "barebones" configuration, which disables the "InBranding" flag, which makes sure that  `inochi-creator` and `inochi-session` (and `inui`)  don't include the mascot image and the icon in the build.

I had to patch `i2d-imgui`, as that dependency has a git submodule that would be fetched when building.

Interestingly, one of the packages statically links `cimgui` and the other one dynamically links it.
The statically linked one needed some patching (to use capital letters for SDL2).

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
- For non-Linux: Is sandboxing enabled in `nix.conf`? (See [Nix manual](https://nixos.org/manual/nix/stable/command-ref/conf-file.html))
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:
  - [NixOS test(s)](https://nixos.org/manual/nixos/unstable/index.html#sec-nixos-tests) (look inside [nixos/tests](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests))
  - and/or [package tests](https://nixos.org/manual/nixpkgs/unstable/#sec-package-tests)
  - or, for functions and "core" functionality, tests in [lib/tests](https://github.com/NixOS/nixpkgs/blob/master/lib/tests) or [pkgs/test](https://github.com/NixOS/nixpkgs/blob/master/pkgs/test)
  - made sure NixOS tests are [linked](https://nixos.org/manual/nixpkgs/unstable/#ssec-nixos-tests-linking) to the relevant packages
- [ ] Tested compilation of all packages that depend on this change using `nix-shell -p nixpkgs-review --run "nixpkgs-review rev HEAD"`. Note: all changes have to be committed, also see [nixpkgs-review usage](https://github.com/Mic92/nixpkgs-review#usage)
[...]

PR #111525, opened 2021-02-01
dmd: remove references to dmd-bootstrap

lionello:
###### Motivation for this change
Fixes #111475 dmd keeps references to boostrap compiler

###### Things done
Add `preFixup` phase and use `remove-references-to` to remove all references to `dmd-bootstrap`.

- [ ] Tested using sandboxing ([nix.useSandbox](https://nixos.org/nixos/manual/options.html#opt-nix.useSandbox) on NixOS, or option `sandbox` in [`nix.conf`](https://nixos.org/nix/manual/#sec-conf-file) on non-NixOS linux)
- Built on platform(s)
   - [ ] NixOS
   - [x] macOS
   - [ ] other Linux distributions
- [ ] Tested via one or more NixOS test(s) if existing and applicable for the change (look inside [nixos/tests](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests))
- [ ] Tested compilation of all pkgs that depend on this change using `nix-shell -p nixpkgs-review --run "nixpkgs-review wip"`
- [ ] Tested execution of all binary files (usually in `./result/bin/`)
- [ ] Determined the impact on package closure size (by running `nix path-info -S` before and after)
- [ ] Ensured that relevant documentation is up to date
- [ ] Fits [CONTRIBUTING.md](https://github.com/NixOS/nixpkgs/blob/master/.github/CONTRIBUTING.md).

PR #324400, opened 2024-07-03
ldc: split includes to separate output

CyberShadow:
## Description of changes

- Fixes trivial cases of #324306.

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/by-name/ld/ldc/package.nix
+++ b/pkgs/by-name/ld/ldc/package.nix
@@ -15,6 +15,7 @@
 , callPackage
 , makeWrapper
 , runCommand
+, writeText
 , targetPackages
 
 , ldcBootstrap ? callPackage ./bootstrap.nix { }
@@ -145,5 +146,36 @@ stdenv.mkDerivation (finalAttrs: {
     mainProgram = "ldc2";
     maintainers = with maintainers; [ lionello jtbx ];
     platforms = [ "x86_64-linux" "i686-linux" "aarch64-linux" "x86_64-darwin" "aarch64-darwin" ];
+  };
+
+  passthru.tests = let
+    ldc = finalAttrs.finalPackage;
+    helloWorld = stdenv.mkDerivation (finalAttrs: {
+      name = "ldc-hello-world";
+      src = writeText "hello_world.d" ''
+        module hello_world;
+        import std.stdio;
+        void main() {
+          writeln("Hello, world!");
+        }
+      '';
+      dontUnpack = true;
+      buildInputs = [ ldc ];
+      dFlags = [];
+      buildPhase = ''
+        ldc2 ${lib.escapeShellArgs finalAttrs.dFlags} -of=test $src
+      '';
+      installPhase = ''
+        mkdir -p $out/bin
+        mv test $out/bin
+      '';
+    });
+  in {
+    # Without -shared, built binaries should not contain
+    # references to the compiler binaries.
+    no-references-to-compiler = helloWorld.overrideAttrs {
+      disallowedReferences = [ ldc ];
+      dFlags = ["-g"];
+    };
   };
 })
--- a/pkgs/by-name/ld/ldc/package.nix
+++ b/pkgs/by-name/ld/ldc/package.nix
@@ -73,8 +73,12 @@ stdenv.mkDerivation (finalAttrs: {
 
   buildInputs = [ curl tzdata ];
 
+  outputs = [ "out" "include" ];
+  outputInclude = "include";
+
   cmakeFlags = [
     "-DD_FLAGS=-d-version=TZDatabaseDir;-d-version=LibcurlPath;-J${pathConfig}"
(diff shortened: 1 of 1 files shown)


CyberShadow:
```
19 packages built:
btdu cheesecutter dstep dtools dub gtkd inochi-creator inochi-session ldc ldc.include literate luneta onedrive onedrivegui onedrivegui.dist rund sambamba serve-d tilix
```

jtbx:
I'm a bit torn on this, what's the point if the default output can't compile anything useful? I think the difference in closure size is negligible.

CyberShadow:
> I'm a bit torn on this, what's the point if the default output can't compile anything useful?

What do you mean? The default output references the `include` output.

> I think the difference in closure size is negligible.

It's not:

```console
$ nix path-info -S "$(nix-build --no-out-link -E '(import <nixpkgs> {}).callPackage ./package.nix {}')"
/nix/store/c7d1kz9rlj38k9w8n8smmp6msq2x0994-ldc-test	 1257047336

$ nix path-info -S "$(nix-build --no-out-link -E '(import ../../../work/extern/nixpkgs {}).callPackage ./package.nix {}')"
/nix/store/rsr7y65w6kirsr88idbqn01yhd90dhwp-ldc-test	   78766032
```

CyberShadow:
To clarify what's going on:

- Programs built by LDC have references to LDC, even without `-g` or `-shared`.
- Without `-g` or `-shared`, these references are all to source files in Phobos. I think they're coming from `__FILE__` instances in exception constructor default arguments.
- LDC itself (the compiler binaries) is big, but it is also pulling in LLVM, which is even bigger.
- By moving the Phobos sources to a separate output (`include`), we can make it so that LDC-built programs only depend on those sources. Those sources in turn don't depend on anything, so we can avoid pulling in LLVM etc.
- Nothing changes from LDC's perspective, other than that the Phobos sources are at a new location. The output containing the compiler binaries now also references the output containing the includes.
- There is room for more improvements to do the same thing for `-shared` builds, if we can move the shared libraries to their own separate output as well. However, that one doesn't seem to be quite as simple as for `include`.

jtbx:
Ah that makes sense now, thanks for all that.

> To clarify what's going on:
> 
>  * Programs built by LDC have references to LDC, even without `-g` or `-shared`.
> 
>  * Without `-g` or `-shared`, these references are all to source files in Phobos. I think they're coming from `__FILE__` instances in exception constructor default arguments.
> 

Yep I think so too, this was discussed in #288841

jtbx (approved):

CyberShadow:
Added two tweaks:
1. Moved `.d` files into `include/d/` so as to not make a mess with `buildEnv` et al
2. Added a test

lionello:
See https://github.com/NixOS/nixpkgs/pull/111525 which had fixes for `dmd`.

CyberShadow:
Thanks, `removeReferencesTo` is a rather blunt tool which I'd rather leave as a last resort.

lionello:
@CyberShadow I'm not sure you can avoid `removeReferencesTo` because IIRC each exception has the filename in it, which means Nix will keep references to bootstrap and/or Phobos around.

CyberShadow:
Well, I haven't looked at the situation with `dmd` / `dmd-bootstrap` specifically. What I did here was to put the Phobos sources in a separate output, which means that programs still refer to those sources, but that is as far as the dependency chain goes, so no more references to LLVM / bootstrap LDC / etc from built programs. Maybe something like that is possible for dmd-bootstrap too, if we tweak its `dmd.conf`?

pkgs/by-name/ld/ldc/package.nix
```diff
@@ -142,4 +148,34 @@ stdenv.mkDerivation (finalAttrs: {
     maintainers = with maintainers; [ lionello jtbx ];
     platforms = [ "x86_64-linux" "i686-linux" "aarch64-linux" "x86_64-darwin" "aarch64-darwin" ];
   };
+
+  passthru.tests = let
```
pbsds, line 152:
```suggestion
  passthru.tests = let
    ldc = finalAttrs.finalPackage;
```

pkgs/by-name/ld/ldc/package.nix
```diff
@@ -1,6 +1,7 @@
 { lib
 , stdenv
 , fetchFromGitHub
+, ldc
```
pbsds, line 4:
```suggestion
```

pkgs/by-name/ld/ldc/package.nix
CyberShadow: Thanks!

PR #324405, opened 2024-07-03
git-instafix: add quodlibetor as maintainer

quodlibetor:
## Description of changes

I write git-instafix and I want to be added to the reviewers list when it gets updated in nixpkgs.

## Things done

No changes or testing except adding myself as a maintainer.

### DIFF
--- a/maintainers/maintainer-list.nix
+++ b/maintainers/maintainer-list.nix
@@ -16479,6 +16479,12 @@
     githubId = 39039420;
     name = "Quinn Dougherty";
   };
+  quodlibetor = {
+    email = "quodlibetor@gmail.com";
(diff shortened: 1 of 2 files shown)


superherointj:
Add yourself to this file:

https://github.com/NixOS/nixpkgs/blob/master/maintainers/maintainer-list.nix

It should be a different commit. And it should come first. This order:

> maintainers: add quodlibetor

> git-instafix: add quodlibetor as maintainer

The order is necessary because second commit depends on first. Or else, eval fails!

quodlibetor:
thanks, lmk if it's not quite right.

superherointj (approved):

PR #324419, opened 2024-07-03
linux: multiple outputs for kernel targets

fpletz:
## Description of changes

Preview of an alternative idea for #244706 and inspired by #287122. Needs #239721.

This adds outputs for all built kernel targets and provides symlinks to all built images in the default output for compatibility. This way we can get rid of `lib.platforms.*.linux-kernel.target` eventually and we don't have to build the kernel multiple times for every target.

Looking for feedback. Is this maybe a way to go?

## TODO

 - [ ] Change NixOS bootloader code to use the outputs
 - [ ] Don't depend on all images in the NixOS system closure
 - [ ] Remove target from `lib.platforms.*.linux-kernel`
 - [ ] Maybe split modules and dtbs in separate outputs
 - [ ] Code cleanups for better readability

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [x] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/os-specific/linux/kernel/generic.nix
+++ b/pkgs/os-specific/linux/kernel/generic.nix
@@ -1,3 +1,12 @@
+let
+  # Map of images representations to legacy Makefile targets.
+  defaultMapKernelToInstallTargets = {
+    "uImage" = "uinstall";
+    "zImage" = "zinstall";
+    "Image.gz" = "zinstall";
+    "vmlinuz.efi" = "zinstall";
+  };
+in
 { buildPackages
 , callPackage
 , perl
@@ -70,6 +79,12 @@ lib.makeOverridable ({ # The kernel source tarball.
 # easy overrides to stdenv.hostPlatform.linux-kernel members
 , autoModules ? stdenv.hostPlatform.linux-kernel.autoModules
 , preferBuiltin ? stdenv.hostPlatform.linux-kernel.preferBuiltin or false
+, kernelTargets ?
+    if stdenv.hostPlatform.isAarch64 then
+      ["Image" "vmlinuz.efi"]
+    else
+      [stdenv.hostPlatform.linux-kernel.target]
+, installTargets ? lib.unique (map (target: defaultMapKernelToInstallTargets.${target} or "install") kernelTargets)
 , kernelArch ? stdenv.hostPlatform.linuxArch
 , kernelTests ? []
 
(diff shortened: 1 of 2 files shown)


pkgs/os-specific/linux/kernel/manual-config.nix
```diff
@@ -128,10 +137,10 @@ let
       ++ optionals withRust [ rustc rust-bindgen ]
       ;
 
-    in (optionalAttrs isModular { outputs = [ "out" "dev" ]; }) // {
+    in (optionalAttrs isModular { outputs = [ "out" "dev" ] ++ (lib.attrValues targetsToOutput); }) // {
```
jmbaur, line 140: We would end up wanting to have outputs even if the kernel wasn't built with module support.

jmbaur:
Overall really like the concept! Would be very nice to be able to easily switch between different kernel image types without incurring the cost of a rebuild. I think using derivation outputs as the differentiator between different image types seems like the right way to go here. I think we will also need to modify https://github.com/nixos/nixpkgs/blob/d5e0de612a503ff9c1a3fe2d0853c0fbbe336fbd/nixos/modules/system/activation/top-level.nix#L104 to take account for this as well. This also probably makes that option (`system.boot.loader.kernelFile`) the best way to go for choosing the kernel image type for a given nixos config, so it probably makes sense to make that option no longer marked as internal.

pkgs/os-specific/linux/kernel/generic.nix
```diff
@@ -70,6 +79,12 @@ lib.makeOverridable ({ # The kernel source tarball.
 # easy overrides to stdenv.hostPlatform.linux-kernel members
 , autoModules ? stdenv.hostPlatform.linux-kernel.autoModules
 , preferBuiltin ? stdenv.hostPlatform.linux-kernel.preferBuiltin or false
+, kernelTargets ?
+    if stdenv.hostPlatform.isAarch64 then
+      ["Image" "vmlinuz.efi"]
+    else
+      [stdenv.hostPlatform.linux-kernel.target]
+, installTargets ? lib.unique (map (target: defaultMapKernelToInstallTargets.${target} or "install") kernelTargets)
```
jmbaur, line 87: I'm not sure how valuable it is to have a separate argument for `installTargets`, as I can't imagine a scenario where the `kernelTargets` passed wouldn't be directly correlated to `installTargets`. My understanding is the linux makefiles accept the kernel image targets as make targets already (e.g. `make bzImage` works directly, so something like `make Image vmlinuz.efi` should work as well)

PR #320502, opened 2024-06-17
spades: 3.15.5 -> 4.0.0

bzizou:
## Description of changes

Upgrade to Spades 4.0.0

- compilation bug with gcc 13 has been fixed upstream
- copytree.patch suggested and accepted upstream
 
## Things done

- Built on platform(s)
  - [X] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
- For non-Linux: Is sandboxing enabled in `nix.conf`? (See [Nix manual](https://nixos.org/manual/nix/stable/command-ref/conf-file.html))
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:
  - [NixOS test(s)](https://nixos.org/manual/nixos/unstable/index.html#sec-nixos-tests) (look inside [nixos/tests](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests))
  - and/or [package tests](https://github.com/NixOS/nixpkgs/blob/master/pkgs/README.md#package-tests)
  - or, for functions and "core" functionality, tests in [lib/tests](https://github.com/NixOS/nixpkgs/blob/master/lib/tests) or [pkgs/test](https://github.com/NixOS/nixpkgs/blob/master/pkgs/test)
  - made sure NixOS tests are [linked](https://nixos.org/manual/nixpkgs/unstable/#ssec-nixos-tests-linking) to the relevant packages
- [ ] Tested compilation of all packages that depend on this change using `nix-shell -p nixpkgs-review --run "nixpkgs-review rev HEAD"`. Note: all changes have to be committed, also see [nixpkgs-review usage](https://github.com/Mic92/nixpkgs-review#usage)
- [X] Tested basic functionality of all binary files (usually in `./result/bin/`) ---> `./result/bin/spades.py --test`
- [24.11 Release Notes](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2411.section.md) (or backporting [23.11](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2311.section.md) and [24.05](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2405.section.md) Release notes)
  - [ ] (Package updates) Added a release notes entry if the change is major or breaking
  - [ ] (Module updates) Added a release notes entry if the change is significant
[...]

PR #324428, opened 2024-07-03
spades: mark broken for darwin

superherointj:
spades: mark broken for darwin

https://hydra.nixos.org/build/264908242/nixlog/1

https://github.com/NixOS/nixpkgs/pull/320502#issuecomment-2207402806

CC @bzizou

### DIFF
--- a/pkgs/by-name/sp/spades/package.nix
+++ b/pkgs/by-name/sp/spades/package.nix
@@ -69,6 +69,6 @@ stdenv.mkDerivation (finalAttrs: {
       "x86_64-darwin"
     ];
     maintainers = with lib.maintainers; [ bzizou ];
-    broken = stdenv.hostPlatform.isMusl;
+    broken = stdenv.hostPlatform.isMusl && stdenv.isDarwin;
   };
 })


wegank:
Oh, that's faster than my fix :-D

pkgs/by-name/sp/spades/package.nix
```diff
@@ -69,6 +69,6 @@ stdenv.mkDerivation (finalAttrs: {
       "x86_64-darwin"
     ];
     maintainers = with lib.maintainers; [ bzizou ];
-    broken = stdenv.hostPlatform.isMusl;
+    broken = stdenv.hostPlatform.isMusl && stdenv.isDarwin;
```
prusnak, line 72: This is wrong. It should have been `||` instead of `&&`

khaneliman:
> Oh, that's faster than my fix :-D

Lol 11 minute unreviewed speed merge

pkgs/by-name/sp/spades/package.nix
superherointj: That's why we have to rest well.

PR #320187, opened 2024-06-16
db4: fix configure with llvm

RossComputerGuy:
## Description of changes

Can be tested by building `pkgsLLVM.db4` (`nix-build -A pkgsLLVM.db4` / `nix build .#pkgsLLVM.db4`)

Fixes #315245

## Things done

- Built on platform(s)
  - [ ] x86_64-linux
  - [x] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
- For non-Linux: Is sandboxing enabled in `nix.conf`? (See [Nix manual](https://nixos.org/manual/nix/stable/command-ref/conf-file.html))
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:
  - [NixOS test(s)](https://nixos.org/manual/nixos/unstable/index.html#sec-nixos-tests) (look inside [nixos/tests](https://github.com/NixOS/nixpkgs/blob/master/nixos/tests))
  - and/or [package tests](https://github.com/NixOS/nixpkgs/blob/master/pkgs/README.md#package-tests)
  - or, for functions and "core" functionality, tests in [lib/tests](https://github.com/NixOS/nixpkgs/blob/master/lib/tests) or [pkgs/test](https://github.com/NixOS/nixpkgs/blob/master/pkgs/test)
  - made sure NixOS tests are [linked](https://nixos.org/manual/nixpkgs/unstable/#ssec-nixos-tests-linking) to the relevant packages
- [ ] Tested compilation of all packages that depend on this change using `nix-shell -p nixpkgs-review --run "nixpkgs-review rev HEAD"`. Note: all changes have to be committed, also see [nixpkgs-review usage](https://github.com/Mic92/nixpkgs-review#usage)
- [ ] Tested basic functionality of all binary files (usually in `./result/bin/`)
- [24.11 Release Notes](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2411.section.md) (or backporting [23.11](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2311.section.md) and [24.05](https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/release-notes/rl-2405.section.md) Release notes)
  - [ ] (Package updates) Added a release notes entry if the change is major or breaking
  - [ ] (Module updates) Added a release notes entry if the change is significant
[...]

PR #324471, opened 2024-07-04
db4: fix pthread configure detection

paparodeo:
## Description of changes

add header stdlib.h so pthread detection doesn't fail do to a missing function prototype for exit.
    
    configure:21863: clang -o conftest -O3   -D_GNU_SOURCE -D_REENTRANT
    conftest.c  >&5
    conftest.c:52:2: error: call to undeclared library function 'exit' with
    type 'void (int) __attribute__((noreturn))'; ISO C99 and later do not
    support implicit function declarations [-Wimplicit-function-declaration]
       52 |         exit (
          |         ^
    conftest.c:52:2: note: include the header <stdlib.h> or explicitly
    provide a declaration for 'exit'

additionally, add a leading space to line 218 to uncorrupt the patch -- git apply was complaining and failed to apply without fix.

this reverts the previous workaround https://github.com/NixOS/nixpkgs/pull/320187

testing: `nix-build -E 'with import ./.{}; db4.override { stdenv = clangStdenv; }` builds

## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/development/libraries/db/clang-4.8.patch
+++ b/pkgs/development/libraries/db/clang-4.8.patch
@@ -153,20 +153,40 @@ diff -ur a/dist/aclocal/clock.m4 b/dist/aclocal/clock.m4
 diff -ur a/dist/aclocal/mutex.m4 b/dist/aclocal/mutex.m4
 --- a/dist/aclocal/mutex.m4	1969-12-31 19:00:01.000000000 -0500
 +++ b/dist/aclocal/mutex.m4	2023-06-05 19:14:47.214158196 -0400
-@@ -4,6 +4,7 @@
+@@ -3,7 +3,9 @@
+ # POSIX pthreads tests: inter-process safe and intra-process only.
  AC_DEFUN(AM_PTHREADS_SHARED, [
  AC_TRY_RUN([
++#include <stdlib.h>
  #include <pthread.h>
 +int
  main() {
  	pthread_cond_t cond;
  	pthread_mutex_t mutex;
-@@ -46,6 +47,7 @@
+@@ -24,6 +26,7 @@ main() {
+ 	pthread_mutexattr_destroy(&mutexattr));
+ }], [db_cv_mutex="$1"],,
+ AC_TRY_LINK([
++#include <stdlib.h>
+ #include <pthread.h>],[
+ 	pthread_cond_t cond;
+ 	pthread_mutex_t mutex;
+@@ -45,7 +48,9 @@ AC_TRY_LINK([
+ ], [db_cv_mutex="$1"]))])
  AC_DEFUN(AM_PTHREADS_PRIVATE, [
  AC_TRY_RUN([
++#include <stdlib.h>
  #include <pthread.h>
 +int
  main() {
+ 	pthread_cond_t cond;
+ 	pthread_mutex_t mutex;
+@@ -64,6 +69,7 @@ main() {
+ 	pthread_mutexattr_destroy(&mutexattr));
+ }], [db_cv_mutex="$1"],,
+ AC_TRY_LINK([
++#include <stdlib.h>
+ #include <pthread.h>],[
  	pthread_cond_t cond;
  	pthread_mutex_t mutex;
 diff -ur a/dist/aclocal/sequence.m4 b/dist/aclocal/sequence.m4
@@ -183,7 +203,7 @@ diff -ur a/dist/aclocal/sequence.m4 b/dist/aclocal/sequence.m4
  			$db_cv_seq_type l;
  			unsigned $db_cv_seq_type u;
 @@ -59,7 +62,9 @@
-				return (1);
+ 				return (1);
  			return (0);
  		}],, [db_cv_build_sequence="no"],
 -		AC_TRY_LINK(,[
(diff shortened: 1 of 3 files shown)


paparodeo:
cc: @@drupol @RossComputerGuy

drupol:
Don't you think this patch should be submitted upstream instead ?

paparodeo:
> Don't you think this patch should be submitted upstream instead ?

it is fixed for db5, tho there are still a bunch of patches in nix for that version as well, one being that main needs to return an int. the patches for the various db versions are:
https://github.com/NixOS/nixpkgs/tree/master/pkgs/development/libraries/db

and if i untar the source for db4 the dates are all `Apr 12 2010` and the homepage url points to https://www.oracle.com/database/technologies/related/berkeleydb.html so not sure where to send a patch to.

so i get the feeling that this version is pretty old so am not sure where to upstream the patch.

drupol:
Fair enough.

pkgs/development/libraries/db/generic.nix
```diff
@@ -67,8 +67,7 @@ stdenv.mkDerivation (rec {
       (if compat185 then "--enable-compat185" else "--disable-compat185")
     ]
     ++ lib.optional dbmSupport "--enable-dbm"
-    ++ lib.optional stdenv.isFreeBSD "--with-pic"
-    ++ (drvArgs.configureFlags or []);
+    ++ lib.optional stdenv.isFreeBSD "--with-pic";
```
drupol, line 70:
```suggestion
    ++ lib.optionals stdenv.isFreeBSD [ "--with-pic" ];
```
Opinionated: I would use `lib.optionals` instead, since `configureFlags` is a list. Semantically, it makes more sense to me.

pkgs/development/libraries/db/clang-4.8.patch
```diff
@@ -183,7 +203,7 @@ diff -ur a/dist/aclocal/sequence.m4 b/dist/aclocal/sequence.m4
  			$db_cv_seq_type l;
  			unsigned $db_cv_seq_type u;
 @@ -59,7 +62,9 @@
-				return (1);
```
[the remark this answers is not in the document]
paparodeo: so i added a space to un-corrupt the diff. i guess generating a diff the first character is a space, + or -, except this line was missing the space and `git apply` as complaining about the patch being corrupted so i fixed it. i can revert it as patch doesn't seem to care.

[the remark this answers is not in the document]
drupol: OK don't, I was just curious.

pkgs/development/libraries/db/generic.nix
paparodeo: so this change is just a revert of the previous workaround for pkgsLLVM so i'd prefer to leave as is (it is the first commit). my change is the 2nd commit which modifies the patch. in general, i do try to prefer optionals over optional.

drupol: It's totally fine and optional, no worries.

paparodeo:
FWIW this is also needed to fix compiling with gcc14.

```
nix-build -E 'with import ./.{}; db4.override { stdenv = gcc14Stdenv; }'
```

fails the same way as if using `clangStdenv`

pkgs/development/libraries/db/clang-4.8.patch
emilazy, line 1: Maybe it would be better to split this out into a separate patch? Or at least rename this to something more generic, if it’s also tackling a modern LLVM thing and recent GCC?

paparodeo: it is the db 4.8 clang patch which fixes similar issues in the same file. eg: main not returning an int. i wouldn't be surprised if new gcc needs all the same fixes in this file too. and the naming matches the other autoconf patches for the db 5.3 and db 6.0.

i'm not quite sure what your asking for -- a re-organization of the package to rename the patches and test them all on new versions of gcc to split out the intersection of fixes needed by both gcc and clang to the patch can be named appropriately?

thanks for taking a look.

emilazy: Ah, sorry, I misinterpreted what the 4.8 was referring to. Something like `db-4.8-modern-compilers` or your `autoconf-modernization` suggestion or whatever seem fine to me.

paparodeo:
change seems fine to me as is and i don't want to put any more time into it so closing.

emilazy:
I’m happy to merge this as‐is if you wanted to reopen. Sorry for nitpicking about the filename when I lacked the full context of the patches.

emilazy (approved):
Successfully built `db4` on `x86_64-darwin` after rebasing this on `master` locally.

vobloeb:
`"--with-mutex=POSIX/pthreads";` used to fix GCC-14, but this PR fixes only Clang, so GCC-14 is broken again

PR #324543, opened 2024-07-04
haskellPackages.haskell-gi: Drop obsolete patches

rvl:
This is for merging into `haskell-updates` (PR #321948)

It should fix the build of `taffybar`.

1. `haskell-gi` doesn't need a patch any more.
2. While we are here, `gi-vte` also doesn't need a patch any more (_I think_ - builds pending).


maralorn:
Great, thank you!

Issue #324396, opened 2024-07-03
Update request: joplin-desktop 2.14.17 → 3.0.12


UlyssesZh:
- Package name: joplin-desktop
- Latest released version: 3.0.12
- Current version on the unstable channel: 2.14.17
- Current version on the stable/release channel: 2.14.17
- [x] Checked the [nixpkgs pull requests](https://github.com/NixOS/nixpkgs/pulls)

**Notify maintainers**
@HugoReeves @QJoly 

There is an auto-generated PR #295450 for upgrading to 3.0.2, but 3.0.2 is a pre-release. Note that 3.0.12 is the first 3.x version that is not a pre-release.

-----

Note for maintainers: Please tag this issue in your PR.

---

Add a :+1: [reaction] to [issues you find important].

[reaction]: https://github.blog/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/
[issues you find important]: https://github.com/NixOS/nixpkgs/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions-%2B1-desc

UlyssesZh:
By the way, why is it built from the downloaded appimage instead of built from source?

pmiddend:
I use this tool daily and unfortunately updated the DB to version 3 a little prematurely. Would be very cool to see this in nixpkgs asap.

yajo:
Android app got updated, and now I cannot sync with desktop app until updated too. Sadness.

deftdawg:
> Android app got updated, and now I cannot sync with desktop app until updated too. Sadness.

Same for iOS app, as a temporary workaround I'm using the appimage w/ the `appimage-run` package until an update gets merged.

```sh
curl -LO "https://github.com/laurent22/joplin/releases/download/v3.0.12/Joplin-3.0.12.AppImage"
appimage-run Joplin-3.0.12.appimage
```

Neon-44:
Was just having a look at it, updating it, and:

It already is updated? In the nixpkgs, it already is declared as version 3.0.12, but somehow the one it ends up installing is still 2.14.17?

This is a bit above my paygrade tbh, would love some help from some actual experts here.

I am using the unstable channel in my flake btw.

aanderse:
> Edit: got it, it was updated in the master channel 2 days ago, it should come to the unstable channel pretty soon I guess.

nailed it

specifically, you can easily see the most recently available commit at https://status.nixos.org/

UlyssesZh:
[...]

PR #324551, opened 2024-07-04
joplin-desktop: 2.14.17 -> 3.0.12

pmiddend:
## Things done

- Built on platform(s)
  - [x] x86_64-linux
  - [ ] aarch64-linux
  - [ ] x86_64-darwin
  - [ ] aarch64-darwin
  - [ ] `sandbox = relaxed`
  - [ ] `sandbox = true`
- [ ] Tested, as applicable:

---

### DIFF
--- a/pkgs/applications/misc/joplin-desktop/default.nix
+++ b/pkgs/applications/misc/joplin-desktop/default.nix
@@ -1,8 +1,8 @@
-{ lib, stdenv, appimageTools, fetchurl, makeWrapper, undmg }:
+{ lib, stdenv, appimageTools, fetchurl, makeWrapper, _7zz }:
 
 let
   pname = "joplin-desktop";
-  version = "2.14.17";
+  version = "3.0.12";
 
   inherit (stdenv.hostPlatform) system;
   throwSystem = throw "Unsupported system: ${system}";
@@ -16,9 +16,9 @@ let
   src = fetchurl {
     url = "https://github.com/laurent22/joplin/releases/download/v${version}/Joplin-${version}${suffix}";
     sha256 = {
-      x86_64-linux = "sha256-u4wEchyljurmwVZsRnmUBITZUR6SxDxyGczZjXNsJkg=";
-      x86_64-darwin = "sha256-KjNwAnJZGX/DvHDPw15vGlSbJ47s6YT59EalARt1mx4=";
-      aarch64-darwin = "sha256-OYpsHPI+7riMVNAp2JpBlmdFdJUSNqNvBmeYHDw6yzY=";
+      x86_64-linux = "sha256-vMz+ZeBHP+9Ugy8KO8lbp8zqC8VHtf1TWw10YytQFSs=";
+      x86_64-darwin = "sha256-XZN1jTv/FhJXuFxZ6D6h/vFMdKi84Z9UWfj2CrMgBBA=";
+      aarch64-darwin = "sha256-lsODOBkZ4+x5D6Er2/paTzAMKZvqIBVkKrWHh5iRvrk=";
     }.${system} or throwSystem;
   };
 
@@ -39,7 +39,7 @@ let
     homepage = "https://joplinapp.org";
     license = licenses.agpl3Plus;
     maintainers = with maintainers; [ hugoreeves qjoly ];
-    platforms = [ "x86_64-linux" "x86_64-darwin" "aarch64-darwin"];
+    platforms = [ "x86_64-linux" "x86_64-darwin" "aarch64-darwin" ];
   };
 
   linux = appimageTools.wrapType2 rec {
@@ -64,7 +64,7 @@ let
   darwin = stdenv.mkDerivation {
     inherit pname version src meta;
 
-    nativeBuildInputs = [ undmg ];
+    nativeBuildInputs = [ _7zz ];
 
     sourceRoot = "Joplin.app";
 


pmiddend:
@ofborg build joplin-desktop

pmiddend:
See #324396 - this is just my attempt at the update.

pkgs/applications/misc/joplin-desktop/default.nix
```diff
@@ -16,9 +16,9 @@ let
   src = fetchurl {
     url = "https://github.com/laurent22/joplin/releases/download/v${version}/Joplin-${version}${suffix}";
     sha256 = {
-      x86_64-linux = "sha256-u4wEchyljurmwVZsRnmUBITZUR6SxDxyGczZjXNsJkg=";
-      x86_64-darwin = "sha256-KjNwAnJZGX/DvHDPw15vGlSbJ47s6YT59EalARt1mx4=";
-      aarch64-darwin = "sha256-OYpsHPI+7riMVNAp2JpBlmdFdJUSNqNvBmeYHDw6yzY=";
+      x86_64-linux = "sha256-vMz+ZeBHP+9Ugy8KO8lbp8zqC8VHtf1TWw10YytQFSs=";
+      x86_64-darwin = "";
+      aarch64-darwin = "";
```
yajo, line 21:
```suggestion
      aarch64-darwin = "sha256-lsODOBkZ4+x5D6Er2/paTzAMKZvqIBVkKrWHh5iRvrk=";
```

yajo (changes requested):

pkgs/applications/misc/joplin-desktop/default.nix
pmiddend: Yup, already saw that. Now just waiting for `x86_64-darwin`, and ideally people with that architectures reporting back that it works for them, since I can only test `x86_64-linux`.

pmiddend:
Given that the mobile app automatically updates and leaves the database in a state that's unusable for users with version <3 - we _have_ to backport this update to 24.05, right?

aanderse:
Screenshot_20240705_112523

:sob:

i'm impacted by this and this is very detrimental to my workflow... (as a temporary workaround ... app image here we come?)

but i can't suggest that we force a major upgrade on users during the `24.05` cycle unless for security reasons... that doesn't seem right
so we can let users pull down an unstable channel/flake input on top of their stable channel/flake input, or, we can offer `joplin-desktop-2_4` and `joplin-desktop_3_0`...

not sure what we expect is the most desirable option here - what is upstream planning to do with `2.x` series? if they are entirely abandoning it then this is not great as `joplin` can be a security issue if you use it with any cloud stuff

of course some people can use it without any cloud stuff at all and then i guess they are probably ok to do whatever

many questions... :thinking:

pmiddend:
> Screenshot_20240705_112523
> 
> 😭
> 
> i'm impacted by this and this is very detrimental to my workflow... (as a temporary workaround ... app image here we come?)
> 
> but i can't suggest that we force a major upgrade on users during the `24.05` cycle unless for security reasons... that doesn't seem right so we can let users pull down an unstable channel/flake input on top of their stable channel/flake input, or, we can offer `joplin-desktop-2_4` and `joplin-desktop_3_0`...
> 
> not sure what we expect is the most desirable option here - what is upstream planning to do with `2.x` series? if they are entirely abandoning it then this is not great as `joplin` can be a security issue if you use it with any cloud stuff
> 
> of course some people can use it without any cloud stuff at all and then i guess they are probably ok to do whatever
> 
> many questions... 🤔

I feel you, and we should think about this. But there is a certain practicality to this. The damage, I suppose, for most users is already done, and I'm not sure if you can go back to a 2.x database easily to mitigate it.

Would be interesting to know how other distributions handle this.

newAM (approved):
Builds and runs on x86_64-linux, thanks for the update!

AshleyYakeley:
I think the proper solution is just to pick certain packages from unstable, even if all your other packages are from 24.05. It works fine, NixOS is designed with this sort of thing in mind.

pmiddend:
But how is it for firefox? That's always updated to the latest version in stable, or is it?

aanderse:
what i am trying to say is that all of our users except you in me, in theory, could use this program completely offline

this is an exaggerative statement meant to demonstrate a point - we would be forcing all of them into an unneeded upgrade

yajo (approved):
Please merge, it's a very important update! Thanks 

For 24.05 we could backport it as a `joplin_3` package

altsalt:
Still impacted by this and haven't yet bitten the bullet to update to the unstable version. That said, I figured I'd respond to a couple of the suggestions with concerns as this will likely not be the only package that faces this problematic pattern.

> I think the proper solution is just to pick certain packages from unstable, even if all your other packages are from 24.05. It works fine, NixOS is designed with this sort of thing in mind.

One of the problems with this solution is that you may forget to remove the unstable version and an update comes down the pipe which puts things out of sync again in a future stable release.

> For 24.05 we could backport it as a `joplin_3` package

This solution may have a similar problem. Migrating to a higher version needs to be taken into account by whoever is maintaining the future named-version packages.

---

p.s. Is there an accessible write-up on how packages get flagged for addition to stable and such? Figure that having a set of rules that can be pointed to, and then updated to handle these sorts of edge cases, would be an easy response in the future.

pmiddend:
laurent22 also released [2.14.23](https://github.com/laurent22/joplin/releases/tag/v2.14.23) with this enhancement

> Disable sync version check to allow compatibility between 2.14 and 3.0 for users who cannot upgrade

Though I'm not sure what that means.

jspiers:
The `_7zz` unarchiver fails with an `ERROR: Dangerous link path` during the unpackPhase on my `x86_64-darwin` system. I have submitted a pull request #361912 to revert to using `undmg` because that fixes the issue on my system. But thought I should give a heads up here in case that would break something on other platforms.

